PR #8, opened 2018-08-30
Added optional height and width to HTMLPanel

jbednar:
HTMLPanel works well, but it doesn't seem to allot any space for the actual object being shown:

```
import numpy as np, pandas as pd, datashader as ds
from datashader import transfer_functions as tf
from datashader.colors import inferno, viridis
from numba import jit
from colorcet import fire

@jit
def clifford_trajectory(a, b, c, d, x0, y0, n):
    xs, ys = np.zeros(n), np.zeros(n)
    xs[0], ys[0] = x0, y0
    for i in np.arange(n-1):
        xs[i+1] = np.sin(a * ys[i]) + c * np.cos(a * xs[i])
        ys[i+1] = np.sin(b * xs[i]) + d * np.cos(b * ys[i])
    return pd.DataFrame(dict(x=xs,y=ys))

def clifford_plot(a, b, c, d, n=10000000, cmap=fire[::-1]):
    cvs = ds.Canvas(plot_width=400, plot_height=400)
    agg = cvs.points(clifford_trajectory(a, b, c, d, 0, 0, n), 'x', 'y')
    return tf.shade(agg, cmap=cmap)
```

```
from panel import Row, Column
from panel.panels import HTMLPanel
import param

class DatashaderPanel(param.Parameterized):
    b = param.Number(1.7, bounds=(-2, 2))
    c = param.Number(0.6, bounds=(-2, 2))
    d = param.Number(1.2, bounds=(-2, 2))

    def view(self):
        return HTMLPanel(clifford_plot(self.a, self.b, self.c, self.d))
    
panel = DatashaderPanel()
Column(panel, panel.view)
```
image

This PR adds optional `height` and `width` parameters to HTMLPanel, allowing users to specify those if they wish:

image

If there is a better way to achieve this; let me know!

Note that we could consider moving the parameters up to the base class and the implementation to `_get_model()` in the base class so that `height` and `width` would always be allowed for any Panel object. I think doing that would let users situate their content in a box of their own choosing.  Maybe that's a better option than this? Probably would have to supply options for centering or justifying the content in that box, though.

### DIFF
--- a/panel/panels.py
+++ b/panel/panels.py
@@ -290,7 +290,9 @@ class ParamMethodPanel(PanelBase):
 
 class DivBasePanel(PanelBase):
     """
-    Baseclass for Panels which render HTML inside a bokeh div.
+    Baseclass for Panels which render HTML inside a Bokeh Div.
+    See the documentation for Bokeh Div for more detail about
+    the supported options like style and sizing_mode.
     """
 
     # DivPanel supports updates to the model
@@ -298,8 +300,20 @@ class DivBasePanel(PanelBase):
 
     __abstract = True
 
+    height = param.Integer(default=None, bounds=(0, None))
+
+    width = param.Integer(default=None, bounds=(0, None))
+
+    sizing_mode = param.ObjectSelector(default=None,
+        objects=["fixed", "scale_width", "scale_height", "scale_both", "stretch_both"], 
+        doc="How the item being displayed should size itself.")
+                                       
+    style = param.Parameter(default=None, doc="""
+        Raw CSS style declaration for this Div.""")
+
     def _get_properties(self):
-        return {}
+        return {p : getattr(self,p) for p in ["width", "height", "sizing_mode", "style"]
+                if getattr(self,p) is not None}
 
     def _get_model(self, doc, root=None, parent=None, comm=None, rerender=False):
         model = Div(**self._get_properties())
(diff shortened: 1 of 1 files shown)


philippjfr:
>Note that we could consider moving the parameters up to the base class and the implementation to _get_model() in the base class so that height and width would always be allowed for any Panel object. I think doing that would let users situate their content in a box of their own choosing

In some cases (e.g. matplotlib) the width/height can be computed from the object being wrapped but I agree that you should be able to override that. I don't mind moving the width/height parameters onto the ``DivBasePanel`` and then applying the explicitly set width/height if it's set.

jbednar:
Ok, I've implemented that, along with some other changes:

- Added Bokeh Div properties height, width, sizing_mode, and style.  These are the properties of a Bokeh Div that seem useful to change:
    *  height/width override anything computed for the object inside, if specified
    * style is a dictionary of CSS property:value pairs, e.g. style=dict(color="red"), applying to what's in the Div
    * sizing_mode may or may not be useful here; just exposes the underlying Bokeh functionality.
git commit -m 'Clarified definition of style parameter' panel/panels.py
- Renamed GGPlotPanel to RGGPlotPanel, since someday we could have support for Python's ggplot port.
- Changed RGGPlotPanel's dpi from 77 to 144 to make the fonts not be unusably small (unrelated change that I didn't want to make in another branch!)

panel/panels.py
```diff
@@ -290,16 +290,30 @@ def _cleanup(self, model):
...
+    height = param.Integer(default=None, bounds=(0, None))
+
+    width = param.Integer(default=None, bounds=(0, None))
+
+    sizing_mode = param.ObjectSelector(default=None,
+        objects=["fixed", "scale_width", "scale_height", "scale_both", "stretch_both"], 
+        doc="How the item being displayed should size itself.")
+                                       
+    style = param.Dict(default=None, doc="""
+        Dictionary of CSS property:value pairs to apply to this Div.""")
```
philippjfr: Great, I somehow missed this one entirely.

PR #9, opened 2018-08-30
Added PNGPanel as a base class and for _repr_png_

jbednar:
Previously, MatplotlibPanel and RGGPlotPanel had separate implementations for rendering PNGs into something Viewable.  This PR pulls that functionality into a new shared base class PNGPanel that makes those classes much simpler (Matplotlib support is now essentially just one call).  

PNGPanel isn't an abstract class, though; it directly renders anything with a `_repr_png_` method, which means that it supports PIL/Pillow objects, Datashader objects, and probably quite a lot of other objects that are designed to display themselves as PNGs in Jupyter.   

PNGPanel computes its size directly from the PNG, which means that it no longer needs an MPL-specific call and Datashader plots no longer need to specify their size (as it's in the PNG already).  So we can now do:

```
import numpy as np, pandas as pd, datashader as ds
from datashader import transfer_functions as tf
from colorcet import kbc
from numba import jit

@jit
def clifford_trajectory(a, b, c, d, x0, y0, n):
    xs, ys = np.zeros(n), np.zeros(n)
    xs[0], ys[0] = x0, y0
    for i in np.arange(n-1):
        xs[i+1] = np.sin(a * ys[i]) + c * np.cos(a * xs[i])
        ys[i+1] = np.sin(b * xs[i]) + d * np.cos(b * ys[i])
    return pd.DataFrame(dict(x=xs,y=ys))

def clifford_plot(a, b, c, d, n=10000000, cmap=kbc[::-1]):
    cvs = ds.Canvas(plot_width=400, plot_height=400)
    agg = cvs.points(clifford_trajectory(a, b, c, d, 0, 0, n), 'x', 'y')
    return tf.shade(agg, cmap=cmap)
```

```
from PIL import Image
logo = Image.open("../../datashader/doc/_static/logo_stacked.png").resize((200,150))
```

```
from panel import Row, Column
import param

class DatashaderPanel(param.Parameterized):
    b = param.Number(1.7, bounds=(-2, 2))
    c = param.Number(0.6, bounds=(-2, 2))
    d = param.Number(1.2, bounds=(-2, 2))

    def view(self):
        return clifford_plot(self.a, self.b, self.c, self.d)
panel = DatashaderPanel()
Row(Column(logo, panel), panel.view)
```

image

### DIFF
--- a/panel/panels.py
+++ b/panel/panels.py
@@ -327,7 +327,41 @@ class DivBasePanel(PanelBase):
         div.update(**self._get_properties())
 
 
-class MatplotlibPanel(DivBasePanel):
+class PNGPanel(DivBasePanel):
+    """
+    Encodes a PNG as base64 and wraps it in a Bokeh Div model.
+    This base class supports anything with a _repr_png_ method, but
+    subclasses can provide their own way of obtaining or generating a PNG.
+    """
+
+    @classmethod
+    def applies(cls, obj):
+        return hasattr(obj, '_repr_png_')
+
+    def _png(self):
+        return self.object._repr_png_()
+    
+    def _pngshape(self, data):
+        """Calculate and return PNG width,height"""
+        import struct
+        w, h = struct.unpack('>LL', data[16:24])
+        return int(w), int(h)
+    
+    def _get_properties(self):
+        data = self._png()
+        b64 = base64.b64encode(data).decode("utf-8")
+        src = "data:image/png;base64,{b64}".format(b64=b64)
+        html = "<img src='{src}'></img>".format(src=src)
+        
+        p = super(PNGPanel,self)._get_properties()
+        width, height = self._pngshape(data)
+        if self.width  is None: p["width"]  = width
+        if self.height is None: p["height"] = height
+        p["text"]=html
+        return p
+
+
+class MatplotlibPanel(PNGPanel):
     """
     A MatplotlibPanel renders a matplotlib figure to png and wraps
     the base64 encoded data in a bokeh Div model.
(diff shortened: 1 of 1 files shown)


panel/panels.py
```diff
@@ -327,32 +327,57 @@ def _update(self, model):
...
         b64 = base64.b64encode(data).decode("utf-8")
         src = "data:image/png;base64,{b64}".format(b64=b64)
-        width, height = self.object.canvas.get_width_height()
         html = "<img src='{src}'></img>".format(src=src)
         
-        p = super(MatplotlibPanel,self)._get_properties()
+        p = super(PNGPanel,self)._get_properties()
+        width, height = self._pngshape(data)
         if self.width  is None: p["width"]  = width
         if self.height is None: p["height"] = height
```
philippjfr: This should be:

```python
p['width'] = self.width or width
p['height'] = self.height or height
```

The width and height should always be provided.

philippjfr:
Looks great, the ``struct`` trick to get the png width and height is really neat.

panel/panels.py
philippjfr: Nevermind it's fine I see the superclass defines it.

jbednar: Shouldn't it be possible for the user to set self.width or self.height to zero, if they want to?  The proposed replacement code doesn't allow that.

jbednar: (Since self.width and self.height are only for user overriding of the computed width,height).

jbednar: Oh, I see; you mean that `p["width"]` and `p["height"]` should always be provided.  So I guess it should be:

```
p["width"]  = width  if self.width  is None else self.width
p["height"] = height if self.height is None else self.height
```

jbednar: Oops; I never saw your "nevermind"; somehow the PR page updated with only your first comment.

PR #11, opened 2018-08-31
Added logo

jbednar:
May not be the most exciting logo ever, but is something to start with.

image

I couldn't see how to make the wordmark be bold and have tighter spacing in Inkscape. The Ubuntu font in gimp supports both, but that would be a bitmap...


jlstevens:
A good start. Maybe you don't actually need the lower green rectangle?

jbednar:
I think disliking that combination of colors is agreed by all, then, including me.  It's tricky; when it's a single color in the logo, it looks odd to have the same color in the wordmark, but I didn't find a good color to go with that green.  I was trying to have the logo be squarish to match all the other PyViz logos, but I agree it will probably look better with just the top two items.  I'll make a quick pass and see if I can get something nicer.

philippjfr:
What do you think of?

panel_logo

philippjfr:
Could you send me your processed wordmark?

jlstevens:
I like Philipp's logo more and Jim's wordmark so yes, that should be a good combination! Still not entirely sure about the color though (it stands out as being the greyscale one out of a bunch of colorful logos)

jlstevens:
I also think the line in the line graph could be a bit thicker...

jbednar:
I'm fine with leaving the final choice to Jean-Luc as an independent party, but note that Bo votes for Philipp's over mine. :-)  I agree that having everything look gray is odd, and that the graph line should be as thick as the slider lines.

Note that neither of these two versions will work well as a favicon, but that doesn't have to be a major concern.

philippjfr:
I've been playing with colors a bit, I'll try to pick a couple and then we can vote. I agree on both the wordmark and the line thickness. For the favicon I was thinking of just using the widget panel.

jlstevens:
I think just the widget panel for the favicon is fine. It is different enough from param by using negative space rather than positive space (it looks like the sliders are a 'hole' in the grey background).

philippjfr:
Any opinions on color?

panel_logo_colors

jlstevens:
Green: Not bad.
Grey: Doesn't fit with the colorful logos
Dark blue: I like it but we have a lot of blue across our logos already.
Purple/Red: Not bad
Blue/Cyan: Similar shade to the numba logo.
Orange: We already have a fair bit of orange across our logos.

So I vote for either the green or purple.

jbednar:
I vote for the dark blue or for the green, though the green is a touch too bright for my taste.  Bo votes for green.

philippjfr:
My vote is also green. I tried playing with a few other shades of green but the ones I liked were either made it look too much like excel or too close to geoviews' shades of green.

jlstevens:
One thought: the green is good for the logo but maybe the text can be a different color? A number of our other logos do that...

jbednar:
Ok, can you please push the logo_stacked.svg and I'll generate the rest?

jbednar:
If the text is a different color, I think it should be black; the dark blue is similar to black but not quite close enough, and people expect black in that case.

philippjfr:
Here's the black:

logo_black

And for comparison here's the green:

logo_green

jbednar:
I think the two-color version looks better.

philippjfr:
Generated all the files and pushed them.

philippjfr:
Should I merge?

Issue #19, opened 2018-09-04
Centering difference between panel and raw bokeh models


jlstevens:
This issue seems to happen when using templates. Here is a small app composed of three files to demonstrate the issue by toggling ``self.use_panel`` in ``CenteringApp`` (a zip of the three files can be found [here](https://github.com/pyviz/panel/files/2349367/Archive.zip)):

Run the app with ``python launch.py``:

``launch.py``:

```python
from dashboard import CenteringApp

if __name__ == '__main__':
    from bokeh.server.server import Server
    server = Server({
        '/center': CenteringApp(name = 'Weird centering')()})
    server.start()

    server.io_loop.add_callback(server.show, "/center")
    server.io_loop.start()

else:
    print("Run dashboard using 'python launch.py'")
```

``dashboard.py`` (note holoviews isn't actually relevant here and could be stripped out too):

```python
import panel
import os
import param

from jinja2 import Environment, FileSystemLoader

from bokeh.application.handlers import FunctionHandler
from bokeh.application import Application
from bokeh.models import Div

import holoviews as hv

renderer = hv.renderer('bokeh')
renderer.mode='server'

class App(object):
    """
    Mixin class to handle boilerplate necessary to turn holoviews
    objects into Bokeh applications.
    """

    name = 'Dashboard'

    def load_template(self, **params):
        appindex = os.path.join('.', 'index.html')
        env = Environment(loader=FileSystemLoader(os.path.dirname(appindex)))
        self.template = env.get_template('index.html')

    def apply_theming(self, doc):
        doc._template = self.template

    def doc_handler(self, doc):
        """
        Callback passed to the Bokeh FunctionHandler.
        """
        viewable = self.viewable(doc)
        if isinstance(viewable, hv.Dimensioned):
            doc = renderer.server_doc(viewable, doc=doc)
        else:
            doc.add_root(viewable)

        doc.title = self.name
        return doc

    def __call__(self):
        """
        Call to return a Bokeh Application to serve.
        """
        return Application(FunctionHandler(self.doc_handler))


class CenteringApp(param.Parameterized, App):

    def __init__(self, **params):
        super(CenteringApp, self).__init__(**params)
        self.load_template()

        self.use_panel = False # Change this
        text = "Should be in the center" * 8
        self.text = "<div style='text-align:left'>{text}.</div>".format(text=text)

    def viewable(self, doc=None):
        self.apply_theming(doc)
        if self.use_panel:
[...]

PR #22, opened 2018-09-04
Wrap raw bokeh model widgets in WidgetBox

philippjfr:
When raw bokeh widgets (of which ``Div`` is one example) are laid out in a Row or Column some strange sizing issues can occur. This can be prevented by wrapping these models in a `WidgetBox`, which is what I've done here.

This additional nesting is quite annoying, particularly as you can see when looking at the tests. The layout work that's currently happening in bokeh promises to get rid of WidgetBox types, which hopefully means all this mess will disappear and the majority of this PR can be reverted once that happens.

### DIFF
--- a/panel/pane.py
+++ b/panel/pane.py
@@ -12,8 +12,8 @@ from io import BytesIO
 
 import param
 
-from bokeh.layouts import Row as _BkRow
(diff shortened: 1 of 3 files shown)


jlstevens:
I can confirm that this PR fixes the example given in issue #19...

PR #27, opened 2018-09-05
Generalized PNG pane for any image-based mimetype and added jpg and gif

jbednar:
The only bits of the PNG pane that were specific to PNGs were the characters "png" and the code for determining the filesize, so I made those overridable at the class level and added implementations for JPG and GIF.

The code for JPG filesizes is a bit complex, so instead of these changes, we could consider adding a pillow-based filesize computation shared across all the Image classes; I have no strong preference.

The resulting images work well in a Column:

```
import panel as pp
pp.extension()

pp.Column("http://pyviz.org/_static/pyviz-logo.png",
          "http://homepages.inf.ed.ac.uk/jbednar/images/20051029_03cx083_[Calton_Hill]_crop.jpg",
          "http://homepages.inf.ed.ac.uk/jbednar/images/NeuralComputation12_7.gif",
          "http://homepages.inf.ed.ac.uk/jbednar/images/981109_or_map_2048MB.020000.spinanim_icon.gif")
```

(where the final image would be animated if this weren't a screenshot).

They don't always work well in a Row, sometimes overlapping and sometimes reserving too much space:

```
pp.Row("http://pyviz.org/_static/pyviz-logo.png",
       "http://homepages.inf.ed.ac.uk/jbednar/images/20051029_03cx083_[Calton_Hill]_crop.jpg",
       "http://homepages.inf.ed.ac.uk/jbednar/images/NeuralComputation12_7.gif"
       "http://homepages.inf.ed.ac.uk/jbednar/images/981109_or_map_2048MB.020000.spinanim_icon.gif")
```

image

However, I don't think that's anything to do with this PR, because for the previously supported (PNG) type I get the same overlapping behavior for master and for this branch's code:

```
import panel as pp
pp.extension()

pp.Row("http://pyviz.org/_static/pyviz-logo.png",
       "http://pyviz.org/_static/pyviz-logo.png",
       "http://pyviz.org/_static/pyviz-logo.png")
```
image

I would guess that's a Bokeh layout issue that's being addressed separately?

### DIFF
--- a/panel/pane.py
+++ b/panel/pane.py
@@ -349,25 +349,34 @@ class DivPaneBase(PaneBase):
         div.update(**self._get_properties())
 
 
-class PNG(DivPaneBase):
+class Image(DivPaneBase):
     """
-    Encodes a PNG as base64 and wraps it in a Bokeh Div model.  This
-    base class supports anything with a _repr_png_ method, a local
-    file with a png file extension or a HTTP url, but subclasses can
+    Encodes an image as base64 and wraps it in a Bokeh Div model.  
+    This is an abstract base class that needs the image type
+    to be specified and specific code for determining the image shape.
+    
+    The imgtype determines the filetype, extension, and MIME type for
+    this image. Each image type (png,jpg,gif) has a base class that
+    supports anything with a `_repr_X_` method (where X is `png`,
+    `gif`, etc.), a local file with the given file extension, or a
+    HTTP(S) url with the given extension.  Subclasses of each type can
     provide their own way of obtaining or generating a PNG.
     """
 
+    imgtype = 'None'
+
     @classmethod
     def applies(cls, obj):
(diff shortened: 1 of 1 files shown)


panel/pane.py
```diff
@@ -376,29 +385,69 @@ def _png(self):
...
+        import struct
+        w, h = struct.unpack("<HH", data[6:10])
+        return int(w), int(h)
+        
+
+class JPG(Image):
+
+    imgtype = 'jpg'
+    
+    def _imgshape(self, data):
```
philippjfr: I'd make these classmethods and then add simple tests by simply embedding some bytestrings for 1x1 pixel images.

jbednar: Ok, done.  Ready for review/merge.

jbednar: (I used a 2x1 image, as it's testing w,h and it's important not to get those two swapped.)

PR #28, opened 2018-09-06
Added yt support

jbednar:
The [yt](https://yt-project.org/) library uses Matplotlib to render one or more plots per object to PNG.  Although yt objects have a `_repr_html_` and can thus be displayed without any special support, the enclosing Bokeh Div does not resize appropriately to the size of the contents, which is very inconvenient.  This PR adds a pane.YT class that is the same as pane.HTML except that it calculates the total height (of all stacked PNGs) and the maximum width (of all stacked PNGs) and uses those to size the Div.  This implementation is somewhat inefficient, in that it renders the images to PNG just to get their sizes, but doing so seemed simpler than trying to parse them out of the HTML representation to find out their sizes.  Example:

```
try:
    import yt
except:
    !conda install -c conda-forge yt

import os
file = "enzo_tiny_cosmology"
if not os.path.exists(file):
    !curl -sSO http://yt-project.org/data/$file.tar.gz
    !tar xzf $file.tar.gz

import yt

import panel as pp
from panel import pane
pp.extension()

ds = yt.load("enzo_tiny_cosmology/DD0046/DD0046")
p = yt.ProjectionPlot(ds, "z", ["density", "temperature"], weight_field="density")

pp.Row(p)
```

image
image

I don't know how to get a tiny YT data file in order to add more official tests than this.

### DIFF
--- a/panel/pane.py
+++ b/panel/pane.py
@@ -540,3 +540,38 @@ class Str(DivPaneBase):
     def _get_properties(self):
         properties = super(Str, self)._get_properties()
         return dict(properties, text='<pre>'+escape(str(self.object))+'</pre>')
+
+
+
+class YT(HTML):
+    """
+    YT panes wrap plottable objects from the YT library.  
+    By default, the height and width are calculated by summing all
+    contained plots, but can optionally be specified explicitly to
+    provide additional space.
+    """
+
+    precedence = 0.5
+
+    @classmethod
+    def applies(cls, obj):
+        return ('yt' in repr(obj) and
+                hasattr(obj, "plots") and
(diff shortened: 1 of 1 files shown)


jbednar:
Given that a yt object can return multiple plots and that stacking them as above could be confusing, one could consider returning a pp.Column object instead of a single pane, which would be trivial to implement given the code in this PR that already effectively generates a list of PNGs.  But that would be a new behavior for Pane, so I'll leave such an implementation up to @philippjfr if he prefers it.  I think yt users can simply define only a single plot per yt object if they want to use the current version in a more reasonable way.

PR #31, opened 2018-09-06
Rough implementation of GridSpec

philippjfr:
Implements a GridSpec layout which allows declaring an NxM grid and assigning objects to the grid cells. Here is an example of a 3x3 grid:

screen shot 2018-09-06 at 6 33 17 pm

- [ ] Invert indexing coordinates (rows then columns)
- [ ] Improve width/height setting logic
- [ ] Fix vertical margin issues
- [ ] Add unit tests
- [ ] Add documentation

### DIFF
--- a/panel/layout.py
+++ b/panel/layout.py
@@ -6,12 +6,13 @@ from __future__ import absolute_import
 
 import param
 
+import numpy as np
 from bokeh.layouts import (Column as BkColumn, Row as BkRow,
                            WidgetBox as BkWidgetBox, Spacer as BkSpacer)
 from bokeh.models.widgets import Tabs as BkTabs, Panel as BkPanel
 
 from .pane import Pane, PaneBase
-from .util import push
+from .util import push, is_rectangle
 from .viewable import Reactive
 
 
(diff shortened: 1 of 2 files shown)


jbednar:
Wow, you are on a roll!

jbednar:
What happens when a grid cell is empty?

jbednar:
Would it be possible for the resulting grid to span the entire page area available (in a server context) or the entire width available (in a notebook context), and to responsively resize, with each item in a grid cell centered by default?  That would be an immediate decent-looking and usable dashboard with very little work for users.

philippjfr:
>What happens when a grid cell is empty?

screen shot 2018-09-06 at 10 35 08 pm

>Would it be possible for the resulting grid to span the entire page area available (in a server context) or the entire width available (in a notebook context), and to responsively resize, with each item in a grid cell centered by default?

I don't think this will work right now, but the new widget work at least theoretically makes that possible if all the other issues can be resolved.

jbednar:
Sounds good, thanks.

philippjfr:
Thanks, I tried the bokeh ``GridSpec`` initially but it didn't do quite what we needed. That said I can't recall why, so I should go back before this is merged and compare/contrast the two implementations.

mattpap:
`GridSpec` was added for a very particular task in one of bokeh's examples (removed at this point), so refining it is an option, especially there are no tests for it and it wasn't every publicized properly.

philippjfr:
Okay, I remember now. One of the main usecases for a `GridSpec` is to assign individual plots to multiple grid cells, the implementation currently in bokeh does not seem to allow for that. When assigning to multiple cells at once you have to provide a list of plots, so really it's just a different way of building the grid yourself and works nothing like matplotlib's `GridSpec`.

jbednar:
Sounds like once panel's GridSpec is solid, Bokeh's GridSpec should either be removed with a pointer to use Panel instead, or else panel's GridSpec should be split into a Bokeh-only implementation with a panel-specific extension on top (i.e. one that understands Panel objects, not just Bokeh models).

philippjfr:
We should definitely consider that. One reason why it might be weird to do that is that this implementation of GridSpec combines both layout and sizing, i.e. inserting an object into the grid defines the size of the plot. This is what makes it possible and what it means to assign a plot/model to multiple grid cells. It may be weird to mess with the sizes of bokeh models directly like that if this were added to bokeh itself, but I'd be happy to file an issue on bokeh to start that discussion.

PR #37, opened 2018-09-14
Minor packaging fixups

ceball:
Some things I just happened to notice:

  * fix summary on anaconda.org (is currently 'Bidirectional communication for PyViz')
  * include ts files in sdist (currently they're missed, although it doesn't really matter yet since we don't yet put a package on pypi)


philippjfr:
Thanks!

PR #46, opened 2018-09-19
Made a start on a Panes user guide

philippjfr:


stonebig:
do you still need that testpath<0.4 ?

stonebig:
do you still need that testpath<0.4 ?

jbednar:
I don't think so; we were able to remove it from datashader without any trouble just now.

PR #47, opened 2018-09-19
Add an implementation of interact

philippjfr:
Made a start on a ``interact`` implementation copying the API from ipywidgets. Large amounts of the implementation are copied directly from ipywidgets.

To do:

- [x] Validate function signature against parameters
- [ ] Capture ``print`` and ``display`` output and put it into display box?!
- [x] Add unit tests
- [x] Add documentation

### DIFF
--- a/panel/interact.py
+++ b/panel/interact.py
@@ -107,8 +107,6 @@ def _matches(o, pattern):
 
 class interactive(PaneBase):
 
-    auto_display = param.Boolean(default=False)
-
     layout = param.ClassSelector(default=Column, class_=Layout, is_instance=False)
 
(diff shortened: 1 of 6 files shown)


panel/interact.py
```diff
@@ -0,0 +1,216 @@
...
+    """Match a pattern of types in a sequence."""
+    if not len(o) == len(pattern):
+        return False
+    comps = zip(o,pattern)
+    return all(isinstance(obj,kind) for obj,kind in comps)
+
+
+class interact(PaneBase):
+
+    layout = param.ClassSelector(default=Row, class_=Layout, is_instance=False)
```
jbednar: I vote for default=Column for interact.

PR #50, opened 2018-09-20
Added support for selecting lists and dicts

jbednar:
Selector widgets currently need to do a reverse lookup on object identity, which requires that objects be hashable, which Bokeh palettes (as lists) are not.  This PR and a [matching PR on param](https://github.com/ioam/param/pull/268) converts lists to tuples when doing lookups, which works:

```
import panel as pp
from colorcet import kbc, fire
pp.extension()

def fn(a=1.7, cmap=kbc):
    return a,cmap[0]

pp.interact(fn, cmap=[kbc,fire])
```
image

```
import param

class C(param.Parameterized):
    cmap = param.ObjectSelector(kbc,[kbc,fire])

    def view(self):
        return self.a, self.cmap[0]
c=C()
pp.Row(c.param,c.view)
```

image

However, note that in both cases the representation of the list is vastly larger than the widget box, which in the second example ends before `(2, '#000000')` starts.  @philippjfr, is there a way to truncate the representation in the widget when it's longer than the box?

For a Param-based widget the solution is to provide string names, but I'm not sure how to do that with `interact`:

```
class C(param.Parameterized):
    cmap = param.ObjectSelector(kbc,dict(kbc=kbc,fire=fire))

    def view(self):
        return self.a, self.cmap[0]
c=C()
pp.Row(c.param,c.view)
```

image

### DIFF
--- a/panel/param.py
+++ b/panel/param.py
@@ -160,7 +160,7 @@ class Param(PaneBase):
             kw['button_type'] = 'primary'
             kw['name'] = value.name
 
-        if hasattr(p_obj, 'get_range') and not isinstance(kw['value'], dict):
+        if hasattr(p_obj, 'get_range'):
             kw['options'] = p_obj.get_range()
 
         if hasattr(p_obj, 'get_soft_bounds'):
--- a/panel/tests/test_widgets.py
+++ b/panel/tests/test_widgets.py
@@ -168,7 +168,8 @@ def test_select_list_constructor():
 
 
 def test_select(document, comm):
-    select = Select(options={'A': 'A', '1': 1}, value=1, name='Select')
+    opts = {'A': 'a', '1': 1}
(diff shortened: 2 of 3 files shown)


philippjfr:
Looks good. Want to add a unit test? :smile:

jbednar:
> For a Param-based widget the solution is to provide string names, but I'm not sure how to do that with interact

Actually, the same solution works with interact too:
```
import panel as pp
from colorcet import kbc, fire
pp.extension()

def fn(a=1.7, cmap=kbc):
    return a,cmap[0]

pp.interact(fn, cmap=dict(kbc=kbc,fire=fire))
```
image

So it is still an error that the widget extends outside the widget box, but at least there's a workaround that's arguably the right thing to do anyway (use string names for any complex objects).

PR #55, opened 2018-09-21
Add control over Matplotlib pane dpi

philippjfr:
Adds a dpi parameter that scales the resolution but not the displayed size of the image.


jbednar:
Works for me, thanks!

PR #57, opened 2018-09-21
Switch to pyviz main channel for pyctdev on CI

ceball:
Means panel will keep using the same version of pyctdev that it's using currently, rather than some potentially disruptive dev releases before the next main release.


ceball:
The test failure appears to be unrelated, so I think this could be merged.

PR #64, opened 2018-09-27
Allow selecting between parameterized subobjects

philippjfr:
- [x] Implements https://github.com/pyviz/panel/issues/68
- [x] Implements https://github.com/pyviz/panel/issues/69 (``toggleable_subobjects`` and ``expand_by_default``)
- [x] Implements https://github.com/pyviz/panel/issues/70
- [x] Adds unit tests


panel/param.py
```diff
@@ -95,9 +96,9 @@ def __init__(self, object, **params):
             params['name'] = object.name
         super(Param, self).__init__(object, **params)
         self._widgets = self._get_widgets()
-        self._widget_box = WidgetBox(*self._widgets.values(), height=self.height,
+        widgets = [widget for widgets in self._widgets.values() for widget in widgets]
+        self._widget_box = WidgetBox(*widgets, height=self.height,
                                      width=self.width, name=self.name)
-
         kwargs = {'name': self.name}
         if self.subpanel_layout is Tabs:
```
jbednar: Should "subpanel" still be appearing in the source code?  I would think it should say "pane" or "viewable"?

panel/param.py
```diff
@@ -260,6 +293,91 @@ def _get_root(self, doc, comm=None):
 
 
 
+class ParamMethod(PaneBase):
+    """
+    ParamMethod panes wrap methods annotated with the param.depends
+    decorator and rerenders the plot when any of the methods parameters
```
jbednar: It's not required for them to be annotated, right?  (Given that the default behavior is to watch all parameters.)

jbednar: "methods" -> "method's"

panel/param.py
philippjfr: Neither seem very descriptive, it's specifically about the layout for the parameterized subobjects. I guess `subobject_layout` would work.

jbednar: That's better for this case, thanks.

philippjfr:
I think there's some error on unwatch I haven't yet handled.

jbednar:
Both new options seem to work well.  I did have one problem, which is that I have a parameter that is defined purely to get a subobject pane (with only one value; `constant=True`), and I can't make the parameter not be displayed.  I can set its precedence to make it go away, but then the corresponding subobject param pane also goes away.  Maybe that can be a special case -- If `expand_by_default=True` and there's a constant parameter that would be expanded, don't show it in the main list of parameters (as it then has no reason to be shown, given that its values are now available on screen anyway and it can't be edited or expanded)?  It's an annoying thing to have to do and tricky, but I don't see how else to make it work well in this case.

jbednar:
> It's an annoying thing to have to do and tricky, but I don't see how else to make it work well in this case.

I don't think we need to do this -- I was wanting to see the parameters of the constant subobject to get around https://github.com/ioam/param/issues/287 , but instead I can use a readonly subobject and thereby put a pane for editing its widgets wherever I like.  So the current behavior for hidden parameters (also hiding their subobject parameters) is fine.

PR #65, opened 2018-09-29
Add Player widget

philippjfr:
Needs some further fixes including debouncing for the periodic events.


jbednar:
Cool.  Be sure to test it on Windows; we've had issues with these unicode characters there.

PR #85, opened 2018-10-08
Added SVG Pane

philippjfr:
Adds an SVG pane, unlike other image types the size and/or aspect ratio of the SVG cannot be easily computed ahead of time, but it still seems default to its native size unless explicitly overridden.

screen shot 2018-10-08 at 2 13 10 am

### DIFF
--- a/panel/pane.py
+++ b/panel/pane.py
@@ -488,7 +488,7 @@ class SVG(Image):
     def _img(self):
         if (isinstance(self.object, basestring) and
             self.object.lstrip().startswith('<svg')):
-            return self.object.encode('utf-8')
+            return self.object
         return super(SVG, self)._img()
 
     def _imgshape(self, data):
@@ -498,6 +498,8 @@ class SVG(Image):
         p = super(Image, self)._get_properties()
         data = self._img()
         width, height = self._imgshape(data)
+        if not isinstance(data, bytes):
+            data = data.encode('utf-8')
         b64 = base64.b64encode(data).decode("utf-8")
         src = "data:image/svg+xml;base64,{b64}".format(b64=b64)
         html = "<img src='{src}' width={width} height={height}></img>".format(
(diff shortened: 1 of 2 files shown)


panel/pane.py
```diff
@@ -476,6 +476,38 @@ def _imgshape(cls, data):
...
 
 
+class SVG(Image):
+
+    imgtype = 'svg'
+
+    @classmethod
+    def applies(cls, obj):
+        return (super(SVG, cls).applies(obj) or
+                (isinstance(obj, basestring) and obj.lstrip().startswith('<svg')))
```
jbednar: What does `(super(SVG, cls))` do here?

jbednar:
Have you tested networkx, which generates graphviz objects that have `_repr_svg_`?

philippjfr:
I tested shapely which has a `_repr_svg_`, don't think networkx will be any different.

panel/pane.py
philippjfr: Adds support for all the usual Image types, e.g. local SVG files, remote SVG files and objects with a `_repr_svg_`.

jbednar: Oh, I see now by looking at the superclass, which I think I wrote; it uses the `imgtype` to allow it to do general checking but specialized for this specific type. :-)  Makes sense now.

jbednar:
Graphviz objects work, but they don't seem to default to their native size as promised:

image

philippjfr:
Not sure there is anything we can do about it tbh, there is simply no reliable way to determine the  the size or aspect ratio of the svg ahead of time.

jbednar:
Not sure what you mean by "ahead of time"; the SVG is static once the Pane is instantiated in this case, right?

philippjfr:
By ahead of time I mean before it is rendered into the browser's DOM.

jbednar:
Ah.  There might be some Python libraries that could at least return the aspect ratio without rendering it, and probably a bounding box, though of course the browser might not respect that.  But not something to work on now...

PR #86, opened 2018-10-08
Add Markdown pane

philippjfr:
The ``Markdown`` pane renders to HTML using the markdown python library. It has a precedence of None which means it will never be used automatically, therefore a user has to explicitly wrap their markdown formatted text in a ``Markdown`` pane.

screen shot 2018-10-08 at 2 30 29 am

### DIFF
--- a/panel/pane.py
+++ b/panel/pane.py
@@ -21,8 +21,7 @@ import param
 from bokeh.layouts import Row as _BkRow, WidgetBox as _BkWidgetBox
 from bokeh.models import LayoutDOM, CustomJS, Widget as _BkWidget, Div as _BkDiv
 
-from .util import (Div, basestring, unicode, get_method_owner, push,
-                   remove_root)
+from .util import Div, basestring, get_method_owner, push, remove_root
 from .viewable import Reactive, Viewable
 
 
--- a/panel/pane.py
+++ b/panel/pane.py
@@ -17,6 +17,7 @@ except:
     from cgi import escape
 
 import param
+import markdown
 
 from bokeh.layouts import Row as _BkRow, WidgetBox as _BkWidgetBox
 from bokeh.models import LayoutDOM, CustomJS, Widget as _BkWidget, Div as _BkDiv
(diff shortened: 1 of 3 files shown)


philippjfr:
>Which markdown library is this? I don't see any new entry in setup.py for a library, and nothing in the stdlib.

I haven't added tests yet so I suspect it's just not running the example yet. It's based on the python markdown library available [on conda](https://anaconda.org/anaconda/markdown). I also already had it installed in my env, which seemed like a good sign.

>Isn't it a problem that there is no precedence value, for when an object has _repr_markdown_? It seems like that's unambiguous, and so should be able to be selected automatically.

Agreed, it's problematic. What I could imagine that instead of a simple True/False boolean `applies` could support returning a precedence value that depends on the type of the object.

>Doesn't Markdown just pass through HTML anyway? It seems to, and if so I think we should make Markdown be the default Pane type to handle strings, requiring HTML to be the one that has to be specified explicitly. I'd much rather be using Markdown strings than HTML strings in any case I can think of.

Seems sensible.

jbednar:
Ok, sounds great!  I'm ready to start using Markdown and almost never HTML!

philippjfr:
>I'm ready to start using Markdown and almost never HTML!

At least in the simple cases HTML embedded in Markdown will still render the same way.

philippjfr:
I guess the question is whether we should make `markdown` a hard dependency. If not what is the alternative fallback?

jbednar:
Oh, good point.  It doesn't look like the markdown library has any dependencies and it's available for py27 and py3, so it seems like a safe dependency to have.  And I'd like to be able to use it throughout our examples and suggested workflow, so I'm ok with making it a dependency.

philippjfr:
>Oh, good point. It doesn't look like the markdown library has any dependencies and it's available for py27 and py3, so it seems like a safe dependency to have.

Agreed, I also think it's pretty lightweight so I'm okay with requiring it.

PR #90, opened 2018-10-09
Added PNG/Matplotlib-based LaTeX pane

jbednar:
Accepts LaTeX strings and objects with `_repr_latex_`, such as `IPython.display.Latex`:

```
import panel as pn
from panel import pane
from IPython.display import Latex
pn.Row(pane.LaTeX(r'$\frac{x}{y^2}$'), Latex(r'$\frac{x}{y^2}$'))
```
image

Uses Matplotlib to render to a PNG and Pillow to crop it and make it transparent.  Could also consider using MathJax to create true (non-image) equations, but that would require a custom Bokeh model to get the JavaScript required.

### DIFF
--- a/panel/pane.py
+++ b/panel/pane.py
@@ -560,7 +560,7 @@ class Matplotlib(PNG):
         return b.getvalue()
 
 
-def latex_to_img(text, size=25):
+def latex_to_img(text, size=25, dpi=100):
     """
     Returns PIL image for LaTeX equation text, using matplotlib's rendering.
     Usage: latex_to_img(r'$\frac(x}{y^2}$')
@@ -571,12 +571,14 @@ def latex_to_img(text, size=25):
     import io
     
     buf = io.BytesIO()
-    plt.rc('text', usetex=True)
-    plt.rc('font', family='serif')
-    plt.axis('off')
-    plt.text(0.05, 0.5, '{text}'.format(text=text), size=size)
-    plt.savefig(buf, format='png')
-    plt.close()
+    with plt.rc_context({'text.usetex': True, 'font.family': 'serif'}):
+        fig = plt.figure()
+        ax = fig.add_subplot(111)
+        ax.axis('off')
+        ax.text(0.05, 0.5, '{text}'.format(text=text), size=size)
+        fig.set_dpi(dpi)
+        fig.canvas.print_figure(buf)
+        plt.close(fig)
 
     im = Image.open(buf)
     bg = Image.new(im.mode, im.size, (255, 255, 255, 255))
@@ -588,7 +590,6 @@ def latex_to_img(text, size=25):
 
 def make_transparent(img, bg=(255, 255, 255, 255)):
     """Given a PIL image, makes the specified background color transparent."""
-    from PIL import Image
     img = img.convert("RGBA")
     clear = bg[0:3]+(0,)
     pixdata = img.load()
@@ -613,6 +614,9 @@ class LaTeX(PNG):
     size = param.Number(default=25, bounds=(1, 100), doc="""
         Scales the size of the rendered equation.""")
 
+    dpi = param.Number(default=100, bounds=(1, 1900), doc="""
+        Matplotlib dpi parameter.""")
+
     @classmethod
     def applies(cls, obj):
         if hasattr(obj, '_repr_latex_'):
@@ -625,7 +629,7 @@ class LaTeX(PNG):
     def _img(self):
         obj=self.object if isinstance(self.object, basestring) else \
             self.object._repr_latex_()
-        return make_transparent(latex_to_img(obj, self.size))._repr_png_()
+        return make_transparent(latex_to_img(obj, self.size, self.dpi))._repr_png_()
 
 
 class RGGPlot(PNG):
--- a/panel/pane.py
+++ b/panel/pane.py
@@ -571,7 +571,7 @@ def latex_to_img(text, size=25, dpi=100):
     import io
     
     buf = io.BytesIO()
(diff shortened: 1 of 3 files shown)


panel/pane.py
```diff
@@ -560,6 +560,74 @@ def _img(self):
...
+    Usage: latex_to_img(r'$\frac(x}{y^2}$')
+    From https://stackoverflow.com/questions/1381741.
+    """
+    import matplotlib.pyplot as plt
+    from PIL import Image, ImageChops
+    import io
+    
+    buf = io.BytesIO()
+    plt.rc('text', usetex=True)
+    plt.rc('font', family='serif')
```
philippjfr: Should not modify rcparams globally use, should instead use:

```python
with plt.rc_context({'text.usetex': True, 'font.family': 'serif'}):
    ...
```

panel/pane.py
```diff
@@ -560,6 +560,74 @@ def _img(self):
...
+    plt.axis('off')
```
philippjfr: Should not use global figure, doesn't need to use savefig and should also make dpi configurable:

```python
fig = plt.figure()
ax = fig.add_subplot(111)
ax.axis('off')
ax.text(0.05, 0.5, '{text}'.format(text=text), size=size)
fig.set_dpi(dpi)
fig.canvas.print_figure(b)
plt.close(fig)
```

panel/pane.py
```diff
@@ -560,6 +560,74 @@ def _img(self):
...
+
+    # Precedence is dependent on the data type
+    precedence = None
+
+    size = param.Number(default=25, bounds=(1, 100), doc="""
+        Scales the size of the rendered equation.""")
+
+    @classmethod
+    def applies(cls, obj):
+        if hasattr(obj, '_repr_latex_'):
```
philippjfr: Should return False if matplotlib not available.

jbednar: That sounds like a good idea, but I can't see how to make it work.  `pane.Matplotlib` checks `sys.modules` for `matplotlib`, which is valid because there shouldn't be any way to have a Matplotlib figure in memory without `matplotlib` having been imported.  Here, however, the user is just supplying a string or maybe an `IPython.display.Latex` object, and so the user need not have imported anything.  So if we wanted `applies` to return False, wouldn't we need to try actually importing matplotlib and pillow within the `applies` method, so that we know whether they are available?  That doesn't seem feasible; I wouldn't want Panel to have to import those two libraries unless we actually have an instance of `pane.LaTeX` already.  So I don't *think* what you are proposing here is feasible.

panel/pane.py
jbednar: Thanks; fixed.

panel/pane.py
jbednar: Thanks; fixed.

panel/pane.py
philippjfr: It's not ideal but importing matplotlib and pillow in the `_repr_latex_` branch is not a major issue unless you expect that we register another pane that handles `_repr_latex_`, because if you're already in that branch it will pick that pane and simply fail loudly later when the pane is actually being rendered.

panel/pane.py
jbednar: Do you have any suggestions for how `dpi` and `size` relate in this code?  I'd like the docstring to describe how to use them, but I can't detect the practical difference between them.

philippjfr: Have a look at the ``Matplotlib`` pane, it computes the size of the plot relative to a fixed dpi. Which means the dpi scales the resolution while the size will scale the actual size of plot on the page.

panel/pane.py
jbednar: Ok.  I guess the benefit of adding that will be that if an object provides multiple repr methods and matplotlib or pillow is not installed, then display will drop back to one of those other repr methods.

panel/pane.py
jbednar: I still can't tell the difference:

```
import panel as pn
from panel.pane import LaTeX
e = r'$\frac{x}{y^2}$'

pn.Row(LaTeX(e, size=25, dpi=100), LaTeX(e, size=50, dpi=100), LaTeX(e, size=25, dpi=200))
```
image

philippjfr: I can't tell where in the code you're scaling with a fixed dpi.

philippjfr: I would have expected something like:

```
    def _imgshape(self, data):
        """Calculate and return image width,height"""
        w, h = super(LaTeX, self)._imgshape(data)
        w, h = (w/self.dpi), (h/self.dpi)
        return int(w*72), int(h*72)
```

Haven't tested it though.

philippjfr: Seems to work:

screen shot 2018-10-09 at 6 04 41 pm

philippjfr: Any idea why they sometimes overlap and have too much padding at other times?

jbednar: Thanks; I'll try it.  I haven't been able to tell how the spacing works out, here or in Panes in general. :-(

jbednar: The spacing is very weird:

```
import panel as pn
from panel.pane import LaTeX
e = r'$\frac{x}{y^2}$'

pn.Column(
    pn.Row(LaTeX(e), LaTeX(e)),
    pn.Row(LaTeX(e), "txt", LaTeX(e)))
```
image

Do you have any idea what's going on there?

jbednar: (Code required for the above just now pushed to the branch)

jbednar: This particular spacing issue doesn't seem to be specific to LaTeX:

image

jbednar: So I'll ignore that issue for this PR.

PR #92, opened 2018-10-10
Add support for Sympy objects to pane.LaTeX

jbednar:
Sympy Expr objects have a `.latex()` method for generating a LaTeX-syntax version of the expression, so it's simple to add support to pane.LaTeX for them:

```
import sympy
x = sympy.symbols('x')
i = sympy.integrate(sympy.sin(x**2), (x, -sympy.oo, sympy.oo))

import panel as pn
pn.Row(x, pn.Spacer(width=-150), i, pn.Spacer(width=30), sympy.sin(x**2)) 
```
image

Other changes:
- As the spacing is still unpredictable, which I think is unrelated to LaTeX, I also allowed Spacer to accept negative widths and heights to let things be lined up properly.
- The error message when explicitly instantiating a pane with an unallowed object was still trimming off the last 5 characters (P,a,n,e,l) despite the long-ago renaming, so I fixed that.

### DIFF
--- a/panel/layout.py
+++ b/panel/layout.py
@@ -262,9 +262,9 @@ class Tabs(Layout):
 
 class Spacer(Reactive):
 
-    height = param.Integer(default=None, bounds=(0, None))
+    height = param.Integer(default=None, bounds=(None, None))
 
-    width = param.Integer(default=None, bounds=(0, None))
+    width = param.Integer(default=None, bounds=(None, None))
 
     _bokeh_model = BkSpacer
 
--- a/panel/pane.py
+++ b/panel/pane.py
@@ -84,7 +84,7 @@ class PaneBase(Reactive):
                 raise ValueError('If a Pane declares no precedence '
                                  'the applies method should return a '
                                  'precedence value specific to the '
-                                 'object type or False, %s pane '
+                                 'object type or False, but the %s pane '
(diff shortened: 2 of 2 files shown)


jbednar:
Note that I didn't add tests for the Sympy support, because I'm guessing that we don't want to require that our test environment include sympy, but of course such tests can be added if needed.

ceball:
> Note that I didn't add tests for the Sympy support, because I'm guessing that we don't want to require that our test environment include sympy, but of course such tests can be added if needed.

In tox.ini, there's a test matrix (although it's not obvious that's so!). You could therefore add a 'with sympy' entry for the sympy tests. Other tests would continue to run fine without sympy, but anyone could request the sympy tests to run in a dev environment (in which case they'd be responsible for installing sympy), and sympy tests would be run when built dists/packages are tested (in which case sympy would be installed by the relevant package manager into the test environment before running the sympy tests).

However, having said all that, there isn't yet a good example of this for you to copy! The first good example will be param, when that's converted to pyctdev (param's got various optional tests, but only with pip and tox at the moment).

jbednar:
Thanks.  That capability does sound particularly useful for Panel, as it is designed to support such a wide range of other libraries that no one installation is likely to need.

philippjfr:
Looks good to me.

PR #94, opened 2018-10-12
Reverted changes to Div layout

philippjfr:
Attempt to restore old Div layout behavior as described in https://github.com/pyviz/panel/issues/93.


philippjfr:
@jbednar If possible and time permits could you do some testing before I merge?

PR #97, opened 2018-10-13
Simplify pip package builds

ceball:
Unlike for geoviews, no need for awkward conda+pip combination.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,9 +26,9 @@ stages:
   - name: docs
     if: (branch = master) AND (type != pull_request)
   - name: pip_dev_package
-    if: tag =~ ^v(\d+|\.)*[a-z]\d*$
+    if: tag =~ ^v(\d+|\.)+[a-z]\d+$
   - name: pip_package
-    if: tag =~ ^v(\d+|\.)*[^a-z]\d*$
+    if: tag =~ ^v(\d+|\.)+[^a-z]\d+$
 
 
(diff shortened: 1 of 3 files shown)


ceball:
To test: push a dev tag, and see if the travis build produces pip packages (which appear on https://testpypi.python.org/).

I tested locally using the same steps as on travis:
```
$ pip install pyctdev && doit ecosystem=pip ecosystem_setup
$ doit ecosystem=pip package_build --test-python=py36 --test-group=unit --sdist-install-build-deps
```

Travis tests on more python versions. The command line options are mysterious, yes, but gradually reducing over time...

I usually find it's easier just to push changes to travis and go do something else in the meantime, waiting for feedback. But if you want immediate feedback, or you wouldn't otherwise get feedback (e.g. because it's a step that would only happen on a tag or only weekly or whatever), or the task is too big for travis, it's handy to be able to run the same commands locally. (It's not perfect yet at all. When more config moves out of .travis and into setup.cfg/equivalent, that should make things easier.)

Also, using conda (as I assume you would be), you should run the above in a fresh conda environment, i.e. something like this first:
```
$ conda create -n panelpip python=3.6
$ conda activate panelpip # or source activate depending on your setup
```

philippjfr:
Thank you so much cleaner. I have to admit I didn't really know what I was doing. Being able to run things locally will be really helpful!

PR #98, opened 2018-10-13
WIP: try to clean up dependencies

ceball:
Couple of questions...

### Which dependencies are not on defaults but are on conda-forge? 

To avoid the conda-forge channel being a requirement, what about requesting those packages be added to defaults? Or if they are straightforward python packages, why not copy them to the pyviz channel instead of adding another channel?

### Dependencies needed to run examples/build docs

In the 'test' dependencies, there's:
```
        # For Panes.ipynb                                                                                                                      
        'plotly',
        'altair',
        'vega_datasets'
```

Are they really test dependencies, or are they needed to run the examples (e.g. for a doc build, or a user who wants to run the examples)?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,8 +11,8 @@ sudo: false
 env:
   global:
     - PYENV_VERSION=3.6
-    - CHANS_DEV="-c pyviz/label/dev -c jbednar"
-    - CHANS_REL="-c pyviz -c jbednar"
+    - CHANS_DEV="-c pyviz/label/dev -c defaults -c conda-forge"
+    - CHANS_REL="-c pyviz -c defaults -c conda-forge"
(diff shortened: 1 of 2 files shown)


philippjfr:
There are a number of optional tests for altair and plotly and they will become doc build dependencies as well. Not even sure the vega datasets are used at all, but I'd have to double check that. Adding plotly and altair to defaults seems like a reasonable request, but I'd by okay mirroring them on pyviz for now.

jbednar:
Actually, they are already mirrored on the jbednar channel for the same reason, so you can substitute conda-forge with jbednar if you'd like.  That way the other packages won't be on pyviz to potentially cause any other confusion...

philippjfr:
Superseded by #98

Issue #99, opened 2018-10-13
Add panel entry point


jbednar:
Passing along a suggestion that I think is a good one: so that Panel users won't be affected if we eventually do split out the parts of Bokeh that Panel uses, we should add an entry point for launching servers like `panel serve file.py`.  For now that would just delegate to `bokeh serve file.py`, but we could then change that to whatever the underlying Bokeh infrastructure becomes, which could even be part of Panel itself.

philippjfr:
This was implemented.

PR #101, opened 2018-10-15
Implementation of panel console script

jlstevens:
This PR aims to address #99. I'll be extending this PR with some experimental functionality with another PR based off this one.

### DIFF
--- a/panel/cli.py
+++ b/panel/cli.py
@@ -1,0 +1,8 @@
+"""
+Commandline interface to Panel
+"""
+
+from bokeh.__main__ import main as bokeh_entry_point
+
+def main():
+    bokeh_entry_point()
--- a/setup.py
+++ b/setup.py
@@ -102,6 +102,10 @@ setup_args = dict(
         "Topic :: Scientific/Engineering",
         "Topic :: Software Development :: Libraries"],
     python_requires=">=2.7",
+    entry_points={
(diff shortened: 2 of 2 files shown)


jbednar:
Looks good so far, thanks.

ceball:
I think you should probably add entry points to the conda recipe too, otherwise users will likely run into problems in some circumstances. In any case, as far as I know, it's recommended by conda build to do this - although I can't immediately find the documentation for that.

You could explicitly add the entry points, or you could template them from setup.py (see e.g. https://github.com/ioam/parambokeh/blob/master/conda.recipe/meta.yaml#L13). When there's only one entry point, and it's unlikely to change, templating might not be worthwhile - up to you.

The next release of pyctdev will remove the need for a conda build recipe - one is generated from the information in setup.py/setup.cfg (it can still be stored/edited if desired). Currently this functionality is only available in a dev release of pyctdev, and although pyviz itself is using the dev release, it's not quite ready to go yet...

ceball:
By the way, it seems like a convention to use `__main__.py` (and bokeh itself follows that, I think, judging from the import you have here). Using __main__ also has some additional effects, e.g. the contents of __main__.py will executed if the module is run with `-m` (https://docs.python.org/3/library/__main__.html#module-__main__). That can also mean something for zipped apps, etc.

PR #102, opened 2018-10-15
Initial example of generating a report with the panel command

jlstevens:
This is an experimental PR that is being used to investigate whether panel could have a 'report' mode that can take any notebook and turn it into a report style page.

For instance, given this notebook `example.ipynb`:

image

You can run:

```
panel report example.ipynb --show
```

To get this page served via bokeh server:

image

Note that 1) code is not shown 2) markdown cells are promoted to markdown panes that are served 3) the markdown pane height is computed via a heuristic via the number of lines (and it isn't currently very good).

This is work in progress and there are quite a few things to do and a few fundamental issues to think about:

- [ ] The simple parsing used should be replaced with `argparse`.
- [ ] The bokeh help for subcommands needs to be shadowed to list the `report` command
- [ ] Panel is repeatedly being imported. Instead a code cell could be inserted at the top of the notebook.
- [ ] The `wrap_cell_expression` function doesn't handle semicolons quite right. 
- [ ] Unit tests of the source to source conversion.

The most fundamental issue is that a report is a vertical format and a lot of output is in the form of text or html (literal output, pandas dataframes, markdown cells) where the height currently needs to be specified. While a heuristic for the markdown height might work reasonably well it cannot generalize to all possible outputs. 

Currently I think this PR should be viewed as a proposal for discussion and a proof-of-concept rather than an implementation that should be merged.

### DIFF
--- a/panel/__init__.py
+++ b/panel/__init__.py
@@ -55,3 +55,11 @@ class extension(param.ParameterizedFunction):
         """
         viewable, model = Viewable._views.pop(msg['id'])
         viewable._cleanup(model)
+
+
+def serve_panel(arg):
+    return Pane(arg).servable()
+
+def serve_markdown(source, lines, height_per_line=30):
+    height = lines * height_per_line
+    return Pane(source, height=height).servable()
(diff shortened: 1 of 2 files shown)


philippjfr:
Closing for now. This is a neat idea but needs to be largely refactored since the CLI handling has changed entirely.

PR #103, opened 2018-10-16
Build and load bokeh compiled model cache

philippjfr:
This PR customizes setup.py in such a way that it compiles custom bokeh models and stores the compiled code alongside the original code. Additionally it registers a hook with bokeh which loads the cached models during bokeh's compilation step.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,7 +15,7 @@ env:
     - CHANS_REL="-c pyviz -c jbednar"
     - LABELS_DEV="--label dev"
     - LABELS_REL="--label dev --label main"
-    - PKG_TEST_PYTHON="--test-python=py27 --test-python=py36"
+    - PKG_TEST_PYTHON="--test-python=py36"
 
 stages:
   - test
@@ -78,7 +78,7 @@ jobs:
       env: PYPI=testpypi PYPIUSER=$TPPU PYPIPASS=$TPPP
       stage: pip_dev_package
       install:
-        - doit develop_install $CHANS_DEV -o tests && pip uninstall -y panel
+        - doit develop_install $CHANS_DEV -o build -o tests && pip uninstall -y panel
         - doit pip_on_conda
         - conda install $CHANS_DEV nodejs=9.11.1
         - doit ecosystem=pip package_build $PKG_TEST_PYTHON --test-group=unit
--- a/.travis.yml
+++ b/.travis.yml
@@ -73,11 +73,11 @@ jobs:
         - doit package_build $CHANS $PKG_TEST_PYTHON --test-group=unit
       script: doit package_upload --token=$ANACONDA_TOKEN $LABELS
 
-
     - &pip_pkg
       env: PYPI=testpypi PYPIUSER=$TPPU PYPIPASS=$TPPP
       stage: pip_dev_package
       install:
+        - pip install pyctdev
         - doit develop_install $CHANS_DEV -o tests && pip uninstall -y panel
         - doit pip_on_conda
         - conda install -c conda-forge nodejs=9.11.1
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,7 +11,7 @@ sudo: false
 env:
   global:
     - PYENV_VERSION=3.6
-    - CHANS_DEV="-c pyviz/label/dev -c jbednar -c bokeh/label/dev"
+    - CHANS_DEV="-c pyviz/label/dev -c conda-forge -c bokeh/label/dev"
     - CHANS_REL="-c pyviz -c jbednar"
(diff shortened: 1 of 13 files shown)


philippjfr:
@ceball Could you have a brief look at my changes to the .travis.yml? nodejs is now a build requirement so I've had to go back to the pip_on_conda approach.

ceball:
The "pip on conda" approach was experimental, for geoviews, where *users* also cannot install geoviews using pip alone without taking additional steps (because of some of its dependencies). In contrast, for panel, isn't it only *package builders* who require more than pip to get all the dependencies? Is that right? If that's right, I would suggest a different approach.

However, either way, the "pip on conda" approach should still work (it's just not ideal), so you could carry on and I can fix up later. (I can't work right now.)

.travis.yml
```diff
@@ -11,11 +11,11 @@ sudo: false
 env:
   global:
     - PYENV_VERSION=3.6
-    - CHANS_DEV="-c pyviz/label/dev -c jbednar"
+    - CHANS_DEV="-c pyviz/label/dev -c jbednar -c bokeh/label/dev"
```
ceball: Just to check: is this addition a build time channel requirement, or do users need it too?

.travis.yml
```diff
@@ -73,15 +73,16 @@ jobs:
         - doit package_build $CHANS $PKG_TEST_PYTHON --test-group=unit
       script: doit package_upload --token=$ANACONDA_TOKEN $LABELS
 
-
     - &pip_pkg
-      env: PYPI=testpypi PYPIUSER=$TPPU PYPIPASS=$TPPP
+      <<: *conda_default
+      env: PYPI="https://test.pypi.org/legacy/" PYPIUSER=$TPPU PYPIPASS=$TPPP
```
ceball: Presumably reverting the shortcut was unintended side effect of copy/paste?

```suggestion
      env: PYPI=testpypi PYPIUSER=$TPPU PYPIPASS=$TPPP
```

I've never tried this 'suggestion' thing before...

.travis.yml
```diff
@@ -90,7 +91,7 @@ jobs:
       env: DESC="" TRAVIS_NOCACHE=$TRAVIS_JOB_ID LABELS=$LABELS_REL CHANS=$CHANS_REL
 
     - <<: *pip_pkg
-      env: PYPI=pypi PYPIUSER=$PPU PYPIPASS=$PPP
+      env: PYPI="https://upload.pypi.org/legacy/" PYPIUSER=$PPU PYPIPASS=$PPP
```
ceball:
```suggestion
      env: PYPI=pypi PYPIUSER=$PPU PYPIPASS=$PPP
```

tox.ini
```diff
@@ -4,7 +4,6 @@
 [tox]
 #          python version             test group                  extra envs  extra commands       
 envlist = {py27,py35,py36}-{flakes,unit,examples,all_recommended}-{default}-{dev,pkg}
-build = wheel
```
ceball: I'd hope you could leave this; please let me know if things don't work with this still present.

philippjfr:
You're right that it's a build time dependency only. Going back to the cleaner approach would be nice but I couldn't work out how to satisfy the nodejs dependency in that case.

.travis.yml
philippjfr: Build time dependency only

.travis.yml
philippjfr: No this was intentional, at some point the package upload started failing with some error about testpypi not being a valid argument.

tox.ini
philippjfr: I'll revert this, I was trying to work around py27 issues and this was just collateral.

.travis.yml
philippjfr: [Here](https://travis-ci.org/pyviz/panel/jobs/442481612)'s the actual error

```
$ doit ecosystem=pip package_upload -u $PYPIUSER -p $PYPIPASS -pypi ${PYPI}
ERROR: Invalid parameter: "testpypi". Must be a command, task, or a target.
Type "doit help" to see available commands.
Type "doit list" to see available tasks.
```

philippjfr: Ah I see my error, will fix it now.

ceball:
Just to be clear, what I meant above is that I think you should be able to continue using the "pip on conda" approach, so don't let my comment that it's not ideal hold up merging if this is otherwise ready. 

(The downside is that the pip package testing will also occur with the package build time dependencies present, i.e. it won't mimic the setup of a typical pip user - nodejs will be present. But I think that's ok for now. I'll revisit later.)

philippjfr:
Since bokeh 1.0 is now released I'll merge.

PR #126, opened 2018-10-24
Add pn.panel and ensure all Panes have layout attribute

philippjfr:
@jbednar Please test this extensively.

- [x] Implements https://github.com/pyviz/panel/issues/124
- [x] Updates unit tests
- [x] Update docs

### DIFF
--- a/examples/user_guide/Panes.ipynb
+++ b/examples/user_guide/Panes.ipynb
@@ -19,7 +19,7 @@
    "source": [
     "``Panel`` makes it possible to display and arrange a wide range of plots and other media on a page, including plots (matplotlib, bokeh, vega/altair, holoviews, and plotly), images (pngs, svgs, gifs, jpegs),  and text (Markdown, HTML etc.).\n",
     "\n",
-    "The ``pn.create`` helper function returns a viewable representation of any object, e.g. a simple string will be interpreted as ``Markdown`` by default:"
+    "The ``pn.panel`` helper function returns a viewable representation of any object, e.g. a simple string will be interpreted as ``Markdown`` by default:"
    ]
   },
   {
@@ -28,7 +28,7 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "layout = pn.create('**Some text**')\n",
+    "layout = pn.panel('**Some text**')\n",
     "layout"
    ]
   },
@@ -36,7 +36,7 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "The ``pn.create`` will always return a Panel layout containing one or more views representing the supplied object. However we can also access the items of the ``Panel``, e.g. to confirm the type of the ``Pane`` that was created:"
+    "The ``pn.panel`` will always return a Panel layout containing one or more views representing the supplied object. However we can also access the items of the ``Panel``, e.g. to confirm the type of the ``Pane`` that was created:"
    ]
   },
   {
@@ -108,7 +108,7 @@
     "fig = figure(width=300, height=300)\n",
     "r = fig.line([1, 2, 3], [1, 2, 3])\n",
     "\n",
-    "bk_layout = pn.create(fig)\n",
+    "bk_layout = pn.panel(fig)\n",
     "bk_layout"
    ]
   },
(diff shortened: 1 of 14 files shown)


panel/layout.py
```diff
@@ -152,23 +158,23 @@ def pop(self, index):
         self.objects = new_objects
```
jbednar: Should we add Layout = Panel for now?  Or is it unlikely that user code would have used Layout by that name?

panel/layout.py
```diff
@@ -313,3 +320,6 @@ def _get_model(self, doc, root=None, parent=None, comm=None):
         model = self._bokeh_model(**self._init_properties())
         self._link_params(model, ['width', 'height'], doc, root, comm)
         return model
+
+
+PaneBase._default_layout = Row
```
jbednar: Should this not be a public class Parameter?  Seems like something one could legitimately want to customize.

panel/pane.py
```diff
@@ -24,6 +24,31 @@
...
+
+    Returns
+    -------
+    layout: Viewable
+       A Viewable representation of the input object
+    """
+    if isinstance(obj, PaneBase):
+        return obj
+    elif isinstance(obj, Viewable):
+        return obj
```
jbednar: Isn't PaneBase a subtype of Viewable?  So why test PaneBase explicitly?

panel/layout.py
```diff
@@ -129,19 +129,25 @@ def _get_model(self, doc, root=None, parent=None, comm=None):
     def __getitem__(self, index):
         return self.objects[index]
 
+    def __len__(self):
```
jbednar: Shouldn't `pane = Pane(pane, _temporary=True)` in _get_objects above be `pane = create(pane, _temporary=True)`?

panel/layout.py
```diff
@@ -10,14 +10,14 @@
                            WidgetBox as BkWidgetBox, Spacer as BkSpacer)
 from bokeh.models.widgets import Tabs as BkTabs, Panel as BkPanel
 
-from .pane import Pane, PaneBase
+from .pane import create, Pane, PaneBase
```
jbednar: Can this import of Pane be deleted?  If not, seems like it should be marked `# noqa`.

panel/pane.py
philippjfr: No reason now, just didn't tidy it up.

panel/layout.py
philippjfr: Seems unlikely that any user would have used an abstract baseclass.

philippjfr:
Always prefer if you warn me about making edits to a branch I'm actively working on. You can now also leave suggested edits directly in the GitHub, which let's me apply the changes when I'm ready.

panel/layout.py
philippjfr: Actually no, this was deliberate.

jbednar:
@jlstevens, consider code like this:

```
header = pn.Row(pn.create(oggm_logo, width=170), pn.Spacer(width=50), 
                pn.Column(pn.Pane(title, height=25, width=400), pn.Spacer(height=-15), pn.create(instruction, width=500)),
                pn.Spacer(width=180), pn.Column(pn.create(dyn_count), clear_button, pn.Spacer(height=-15)), 
                pn.create(pv_logo, width=80))
```

To me `create` reads oddly there; I'd rather see `panel` so that I can see that it's a row of rows, spacers, and panels, not a row of creates...

panel/layout.py
philippjfr: Nevermind, it wasn't.

examples/user_guide/Panes.ipynb
```diff
@@ -61,14 +78,23 @@
...
    "source": [
-    "pane.object = '<b>Some updated text</b>'"
+    "pane.object = 'Some updated text'"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "One thing to note is that since a ``Pane`` may define multiple views, when working with an explicitly constructed ``Pane`` care should be taken to compose the component of the ``Pane.layout`` manually."
```
jbednar: the components?

PR #130, opened 2018-10-26
Changed default Param.expand_layout

philippjfr:


jbednar:
I *think* this is the best default behavior...

PR #133, opened 2018-10-31
Transform command-line arguments passed by anaconda project

jlstevens:
This PR helps transform arguments that may be passed to the `panel` console script that would not be accepted by bokeh without transformation.


jbednar:
Note that using --allow-websocket-origin in this way requires bokeh 0.12.7 or later, which anaconda-project itself can't do (as it is meant to work with any Bokeh version, even though it doesn't currently), but it should be ok here because Panel requires a later version of bokeh anyway.

PR #136, opened 2018-11-05
Interactive layout fixes

philippjfr:
The interactive layout was previously self-referential which causes issues when trying to display the plot/display component on its own. Additionally this ensures that a small fixes to assign a minimal height when no height is specified for the last item in a Column.

### DIFF
--- a/panel/layout.py
+++ b/panel/layout.py
@@ -11,7 +11,7 @@ from bokeh.layouts import (Column as BkColumn, Row as BkRow,
 from bokeh.models import Box as BkBox
 from bokeh.models.widgets import Tabs as BkTabs, Panel as BkPanel
 
-from .util import push, abbreviated_repr
+from .util import param_reprs, push
 from .viewable import Reactive, Viewable
 
 
@@ -161,9 +161,7 @@ class Panel(Reactive):
     def __repr__(self, depth=0):
         spacer = '\n' + ('    ' * (depth+1))
         cls = type(self).__name__
-        params = ['%s=%s' % (p, abbreviated_repr(v)) for p, v in sorted(self.get_param_values())
-                  if v is not self.params(p).default and v not in ('', None)
-                  and p != 'objects' and not (p == 'name' and v.startswith(cls))]
+        params = param_reprs(self, ['objects'])
         objs = ['[%d] %s' % (i, obj.__repr__(depth+1)) for i, obj in enumerate(self.objects)]
         if not params and not objs:
             return super(Panel, self).__repr__(depth+1)
--- a/panel/layout.py
+++ b/panel/layout.py
@@ -15,6 +15,26 @@ from .util import push, abbreviated_repr
 from .viewable import Reactive, Viewable
 
 
+def no_height(layout):
+    """
+    Whether the supplied layout has no height
+    """
+    if isinstance(layout, BkBox):
+        for child in layout.children:
+            if isinstance(child, BkBox):
+                if no_height(child):
+                    return False
+
+            height = getattr(child, 'height', None)
+            if height is not None:
+                return False
+    elif isinstance(BkWidgetBox):
+        return False
+    elif getattr(child, 'height', None) is not None:
+        return False
+    return True
+
+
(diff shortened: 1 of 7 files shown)


panel/layout.py
```diff
@@ -15,6 +15,26 @@
...
+            height = getattr(child, 'height', None)
+            if height is not None:
+                return False
+    elif isinstance(BkWidgetBox):
+        return False
+    elif getattr(child, 'height', None) is not None:
+        return False
+    return True
+
+
```
jbednar: It really hurts my brain to follow the logic of a method named `no_height`; is there any way it could be changed to `has_height`?

panel/layout.py
```diff
@@ -118,8 +138,7 @@ def _get_model(self, doc, root=None, parent=None, comm=None):
         objects = self._get_objects(model, [], doc, root, comm)
 
         # HACK ALERT: Insert Spacer if last item in Column has no height
-        if (isinstance(self, Column) and objects and not isinstance(objects[-1], (BkWidgetBox, BkBox))
-            and getattr(objects[-1], 'height', False) is None):
+        if (isinstance(self, Column) and objects and no_height(objects[-1])):
             objects.append(BkSpacer(height=50))
```
jbednar: Add a parameter `default_height`?  Could be useful e.g. for the report mode dashboards...

panel/layout.py
philippjfr: Fixed.

jbednar: Much better now, and shorter!  Now I can follow the logic without any brain meltdowns. :-)

jbednar:
I'm seeing errors on this PR with "ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()" with https://anaconda.org/jbednar/datashadercliffordinteract/notebook

philippjfr:
Not sure how that could be caused by this PR.

jbednar:
Not sure either, just that it works with master (v0.2.0a4-1-ge98e151) and not with this branch (v0.2.0a4-6-g40bc2ba).

panel/layout.py
philippjfr: Where would the parameter live? I'd be happy to add it but maybe in a separate PR.

jbednar: Well, height=50 is used here on Panel, so presumably on Panel, so that it can be height=self.default_height

philippjfr: It's specifically behind an isinstance check for Column to avoid having to duplicate the whole method there.

philippjfr: That said I could add an equivalent check for Row:

``(isinstance(self, Row) and objects and not any(has_height(o) for o in objects))``

philippjfr: Merging this, but I can follow up on this tomorrow.

jbednar:
This change really improved the usability of what's returned by interact(), and I've updated https://anaconda.org/jbednar/datashadercliffordinteract/notebook to reflect that.  Now, the repr is something even more clear and clearly usable:

```
>>> i = pn.interact(clifford_plot, a=(-2,2), b=(-2,2), c=(-2,2), d=(-2,2), 
              n=(1,20000000), colormap=ps, panel_layout=pn.Row)
>>> print(i)


Column
    [0] WidgetBox
        [0] FloatSlider(end=2, name='a', start=-2, value=1.7)
        [1] FloatSlider(end=2, name='b', start=-2, value=1.7)
        [2] FloatSlider(end=2, name='c', start=-2, value=0.6)
        [3] FloatSlider(end=2, name='d', start=-2, value=1.2)
        [4] IntSlider(end=20000000, name='n', start=1, value=2000000)
        [5] Select(name='colormap', options={'bgy': ['#fff123', ...}, value=['#b3fef5', '#b0fef5', ...])
    [1] Row
        [0] PNG(Image, name='interactive00026')
```

So it's now obvious that one can do `pn.Column(i[0][5], i[1][0])` to get a column with just the Select widget above the PNG image, with no opaque `interact` objects.  Very nice!

PR #155, opened 2018-11-15
Added Player step parameter

philippjfr:


jbednar:
Is there a Param attribute that maps to this?  Seems like we need step for both Player and sliders.

philippjfr:
There is not, I thought there was an issue about that somewhere but I can't find it now.

kcpevey:
https://github.com/pyviz/EarthSim/issues/196 That was related to parambokeh but is now relevant to panel.

PR #156, opened 2018-11-15
Added collage

jbednar:
image

Right now, the collage items link to the Anaconda.org rendered notebook, but @philippjfr, you may want to edit this PR to make it link to a live notebook?  Seems like the code is also important to show, though.


philippjfr:
I'm working on getting these served as apps on mybinder, i.e. no notebooks and therefore no code. I think that's fine for a landing page and we can have a gallery page which links to the code and the app versions.

jbednar:
I've added separate titles for each thumbnail, so if you want we could have the image link to the running app, and the name link to the notebook.  We can also have each running app link back to the notebook...

jbednar:
In the meantime, consider just merging it so that the website has these examples, and updating in a separate PR.

Issue #109, opened 2018-10-19
"panel examples" command does not work


ceball:
Following http://panel.pyviz.org/#usage, I couldn't get the examples:
```
$ panel --install-examples
usage: /home/ceball/mc3/envs/panel/bin/panel [-h] [-v] {html,info,json,png,sampledata,secret,serve,static,svg} ...
/home/ceball/mc3/envs/panel/bin/panel: error: unrecognized arguments: --install-examples
```

I was going to add pyct to make `panel examples` available (in common with other pyviz projects), but I'm not sure how I ought to do that, because there doesn't seem to be any existing argparse parser I can add to - bokeh's main() is  just called.

jbednar:
@jlstevens is going to need to add argparse-based options anyway, so can you please set it up to be like all our other projects except that if a command doesn't match and it would otherwise show a usage message, instead it call's bokeh's main()?

jbednar:
@jlstevens, it looks like our instructions still don't match what panel currently does, and everything will be more consistent if we change what panel does rather than the instructions...

jlstevens:
It should now be working, though I think the command is just `panel examples`. Closing.

PR #176, opened 2018-12-04
Extended parser to support pyct commands

jlstevens:
Addresses #109. It isn't obvious how to do the right thing here but this seems to me like the easiest way to splice together two external parsers. It does mean if you mistype 'serve', then it is the pyct parser that will complain.

### DIFF
--- a/panel/cli.py
+++ b/panel/cli.py
@@ -36,6 +36,7 @@ if __name__ == "__main__":
     main()
 
 def main(args=None):
+    """Special case: Allow Bokeh to handle the `serve` command; rest is handled by pyct."""
     if len(sys.argv) > 1 and 'serve' == sys.argv[1]:
         sys.argv = transform_cmds(sys.argv)
         bokeh_entry_point()
--- a/panel/cli.py
+++ b/panel/cli.py
@@ -8,7 +8,7 @@ from bokeh.__main__ import main as bokeh_entry_point
 
 def transform_cmds(argv):
     """
-    Allows usage with anaconda-project by remapping the argv list provided 
+    Allows usage with anaconda-project by remapping the argv list provided
     into arguments accepted by Bokeh 0.12.7 or later.
     """
     replacements = {'--anaconda-project-host':'--allow-websocket-origin',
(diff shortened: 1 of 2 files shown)


panel/cli.py
```diff
@@ -32,7 +32,17 @@ def transform_cmds(argv):
             transformed.append(arg)
     return transformed
 
+if __name__ == "__main__":
+    main()
 
-def main():
-    sys.argv = transform_cmds(sys.argv)
-    bokeh_entry_point()
+def main(args=None):
```
jbednar:
```suggestion
def main(args=None):
    """Special case: Allow Bokeh to handle the `serve` command; rest is handled by pyct."""
```

panel/cli.py
```diff
@@ -33,6 +33,18 @@ def transform_cmds(argv):
     return transformed
 
 
-def main():
-    sys.argv = transform_cmds(sys.argv)
-    bokeh_entry_point()
+def main(args=None):
+    """Special case: Allow Bokeh to handle the `serve` command; rest is handled by pyct."""
+    if len(sys.argv) > 1 and 'serve' == sys.argv[1]:
```
philippjfr: Is 'serve' the only bokeh command we are supporting?

philippjfr:
I'll merge, I'm not sure if we should support any bokeh commands other than ``serve`` but we should at least decide deliberately.

jbednar:
The commands currently offered by Bokeh are:

```
0172-jbednar:~/notes> bokeh --help
usage: /Users/jbednar/anaconda/envs/pyviz/bin/bokeh [-h] [-v]
                                                    {html,info,json,png,sampledata,secret,serve,static,svg}
                                                    ...

positional arguments:
  {html,info,json,png,sampledata,secret,serve,static,svg}
                        Sub-commands
    html                Create standalone HTML files for one or more applications
    info                print information about Bokeh and Bokeh server configuration
    json                Create JSON files for one or more applications
    png                 Create standalone PNG files for one or more applications
    sampledata          Download the bokeh sample data sets
    secret              Create a Bokeh secret key for use with Bokeh server
    serve               Run a Bokeh server hosting one or more applications
    static              Serve bokehjs' static assets (JavaScript, CSS, images, fonts,
                        etc.)
    svg                 Create standalone SVG files for one or more applications
```

I'm happy to be corrected, but I don't think there is any use for `info` or `sampledata`; people can call bokeh directly for those. I don't personally know any use for `json` or `static`, but someone who knows more might.   `secret`, `html`, `png`, and `svg` sound useful to me, and something potentially useful for panel to provide.

PR #177, opened 2018-12-05
Possible fixes for example packaging

jlstevens:
Attempt to fix issue where the examples aren't in the package and `panel examples` isn't working.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,12 +1,14 @@
 #!/usr/bin/env python
 
 import os
+import sys
 import json
 import hashlib
 from setuptools import setup, find_packages
 from setuptools.command.develop import develop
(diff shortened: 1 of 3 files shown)


setup.py
```diff
@@ -163,4 +166,10 @@ def run(self):
 
 
 if __name__=="__main__":
+
+    example_path = os.path.join(os.path.dirname(os.path.abspath(__file__)),
+                                'panel','examples')
+    if 'develop' not in sys.argv:
```
philippjfr: Don't really get this, develop installs the project in place which means you're basically just copying the files to the same location.

philippjfr: Sorry my mistake, missed the **not**.

PR #191, opened 2018-12-28
Enable Widget links to Hv Bokeh plots

xavArtley:
This PR is a work in progress
It allows to enable links between panel widgets and holoviews plots with bokeh backend
It add examples in WidgetLinks user guide notebook based on https://anaconda.org/philippjfr/widgetlinks/notebook
Tests must be added
Doc must be extended
Discussion is opened on the organisation of codes
And even if tests passed I thin I breaked previous links between hv elements.

### DIFF
--- a/panel/links.py
+++ b/panel/links.py
@@ -5,11 +5,10 @@ import weakref
 
 from .layout import Viewable, Panel
 from .widgets import Widget
-from .holoviews import HoloViews
+from .holoviews import HoloViews, is_bokeh_element_plot
 
 from holoviews.plotting.links import Link
 from bokeh.models import CustomJS
-from panel.holoviews import generate_hvelems_bkplots_map
 
 
 class WidgetLink(Link):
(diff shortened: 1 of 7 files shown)


xavArtley:
Example of result in an exported notebook
links_in_panel

xavArtley:
Now this works
image

philippjfr:
Thank you for working on this. I've been meaning too but have been annoyed I lost my initial progress. Note that I'm out for the next week, but will try to fit reviewing this in at some point.

panel/links.py
```diff
@@ -0,0 +1,151 @@
+"""
+"""
+import param
+import weakref
+
+from .layout import Viewable, Panel
+from .widgets import Widget
+from .holoviews import HoloViews, is_bokeh_element_plot
+
+from holoviews.plotting.links import Link
```
philippjfr: Panel should not require HoloViews, so we'll have to think about this a bit.

panel/links.py
```diff
@@ -0,0 +1,151 @@
...
+    linked to.
+    """
+    registry = weakref.WeakKeyDictionary()
+    
+    code = param.String(default=None)
+    
+    target_model = param.String(default=None)
+    
+    target_property = param.String(default=None)
+
```
philippjfr: Spacing should be consistent.

PR #195, opened 2018-12-31
Added links notebook

xavArtley:


philippjfr:
Thanks this looks like a really good start of an important user guide. I think we need to clarify the name a bit so it isn't confused with the JS based links we are about to add, maybe this is ``Linking in Python`` and then we also add ``Linking in Javascript``, not entirely sure yet.

xavArtley:
Indeed the title is not well chosen. When I began to write the notebook I wanted to regroup all links but finally I thought smaller guide was a better choice.

philippjfr:
I think I'll merge this as is and then update it in the JS link PR.

PR #199, opened 2019-01-06
Added general support for linking Panel objects

xavArtley:
Following PR #191
It focus only on the Links between widgets and Hv bokeh plots

I removed `AxesLink` which was a link between Hv bokeh plots
I removed dependencies to holoviews (duplicate code of the `Link` class of holoviews)
I will create another PR with some refactor of the links between Hv Plots to decorrelate Links from holoviews and links developed in Panel.


xavArtley:
Now it is possible to link directly bokeh widgets or panel widgets to bokeh figures
image

philippjfr:
@xavArtley I've started generalizing the LinkCallback a bit so we can eventually apply arbitrary linking. Next I'll be adding a simpler API to specify widget links. Please let me know if you disagree with any of my changes so far.

panel/links.py
```diff
@@ -91,132 +91,212 @@ def unlink(self):
...
-                             callback=WidgetLinkCallback)
 
-Viewable._preprocessing_hooks.append(WidgetLink.make_links)
+class WidgetLinkCallback(LinkCallback):
+
+    source_code = """
+    target['{value}'] = source.value
+    """
+
+    def _get_target_spec(cls, link):
```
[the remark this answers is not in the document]
philippjfr: Thanks I made it an instancemethod but did not update the signature.

panel/links.py
```diff
@@ -0,0 +1,295 @@
...
+
+    def _initialize_models(self, link, src_model, tgt_model):
+        if link.target_property and tgt_model:
+            setattr(tgt_model, link.target_property, src_model.value)
+        if tgt_model is None and not link.code:
+            raise ValueError('Model could not be resolved on target '
+                             '%s and no custom code was specified.' % 
+                             type(self.target).__name__)
+
+    def _process_references(self, references):
```
xavArtley: Should not a default method be added in the Parent class?

philippjfr: Yes, my oversight.

panel/links.py
```diff
@@ -92,23 +92,27 @@ def _process_links(cls, root_view, root_model):
...
+                    for link in cls.registry[linkable]
+                    for src in ([link.source] if not isinstance(link.source, HvElement) else map_hve_bk[link.source])
+                    for tgt in ([link.target] if not isinstance(link.target, HvElement) else map_hve_bk[link.target])]
+        else:
+            if not linkables:
+                return
+            found = [(link, src, link.target) for src in linkables if src in cls.registry
+                    for link in cls.registry[src] if link.target in linkables or
+                    not link._requires_target]
+
```
philippjfr: I don't completely follow this change. Could you explain the difference?

xavArtley: If you register an holoviews element as source the LinkCallBack __init__ fail because the source is an holoview element and not a Plot.
Now if we have an HvElement in link.source we find all bokeh plots(map_hve_bk[link.source]) and we create the link for each one

philippjfr: Thanks.

philippjfr:
@xavArtley I've once again refactored the code in order to make linking a completely general concept which can be applied both in Python and in JS. Have a look at the updated Links.ipynb to see what you think.

@jbednar When you get a minute I'd appreciate it if you could review and edit that notebook as well to see what you think.

xavArtley:
@philippjfr  Great work it's very intuitive.

jbednar:
If those changes look ok to you, I'm happy to merge it!

examples/user_guide/Links.ipynb
```diff
@@ -192,21 +273,49 @@
...
   {
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "`param.watch` return a reference to the watcher which can be used to remove it."
+    "As you can see the signature is identical to the ``link`` example above, but here Panel translates the specification into a JS code snippet which syncs the properties on the underlying Bokeh properties. But now if you edit the widget and press Return, the Markdown display will automatically update even in a static HTML web page.\n",
+    "\n",
+    "#### Linking using custom JS code\n",
+    "\n",
+    "Since everything happens in JS for a `jslink`, we can't provide a Python callback. Instead, we can define a JS code snippet, which is executed when a property changes. E.g. we can define a  little code snippet which adds HTML bold tags (``<b>``) around the text before setting it on the target. The code argument should map from the parameter/propery on the source object to the JS code snippet to execute:"
```
philippjfr:
```suggestion
    "Since everything happens in JS for a `jslink`, we can't provide a Python callback. Instead, we can define a JS code snippet, which is executed when a property changes. E.g. we can define a little code snippet which adds HTML bold tags (``<b>``) around the text before setting it on the target. The code argument should map from the parameter/property on the source object to the JS code snippet to execute:"
```

examples/user_guide/Links.ipynb
```diff
@@ -50,8 +67,19 @@
...
+    "text_input.link(markdown, value='object')\n",
+    "\n",
+    "pn.Row(text_input, markdown)"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "Now, if Python is running and you type something in the box above and press Return, the corrresponding Markdown pane should update to match.  And if you use Python to directly manipulate the value of the text input (e.g. by editing the following cell to have new text in it and then executing it), then the Markdown should also update to match:"
```
jbednar:
```suggestion
    "Now, if Python is running and you type something in the box above and press Return, the corresponding Markdown pane should update to match.  And if you use Python to directly manipulate the value of the text input (e.g. by editing the following cell to have new text in it and then executing it), then the Markdown should also update to match:"
```

philippjfr:
@xavArtley If you're around could you apply the suggestions? Only the PR author can do it.

philippjfr:
Thanks @xavArtley at working out the initial cut of this and thanks @jbednar for tidying up the docs.

PR #201, opened 2019-01-06
Decorrelate Links defined in panel from holoviews ones.

xavArtley:
Create a PanelLink class to define links specific to panel package.
For example AxesLink is not require in holoviews due to Layout
But is needed to link plots across panels
Add AxesLink + test
I refactor find_links to remove dependency to holoviews. 
Tests passed but I may skipped some use cases.

### DIFF
--- a/panel/links.py
+++ b/panel/links.py
@@ -1,0 +1,132 @@
+"""
+"""
+import param
+import weakref
+
+from collections import defaultdict
+from functools import partial
+
+from .viewable import Viewable
+from .holoviews import find_links
+
+
+class Link(param.Parameterized):
+    """
+    A Link defines some connection between a source and target object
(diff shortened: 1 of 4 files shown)


panel/links.py
```diff
@@ -0,0 +1,132 @@
...
+    plot and the axes on the target plot. By default it will
+    link along the x-axis but using the axes parameter both axes may
+    be linked to the tool.
+    """
+
+    axes = param.ListSelector(default=['x', 'y'], objects=['x', 'y'], doc="""
+        Which axes to link the tool to.""")
+
+
+class RangeAxesLinkCallback(param.Parameterized):
```
philippjfr: Could you delete these again, this type of link is definitely specific to HoloViews and should be added there instead.

Issue #202, opened 2019-01-07
entering two different invalid values into LiteralInput consecutively causes '(invalid)' message to disappear


lebedov:
When using Panel 0.3.1, if I create a `panel.widgets.LiteralInput`, enter an invalid value (e.g., `a`) and press Enter, and then enter a different invalid value (e.g., `b`) and press Enter, the `(invalid)` message above the widget appears after the first entry and then disappears after the second. Shouldn't the message remain present after the second entry?

philippjfr:
I think this was addressed by the linked PR.

PR #203, opened 2019-01-07
Fix reporting of bad inputs in LiteralInput widget (#202).

lebedov:


philippjfr:
Thanks for the PR. Would you mind adding a test to try to cover the case this is trying to fix?

lebedov:
Assuming I've understood the unit tests correctly, the last assert in the following code should fail on the bug I observed:
```python
from pyviz_comms import Comm
from panel.widgets import LiteralInput

doc = Document()
comm = Comm()

literal = LiteralInput(value=0, type=int, name='Literal')
box = literal._get_model(doc, comm=comm)
widget = box.children[0]
assert widget.title == 'Literal'
literal._comm_change({'value': "invalid"})
assert widget.title == 'Literal (invalid)'
literal._comm_change({'value': "also invalid"})
assert widget.title == 'Literal (invalid)'
```
However, I don't observe any error when I run it at the command line with unpatched Panel 0.3.1 and pyviz_comm 0.7.0 installed; I can only observe the error (i.e., the title changing from "Literal (invalid)" to "Literal" when I change the contents of the box from "invalid" to "also invalid")  if I create a widget in Jupyter Notebook or Lab as follows and try modifying the data in the displayed input box as described:
```python
import panel
from panel.widgets import LiteralInput
panel.extension()
w = LiteralInput(value=0, type=int, name='Literal')
w
``` 
Thoughts?

philippjfr:
I'll try to look into it, also worth noting that the bug doesn't occur on bokeh server so it's something going wrong when syncing the model state in the notebook.

philippjfr:
I'll merge this without the unit test for now, but will look into the syncing issues this PR highlights.

PR #204, opened 2019-01-08
Add appveyor build

xavArtley:
Example of build:
https://ci.appveyor.com/project/xavArtley/panel/builds/21457101


philippjfr:
Thanks looks good, will try that now.

PR #209, opened 2019-01-10
Replace special charactere of player.ts by their unicode equivalent

### LINKED ISSUE #208: Interaction between Panel, bokeh and Jupyter Notebook format prevents bokeh serve from running with charmap error
Hi,
I'm trying to run the following script using Holoviews (with Bokeh) and Panel in a .ipynb.

```python
import numpy as np
import panel as pn
import param
import holoviews as hv
hv.extension('bokeh')

def gaussian_func(x, x0, amplitude, sigma):
    return amplitude * np.exp(-(x - x0)**2/sigma)

class App(param.Parameterized):
    sigma = param.Number(default=0.1, bounds=(0, 20))
    
    def view(self):
        x = np.linspace(0, 10)
        vals = gaussian_func(x, 5, 1., self.sigma)
        return hv.Curve((x, vals))

app = App(name="Test app")
panel = pn.Row(app, app.view)
panel.servable()
```
If I copy this in a Jupyter Notebook cell, execute the content of the cell, save the result to a notebook called 'bokeh_serve_bug' and then try to serve this file with:
`bokeh serve --show "bokeh_serve_bug.ipynb"`
I get an error:
`ERROR: 'charmap' codec can't decode byte 0x9d in position 1052656: character maps to <undefined>`
Thanks to @xavArtley, we've been able to track down the bug to this by jumping to position 1052656:
`"              first.appendChild(document.createTextNode('❚◀◀'));\n",`

As far as I understand the problem, the analysis is then:
- the Jupyter notebook serializes the panel
- therefore including the JS sources for the Player element (unused here, but still included)
- the player uses a unicode character which maps to undefined
- causing bokeh serve to not work

From my point of view as an user, this is a problem since I expect the bokeh server to run based on the source code, and should ignore any output!

What should be done with this? How could it be fixed?

Thanks for your help

Florian

xavArtley:
Fix #208
Ensure windows compatibility
A simple test which failed on windows was trying to open().read() player.ts file

### DIFF
--- a/panel/models/player.ts
+++ b/panel/models/player.ts
@@ -58,43 +58,43 @@ export class PlayerView extends WidgetView {
 
     const first = document.createElement('button')
     first.style = button_style
-    first.appendChild(document.createTextNode('❚◀◀'))
+    first.appendChild(document.createTextNode('\u275a\u25c0\u25c0'))
     first.onclick = () => this.first_frame()
     button_div.appendChild(first)
 
     const previous = document.createElement('button')
     previous.style = button_style
-    previous.appendChild(document.createTextNode('❚◀'))
+    previous.appendChild(document.createTextNode('\u275a\u25c0'))
     previous.onclick = () => this.previous_frame()
     button_div.appendChild(previous)
 
     const reverse = document.createElement('button')
     reverse.style = button_style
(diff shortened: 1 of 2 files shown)


philippjfr:
Thanks!

PR #225, opened 2019-01-28
Python2 compatibility

philippjfr:


philippjfr:
>I'm glad there weren't more changes needed, if we haven't been running python2 tests!!!

Bit premature but I think I've got it now.

fmaussion:
Devil's advocate here. Was this really necessary? Xarray just merged a PR to *remove* py2 compat ;-)

philippjfr:
After going through all the effort I'd be very tempted to say no, but we'll probably drop py2 compatibility for all projects at the same time.

fmaussion:
Related: https://twitter.com/mrocklin/status/1083889299211223040

*"Here's a new one. The Dask-Kubernetes subproject just accepted a patch to support Python 2 (Dask's Kubernetes was Python 3 only), then cut a release, and immediately dropped Python 2 again."*

jbednar:
Panel replaces ParamNB and ParamBokeh, which both supported Python2, and I don't want people ever to have any reason to turn to those older projects.  When we drop Python2 support in Panel (which we will!), we can now safely point people who are stuck on py2 to use an older Panel release, rather than having them pointlessly invest in ParamNB or ParamBokeh.

PR #234, opened 2019-02-07
Updating meta.yaml to make it pure template

jsignell:
Getting rid of dependencies for test and nodejs requirement. If nodejs is needed it should be moved to setup.py

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -28,6 +28,46 @@ def get_setup_version(reponame):
         print("WARNING: param>=1.6.0 unavailable. If you are installing a package, this warning can safely be ignored. If you are creating a package or otherwise operating in a git repository, you should install param>=1.6.0.")
         return json.load(open(version_file_path, 'r'))['version_string']
 
+def build_custom_models():
+    """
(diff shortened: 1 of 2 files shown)


jbednar:
Sounds good to me!

philippjfr:
It is indeed required, doesn't need to be pinned though. Would be good to add it to ``extras_require['build']`` in this PR.

setup.py
```diff
@@ -28,6 +28,46 @@ def get_setup_version(reponame):
         print("WARNING: param>=1.6.0 unavailable. If you are installing a package, this warning can safely be ignored. If you are creating a package or otherwise operating in a git repository, you should install param>=1.6.0.")
         return json.load(open(version_file_path, 'r'))['version_string']
 
+def build_custom_models():
```
jsignell: I moved this up in the file to make the setup.py easier to read.

jbednar: Ok

jbednar: Merge at your leisure!

PR #238, opened 2019-02-12
Use parameter definition order for Python 3.6+

jlstevens:
Since Python dictionaries preserve order since 3.6 (and it became part of the CPython spec in 3.7), you no longer have to use parameter precedence to order your widgets as they can appear in definition order. This PR implements this behavior for Python 3.6+.

### DIFF
--- a/panel/param.py
+++ b/panel/param.py
@@ -359,7 +359,8 @@ class Param(PaneBase):
                     if ((p.precedence is None) or (p.precedence >= self.display_threshold))]
         groups = itertools.groupby(filtered, key=key_fn)
         # Params preserve definition order in Python 3.6+
-        dict_ordered = sys.version_info.major == 3 and sys.version_info.minor >= 6
+        dict_ordered_py3 = (sys.version_info.major == 3 and sys.version_info.minor >= 6)
+        dict_ordered = dict_ordered_py3 or (sys.version_info.major > 3)
         ordered_groups = [list(grp) if dict_ordered else sorted(grp) for (k,grp) in groups]
         ordered_params = [el[0] for group in ordered_groups for el in group]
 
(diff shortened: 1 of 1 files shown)


jlstevens:
@philippjfr I think this is ready to merge once the tests pass.

panel/param.py
```diff
@@ -357,8 +358,10 @@ def _get_widgets(self):
         filtered = [(k,p) for (k,p) in sorted_precedence
                     if ((p.precedence is None) or (p.precedence >= self.display_threshold))]
         groups = itertools.groupby(filtered, key=key_fn)
-        sorted_groups = [sorted(grp) for (k,grp) in groups]
-        ordered_params = [el[0] for group in sorted_groups for el in group]
+        # Params preserve definition order in Python 3.6+
+        ordered_params = sys.version_info.major == 3 and sys.version_info.minor >= 6
+        ordered_groups = [list(grp) if ordered_params else sorted(grp) for (k,grp) in groups]
+        ordered_params = [el[0] for group in ordered_groups for el in group]
```
jbednar: One more change: it's confusing to use ordered_params as both the boolean flag and as the set of ordered params.  Maybe rename the first one to `params_already_ordered`?

panel/param.py
```diff
@@ -357,8 +358,10 @@ def _get_widgets(self):
         filtered = [(k,p) for (k,p) in sorted_precedence
                     if ((p.precedence is None) or (p.precedence >= self.display_threshold))]
         groups = itertools.groupby(filtered, key=key_fn)
-        sorted_groups = [sorted(grp) for (k,grp) in groups]
-        ordered_params = [el[0] for group in sorted_groups for el in group]
+        # Params preserve definition order in Python 3.6+
+        ordered_params = sys.version_info.major == 3 and sys.version_info.minor >= 6
```
philippjfr: Since Guido says Python 4.0 will be the version after 3.9, let's include >=4.x here.

panel/param.py
philippjfr: I'd go with ``if dict_ordered``.

jlstevens: Fixed in 791778d

panel/param.py
jlstevens: Fixed in ec923e7

philippjfr:
Thanks for making the requested changes.

PR #250, opened 2019-02-22
Allow embedding app state

philippjfr:
This allows exporting panel objects with embedded state, much like a HoloMap in HoloViews. This works by searching for all widgets which define a fixed set of options, taking their cross-product,  evaluating the widget values and recording the corresponding app state. The app state can then be replayed by applying the state to the document whenever a widget value changes.

Here is a demo of some embedded state: https://anaconda.org/philippjfr/state_embed/notebook


panel/holoviews.py
```diff
@@ -210,7 +210,6 @@ def widgets_from_dimensions(cls, object, widget_types={}, widgets_type='individu
                 widgets.append(widget)
         if widgets_type == 'scrubber':
             widgets = [Player(length=nframes)]
-        print(widgets)
```
jbednar: Oops!

panel/util.py
```diff
@@ -290,6 +296,91 @@ def add_to_doc(obj, doc, hold=False):
...
+    restore = [w.value for w, _, _ in values]
+    init_vals = [m.value for _, m, _ in values]
+    cross_product = list(product(*[vals[::-1] for _, _, vals in values]))
+
+    if len(cross_product) > max_states:
+        raise RuntimeError('The cross product of different application '
+                           'states is too large to explore (N=%d), either reduce '
+                           'the number of options on the widgets or increase '
+                           'the max_states specified on static export.' %
+                           len(cross_product))
```
jbednar: Is there any hope that exceeding the max_states would be a warning rather than an error, such that instead of aborting Panel decimates the widget options included (at least the numeric ones) until max_states is satisfied?

PR #251, opened 2019-02-23
Minor fixes to docs and website

jbednar:
The Installation section on http://panel.pyviz.org was being formatted at the wrong level:

image

Also cleaned up some wording and clarified in the intro that there are three completely different ways of using Panel, and that users should pick one of them for any particular application.

### DIFF
--- a/doc/index.rst
+++ b/doc/index.rst
@@ -6,10 +6,9 @@
 A high-level app and dashboarding solution for Python
 -----------------------------------------------------
 
+Panel is an `open-source <https://github.com/pyviz/panel/blob/master/LICENSE.txt>`_ Python library that lets you create custom interactive web apps and dashboards by connecting user-defined widgets to plots, images, tables, or text.
 
-Panel is an `open-source <https://github.com/pyviz/panel/blob/master/LICENSE.txt>`_ Python library.  
-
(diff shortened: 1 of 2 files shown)


doc/index.rst
```diff
@@ -21,16 +20,11 @@ Panel makes it simple to make:
...
+Panel lets you move the same code freely between an interactive `Jupyter Notebook <http://jupyter.org>`__ prompt and a fully deployable standalone server.  That way you can easily switch between exploring your data, building visualizations, adding custom interactivity, sharing with non-technical users, and back again at any point, using the same tools and the same code throughout. Panel thus helps support your entire workflow, so that you never have to commit to only one way of using your data and your analyses, and don't have to rewrite your code just to make it usable in a different way.
 
-Some example Panel apps: (click on the title to see the code in each case)
+Some example Panel apps: (Click on the title to see the code, or the image to deploy on mybinder if it has resources available.)
```
philippjfr: The errors are not actually related to resources I think it's a race condition between the binder starting bokeh server and the client connecting to it, but I'm okay not explaining that and eventually replacing binder with AE or AWS deployments.

philippjfr:
Looks good, made one small comment but we can just gloss over that if you want.

doc/index.rst
jbednar: Sure. I was trying to think of a sufficiently vague way to say that errors should be expected, without getting into details. :-) I'll merge as-is, but if you can think of a short way to explain the issue better, please do! (And if you can think of a way to avoid the race condition, that would be even better!)

PR #258, opened 2019-03-01
Improved behavior for fixed widths and heights

philippjfr:
This PR forces a fixed height to be respected if specified either by setting sizing_mode='fixed' or by setting the corresponding `min_width` or `min_height` property at the same time.

Fixes https://github.com/pyviz/panel/issues/23


jsignell:
Fixes the issue I was seeing locally thanks!

PR #276, opened 2019-03-05
Refactored server handling

philippjfr:
This PR refactors the server handling such that all existing methods that start a bokeh server simply call the new ``Viewable.get_server`` method and then publish the server session in different ways. This also means that we now take full control over embedding servers inline in the notebook, which makes it possible to automatically clean up server sessions as cells are deleted or re-executed.

Depends on new versions of both the JLab extension and pyviz_comms: https://github.com/pyviz/pyviz_comms/pull/28

### DIFF
--- a/panel/__init__.py
+++ b/panel/__init__.py
@@ -4,6 +4,9 @@ import sys
 
 import param as _param
 from bokeh.document import Document as _Document
+from pyviz_comms import (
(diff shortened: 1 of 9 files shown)


philippjfr:
Fixes https://github.com/pyviz/panel/issues/231

PR #279, opened 2019-03-05
Adding some more list methods for layout and tests

### LINKED ISSUE #278: Rows and Columns should have the `extend` method
To update the contents of a Row or Columns, you can use list methods like append and pop. For instance:

```python
import panel as pn

col = pn.Column('123')
col.append('new')
col.pop(col[0])
col
```
screen shot 2019-03-05 at 2 32 10 pm

It'd be nice to also accept `col.extend`

jsignell:
Closes #278  (adding `extend`) and adds `clear` and `reverse` as well. I think the only other missing list methods are copy, count, index, and sort of which only sort affects the list in-place. Sort doesn't work on most panel objects and probably doesn't make much sense (might later if we wanted to compare names or something).

### DIFF
--- a/panel/layout.py
+++ b/panel/layout.py
@@ -12,7 +12,7 @@ from bokeh.models.widgets import Tabs as BkTabs, Panel as BkPanel
 
 from .io import state
 from .util import param_name, param_reprs, push
-from .viewable import Reactive, Viewable
+from .viewable import Reactive
 
 
 class Panel(Reactive):
@@ -326,7 +326,6 @@ class Tabs(Panel):
         return new_models
 
     def __setitem__(self, index, panes):
-        from .pane import panel
         new_objects = list(self)
         if not isinstance(index, slice):
             if index > len(self.objects):
--- a/panel/layout.py
+++ b/panel/layout.py
@@ -276,7 +276,7 @@ class Tabs(Panel):
         if isinstance(item, tuple):
             name, item = item
         else:
-            name = getattr(item, 'name')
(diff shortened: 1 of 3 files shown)


jbednar:
Sounds good. Sort would make sense for a column of widgets or anything else that has a name attribute, but I'm not sure if that's worth adding now.

philippjfr:
Thanks for this it's looking good. I agree we should punt on sort for now. 

Before merging this we will need a different implementation for Tabs because the methods must also accept tuples of the form ``(title, object)`` which lets you set a tab title independently of the object's name.

philippjfr:
Minor merge conflict, apart from that this is looking good.

jsignell:
Fixed the merge conflict. Passes tests locally -  so hopefully they'll pass on CI.

panel/layout.py
```diff
@@ -254,21 +267,27 @@ class Tabs(Panel):
     _linked_props = ['active']
 
     def __init__(self, *items, **params):
+        objects, self._names = self._to_objects_and_names(items)
+        super(Tabs, self).__init__(*objects, **params)
+        self.param.watch(self._update_names, 'objects')
+
+    def _to_object_and_name(self, item):
```
jsignell: This method and the one below are helper functions and don't depend on self, so they could be moved out of the class, but I thought it was clearer to have them as class methods.

philippjfr: Seems fine to me, could be classmethod but I don't really mind.

philippjfr:
I'll just go ahead and merge.

PR #290, opened 2019-03-08
Param formatting via the new label property

jlstevens:
Now that https://github.com/pyviz/param/pull/319 is merged, there is no need for the label formatter in panel to format parameter names. Merging this PR will bump the param version requirements so I will update the pinning too.

### DIFF
--- a/panel/param.py
+++ b/panel/param.py
@@ -266,8 +266,7 @@ class Param(PaneBase):
             widget_class = self.widgets[p_name]
         value = getattr(self.object, p_name)
 
-        kw = dict(value=value, disabled=p_obj.constant)
-        kw['name'] = p_obj.label
+        kw = dict(value=value, disabled=p_obj.constant, name=p_obj.label)
 
         if hasattr(p_obj, 'get_range'):
             options = p_obj.get_range()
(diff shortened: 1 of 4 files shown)


panel/param.py
```diff
@@ -216,7 +213,7 @@ def toggle_pane(change, parameter=pname):
                     pane = Param(parameterized, name=parameterized.name,
                                  _temporary=True, **kwargs)
                     if isinstance(self._expand_layout, Tabs):
-                        title = pname if self.label_formatter is None else self.label_formatter(pname)
+                        title = self.params(pname).label
```
jlstevens: Not sure this is correct....

```suggestion
                        title = self.object.param[pname].label
```

philippjfr:
Should cut a new param dev release and pin that in this PR.

panel/param.py
jlstevens: Thanks

panel/param.py
```diff
@@ -270,11 +267,7 @@ def widget(self, p_name):
         value = getattr(self.object, p_name)
 
         kw = dict(value=value, disabled=p_obj.constant)
-
-        if self.label_formatter is not None:
-            kw['name'] = self.label_formatter(p_name)
-        else:
-            kw['name'] = p_name
+        kw['name'] = p_obj.label
```
philippjfr: Just inline that in ``kw = dict(value=value, disabled=p_obj.constant, name=p_obj.label)``

jlstevens: Sure.

jlstevens: Fixed in ba512b0

PR #299, opened 2019-03-12
Adding a silencer to layouts

### LINKED ISSUE #285: Allow empty layouts
I think you mentioned this @philippjfr, but just to capture it. We should hide the bokeh warning since it can be very useful to create an empty layout and then add content to it.

screen shot 2019-03-06 at 12 52 42 pm

jsignell:
Not positive this is the right place - and will only work after https://github.com/bokeh/bokeh/pull/8739 lands

fixes #285

### DIFF
--- a/panel/layout.py
+++ b/panel/layout.py
@@ -13,6 +13,14 @@ from bokeh.models.widgets import Tabs as BkTabs, Panel as BkPanel
 from .util import param_name, param_reprs
 from .viewable import Reactive
 
+from bokeh.core.validation import silence
+from bokeh.core.validation.warnings import EMPTY_LAYOUT
+
+
+# Silence the bokeh warning for empty layouts since empty layouts
+# are a valid use case for panel
+silence(EMPTY_LAYOUT)
+
 
 class Panel(Reactive):
     """
--- a/panel/layout.py
+++ b/panel/layout.py
@@ -17,11 +17,6 @@ from bokeh.core.validation import silence
 from bokeh.core.validation.warnings import EMPTY_LAYOUT
 
 
-# Silence the bokeh warning for empty layouts since empty layouts
-# are a valid use case for panel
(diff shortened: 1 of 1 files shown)


jsignell:
This does affect bokeh after panel is imported - which is perhaps a bad and unintended consequence.

Screen Shot 2019-03-12 at 1 23 20 PM

jsignell:
We could move the silence call into the init method to at least delay this side effect

philippjfr:
Will have to wait until a new bokeh dev release is available.

panel/layout.py
```diff
@@ -107,6 +111,7 @@ def _cleanup(self, root):
         super(Panel, self)._cleanup(root)
         for p in self.objects:
             p._cleanup(root)
+        silence(EMPTY_LAYOUT, False)
```
philippjfr: Could you explain this? This seems weird, I can imagine creating a panel, displaying it, deleting the cell or replacing the panel in the plot which triggers cleanup and then a subsequent display of the panel would again have the warning enabled again.

jsignell: I left that in by mistake. I was trying to make it cleanup after itself. But this isn't the right way.

philippjfr:
It seems like some of my recent changes caused the warnings to be bypassed entirely, so before implementing this I think we need to reenable them since I assume at least some of them will be helpful for us.

jsignell:
I am leaning towards my first approach of silencing on import rather than on init - what do you think @philippjfr

philippjfr:
>I am leaning towards my first approach of silencing on import rather than on init - what do you think @philippjfr 

I agree, but currently it seems like the validation is not enabled at all with or without this PR. I think in refactoring some of the rendering code they are now bypassed or something.

philippjfr:
Afaik warnings are already silenced as a side-effect of the way the rendering is done.

PR #301, opened 2019-03-13
Add support for saving and loading embedded data from JSON

philippjfr:
Makes it possible to export the embedded state to JSON files which can be loaded independently. Unlike the equivalent functionality in HoloViews each state is stored as a separate file which means it can be loaded only when it is needed, allowing a huge state space to be exported with a minimal file size penalty.


jbednar:
Nice!

PR #303, opened 2019-03-14
Allow passing Parameter instances to Param pane

philippjfr:
This is a bit of convenience to easily add single parameter widgets, i.e. this code:

```python
 pn.pane.Param(parameterized, parameters=['parameter'], show_name=False)
```

 becomes:

```python
pn.pane.Param(parameterized.param.parameter)
```

### DIFF
--- a/panel/param.py
+++ b/panel/param.py
@@ -134,6 +134,11 @@ class Param(PaneBase):
     _rerender_params = []
 
     def __init__(self, object=None, **params):
+        if isinstance(object, param.Parameter):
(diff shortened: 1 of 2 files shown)


jbednar:
From the code, it looks like this would also work?

```
pn.Column(parameterized.param.param1, parameterized.param.param2)
```

That seems like a more common pattern than `pn.pane.Param(parameterized.param.param1)`.  But then there would be no label for each widget?

jbednar:
Sounds good!

PR #305, opened 2019-03-14
Reverse map discrete options using identity

philippjfr:
Using a dictionary to map options from their value to their label did not work well because the values may be non-hashable or hash inconsistently. Instead this PR switches to using the value's identity to do the reverse mapping which should work reliably.

### DIFF
--- a/panel/widgets/player.py
+++ b/panel/widgets/player.py
@@ -6,6 +6,7 @@ from __future__ import absolute_import, division, unicode_literals
 import param
 
 from ..models.widgets import Player as _BkPlayer
+from ..util import isIn, indexOf
 from .base import Widget
 from .select import SelectBase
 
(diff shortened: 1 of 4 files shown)


jsignell:
This seems like a more straightforward approach. Are there downsides? Is it slower?

jbednar:
Sounds good. Should Param do the same for selectors?

philippjfr:
I doubt it's much slower, if anything I'd expect it to be cheaper not to do the hashing. That said the approach isn't quite right since ``__contains__`` will fall back to testing for equality which will break for certain types (e.g. numpy arrays). I'll have to add a small ``is_in`` utility.

panel/widgets/player.py
```diff
@@ -70,16 +70,16 @@ class DiscretePlayer(PlayerBase, SelectBase):
     _rename = {'name': None, 'options': None}
 
     def _process_param_change(self, msg):
-        values = [hashable(v) for v in self.values]
+        values = [v for v in self.values]
```
philippjfr:
```suggestion
        values = self.values
```

philippjfr:
>Sounds good. Should Param do the same for selectors?

Yes, the ``named_objs`` implementation in param seems to suffer from the same issues.

PR #311, opened 2019-03-16
Add a KaTeX and MathJax based LaTeX pane

philippjfr:
I think this should probably replace the LaTeX pane because it is much more capable. That said I still need to assess how expensive it would be to load KaTeX all the time, if it's too large I will likely have to let the user request it explicitly using ``pn.extension('katex')`` (not required on bokeh server or when saving).

Screen Shot 2019-03-16 at 2 45 36 AM

### DIFF
--- a/examples/gallery/panes/LaTeX.ipynb
+++ b/examples/gallery/panes/LaTeX.ipynb
@@ -1,0 +1,115 @@
+{
+ "cells": [
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "import panel as pn\n",
+    "\n",
+    "pn.extension('katex', 'mathjax')"
+   ]
+  },
+  {
+   "cell_type": "markdown",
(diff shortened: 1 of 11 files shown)


philippjfr:
katex.min.js + katex.min.css are ~137 KB combined. Is it worth it to always bundle that or should I require the user to run ``pn.extension('katex')`` before using the component in the notebook? A helpful warning would tell the user they forgot to do so. I'm on the edge, I don't quite think using LaTeX is common enough to make everyone pay a 137 KB penalty. If I make it optional I could also allow toggling between MathJax and KaTeX and use whichever one is loaded.

jbednar:
I'd say it's worth saving the 137KB at a cost of having to load the extension explicitly

philippjfr:
Okay, I've now implemented MathJax as well and refactored in such a way that the LaTeX pane will use either the MathJax or the KaTeX model depending on which was loaded. If an explicit renderer is defined on the pane it will use that and if neither is loaded it will warn.

PR #313, opened 2019-03-16
Improvements and Documentation for Panes

philippjfr:
This PR provides polish to all existing pane classes and then adds reference sections for each. I'll outline all the panes and the improvements made here:

* HoloViews
  - [x] Add support for widget state overrides as dictionary in widgets parameter (fixes https://github.com/pyviz/panel/issues/302)
  - [x] Ensure that initial plot state matches widget state
  - [x] Add reference gallery entry
* Plotly
  - [x] Fixed and improved efficiency of updates when only data or only layout properties are changed
  - [x] Ensure it is rendered correctly on initialization
  - [x] Fixes issues sending arrays of different shapes (fixes https://github.com/pyviz/panel/issues/226)
  - [x] Add reference gallery entry
* Vega
  - [x] Ensure it is rendered correctly on initialization
  - [x] Add reference gallery entry

### DIFF
--- a/doc/_templates/layout.html
+++ b/doc/_templates/layout.html
@@ -46,10 +46,6 @@
   <script type="text/javascript" async
   src="https://cdnjs.cloudflare.com/ajax/libs/mathjax/2.7.1/MathJax.js?config=TeX-MML-AM_CHTML">
   </script>
-  <script src="https://cdn.plot.ly/plotly-latest.min.js"></script>
-  <script src="https://cdn.jsdelivr.net/npm/vega@4.2.0?noext"></script>
-  <script src="https://cdn.jsdelivr.net/npm/vega-lite@3.0.0-rc4?noext"></script>
-  <script src="https://cdn.jsdelivr.net/npm/vega-embed@3.18.2?noext"></script>
   <script src="{{ pathto('_static/' + theme_js, 1) }}"></script>
   {% for link in js_includes %}
      <script src="{{ pathto('_static/' + link, 1) }}"></script>
--- a/examples/gallery/panes/HoloViews.ipynb
+++ b/examples/gallery/panes/HoloViews.ipynb
@@ -1,0 +1,204 @@
+{
+ "cells": [
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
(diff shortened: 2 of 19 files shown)


philippjfr:
katex.min.js + katex.min.css are ~137 KB combined. Is it worth it to always bundle that or should I require the user to run ``pn.extension('katex')`` before using the component in the notebook? A helpful warning would tell the user they forgot to do so. I'm on the edge, I don't quite think using LaTeX is common enough to make everyone pay a 137 KB penalty.

philippjfr:
So after refactoring the various models I now realize that loading the JS in this does not work for the classic notebook as noted in https://github.com/bokeh/bokeh/issues/5227.

Not sure what to do about this yet, but it may require a fallback condition implemented using jQuery or similar.

philippjfr:
In the worst case scenario which is that we find no non-hacky solution that could be added to bokeh itself I will likely just insert this hack in our JS:

```javascript
if (window.requirejs) {
  window._define = window.define;
  // Stub does not define amd and is therefore skipped by UMD loaders
  // however Notebook bundle does not check for window.define.amd
  function _define_stub(name, deps, callback) {
    window._define(name, deps, callback)
  }
  window.define = _define_stub;
  function _reset_define() {
    console.log("RESET")
    window.define = window._define;
  }
  window._bokeh_onload_callbacks.push(_reset_define)
}
```

It temporarily replaces ``window.define`` with a stub which is not recognizable by a UMD loader but will still be used to fetch the classic notebook webpack bundle.

philippjfr:
I decided to trial a fix for the issue loading dependencies in classic notebooks by adding a ``__js_require__`` attribute to custom models, which is used in the autoload_panel_js.js template to generate a requirejs call which satisfies the dependencies. If this works well for us I'll fully flesh out the ``__js_require__`` API and contribute it upstream.

@xavArtley If you want the VTK model to work in the notebook you'll be able to add something like this to your VTK model to load vtk.js correctly in the classic notebook:

```python
__js_require__ = {"paths": {"vtk": "//unpkg.com/vtk.js@8.3.15/dist/vtk"},
                             "shim": {"vtk": {"exports": "vtk"}}}
```

PR #314, opened 2019-03-17
Simple Iris dataset Visualization

rtmatx:
Hi,

I have created a simple visualization using Iris Dataset similar to the visualization available at [https://shiny.rstudio.com/gallery/kmeans-example.html](https://shiny.rstudio.com/gallery/kmeans-example.html).

Let me know the format the I should follow to create few more examples for the gallery section. I want to start with something simple following the guidelines you have mentioned here [https://github.com/pyviz/panel/issues/286](https://github.com/pyviz/panel/issues/286)

Thanks
Ravi


jbednar:
Thanks!  Are you planning to extend it to include k-means clustering, using the import from sklearn that's already there and following the Shiny example?  If so I'll hold off on review until that's added.

jbednar:
I've done a pass over it for you to use if you see fit (rename to remove .txt):

[Iris_dataset.ipynb.txt](https://github.com/pyviz/panel/files/2975401/Iris_dataset.ipynb.txt)

Here I've replaced the numerical labels with the species name, and cleaned up the descriptions a bit. It's not ready to merge, because the text is stating that it's using clustering, but it currently isn't. I think that it should either keep using sklearn but add clustering, or else it should use the iris dataset from Bokeh (see e.g. http://holoviews.org/gallery/demos/bokeh/iris_example.html) and avoid having to install and import sklearn.  Either approach is fine, but right now it's stuck in the middle, so it will have go in one direction or the other.

@philippjfr, I'm not sure about the programming style to encourage here.  Seems like this example is simple enough it could just use `interact`, maybe?

jbednar:
Oh, and I changed `name='Iris Dashboard'` to `name='Iris_Dashboard'` to avoid messed-up formatting for the name string:

image

I think that may be fixed in some later Bokeh version than my 1.1.0dev8, so that change may no longer be needed.

rtmatx:
> Thanks! Are you planning to extend it to include k-means clustering, using the import from sklearn that's already there and following the Shiny example? If so I'll hold off on review until that's added.

Yeah. I would like to check if this is the format that I have to follow before proceeding further on this example or any others than I will do later. My idea is to create visualization available at shiny and show them in Panel. Please correct me on this, if the objective is different

rtmatx:
I have updated the notebook to use bokeh dataset and then developed the panel dashboard just to show `selector` widget and `hvplot` scatter functionality

philippjfr:
>My idea is to create visualization available at shiny and show them in Panel. Please correct me on this, if the objective is different

I think they provide good templates, but the goal isn't simply to replicate them but to demonstrate different aspects of the library.

>I'm not sure about the programming style to encourage here. Seems like this example is simple enough it could just use interact, maybe?

This is a good question, I think there should be sections for interact, param and lower-level usage. I'm not sure how to decide which one is more appropriate for a particular example. We could decide to show off different approaches in one notebook, or decide arbitrarily which one to us, or make a separate notebook for each approach.

jbednar:
> but to demonstrate different aspects of the library.

I think we can have a slightly larger goal as well, i.e. to have compelling, exciting examples that motivate people. Even if a given example doesn't show off some particular novel feature of the library compared to other examples, if it's something that inspires people to think that they should be trying out, it's still worthwhile.

I'm not sure how best to handle the different interfaces either.

philippjfr:
Could we move this example to ``examples/gallery/param`` then I'd be happy to merge and refine it afterwards.

philippjfr:
Thanks, I'll probably take a look at the gallery over the weekend.

PR #334, opened 2019-03-25
Allow using a 4 tuple of min, max, step, value for interact

### LINKED ISSUE #333: Interact should accept tuple of min, max, and value as input
At the moment interact only supports passing value OR (min, max). It would be very helpful to be able to pass all three. For context - this came up in the context of making widgets from intake user parameters which have default values in addition to min, and max.

jsignell:
Fixes #333

Allows step to be None


philippjfr:
Looks good to me. Ready to merge?

jsignell:
I was adding a cell to the interact notebook. Think it's necessary?

jbednar:
Yes, please do, because this is functionality beyond what ipywidgets supports and thus won't be expected by many people.

jsignell:
I did a little formatting cleanup in the options table and cleaned up the notebook state.


## Before
Screen Shot 2019-03-25 at 2 19 17 PM

## After
Screen Shot 2019-03-25 at 2 21 21 PM

jsignell:
I'm ready for this to go in now.

philippjfr:
Definitely not needed, I think the jq based script ignores it or I was lazy about clearing it.

PR #337, opened 2019-03-27
Refactored and improved VTK model and pane [doc-build]

philippjfr:
Does some refactoring for the VTK pane and adds some features:

* VTK pane now supports responsive sizing modes
* VTK pane now supports syncing (and therefore linking) camera state

@xavArtley, review would be appreciated but I'll probably merge some time tomorrow in any case.

@dharhas, it currently only supports uni-directional links between objects, which is something I need to think about and address in general, but you could now link multiple plots together and then interact with the plot they are linked to:

vtk_link

### DIFF
--- a/panel/models/vtk.ts
+++ b/panel/models/vtk.ts
@@ -1,39 +1,46 @@
 import * as p from "core/properties"
+import {clone} from "core/util/object";
 import {HTMLBox, HTMLBoxView} from "models/layouts/html_box";
 import {div} from "core/dom";
 
 export class VTKPlotView extends HTMLBoxView {
   model: VTKPlot
   protected _vtk: any
-  protected _rendererEl: any
   protected _container: HTMLDivElement
-  protected _vtk_initialized: boolean
+  protected _rendererEl: any
+  protected _renderer: any
+  protected _camera: any
+  protected _setting: boolean = false
 
   initialize(): void {
     super.initialize()
-    const width = this.model.width ? "100%" : "300px"
-    const height = this.model.height ? "100%" : "300px"
     this._vtk = (window as any).vtk
     this._container = div({
       style: {
-        width,
-        height
+        width: "100%",
+        height: "100%"
       }
-    })
-    this._vtk_initialized = false
+    });
   }
 
-
-  after_layout() {
+  after_layout(): void {
     super.after_layout()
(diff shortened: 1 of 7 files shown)


xavArtley:
In notebook the vtk panel does not appear for me
image
If I use the show metho it seems to work
image

philippjfr:
Is that JupyterLab or classic notebook? I can't reproduce issues in either.

xavArtley:
jupyter lab

philippjfr:
Ah, have you updated the JupyterLab extension recently? We made a major change to the way output is published in the notebook.

philippjfr:
You'll need:

@pyviz/jupyterlab_pyviz v0.7.1

xavArtley:
Ok greats it works now.

philippjfr:
Actually bi-directional linking works okay, and there's no need for the trigger:

```
vtk = pn.pane.VTK('https://raw.githubusercontent.com/Kitware/vtk-js/master/Data/StanfordDragon.vtkjs',
                  sizing_mode='stretch_both')
vtk2 = pn.pane.VTK('https://raw.githubusercontent.com/Kitware/vtk-js/master/Data/StanfordDragon.vtkjs',
                   sizing_mode='stretch_both')

vtk.jslink(vtk2, camera='camera')
vtk2.jslink(vtk, camera='camera')

pn.Row(vtk, vtk2)
```

I'll probably add a bi-directional flag to the ``jslink`` method.

philippjfr:
@xavArtley Finally got vtk to run on Travis so we'll be able to add some more tests.

panel/models/vtk.ts
```diff
@@ -1,60 +1,90 @@
...
+  _get_camera_state(): void {
+    if (!this._setting) {
+      this._setting = true;
+      const state = clone(this._camera.get());
+      delete state.classHierarchy;
+      delete state.vtkObject;
+      delete state.vtkCamera;
+      delete state.viewPlaneNormal;
+      this.model.camera = state;
+      this.model.properties.camera.change.emit();
```
xavArtley: Is it really necessary to emit change explicitly?
As we set a new value to model.camera, I would excpect the change event would be already emit

philippjfr: I think you're right, thanks!

PR #338, opened 2019-03-27
Add GridSpec layout [doc-build]

philippjfr:
Implements a ``GridSpec`` layout based on the new GridBox model in bokeh. It works quite differently from the previous approach, you no longer have to declare the number of cols and rows, you simply declare a size (or sizing_mode='stretch_both') and then start filling the grid:

Screen Shot 2019-03-27 at 6 53 39 PM

Screen Shot 2019-03-27 at 6 53 55 PM

Supersedes https://github.com/pyviz/panel/pull/31

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -54,6 +54,7 @@ jobs:
       install:
         - doit env_create $CHANS_DEV --python=$PYENV_VERSION
         - source activate test-environment
+        - conda install -c conda-forge mesalib
         - doit develop_install -o recommended -o tests -o build $CHANS_DEV
         - python setup.py develop --no-deps
         - doit env_capture
--- a/doc/user_guide/index.rst
+++ b/doc/user_guide/index.rst
@@ -43,6 +43,7 @@ Supplementary guides
     :maxdepth: 2
 
     Introduction <Introduction>
+    Components <Components>
     Deploy & Export <Deploy_and_Export>
     Interact <Interact>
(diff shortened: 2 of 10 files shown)


jbednar:
Seems very cool! We should have a demo example using this approach, with some numeric readouts and other things that aren't plots, as in the various dashboards you find if you search for "dashboard" images.

philippjfr:
Here's a more complex example:

Screen Shot 2019-03-28 at 4 27 21 AM

and here's an error when trying to assign to a region of the grid that's already occupied:

```
IndexError: Specified region overlaps with the following existing object(s) in the grid:

    (0, 0): HTML(str, background='#FF0000')    (1, 0): HTML(str, background='#00FF00')    (1, 1): VTK(str, camera={'position': [-21.49009035...})

The following shows a view of the grid (empty: 0, occupied: 1, overlapping: 2):

[[2 2 2]
 [2 2 2]
 [2 2 2]
 [1 1 0]
 [1 1 1]]
```

jbednar:
Is there a simple way to center content in each grid cell's region, both vertically and horizontally?  Presumably no one would want A, B, C, D to be in the upper left like that.

philippjfr:
> Is there a simple way to center content in each grid cell's region, both vertically and horizontally? 

There is probably some combination of CSS that will do it, worth looking into but not tied up with this PR.

PR #355, opened 2019-04-01
[vtk] correction of the texture serialization

xavArtley:
fixes :  https://github.com/pyviz/panel/issues/350

However when renderding the vtkjs panel the texture does not appear unless an interaction is done
I tried to export the vtkjs file into the official scene explorer (https://kitware.github.io/vtk-js/examples/SceneExplorer.html) and the same issue arise

ezgif com-video-to-gif (5)

May be we need to open an issue on the main repo

### DIFF
--- a/panel/models/vtk.ts
+++ b/panel/models/vtk.ts
@@ -89,10 +89,11 @@ export class VTKPlotView extends HTMLBoxView {
           renderer: this._rendererEl.getRenderer(),
           dataAccessHelper,
         })
-        sceneImporter.setUrl('index.json');
-        sceneImporter.onReady(() => {
-          setTimeout(() => this._rendererEl.getRenderWindow().render(),100)
-        })
+        const fn = this._vtk.macro.debounce(() => setTimeout(
+          this._rendererEl.getRenderWindow().render,
+           100), 100)
+        sceneImporter.setUrl('index.json')
+        sceneImporter.onReady(fn)
       }
     })
   }
--- a/panel/models/vtk.ts
+++ b/panel/models/vtk.ts
@@ -91,7 +91,7 @@ export class VTKPlotView extends HTMLBoxView {
         })
         sceneImporter.setUrl('index.json');
         sceneImporter.onReady(() => {
-          this._rendererEl.getRenderWindow().render()
+          setTimeout(() => this._rendererEl.getRenderWindow().render(),100)
         })
       }
     })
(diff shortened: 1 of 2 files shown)


jourdain:
I guess a call to render is missing when images have been downloader+decoded and connected to the pipeline. You could solve that issue, by triggering a render() call via a timeout.

xavArtley:
How long should be the timeout?

> I guess a call to render is missing when images have been downloader+decoded and connected to the pipeline. You could solve that issue, by triggering a render() call via a timeout.

jourdain:
Can you point me to the section of code that do the loading. I might be able to provide a better solution than a timeout with a more appropriate callback. 

Otherwise you can simply do:

```
setTimeout(rw.render, 100);
setTimeout(rw.render, 500);
setTimeout(rw.render, 1000);
```

xavArtley:
https://github.com/pyviz/panel/blob/67b90c42cb382c38d2ae5355f46d6bfdb217d0d3/panel/models/vtk.ts#L85-L97

jourdain:
The onReady will be called several time (on for each dataset) so this could be good.
The setURL also return a promise, although image loading/decoding is indeed trickier to properly capture when everything is ready.

Can you try that?

```
const fn = macro.debounce(() => setTimeout(this._rendererEl.getRenderWindow().render, 100), 100);
sceneImporter.setUrl('index.json').then(fn);
sceneImporter.onReady(fn);
```

xavArtley:
Uncaught (in promise) TypeError: Cannot read property 'then' of undefined
    at Object.callback (<anonymous>:663:57)
    at VM57184 vtk.js:7

setUrl return undefined  apparently. However if I use just 
```
const fn = macro.debounce(() => setTimeout(this._rendererEl.getRenderWindow().render, 100), 100);
sceneImporter.setUrl('index.json')
sceneImporter.onReady(fn);
```
it seems to work

jourdain:
Great!

banesullivan:
This is awesome, thanks @xavArtley!

Once this is merged could there be a release staged for `panel`? As soon as this is available on conda, I could update all of [these examples](https://github.com/OpenGeoVis/PVGeo-Examples) and get the [`vtki` examples](http://docs.vtki.org/en/latest/examples/index.html) on Binder to utilize all your work!

philippjfr:
>Once this is merged could there be a release staged for panel?

I'm happy to cut a dev release, a real release is still waiting on a new bokeh release which should be very soon.

philippjfr:
Thanks for the fix @xavArtley!

PR #359, opened 2019-04-02
Ace code editor

xavArtley:
This PR add a panel of type code editor
It's based on this library https://github.com/ajaxorg/ace/

I have activated the language and theme selection
basic autocompletion with ctrl+space
and possibility to add anotations
ezgif com-video-to-gif (6)


philippjfr:
@xavArtley I'd be happy to merge this once there is a reference gallery entry. At some point we will have to have a discussion which components make sense in panel itself and which could be added somewhere else.

xavArtley:
Indeed I think this model should be external to panel more in some custom extensions

philippjfr:
We'll need to work out a plan for managing external components like this.

philippjfr:
I'm actually happy to merge this now and move it later if we decide on a repo to extend panel. Does that sound okay @xavArtley ?

xavArtley:
Ok for me

PR #368, opened 2019-04-04
add spinner widget

xavArtley:
image

### DIFF
--- a/panel/widgets/input.py
+++ b/panel/widgets/input.py
@@ -131,11 +131,11 @@ class ColorPicker(Widget):
 
 class Spinner(Widget):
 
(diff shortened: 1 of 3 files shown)


panel/widgets/input.py
```diff
@@ -129,6 +129,25 @@ class ColorPicker(Widget):
     _rename = {'value': 'color', 'name': 'title'}
 
 
+class Spinner(Widget):
+
+    low = param.Number(default=None, doc="""
```
philippjfr: Let's make these consistent with other similar widgets and call them `start` and `end`.

philippjfr:
Looks great thanks, one minor comment otherwise I'm happy to merge.

PR #379, opened 2019-04-12
Vtk key binds option

### LINKED ISSUE #378: [vtk] keypress event bug
When using `panel`'s new VTK functionality in a Joptyer notebook, the panel listens to key events outside of its scope. 

When writing new code in a different cell, buttons like `v` and `s` alter the scene - this is unexpected behavior. See example below using `vtki` which sends its renderer to panel's VTK code:

ezgif com-video-to-gif-3

xavArtley:
fixes #378
correct a display error in the doc reference introduced by #372 :
image

image


panel/models/vtk.py
```diff
@@ -31,6 +31,8 @@ class VTKPlot(HTMLBox):
 
     camera = Dict(String, Any)
 
+    keybinds = Bool(default=True)
```
philippjfr: Since this causes such issues let's default to False and call it ``keybinding``, or maybe @jbednar has a better suggestion for naming?

philippjfr:
Thanks for the other fix, I wondered what happened there.

panel/models/vtk.py
jbednar: I agree it should be false by default; people can enable it in a standalone dashboard but it seems very unlikely to be useful in a notebook until it can be made to respect focus.  Seems like it could have a very descriptive name like `capture_keypress_events`; it won't come up often and something like that will be very clear.

philippjfr: ``enable_keybindings``?

jbednar: That's fine, but to me it doesn't convey that it will be listening to all keypresses.  Whereas if someone explicitly enables `capture_keypress_events`, they surely won't come complaining to us when they see it doing just that. :-)  But I'll leave it to your discretion.

philippjfr: I think that's okay for an option you have to enable explicitly, as long as it's properly documented that is.

panel/pane/vtk.py
```diff
@@ -331,6 +331,8 @@ class VTK(PaneBase):
 
     camera = param.Dict(doc="""State of the rendered VTK camera.""")
 
+    keybinds = param.Boolean(default=True, doc="""Activate/Deactivate keys binding""")
```
philippjfr: Could we add a quick note about the issues with this here?

panel/models/vtk.py
philippjfr: @xavArtley Sorry to have been confusing here, I think ``enable_keybindings`` is the best name for the parameter/property.

philippjfr:
Thanks, will merge shortly.

PR #394, opened 2019-04-23
Adding tests to make sure the reference gallery is complete

jsignell:
I thought it'd be useful to have a unit test that asserts that for every panel object (layout, widget, pane) there is a corresponding reference gallery entry.

Found that ToggleGroup, YT, and RGGPlot are missing. Allowing those ones to be missing for now.

### DIFF
--- a/panel/tests/test_docs.py
+++ b/panel/tests/test_docs.py
@@ -1,0 +1,47 @@
+"""
+These tests verify that all of the panes, layouts, and widgets defined by panel are
+represented in the reference gallery.
+"""
+import os
+from inspect import isclass
+import panel as pn
+
+
+here = os.path.abspath(os.path.dirname(__file__))
+ref = os.path.join(here, '..', '..', 'examples', 'reference')
+
+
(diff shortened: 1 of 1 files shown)


jbednar:
The YT and RGGPlot ones have requirements that we are unlikely to want to fulfill for our usual website building, so I'm not sure how to handle those.  ToggleGroup seems like just an oversight; good catch!

philippjfr:
ToggleGroup is a factory class which creates the appropriate widget type depending on two options. I'd be in favor of deprecating it to be honest.

philippjfr:
Should we just have a list of exclusions?

jbednar:
The PR already has exclusions; just merge then?

philippjfr:
I'll rebase and see if it passes.

jsignell:
Sorry I thought this got in a while ago. Seems like an unrelated failure. Should I merge?

philippjfr:
Yes feel free to merge.

philippjfr:
Sadly these tests don't work while packaging which means they are currently breaking releases.

PR #395, opened 2019-04-24
Add Video Stream widget to capture local input video

xavArtley:
I open this pull request to discuss about a widget to retrieve video stream.
For the moment this widget answer some personnal needs for a work project where I need to have the video stream and get a snapshot of this stream on request.

ezgif com-video-to-gif (9)

However I'm not really good at thinking about user friendly API

Moreover I have some problems with the bokeh layout. The width and height are not reactive with the size of the video


philippjfr:
The API seems okay but the parameter docstrings could be expanded on a bit. I'll try to take a look to see if I can get responsive behavior working. Did you try similar tricks like you did for the VTK model already, e.g. by creating a container with 100% width and height?

philippjfr:
Perhaps we can somehow add a method on the ``VideoStream`` pane that triggers a snapshot.

philippjfr:
Yes, I think that's the right approach, I'm going to add both a timeout parameter, which refreshes the image at a fixed interval and a snapshot method which syncs the state only when requested.

PR #402, opened 2019-04-25
Getting rid of duplicate `python setup.py develop`

jsignell:
Now that pyctdev uses `python setup.py develop --no-deps` in pyctdev, we no longer need to duplicate that line in the travis file.


philippjfr:
Thanks, good we no longer have to do this. I never quite understood why pip didn't trigger a normal dev install.

jsignell:
Looks like this is working.

philippjfr:
Thanks @jsignell!

PR #409, opened 2019-05-03
Add possibility to set custom vtkjs serializer

xavArtley:
In reference to https://github.com/pyviz/panel/issues/408
I separate the vtkjs serialization implementation from the pane implementation

I add a method set_serializer which register a custom serializer.

I add a try catch to wait for vtki serializer implementation. If it fails (and it fails for the moment). I register the actual serializer

a serializer expect a vtk render window as input and the vtkjs file binary stream output

I factorized the base64 encoding to send the stream to vtkjs

the _vtk_serializer.py should not be in pane package but I don't know where to put it

### DIFF
--- a/panel/pane/vtk/__init__.py
+++ b/panel/pane/vtk/__init__.py
@@ -1,9 +1,2 @@
 
 from .vtk import VTK
-
-try:
-    import vtk
-    from .vtkjs_serializer import render_window_serializer
-    VTK.set_serializer(render_window_serializer)
-except ImportError:
-    pass
--- a/panel/pane/vtk/__init__.py
+++ b/panel/pane/vtk/__init__.py
@@ -1,0 +1,9 @@
+
+from .vtk import VTK
+
+try:
+    import vtk
+    from .vtkjs_serializer import render_window_serializer
+    VTK.set_serializer(render_window_serializer)
+except ImportError:
+    pass
--- a/panel/pane/vtk/__init__.py
+++ b/panel/pane/vtk/__init__.py
@@ -1,2 +1,2 @@
 
-from .vtk import VTK
+from .vtk import VTK # noqa
--- a/panel/pane/vtk/vtk.py
+++ b/panel/pane/vtk/vtk.py
@@ -19,7 +19,7 @@ import param
 
 from pyviz_comms import JupyterComm
 
-from .base import PaneBase
+from ..base import PaneBase
 
 if sys.version_info >= (2, 7):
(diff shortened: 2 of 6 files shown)


philippjfr:
>the _vtk_serializer.py should not be in pane package but I don't know where to put it

I'd be fine making a vtk sub-package in pane where this could go.

What's the status of this PR otherwise?

philippjfr:
@xavArtley Any updates on this? I know you were/are on vacation but is there any chance this could get into a release due this weekend?

xavArtley:
I have a day off tomorrow I'll worked on it

<https://www.avast.com/sig-email?utm_medium=email&utm_source=link&utm_campaign=sig-email&utm_content=webmail>
Garanti
sans virus. www.avast.com
<https://www.avast.com/sig-email?utm_medium=email&utm_source=link&utm_campaign=sig-email&utm_content=webmail>
<#DAB4FAD8-2DD7-40BB-A1B8-4E2AA1F9FDF2>

Le jeu. 30 mai 2019 à 11:38, Philipp Rudiger <notifications@github.com> a
écrit :

xavArtley:
I should work as before but with possibility to change the serializer

panel/pane/vtk/__init__.py
```diff
@@ -0,0 +1,9 @@
+
+from .vtk import VTK
+
+try:
+    import vtk
```
philippjfr: Large libraries like vtk should not be imported until the VTK pane is actually used.

panel/pane/vtk/vtk.py
```diff
@@ -0,0 +1,107 @@
...
+            if os.path.isfile(self.object):
+                with open(self.object, 'rb') as f:
+                    vtkjs = f.read()
+            else:
+                data_url = urlopen(self.object)
+                vtkjs = data_url.read()
+        elif hasattr(self.object, 'read'):
+            vtkjs = self.object.read()
+        else:
+            vtkjs = VTK._serializer(self.object).read()
```
philippjfr: I'd suggest this should simply check if `_serializer` is None and then fall back to the inbuilt ``render_window_serializer``.

philippjfr:
Thanks for the updates, made some small review comments, but otherwise this is ready to merge.

philippjfr:
Looks good to me, will merge if it passes this time.

philippjfr:
Thanks for doing this @xavArtley, hugely appreciated!

PR #432, opened 2019-05-14
Add ParamFunction pane

philippjfr:
As outlined in https://github.com/pyviz/param/pull/334 this PR adds support for including reactive functions in Panel (i.e. functions which re-render when one of their dependencies change). This relies on an extension of ``param.depends`` which allows annotating functions with parameter objects.

This makes it possible to build reactive apps without requiring a user to define a whole class. A simple example is the following implementation of a stock selector:

reactive_plot

- [x] Add tests
- [ ] Add documentation
- [x] Add examples


jbednar:
Looks great. I think the docs should have a non-trivial example implemented using all of the now five different Panel development styles:

1. `interact`
2. `interact` but re-packed into a compact dashboard
3. `@depends` with bare functions
4. `@depends` with a Parameterized class
5. `.watch` with bare widgets

Having the same thing done each way will help people figure out the similarities and differences between these approaches, to choose the right one for their own projects.

jbednar:
> I strongly agree, not quite sure how far to push it on the non-trivial front yet though.

Well, it needs to be something where there is a reason not to simply use `interact`, i.e. some benefit or visible difference from the non-interact approaches.

philippjfr:
>Well, it needs to be something where there is a reason not to simply use interact, i.e. some benefit or visible difference from the non-interact approaches.

Apart from the more complex things you can do, I think the main reason you might prefer it over interact is that it's more explicit directly tying widget to computation to output. The comparison with `.watch` is easy because reactive patterns are almost always preferable to callback hell.

philippjfr:
In the end ``interact`` is very similar except it does extra magic by auto-generating the widgets and automatically setting up the dependencies. In fact from a code perspective it may be cleaner if I reimplemented ``interact`` on top of this PR so it's not duplicating the same thing.

jbednar:
Reimplementing in that way like the right approach to me, and will make it simpler to explain to people (both to explain `interact` itself and to explain how the approaches relate to each other).

philippjfr:
>Reimplementing in that way like the right approach to me

Just had a look, really not worth it from a code perspective but doesn't mean we can't explain it that way in the docs.

philippjfr:
Going to merge this and improve the docs later.

PR #465, opened 2019-05-31
Add tight option bbox

ahuang11:
https://github.com/pyviz/panel/issues/454

```
import panel as pn
from mpl_toolkits.axes_grid1 import make_axes_locatable
pn.extension()

fig = plt.figure(figsize=(8, 5))
ax = plt.axes()
im = ax.scatter([0, 1, 2], [3, 4, 5], c=[5, 6, 7])
ax.set_title('test')
ax.set_xlabel('someth', fontsize=40)
ax.set_ylabel('someth', fontsize=40)
divider = make_axes_locatable(ax)
cax = divider.append_axes("right", size="5%", pad=0.05)
cbar = plt.colorbar(im, label='colorbar', cax=ax)
plt.suptitle('somethingsomething loooong somethingsomething loooongsomethingsomething loooongsomethingsomething loooong')

pn.Pane(fig)

pn.Pane(fig, tight=True)
```

top is default render from Jupyter, second is wrapped fig with panel, and third is wrapped fig again but with tight
image

### DIFF
--- a/examples/reference/panes/Matplotlib.ipynb
+++ b/examples/reference/panes/Matplotlib.ipynb
@@ -15,11 +15,12 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "The ``Matplotlib`` pane allows displaying any displayable matplotlib figure inside a Panel app. It will render the plot to png at the declared DPI and then embed it.\n",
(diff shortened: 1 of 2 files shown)


philippjfr:
Looks good thanks!

PR #466, opened 2019-06-01
User guide review [doc-build]

jbednar:
Reviewing all user guide sections:

- [x] index
- [x] Overview.ipynb
- [x] Components.ipynb
- [x] APIs.ipynb
- [x] Customization.ipynb
- [x] Deploy_and_Export.ipynb
- [x] Interact.ipynb
- [x] Widgets.ipynb
- [x] Param.ipynb
- [x] Links.ipynb
- [x] Pipelines.ipynb
- [x] Django_Apps.ipynb


jbednar:
Notes and issues:

- [ ] Is there a Reactive user guide?  Links compares callbacks to Param API, but really it's more similar to the reactive API, right?
- [ ] Customization.ipynb: it starts out immediately talking about CSS; shouldn't that be the last topic covered, as it's complicated and probably not needed except for theming?  It's off putting right now, particularly since this is the first notebook alphabetically after the core guides.

philippjfr:
For a bit of context, the layouts user guide content is now in the customization guide.

philippjfr:
> Deploy_and_Export.ipynb: Is this how .app is supposed to work?

You specified the wrong port number.

> Deploy_and_Export.ipynb: Not sure how to avoid this:

That's the warning from the previous cell that's warning you about having specified the wrong port number.

examples/gallery/external/deck.gl.ipynb
```diff
@@ -111,7 +111,7 @@
     "\n",
     "widget.jslink(deckgl, code={'active': update_city});\n",
     "\n",
-    "pn.Column(widget, deckgl)"
+    "pn.Column(widget, deckgl).servable()"
```
philippjfr: jslink based examples do not need servable.

```suggestion
    "pn.Column(widget, deckgl)"
```

jbednar:
> You specified the wrong port number.

I didn't change the port number; do you mean I didn't edit the number to match my running notebook?  If so it probably needs to say that explicitly.

examples/gallery/external/deck.gl.ipynb
jbednar: Ok, but it might be good to emphasize in that notebook how to make a pure HTML exported version (if not using panel serve as in other examples).  Maybe should .servable() still be used, for use with `panel html` (which currently isn't supported), so that someone can write a notebook and then serve it as HTML in the same way as `panel serve` does for a live Python server?

Issue #408, opened 2019-05-01
[VTK] Converter is buggy - propose move to PyVista for maintainability


banesullivan:
First, I love what you have created for converting VTK render windows to VTKjs scenes that can be embedded in a panel! It's absolutely awesome!

I've been working towards incorporating `panel` as an optional dependency for [`vtki`](https://github.com/vtkiorg/vtki) to allow users to view scenes they create via `vtki` directly in a notebook - I have examples live at http://playground.vtki.org

Unfortunately, there are a lot of nuances to the conversion between VTK render scenes and VTKjs scenes that aren't covered in your `panel/pane/vtk.py` script - also I've been experiencing lots of bugs regarding scalar mapping and `vtkStructuredGrid`s outright don't work with your script.  These issues bring me hesitation in fully supporting panel as the default notebook viewer for `vtki`

A while back in https://github.com/vtkiorg/vtki/pull/57 (with many later bug fixes), I created the same VTKjs export script directly in `vtki` that handles a lot of the VTK nuances... I'm curious if you all would be interested in migrating your VTKjs conversion needs over to `vtki`? This way the developers for `vtki` could maintain the nitty-gritty details of how VTK scenes are converted to VTKjs and `panel` could simply depend on `vtki` for that conversion. 

*The exact details of this dependency might get complicated since `panel` would rely on `vtki` for conversion and `vtki` would rely on `panel` for displaying scenes in a notebook - we'll have to figure out a clean solution for this...*

xavArtley:
I cannot work on this for the moment (I'm in holidays for the next weeks) however we could refactor a little bit the class to set custom vtkjs serializers. We could keep the panel serializer as default for backward compatibility and not add extra dependencies. And in vtki you could set your own serializer

philippjfr:
>A while back, I created the same VTKjs export script directly in vtki that handles a lot of the VTK nuances... I'm curious if you all would be interested in migrating your VTKjs conversion needs over to vtki? This way the developers for vtki could maintain the nitty-gritty details of how VTK scenes are converted to VTKjs and panel could simply depend on vtki for that conversion.

I think that was the long term plan anyway, we just needed something to be merged in the shorter term for some internal demos so I pushed ahead with merging @xavArtley's implementation.

>We could keep the panel serializer as default for backward compatibility and not add extra dependencies. And in vtki you could set your own serializer

This seems like a decent plan to me, it would probably default to the `vtki` serializer if installed and otherwise fall back to the existing one.

[...]

PR #475, opened 2019-06-10
custom serializer registration improvement

xavArtley:
This PR improve the one to set a custom serializer for vtk RenderWindow (#409)
Here we can register serializer for different class not just the vtk RenderWindow wich is limitatting if as input of the serializer we have a custom class.

With this PR for example we can use the pyvista (#408) export function without major change in code:
```python
import panel as pn
pn.extension('vtk')

import pyvista as pv
from pyvista import examples

mesh = examples.download_st_helens().warp_by_scalar()

# First a default plot with jet colormap
p = pv.Plotter()
# Add the data, use active scalar for coloring, and show the scalar bar
p.add_mesh(mesh)

def pyvista_serializer(plotter, delete_tmp_file=False):
    filename = 'pyvista_tmp'
    plotter.export_vtkjs(filename)
    with open(filename + '.vtkjs', 'rb') as f:
        vtkjs = f.read()
    if delete_tmp_file:
        import os
        os.remove(filename + '.vtkjs')
    return vtkjs
pn.pane.VTK.register_serializer(pv.plotting.Plotter, pyvista_serializer)

pn.panel(p)
```
image

### DIFF
--- a/panel/pane/vtk/vtk.py
+++ b/panel/pane/vtk/vtk.py
@@ -102,11 +102,14 @@ class VTK(PaneBase):
         elif hasattr(self.object, 'read'):
             vtkjs = self.object.read()
         else:
-            if not VTK._serializers:
+            available_serializer = [v for k, v in VTK._serializers.items() if isinstance(self.object, k)]
+            if len(available_serializer) == 0:
                 import vtk
                 from .vtkjs_serializer import render_window_serializer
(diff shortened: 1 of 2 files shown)


philippjfr:
Ready to merge?

xavArtley:
Yes

Le mar. 25 juin 2019 à 00:09, Philipp Rudiger <notifications@github.com> a
écrit :

xavArtley:
I could remove linter warning in the same time.

PR #489, opened 2019-06-24
Added on_click example for Button widget

hoseppan:
Added example on how to use on_click method to trigger function when button is pressed.

I could not find any other working example how to use button in Panel documentation.


philippjfr:
Thanks for doing this! If you don't mind could you remove the additional metadata that was added? We generally strip this out. Don't worry if not though, I'll just merge and remove it in a subsequent PR.

philippjfr:
Thanks again!

PR #490, opened 2019-06-25
panel pass kwargs to the widget

henriqueribeiro:
Implements https://github.com/pyviz/panel/issues/456

Following @philippjfr suggestion, now it's also possible to pass a dictionary to the widgets. This dictionary will be passed as kwargs to the widget itself.


philippjfr:
@jbednar Do we prefer 'widget' or 'type' to override the widget type as a dictionary key?

philippjfr:
We should document this in the widget override section of the Param user guide.

philippjfr:
Thanks for the PR @henriqueribeiro, hugely appreciated!

jbednar:
Isn't `type` a reserved word?  I do prefer `type` to `widget`, I think, but I'd have to see an example of usage (as you requested).

philippjfr:
Here's an example:

```
class Example(param.Parameterized):

    test = param.ListSelector(default=['a'], objects=['a', 'b', 'c', 'd', 'e', 'f'], height=10)

    @param.depends("test")
    def view(self):
        return self.test

example = Example()
pn.Row(pn.Param(example.param, widgets={'test': {'type': pn.widgets.CheckBoxGroup, 'inline': False}} , example.view)
```

jbednar:
I have no preference between `type` and `widget` in this case; they are both accurate.  We should be sure to clarify in the same section of the docs that it's a lot easier to change the `inline` parameter at the class level, for the cases where that is sufficient:

```
import panel.widgets as pw
pw.CheckBoxGroup.inline=False
```

henriqueribeiro:
Hey! I do agree with 'type' instead of widget. I will change it and will document it. I just wanted you to check if you agree with this first version :smile:

panel/param.py
```diff
@@ -285,9 +285,17 @@ def update_pane(change, parameter=pname):
...
         """Get widget for param_name"""
         p_obj = self.object.param[p_name]
+        kw_widget = {}
 
         if self.widgets is None or p_name not in self.widgets:
             widget_class = self.widget_type(p_obj)
+        elif isinstance(self.widgets[p_name], dict):
+            if 'widget' in self.widgets[p_name]:
+                widget_class = self.widgets[p_name]['widget']
+                self.widgets[p_name].pop('widget')
```
philippjfr: You can do this in a single line:

```python
widget_class = self.widgets[p_name].pop('widget')
```

philippjfr:
Looks good to me. Would love to see some documentation and a test but won't let that hold up merging.

henriqueribeiro:
I will add some tests later

philippjfr:
Thanks so much!

PR #505, opened 2019-07-08
DOC: remove pyplot from mpl API example

tacaswell:
Create Axes objects and pass them into pandas to avoid registering
them with pyplot.  This avoids:

 - needing to close the figures to avoid memory leaks (as no one else
   is holding a reference to it)
 - needing to close the figure to avoid the `inline` backend
   automatically displaying the figure.

The second problem could be avoided by switch to a non-interactive
backend (such as Agg) but would not solve the problem of the memory
leak or getting the figure object implicitly created by pandas.


philippjfr:
Thanks @tacaswell, much appreciated. We should probably capture this guidance in the [Matplotlib pane reference](http://panel.pyviz.org/reference/panes/Matplotlib.html) as well.

philippjfr:
CI issues are unrelated, merging.

PR #526, opened 2019-07-13
Widget box orientation

gwbischof:
Fixes https://github.com/pyviz/panel/issues/500

### DIFF
--- a/panel/layout.py
+++ b/panel/layout.py
@@ -342,24 +342,26 @@ class Column(ListPanel):
     _bokeh_model = BkColumn
 
 
-class WidgetBox(Column):
+class WidgetBox(ListPanel):
     """
     Vertical layout of widgets.
     """
 
+    _bokeh_model = BkColumn
+
     @property
     def horizontal(self):
-        return _bokeh_model == BkRow
+        return self._bokeh_model == BkRow
 
     @horizontal.setter
     def horizontal(self, value):
         if not isinstance(value, bool):
-            raise ValueError(f"Invalid value {value}. The horizontal attribute
-                               must be a boolean.")
+            raise ValueError(f"Invalid value {value}. The horizontal attribute "
+                              "must be a bool.")
         if value:
-            _bokeh_model = BkRow
+            self._bokeh_model = BkRow
         else:
-            _bokeh_model = BkColumn
+            self._bokeh_model = BkColumn
 
     css_classes = param.List(default=['widget-box'], doc="""
         CSS classes to apply to the layout.""")
--- a/panel/layout.py
+++ b/panel/layout.py
@@ -347,21 +347,14 @@ class WidgetBox(ListPanel):
     Vertical layout of widgets.
     """
 
-    _bokeh_model = BkColumn
+    _rename = {'objects': 'children', 'horizontal': None}
+
+    horizontal = param.Boolean(default=False, doc="""Whether to lay out the
+                    widgets in a Row layout as opposed to a Column layout.""")
 
     @property
-    def horizontal(self):
-        return self._bokeh_model == BkRow
-
-    @horizontal.setter
-    def horizontal(self, value):
-        if not isinstance(value, bool):
-            raise ValueError("Invalid value " +  value + ". The horizontal attribute "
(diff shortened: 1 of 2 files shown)


philippjfr:
Thanks for this @gwbischof! It's very much appreciated, however ``horizontal`` should also be settable via the constructor, so we should make it a parameter and the `_bokeh_model`` a property.

```python
    horizontal = param.Boolean(default=False, doc="""
       Whether to lay out the widgets in a Row layout as opposed to a Column layout.""")

    @property
    def _bokeh_model(self):
        return BkRow if self.horizontal else BkColumn
```

A simple test that call `widget_box._get_root()` and does an isinstance check on the returned model would also be appreciated.

gwbischof:
I get a bokeh error if I make horizontal a property:
'''---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
~/anaconda3/envs/caproto/lib/python3.7/site-packages/IPython/core/formatters.py in __call__(self, obj, include, exclude)
    968 
    969             if method is not None:
--> 970                 return method(include=include, exclude=exclude)
    971             return None
    972         else:

~/dama/panel/panel/viewable.py in _repr_mimebundle_(self, include, exclude)
    270         doc = _Document()
    271         comm = state._comm_manager.get_server_comm()
--> 272         model = self.get_root(doc, comm)
    273         if config.embed:
    274             embed_state(self, model, doc,

~/dama/panel/panel/viewable.py in get_root(self, doc, comm)
    398         """
    399         doc = doc or _curdoc()
--> 400         root = self._get_model(doc, comm=comm)
    401         self._preprocess(root)
    402         ref = root.ref['id']

~/dama/panel/panel/layout.py in _get_model(self, doc, root, parent, comm)
    105         objects = self._get_objects(model, [], doc, root, comm)
    106         props = dict(self._init_properties(), objects=objects)
--> 107         model.update(**self._process_param_change(props))
    108         self._models[root.ref['id']] = (model, parent)
    109         self._link_props(model, self._linked_props, doc, root, comm)

~/anaconda3/envs/caproto/lib/python3.7/site-packages/bokeh/core/has_props.py in update(self, **kwargs)
    374         '''
    375         for k,v in kwargs.items():
--> 376             setattr(self, k, v)
    377 
    378     def update_from_json(self, json_attributes, models=None, setter=None):

~/anaconda3/envs/caproto/lib/python3.7/site-packages/bokeh/core/has_props.py in __setattr__(self, name, value)
    286 
    287             raise AttributeError("unexpected attribute '%s' to %s, %s attributes are %s" %
--> 288                 (name, self.__class__.__name__, text, nice_join(matches)))
    289 
    290     def __str__(self):

AttributeError: unexpected attribute 'horizontal' to Column, possible attributes are align, aspect_ratio, background, children, css_classes, disabled, height, height_policy, js_event_callbacks, js_property_callbacks, margin, max_height, max_width, min_height, min_width, name, rows, sizing_mode, spacing, subscribed_events, tags, visible, width or width_policy```

philippjfr:
Okay, add this to the class:

```python
    _rename = {'objects': 'children', 'horizontal': None}
```

philippjfr:
Looks perfect thanks, I'll merge once tests pass. Ideally you'd be able to set horizontal and it would update existing views but I can't see any nice way to feasibly do that.

gwbischof:
No prob!
I think panel will be a good solution for us, so I would like to keep contributing. Could you recommend some issues for me to work on?
Thanks!

philippjfr:
>I think panel will be a good solution for us, so I would like to keep contributing.

That's great to hear! Glad it's useful to you and your contributions are super welcome. I'd be happy to make you a contributor if you're interested.

>Could you recommend some issues for me to work on?

Definitely, I'll go through and mark some issues with the "Good first issue tag" so you can filter on those.

jbednar:
Plus in general, any time you spot something confusing or missing on the website or other docs, it's super helpful if you make a PR to fix the problem.  If you have become productive and able to solve your problems with Panel, even  if you aren't an expert, you are very likely to be able to spot things that will trip up newer users and can easily address them.

PR #527, opened 2019-07-14
fixed DateRangeSlider crash

gwbischof:
Fixes https://github.com/pyviz/panel/issues/495
This small change gives the DateRangeSlider the same behavior as the DateSlider


philippjfr:
Thanks @gwbischof, much appreciated!

PR #529, opened 2019-07-14
plotly callback and in-place update support

jonmmease:
This PR contains initial work on providing plotly callback support in Panel.

To start with, I added callback properties that match what Dash uses (https://dash.plot.ly/interactive-graphing):
    - `relayout_data`
    - `restyle_data`
    - `click_data`
    - `hover_data`
    - `clickannotation_data`
    - `selected_data`

These properties are successfully updated on the Python side when interacting with a displayed figure. For example:

```python
import panel as pn
import plotly.graph_objs as go
pn.extension('plotly')

fig_pane = pn.pane.Plotly(
    go.Figure(
        data=[go.Scatter(y=[1, 3, 2], mode='markers'),
              go.Scatter(y=[3, 2, 4], mode='markers')],
    ),
)
fig_pane
```
newplot (14)

Then perform a box selection and look at the resulting `selected_data` property:
```python
fig_pane.selected_data
```
```
{'points': [{'curveNumber': 0,
   'pointNumber': 1,
   'pointIndex': 1,
   'x': 1,
   'y': 3},
  {'curveNumber': 0, 'pointNumber': 2, 'pointIndex': 2, 'x': 2, 'y': 2},
  {'curveNumber': 1, 'pointNumber': 1, 'pointIndex': 1, 'x': 1, 'y': 2}],
 'range': {'x': [0.8510443353259319, 2.0522968260745795],
  'y': [1.5626990445859872, 3.4990047770700636]}}
```

One thing I'm not sure of, is how to use these callback properties in a declarative context. For example, I'd like to be able to write a class something like this:

```python
class StockExplorer(param.Parameterized):
    ticker = param.Selector(default='AAPL', objects=tickers)
    window_size = param.Integer(default=6, bounds=(1, 21))
    
    @param.depends('ticker', 'window_size')
    def plot(self):
        return get_plot(self.ticker, self.window_size)
```

But have the `plot` function also depend on the `selected_data` property.  Is there some way to pre-declare the `plot` will return a figure so that the `selected_data` property can be used declaratively?

### DIFF
--- a/panel/pane/plotly.py
+++ b/panel/pane/plotly.py
@@ -5,6 +5,7 @@ bokeh model.
 from __future__ import absolute_import, division, unicode_literals
 
 import sys
+from functools import partial
 
 import numpy as np
 
@@ -32,14 +33,6 @@ class Plotly(PaneBase):
     clickannotation_data = param.Dict(doc="""clickannotation callback data""")
     selected_data = param.Dict(doc="""selected callback data""")
 
-    _py2js_addTraces = param.Dict(doc="""addTraces message data""")
-    _py2js_restyle = param.Dict(doc="""restule message data""")
-    _py2js_relayout = param.Dict(doc="""relayout message data""")
-    _py2js_update = param.Dict(doc="""update message data""")
-    _py2js_animate = param.Dict(doc="""animate message data""")
-    _py2js_deleteTraces = param.Dict(doc="""deleteTraces message data""")
-    _py2js_moveTraces = param.Dict(doc="""moveTraces message data""")
-
     _updates = True
 
     priority = 0.8
@@ -60,13 +53,13 @@ class Plotly(PaneBase):
             # We only patch `Figure` objects (not subclasses like FigureWidget) so
             # we don't interfere with subclasses that override these methods.
             fig = object
-            fig._send_addTraces_msg = self._send_addTraces_msg
-            fig._send_moveTraces_msg = self._send_moveTraces_msg
-            fig._send_deleteTraces_msg = self._send_deleteTraces_msg
-            fig._send_restyle_msg = self._send_restyle_msg
-            fig._send_relayout_msg = self._send_relayout_msg
-            fig._send_update_msg = self._send_update_msg
-            fig._send_animate_msg = self._send_animate_msg
+            fig._send_addTraces_msg = partial(self._update_figure, fig)
+            fig._send_moveTraces_msg = partial(self._update_figure, fig)
+            fig._send_deleteTraces_msg = partial(self._update_figure, fig)
+            fig._send_restyle_msg = partial(self._update_figure, fig)
+            fig._send_relayout_msg = partial(self._update_figure, fig)
+            fig._send_update_msg = partial(self._update_figure, fig)
+            fig._send_animate_msg = partial(self._update_figure, fig)
 
     def _to_figure(self, obj):
         import plotly.graph_objs as go
@@ -141,14 +134,7 @@ class Plotly(PaneBase):
         self._link_props(
             model, [
                 'config', 'relayout_data', 'restyle_data', 'click_data',  'hover_data',
-                'clickannotation_data', 'selected_data',
-                '_py2js_addTraces',
-                '_py2js_restyle',
-                '_py2js_relayout',
-                '_py2js_update',
-                '_py2js_animate',
-                '_py2js_deleteTraces',
-                '_py2js_moveTraces',
+                'clickannotation_data', 'selected_data'
             ],
             doc,
             root,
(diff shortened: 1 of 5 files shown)


philippjfr:
Excited to start playing with this.

jonmmease:
Now this PR supports updating the Pane in response to property updates on the figure itself. e.g.

This is mostly reuse from the `FigureWidget` class that is included with plotly.py.

```
import panel as pn

pn.extension('plotly')

fig = go.Figure()
fig_pane

# _py2js_addTraces
fig.add_scatter(y=[1, 3, 2])
fig.add_bar(y=[1, 0.4, 0.7]);
fig.add_scatter(y=[2, 1.4, 1.7], marker_size=20, mode='markers');

# _py2js_moveTraces
fig.data = [fig_pane.object.data[i] for i in [0, 2, 1]]

# _py2js_deleteTraces
fig.data = [fig_pane.object.data[1]]

# _py2js_restyle
fig.data[0].marker.color = 'green'

# _py2js_update
with fig.batch_update():
    fig.data[0].marker.color = 'green'
    fig.data[0].marker.size = 30

# _py2js_animate
with fig_pane.object.batch_animate():
    fig.data[0].marker.color = 'orange'
    fig.data[0].marker.size = 40
```

Unlike `FigureWidget`, this does not sync default properties from the JavaScript side back to the Python side.  It also does not support the `on_change`, `on_selection`, etc. figure callbacks.  There's no reason these couldn't be supported, but they would involve a bit more logic to be borrowed from the FigureWidget implementation.

jonmmease:
After some experimentation and gaining some more understanding of the bokeh model system, I changed how the property updates are performed.  The prior iteration used the Plotly.js messages generated by plotly.py to update Plotly.js on the JavaScript side.

But this approach doesn't keep the models themselves up to date with the changes in the figure. In FigureWidget this is handled by manually updating the model in response to each message.  This is done on the JavaScript side, as ipywidgets maintains only a single model in JavaScript that may correspond to multiple views.

If I'm understanding bokeh properly, it looks like a separate model is created for each view.  And the model has both a Python object and a JavaScript object.  So manually updating all of these models in response to Plotly.js messages would be a bit challenging.

The new approach updates to full figure `object` in response to any of these changes.  This passes the synchronization through the already somewhat optimized `_update` method.  In the future we could work on making this method even more efficient by sending only patches, and applying the patches with Plotly.js methods, when small changes are detected.

jonmmease:
@philippjfr, I think this is ready for review when you have time.

I am have one issue that I might need help tracking down.  It seems that occasionally the `selected_data` property is not successfully synchronized from JavaScript to Python.  Are there any race conditions or throttling situations I should be considering when updating these model properties in JavaScript?

philippjfr:
>So manually updating all of these models in response to Plotly.js messages would be a bit challenging.

Could you expand on this? I'm not quite following why this is problematic or challenging.

>Are there any race conditions or throttling situations I should be considering when updating these model properties in JavaScript?

Not afaik, I'll play around.

panel/pane/plotly.py
```diff
@@ -33,6 +43,24 @@ def applies(cls, obj):
...
+            # We only patch `Figure` objects (not subclasses like FigureWidget) so
+            # we don't interfere with subclasses that override these methods.
+            fig = object
+            fig._send_addTraces_msg = partial(self._update_figure, fig)
+            fig._send_moveTraces_msg = partial(self._update_figure, fig)
+            fig._send_deleteTraces_msg = partial(self._update_figure, fig)
+            fig._send_restyle_msg = partial(self._update_figure, fig)
+            fig._send_relayout_msg = partial(self._update_figure, fig)
+            fig._send_update_msg = partial(self._update_figure, fig)
+            fig._send_animate_msg = partial(self._update_figure, fig)
```
philippjfr: These can all use `lambda: self.param.trigger('object')` or similar.

panel/pane/plotly.py
```diff
@@ -143,3 +212,6 @@ def _update(self, model):
 
         if update_layout:
             model.layout = json.get('layout')
+
+    def _update_figure(self, fig, *args, **kwargs):
+        self.object = fig
```
philippjfr: Won't need this once you use trigger above.

panel/pane/plotly.py
philippjfr: I think this also needs to happen not only on the constructor but whenever a user updates the ``object``. I think we should probably add a callback on changes to ``.object`` which calls `_to_figure` and caches the created figure. Something like:

```python
@param.depends('object')
def _update_figure(self):
     if self.object is not None and id(self.object) not in self._figure:
         fig = self._to_figure(self.object)
         self._figure = {id(self.object): fig}
         fig._send_addTraces_msg = lambda: self.param.trigger('object')
         ...
```

jonmmease:
> Could you expand on this? I'm not quite following why this is problematic or challenging.

Sure. Let's say a user updates a trace with property assignment

```python
fig.data[1].marker.line.width = 5
```
The `Figure` object will update it's internal `dict` representation and call it's `_send_restyle_msg` method with arguments 

```
style={'marker.line.width': 5}, trace_indexes=[1]
```

These arguments are exactly what plotly.js needs in order to update the JavaScript figure with 
```
Plotly.restyle(this.el, {'marker.line.width': 5}, [1])
```

In https://github.com/pyviz/panel/pull/529/commits/259c932846f67a2100d346c8f651ff05edc14178, I transferred these arguments to the JavaScript side with the `_py2js_restyle` property and then called `Plotly.restyle` whenever this property changed.  This approach successfully updates the original figure view, but doesn't update the backing model so when the pane is displayed again, the original out-of-date view is shown.

In the `FigureWidget` case, the `FigureWidget` class is an ipywidgets subclass and so the source-of-truth dictionary for the figure is the model on the Python side (So there is no need to separately update a model in Python).  To update the figure model in JavaScript, there is a custom `do_restyle` function that inputs the `_send_restyle_msg` message data and then performs the update on the JavaScript model. As long as this function is equivalent to what Plotly.js does, then everything stays in sync.

To do something similar with Bokeh models, I think we would need to have a separate Python implementation of `restyle` that updates each model in-place. And we would need a JavaScript implementation of restyle to update each JavaScript model in-place (this we could reuse from FigureWidget if we can include lodash as a dependency).  If the Python and JavaScript restyle functions are correct then everything would stay in sync.

The restyle function isn't complicated to implement, but it's just one of 7 functions that would need to be implemented, both in Python and JavaScript, to make this work (restyle, relayout, addTraces, moveTraces, removeTraces, update, and animate).  So perhaps I should have said that this would be "tedious" rather than "challenging" 🙂 

Though maybe Bokeh has something that would make this simpler than I'm picturing.

Does that make sense?

panel/pane/plotly.py
jonmmease: That all makes sense. Will do

jonmmease:
> I think this also needs to happen not only on the constructor but whenever a user updates the object. I think we should probably add a callback on changes to .object which calls _to_figure and caches the created figure. Something like:

I reworked the installation of the message callbacks to happen in an `_update_figure` method that depends on `object`.  At this point I'm only keeping track of the most recent figure so that we can skip the callback registration if a Pane's figure is assigned back to itself. And I'm not converting dicts to `Figure`s in here, just checking if `object` is a `Figure` and installing the callbacks if it is.

I'm not confident I fully understood your suggestion, so let me know if there are some more things that should change here.

jonmmease:
Here is a selection example that sometimes doesn't update for me

```python
import panel as pn
import plotly.express as px
import numpy as np
import pandas as pd
pn.extension("plotly")

df = px.data.iris()

fig = (go.Figure(layout_showlegend=False)
       .add_scatter(x=df.sepal_width,
                    y=df.petal_length,
                    mode='markers',
                    hoverinfo='x')
       .update_layout(dragmode='select'))

md = pn.pane.Markdown()

def update_selected(event):
    selected_data = fig_pane.selected_data
    if  selected_data:
        point_inds = [point['pointNumber']
                      for point in selected_data['points']
                      if point['curveNumber'] == 0]
    else:
        point_inds = []
    md.object = str(point_inds)        

fig_pane.param.watch(update_selected, 'selected_data')

pn.layout.Column(fig_pane, md)
```
Screen Shot 2019-07-23 at 1 05 03 PM

**Note:** If I disable setting the hover properties from the client, then I don't see the issue at all.  So it looks like it has something to do with other properties changing quickly around the same time as the selection.

philippjfr:
Should I merge as a first cut and then we can address issues in another PR?

philippjfr:
Will just go ahead with that. We should probably keep track of any problems and long-term plans in a separate issue.

jonmmease:
Sounds good, thanks!

PR #543, opened 2019-07-21
Add fancy_layout option to HoloViews pane

philippjfr:
The current HoloViews pane generates a simple layout of the plot and the set of widgets generated from that plot. This makes it easy to compose these two components in different ways but it doesn't match the look that the old-style HoloViews widgets use. Since we want to swap out the widgets in HoloViews with as little disruption as possible this PR adds a ``fancy_layout`` option which tries to match the look and feel of the original widget implementation.

As a simple example here is the output of a HoloMap with this option enabled:

Screen Shot 2019-07-21 at 1 23 32 PM

### DIFF
--- a/panel/pane/holoviews.py
+++ b/panel/pane/holoviews.py
@@ -210,12 +210,15 @@ class HoloViews(PaneBase):
         for i, dim in enumerate(dims):
             widget_type, widget, widget_kwargs = None, None, {}
             if fancy:
-                if i == 0:
-                    kwargs = {'margin': (20, 20, 5, 20)}
+                if i == 0 and i == (len(dims)-1):
+                    margin = (20, 20, 20, 20)
+                elif i == 0:
+                    margin = (20, 20, 5, 20)
                 elif i == (len(dims)-1):
-                    kwargs = {'margin': (5, 20, 20, 20)}
+                    margin = (5, 20, 20, 20)
                 else:
-                    kwargs = {'margin': (0, 20, 5, 20)}
+                    margin = (0, 20, 5, 20)
+                kwargs = {'margin': margin, 'width': 250}
             else:
                 kwargs = {}
 
--- a/panel/pane/holoviews.py
+++ b/panel/pane/holoviews.py
@@ -10,6 +10,7 @@ from collections import OrderedDict, defaultdict
 from functools import partial
 
 import param
+
 from bokeh.models import Spacer as _BkSpacer
 
 from ..io import state
@@ -189,9 +190,10 @@ class HoloViews(PaneBase):
         return isinstance(obj, Dimensioned)
 
     @classmethod
-    def widgets_from_dimensions(cls, object, widget_types={}, widgets_type='individual', fancy=False):
+    def widgets_from_dimensions(cls, object, widget_types={}, widgets_type='individual',
+                                fancy=False):
         from holoviews.core import Dimension
-        from holoviews.core.util import isnumeric, unicode, datetime_types
+        from holoviews.core.util import isnumeric, unicode, datetime_types, unique_iterator
         from holoviews.core.traversal import unique_dimkeys
         from holoviews.plotting.util import get_dynamic_mode
         from ..widgets import Widget, DiscreteSlider, Select, FloatSlider, DatetimeInput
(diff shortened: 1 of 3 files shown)


jbednar:
I se this is merged already, but I'm wondering whether `fancy` is the right term for this arrangement.  Fancy makes it sound like it's better than the default, but really it's just different, as far as I can see.  Could we use a more neutral word here, like `shaded_widgets`?

philippjfr:
Happy to change it, about to open another PR to improve the scrubber layout anyway, but -1 on shaded. Open to other suggestions.

philippjfr:
The differences are:

1. The widgets are inside a widget box (which is shaded and gray)
2. The plot and widgets are horizontally centered and the margins are responsive
3. The widget box is vertically centered

philippjfr:
Also for context, the "fancy" bit is borrowed from matplotlib's `fancybox` option on legends: https://matplotlib.org/gallery/recipes/transparent_legends.html

jbednar:
I don't think it shares much in common with the mpl fancybox option, which isn't grey or offset like that. Maybe `center_shaded`?  I'm just looking for something more literal and with no valence.  But it's only a slight preference, so if there's no literal name you are happy with, then I guess just keep it fancy.

philippjfr:
Cc: @jsignell @jlstevens @jonmmease Any opinions?

jsignell:
What about `original_layout` or `classic_layout`

philippjfr:
I like ``classic_layout``.

jbednar:
`classic_layout` is fine if it's not something we plan to support forever or expect new users to consider using.

philippjfr:
>classic_layout is fine if it's not something we plan to support forever or expect new users to consider using.

We do plan to support it forever, it's the default layout HoloViews will use to render its widgets. Panel users should generally use the default layout so they can compose it themselves more easily.

jonmmease:
I don't have strong feeling on the name itself, but should we consider this to be an enumeration rather than a boolean?  Something like `layout_style='holoviews'`, and then leave open the possibility of other styles in the future.

philippjfr:
>Something like layout_style='holoviews', and then leave open the possibility of other styles in the future.

Not a bad idea either.

philippjfr:
Currently the options would be 'simple' and 'classic' or 'holoviews'.

jonmmease:
As a disclaimer, I haven't looked at it closely enough to have an opinion on whether we would want to have more options like this in the future.  I just know I've often come to regret boolean arguments 🙂

philippjfr:
>As a disclaimer, I haven't looked at it closely enough to have an opinion on whether we would want to have more options like this in the future

I would certainly have to refactor the code and come up with some better representation to express custom layouts because the code is already getting unwieldy with two options but I think more flexibility would be a good thing. People certainly have asked for more flexibility in HoloViews and while they can already achieve that if they use Panel directly and lay things out themselves I don't think it would hurt to include a few more canned layouts they can pick from.

philippjfr:
If no one objects I'm going with:

```python
layout_style = param.ObjectSelector(default='simple', objects=['simple', 'classic'], doc="""
    The default layout of the plot and widgets, by default it uses the 'simple' layout to make
    it easy to compose the plot and the widgets manually.""")
```

jlstevens:
The last suggestion above looks good to me in terms of a `layout_style` parameter. I think 'simple' is clear enough as a name (the current panel behavior afaik) but 'classic' isn't too informative as to what that really means.

I am not a fan of 'holoviews' as a value and 'adjoint' is probably confusing given holoviews uses that word as well. I'm wouldn't let 'classic' block this PR but I would like to try and think of something else if possible (tricky!).

philippjfr:
Could call it centered I guess but there could be a number of centered arrangements. One problem is that it also interacts with the widget_type parameter. Maybe we could add center_right, center_left, center_bottom, center_top, left right, bottom and top options instead. HoloViews would use center_right for regular widget and center_bottom for the scrubber.

jlstevens:
Those options make sense to me and generalize panel nicely. Of course it is also more work to do but this would be my preferred approach...

jbednar:
I was reading the recent discussion starting from the top and independently thought of `center_right` before reading @philippjfr's comment, so I vote for that -- that way people can declare an appropriate default mapping to a layout that they prefer in general, and of course then they can always repack things manually using Panel if they wish.

PR #549, opened 2019-07-26
Add viewport property to plotly pane to hold the axis range for all axes

jonmmease:
This PR is a follow on to the initial plotly callback support added in https://github.com/pyviz/panel/pull/529.

This adds a new `viewport` property to the plotly pane that holds the current axis ranges of all 2D cartesian axes in the figure.  This information isn't alway available through the `relayout_data` callback, for example when a figure is first displayed. This will be required to implement `Range*` streams in HoloViews, and by extension to support Datashader

Also added `viewport_update_policy` to control whether to update viewport
property with "continuous", "mouseup", or "throttle" policies.  When `viewport_update_policy='throttle'`, the throttle duration is controled by the new `viewport_update_throttle` property.  Here's an example of it in action.

```python
import panel as pn
from plotly.subplots import make_subplots
pn.extension('plotly')

fig = make_subplots(rows=1, cols=2)
fig.add_scatter(y=[1, 3, 2], row=1, col=1);
fig.add_bar(y=[3, 2, 10], row=1, col=2);

viewport_pane = pn.pane.Str()

pn.layout.Column(
    fig_pane,
    fig_pane.param.viewport,
    fig_pane.param.viewport_update_policy,
    fig_pane.param.viewport_update_throttle,
)
```
viewport_plotly

The `viewport` property can also be set from Python to set the figure viewport.

SetViewport

With this, the viewports of multiple views of the same pane will stay in sync.

```python
pn.layout.Column(fig_pane, fig_pane)
```

viewports_sync

I added `lodash` as a dependecy of the Plotly pane so handle deep object equality and throttling.  Including lodash will also make it easier to port more of the `FigureWidget` logic to the panel pane in the future.

### DIFF
--- a/panel/pane/plotly.py
+++ b/panel/pane/plotly.py
@@ -157,6 +157,7 @@ class Plotly(PaneBase):
         model = PlotlyPlot(data=json.get('data', []),
                            layout=json.get('layout', {}),
                            config=self.config,
+                           viewport=self.viewport,
                            viewport_update_policy=self.viewport_update_policy,
                            viewport_update_throttle=self.viewport_update_throttle,
                            data_sources=sources)
--- a/panel/pane/plotly.py
+++ b/panel/pane/plotly.py
@@ -33,8 +33,22 @@ class Plotly(PaneBase):
     selected_data = param.Dict(doc="""selected callback data""")
     viewport = param.Dict(doc="""current viewport state""")
     viewport_update_policy = param.Selector(
-        objects=["continuous", "mouseup", "throttle"])
-    viewport_update_throttle = param.Integer(bounds=(0, None), default=200)
+        objects=["continuous", "mouseup", "throttle"],
+        default="continuous",
+        doc="""\
+Policy by which the viewport parameter is updated during user interactions:
+ - "continuous": updates are synchronized continually while panning
+ - "mouseup": updates are synchronized when mouse button is released after panning
+ - "throttle": updates are synchronized while panning, at intervals determined by the
+               viewport_update_throttle parameter"""
+    )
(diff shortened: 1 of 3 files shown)


jonmmease:
Don't merge yet, there's a feedback bug that can cause updates to get stuck in an infinite loop when multiple views are moved together.  Investigating...

jonmmease:
Testing again with #553 and there is still a problem.  The runaway feedback loop has gone away, but after panning one plot for a few seconds the other stops responding and I get the following error in Chrome:

Screen Shot 2019-07-28 at 4 22 44 PM

jonmmease:
Alright, this is working reliable now 🎉  No more feedback loops or console errors when dragging in continuous mode when multiple views of the same figure are displayed together.

@philippjfr ready for review whenever you have time.  There are a couple of other tweaks I want to make to the plotly pane, but I'll do that in another PR since this one is pretty well self-contained now.

panel/pane/plotly.py
```diff
@@ -31,6 +31,10 @@ class Plotly(PaneBase):
     hover_data = param.Dict(doc="""hover callback data""")
     clickannotation_data = param.Dict(doc="""clickannotation callback data""")
     selected_data = param.Dict(doc="""selected callback data""")
+    viewport = param.Dict(doc="""current viewport state""")
+    viewport_update_policy = param.Selector(
+        objects=["continuous", "mouseup", "throttle"])
```
philippjfr: Should add a docstring and make the default explicit.

philippjfr:
One minor comment, otherwise I'm happy to merge.

jonmmease:
Thanks for taking a look. docs and defaults added.

Issue #556, opened 2019-07-29
DatePicker value type error in notebook context


randomstuff:
#### ALL software version info (Panel, bokeh, python, notebook, OS, browser, any other relevant packages)

* Debian testing
* Python 3.7.4
* `pip install panel`

<pre>
bokeh           1.3.0  
Jinja2          2.10.1 
Markdown        3.1.1  
MarkupSafe      1.1.1  
numpy           1.17.0 
packaging       19.0   
panel           0.6.0  
param           1.9.1  
Pillow          6.1.0  
pip             18.1   
pkg-resources   0.0.0  
pyct            0.4.6  
pyparsing       2.4.1.1
python-dateutil 2.8.0  
pyviz-comms     0.7.2  
PyYAML          5.1.1  
setuptools      40.8.0 
six             1.12.0 
tornado         6.0.3 
</pre>

#### Complete, minimal, self-contained example code that reproduces the issue

```python
import datetime
import panel as pn
import panel.widgets as pnw
import panel.widgets.slider

a = datetime.date(2019, 1, 1)
b = datetime.date(2019, 12, 31)

pn.extension()
pnw.DateSlider(name="x", start=a, end=b, value=b)
```

#### Stack traceback and/or browser JavaScript console output

<pre>
Python failed with the following traceback: 
/home/gcorona/src/panel/panel/viewable.py _change_event L653
/home/gcorona/temp/test-toto/venv/lib/python3.7/site-packages/param/parameterized.py inner L1219
/home/gcorona/temp/test-toto/venv/lib/python3.7/site-packages/param/parameterized.py set_param L2572
/home/gcorona/temp/test-toto/venv/lib/python3.7/site-packages/param/parameterized.py set_param L1358
/home/gcorona/temp/test-toto/venv/lib/python3.7/site-packages/param/parameterized.py _f L294
/home/gcorona/temp/test-toto/venv/lib/python3.7/site-packages/param/parameterized.py _f L296
/home/gcorona/temp/test-toto/venv/lib/python3.7/site-packages/param/__init__.py __set__ L617
/home/gcorona/temp/test-toto/venv/lib/python3.7/site-packages/param/parameterized.py _f L296
/home/gcorona/temp/test-toto/venv/lib/python3.7/site-packages/param/parameterized.py __set__ L823
/home/gcorona/temp/test-toto/venv/lib/python3.7/site-packages/param/__init__.py _validate L1830
[...]

Issue #555, opened 2019-07-29
DatePicker value type error in Bokeh Server context


randomstuff:
#### ALL software version info (Panel, bokeh, python, notebook, OS, browser, any other relevant packages)

* Debian testing
* Python 3.7.4
* `pip install panel`

<pre>
bokeh           1.3.0  
Jinja2          2.10.1 
Markdown        3.1.1  
MarkupSafe      1.1.1  
numpy           1.17.0 
packaging       19.0   
panel           0.6.0  
param           1.9.1  
Pillow          6.1.0  
pip             18.1   
pkg-resources   0.0.0  
pyct            0.4.6  
pyparsing       2.4.1.1
python-dateutil 2.8.0  
pyviz-comms     0.7.2  
PyYAML          5.1.1  
setuptools      40.8.0 
six             1.12.0 
tornado         6.0.3 
</pre>

#### Description of expected behavior and the observed behavior

When changing the value of a `Datepicker`, it fails with: `TypeError: 'datetime.date' object is not subscriptable`.

#### Complete, minimal, self-contained example code that reproduces the issue

```python
import panel.widgets as pnw
pnw.DatePicker().servable()
```

```sh
panel serve test.py
```

#### Stack traceback and/or browser JavaScript console output

<pre>
2019-07-29 10:08:31,618 Exception in callback functools.partial(<bound method IOLoop._discard_future_result of <tornado.platform.asyncio.AsyncIOMainLoop object at 0x7f5a885a7150>>, <Future finished exception=TypeError("'datetime.date' object is not subscriptable")>)
Traceback (most recent call last):
  File "/home/gcorona/temp/test/venv/lib/python3.7/site-packages/tornado/ioloop.py", line 743, in _run_callback
    ret = callback()
  File "/home/gcorona/temp/test/venv/lib/python3.7/site-packages/tornado/ioloop.py", line 767, in _discard_future_result
    future.result()
  File "/home/gcorona/temp/test/venv/lib/python3.7/site-packages/tornado/gen.py", line 748, in run
    yielded = self.gen.send(value)
  File "/home/gcorona/temp/test/venv/lib/python3.7/site-packages/bokeh/server/session.py", line 70, in _needs_document_lock_wrapper
    result = yield yield_for_all_futures(func(self, *args, **kwargs))
  File "/home/gcorona/temp/test/venv/lib/python3.7/site-packages/bokeh/server/session.py", line 191, in with_document_locked
    return func(*args, **kwargs)
[...]

PR #557, opened 2019-07-29
Fix dates

randomstuff:
Fixes for #556 and #555.

### DIFF
--- a/panel/tests/widgets/test_input.py
+++ b/panel/tests/widgets/test_input.py
@@ -1,7 +1,7 @@
 from __future__ import absolute_import, division, unicode_literals
 
 import pytest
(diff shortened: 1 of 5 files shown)


philippjfr:
Thanks so much for the fix! Any chance we could get a test or two?

randomstuff:
@philippjfr: added some tests for these problems.

randomstuff:
Updated for compatibility with Python2.

Note: After this change, `DateSlider` is `date`-valued after user assignment (because this was the case for the working case) but this is not consistent with `DateRangeSlider` (which is `datetime`-valued). I guess it would be better to make `DateRangeSlider` `date`-valued.

PR #558, opened 2019-07-29
Plotly: Support preserving user-interaction states across updates with `uirevision` property

jonmmease:
The goal of this PR is to make it possible to preserve user interaction states across updates to the plotly pane.  Examples of user interaction states include the current zoom level, the active modebar buttons, and the traces that were hidden by clicking on the legend.

By default, these user interaction states are reset during calls to `Plotly.react`, but Plotly.js provides `uirevision` properties for controlling which properties should persist across calls to `react`. See https://community.plot.ly/t/preserving-ui-state-like-zoom-in-dcc-graph-with-uirevision for more details.

Prior to this PR, there were a few issues that prevented `uirevision` from working properly.

 1. First, the `render` function in `panel/models/plotly.ts` sometimes called `Plotly.purge` prior to `Plotly.react`.  This deletes the prior plot state before `Plotly.react` has access to it, so it's not possible to preserve user interaction states.  This also takes away the efficiency of using `Plotly.react`.  The `Plotly.react` function is responsible for transforming whatever the prior figure state was into the new figure state, so the `purge` call shouldn't be needed.
 2. The `super.render()` call in the render function also has the effect of deleting the prior plot state.
 3. When an existing figure was updated, the traces, layout, and data sources were all updated independently.  This is advantageous in terms of synchronizing as little data as possible from Python  to JavaScript, but updating the figure using a combination of `relayout`, `restyle`, and `react` API calls made it hard to reason about when the `uirevision` property would be passed to the `react` function.

This PR addresses (1) and (2) by removing the calls to `purge` and `super.render()`.  Everything seems to work alright without the call to `super.render()`, is there anything to be concerned about with bypassing this logic?

To address (3), a new private `_render_count` param was added to store the number of times the figure has been rendered. The Python logic for updating the `layout`, `data`, and `data_sources` params has remained unchanged, but now the `_render_count` param is incremented if there is any change anywhere in the figure.  On the JavaScript side, `layout`, `data`, and `data_sources` properties are no longer watched for changes. Instead, the `_render_count` param is watched and when this changes, the figure is updated using `Plotly.react`.

Here's an example of responding to selection events to overlay a second trace with the selected points.  Without these changes, and without the `layout.uirevision=True` property on the figure, the modebar tool would reset to zoom after each selection.

```python
import panel as pn
import plotly.express as px
import numpy as np
import pandas as pd
import plotly.io as pio
pn.extension("plotly")

df = px.data.iris()

scatter_opts = dict(mode='markers', unselected_marker_opacity=1.0, selected_marker_opacity=1.0)
fig = (go.Figure(layout_showlegend=False)
      .add_scatter(x=df.sepal_width, y=df.petal_length, marker_size=10, **scatter_opts)
      .add_scatter(x=[None], y=[None], **scatter_opts)
      .update_layout(
          uirevision=True
      ))

def update_selected(event):
    selected_data = fig_pane.selected_data
    if not selected_data:
        point_inds = []
    else:
        point_inds = [point['pointNumber']
                      for point in selected_data['points']
                      if point['curveNumber'] == 0]
        
    selected_trace = fig.data[0]

    with fig.batch_update():
        fig.data[0].update(selectedpoints=[])
        fig.data[0].update(selectedpoints=None)
        fig.data[1].update(x=selected_trace.x[point_inds], y=selected_trace.y[point_inds])

fig_pane.param.watch(update_selected, 'selected_data')

fig_pane
```
select_uirevision

### DIFF
--- a/panel/models/plotly.py
+++ b/panel/models/plotly.py
@@ -43,6 +43,7 @@ class PlotlyPlot(LayoutDOM):
     viewport = Dict(String, Any)
     viewport_update_policy = Enum("continuous", "mouseup", "throttle")
     viewport_update_throttle = Int()
+    _render_count = Int()
 
 
 CUSTOM_MODELS['panel.models.plotly.PlotlyPlot'] = PlotlyPlot
--- a/panel/models/plotly.ts
+++ b/panel/models/plotly.ts
@@ -101,6 +101,7 @@ export class PlotlyPlotView extends HTMLBoxView {
   model: PlotlyPlot
   _setViewport: Function
   _settingViewport: boolean = false
+  _plotInitialized: boolean = false
 
   connect_signals(): void {
     super.connect_signals();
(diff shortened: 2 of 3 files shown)


jonmmease:
looking into the test failure...

jonmmease:
Travis tests passing now. Looks like AppVeyor is a bit backed up.

PR #567, opened 2019-07-31
HoloViews: Fallback to holoviews.Store.backend in _get_model if backend undefined

jonmmease:
This ensures that we construct a specific Pane subclass as `child_pane` whenever possible.  This is important for the current design of https://github.com/pyviz/holoviews/pull/3866.  But even if the design of that PR changes I think this is a reasonable thing to do.

### DIFF
--- a/panel/pane/holoviews.py
+++ b/panel/pane/holoviews.py
@@ -116,6 +116,7 @@ class HoloViews(PaneBase):
     #----------------------------------------------------------------
 
     def _get_model(self, doc, root=None, parent=None, comm=None):
+        from holoviews import Store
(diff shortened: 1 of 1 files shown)


philippjfr:
Oh I thought that was already the case, definitely the right behavior.

jonmmease:
The logic was already in there for the `_render` method, so the display looked fine and it's probably never caused anyone problems.

But the `_plots` entry would be set to a `Pane` object instead of the specific plot pane type.  I only noticed it when trying to update the plot pane in https://github.com/pyviz/holoviews/pull/3866.

PR #570, opened 2019-08-01
Added definition_order parameter to CrossSelector

jlstevens:
Right now you can't specify any chosen ordering in the selected list on the right side of a cross-selector as the filtering forces the items into the definition order for `options`. Note that when you set the order via `value`, the widget temporarily accepts the chosen ordering before reverting to definition order.

In some situations, you want to specify the ordering of the selected list based on the ordering you add/remove items to/from it. This PR adds a `definition_order` parameter that you can set to `False` to allow this use case.

I'm not entirely sure that `self._lists[selected].values` is the right set of values here but it does seem to work correctly in my (limited) testing. At the very least I no longer see the reordering back to definition order.

### DIFF
--- a/panel/tests/widgets/test_select.py
+++ b/panel/tests/widgets/test_select.py
@@ -302,3 +302,13 @@ def test_cross_select_move_unselected_to_selected():
 
     assert cross_select.value == ['A', 1, 'B', 3]
     assert cross_select._lists[True].options == ['A', 'B', '1', '3']
+
+
+def test_cross_select_move_unselected_to_selected_not_definition_order():
+    cross_select = CrossSelector(options=['B', 'A', 'C', 1, 2, 3], value=['A', 1], size=5, definition_order=False)
+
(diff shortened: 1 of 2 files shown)


philippjfr:
Two comments:

1) If you add options dynamically do we need special handling to ensure that this option is respected?

2) I'd like to see a test.

jlstevens:
> If you add options dynamically do we need special handling to ensure that this option is respected?

I don't think so. Adding options dynamically seems to work though any selection is cleared (true before this PR as well).

> I'd like to see a test.

Agreed.

jlstevens:
To recap the behavior I observed prior to this PR (and that is still present):

* Adding to the `options` dynamically resets any selection even if the selected items are members of the new options.
* Shrinking `options` does seem to filter out the selection on the right side without deselecting them and moving them to the left side.

The former point is probably a bug but I think the fix is outside the scope of this PR.

philippjfr:
Looks good. Happy to merge.

PR #571, opened 2019-08-05
Improved fancy layout for scrubber

philippjfr:
HoloViews pane now offers a number of options for arranging the plot and widgets

## left

Screen Shot 2019-08-06 at 10 46 10 AM

## top

Screen Shot 2019-08-06 at 10 46 19 AM

## right

Screen Shot 2019-08-06 at 10 45 10 AM

## bottom

Screen Shot 2019-08-06 at 10 46 02 AM

## center_left

Screen Shot 2019-08-06 at 10 46 37 AM

## center_top

Screen Shot 2019-08-06 at 10 46 52 AM

## center_right

Screen Shot 2019-08-06 at 10 47 09 AM

## center_bottom

Screen Shot 2019-08-06 at 10 47 25 AM

### DIFF
--- a/examples/reference/panes/HoloViews.ipynb
+++ b/examples/reference/panes/HoloViews.ipynb
@@ -21,7 +21,10 @@
     "For layout and styling related parameters see the [customization user guide](../../user_guide/Customization.ipynb).\n",
     "\n",
     "* **``backend``** (str): Any of the supported HoloViews backends ('bokeh', 'matplotlib', or 'plotly')\n",
+    "* **``center``** (boolean): Whether to center the plot\n",
     "* **``object``** (object): The HoloViews object being displayed\n",
+    "* **``widget_location``** (str): Where to lay out the widget relative to the plot \n",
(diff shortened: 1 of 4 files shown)


philippjfr:
@jbednar @jlstevens Thinking of renaming the option to `widget_layout`, how does that sound? Also should I add top_left, top_right, bottom_left, and bottom_right?

philippjfr:
Or maybe should we have separate `widget_layout` and `center` parameters?

philippjfr:
The full set of options in that case would be:

center = True/False
widget_layout = top_left, top, top_right, right_top, right, right_bottom, bottom_right, bottom, bottom_left, left_bottom, left, left_top

jbednar:
Yes, it seems like the two flags need to be separated if all those other options are supported, or there will be too many combinations.  Are those location options identical to what Bokeh supports for legends?  If so that sounds reasonable...

philippjfr:
>Are those location options identical to what Bokeh supports for legends? If so that sounds reasonable...

No, bokeh has a relative positioning system for legends which is expressed in pixels (and therefore pretty useless imo).

PR #573, opened 2019-08-06
Add periodic callback example

ahuang11:
random_number_generator
I wasn't really sure where to demonstrate it (reference gallery? user guide?) so I just dumped it under gallery simple examples

### DIFF
--- a/panel/examples/gallery/simple/random_number_generator.ipynb
+++ b/panel/examples/gallery/simple/random_number_generator.ipynb
@@ -3,12 +3,7 @@
   {
    "cell_type": "code",
    "execution_count": null,
-   "metadata": {
-    "ExecuteTime": {
-     "end_time": "2019-08-06T02:39:22.328125Z",
-     "start_time": "2019-08-06T02:39:22.238841Z"
-    }
-   },
+   "metadata": {},
    "outputs": [],
    "source": [
     "import panel as pn\n",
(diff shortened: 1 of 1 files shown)


philippjfr:
Thanks, I think that's a good place for this example.

panel/examples/gallery/simple/random_number_generator.ipynb
```diff
@@ -0,0 +1,95 @@
...
+   },
+   "types_to_exclude": [
+    "module",
+    "function",
+    "builtin_function_or_method",
+    "instance",
+    "_Feature"
+   ],
+   "window_display": false
+  }
```
philippjfr: Need to remove all this metadata. We use this little bash script to clear notebook metadata before committing it:

```
#!/bin/bash
jq --indent 1 \
    '
    (.cells[] | select(has("outputs")) | .outputs) = []
    | (.cells[] | select(has("execution_count")) | .execution_count) = null
    | .metadata = {"language_info": {"name":"python", "pygments_lexer": "ipython3"}}
    | .cells[].metadata = {}
    ' $1 > /tmp/stripped.ipynb
```

Simply run it and then copy the cleared notebook from /tmp/stripped.ipynb

panel/examples/gallery/simple/random_number_generator.ipynb
```diff
@@ -0,0 +1,62 @@
...
+    "col.servable()"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "periodic_cb.stop()  # finish generating random numbers automatically"
```
philippjfr: Any chance you could you put this behind another button. Right now when you run this as an app (which is how it'll appear in the docs) it will immediately stop the periodic callback.

philippjfr:
Thanks for making the changes and sorry for the last bit of feedback!

ahuang11:
No problem! The package has been extremely useful

PR #586, opened 2019-08-09
Add preprocessing hook to link HoloViews plot axes across panel layout

philippjfr:
As the title says. Should decide whether this should be the default and perhaps handle axiswise option on holoviews objects.

- [x] Add tests

### DIFF
--- a/panel/pane/holoviews.py
+++ b/panel/pane/holoviews.py
@@ -36,6 +36,10 @@ class HoloViews(PaneBase):
     center = param.Boolean(default=False, doc="""
         Whether to center the plot.""")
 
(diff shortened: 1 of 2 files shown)


jbednar:
I vote for this being the default, and for supporting the option to disable it if needed.

johnzzzzzzz:
I would vote to follow the defaults for the underlying library. 
In the case of Holoviews, the axes seem shared by defualt for each axes kind, i.e when two plots share the same underlying column for the x axis.

For example, if the shared data has columns: time, q, z1, z2.
Then the plots for time vs z1 and time vs z1 would share axes AND for the same layout, plots for q vs z1 and q vs z2 would share their (different) axes.

PR #345, opened 2019-03-29
Ensure that events are not launched from the wrong thread

philippjfr:
Fixes #344 

@xavArtley Could you test this?

PR #605, opened 2019-08-19
VTK panel scalar array choice correction

xavArtley:
Work in progress
Need to test if the correction handle correctly the colormap of the vtk mesh in all case and does not bring regression
Should correct in https://github.com/pyvista/pyvista/pull/268 : 
- https://github.com/pyvista/pyvista-support/issues/15
- Make sure scalar/color mapping is working - sometimes the mapping is totally wrong. Blood vessels example shows this
- custom colormaps (see #345) (except we must create a pyvista plotter since mesh.plot(use_panel=True) does not work anymore due to auto_close=True
image

Concerning :
- Send text/labels to VTKjs or at least don't throw an error (text added with add_point_labels and add_text)
For me it does not throw error but the text does not show

@banesullivan it will be great if you could test it

### DIFF
--- a/panel/pane/vtk/enums.py
+++ b/panel/pane/vtk/enums.py
@@ -1,0 +1,21 @@
+
+from enum import Enum
+
+
+class SCALAR_MODE(Enum):
+    Default = 0
+    UsePointData = 1
+    UseCellData = 2
+    UsePointFieldData = 3
+    UseCellFieldData = 4
+    UseFieldData = 5
+
+
+class COLOR_MODE(Enum):
+    DirectScalars = 0
+    MapScalars = 1
+
+
+class ACCESS_MODE(Enum):
+    ById = 0
+    ByName = 1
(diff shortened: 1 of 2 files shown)


banesullivan:
These changes are great and much needed, thanks @xavArtley! I'll try to check this out and test when I have a minute.

Not to throw a curveball at you, but what if we reshifted focus to expanding [`itkwidgets`](https://github.com/InsightSoftwareConsortium/itkwidgets) and added support for `itkwidgets` in `panel` instead of using this VTKjs conversion script? `itkwidgets` is definitely the path forward for PyVista and it has a few open issues to make it more interactive and easier to treat like a PyVista plotter.

See my comment (https://github.com/pyvista/pyvista/issues/364#issuecomment-522080389) for some useful info/links to demos of `itkwidgets`

philippjfr:
@xavArtley Ready to merge?

xavArtley:
I think so I haven't encounter regressions

PR #607, opened 2019-08-20
Open url

hoseppan:
Added example of opening URL into /examples/user_guide/Links.ipynb

### DIFF
--- a/examples/user_guide/Links.ipynb
+++ b/examples/user_guide/Links.ipynb
@@ -453,7 +453,7 @@
    ]
   }
  ],
-  "metadata": {
+ "metadata": {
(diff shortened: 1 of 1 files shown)


examples/user_guide/Links.ipynb
```diff
@@ -432,7 +453,7 @@
    ]
   }
  ],
- "metadata": {
+  "metadata": {
```
jbednar:
```suggestion
 "metadata": {
```

(not sure how that stray space got there)

philippjfr: @hoseppan could you apply this fix?

PR #624, opened 2019-08-28
Allow embedding local files

ahuang11:


philippjfr:
I'm not totally sure embed should be the default now. @jbednar any opinion?

ahuang11:
I guess I made it True because it was always the default for a local file.

Also I moved the PR to:
https://github.com/pyviz/panel/pull/625

jbednar:
Makes sense. All things considered, I vote for False, if my more complicated default proposed above is infeasible.

PR #625, opened 2019-08-28
Allow not embedding local file

ahuang11:
I want to not embed local images so that the output HTML size is small, but I'm not sure the best way to update the defaults.

https://github.com/pyviz/panel/issues/617


panel/pane/image.py
```diff
@@ -82,7 +82,7 @@ def _get_properties(self):
         elif self.height is not None:
             width = int((self.height/height)*width)
             height = self.height
-        if self._is_url(self.object) and not self.embed:
+        if self.embed:
```
philippjfr: Isn't this logic the wrong way around?

```suggestion
        if not self.embed:
```

ahuang11: Right, I mistakened self.object as the entire image, but it's actually just the path.

PR #633, opened 2019-09-03
[COMPAT] Force dtype for older numpy

### LINKED ISSUE #632: [CONDA] Require minimum numpy version
[This line](https://github.com/pyviz/panel/blob/72e6fb47c53afc606fe7a96e4da6a3513f5c16c6/panel/layout.py#L818) does not work when using older versions of numpy.

Can be added [here](https://github.com/pyviz/panel/blob/72e6fb47c53afc606fe7a96e4da6a3513f5c16c6/setup.py#L87-L91).

Otherwise, add `dtype=object` arg.

rs2:
Fixes #632

### DIFF
--- a/panel/layout.py
+++ b/panel/layout.py
@@ -807,7 +807,7 @@ class GridSpec(Panel):
 
     @property
     def _object_grid(self):
-        grid = np.full((self.nrows, self.ncols), None)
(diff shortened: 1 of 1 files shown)


jbednar:
Looks good to me, thanks!

PR #639, opened 2019-09-09
Make pn.depends and pn.interact accept widgets

### LINKED ISSUE #638: Add `pn.panelize` or at least improvements to `pn.depends`
## Background
I have been playing around with the [stock APIs example](http://panel.pyviz.org/gallery/apis/stocks_hvplot.html) which is totally great and has really been helping me understand the various APIs. 

In the course of my experiments I was surprised that while you can do:

```python 
def get_plot(ticker, window_size):
    df = get_df(ticker, window_size)
    return df.hvplot.line('date', 'close', grid=True, width=600)

pn.interact(get_plot, ticker=tickers, window_size=(1, 21, 5))
```

When working with `depends` the syntax becomes:

```python
ticker = pn.widgets.Select(name='Ticker', options=tickers)
window = pn.widgets.IntSlider(name='Window Size', value=6, start=1, end=21)

pn.depends(ticker=ticker.param.value, window_size=window.param.value)(get_plot))
```

This is just one example of how the interact and depends methods are different in kind of strange and surprising way. I know that these methods have been discussed at length, but it seems worth ironing out a little more since these are the first places that users get stuck when first starting out with panel.

## Proposal 1 (less ambitious)
To make depends more user-friendly, `pn.depends` should: 
1) accept a function or class method as the first argument, so that the syntax is the same as `pn.interact`
2) accept widgets in addition to parametrized values.
3) return a panel object (as if you had done `pn.Pane(get_plot)`)

### Example 1
PROPOSED:
```python
pn.depends(get_plot, ticker=ticker, window_size=window)
``` 
CURRENT:
```python
pn.depends(ticker=ticker.param.value, window_size=window.param.value)(get_plot)
``` 

### Example 2
PROPOSED:
```python
ticker = pn.widgets.Select(name='Ticker', options=tickers)
window = pn.widgets.IntSlider(name='Window Size', value=6, start=1, end=21)

@pn.depends(ticker=ticker, window_size=window)
def get_plot(ticker, window_size):
    df = get_df(ticker, window_size)
    return df.hvplot.line('date', 'close', grid=True, width=600)

get_plot
```
CURRENT:
```python
ticker = pn.widgets.Select(name='Ticker', options=tickers)
window = pn.widgets.IntSlider(name='Window Size', value=6, start=1, end=21)

@pn.depends(ticker=ticker.param.value, window_size=window.param.value)
def plot(ticker, window_size):
    df = get_df(ticker, window_size)
    return df.hvplot.line('date', 'close', grid=True, width=600)

pn.Pane(plot)
```

## Proposal 2 (more ambitious)
I'd like for there to be a combined decorator (perhaps called `pn.panelize`) that would act like `pn.interact` or like `pn.depends` based on what the inputs to the decorator are. 

### Example 1 (maps to interact)
```python
@pn.panelize(ticker=tickers, window_size=(1, 21, 5))
def get_plot(ticker, window_size):
    df = get_df(ticker, window_size)
    return df.hvplot.line('date', 'close', grid=True, width=600)

get_plot
```

### Example 2 (maps to depends)
```python
ticker = pn.widgets.Select(name='Ticker', options=tickers)
window = pn.widgets.IntSlider(name='Window Size', value=6, start=1, end=21)

@pn.panelize(ticker=ticker, window_size=window)
def get_plot(ticker, window_size):
    df = get_df(ticker, window_size)
    return df.hvplot.line('date', 'close', grid=True, width=600)

get_plot
```

jsignell:
Closes #638 

These all work now:

```python
ticker = pn.widgets.Select(name='Ticker', options=tickers)
window = pn.widgets.IntSlider(name='Window Size', value=6, start=1, end=21)

@pn.depends(ticker, window)
def get_plot(ticker, window_size):
    df = get_df(ticker, window_size)
    return df.hvplot.line('date', 'close', grid=True)

pn.Row(
    pn.Column(title, ticker, window),
    get_plot
)
```

```python
pn.interact(get_plot, ticker=ticker, window_size=window)
```

```python
pn.interact(get_plot, ticker=ticker.param.value, window_size=window.param.value)
```


jsignell:
I haven't fixed all the examples or written tests, just wanted to check with @philippjfr that this seems sane and helpful.

philippjfr:
It looks like you didn't push the depends.py module.

panel/depends.py
```diff
@@ -0,0 +1,14 @@
+import param
```
philippjfr:
```suggestion
from __future__ import absolute_import, division, unicode_literals

import param
```

panel/interact.py
```diff
@@ -447,6 +447,14 @@ def square(num=2):
             def square(num=2):
                 print("{} squared is {}".format(num, num*num))
         """
+        for k, v in kwargs.items():
+            if isinstance(v, param.parameterized.Parameter):
+                pass
+            elif isinstance(v, Widget):
+                kwargs[k] = v.param.value
```
philippjfr: I actually thought interact already accepted widgets. Is that not the case?

jsignell: On second look I believe it is the case. How silly of me.

jsignell:
I don't really think there is any value in `interact` accepting `widget.param.value` in addition to `widget`, so I got rid of all the changes in `interact`

PR #642, opened 2019-09-11
Added server deployment guide

philippjfr:
## Topics

- [x] Server Architecture
- [ ] Authentication (will add example once bokeh 1.4 is released)
- [x] Local usage
- [x] SSH tunneling (copied from bokeh docs) 

## Reverse Proxies

- [x] Nginx (pointed to bokeh docs) 
- [x] Apache (pointed to bokeh docs)

## Platforms

- [ ] AWS
- [ ] Google Cloud
- [x] Heroku
- [ ] Digital Ocean
- [ ] Azure
- [x] Binder (or just myBinder?)
- [ ] Anaconda Enterprise

## Frameworks

- [ ] Flask
- [ ] Django


neuralvis:
is there an easy way to deploy panel apps using tornado ?

philippjfr:
Panel is built on bokeh, which ships with a Tornado server so yes it is.  You can find more details here: https://docs.bokeh.org/en/latest/docs/user_guide/server.html

neuralvis:
Thanks. I am trying to implement the Panel equivalent of the embedded server as illustrated here - https://github.com/bokeh/bokeh/blob/1.4.0/examples/howto/server_embed/tornado_embed.py

Would I just replace the line here `doc.add_root(column(slider, plot))` with a panel object ? i.e. 
`doc.add_root(pn.Column(widget_box, heat map_fn))` for example ?

philippjfr:
You'd do something like (where you replace the panel_obj with whatever panel app you want:

```
def bkapp(doc):
    panel_obj = pn.panel('ABC')
    panel_obj.server_doc(doc)
```

neuralvis:
Many thanks!! I’ll make up an example and contribute it back to the documentation!

PR #643, opened 2019-09-11
Add Comparison section to docs

philippjfr:


jbednar:
I might suggest renaming to Comparisons, as I added other comparisons.  The new stuff could be fleshed out a good bit better, and should be checked for accuracy, but it's a start!

PR #645, opened 2019-09-12
Add run() to StoppableThread class

### LINKED ISSUE #634: Port not being released after stopping threaded holoviz panel app server
Closing a threaded panel app holds on to the port it started on. This is different behavior than closing an app initialized without threading.

```
usgs_logo = pn.panel('../assets/usgs_logo.png', height=130)

column = pn.Column(usgs_logo)

app = column.show(port=8889)

app.stop()
```

Port 8889 is released.

```
app = row.show(port=8889, threaded=True)

app.stop()
```
Port 8889 is not released.

lsetiawan:
## Overview

This PR adds a `run()` method to the `StoppableThread` class. The method overrides the default `run()` from [threading library](https://github.com/python/cpython/blob/master/Lib/threading.py#L871-L886). The result from `_target` is used stop the bokeh server completely, thereby releasing the port.

This should fix #634. 

@friedrichknuth has tested this with his work, and it seemed to solve the issue.

### DIFF
--- a/panel/io/server.py
+++ b/panel/io/server.py
@@ -122,6 +122,15 @@ class StoppableThread(threading.Thread):
         if self.stopped:
             self._cb.stop()
             self.io_loop.stop()
+    
+    def run(self):
+        try:
+            if self._target:
+                bokeh_server = self._target(*self._args, **self._kwargs)
+        finally:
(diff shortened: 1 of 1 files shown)


philippjfr:
Thanks, this is hugely appreciated.

PR #650, opened 2019-09-23
Show should allow a kwarg for title

jsignell:
This doesn't work yet, but I think show should be extended to accept any argument to `panel serve`. It seems that this is how `.show()` works in bokeh.

### DIFF
--- a/panel/base.py
+++ b/panel/base.py
@@ -1,0 +1,74 @@
+from __future__ import absolute_import, division, unicode_literals
+
+from bokeh.io import curdoc as _curdoc
+
+from .io.server import StoppableThread, get_server
+
+
+class ServableMixin(object):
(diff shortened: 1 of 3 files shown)


panel/base.py
```diff
@@ -0,0 +1,74 @@
...
+        """
+        if threaded:
+            from tornado.ioloop import IOLoop
+            loop = IOLoop()
+            server = StoppableThread(
+                target=self._get_server, io_loop=loop,
+                args=(port, websocket_origin, loop, True, True))
+            server.start()
+        else:
+            self.server_doc(title=title)
```
jsignell: This doesn't work FYI

philippjfr:
Superseded by #650

PR #666, opened 2019-09-27
Add ability to render Template inside notebook

philippjfr:
This makes it possible to render custom Templates inside the notebook to see what they look like and to achieve more complex layouts.

### DIFF
--- a/panel/_templates/nb_template.html
+++ b/panel/_templates/nb_template.html
@@ -1,0 +1,43 @@
+{#
+Renders Bokeh models into a basic .html file.
+
+:param title: value for ``<title>`` tags
+:type title: str
+
+:param plot_resources: typically the output of RESOURCES
+:type plot_resources: str
+
+:param plot_script: typically the output of PLOT_SCRIPT
+:type plot_script: str
+
+Users can customize the file output by providing their own Jinja2 template
+that accepts these same parameters.
+
+#}
+
+{% from macros import embed %}
+
+{% block inner_head %}
+  {% block preamble %}{% endblock %}
+  {% block resources %}
+    {% block js_resources %}
+      {{ bokeh_js | indent(8) if bokeh_js }}
+    {% endblock %}
+  {% endblock %}
+{% endblock %}
+{% block postamble %}{% endblock %}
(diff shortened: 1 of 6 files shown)


panel/_templates/nb_template.html
```diff
@@ -0,0 +1,43 @@
...
+        {% for root in doc.roots %}
+          {% block root scoped %}
+            {{ embed(root) | indent(10) }}
+          {% endblock %}
+        {% endfor %}
+      {% endfor %}
+    {% endblock %}
+    {{ plot_script | indent(8) }}
+  {% endblock %}
+{% endblock %}
```
mattpap: Is there a reason for copying this, instead of loading this template from bokeh? Then it wouldn't be necessary it reimplement `html_page_for_render_items()`.

panel/io/notebook.py
```diff
@@ -66,6 +70,41 @@
...
+  var plot = Bokeh.index[plot_id];
+}}
+
+if (plot == null) {{
+  return
+}}
+
+if (plot_id in window.PyViz.receivers) {{
+  var receiver = window.PyViz.receivers[plot_id];
+}} else if (Bokeh.protocol === undefined) {{
```
mattpap: Was this ever an issue? `Bokeh.protocol` should always be present.

panel/_templates/nb_template.html
philippjfr: Yes, this template is stripped from all the standalone HTML page boilerplate including `<html>`, `<head>`, `<body>` tags.

panel/io/notebook.py
philippjfr: `Bokeh.protocol` was not exposed until this PR: https://github.com/bokeh/bokeh/pull/7567

You're right that this is well before the minimum supported version though and can now be removed.

jsignell:
Seems like the panel dependency in holoviews is causing an issue on appveyor.

philippjfr:
>Seems like the panel dependency in holoviews is causing an issue on appveyor.

Fixing appveyor builds in a separate PR.

PR #670, opened 2019-10-01
Updated Markdown example

Jacob-Barhak:
The new example shows a markdown hyper link and how to style table borders using various methods and discusses their advantages and disadvantages

### DIFF
--- a/examples/reference/panes/Markdown.ipynb
+++ b/examples/reference/panes/Markdown.ipynb
@@ -126,7 +126,7 @@
     "\n",
     "It is possible to use html to tags within the markdown text. For example. If we want to define the borders of the table to be thick we can add &lt;style&gt; tag as shown below. However, this technique is recommedned only when working outside notebook environment for the purposes of generating an html file since in notebook environment in will affect the entire notebook. \n",
     "\n",
-    "The code below generates a table with thick borders and saves it as an html file. It is not executed in this notebook to avoid changing the style for all other tables."
+    "(For this reason the code here saves the result to a separate HTML file, to avoid changing the style for all other tables)."
    ]
   },
   {
--- a/examples/reference/panes/Markdown.ipynb
+++ b/examples/reference/panes/Markdown.ipynb
@@ -122,7 +122,7 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "However, the **Style** method only affects the top most object. So we can only set the borders of the table this way, not the cell borders. \n",
+    "However, styles specified in this way will only be applied to the outermost Div, and there is not currently any way to apply styling in this way to specific internal elements of the HTML.  In this case, we cannot use the `style` parameter to control styling of the rows or headings of the generated table. \n",
     "\n",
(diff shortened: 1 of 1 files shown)


philippjfr:
Thanks @Jacob-Barhak. Would you mind clearing the notebook before committing it. Storing the output would bloat the repository and makes it impossible to review.

Jacob-Barhak:
Sure Philip,

James did not mention these details and I was not aware of your continuous
integration system. Will be glad to do it.

         Jacob

examples/reference/panes/Markdown.ipynb
```diff
@@ -89,14 +89,130 @@
...
+    "\"\"\", style={'border': \"4px solid blue\"})"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "However, the **Style** method only affects the top most object. So we can only set the borders of the table this way, not the cell borders. \n",
+    "\n",
+    "It is possible to use html to tags within the markdown text. For example. If we want to define the borders of the table to be thick we can add &lt;style&gt; tag as shown below. However, this technique is recommedned only when working outside notebook environment for the purposes of generating an html file since in notebook environment in will affect the entire notebook. \n",
```
jbednar:
```suggestion
    "If we do want to change specific internal elements of the generated HTML, we can do so by providing an HTML/CSS &lt;style&gt; section. For instance, we can change the border thickness for headers and data as follows, but note that the changes will apply to subsequent Markdown as well, including other cells if in a notebook context:\n",
```

examples/reference/panes/Markdown.ipynb
```diff
@@ -89,14 +89,130 @@
...
+    "\n",
+    "The code below generates a table with thick borders and saves it as an html file. It is not executed in this notebook to avoid changing the style for all other tables."
```
jbednar:
```suggestion
    "(For this reason the code here saves the result to a separate HTML file, to avoid changing the style for all other tables)."
```

examples/reference/panes/Markdown.ipynb
```diff
@@ -89,14 +89,130 @@
...
+   "metadata": {},
+   "source": [
+    "![image.png](attachment:image.png)"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "However, a better solution that allows styling is using css. The code below shows how to style only the table to have a red border without affecting the entire notebook. "
```
jbednar:
```suggestion
    "If you want to change styling only for a specific bit of Markdown text, you can do so, but it requires making a special type of Div as a target for the styling. E.g. here we can have only the following table use a red border, without affecting the entire notebook:"
```

Jacob-Barhak:
Thanks James,

The text in the suggested changes are much better.

philippjfr:
@Jacob-Barhak Could you apply @jbednar's suggestions?

jbednar:
Looks good to me now!

PR #686, opened 2019-10-12
Support file like objects for images

### LINKED ISSUE #610: Support reading from BytesIO/StringIO
I'm trying to write a GIF to memory and use panel to read from memory rather than storage.
`ValueError: GIF pane does not support objects of type 'BytesIO'`

```
from io import BytesIO
import imageio

memory = BytesIO()

save_fps = [..]

with imageio.get_writer(memory, mode='I', format='gif', fps=8) as writer:
    for fp in save_fps:
        image = imageio.imread(fp)
        writer.append_data(image)

memory.seek(0)
pn.pane.GIF(memory)
```

RedBeardCode:
Should fix #610


philippjfr:
Thanks for joining the sprint!

PR #692, opened 2019-10-12
Issue 488/remove mutable default args in io model

xtaje:


panel/io/save.py
```diff
@@ -62,8 +62,12 @@ def save(panel, filename, title=None, resources=None, template=None,
       Optional title for the plot
     resources: bokeh resources
       One of the valid bokeh.resources (e.g. CDN or INLINE)
+    template:
+      template file, as used by bokeh.file_html. If None will use bokeh defaults
+    template_variables:
+      template_variables file dict, as used by bokeh.file_html
     embed: bool
-      Whether the state space should be embedded in the saved file.
+      Whether the state space should be embedded in the saved file. If None use bokeh defaults
```
philippjfr: There is no bokeh default for the `embed` functionality.

```suggestion
      Whether the state space should be embedded in the saved file.
```

PR #702, opened 2019-10-13
Prepare support for python 3.8

### LINKED ISSUE #701: Python3.8 compatibility
python3.8 comes with a number of smaller API breaks that we need to account for, lest they become bugs as soon as people try running panel with the new version.

a-recknagel:
closes #701

### DIFF
--- a/panel/util.py
+++ b/panel/util.py
@@ -9,7 +9,8 @@ import inspect
 import numbers
 import datetime as dt
 
-from collections import defaultdict, MutableSequence, MutableMapping, OrderedDict
+from collections import defaultdict, OrderedDict
+from collections.abc import MutableSequence, MutableMapping
 from datetime import datetime
 from six import string_types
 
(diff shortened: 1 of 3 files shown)


philippjfr:
Thanks @a-recknagel, much appreciated!

panel/util.py
```diff
@@ -9,9 +9,13 @@
...
 import datetime as dt
 
-from collections import defaultdict, MutableSequence, MutableMapping, OrderedDict
 from datetime import datetime
 from six import string_types
+from collections import defaultdict, OrderedDict
+try:
+    from collections.abc import MutableSequence, MutableMapping
+except ImportError:  # support for python>3.8
+    from collections import MutableSequence, MutableMapping
```
jbednar: Isn't this comment backwards?  I.e., isn't the `try` clause (collections.abc) required for python>3.8, and the `except` clause is for some older version (<3.3)?

a-recknagel: yeah, I put the comparison the wrong way round, should have been `python<3.8`

a-recknagel: oh, it was already merged. I'll include it in the docs PR

jbednar: Ok, but it's not true the other way around either; the `except` isn't needed for python<3.8, only for python<3.3.

a-recknagel: Ah, thanks for pointing that out.

PR #703, opened 2019-10-13
Audio widget uses numpy array and sample_rate  as input

Karamya:
Fixes the issue : https://github.com/pyviz/panel/issues/351


philippjfr:
Thanks again @Karamya, much appreciated!

PR #704, opened 2019-10-13
Make the project ✨🖤✨

### LINKED ISSUE #693: Blacken project
Formatting issues come up every now and then, it's nice to get rid of them by using black. I created https://github.com/pyviz/panel/tree/ar/blacken where I ran black with no config and only touched up a few params in viewable.py to have something we can look at to identify issues.

a-recknagel:
Would fix #693, but needs to be re-build because of millions of merge conflicts in case it is accepted. This draft right now only serves to help with discussion.

### DIFF
--- a/panel/__init__.py
+++ b/panel/__init__.py
@@ -2,21 +2,24 @@ from __future__ import absolute_import, division, unicode_literals
 
 import param as _param
 
-from . import layout # noqa
-from . import links # noqa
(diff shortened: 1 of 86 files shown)


philippjfr:
Thanks again @a-recknagel. No decision on blackening yet but I'll close for now since it has to be redone anyway.

PR #708, opened 2019-10-14
Fix panes link in Parameters user guide docs

### LINKED ISSUE #628: [DOC] /user_guides/Panes.html is missing
#### ALL software version info:
Whatever was used to generate https://panel.pyviz.org/.

No hyperlinks in https://github.com/pyviz/panel/blob/master/examples/user_guide/Param.ipynb yield a 404.

https://panel.pyviz.org/user_guide/Panes.html is not available, and I could not find suitable candidates in https://github.com/pyviz/panel/tree/master/examples/user_guide. See https://github.com/pyviz/panel/pull/466/files#diff-ce01fb962ef569b34c1649450526e27cR238. for the diff that introduced it.

a-recknagel:
Fixes #628

### DIFF
--- a/examples/user_guide/Param.ipynb
+++ b/examples/user_guide/Param.ipynb
@@ -266,7 +266,7 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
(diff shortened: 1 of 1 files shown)


philippjfr:
Thanks for tracking that down!

PR #709, opened 2019-10-14
Fixing Introduction link to work in both rendered docs as well as notebooks.

a-recknagel:
See https://github.com/pyviz/panel/pull/705#discussion_r334467157


philippjfr:
Thanks, we'll have to see if this actually works in a website build.

PR #719, opened 2019-10-22
Drop 2.7 and 3.5 support and prepare for mypy

mattpap:
I went ahead and dropped support for 2.7 in light of inevitable 2.7's EOL and bokeh's 2.0 release. I dropped Python 3.5 as well, per BEP 6 and NEP 29. Given the extent of touched files, I added `__all__` across `panel`. For that purpose I introduced `@public` decorator, which I originally implemented in `sympy`, that follows DRY principle and fills `__all__` automatically. I also added tests, to make sure all modules are covered. Finally I added rudimentary support for static typing with mypy. The initial goal is to help with dropping old Python versions and I would like `mypy panel` to result in no errors on weakest settings. From there we can discuss if static typing is actually desired in this project, and if so, proceed in future PRs with using stricter settings and more type coverage.


mattpap:
Looks like testing on Windows is broken. I wasn't able to make it past conda install at least once.

philippjfr:
>Looks like testing on Windows is broken. I wasn't able to make it past conda install at least once.

Yes, something is definitely broken on appveyor. Will need to investigate.

philippjfr:
Closing for now since there are a lot of conflicts.

PR #721, opened 2019-10-25
Adding min_characters to AutocompleteInput

jsignell:
This is to match the bokeh model which exposes this option.

### DIFF
--- a/panel/widgets/select.py
+++ b/panel/widgets/select.py
@@ -137,15 +137,15 @@ class MultiSelect(Select):
 
 class AutocompleteInput(Widget):
 
+    min_characters = param.Integer(default=2, doc="""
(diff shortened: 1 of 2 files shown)


xavArtley:
It works well for me

jsignell:
Yeah I updated my comment. I had the wrong version.

panel/widgets/select.py
```diff
@@ -143,6 +143,10 @@ class AutocompleteInput(Widget):
 
     value = param.Parameter(default=None)
 
+    min_characters = param.Integer(default=2, doc="""
+        The number of characters a user must type before
+        completions are presented.""")
+
```
philippjfr: Nitpick, let's keep parameters in alphabetical order.

PR #724, opened 2019-10-29
Select behavior

xavArtley:
Related to https://github.com/pyviz/panel/issues/718

```python
import panel as pn
pn.extension()
sel = pn.widgets.Select(value='val1', options={'lab1': 'val1', 'lab2': 'val2'})
sel.jscallback(value="""console.log(source.value); console.log(...source.options)""")
sel
```
With this PR on js side the change of the select value gives
image

### DIFF
--- a/panel/tests/test_param.py
+++ b/panel/tests/test_param.py
@@ -291,32 +291,32 @@ def test_object_selector_param(document, comm):
     test_pane = Pane(test)
     model = test_pane.get_root(document, comm=comm)
 
-    slider = model.children[1]
-    assert isinstance(slider, Select)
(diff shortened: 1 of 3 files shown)


philippjfr:
Could you explain what happens here if the actual value cannot be represented in Javascript?

philippjfr:
That's what I assumed your issue was but it doesn't really answer my question. My question was about what happens if the value of the Select widget is some complex Python object that cannot be serialized to Javacript?

xavArtley:
It pass just a string on the javascript side wich is useless but it does not affect the current behavior

Issue #727, opened 2019-10-31
panel save not working with last bokeh and panel


xavArtley:
after using `pan.save('test.html')` when I try to reload the file exported I get this error:
```
bokeh-1.4.0rc2.min.js:83 Uncaught Error: Cannot find module '@bokehjs/models/widgets/widget'
    at require (bokeh-1.4.0rc2.min.js:81)
    at Object._ (test.html:126)
    at require (bokeh-1.4.0rc2.min.js:88)
    at Object._ (test.html:124)
    at require (bokeh-1.4.0rc2.min.js:88)
    at Object._ (test.html:123)
    at require (bokeh-1.4.0rc2.min.js:88)
    at Object.main.register_plugin (bokeh-1.4.0rc2.min.js:115)
    at test.html:117
    at test.html:122
bokeh-1.4.0rc2.min.js:243 Uncaught Error: Model 'panel.models.vtk.VTKPlot' does not exist. This could be due to a widget or a custom model not being registered before first usage.
    at Object.o.Models (bokeh-1.4.0rc2.min.js:243)
    at Function.e._instantiate_object (bokeh-1.4.0rc2.min.js:242)
    at Function.e._instantiate_references_json (bokeh-1.4.0rc2.min.js:242)
    at Function.e.from_json (bokeh-1.4.0rc2.min.js:242)
    at v (bokeh-1.4.0rc2.min.js:240)
    at bokeh-1.4.0rc2.min.js:240
```

xavArtley:
did you try with custom extension?
Like :
```
import panel as pn
pn.extension('vtk')

dragon = pn.pane.VTK('https://raw.githubusercontent.com/Kitware/vtk-js/master/Data/StanfordDragon.vtkjs',
                     sizing_mode='stretch_width', height=400)
dragon.save('test.html')
```

philippjfr:
Yes I did.

philippjfr:
Oh right, I know what you mean now. Yes, I'm seeing the same thing.

philippjfr:
Raised the issue here: https://github.com/bokeh/bokeh/issues/9343

philippjfr:
Fixed upstream.

PR #626, opened 2019-08-28
Update panel to use `bokeh build`

mattpap:
This is a preview based on a WIP PR https://github.com/bokeh/bokeh/pull/9198, where the ad-hoc extensions compiler was replaced with a new, more robust one. One can build panel with `bokeh build panel` (assuming cwd is the repository) or with `bokeh.ext.build("panel")` (which is used in `setup.py`). The changes in this PR reflect how the extension should (almost) ideally look, barring any backwards compatibility issues. Almost, because e.g. `require_components()` could be managed by bokeh, or `__module__` fields could be inserted automatically, similarly to `__name__`. The extension layout isn't set in stone, so if there are any reservations regarding the current approach, then things can still change.

Example output:
```
~/repo/panel $ bokeh build panel/
Working directory: /home/mateusz/repo/panel/panel
Using different version of bokeh, rebuilding from scratch.
Running npm install.
added 32 packages from 50 contributors and audited 272 packages in 15.457s
found 0 vulnerabilities

Using /home/mateusz/repo/panel/panel/tsconfig.json
Compiling TypeScript (13 files)
Linking modules
Output written to /home/mateusz/repo/panel/panel/dist
All done.
```

PR #729, opened 2019-10-31
update vtkvolume model to works with new bokeh build

xavArtley:
fix : https://github.com/pyviz/panel/issues/728


mattpap:
Looks good. I must have overlooked this recent addition when rebasing PR #626. It would be good to test if this and all other models are importable in notebooks and under other embedding methods, especially in light of issue #727. Though I wouldn't block merging this PR on that.

Issue #734, opened 2019-10-31
Pipeline network diagram should be clickable


jbednar:
At some point, the nodes in the pipeline layout diagram were clickable, selecting that stage.

![image](https://user-images.githubusercontent.com/1695496/67988199-3d954080-fbfd-11e9-8392-06954e69b94d.png)

Hover currently (in Panel and HoloViews master) works for the nodes, but clicking doesn't do anything.  I think that clicking should be supported in at least the following cases:
- [ ] clicking on the previous and next nodes should work precisely the same as clicking on the Previous and Next buttons.  Clicking on the diagram is a very clear way for a user to unambiguously select a stage, and supporting that also lets a panel designer suppress those buttons to save screen real estate without losing any functionality.
- Clicking on arbitrary stages should also be supported, if one could already just click forward or back to get there, and it should have the same semantics as doing so (being supported in all the same cases and having the same effect).
- Clicking on the start node should always be supported, even for a Pipeline that defines a `ready_parameter`, with results equivalent to reloading the page if moving backwards to that node is not otherwise allowed using the Previous button.  I.e., it always makes sense to start completely over (right?), given that a user can already reload the page, so I think the diagram should allow that behavior.  Otherwise it's frustrating for users -- they can see what they want to do, but no way to make the pipeline do it, even when what they want to do is inherently appropriate and reasonable.

Issue #733, opened 2019-10-31
Pipeline should display itself, without requiring .layout


jbednar:
A Pipeline object can be displayed in Panel using the `.layout` attribute:

![image](https://user-images.githubusercontent.com/1695496/67987772-43d6ed00-fbfc-11e9-87df-a5745d706a4e.png)

Can't the Pipeline simply _be_ a Layout (inheriting from ListPanel?), or at least have the layout as its rich display, so that typing `pipeline` here would work the same as `pipeline.layout`?  That seems much more obvious and straightforward compared to remembering to use `.layout`.

Issue #732, opened 2019-10-31
Pipeline should be responsive in width 


jbednar:
The pipeline title, network, and buttons don't fit into a default Jupyter notebook width, cutting off the buttons (in some cases so that I didn't even realize there were any buttons). E.g.:
![image](https://user-images.githubusercontent.com/1695496/67987772-43d6ed00-fbfc-11e9-87df-a5745d706a4e.png)
@philippjfr suggested that there was a minimum width being enforced, and if so it should at least be narrower than the default Jupyter notebook width.

Issue #731, opened 2019-10-31
Can Pipeline automatically hide widgets for already-connected inputs?


jbednar:
The first example in the Pipelines user manual page shows a two-stage pipeline with Stage1's `c` output connected to Stage2's `c` input.  By default, displaying parameters for Stage2 in a Pipeline would generate a widget for `Stage2.c`, which would be confusing (and arguably incorrect) because that parameter is actually being driven by Stage1, and is _not_ appropriate for the user to manipulate as a widget.  To avoid confusion, the suggested example suppresses `Stage2.c`'s widget by setting it to a negative precedence.
```
class Stage2(param.Parameterized):
    c = param.Number(default=5, precedence=-1, bounds=(0, None))
```
That works, but it isn't ideal, because it means that Stage2 can't fully be tested, evaluated, and used on its own (as Panel will only create widgets for the other parameters, not for `c`).  It's also a bit confusing for users, who don't often need to know about precedence.

More seriously (or more abstractly?), having to edit the definition of `c` means that we can't use the same class definition for a Pipeline stage and just as a Parameterized object or Panel component in other contexts.  If it were possible to hide `c` automatically when it's used in a Pipeline, then making a given Parameterized object usable as a Pipeline stage won't affect any other purposes it already fills, which means that people could set up whole suites of stage-ready Parameterized objects usable on their own, in any combination in a Panel or notebook, and/or in a Pipeline (perhaps one created using [boxflow](https://github.com/ioam/boxflow)).

I'm not sure what the best solution would be.  One approach would be for Pipeline's `add_stage` method to check the precedence for any parameter it is driving from a previous stage, and set the precedence on that instance's parameter object to negative if the precedence was at its default value.  That's a bit intrusive, though, because when that instance is displayed outside the Pipeline the widget will now be gone, but at least it would mean that at the source code level the Parameterized object won't have to know which of its parameters might be being driven by a previous stage, as desired.  

Other approaches might require changes at the Param level, to support some way other than precedence for Pipeline to indicate that a given Parameter should temporarily not appear as a widget in the context of a Pipeline, even though it would appear as a widget in other contexts.

PR #735, opened 2019-10-31
Made Pipelines user guide consistent and clearer

jbednar:
Main changes:

- The first section was written as if there was only one output, but the code had two; looks like the markdown wasn't updated when the code was
- Removed stray `, ready_parameter='ready', auto_advance=True` in early example; looks like something meant for the last example
- Clarified hierarchical arrangement of `.layout`
- Cleaned up descriptions and narrative flow

I opened some issues found while reviewing this (#731, #732, #733, #734). When those are fixed, the text in this notebook will need to be updated accordingly, so this PR should probably be merged before addressing any of those (or updated each time).  

One additional issue is that I wasn't sure why `.init()` needed to be called or precisely when it should be called; isn't there some way that it can be called automatically?  It's a confusing, stateful thing...

### DIFF
--- a/examples/user_guide/Pipelines.ipynb
+++ b/examples/user_guide/Pipelines.ipynb
@@ -16,7 +16,9 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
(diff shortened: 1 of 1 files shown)


philippjfr:
The init issue can probably be solved thinking about it. The issue was that it wasn't obvious how to set up the graph when you declare a simple linear pipeline without calling define_graph, but if we simply build up the linear pipeline as stages are added and then override if define_graph is called there is no problem.

jbednar:
That sounds right from what I remember of that code, thanks.

philippjfr:
Changes look good thanks!

PR #737, opened 2019-11-01
Various Pipeline fixes

### LINKED ISSUE #563: Disable pan tool on pipeline subway map
In https://panel.pyviz.org/user_guide/Pipelines.html, you can drag the pipeline object around.

Would it be possible to also drag and resize gridspecs (make the top pane larger and bottom pane smaller by dragging the center point, like how you can move/resize windows in an OS).

```
import panel as pn
pn.extension()

gs = pn.GridSpec()
gs[0:25, 0:50] = pn.Spacer(background='Red')
gs[25:50, 0:50] = pn.Spacer(background='Blue')
gs
```

image

### LINKED ISSUE #731: Can Pipeline automatically hide widgets for already-connected inputs?

### LINKED ISSUE #732: Pipeline should be responsive in width

### LINKED ISSUE #733: Pipeline should display itself, without requiring .layout

### LINKED ISSUE #734: Pipeline network diagram should be clickable

### LINKED ISSUE #736: Call pipeline.init() automatically
The pipeline.init() method is confusing; it's hard to tell when it needs to be call, what it does, or why it can't be called automatically.  Basically, it introduces state that makes the user have to reason about it, without extensive docs that explain how to do that reasoning.

@philippjfr suggests that it might indeed be possible to avoid having to call it explicitly.  "The issue was that it wasn't obvious how to set up the graph when you declare a simple linear pipeline without calling define_graph, but if we simply build up the linear pipeline as stages are added and then override if define_graph is called there is no problem."

philippjfr:
- [x] Fixes #563 
- [x] Fixes #731
- [x] Fixes #732
- [x] Fixes #733
- [x] Fixes #736

### DIFF
--- a/examples/user_guide/Pipelines.ipynb
+++ b/examples/user_guide/Pipelines.ipynb
@@ -147,7 +147,7 @@
    "source": [
     "class Stage2(param.Parameterized):\n",
     "    \n",
-    "    c = param.Number(default=5, precedence=-1, bounds=(0, None))\n",
+    "    c = param.Number(default=5, bounds=(0, None))\n",
     "\n",
     "    exp = param.Number(default=0.1, bounds=(0, 3))\n",
     "    \n",
@@ -195,14 +195,14 @@
     "pipeline = pn.pipeline.Pipeline(debug=True)\n",
     "pipeline.add_stage('Stage 1', Stage1())\n",
     "pipeline.add_stage('Stage 2', Stage2)\n",
-    "pipeline.layout"
+    "pipeline"
    ]
   },
   {
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "As you can see the ``Pipeline`` renders a little diagram displaying the available stages in the workflow along with previous and next buttons to move between each stage. This allows setting up complex workflows with multiple stages, where each component is a self-contained unit, with minimal declarations about stage outputs (using the ``param.output`` decorator) and how to render the stage (by declaring a ``panel`` method).\n",
+    "As you can see the ``Pipeline`` renders a little diagram displaying the available stages in the workflow along with previous and next buttons to move between each stage. This allows setting up complex workflows with multiple stages, where each component is a self-contained unit, with minimal declarations about stage outputs (using the ``param.output`` decorator) and how to render the stage (by declaring a ``panel`` method).  Note also when progressing to Stage 2, the `c` parameter widget is not rendered because it's value has been provided by the previous stage. \n",
     "\n",
     "Above we created the ``Pipeline`` as we went along, which makes some sense in a notebook to allow debugging and development of each stage.  When deploying the Pipeline as a server app or when there's no reason to instantiate each stage separately; instead we can declare the stages as part of the constructor:"
    ]
@@ -214,7 +214,7 @@
    "outputs": [],
    "source": [
     "pipeline = pn.pipeline.Pipeline([('Stage 1', Stage1), ('Stage 2', Stage2)])\n",
-    "pipeline.layout"
+    "pipeline"
    ]
   },
   {
@@ -310,7 +310,7 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "dag.layout"
+    "dag"
    ]
   },
   {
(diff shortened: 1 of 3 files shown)


jbednar:
Thanks! Looks good so far. One more thing:

- [ ] Pipelines user guide updated to reflect these changes.

examples/user_guide/Pipelines.ipynb
```diff
@@ -195,14 +211,15 @@
...
+    "\n",
+    "pipeline"
    ]
   },
   {
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "As you can see the ``Pipeline`` renders a little diagram displaying the available stages in the workflow along with previous and next buttons to move between each stage. This allows setting up complex workflows with multiple stages, where each component is a self-contained unit, with minimal declarations about stage outputs (using the ``param.output`` decorator) and how to render the stage (by declaring a ``panel`` method).\n",
+    "As you can see the ``Pipeline`` renders a little diagram displaying the available stages in the workflow along with previous and next buttons to move between each stage. This allows setting up complex workflows with multiple stages, where each component is a self-contained unit, with minimal declarations about stage outputs (using the ``param.output`` decorator) and how to render the stage (by declaring a ``panel`` method).  Note also when progressing to Stage 2, the `c` parameter widget is not rendered because it's value has been provided by the previous stage. \n",
```
philippjfr:
```suggestion
    "As you can see the ``Pipeline`` renders a little diagram displaying the available stages in the workflow along with previous and next buttons to move between each stage. This allows setting up complex workflows with multiple stages, where each component is a self-contained unit, with minimal declarations about stage outputs (using the ``param.output`` decorator) and how to render the stage (by declaring a ``panel`` method).  Note also when progressing to Stage 2, the `c` parameter widget is not rendered because its value has been provided by the previous stage. \n",
```

jbednar:
```suggestion
    "As you can see the ``Pipeline`` renders a little diagram displaying the available stages in the workflow along with previous and next buttons to move between each stage. This allows setting up complex workflows with multiple stages, where each component is a self-contained unit, with minimal declarations about stage outputs (using the ``param.output`` decorator) and how to render the stage (by declaring a ``panel`` method).  Note also when progressing to Stage 2, the `c` parameter widget is not rendered because its value has been provided by the previous stage. \n",
```

PR #743, opened 2019-11-06
Remove object attribute from Panel class

### LINKED ISSUE #546: GridSpec objects attribute violates Panel interface contract
The `Panel` class provides an `objects` attribute that is expected to contain a list of child objects:

```python
class Panel(Reactive):
   ...
    objects = param.Parameter(default=[], doc="""
        The list of child objects that make up the layout.""")
```

but `GridSpec` overrides this with a dictionary:

```python
class GridSpec(Panel):
   ...
    objects = param.Dict(default={}, doc="""
        The dictionary of child objects that make up the grid.""")
```

Consequently any code that is meant to operate on a `Panel .object` list is likely to break when applied to a `GridSpec`. In particular, the `Panel._cleanup`, which is inherited by `GridSpec` will fail because iterating over the objects will return tuples instead of the actual child objects.

For this case, you could probably fix this by overriding `_cleanup` in `GridSpec` but it would not fix the underlying issue.

Observed using Panel 0.6.0.

kleavor:
Closes #546 

Inheriting classes define this attribute individually.

### DIFF
--- a/panel/layout.py
+++ b/panel/layout.py
@@ -24,9 +24,6 @@ class Panel(Reactive):
     Abstract baseclass for a layout of Viewables.
     """
 
-    objects = param.Parameter(default=[], doc="""
-        The list of child objects that make up the layout.""")
-
     _bokeh_model = None
 
     __abstract = True


philippjfr:
Thanks! I think the _get_objects method should probably also be moved to the ListPanel class.

PR #752, opened 2019-11-07
Fix Django example app

### LINKED ISSUE #445: panel/examples/apps/django2/ example doesn't work
The django2 example doesn't work at least for Django 2.2. The interactive plot doesn't show up and there are no clear error messages either. However, the same example provided by ParamBokeh works fine. But I prefer Panel if this problem can be solved.

eddienko:
Changed _get_root to server_doc.

Fixes #445


philippjfr:
Thanks for the fix.

PR #774, opened 2019-11-13
Various Template improvements

philippjfr:


panel/template.py
```diff
@@ -56,7 +57,8 @@ class Template(object):
     served as a standalone server and when used in the notebook.
```
jbednar: I'm not sure what "The Template object should be served as a standalone server and when used in the notebook" means; can you explain?  And can you explain how Template the class relates to a string template, here in the docstring?

PR #785, opened 2019-11-14
Documentation fixes

philippjfr:


doc/index.rst
```diff
@@ -12,10 +12,10 @@ A high-level app and dashboarding solution for Python
    <table>
      <tr>
        <td border=1><a href="https://examples.pyviz.org/attractors/clifford_panel.html"><b>Interact</b></a><br><a href="https://clifford.pyviz.demo.anaconda.com"><img src="_images/clifford.png" /></a></td>
-       <td border=1><a href="https://examples.pyviz.org/gapminders/gapminders.html"><b>Gapminders</b></a><br><a href="https://gapminder.pyviz.demo.anaconda.com"><img src="_images/gapminders.png" /></a></td>
+       <td border=1><a href="https://examples.pyviz.org/gapminders/gapminders.html"><b>Gapminders</b></a><br><a href="https://gapminder.pyviz.demo.anaconda.com"><img src="_images/gapminder.png" /></a></td>
```
jbednar: The actual app is called "gapminders" with an "s", so it's odd that the thumbnail would be called "gapminder"; maybe rename the file instead?

doc/Comparisons.rst
```diff
@@ -27,9 +27,10 @@ Panel and Dash can both be used to create dashboards in Python, but take very di
...
+
+  * Dash's approach is more highly scalable in some cases, allowing many simultaneous client sessions without necessarily using up resources on the server for each new client. (That said, Dash's server only processes a single request at any one time and is `"not suitable for production" <https://dash.plot.ly/deployment>`__, so the differences may not be that large in practice.)
```
jbednar:
```suggestion
  * Dash's approach is more highly scalable in some cases, allowing many simultaneous client sessions without necessarily using up resources on the server for each new client.
```

PR #795, opened 2019-11-16
Update packaging metadata

ceball:
I came here just to add python 3.7 to the list. However, I noticed a few other things at the same time. So close to the release, you might want to stick to the python 3.7 change only and make the others after the release - let me know if so and I'll re-do this PR.

Other changes:
  * pyproject.toml was a bit out of date vs. `extras_require['build']`.
  * Removed older python 3 versions - I looked at bokeh's list, assuming it's likley the same for panel (?).
  * While I was looking at bokeh, I noticed it had a lot more relevant trove classifiers - pyviz's trove classifiers still only ever talk about science...and don't mention visualization. (At the same time, no idea if anyone uses these classifiers now.)

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,7 +1,8 @@
 [build-system]
 requires = [
-    "param >=1.7.0",
-	"bokeh >=1.4.0",
+    "param >=1.9.0",
     "pyct >=0.4.4",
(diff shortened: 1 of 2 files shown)


philippjfr:
Thanks for catching this before the release. All the changes look good.

PR #812, opened 2019-11-25
Fixed handling of state._views for Templates

philippjfr:
In order for pre-processing to work correctly on Templates we need to create a fake root that is shared by all sub-objects. Previously this fake root was not added to the `state._views` which is a global index of all roots, which ends up causing issues when updating a root.

Cc: @dharhas this should fix the issues you've been seeing


dharhas:
@philippjfr I can confirm that this fixes the lsst_dashboard issue.

Issue #808, opened 2019-11-21
`value` not updating (on objects with matching hashes?)


jsignell:
`value` is not updating when a new value is selected for the selection widget. I am fairly confident this is a regression having to do with how panel hashes objects in widget option dicts.

### example
```python
import geoviews as gv
import geoviews.tile_sources as gvts
import panel as pn

gv.extension('bokeh')
basemap = pn.widgets.Select(options=gvts.tile_sources)

@pn.depends(basemap=basemap)
def view(basemap):
    return basemap

pn.Row(basemap, view)
```

### context
This is the repr of gvts.tile_sources:

```python
{'CartoDark': :WMTS   [Longitude,Latitude],
 'CartoEco': :WMTS   [Longitude,Latitude],
 'CartoLight': :WMTS   [Longitude,Latitude],
 'CartoMidnight': :WMTS   [Longitude,Latitude],
 'StamenTerrain': :WMTS   [Longitude,Latitude],
 'StamenTerrainRetina': :WMTS   [Longitude,Latitude],
 'StamenWatercolor': :WMTS   [Longitude,Latitude],
 'StamenToner': :WMTS   [Longitude,Latitude],
 'StamenTonerBackground': :WMTS   [Longitude,Latitude],
 'StamenLabels': :WMTS   [Longitude,Latitude],
 'EsriImagery': :WMTS   [Longitude,Latitude],
 'EsriNatGeo': :WMTS   [Longitude,Latitude],
 'EsriUSATopo': :WMTS   [Longitude,Latitude],
 'EsriTerrain': :WMTS   [Longitude,Latitude],
 'EsriReference': :WMTS   [Longitude,Latitude],
 'EsriOceanBase': :WMTS   [Longitude,Latitude],
 'EsriOceanReference': :WMTS   [Longitude,Latitude],
 'OSM': :WMTS   [Longitude,Latitude],
 'Wikipedia': :WMTS   [Longitude,Latitude]}
```

### workaround
```python
basemap = pn.widgets.Select(options=list(gvts.tile_sources.keys()))

@pn.depends(basemap=basemap)
def view(basemap):
    return getattr(gvts, basemap)()

pn.Row(basemap, view)
```

PR #816, opened 2019-11-25
Ensure that JS-side Select values are unique

philippjfr:
Fixes https://github.com/holoviz/panel/issues/808

Cc: @xavArtley I've tried to retain the previous behavior when the stringified values are unique but switch back to the guaranteed to be unique labels otherwise.


xavArtley:
Very well, Thanks for keeping both behaviors

philippjfr:
Not totally happy about it tbh, but can't see a better solution.

PR #817, opened 2019-11-26
[VTK] Add axes coordinates to VTK view

xavArtley:
First prototype of axes for VTK panel:

```python
import panel as pn
import numpy as np

axes = dict(
    origin = [10, 0, -5],
    xticks = np.linspace(-10,10,5),
    yticks = np.linspace(0,12,11),
    zticks = np.linspace(-5,5,5),
)
dragon = pn.pane.VTK('https://raw.githubusercontent.com/Kitware/vtk-js/master/Data/StanfordDragon.vtkjs',
                     sizing_mode='stretch_width', height=400, enable_keybindings=True, orientation_widget=True,
                     serialize_on_instantiation=True, axes=axes)
dragon.show()
```
ezgif com-video-to-gif (18)

In the same time I restructure typescript code of VTK pane by splitting in smaller files in a repository specific to vtk

TODO:
- [x] : grid opacity, axes opacity
- <del> axes color, grid color </del>
- [x] : separate ticks position and ticks label (number and text)
- [x] : add possibility to set tick label style properties (font
- [x] : add listener on axes change
- [x] : don't remove axes actor when panel `object` is changed
- [x] : Add documentation


panel/package.json
```diff
@@ -5,7 +5,9 @@
   "license": "BSD-3-Clause",
   "repository": {},
   "dependencies": {
-    "bokehjs": "^1.3.4"
+    "@types/gl-matrix": "^2.4.5",
+    "bokehjs": "^1.3.4",
+    "gl-matrix": "^3.1.0"
```
philippjfr: I wasn't aware that you could actually load dependencies like this yet. What ends up loading the dependency?

xavArtley: I suppose node and the typescript compiler to create the .js

xavArtley:
ezgif com-video-to-gif (2)

xavArtley:
ezgif com-video-to-gif (3)

philippjfr:
Thanks for the PR! At some point we need to figure out how to test the Vtk code.

PR #820, opened 2019-11-27
Add ability to declare OAuth provider

philippjfr:
Adds a number of inbuilt OAuth providers to Panel which you can activate via the config and eventually by supplying options to `panel serve`.

- [x] Add GitHub OAuth Provider
- [x] Add AzureAd OAuth Provider 
- [x] Add Bitbucket OAuth Provider 
- [x] Add Google OAuth Provider
- [x] Add GitLab OAuth Provider
- [x] Add ability to declare OAuth provider in config and via env variables
- [x] Ensure the OAuth providers can be declared using `panel serve`
- [x] Expose the current user info in a convenient way
- [x] Make AuthProviders pluggable
- [x] Add docs 
- [x] Add ability to logout
- [ ] Add examples

e.g. given an app like this:

```python
import hvplot.pandas
import panel as pn
import holoviews as hv

from bokeh.sampledata.sea_surface_temperature import sea_surface_temperature

bins = pn.widgets.IntSlider(name='Number of bins', start=5, end=25, step=5, value=10)
kde = pn.widgets.Checkbox(name='Show density estimate')
observations = pn.widgets.Checkbox(name='Show individual observations')
bandwidth = pn.widgets.FloatSlider(name='KDE Bandwidth', start=0.1, end=1)

@pn.depends(bins, kde, observations, bandwidth)
def get_plot(bins, kde, obs, bw):
    plot = sea_surface_temperature.hvplot.hist(bins=bins, normed=True, xlabel='Temperature (C)', padding=0.1)
    if kde:
        plot *= sea_surface_temperature.hvplot.kde().opts(bandwidth=bw)
    if obs:
        plot *= hv.Spikes(sea_surface_temperature, 'temperature').opts(
            line_width=0.1, alpha=0.1, spike_length=-0.01)
    return plot

widgets = pn.WidgetBox('## Sea surface temperatures', bins, observations, kde, logout)

def add_bandwidth(event):
    if event.new:
        widgets.append(bandwidth)
    else:
        widgets.remove(bandwidth)

kde.param.watch(add_bandwidth, 'value')

pn.Row(widgets, get_plot).show()
```

You can run it with:

```
PANEL_OAUTH_PROVIDER='github' PANEL_OAUTH_KEY=... PANEL_OAUTH_SECRET=... python app.py
```

or:

```
panel serve test_oauth.py --oauth-provider="github" --oauth-key="..." --oauth-secret="..."
```

and get something like:

Screen Shot 2019-11-27 at 4 26 42 PM

Cc: @ceball @sdc50 @jbednar


philippjfr:
I'd particularly value any opinions on how best to make AuthProviders pluggable. Bokeh allows defining a module alongside the app which provides the auth provider but I'd prefer a cleaner way of providing one.

jbednar:
Cool. I don't have an immediate ideas about making it pluggable.

sdc50:
> I'd particularly value any opinions on how best to make AuthProviders pluggable.

Would it work to use namespace plugins, similar to what Intake does (https://github.com/intake/intake/blob/master/intake/source/discovery.py#L271)? 

Something like this might work:

```python
panel_oauth_login_handlers = dict('github': GithubLoginHandler)

import panel_oauth
for _, modname, ispkg in pkgutil.iter_modules(panel_oauth.__path__):
    plugin_path = f'panel_oauth.{modname}'
    plugin_module = importlib.import_module(plugin_path)
    for name, cls in inspect.getmembers(plugin_module, inspect.isclass):
        if issubclass(cls, tornado.web.RequestHandler):
            panel_oauth_login_handlers[modname] = cls




class OAuthProvider(AuthProvider):

    ...

    @property
    def login_handler(self):
        return panel_oauth_login_handlers.get(config.oauth_provider)
```

This would require that the AuthProvider plugins would need to be packaged and installed, which is maybe not as simple as you were hoping for.

MarcSkovMadsen:
FYI I'm not very knowledgable in this. But one of the next steps for me would be to find out how to have users securely sign on via LDAP, Azure AD or SSO. I'm able to do LDAP with django applications. The other methods I've not tried yet.

Hope the methods are or can be supported. Would be great if there was some kind of docs, examples or tutorials on this.

MarcSkovMadsen:
And in the end I would like to be able to build a single page application. But with different "pages" in (like awesome-panel.org has a gallery of apps). Some of these pages everybody can have access to. And others are restricted by AD Groups. I.e. there could be multiple AD groups giving access to different "pages" and data in the application.

MarcSkovMadsen:
I am so much looking forward to this.

And it’s a unique feature. No other Python Analytics framework provides this in the free, open source version I believe.

philippjfr:
@MarcSkovMadsen If you get a chance it would be fabulous if you could go through the new docs I've added here and do some testing. I'm going to merge this PR due to shorter term deadlines but this definitely needs some Q&A and actual unit tests before release.

Issue #845, opened 2019-12-04
Only first tab of Tabs re-renders datashader plot


timothydmorton:
#### ALL software version info (Panel, bokeh, python, notebook, OS, browser, any other relevant packages)

panel: latest master (panel 0.7.0.post7+g088e33c)
bokeh: 1.4.0
holoviews: 1.12.7
datashader: 0.8.0

#### Description of expected behavior and the observed behavior

I expect that all tabs should re-render the datashader plots as zoom/pan changes, but only the first tab does.

#### Complete, minimal, self-contained example code that reproduces the issue

```python
import holoviews as hv
import numpy as np
import panel as pn
import pandas as pd
from holoviews.operation.datashader import datashade
import datashader as ds

pn.extension()

N = 1000000
x = np.random.random(N)
y = np.random.random(N)

dset = hv.Dataset(pd.DataFrame({'x': x,
                              'y': y, 
                              'z1': x*y,
                              'z2': -x*y}))

pts1 = hv.Points(dset, kdims=['x', 'y'], vdims=['z1'])
pts2 = hv.Points(dset, kdims=['x', 'y'], vdims=['z2'])

opts = dict(height=800, width=800)
plot1 = datashade(pts1, aggregator=ds.mean('z1')).options(**opts)
plot2 = datashade(pts2, aggregator=ds.mean('z2')).options(**opts)

pn.Tabs(*[('z1', pn.panel(plot1)), ('z2', pn.panel(plot2))])
```

#### Screenshots or screencasts of the bug in action
Axes are linked properly, but only the first tab re-renders on zoom:
<img width="881" alt="Screenshot 2019-12-04 14 15 09" src="https://user-images.githubusercontent.com/1895387/70186350-95addf80-16a0-11ea-9573-03230559efc2.png">
<img width="892" alt="Screenshot 2019-12-04 14 15 15" src="https://user-images.githubusercontent.com/1895387/70186348-95addf80-16a0-11ea-96b5-250de32e8d43.png">

timothydmorton:
Update: seems to still be broken after #846 when tabs are overlays, e.g.:
```python
import holoviews as hv
import numpy as np
import panel as pn
import pandas as pd
from holoviews.operation.datashader import dynspread, datashade, rasterize, shade
from holoviews.operation import decimate
import datashader as ds

pn.extension()

N = 1000000
x = np.random.random(N)
y = np.random.random(N)

dset = hv.Dataset(pd.DataFrame({'x': x,
[...]

PR #846, opened 2019-12-04
Ensure HoloViews callbacks are reinitialized after linking

philippjfr:
Fixes https://github.com/holoviz/panel/issues/845


timothydmorton:
Thanks!

Issue #778, opened 2019-11-13
Different app reactivity between .show, .servable and bokeh serve 


fmaussion:
We have an app (https://github.com/pat-schmitt/glacier_simulator) which works as follows: when the user presses the "run the model" button, a relatively expensive numerical model runs in the background. In a loop, we report back at each model time step and use `.event()` to update the figure, so that the user sees the glacier grow. 

The expected behavior looks like this:

![servable](https://user-images.githubusercontent.com/10050469/68807546-490a5200-0668-11ea-9e27-61fce051b9f9.gif)

and **this is what we get when running the app locally with `.servable()` or (strangely enough), when running the app via mybinder using `panelserverextension.py`** (which uses `panel serve` internally?)
 
However, when we run the app locally with `.show()` or when we serve the app on google cloud using `bokeh serve`, the app now doesn't update during the simulation but updates all frames at the end of the simulation, looking like this:

![show](https://user-images.githubusercontent.com/10050469/68807922-08f79f00-0669-11ea-8610-2a3005a8ab2f.gif)

If you want to see the app running, just click "run the model" and you can compare them depending on how we serve them:
-  mybinder: https://mybinder.org/v2/gh/pat-schmitt/glacier_simulator/master?urlpath=/proxy/5006/simulator
- google cloud: https://bokeh.oggm.org/simulator/simulator

Obviously, we would expect our own server to be faster and slicker than mybinder, but because of this glitch this is not the case. Where do these differences come from? Why is `.show()` different than `.servable()` locally? Thanks for any hint!

cc @pat-schmitt who wrote the app and @TimoRoth who maintains the server

philippjfr:
@fmaussion Could you report all versions? I'm particularly interested in the version of param in this case.

fmaussion:
Sorry I don't have access to this info right now. @pat-schmitt and @TimoRoth will report tomorrow. 

Regarding `panel serve`: is this any different than `bokeh serve`, and would you recommend to use `panel serve` on our server?

TimoRoth:
You can run the app in the exact same fashion it's running on bokeh.oggm.org via the following:
```
docker run -e BOKEH_ALLOW_WS_ORIGIN=127.0.0.1 -p 8080:8080 oggm/bokeh:20191108 git+https://github.com/pat-schmitt/glacier_simulator.git simulator.ipynb
```

It's on 127.0.0.1:8080 afterwards.

pip freeze output for the packages in that container:

<details>
[...]

PR #857, opened 2019-12-08
Unlock Document and dispatch events if we have lock

philippjfr:
This PR ensures that when even on the server events are dispatched immediately. This is achieved by adding an `unlock` context manager which dispatches all ModelChangedEvents triggered in the context body to all sockets on current sessions. I have to do significantly more testing to figure out if this is actually safe.

Fixes https://github.com/holoviz/panel/issues/778


philippjfr:
Going to merge and make sure I test it extensively before release.

PR #867, opened 2019-12-15
Attempt optimizing layout solver

philippjfr:
Attempt at caching layout solver state for HTML model.

@MarcSkovMadsen would you mind trying this PR out? It should give a huge performance speedup for the layout solver but I don't yet know if it has unintended consequences. You'll have to check out the PR and then build the models with `pip install -e .` or `python setup.py develop`

### DIFF
--- a/panel/models/html.ts
+++ b/panel/models/html.ts
@@ -12,23 +12,37 @@ function htmlDecode(input: string): string | null {
 }
 
 export class CachedVariadicBox extends Layoutable {
-  _cache: (Sizeable | null)
+  _cache: {[key: string]: Size}
 
   constructor(readonly el: HTMLElement) {
     super()
-    this._cache = null;
+    this._cache = {};
   }
 
   protected _measure(viewport: Size): SizeHint {
-    if (this._cache)
-      return this._cache
+    const {width_policy, height_policy} = this.sizing
+
+    let width: number | null = null;
+    let height: number | null = null;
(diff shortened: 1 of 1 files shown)


philippjfr:
Going to merge this and test more before release.

MarcSkovMadsen:
Hi @philippjfr 

I've tested the current master branch and 

- All my pytests pass
- All my pages on the awesome-streamlit.org repo looks as expected in Chrome and Firefox
- The feel of the website is that its more responsive but I have not timed this.
- Firefox is still significanly faster than Chrome.
- There was a glitch. The first time I tested the new branch on master in Firefox it went into some sort of maximized, presentation mode. But F11 could not bring it back to the normal mode and it looked wird. Furthermore it did not maximize the width fully. There was a sort of margin on the left and right of a few pixels wide outside the Firefox window where I could see the background. I could not reproduce this though even though I tried a lot. I didn't take any screenshots because I thought I could do that when I reproduced the issue. My feeling is that this can be ignored. Maybe I hit a key or something. But if you hear this from others then I would get worried.

philippjfr:
Thanks for your testing, it is much appreciated!

>The feel of the website is that its more responsive but I have not timed this.

For me it's considerably more responsive but actual timing is difficult.

>But if you hear this from others then I would get worried.

I'll look out for this, I'll do some extensive testing before the next release to ensure there haven't been any regressions.

PR #875, opened 2019-12-17
Switch to holoviz.org domain

philippjfr:
Updating the DNS records now.


philippjfr:
This was already done.

Issue #878, opened 2019-12-18
Explore making reactive callbacks smarter at diffing contents


philippjfr:
Currently reactive callbacks (defined using pn.depends) make some basic attempts at diffing the contents by checking if the returned type is a Pane of the same type that supports and update call. However I think we can do even better than that by making it possible to diff Panels as well. This would make it possible to compose complex layouts inside a callback and avoid having to rerender the whole thing.

PR #879, opened 2019-12-18
Allow recursive reactive updates

philippjfr:
Implements #878 

Recursive diffing means you can write apps like this and instead of replacing the components in the returned layout each time it will simply update them:

```python
slider = pn.widgets.IntSlider(name='Slider', start=0, end=10)

@pn.depends(slider)
def example(value):
    return pn.Row(slider, pn.widgets.Checkbox(value=bool(value%2)), value)

pn.panel(example)
```

### DIFF
--- a/panel/pane/base.py
+++ b/panel/pane/base.py
@@ -21,8 +21,6 @@ from ..io import (
     init_doc, push, state, unlocked,
 )
 from ..layout.base import NamedListPanel, Panel, Row
-from ..io import push, state, unlocked
-from ..layout import Panel, Row
 from ..links import Link
 from ..models import ReactiveHTML as _BkReactiveHTML
 from ..reactive import Reactive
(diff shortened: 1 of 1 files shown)


dataist:
Seems a little faster. But I get a strange text artifact flicker now where I didn't before--like the entire dash reverts to a text representation for a half second. 

Attached comparison vids
[comparison_vids.zip](https://github.com/holoviz/panel/files/3976245/comparison_vids.zip)

philippjfr:
Yep, I just noticed pushing a fix in a minute.

philippjfr:
Actually your issue might be unrelated to the bug I just fixed but worth trying again anyway.

dataist:
that fixed the text flicker! However the kernel stays busy now 20x longer on an interaction than it used to

philippjfr:
>However the kernel stays busy now 20x longer on an interaction than it used to

Well that's not good. Is the example you've built something you can share?

dataist:
unfortunately not, following up via DM with other options

philippjfr:
It may be the case that I should batch updates. When we were replacing everything that meant that everything happened in one go but now that we are actually updating components each update triggers a separate update.

dataist:
it isnt clear what the kernel is actually up to since the processing and rerendering finishes in the same time as before just the kernel light stays up

dataist:
pulling up the dev console, the browser is really busy with "rendering" during that time even though there are no visual changes. Here are the last few seconds of the busy signal performance profiled: 
Screen Shot 2019-12-17 at 9 38 14 PM

philippjfr:
This ends up triggering a ton of layout events which make this very expensive. Bokeh should wait to apply a new layout until all the events have been applied but until that is implemented I'll put this on hold.

PR #887, opened 2019-12-18
Added FontAwesome icon and pane

philippjfr:
```python
modifiers = ['', 'fa-spin']
sizes = ['' 'fa-2x', 'fa-5x', 'fa-10x']
icons = ['fa-spinner', 'fa-cog', 'fa-heart', 'fa-clock', 'fa-check', 'fa-home', 'fa-volume-up']

pn.GridBox(*[pn.panel(' '.join([i, s, m]), margin=20) for m in modifiers for s in sizes for i in icons], ncols=10).show()

pn.Column(*((pn.pane.Markdown('# Buttons'),)+tuple(pn.widgets.Button(icon='%s fa-2x' % i) for i in icons))).show()
```

fa

Screen Shot 2019-12-18 at 8 07 47 PM

### DIFF
--- a/panel/_styles/widgets.css
+++ b/panel/_styles/widgets.css
@@ -139,3 +139,11 @@ progress:not([value])::before {
   from {background-position: 0%}
   to {background-position: 100%}
 }
+
+.fas {
+  vertical-align: bottom;
+}
+
+.only-icon {
+  margin-bottom: -1em !important;
+}
--- a/panel/config.py
+++ b/panel/config.py
@@ -26,6 +26,9 @@ from .io.state import state
 
 _PATH = os.path.abspath(os.path.dirname(__file__))
 _CSS_FILES = glob.glob(os.path.join(_PATH, '_styles', '*.css'))
+_CSS_FILES += [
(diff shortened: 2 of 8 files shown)


jbednar:
Cool! Is it possible to make the button resize to match the icon's size?  And could it be possible to make the button itself disappear as an option, leaving only the icon (but still clickable)?

philippjfr:
>Is it possible to make the button resize to match the icon's size?

That's already what it does.

>And could it be possible to make the button itself disappear as an option, leaving only the icon (but still clickable)?

Probably using some css tricks which I could expose as a parameter on buttons.

jbednar:
That's not what it's doing above (horizontally)...

philippjfr:
>That's not what it's doing above (horizontally)...

Oh, right, currently the widgets default to width=300, it's possible it'll do the right thing with width=None. Will give it a spin later.

philippjfr:
Going to close, you can easily add FontAwesome icons with ReactiveHTML now and this PR is beyond hope.

PR #888, opened 2019-12-18
Use optimized HTML model consistently

philippjfr:
So it turns out I hadn't actually used the optimized HTML model consistently throughout so this PR should provide further speedups but also potential to break things.

@MarcSkovMadsen would you mind trying this out again?

### DIFF
--- a/panel/pane/image.py
+++ b/panel/pane/image.py
@@ -12,7 +12,7 @@ from six import string_types
 
 import param
 
-from .markup import DivPaneBase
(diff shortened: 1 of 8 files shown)


philippjfr:
Cc: @dharhas and @timothydmorton any chance you could also give this a spin on the LSST dashboard?

MarcSkovMadsen:
Hi @philippjfr 

Tested with success in Chrome and Firefox on Windows on all pages of the awesome-panel.org code. 

Awesome!

Issue #471, opened 2019-06-04
Feature Request: Searchable Multi-Select Dropdown


sr2424:
I would like to ask for a new widget that behaves similar to the Dash dropdown or Select2 dropdown.

Dash Dropdown:  https://dash.plot.ly/dash-core-components/dropdown
Select2:  https://select2.org/

The CrossSelector widget is close to what I'm looking for, but it doesn't give the full functionality.  I would like to be able to search/filter the dropdown and select multiple entries.  I use the Dash version to select individual parameters that are a various temperature, pressure, speed, etc... telemetry with datetime, value pairings and quickly plot the parameter. However, I would prefer to use panel and have the ability to use hvplot and datashader for larger datasets.

jbednar:
Can you elaborate a bit?  The [CrossSelector](https://panel.pyviz.org/reference/widgets/CrossSelector.html#gallery-crossselector) seems like it already has filtering and selecting multiple entries.  Is searching (as for Autocompleteinput) the missing bit?

sr2424:
Sorry for the slow response.  Maybe it's an option/customization that I don't realize. If I have a list of options that is hundreds of items long, then scrolling through everything isn't possible. If each option has some built in structure, i.e. TX Apple Green ("State Fruit Color"), then you can filter using the text in order, but not out of order.  The other difficulty is that the filter, selects options, but doesn't filter the list of options.  If you are trying to select just one or two out of hundreds with an expected return on your filter being 10 or so, then it is difficult to accomplish this.  I am using this for selecting data parameters to chart.  The options list I have has location, hardware set, component, parameter/sensor type information in the option text.  This can all be broken down into multiple drop down boxes and linked, but I prefer the simplicity of a single box for searching, selecting, and tracking of the selections.

xavArtley:
Like here?
https://github.com/pyviz/panel/issues/392

sr2424:
The new multichoice widget in bokeh 2.0 adds this feature.  It looks like it will be available with 0.9 release.
https://github.com/bokeh/bokeh/pull/9546

PR #910, opened 2019-12-27
Added Pillbox widget

philippjfr:
Adds a `Pillbox` widget which is a more compact and more modern looking multi-select widget.

Addresses: https://github.com/bokeh/bokeh/issues/9532 and https://github.com/holoviz/panel/issues/471

Screen Shot 2019-12-27 at 7 21 22 PM

Cc: @MarcSkovMadsen

- [ ] Add tests
- [ ] Add reference entry


philippjfr:
Going to close in favor of adding this directly to bokeh and then exposing it in Panel.

Issue #565, opened 2019-07-31
working links in auto-generated SVGs


ltalirz:
I noticed that hyperlinks defined in SVGs using `xlink:href` don't work when the SVG is base64-encoded. Since panel encodes the SVG in base64, this means those links stop working.

Panel *does* provide a solution for cases where you want to link to an external file ([using `embed=False`](https://panel.pyviz.org/reference/panes/SVG.html)) but I believe it does not cover the case where you generate the SVG on the fly from an object with a `_repr_svg` method, e.g. a graph created using a library like [graphviz](https://pypi.org/project/graphviz/).

It might make sense to give users the option to disable encoding via base64 also when the link is not external. Anyhow, below is the hack I added to work around this in case other people hitting the same issue find this useful:
```python
import panel as pn

class PlainSVG(pn.pane.SVG):
    """Drop-in replacement for pn.pane.SVG that outputs plain SVG (not base64-encoded).
    
    Note: links encoded with xlink:href are not displayed in base64-encoded SVGs.
    """
    
    def _get_properties(self):
        p = super(PlainSVG, self)._get_properties()
        data = self._img()
        
        if self.object is None:
            return dict(p, text='<img></img>')
        if not isinstance(data, bytes):
            data = data.encode('utf-8')
        return dict(p, text=data)
```

Below an example to play around with.

```python
from graphviz import Digraph
g = Digraph('G')

g.attr(rankdir='TB')
g.attr("node",style='filled', fillcolor='white:gray', gradientangle='45')

g.node("Reference\npublication", shape="oval", href=link_test)g.node("GitHub", shape="oval", href=link_github)
g.node("Original\nstructure", shape="oval", href=link_test)
g.node("geo1",label="Geometrical\nproperties", shape="oval", href=link_test)
g.node("DFT optimization\n& DDEC charges", shape="box", href=link_test)
g.node("Optimized structure\n W/DDEC charges", shape="oval", href=link_test)
g.node("geo2",label="Geometrical\nproperties", shape="oval", href=link_test)
g.node("Adsorption calculation\nCO2", shape="box", href=link_test)
g.node("Adsorption calculation\nN2", shape="box", href=link_test)
g.node("Results CO2", shape="oval", href=link_test)
g.node("Results N2", shape="oval", href=link_test)
g.node("CCS process\nperformances", shape="oval", href=link_test)

g.edge("Reference\npublication",'GitHub')
g.edge('GitHub', 'Original\nstructure')
[...]

PR #913, opened 2019-12-29
Added encode parameter to SVG pane

philippjfr:
Fixes https://github.com/holoviz/panel/issues/565


examples/reference/panes/SVG.ipynb
```diff
@@ -21,7 +21,8 @@
     "\n",
     "For layout and styling related parameters see the [customization user guide](../../user_guide/Customization.ipynb).\n",
     "\n",
-    "* **``embed``** (boolean, default=False): If given a URL to an image this determines whether the image will be embedded as base64 or merely linked to.\n",
+    "* **``embed``** (boole, default=False): If given a URL to an image this determines whether the image will be embedded as base64 or merely linked to.\n",
+    "* **``encode``** (bool, default=True): Whether to base64 encode the SVG, when enabled SVG links may not work.\n",
```
jbednar: Why is it true by default if it lacks linking? Presumably encoding has some benefit, but what it is is not stated.

philippjfr: Just didn't change the default but I guess there's no compelling reason not to, base64 is slightly more efficient I guess but probably not worth it if certain features don't work.

jbednar: Ok, I vote to change the default then.

Issue #839, opened 2019-12-03
Downloadable DataFrame


apahl:
Hi, thanks for this great dashboarding framework.  
I have used it now for the first time to offer a PCA / t-SNE Dimension reduction app for one of our in-house data sets and it works great.

In addition to the PCA / t-SNE plot the tool also displays the calculated table as a panel.widgets.DataFrame, which also looks very nice.

But of course, the first request by my users was to have a link or button to download the table for further processing.

I unsuccessfully tried to implement a download link, is there any possibility to achieve something like this?  
Like, saving the dataframe as a csv, then offer a download link to it?

The app is served with `panel serve app_*.py` .

Many thanks again.

Kind regards,  
Axel

MarcSkovMadsen:
I would definately need this. 

What I really would like is to be able to expose url endpoints that when clicked starts the download of the file. That would enable the use to store the link an use it again and again in Python, Excel, etc. An alternative is just a button, that when clicks starts the download.

The files types I would expect to expose would be csv, xlsx, txt, parquet, json and maybe one day docx.

See also discussion on Gitter

![image](https://user-images.githubusercontent.com/42288570/70080026-21bef900-1606-11ea-82d8-b04981bdd97a.png)

ahuang11:
Here's code to download Ace content:
```
ace = pn.pane.Ace()
btn = pn.widgets.Button(name='Download')
btn.jscallback(args={'ace':ace}, clicks="""
filename = 'ace_content.txt'
var element = document.createElement('a');
element.setAttribute('href', 'data:text/plain;charset=utf-8,' + encodeURIComponent(ace.code));
element.setAttribute('download', filename);

element.style.display = 'none';
document.body.appendChild(element);

element.click();

document.body.removeChild(element);

""")
pn.Row(ace, btn)
```
credits to @xavArtley

MarcSkovMadsen:
The functionality I would like could be inspired by `pn.pane.img`. When using that Panel/ Bokeh exposes an url to the image. I would like to be able to do exactly the same with a file.

I'm just brainstorming. Maybe I need an `FileEndpointWidget` that has an `on_request` method where I get the request context like full url including parameters in my event handler. And I can the code the file response including filename. 

Or maybe I just need to be able to annote a `file_response_function` with a `@down_load_file` annotation.

apahl:
@MarcSkovMadsen : Your `data:file` example actually works for me. I did not know that was possible, thank you.  
[...]

Issue #680, opened 2019-10-07
Download functionality


coobas:
A feature that would enable an easy way to download an app (or its part) as html would be useful, especially when running on a server.

Example use case: Report generator. User provides inputs, clicks on a generate button, a generated report is diplayed in the app, optionally interacts with plots in the generated report, the report *can be downloaded by the user as an html file* by clicking on a download button.

So far it seems we could use the `save` method, which generates a local html file, and use some custom JS to serve this file to the user (e.g. something like https://stackoverflow.com/questions/49388511/send-file-from-server-to-client-on-bokeh).

philippjfr:
This seems like a great idea!

coobas:
I may have some time to help but I have no contribution to Panel / Bokeh (yet ;) but would welcome getting an experience) and minimum JS skills hence would need hints to contribute efficiently ;)

philippjfr:
We now have the `FileDownload` widget and you can define static directories to serve alongside your app.

PR #915, opened 2019-12-30
Add a FileDownload widget

philippjfr:
This adds a `FileDownload` which makes it possible to download some data which is set on the backend. By default the data is not transferred until a user clicks on a link which will start the transfer, once the data is transferred the widget will provide a link which actually downloads the data. I still want to improve the visual styling as currently it just uses bare text.

- Closes https://github.com/holoviz/panel/issues/839 and https://github.com/holoviz/panel/issues/680


MarcSkovMadsen:
I now have a use case for this. 

I have some CSV files that are 6MB. Chrome has a limit of 2MB for base64 encoded urls. So I cannot use that workaround.

maximlt:
For sure this will prove to be very useful!
Is it going to work for Office files (.xls, .xlsx, .doc, .docx, etc.)

philippjfr:
>Is it going to work for Office files (.xls, .xlsx, .doc, .docx, etc.)?

Should work fine yes.

apahl:
Very cool!

Thanks a lot.

PR #927, opened 2020-01-07
Correctly unpack vega datasets

philippjfr:


panel/models/vega.ts
```diff
@@ -55,6 +55,12 @@ export class VegaPlotView extends HTMLBoxView {
         this.model.data.data['values'] = datasets['data']
         delete datasets['data']
       }
+      for (const d of this.model.data.data) {
+        if (d.name in datasets) {
```
mattpap: Being pedantic, this should use `datasets.hasOwnProperty(d.name)`. This way, if `d.name` overlaps with an inherited member, but `d.name` isn't generated by `this._fetch_datasets()`, then you won't get a false positive in this clause. Alternatively `_fetch_datasets()` could return a plain object created with `Object.create(null)` instead of `{}`. Then you can use `in` operator freely.

Issue #943, opened 2020-01-14
Create a pn.Param reference example


MarcSkovMadsen:
# My Pain

I think that pn.Param is very important and does not get the attention or detailed explanation that it needs.

As a newbie to Panel I read quickly through the *getting started* and *user guide* and then started experimenting to learn. After that when I needed to understand Panel I wen't to the Reference Gallery as I thought this was the API/ Reference Documentation. It confused me a lot that I could not find pn.Param in there.

Also the example use case of `pn.Param` in the *User Guide - Parameters* section has not provided the level of detail for my use cases. So i've struggled a bit.

- I did not understand how to include only selected parameters via the `parameters` attribute.
- I did not understand how to change the `default_layout`.

I also think that `pn.Param` cannot really layout a complicated Parent-Child Child alone. You need to add in some more layout. That should be in the mix as well.

# Solution

Add a more detailed `pn.Param` example to the Reference Gallery. As it's not a panel, layout or widget (what is it? Why is it not a Pane?) I would suggest adding a section called Other.

philippjfr:
>Why is it not a Pane?

It is in fact a Pane, but a special one. That said I should probably move into the pane sub-package regardless.

PR #944, opened 2020-01-14
pn.Param reference notebook

MarcSkovMadsen:
This is a first version of a Param reference notebook that I have requested added to the Reference Gallery.

Please read the motivation here in the feature request  https://github.com/holoviz/panel/issues/943

image

This is only a first version and there is no link and image to this notebook as I could not find out where it was. I did not spend that much time on it since this is my first contribution and I hope to get a little bit of guidance in order to create something inline with the overall ideas and thoughts behing the Panel documentation. I would be very much prepared to give it a few iterations based on some feedback.

Please also note that this pull request includes a .gitignore file. It seems the repository does not have one? But I need it as VS Code etc. creates a .vscode folder. And I can also see other files change that I don't wan't to merge into master.

I hope you would provide some helpfull feedback and maybe it should not in the Reference Gallery but please consider it for the Example Gallery.


.gitignore
```diff
@@ -0,0 +1,7 @@
+.vscode/settings.json
+.vscode/panel.code-workspace
+__pycache__/dodo.cpython-37.pyc
+__pycache__/setup.cpython-37.pyc
+0.5.0
+panel/.version
```
xavArtley: Patterns which are specific to a particular repository but which do not need to be shared with other related repositories (e.g., auxiliary files that live inside the repository but are specific to one user’s workflow) should go into the $GIT_DIR/info/exclude file.
https://git-scm.com/docs/gitignore

philippjfr: I'm going to be adding a gitignore myself shortly so let's drop this one for now.

philippjfr:
Thanks for this! Seems like a major oversight that I didn't add this before.

MarcSkovMadsen:
#### Task List

What I **must have** is help on how to

1. Contribute an image and link to the https://panel.holoviz.org/reference/index.html page

I simply cannot find that when searching through the documentation.

It would also be nice to get 

2. some feedback from a second pair of eyes on the notebook it self and some feedback.

Then I will 

3. Contribute the changes including cleaning the Notebook.

philippjfr:
Could you clear the notebook first? That will make it easier for me to review it.

philippjfr:
>Contribute an image and link to the https://panel.holoviz.org/reference/index.html page

That currently requires a separate step from me to upload a thumbnail to an S3 bucket so we can do that after merging.

MarcSkovMadsen:
Hi @philippjfr . The notebook has been slightly improved and then cleared. 

Looking forward to some helpfull suggestions. 

But no hurry I can wait. :+1:

philippjfr:
Merging this as a first cut and will revise in the general documentation overhaul.

philippjfr:
@MarcSkovMadsen How do you feel about something like this:

```
pn.GridBox.subclass('GridBoxWithTwoColumns', ncols=2)
```

or similar, to replace this:

```
class GridBoxWithTwoColumns(pn.GridBox):
        super().__init__(*objects, **params, ncols=2)
```

MarcSkovMadsen:
@philippjfr. The ability to easily create a GridBox subclass would be great.

My questions are
- will an `nrows` parameter be supported?
- What purpose does specifying the 'GridBoxWithTwoColumns' serve?

philippjfr:
> will an nrows parameter be supported?

This would be a general utility which creates a subclass where the default values of the parameters would be overridden by the new values. The 'GridBoxWithTwoColumns' is the name of the new class, basically it's pretty much the the exact same as your code just dynamic.

philippjfr:
In fact if you know about Python internals you can do the same thing now using:

```python
type('GridBoxWithTwoColumns', (pn.GridBox,), {'ncols': 2})
```

Issue #936, opened 2020-01-10
Add json widget to Panel in order to display and navigate nested objects


MarcSkovMadsen:
#### My Pain

I'm trying to port the Yahoo Query App from awesome-streamlit.org to awesome-panel.org but there is no easy way in Panel to display and navigate the dict/ json response from the yahoo finance nicely.

![image](https://user-images.githubusercontent.com/42288570/72127995-ab828480-3371-11ea-9ce6-3bdc30c64e89.png)

In general I've experienced it very nice for internal and super user tools to be able to display dicts/ json objects and this functionality is not currently available in Panel.

#### Additional Context

- [react-json-view npm package](https://www.npmjs.com/package/react-json-view)
- [Streamlit json widget](https://docs.streamlit.io/api.html?highlight=json#streamlit.json)

I would be willing to try to implement and contribute this to Panel if I can get some guidance. But (disclaimer)

- I don't know where to start
- I don't know if I can implement this robustly

philippjfr:
Best candidate seems to be https://www.npmjs.com/package/json-tree-view

philippjfr:
Would be nice to use the react component but I'm not sure how feasible that is. In any case it seems clear this would have to be a custom bokeh model which uses an external library to perform the rendering. I'd be happy to guide you through that process and maybe write up a section in the developer guide for future users.

MarcSkovMadsen:
Thanks. That sounds like good ideas.

- How do I find a good component? What are the requirements?
- Where do I start? Is there a way to do some kind of a first, simple POC to learn how its done and then do the right implementation in a second step? 

I would like to contribute to the documentation as well. I'm a little bit passionate about this because the panes, layouts and widgets are what meets the users. It's their first impression.

philippjfr:
>How do I find a good component? What are the requirements?

Reasonably small size and small number of dependencies is one metric (which excludes the React components for example). Otherwise it just comes down to the functionality.

>Where do I start? Is there a way to do some kind of a first, simple POC to learn how its done and then do the right implementation in a second step?

I'd just start by copying an existing model in `panel/models/` and adapting it a little bit. In this case I'd recommend the following:

1. Make a copy of `panel/models/katex.ts` and name it `panel/models/json_tree.ts`
2. Perform a search and replace from `KaTeX` -> `JSONTree`.
3. Replace the contents of the `render` method with some placeholder that renders something into the `this.markup_el`
4. Import the model into `panel/models/index.ts`
[...]

PR #953, opened 2020-01-17
Add JSON model and Pane

philippjfr:
Implements https://github.com/holoviz/panel/issues/936

Screen Shot 2020-01-17 at 1 40 09 PM

- [x] Add tests
- [x] Add reference entry

### DIFF
--- a/panel/_styles/json.css
+++ b/panel/_styles/json.css
@@ -1,0 +1,194 @@
+.json-formatter-row {
+  font-family: monospace;
+}
+.json-formatter-row,
+.json-formatter-row a,
+.json-formatter-row a:hover {
+  color: black;
+  text-decoration: none;
+}
+.json-formatter-row .json-formatter-row {
+  margin-left: 1rem;
+}
+.json-formatter-row .json-formatter-children.json-formatter-empty {
+  opacity: 0.5;
+  margin-left: 1rem;
+}
+.json-formatter-row .json-formatter-children.json-formatter-empty:after {
+  display: none;
+}
+.json-formatter-row .json-formatter-children.json-formatter-empty.json-formatter-object:after {
+  content: "No properties";
+}
+.json-formatter-row .json-formatter-children.json-formatter-empty.json-formatter-array:after {
+  content: "[]";
+}
+.json-formatter-row .json-formatter-string,
+.json-formatter-row .json-formatter-stringifiable {
+  color: green;
+  white-space: pre;
+  word-wrap: break-word;
+}
+.json-formatter-row .json-formatter-number {
+  color: blue;
+}
+.json-formatter-row .json-formatter-boolean {
+  color: red;
+}
+.json-formatter-row .json-formatter-null {
+  color: #855A00;
+}
+.json-formatter-row .json-formatter-undefined {
+  color: #ca0b69;
+}
+.json-formatter-row .json-formatter-function {
+  color: #FF20ED;
+}
+.json-formatter-row .json-formatter-date {
+  background-color: rgba(0, 0, 0, 0.05);
+}
+.json-formatter-row .json-formatter-url {
+  text-decoration: underline;
+  color: blue;
+  cursor: pointer;
+}
+.json-formatter-row .json-formatter-bracket {
+  color: blue;
+}
+.json-formatter-row .json-formatter-key {
+  color: #00008B;
+  padding-right: 0.2rem;
+}
+.json-formatter-row .json-formatter-toggler-link {
+  cursor: pointer;
+}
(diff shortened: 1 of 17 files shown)


jbednar:
Cool!

philippjfr:
Okay, I also took this opportunity to factor out a `PanelMarkupView` which contains the layout optimizations and improved handling for sizing modes. The previous caching approach did not handle changes to the sizing mode well but I've now modified such that switching to a dynamic sizing mode dynamically works.

panel/models/json.ts
```diff
@@ -0,0 +1,59 @@
...
+    }
+    const config = {hoverPreviewEnabled: this.model.hover_preview, theme: this.model.theme}
+    const formatter = new JSONFormatter(json, this.model.depth, config)
+    const rendered = formatter.render()
+    let style = "border-radius: 5px; padding: 10px;";
+    if (this.model.theme == "dark")
+      rendered.style.cssText = "background-color: rgb(30, 30, 30);" + style;
+    else
+      rendered.style.cssText = style;
+	this.markup_el.append(rendered)
```
mattpap: Bad indentation or missing `{}`?

philippjfr: Bad indentation, thanks!

panel/models/declaration.d.ts
```diff
@@ -0,0 +1 @@
+declare module "json-formatter-js";
```
mattpap: This package has type information, so this shouldn't be needed.

philippjfr: Can't seem to find it:

```
panel/models/json.ts:3:27 - error TS7016: Could not find a declaration file for module 'json-formatter-js'. '/Users/philippjfr/development/pyviz_panels/panel/node_modules/json-formatter-js/dist/json-formatter.js' implicitly has an 'any' type.
```

panel/models/json.ts
```diff
@@ -0,0 +1,59 @@
...
+  }
+}
+
+export namespace JSON {
+  export type Attrs = p.AttrsOf<Props>
+  export type Props = Markup.Props & {
+    depth: p.Property<number>
+    hover_preview: p.Property<boolean> 
+    theme: p.Property<string>
```
mattpap: Seems there are only two choices, so `p.Property<"light" | "dark">` would be better.

panel/_styles/json.css
```diff
@@ -0,0 +1,194 @@
+.json-formatter-row {
```
mattpap: Unless this is intentionally added, it should be possible to include styles from an npm package.

philippjfr: What's the syntax for that?

panel/models/declaration.d.ts
mattpap: Scratch that, they messed up the configuration, so no type info. Also on a related note, looking closer at this library, it doesn't seem to be maintained anymore.

panel/models/json.ts
mattpap: I need this finish linting support in the extensions' compiler, then this kind of error can be more easily detected.

panel/models/declaration.d.ts
philippjfr: Since it doesn't have any upstream runtime dependencies I can live with it not being maintained. Do you think I **should** be concerned?

panel/models/json.ts
philippjfr: I also really need to set up emacs in a way that it doesn't use tabs for ts files...should have done that a long time ago.

panel/_styles/json.css
mattpap: That's a good question, as I didn't test this outside bokehjs. I will get back to you on this later (so perhaps in another PR, if this one progresses quickly).

panel/models/declaration.d.ts
mattpap: It depends, though (TL;DR) in this case no. Given that it doesn't have dependencies is a good thing and avoids security issues, patching up `package-lock`, etc. Type information can be defined locally, so that's not a big issue. So the only real issue is fixing any potential bug and/or adding new features. If you have to change it, you will have to fork it and continue development on your own. Given that this is JS, some things can be patched up without changing the source code. This is (I presume) a "leaf" feature, i.e. you aren't going to build on top of it, so it's a low risk. To the contrary, in bokeh we depend on canvas2svg, which also isn't maintained anymore and it's a big problem (essentially every new issue is a "won't fix").

panel/_styles/json.css
philippjfr: As far as I can tell they don't actually ship the CSS with the npm bundle anyway.

PR #954, opened 2020-01-17
Minify bundled CSS files during build

philippjfr:
If the csscompressor library is available this will minify the contents of the bundled css files at build time, and switches to using those files instead if they are available.


philippjfr:
@mattpap Do you have any suggestions for minifying CSS files at build time? Not loving the idea of bundling csscompressor, which I was forced to do since it's not available from a reasonable channel on conda.

PR #956, opened 2020-01-17
let FileInput widget set filename

### LINKED ISSUE #955: Fileinput widget always returns `None` for name of uploaded file
As of panel 0.7.0 (with bokeh 1.4.0), the Fileinput widget does not set the name of the uploaded file.

E.g. in the following code
```python
import panel as pn

inp = pn.widgets.FileInput(name='Upload')
btn = pn.widgets.Button(name='Upload', button_type='primary')
r= pn.Column(pn.Row(inp, btn),)

def on_click_parse(event):
    #print(inp.get_param_values())
    print(inp.filename)
btn.on_click(on_click_parse)

r.servable()
```
it always prints `None` to the terminal.

I can already see where the issue is and will open a PR.

ltalirz:
fixes #955 

The value of the `filename` property of the `FileInput` always remained
`None` (also after uploading a file).

I could add a test [here](https://github.com/holoviz/panel/blob/master/panel/tests/widgets/test_input.py#L63), if you let me know what the filename I should be testing for here.

### DIFF
--- a/panel/tests/widgets/test_input.py
+++ b/panel/tests/widgets/test_input.py
@@ -59,10 +59,11 @@ def test_file_input(document, comm):
 
     assert isinstance(widget, BkFileInput)
 
-    file_input._comm_change({'mime_type': 'text/plain', 'value': 'U29tZSB0ZXh0Cg=='})
+    file_input._comm_change({'mime_type': 'text/plain', 'value': 'U29tZSB0ZXh0Cg==', 'filename': 'testfile'})
     assert file_input.value == b'Some text\n'
     assert file_input.mime_type == 'text/plain'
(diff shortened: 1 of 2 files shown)


philippjfr:
This is great thank you!

>I could add a test here, if you let me know what the filename I should be testing for here.

Just do something like this:

```python
file_input._comm_change({'mime_type': 'text/plain', 'value': 'U29tZSB0ZXh0Cg==', 'filename': 'testfile'})

assert file_input.filename == 'testfile'
```

ltalirz:
Done!

philippjfr:
Looks like a syntax error snuck in there.

philippjfr:
Thanks again!

PR #958, opened 2020-01-18
Add .gitignore

philippjfr:


.gitignore
```diff
@@ -0,0 +1,121 @@
...
+# Distribution / packaging
+.Python
+env/
+build/
+develop-eggs/
+dist/
+downloads/
+eggs/
+.eggs/
+lib/
```
mattpap: My overall preference and advice is to avoid catch-all ignore files like this one. However, I understand the convenience. To future proof this `.gitignore`, and reduce future head scratching (why my files are seemingly gone), I will strongly recommend to use `/lib/` syntax. This limits the scope of such entry to the directory containing `.gitignore`. I chose `lib` in particular, because that's a good example of a name that can be both ignored and included, depending on its location in the tree.

philippjfr: Point well taken, I'll make another pass over this.

PR #978, opened 2020-01-22
Clarify developer installation instructions, and fix up some metadata.

ceball:
* Clarify developer installation instructions.
* Fix nodejs dependency (I don't _think_ it is required to run the tests...right? It is required for developers, but they are instructed to install the build extra, which includes nodejs.)

The changes to the developer guide should definitely be reviewed. Also, I've never created or edited a panel model (as far as I can recall!), so I'm sure someone could add something more useful...

Also, because I was in there:
  * Update license years and team name.
  * Update package metadata.
  * ~Removed `postBuild` (am I right it's unused?)~


.travis.yml
```diff
@@ -98,7 +98,6 @@ jobs:
         - source activate test-environment
         - doit develop_install $CHANS_DEV -o build && pip uninstall -y panel
         - doit pip_on_conda
-        - conda install $CHANS_DEV nodejs
```
ceball: The develop_install command above will install it (`-o build`).

LICENSE.txt
```diff
@@ -1,4 +1,4 @@
-Copyright (c) 2018, PyViz contributors (pyviz.org)
+Copyright (c) 2018-2020, HoloViz team (holoviz.org).
```
ceball: By team I mean you 😂

philippjfr: 😄

philippjfr:
The postBuild should have a note in it to mention that it's used for binder. It is indeed still needed for that reason.

philippjfr:
Thanks, looks good to me!

PR #979, opened 2020-01-22
CLI fixups: fix handling of no args; restore pyct commands (were being ignored).

ceball:
Aside: I'd like to add at least a basic test of the cli (e.g. as is done for holoviz; https://github.com/holoviz/holoviz/blob/master/tox.ini#L36), but I think that should be part of a larger effort to get better at testing from the user's point of view. (E.g. does conda install panel currently work? And then does `panel install-examples` or whatever work? And then do the examples run? Etc. This topic is on the pyctdev "roadmap".)

### DIFF
--- a/panel/cli.py
+++ b/panel/cli.py
@@ -67,21 +67,21 @@ def main(args=None):
     for cls in bokeh_commands:
         subs.add_parser(cls.name, help=cls.help)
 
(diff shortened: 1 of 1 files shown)


philippjfr:
Thanks, should have added tests while refactoring this.

Issue #987, opened 2020-01-22
VTK pane not playing well with PanelHTMLBoxView


xavArtley:
As an example we can notice vtkvolume panels are not well displayed in the developer documentation
https://pyviz-dev.github.io/panel/reference/panes/VTKVolume.html#panes-gallery-vtkvolume
I suppose it is due to the way the PanelHTMLBoxView compute the size of the panel content.
As the vtk panels are filled after the computation of the layout they do not behave like other panes.

xavArtley:
Sorry to reopen but I don't think the issue is solved:
```
import panel as pn
pn.extension('vtk')
dragon = pn.pane.VTK('https://raw.githubusercontent.com/Kitware/vtk-js/master/Data/StanfordDragon.vtkjs',
                     sizing_mode='stretch_both', height=400, enable_keybindings=True, orientation_widget=True,
                     serialize_on_instantiation=True)
dragon
```

result:
- notebook
![image](https://user-images.githubusercontent.com/18531147/72967312-5c047580-3dc1-11ea-8636-5002be48d9b1.png)
- served 
![image](https://user-images.githubusercontent.com/18531147/72967361-78081700-3dc1-11ea-84ed-9b13cd3f1ae3.png)

philippjfr:
I can't reproduce:

![Screen Shot 2020-01-23 at 11 58 48 AM](https://user-images.githubusercontent.com/1550771/72979056-ccb68c80-3dd7-11ea-98d0-50da8a766c76.png)

![Screen Shot 2020-01-23 at 11 57 38 AM](https://user-images.githubusercontent.com/1550771/72978938-96790d00-3dd7-11ea-861e-63c0eba4b467.png)

philippjfr:
Hmm, I see, you're using 'stretch_both', missed that sorry.

PR #988, opened 2020-01-22
Fixes for VTK layout

philippjfr:
Fixes https://github.com/holoviz/panel/issues/987


panel/models/vtk/vtk.ts
```diff
@@ -79,48 +79,43 @@ export class VTKPlotView extends PanelHTMLBoxView {
...
-      })
-      this._plot()
-      this._rendererEl.getRenderer().getActiveCamera().onModified(() => this._get_camera_state())
-      this._remove_default_key_binding()
-      this.model.renderer_el = this._rendererEl
-    }
-    super.after_layout()
+  render(): void {
+    super.render()
+    const container = div()
```
xavArtley: Not sure If the div is really necessay or if we can replace it by this.el and the vtk root container by this.root.el

philippjfr: I think in certain grid plots there might not be any actual guarantee `this.root.el` exists.

philippjfr: Generally other components also seem to add a subcontainer to `this.el` so I'm inclined to leave it.

xavArtley: ok

panel/models/vtk/vtk.ts
```diff
@@ -79,48 +79,42 @@ export class VTKPlotView extends PanelHTMLBoxView {
...
+      const dims = container.getBoundingClientRect()
+      const width = Math.floor(dims.width * window.devicePixelRatio)
+      const height = Math.floor(dims.height * window.devicePixelRatio)
+      axes_canvas.setAttribute('width', width.toFixed())
+      axes_canvas.setAttribute('height', height.toFixed())
+    })
+    this._plot()
+    this._rendererEl.getRenderer().getActiveCamera().onModified(() => this._get_camera_state())
+    this._remove_default_key_binding()
+    this.model.renderer_el = this._rendererEl
```
philippjfr: @xavArtley While I'm looking at this code, I'm assuming this is added to make JS linking possible in some form?

philippjfr: It's usually bad form to attach anything related to the view to a model but if it's about linking I get why you're doing this.

xavArtley: Indeed

philippjfr: Thanks, glad I have at least a rudimentary understanding of the VTK models having done this.

PR #992, opened 2020-01-23
Fixed usage of old param API in util

jlstevens:
Discovered while trying to update param API usage in https://github.com/holoviz/holoviews/pull/4201. There may be others so don't merge the PR just yet...


philippjfr:
Thanks.

PR #993, opened 2020-01-23
vtk layout proposition

xavArtley:
fix : https://github.com/holoviz/panel/issues/987


philippjfr:
Thanks for this, can confirm that it fixes #987 at least for the VTK model, the VTKVolume model on the other hand still seems to exhibit the issues. I did some minor refactoring.

philippjfr:
Can't figure that part out either but it doesn't seem to be a regression since 0.7.0 exhibits the same behavior for the VTKVolume pane.

PR #1017, opened 2020-01-27
Add progress bar to embed save code

Jacob-Barhak:
Remove hard limit of max_states when saving embeded files and turn it into a warning while adding a progress bar and enable stopping on keyboard interrupt to avoid long comuptations that cannot be stopped.

### DIFF
--- a/panel/io/embed.py
+++ b/panel/io/embed.py
@@ -277,9 +277,7 @@ def embed_state(panel, model, doc, max_states=1000, max_opts=3,
             w, m, _, g = values[i]
             try:
                 w.value = k
-            except (KeyboardInterrupt, SystemExit):
-                raise
-            except:
+            except Exception:
                 skip = True
                 break
             sub_dict = sub_dict[g(m)]
@@ -300,7 +298,7 @@ def embed_state(panel, model, doc, max_states=1000, max_opts=3,
     for (w, _, _, _), v in zip(values, restore):
         try:
             w.param.set_param(value=v)
-        except:
+        except Exception:
             pass
 
     if json:
--- a/panel/io/embed.py
+++ b/panel/io/embed.py
@@ -6,6 +6,7 @@ from __future__ import absolute_import, division, unicode_literals
 import os
 import json
 import uuid
+import param
 
 from collections import defaultdict
 from itertools import product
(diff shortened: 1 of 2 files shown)


panel/io/embed.py
```diff
@@ -258,22 +260,25 @@ def embed_state(panel, model, doc, max_states=1000, max_opts=3,
...
+    for key in tqdm(cross_product):
         sub_dict = state_dict
         skip = False
         for i, k in enumerate(key):
             w, m, _, g = values[i]
             try:
                 w.value = k
+            except (KeyboardInterrupt, SystemExit):
+                raise
             except:
```
mattpap: This should be replaced with simpler and idiomatic `except Exception`, which filers out system-type exceptions.

Jacob-Barhak: The keyboard expedition was added to allow Control C breaking a long process- it was tested to work. Without this code, one could not stop a long process running. Is Exception setup to handle those interrupts? 
If yes, I can test it.

mattpap: They way this works is that bare `except` is equivalent to `except BaseException`, which is a catch-all clause. System-type exceptions (interrupt, exit, etc.) derive directly from it. There is also `Exception` to cover all other (user) exceptions. See Python's [exception hierarchy](https://docs.python.org/3/library/exceptions.html#exception-hierarchy) for details. So, if you change this to `except Exception` is will catch only user exceptions and allow to pass-though system exceptions. In general it's a bad practice in Python to use bare `except`, because the side effect is that Ctrl+C, etc. don't work as expected. In fact panel uses those a lot (some 50 cases), so I will go ahead and fix that and set up linting to cover this issue.

Jacob-Barhak: Thanks Mateusz, 

You explained well this time - you meant that instead of isolating the keyboard error, just avoid using a bare except - this makes sense. I was looking at the code you highlighted rather than the code below. If I understand you correctly, you want the code to look like:

````

        for i, k in enumerate(key):
            w, m, _, g = values[i]
            try:
                w.value = k
            except Exception:
                skip = True
                break
````
I just fixed one problem, not the code everywhere as you are suggesting.

Hopefully we are on the same page now. 

It yes, will be happy to fix it.

PR #1018, opened 2020-01-27
Set up flake8 and clean up bare except and other errors

mattpap:
This adds a minimal setup derived from bokeh's `setup.cfg`. This PR is a side-effect of PR #1017.


philippjfr:
Thanks @mattpap!

Issue #957, opened 2020-01-18
Deck.GL Pane


philippjfr:
Deck.GL is a really nice library for displaying geographic 2.5 and 3D plots, it would be nice if we could support Deck.GL natively in Panel by exposing a Deck.GL pane.

PR #960, opened 2020-01-19
Add pydeck

MarcSkovMadsen:
Defines a PyDeck Pane which renders a PyDeck plot using a PyDeckPlot
bokeh model.

For now I've just experimented a bit.

There is a simple implementation of an app using PyDeck in the gallery at
[awesome-streamlit.org](https://awesome-streamlit.org).

image

See

- [pydeck implementation](https://github.com/MarcSkovMadsen/awesome-panel/blob/master/package/awesome_panel/express/pane/pydeck.py)
- [test_pydeck app](https://github.com/MarcSkovMadsen/awesome-panel/blob/master/src/pages/gallery/awesome_panel_express_tests/test_pydeck.py)
- [test_pydeck test](https://github.com/MarcSkovMadsen/awesome-panel/blob/master/package/tests/test_pydeck.py)

It's based on the following resources

- [PyDeck Github](https://github.com/uber/deck.gl/tree/master/bindings/pydeck)
- [PyDeck Docs](https://deckgl.readthedocs.io/en/latest/)
- [PyDeck Deck Implementation](https://github.com/uber/deck.gl/blob/master/bindings/pydeck/pydeck/bindings/deck.py)
- [PyDeck render_json_to_html Implementation](https://github.com/uber/deck.gl/blob/master/bindings/pydeck/pydeck/io/html.py)

When I move on I believe I can find inspiration in

- The Panel implementation of the Plotly Pane and Plotly Bokeh Model
- [Extending Bokeh](https://docs.bokeh.org/en/latest/docs/user_guide/extensions.html)
- [PyDeck Jupyter Implementation](https://github.com/uber/deck.gl/blob/master/bindings/pydeck/pydeck/widget/widget.py)

There is a feature request for this at [Github 957](https://github.com/holoviz/panel/issues/957)

PR #1019, opened 2020-01-27
Add DeckGL pane [doc-build]

philippjfr:
Supersedes #960 

Fixes https://github.com/holoviz/panel/issues/957


philippjfr:
@MarcSkovMadsen Thanks so much for getting this started! Getting this to work reliably in Jupyter notebooks and JupyterLab was a huge pain but I've worked out a solution. In JupyterLab (and on the server) we load the individual dependencies while in Jupyter notebooks we load the pydeck jupyter-widget bundle (as you were doing). The former is more efficient overall and really this is kind of hacky but I hope eventually we can have smaller bundles which consistently load the components using npm or get a better mechanism for loading the JS dependencies lazily. For now I'm very happy to have this in 0.8! This should be the last code PR to go in and I should be ready to release tomorrow.

MarcSkovMadsen:
Hi @philippjfr 

I'm so impressed. I could have spent a week on this. Don't know how you know how to do all these things :-)

One thing that might be interesting is that I have a more advanced example of using the DeckGL pane. I will create a new pull request to include that.

philippjfr:
>I'm so impressed. I could have spent a week on this. Don't know how you know how to do all these things :-)

Ha, don't put yourself down, you got things 90% there, I just did all the tedium of making things work across across different environments (honestly I wanted to pull my hair out on this) and updating it to the new build infrastructure.

>One thing that might be interesting is that I have a more advanced example of using the DeckGL pane. I will create a new pull request to include that.

Yes, we need to get the gallery more populated somehow. At some point soon I'll ask people to contribute their examples back to the gallery. A more complex DeckGL example would be hugely appreciated.

jbednar:
> A more complex DeckGL example would be hugely appreciated.

More complex examples should probably be part of examples.pyviz.org, not Panel itself, so that they can have their own separate pinned environment rather than necessarily being updated during every Panel release...

PR #1020, opened 2020-01-28
Merge two flake8 config files

ceball:
Not quite sure what happens when config is in multiple places - probably just gets read from one of them, so need to stick to one. I _think_ setup.cfg is a more expected location than tox.ini for such config (that was my opinion a few years ago, and still is).

(Also, I've dropped the "ignore run_test.py" bit as I think that bug was fixed some time ago.)


philippjfr:
Thanks, I did wonder about this when I merged the PR just now but thought I'd merge and deal with it in a subsequent PR. You beat me to it.

PR #1021, opened 2020-01-28
Add rerender option for HoloViews pane

philippjfr:
@Jacob-Barhak Want to try this out, basically just do this with your hv objects:

```python
pn.pane.HoloViews(holomap, rerender=True).save('test.html', embed=True)
```

### DIFF
--- a/panel/io/embed.py
+++ b/panel/io/embed.py
@@ -13,11 +13,10 @@ from itertools import product
 
 from bokeh.models import CustomJS
 from param.parameterized import Watcher
+from tqdm import tqdm
 
 from .model import add_to_doc, diff
 from .state import state
-from tqdm import tqdm
-
 
 #---------------------------------------------------------------------
 # Private API
@@ -266,6 +265,7 @@ def embed_state(panel, model, doc, max_states=1000, max_opts=3,
                            'increase the max_states specified in the function '
                            'to remove this warning' %
                            len(cross_product))
+
 
     nested_dict = lambda: defaultdict(nested_dict)
     state_dict = nested_dict()
--- a/panel/pane/holoviews.py
+++ b/panel/pane/holoviews.py
@@ -41,6 +41,9 @@ class HoloViews(PaneBase):
         Whether to use link the axes of bokeh plots inside this pane
         across a panel layout.""")
 
+    rerender = param.Boolean(default=False, doc="""
+        Whether to rerender the plot on each update.""")
+
     renderer = param.Parameter(default=None, doc="""
         Explicit renderer instance to use for rendering the HoloViews
         plot. Overrides the backend.""")
@@ -72,7 +75,7 @@ class HoloViews(PaneBase):
     _rename = {
         'backend': None, 'center': None, 'linked_axes': None,
         'renderer': None, 'widgets': None, 'widget_layout': None,
-        'widget_location': None, 'widget_type': None
+        'widget_location': None, 'widget_type': None, 'rerender': None
     }
 
     _rerender_params = ['object', 'backend']
@@ -165,7 +168,7 @@ class HoloViews(PaneBase):
             not self._initialized):
             self._update_layout()
 
-    def _update_plot(self, plot, pane):
+    def _update_plot(self, plot, pane, backup):
         from holoviews.core.util import cross_index, wrap_tuple_streams
 
         widgets = self.widget_box.objects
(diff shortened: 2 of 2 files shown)


Jacob-Barhak:
You will have to better explain what this option does, yet installing the new version from master and using this new rerender option only makes things worse - moving sliders seem not to work at all now at least in one case - I will run a few more tests to make sure, yet if there is any difference than before, it is only worse.
Hopefully the problem is on my side.

philippjfr:
It will almost certainly be slower and result in much larger files but I've not encountered a situation where it doesn't work.

philippjfr:
Just to explain both the issues you are having with HoloViews and what this workaround is meant to achieve. Embedding in Panel (and previously the HoloMap support in HoloViews) work by collecting all the widgets in an app that have a discrete set of values and then taking their cross product to come up with the state space. The embedding code then iterates over that state space and records the events that are generated as each change is applied. This works well for simple cases but **can** exhibit issues with more complex cases when combined with HoloViews. Specifically HoloViews tries to compute the minimal set of events to go from one state to the next. Imagine you have states `A`, `B` and `C`, the embedding code will go from `A->B` and record the events, and then go from `B->C` and record the events, however since it's computing minimal diffs the transition from `C->B` may not include all the events to fully restore the view to the original state `A`, because it has only recorded the diff to go from `A->B` and A and B may be more similar than `C` and `A`. This problem is particularly acute when we change column labels.

 The correct solution would be to ensure that during embedding HoloViews always includes all the events necessary to fully restore the state. This is a very difficult problem however, particularly while also still minimizing the size of the resulting data. This PR is a total hammer to solve this problem, instead of saying "let's try to compute a minimal diff between A and B" it says, let's just embed the entire plot for each state change. This will result in much bigger file sizes and trigger a full rerender of the plot but it also guarantees all the state changes are recorded in full.

Jacob-Barhak:
This is a very good explanation, thank you. Yet I have a few questions.
Does state A represent a certain value in Widget 1 or does it represent a
set of values in all widgets?

If a state represents a combination of widget values, then have you
considered creating a null state that you always pass through that resets
everything before moving to the desired end state? This may reduce the
combination numbers - yet again, this is only if I understand the term
state correctly.

However, the problem I am experiencing may be holoviews related or both
holoviews and panel related.

The bottom line is that I can not generate using the newer versions a plot
that works like [this one](
https://jacob-barhak.github.io/PosterIMAG2019Resources/PopulationPlotActive.html
)

Somehow changing combo boxes / sliders fail to do one of the following:
- update the plot image
- update the title properly
- update the axes

And there may be multiple issues here.

Your approach of embedding all states seems nice, since I do not care about
file size, yet this approach does not work for me - the plot just does not
change anything.

Any working solution is welcome, and I am happy to invest time and help
resolve it if you point out the relevant files and I can comprehend them.

I can start by creating a simple reproducible example it you think it will
help.

philippjfr:
>Does state A represent a certain value in Widget 1 or does it represent a
set of values in all widgets?

It represents a set of values across all widgets.

>If a state represents a combination of widget values, then have you
considered creating a null state that you always pass through that resets
everything before moving to the desired end state? This may reduce the
combination numbers - yet again, this is only if I understand the term
state correctly.

This may be a very good idea, can't say offhand if it'll work but it's definitely worth investigating

>Your approach of embedding all states seems nice, since I do not care about
file size, yet this approach does not work for me - the plot just does not
change anything.

Do you have a smaller example I could try?

Jacob-Barhak:
If I understand you correctly now, it seems that your Render solution is
probably similar to what I meant by passing through a null state -
eventually it ends up as code replicated as many times as you have states.

I will send you a test example that will show some of the problems by
tomorrow. - I need to cut down the code to reach a small enough example and
it will take some time. Until then many thanks for trying to resolve it
quickly.

philippjfr:
>If I understand you correctly now, it seems that your Render solution is
probably similar to what I meant by passing through a null state

It's similar but in fact a null state may be better overall. Instead of embedding a new plot each time the null state could iterate over all the plots and reset all properties to their defaults. So the basic layout and structure of plots stays the same but everything else about them is reset.

Jacob-Barhak:
And also, the number of transitions should be reduced by square root, if I
understand this correctly.

If one has N combined states, then there are O(N^2) number of transitions
among those states if I understood the current implementation correctly.

If one uses a null state then there are O(N) transitions between states.

If I am correct, this may be a partial reason for the huge static files. So
adding a Null state, will probably reduce the file.size as well.

I do hope that you have support in double buffering to avoid flickering
graphics - yet I may misunderstand something, so  let us do it properly
with some example that currently does not work and make it work. I hope I
can provide one by tomorrow.

PR #1022, opened 2020-01-28
VTK 3D multi slice 

xavArtley:
PR to implement a multi slice of 3d data volumes
In the same time I refactor some parts of VTK models


xavArtley:
New volume control exposed 
image

xavArtley:
I merged the 3d slicer with the volume pan
ezgif com-video-to-gif (10)

xavArtley:
I exposed the interpolation type, now we have real boxes when using nearest interpolation
image

xavArtley:
Last thing to do is to update the reference notebook

philippjfr:
Thanks so much @xavArtley! I think the docs are largely fine for now and we can expand on them later.

PR #1024, opened 2020-01-28
Add support for setting debug=False

philippjfr:


jbednar:
Is "debug" really the correct name for this concept? Seems more like "stdout" or "output", to me.

philippjfr:
Output seems meaningless and stdout is obscure (and technically doesn't cover exceptions).

jbednar:
Well, debugging is only one of the many things this feature controls; it literally controls all output to stdout and stderr, right?  I agree stdout is problematic.  text_output? console_output?  messages?  "debug" really doesn't convey to me that it's about console output; it could just as easily be about breakpoints, logging to a file, or a host of other concepts that don't mean "what happens when I do `print()` in Python" or what happens to exceptions.

PR #1029, opened 2020-01-29
Implemented __signature__ for Viewable objects

philippjfr:
Replaces the `param` docstring signature with a ``__signature__`` attribute, which should allow for tab-completion in newer versions of IPython. The new signature is applied on extension load as long as `pn.config.apply_signature` is set.

Additionally this enforces strict rules on what `Pane`, `Widget`, and `Panel` component signatures must look like during testing. This ensures no new components are added without considering how the signature code affects them.

### DIFF
--- a/panel/config.py
+++ b/panel/config.py
@@ -57,7 +57,7 @@ class _config(param.Parameterized):
 
     apply_signatures = param.Boolean(default=True, doc="""
         Whether to set custom Signature which allows tab-completion
-        in some IDEs and environements.""")
+        in some IDEs and environments.""")
 
     css_files = param.List(default=_CSS_FILES, doc="""
         External CSS files to load.""")
@@ -318,10 +318,10 @@ class panel_extension(_pyviz_extension):
                 prefix = cls.__doc__.split('\n')[0]
                 cls.__doc__ = cls.__doc__.replace(prefix, '')
             sig = inspect.signature(cls.__init__)
-            sig_params = list(sig.parameters)
-            if not sig_params or 'params' != sig_params[-1]:
+            sig_params = list(sig.parameters.values())
+            if not sig_params or sig_params[-1] != Parameter('params', Parameter.VAR_KEYWORD):
                 continue
-            parameters = list(sig.parameters.values())[:-1]
+            parameters = sig_params[:-1]
 
             processed_kws, keyword_groups = set(), []
             for cls in reversed(cls.mro()):
@@ -336,7 +336,7 @@ class panel_extension(_pyviz_extension):
             parameters += [
                 Parameter(name, Parameter.KEYWORD_ONLY)
                 for kws in reversed(keyword_groups) for name in kws
-                if name not in sig_params
+                if name not in sig.parameters
             ]
             parameters.append(Parameter('kwargs', Parameter.VAR_KEYWORD))
             cls.__init__.__signature__ = Signature(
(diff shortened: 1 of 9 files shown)


panel/config.py
```diff
@@ -54,6 +55,10 @@ class _config(param.Parameterized):
         os.environ['PANEL_EMBED'] = 'True'
     """
 
+    apply_signatures = param.Boolean(default=True, doc="""
+        Whether to set custom Signature which allows tab-completion
+        in some IDEs and environements.""")
```
jlstevens: Typo: environements

panel/config.py
```diff
@@ -294,13 +300,50 @@ def __call__(self, *args, **params):
...
         panel_extension._loaded = True
 
 
+    def _apply_signatures(self):
+        from inspect import Parameter, Signature
+        from .viewable import Viewable
+
+        descendants = param.concrete_descendents(Viewable)
+        for cls in reversed(list(descendants.values())):
+            if cls.__doc__.startswith('params'):
```
jlstevens: Is this to remove the docstring hint? If so I *believe* you can just disable this behavior at the param level with a flag somewhere (though it is possible that I am misremembering)

panel/config.py
```diff
@@ -294,13 +300,50 @@ def __call__(self, *args, **params):
...
+                prefix = cls.__doc__.split('\n')[0]
+                cls.__doc__ = cls.__doc__.replace(prefix, '')
+            sig = inspect.signature(cls.__init__)
+            sig_params = list(sig.parameters)
+            if not sig_params or 'params' != sig_params[-1]:
+                continue
+            parameters = list(sig.parameters.values())[:-1]
```
jlstevens: `[:-1]` is to get rid of `**kwargs``? Are you sure it is always present?

philippjfr: Yes, that's what this asserts:

```
if not sig_params or 'params' != sig_params[-1]:
    continue
```

panel/config.py
philippjfr: >If so I believe you can just disable this behavior at the param level with a flag somewhere

There is a flag but it would have to be disabled on import.

jlstevens: Right. You could set the flag as soon as param is imported.

panel/config.py
jlstevens: I would check the `kind` to be safe i.e make sure it is `VAR_KEYWORDS`.

panel/config.py
philippjfr: I think it's problematic, because if any other param based library gets imported in the process of importing panel then all those classes are affected too (don't think that does happen but I may very well be wrong).

panel/config.py
philippjfr: Sure, better to be explicit that said I also have tests asserting that all Viewable subclasses (except one) define `params` and that they are all `VAR_KEYWORDS`.

philippjfr: Those tests don't cover user defined Viewables so it's definitely a good idea.

panel/config.py
jlstevens: I don't think that would matter too much as those docstring hints are no longer supported by IPython and we'll be dropping Python 2 soon enough (and assuming Python versions that support ``__signature__``). Still just a suggestion and up to you!

philippjfr: Will leave it for now.

PR #1031, opened 2020-01-29
Ensure JS dependencies are skipped if already satisfied

philippjfr:
This adds a mechanism by which the dynamic loading code can check if a Javascript component has already been loaded and will then avoid loading it again. This avoids issues in JupyterLab where loading the same module twice can have side-effects, specifically the DeckGL dependency has issues when loaded twice as it seems to unload the JSON loader dependency causing certain examples two fail if two JupyterLab tabs are both running DeckGL examples.

### DIFF
--- a/panel/models/deckgl.py
+++ b/panel/models/deckgl.py
@@ -28,7 +28,7 @@ class DeckGLPlot(HTMLBox):
                       "https://cdn.jsdelivr.net/npm/mapbox-gl@1.6.1",
     ]
 
-    __js_skip__ = {'deck': __javascript__[:-1], 'mapboxgl': __javascript__[:-1:]}
+    __js_skip__ = {'deck': __javascript__[:-1], 'mapboxgl': __javascript__[-1:]}
 
     __js_require__ = {
         'paths': OrderedDict([
@@ -44,4 +44,3 @@ class DeckGLPlot(HTMLBox):
 
     height = Override(default=400)
     width = Override(default=600)
-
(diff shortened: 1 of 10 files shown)


panel/models/deckgl.py
```diff
@@ -22,10 +22,13 @@ class DeckGLPlot(HTMLBox):
...
                       "https://cdn.jsdelivr.net/npm/@deck.gl/json@8.1.0-alpha.1/dist/dist.dev.js",
-                      "https://cdn.jsdelivr.net/npm/mapbox-gl@1.6.1",
                       "https://cdn.jsdelivr.net/npm/@loaders.gl/csv@2.0.2/dist/dist.min.js",
                       "https://cdn.jsdelivr.net/npm/@loaders.gl/json@2.0.2/dist/dist.min.js",
-                      "https://cdn.jsdelivr.net/npm/@loaders.gl/3d-tiles@2.0.2/dist/dist.min.js"]
+                      "https://cdn.jsdelivr.net/npm/@loaders.gl/3d-tiles@2.0.2/dist/dist.min.js",
+                      "https://cdn.jsdelivr.net/npm/mapbox-gl@1.6.1",
+    ]
+
+    __js_skip__ = {'deck': __javascript__[:-1], 'mapboxgl': __javascript__[:-1:]}
```
philippjfr:
```suggestion
    __js_skip__ = {'deck': __javascript__[:-1], 'mapboxgl': __javascript__[-1:]}
```

PR #1052, opened 2020-01-31
os separator correction in test file selector

xavArtley:
Correction of `test_file_selector.py` which failed on windows due to  the os separator


philippjfr:
Thanks we need to get a windows CI set up again.

PR #1054, opened 2020-01-31
Improve VTK tests and coverage

xavArtley:


philippjfr:
Looks like there's a py2 issue:

```
        for renderer in renderers:
            for renProp in renderer.GetViewProps():
>               if not renProp.GetVisibility() or not isinstance(renProp, vtk.vtkActor):
E               AttributeError: 'module' object has no attribute 'vtkActor'
```

Sorry about still having to deal with py2 issues, excited to drop it soon.

xavArtley:
It will be hard for me to debug since there is no vtk binaries for python 2 for windows
I can found binaries but only for vtk-7.1

xavArtley:
when using `import vtk` in python 2.7 result in `panel.pane.vtk.vtk`

philippjfr:
Thanks again, you rock!

PR #1055, opened 2020-01-31
Error handling in Javascript Linking

philippjfr:


panel/links.py
```diff
@@ -394,7 +396,11 @@ class JSLinkCallbackGenerator(JSCallbackGenerator):
       console.log('WARNING: Could not set {tgt_attr} on target, raised error: ' + err);
       return;
     }}
-    target['{tgt_attr}'] = value;
+    try {{
+      target['{tgt_attr}'] = value;
+    }} catch(err) {{
+      console.log(err)
+    }}
```
mattpap: I'm genuinely wondering what errors caused this and other changes, because unless `target` is undefined, I don't see much that can happen here. Possibly overriding a read-only property would throw, but that should have been filtered out at the validation stage.

philippjfr: Depends on the component, certain Panel components are not yet sufficiently typed to detect all issues, e.g. the DeckGL model will throw all kinds of errors if you give it an invalid JSON spec.

PR #1056, opened 2020-01-31
[DOC] remove pane affectation to display it

xavArtley:
No display because affected to `vol`
image


philippjfr:
Damn, I thought I had removed that. Too late for 0.8 now.

PR #1064, opened 2020-02-03
bump param dependency

### LINKED ISSUE #1063: outdated param dependency
it seems panel 0.8 uses `CalendarDateRange` from param. This [was introduced in param 1.9.2](https://github.com/holoviz/param/releases/tag/v1.9.2), but the param dependency is still at >=1.9.0
https://github.com/holoviz/panel/blob/master/setup.py#L93

This can lead to errors like 
```
    param.CalendarDateRange: DateRangeSlider,
AttributeError: module 'param' has no attribute 'CalendarDateRange'
```
when upgrading to panel 0.8.0.
Will make a simple PR to fix this

ltalirz:
fix #1063

panel 0.8.0 uses CalendarDateRange from param, which was introduced in
param 1.9.2

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -142,7 +142,7 @@ extras_require['all'] = sorted(set(sum(extras_require.values(), [])))
 # non-python dependencies). Note that setup_requires isn't used
 # because it doesn't work well with pip.
 extras_require['build'] = [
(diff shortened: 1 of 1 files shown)


philippjfr:
Thanks!

PR #1065, opened 2020-02-03
Fix param dependency update from #1064.

ceball:
I guess this is what was intended (i.e. increasing the run time version rather than the build time one).

The build dep version could presumably be lower, but it probably doesn't matter too much.


philippjfr:
Thanks, shouldn't review PRs on my phone I suppose.

PR #1068, opened 2020-02-03
Enable orientation widget and camera control for vtkvolume

xavArtley:
Orientation widget and camera control can be usefull for vtk volume plot.


xavArtley:
PR #1073 is built on top of this one so don't need to keep this one open

PR #1070, opened 2020-02-04
WIP: Allow VTK Pane to partially update

tjcorona:


xavArtley:
It will be hard to review.
I check out your branch but first it does not compile when running `python setup.py develop`
However with slight modifcation I succeed to compile.
Next when I tried to run an example I had this error:
```
---> 26     partitioned_exporter = vtk.vtkJSONRenderWindowExporter()
     27     partitioned_archiver = vtk.vtkPartitionedArchiver()
     28     partitioned_exporter.SetArchiver(partitioned_archiver)

AttributeError: module 'vtkmodules.all' has no attribute 'vtkJSONRenderWindowExporter'
```
It lets me suppose you work with a custom build of vtk
(my installation is the following one :
```
Name: VTK
Version: 8.2.0
Summary: The Visualization Toolkit
Home-page: http://www.vtk.org/
Author: None
Author-email: None
License: BSD
Location: d:\python\python37-64\pyenv\py37\lib\site-packages
Requires: 
Required-by: vtki, pyvista
```)

**Comment by tjcorona** [comment]
Ah, true. I am working off of VTK's master branch, which has `vtkJSONRenderWindowExporter` in it. Sorry I didn't mention that earlier.

VTK is due to cut a release in the next week or so, and with it we will distribute new python packages for VTK. The easiest path is probably to wait until this happens before trying to build.

**Comment by xavArtley** [comment]
@tjcorona hello, I compiled vtk 9 however I have now the vtkJSONRenderWindowExporter
however I have not the vtkPartitionedArchiver

**Comment by tjcorona** [comment]
> @tjcorona hello, I compiled vtk 9 I have now the vtkJSONRenderWindowExporter
> however I have not the vtkPartitionedArchiver is there something I have not selected in cmake
> Capture d’écran de 2020-03-26 17-17-43

To build vtkPartitionedArchiver, you must set the flag `VTK_MODULE_ENABLE_VTK_CommonArchive` to `YES`. Please note, this adds libarchive as a dependency.

**Comment by xavArtley** [comment]
hello @tjcorona ,
I'm currently playing with your PR
I put your modifications on top of panel master in a new branch (xav/python-link-interactivity)
I made some modification which should not change the current behavior
I rename your model in `VTKSynchronizedPlot`

For the moment I play directly with the bokeh model (`panel.models.vtk.VTKSynchronizedPlot`)

However I notice several problems.
I made a .gif to show a first problem:
vtk_sync_camera

On the left you have the result with the `VTKSynchronizedPlot`

 and on the right the result with the current panel model.
As you can see the camera parameters are not taken into account and the second time we display the scene the camera is different from the first display

I give you the code I used in the notebook:
```python
import vtk
import panel as pn
import numpy as np
import pyvista as pv
from panel.models.vtk import VTKSynchronizedPlot
import base64, ctypes, io, json, os, sys, zipfile
pn.extension('vtk')

base64encode = lambda x: base64.b64encode(x).decode('utf-8')


def render_window_serializer(render_window):
    """
    Function to convert a vtk render window in a vtk-js scene.
    """
    render_window.OffScreenRenderingOn()
    render_window.Render()

    partitioned_exporter = vtk.vtkJSONRenderWindowExporter()
    partitioned_archiver = vtk.vtkPartitionedArchiver()
    partitioned_exporter.SetArchiver(partitioned_archiver)
    partitioned_exporter.SetRenderWindow(render_window)
    partitioned_exporter.Write()

    arrays = {}
    scene = None
    
    for i in range(partitioned_archiver.GetNumberOfBuffers()):
        buffer_name = partitioned_archiver.GetBufferName(i)
        ptr = partitioned_archiver.GetBufferAddress(buffer_name)
        address = int(ptr[1:-7], 16)
        ArrayType = ctypes.c_byte*partitioned_archiver.GetBufferSize(buffer_name)
        b = ArrayType.from_address(address)
    
        stream = io.BytesIO(b)
    
        filename = os.path.split(buffer_name)[-1]
        if len(filename) != 32:
            with zipfile.ZipFile(stream) as zf:
                encoded_scene = zf.read(zf.infolist()[0])
                scene = encoded_scene.decode("utf-8")
        else:
            if not filename in arrays:
                arrays[filename] = base64encode(stream.read())

    return (scene, arrays)

cylinder = vtk.vtkCylinderSource()
cylinder.SetResolution(8)
cylinderMapper = vtk.vtkPolyDataMapper()
cylinderMapper.SetInputConnection(cylinder.GetOutputPort())
cylinderActor = vtk.vtkActor()
cylinderActor.SetMapper(cylinderMapper)
cylinderActor.RotateX(30.0)
cylinderActor.RotateY(-45.0)
ren = vtk.vtkRenderer()
renWin = vtk.vtkRenderWindow()
renWin.AddRenderer(ren)
ren.AddActor(cylinderActor)

```

```python
scene, arrays = render_window_serializer(renWin)
pn.Row(VTKSynchronizedPlot(scene=scene, arrays=arrays), renWin)
```

```python
cylinder.SetResolution(3)
scene, arrays = render_window_serializer(renWin)
pn.Row(VTKSynchronizedPlot(scene=scene, arrays=arrays), renWin)
```

xavArtley:
I tried to create a new SynchronizedContext for each vtk panel
howver the vtkCamera can be construct only one time if it's `id` does not change:
https://github.com/Kitware/vtk-js/blob/26c9cd04a7674b9ca3c03b127e11a472e722cd82/Sources/Rendering/Misc/SynchronizableRenderWindow/vtkObjectManager.js#L409-L412

The camera `id` in a vtkjs global variable.
A workaround is to call `vtk.Rendering.Misc.vtkSynchronizableRenderWindow.vtkObjectManager.clearOneTimeUpdaters()`

tjcorona:
@jourdain

xavArtley:
vtkActor does not seems to take into acount the orientation property

jourdain:
Your workaround is not bad, but ideally we should provide another setup and with the current public API inside vtkObjectManager, it is tricky to reset to the genericUpdater.

I'll update vtk.js to enable a simpler call. Something like below

```
const vtkCamera = vtk.Rendering.Core.vtkCamera';
vtk.Rendering.Misc.vtkSynchronizableRenderWindow.vtkObjectManager.setTypeMapping('vtkCamera', vtkCamera.newInstance);
vtk.Rendering.Misc.vtkSynchronizableRenderWindow.vtkObjectManager.applyDefaultAliases();
```

For the colorBy issue, that is strange but it might be related to the fact that you are using SetScalars and let the mapper fallback to pick it for its colorBy rather than providing the array selection in the mapper.

Regarding the actor orientation, it is possible that we are skipping that property (not sure) @tjcorona.

xavArtley:
For the scalar color I remember I had to make some changes to the historical vtkjs serializer to make it works in most cases.

xavArtley:
@jourdain , @tjcorona another problem I notice is the support of textures:
texture_pb

xavArtley:
Since Lights are not deserialized on javascript side :
https://github.com/Kitware/vtk-js/blob/26c9cd04a7674b9ca3c03b127e11a472e722cd82/Sources/Rendering/Misc/SynchronizableRenderWindow/vtkObjectManager.js#L467-L470

we should force renderer AutomaticLightCreation to 1
Capture d’écran de 2020-04-01 17-13-53

jourdain:
There is already the automatic light creation. The issue that you are seeing is similar to the Active type of arrays Scalars/TextureCoordinates/Normals.

jourdain:
You are just providing several ways to highlight the same bug.

jourdain:
@tjcorona  we should talk so I can tell you what might be missing in the VTK serializer.

tjcorona:
> I put your modifications on top of panel master in a new branch (xav/python-link-interactivity)

@xavArtley I cannot find this branch on your fork. Can you please provide a link?

xavArtley:
it's on panel repository not on my fork

jourdain:
vtk.js 13.12.6 should be good to go…

tjcorona:
> vtk.js 13.12.6 should be good to go…

Pending https://gitlab.kitware.com/vtk/vtk/-/merge_requests/6675

xavArtley:
Hello @tjcorona, as your PR insterests me a lot
Hello, I worked a bit more on the Synchronized panel.
I pushed my current changes on the same branch
Since I'm back on windows and don't know how to compile VTK with libarchive, I have reimplemented the python serializer https://gitlab.kitware.com/vtk/vtk/-/blob/master/Web/Python/vtkmodules/web/render_window_serializer.py (with some slight modifications)
Here is some examples of current possibilities:
ezgif com-video-to-gif

ezgif com-video-to-gif (1)

image

tjcorona:
> Here is some examples of current possibilities:

Cool! Once vtk 9.0.0 is released (any day now, I promise!) I will push to have its python wheels be built with libarchive enabled. Your updated serialization routine would be nice as a fallback for vtk implementations without libarchive, though.

If you are ok with it, I would like your branch to replace mine as the content of this MR. If you prefer, you could post a MR with your branch and I would happily close this one in favor of yours.

tjcorona:
> Since I'm back on windows and don't know how to compile VTK with libarchive, I have reimplemented the python serializer

Another, potentially lighter-weight option could be to subclass `vtkArchiver` in Python and reimplement `vtkPartitionedArchiver` and its compression routines entirely in Python. I have experience subclassing C++ classes in Python to be useable in C++ using [pybind11](https://pybind11.readthedocs.io/en/stable/), so it is definitely doable. This would provide the additional benefit of confining the temperamental logic of serializing a VTK scene into vtk-js to one place.

jourdain:
@xavArtley could you share your changes on the python serializer so it could eventually replace the one available in VTK. Even better if you could make a PR so I can review it and make sure the changes we've done with TJ on the C++ side are actually managed (Which is probably what you changed).

Thanks!

xavArtley:
@jourdain  The serializer is use is here:
 https://github.com/holoviz/panel/blob/xav/python-link-interactivity/panel/pane/vtk/synchronizable_serializer.py

I still have a problem with the lookup table

jourdain:
The color issue might be related to the actual class used that is not defined here:
https://github.com/holoviz/panel/blob/xav/python-link-interactivity/panel/pane/vtk/synchronizable_serializer.py#L195-L196

Thanks for sharing... I'll see if I can port back some of your changes into VTK proper. ;-)

xavArtley:
On javascript side I had implented this but there was still some issues:
```
vtkns.SynchronizableRenderWindow
       .vtkObjectManager
       .setTypeMapping('vtkLookupTable', vtkns.LookupTable.newInstance, lookupTableUpdater)

}

function lookupTableUpdater(instance: any, state: any, context: any) {
  if(state.properties.table){
    state.properties.table = vtkns.DataArray.newInstance(state.properties.table);
  }
  vtkns.SynchronizableRenderWindow.vtkObjectManager.genericUpdater(instance, state, context);
}
```

xavArtley:
@tjcorona we can use my branch, however I don't think you have the access

tjcorona:
> @tjcorona we can use my branch, however I don't think you have the access

I am happy to abandon this MR in favor of yours if that's easier

xavArtley:
@philippjfr is it possible to give permissions to @tjcorona to push on the repo?

philippjfr:
Done.

xavArtley:
Thx

jourdain:
For the vtkLookupTable, I think you want to create `vtk.Rendering.Core.vtkColorTransferFunction` not the basic one that use the HueRange.

PR #1071, opened 2020-02-04
Retrieve vtk slices on python side

xavArtley:
This PR allow to get on the python side the slices computed by vtk.js in a vtkvolume
```python
import panel as pn
pn.extension('vtk')
import holoviews as hv
hv.extension('bokeh')
import numpy as np

data_matrix = np.zeros([75, 75, 75], dtype=np.uint8)
data_matrix[0:35, 0:35, 0:35] = 50
data_matrix[25:55, 25:55, 25:55] = 100
data_matrix[45:74, 45:74, 45:74] = 150

vol = pn.pane.VTKVolume(data_matrix, sizing_mode='stretch_width', height=400, spacing=(3,2,1),
                        interpolation='nearest', edge_gradient=0, sampling=0, display_slices=True,
                        display_volume=False, rescale=True, orientation_widget=True)
btn = pn.widgets.Button(name='debug')
btn.jscallback(args={'vtkpan':vol}, clicks="debugger")

@pn.depends(vol.param.slices_data_i, vol.param.slices_data_j, vol.param.slices_data_k)
def plot_rgb(data_i, data_j, data_k):
    if data_i and data_j and data_k:
        image_i = hv.RGB(np.array(list(data_i['values'].values()), dtype=np.uint8).reshape(75,75,4)[:,:,:3])
        image_j = hv.RGB(np.array(list(data_j['values'].values()), dtype=np.uint8).reshape(75,75,4)[:,:,:3])
        image_k = hv.RGB(np.array(list(data_k['values'].values()), dtype=np.uint8).reshape(75,75,4)[:,:,:3])
        return image_i + image_j + image_k
    else:
        return hv.RGB([]) + hv.RGB([]) + hv.RGB([])

pn.Column(pn.Row(pn.Column(btn, vol.controls(jslink=False, parameters=['slice_i', 'slice_j', 'slice_k'])), vol), plot_rgb)
```

ezgif com-video-to-gif

However I need help concerning the way to communicate through bokeh api

### DIFF
--- a/panel/models/vtk.py
+++ b/panel/models/vtk.py
@@ -4,10 +4,36 @@ Defines custom VTKPlot bokeh model to render VTK objects.
 """
 from bokeh.core.properties import (String, Bool, Dict, Any, Override,
                                    Instance, Int, Float, PositiveInt, Enum)
+from bokeh.core.has_props import abstract
 from bokeh.core.enums import enumeration
 from bokeh.models import HTMLBox, Model
 
 vtk_cdn = "https://unpkg.com/vtk.js"
+
+@abstract
+class AbstractVTKPlot(HTMLBox):
+    """
+    Abstract Bokeh model for vtk plots that wraps around a vtk-js library and
+    renders it inside a Bokeh plot.
+    """
+
+    __javascript__ = [vtk_cdn]
+
+    __js_skip__ = {'vtk': [vtk_cdn]}
+
+    __js_require__ = {
+        "paths": {"vtk": vtk_cdn[:-3]},
+        "exports": {"vtk": None},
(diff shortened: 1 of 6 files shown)


philippjfr:
So this data is basically the colormapped RGB of the input array? My main concern here is only communicating this data if requested to avoid sending it all of the time.

xavArtley:
> So this data is basically the colormapped RGB of the input array? My main concern here is only communicating this data if requested to avoid sending it all of the time.

You're right, I found more efficient to retrive colormap information on python side. I'll open an other PR with this solution

ezgif com-video-to-gif (1)

Issue #1089, opened 2020-02-11
Show command with threaded = True raise Error in panel 0.8.0


xavArtley:
Hello,
This code:
```
import panel as pn
pn.extension()

pn.panel(pn.widgets.Button(name='test')).show(threaded=True)
```
raise the following error.
![image](https://user-images.githubusercontent.com/18531147/74242493-53c39a00-4cde-11ea-8379-0c75744baea4.png)

I have not this error with panel 0.7.0

PR #1090, opened 2020-02-11
Fix threaded server

xavArtley:
fix: https://github.com/holoviz/panel/issues/1089


philippjfr:
Need to figure out what's going on with the py2 env but thanks for the fix. Ideally we'd put this under test.

PR #1092, opened 2020-02-12
Swap HTML with DIV

bstadlbauer:
Do this not only for `.png` export but for all others aswell (`.html`) to allow workflows that use `phantomjs` to render a `pdf` from the created `html`

### DIFF
--- a/panel/io/save.py
+++ b/panel/io/save.py
@@ -49,7 +49,7 @@ def save_png(model, filename):
 
 
 @contextmanager
-def swap_html_model(swap):
+def swap_html_model():
     """
     Temporary fix to swap HTML model for Div model during png export
     to avoid issues with DOMParser compatibility in PhantomJS.
@@ -64,7 +64,7 @@ def swap_html_model(swap):
     for viewable in param.concrete_descendents(Viewable).values():
         model = getattr(viewable, '_bokeh_model', None)
         try:
-            swap_model = issubclass(model, HTML) and swap
+            swap_model = issubclass(model, HTML)
             assert swap_model
         except Exception:
             continue
@@ -130,7 +130,7 @@ def save(panel, filename, title=None, resources=None, template=None,
     doc = Document()
     comm = Comm()
     with config.set(embed=embed):
-        with swap_html_model(as_png):
+        with swap_html_model():
             model = panel.get_root(doc, comm)
         if embed:
             embed_state(


bstadlbauer:
Solves #1093

philippjfr:
Don't have any particular objections here since the reason HTML was used was to make the work of the layout solver easier when building dynamic GUIs. That said this was only meant to be a temporary workaround until Bokeh drops PhantomJS support (in Bokeh 2.0 due out this month). Could you walk me through your workflow of generating PDFs? I'd prefer not to keep this hack around for too long.

bstadlbauer:
So at the moment we create a layout of a few plots, tables and text elements using rows and columns, then save the layout to `html` and use the following javascript snippet with `phantomjs` to create an A4 sized report.
<details>
    <summary>JS snippet</summary>

```js
"use strict";
var page = require('webpage').create(),
    system = require('system'),
    address, output, size, pageWidth, pageHeight;

if (system.args.length < 3 || system.args.length > 5) {
    console.log('Usage: filename [paperwidth*paperheight|paperformat] [zoom]');
    console.log('  paper (pdf output) examples: "5in*7.5in", "10cm*20cm", "A4", "Letter"');
    console.log('  image (png/jpg output) examples: "1920px" entire page, window width 1920px');
    console.log('                                   "800px*600px" window, clipped to 800x600');
    phantom.exit(1);
} else {
    address = system.args[1];
    output = system.args[2];
    page.viewportSize = { width: 600, height: 600 };
    if (system.args.length > 3 && system.args[2].substr(-4) === ".pdf") {
        size = system.args[3].split('*');
        page.paperSize = size.length === 2 ? { width: size[0], height: size[1], margin: '0px' }
                                           : { format: system.args[3], orientation: 'portrait', margin: '1cm' };
    } else if (system.args.length > 3 && system.args[3].substr(-2) === "px") {
        size = system.args[3].split('*');
        if (size.length === 2) {
            pageWidth = parseInt(size[0], 10);
            pageHeight = parseInt(size[1], 10);
            page.viewportSize = { width: pageWidth, height: pageHeight };
            page.clipRect = { top: 0, left: 0, width: pageWidth, height: pageHeight };
        } else {
            console.log("size:", system.args[3]);
            pageWidth = parseInt(system.args[3], 10);
            pageHeight = parseInt(pageWidth * 3/4, 10); // it's as good an assumption as any
            console.log ("pageHeight:",pageHeight);
            page.viewportSize = { width: pageWidth, height: pageHeight };
        }
    }
    if (system.args.length > 4) {
        page.zoomFactor = system.args[4];
    }
    page.open(address, function (status) {
        if (status !== 'success') {
            console.log('Unable to load the address!');
            phantom.exit(1);
        } else {
            window.setTimeout(function () {
                page.render(output);
                phantom.exit();
            }, 200);
        }
    });
}
```
</details>

However, this is not ideal anyway as we have to make sure that all elements have the correct size, etc. to fit on the page and correct pagebreaks are quite next to impossible. That said, we found no better options (just rendering out the plots as `png` and then creating a report in `latex` is not feasible, as we already have a lot of styling in CSS). If you have a better idea though, I am happy for suggestions!

philippjfr:
Have you played around with Chrome headless? I haven't tried it but something like this might be a solution: https://www.npmjs.com/package/chrome-headless-render-pdf

bstadlbauer:
Yes I actually tried and we have a production system using it (regular headleass chrome, not this specific extension), and the html to pdf problems are similar (pagebreaks/correct formatting). 

The main reason behind `phantomjs` was because I was used to it from `bokeh` (as it is/was the default renderer) - but as the support is running out I will try recreating the same results with headless chrome (tomorrow), so if it works the PR might be obsolete (and I got it in my own fork as quickfix anyway)

philippjfr:
I'm happy to merge for now but as I said this will get removed in 0.9, which will only support bokeh>=2.0. Good luck getting the chrome based solution working!

Issue #1102, opened 2020-02-16
Panel widgets with unit of measurement


1081:
There is no unambiguous approach how to work with different unit of measurement in one panel application. As a workaround, the symbols can be added in "brackets" after the widget name/description. For example:

_Float Slider name [kg/s]: 1.23 
Another Float Slider name [kg/s]: 1.24 
Yet another Float Slider name [kg/s]: 3.2_ 

But this results in a cluttered user interface. Although the use of brackets is not the standard-conform way to write si-unit based symbols!

A clear representation could look like this:
`float_slider = pn.widgets.FloatSlider(name=‘Float Slider’, value=1.23, symbol=‘kg/s’)
`
![example screenshot](https://discourse.holoviz.org/uploads/default/original/1X/38b6768bc7f983faafb0db332adf609fdf91ac50.png)

The symbols and (widget value) are aligned to the right.
The result would be an uncluttered and clear unser interface with the possibility to add units of measurement.

PR #1103, opened 2020-02-16
add suffix parameter to sliders

xavArtley:
adress https://github.com/holoviz/panel/issues/1102
Discussion for the possilbiltes of suffix styles (position emphasis...) are opened
And I wonder if this enhancement should not belong to bokeh directly
Peek 16-02-2020 13-18


philippjfr:
Would be okay merging this for now, but I agree this is better added to bokeh.

mattpap:
Maybe https://github.com/bokeh/bokeh/pull/9625 already does what's requested?

PR #1106, opened 2020-02-18
fix viewable show keyword arguments

gmoutso:
keyword arguments show and title are now passed to the get_server.

### DIFF
--- a/panel/viewable.py
+++ b/panel/viewable.py
@@ -244,7 +244,7 @@ class ServableMixin(object):
         return self
 
     def show(self, port=0, websocket_origin=None, threaded=False,
-             title=None, verbose=True, show=True, **kwargs):
+             title=None, verbose=True, open=True, **kwargs):
         """
         Starts a Bokeh server and displays the Viewable in a new tab.
 
@@ -264,7 +264,7 @@ class ServableMixin(object):
           A string title to give the Document (if served as an app)
         verbose: boolean (optional, default=True)
           Whether to print the address and port
-        show : boolean (optional, default=True)
+        open : boolean (optional, default=True)
           Whether to open the server in a new browser tab
 
         Returns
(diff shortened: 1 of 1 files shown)


panel/viewable.py
```diff
@@ -244,7 +244,7 @@ def servable(self, title=None):
         return self
 
     def show(self, port=0, websocket_origin=None, threaded=False,
-             title=None, verbose=True, **kwargs):
+             title=None, verbose=True, show=False, **kwargs):
```
jbednar: I don't see the `show` argument documented in the `show()` method docstring, which seems important given how unintuitive `show(show=False)` is. :-)

gmoutso: Added the docstring and changed the default to the previous behaviour.

gmoutso: I copied the docstring from `server.py`. Should kwarg show be popup or open?

panel/viewable.py
```diff
@@ -264,6 +264,8 @@ def show(self, port=0, websocket_origin=None, threaded=False,
           A string title to give the Document (if served as an app)
         verbose: boolean (optional, default=True)
           Whether to print the address and port
+        show : boolean (optional, default=True)
+          Whether to open the server in a new browser tab on start
```
jbednar: Thanks!  I guess this is called "show" because it does what `bokeh serve --show` does.  But it's still very confusing to see `.show(show=False)`.  Is there any enthusiasm for renaming `show` here to `open` or `launch` or `browse`?

panel/viewable.py
jbednar: Oops; responded in different thread, proposing `open` or `launch` or `browse`.  Probably should let @philippjfr decide.

philippjfr: No strong preference but I vote for `open` and I'd probably only rename it here.

gmoutso: `show(show=True)` does look somewhat absurd (but I really want you to show...). I think the votes are in.

gmoutso: Or even more strange to request `show(show=False)`

gmoutso: I will push `open` but feel free to revert (`open` has the disadvantage of being a python command as well).

philippjfr:
Looks good thanks, those py2 CI issues will be resolved separately.

PR #1112, opened 2020-02-21
[PYVISTA] switch to active_scalar until 0.9

xavArtley:
I made the choice to keep using the deprecated property until we stop py2 support


philippjfr:
Thanks @xavArtley.

PR #1122, opened 2020-02-28
WebComponent pane to dramatically speed up creating new widgets in Panel

MarcSkovMadsen:
I believe I have created a proof of concept for how to use web components in panel to speed up the time to market for creating new widgets.

If you run something like

```bash
python -m panel serve 'panel\components\wired\buttons.py' --dev --show
```

Then you can explore the below

It's not all fully working yet - but the principles have been identified.

web_component_poc


philippjfr:
This is amazing. Just for some background for me. Is there some automated way we could generate a set of components for some library including dynamic generation of parameters? For example if the underlying library ships some JSON spec of the components we could use that to auto-generate the components.

MarcSkovMadsen:
I've added 3 reviewers. Mostly because I think this is very important functionality (Otherwise please tell me why not) and if you have the time to review it a bit to improve and learn it was bit so awesome.

This is work in progress and will be improved dramatically. I will let you now when it's ready for review. 

 @philippjfr , @jbednar , @xavArtley

philippjfr:
As another maybe simpler approach instead of actually generating panel components from some JSON spec it might be nice to at least parse the HTML and make the attributes of the component accessible as a dictionary.

xavArtley:
Using python setup.py develop raise typescript compiler errors
Capture d’écran de 2020-02-28 21-26-33

philippjfr:
Yeah the PR is still using the old deprecated approach to building extensions.

xavArtley:
Was wondering if it was on purpose to not incrase size of the bundle

philippjfr:
No, I think @MarcSkovMadsen is still learning how to develop extensions and we still don't have a decent developer guide for it.

xavArtley:
@MarcSkovMadsen 
I allowed myself to make some changes so that the extension compiles

panel/components/wired/buttons.py
```diff
@@ -0,0 +1,48 @@
...
+        super().__init__(min_height=min_height, **params)
+
+class Video(WebComponent):
+    html = param.String("""<wired-video autoplay="" playsinline="" muted="" loop="" style="height: 100%; margin: 0px" src="https://file-examples.com/wp-content/uploads/2017/04/file_example_MP4_480_1_5MG.mp4" class="wired-rendered"></wired-video>""")
+
+if __name__.startswith("bk"):
+    import panel as pn
+    pn.config.sizing_mode="stretch_width"
+
+    js = """
```
philippjfr: Would be nicer to append these urls to `pn.config.js_files`.

MarcSkovMadsen:
Thanks for all your comments and improvements above. I have briefly read them but I wan't to comment how I expect to improve/ finalize the WebComponent concept.

The overall idea/ goal is to allow the user to inherit the WebComponent and

- add custom parameters to the child class
    - At a minimum parameters of type `String`, `Number` and `Boolean` should be supported
    - if a `clicks= param.Number()` attribute is added it will keep track of the number of clicks
- specify in an `attributes_to_sync =  param.Class()` parameter (or similar) that keeps track of the attributes/ properties of the python ChildWebComponent and the html/ js web component to sync.
    - The user can specify a `list` if there is a 1-1 correspondance between all python ChildWebComponent attributes and all the html/ js web component attributes/ properties to sync.
    - The user can specify a `dict` if there is not a 1-1 correspondance between all python ChildWebComponent attributes and all the html/ js web component attributes/ properties to sync.
  - If a `param.DataFrame` attribute is specified the data will be transfered effectively back and forth.
    - An example could be a ChildWebComponent attribute `table: param.DataFrame` that would be made available for use on the js side as a variable table that is used like `<my-html-tag table="[[table]]"></my-html-tag>" or something similar. (I will have to experiment a bit to understandhow this can be implemented).
    - Automatically parse any new `html` value received from the js side by a function called `parse_html` or similar. This function will parse the new `html` value and set the attributes of the ChildWebComponent.
    - The user can override the `parse_html` function if they need more than simple 1-1 parsing.

I hope this will all work out. 

The perspective is that it will be much easier to contribute new widgets to Panel

- Ordinary users can stick to simple html and python for creating or contributing new widgets.
- If you have access to a front end developer, the she does not need to learn of Bokeh or Panel. She can just convert an existing js, ts, react, vue or angular component into a web component or alternative develop the new component from scratch as a web component. The user can the integrate it into Panel them selves I believe.

panel/components/wired/buttons.py
MarcSkovMadsen: Yes. Eventually they should.

If you take a look the value of `js` you can see only one of the two imports can be done via `pn.config.js_files`. In one of the imports I need to specify the `type`.

```python
js = """
<script src="https://unpkg.com/@webcomponents/webcomponentsjs@2.0.0/webcomponents-loader.js"></script>
<script type="module" src="https://unpkg.com/wired-elements@0.6.4/dist/wired-elements.bundled.js"></script>
"""
```

In general the import of `js` web component libraries does not always work for me. It needs some experimentation before it works. Sometimes I need to 

- add `type="module"`
- add ``?module` to the url.
- add additional javascript libraries than what is described on npm or in the documentation.

(I believe)

So I guess there will be a request to improve the `pn.config.js_files` at some stage. At a minimum be able to specify the `type`.

MarcSkovMadsen:
> @MarcSkovMadsen
> I allowed myself to make some changes so that the extension compiles

@philippjfr and @xavArtley . Just to be sure. There is only

panel/components/wired/buttons.py
philippjfr: I was wondering about that. Will have to do some reading before making any concrete suggestions.

MarcSkovMadsen:
One more thing I would like to provide is a `js_files: param.Dict` (or similar) attribute where the user can specify which javascript files are need an any attributes like `type` that are needed.

```python
`js_files =[
    { "src": "https://unpkg.com/wired-elements@0.6.4/dist/wired-elements.bundled.js", 
      "type"= "module" }, 
}
```

The user still needs to remember to add them via `pn.config._js_files` or similar I belive.

MarcSkovMadsen:
@philippjfr 

After the model web_component.py file was modified in https://github.com/holoviz/panel/pull/1122/commits/bcff9aeb2f26f038c108e85ef76e979e670787aa I get the error 

image

How do I build and register the model before I run `panel serve`?

Before the the implementation used the `__implementation__` attribute and the model was automatically compiled.

```python
from bokeh.core import properties
from bokeh.models.layouts import HTMLBox	from bokeh.models.layouts import HTMLBox


RADIOBUTTON_TS_PATH = pathlib.Path(__file__).parent / "radio_button.ts"	WEBCOMPONENT_TS_PATH = pathlib.Path(__file__).parent / "web_component.ts"
RADIOBUTTON_TS_STR = str(RADIOBUTTON_TS_PATH.resolve())	WEBCOMPONENT_TS_STR = str(WEBCOMPONENT_TS_PATH.resolve())




class RadioButton(HTMLBox):	class WebComponent(HTMLBox):
    """A Predix App Navigation Widget	    """A Predix App Navigation Widget
    See https://www.predix-ui.com/#/elements/px-app-nav"""	    See https://www.predix-ui.com/#/elements/px-app-nav"""


    __implementation__ = RADIOBUTTON_TS_STR	    __implementation__ = WEBCOMPONENT_TS_STR


    html = properties.String('{"label":"Home","id":"home","icon":"px-fea:home"}')	    innerHTML = properties.String('')
```

philippjfr:
Did you compile the models with `panel build panel`?

MarcSkovMadsen:
Nope. I forget that was what I had to do.

Now it works. Thanks.

MarcSkovMadsen:
The manual test has been changed to `python -m panel serve 'panel\tests\components\widgets\test_wired.py' --dev --show`.

The synchronization of the Slider now works both ways. For some reason the ComboBox does not.

web_component_poc

panel/components/wired.py
```diff
@@ -0,0 +1,37 @@
+"""Implementation of the wired WebComponents"""
+import param
+
+from panel.pane import WebComponent
+
+
+class RadioButton(WebComponent):
+    """A Wired RadioButton"""
+    html = param.String('<wired-radio checked id="1" checked>Radio Two</wired-radio>')
```
xavArtley: for me ids should be unique in a document

xavArtley:
> The manual test has been changed to `python -m panel serve 'panel\tests\components\widgets\test_wired.py' --dev --show`.
> 
> The synchronization of the Slider now works both ways. For some reason the ComboBox does not.
> 
> The slider value only updates when I release the mouse click though.
> 
> web_component_poc

I think combo box does not work because the wired bundle used is too old

I succeed to make it appear correcly by using this code :
```python
import param
import panel as pn
js_files = {"wired": "https://wiredjs.com/dist/showcase.min.js"}
pn.extension(js_files=js_files)

from panel.components.wired import RadioButton, CheckBox, Slider, ComboBox

radio_button = RadioButton()
check_box = CheckBox()
slider = Slider()
combobox = ComboBox()
# video = Video(height=500)
pn.Column(
    radio_button, pn.Param(radio_button.param.html),
    check_box, pn.Param(check_box.param.html),
    slider, pn.Param(slider.param.html),
    combobox, pn.Param(combobox.param.html),
    # video, pn.Param(video.param.html),
    ).servable()
```
Capture d’écran de 2020-02-29 19-04-49

But it raises other problems for example for the slider

MarcSkovMadsen:
Synchronization from

- html parameter to all parameters in attributes_to_watch vice versa
- server to client and vice versa

should now be working.

I've added a lot of tests. The Param package, Panel and test driven development works so great. So nice to be able to write most tests in Python only.

I've also decided that parameter values of parameters in attributes_to_watch takes precedence over attributes of the html parameter value on construction. And I've implemented functionality to support this. To showcase this I've added two checkboxes in the example below.

I use the lxml to set and delete html attributes for now. I don't know if that would be ok to add to Panel as a requirement or if we should investigate some other option like custom code, regex or ..? @philippjfr ?

I have not looked into changing the js library yet (cf @xavArtley  comment above) and I've also noted that a param.Number needs to be rounded. This is not currently handled which is why the slider text turns red below.

web_component_poc

MarcSkovMadsen:
If investigated the `param.Number` problem above a bit. I can see that

- Both in Chrome (left below) and Firefox (right below) the slider value is shown round to 0 decimals
- In Firefox but not Chrome an error is indicated



image

I don't know where to start

- Is the bug in
    - `pn.pane.WebComponent`
        - Should I round number values when parsing from html attribute to parameter value?
    - `param.Number`
    - `pn.Param`
    - my showcase example.
        - Should I not use `param.Number` for the slider value?
        - Should I just customize the `pn.Param` widget used for slider value to have an appropriate number of decimals?
- param

Any suggestions @philippjfr ?

MarcSkovMadsen:
I've investigated the wired js package a bit. In different src's have different behaviour

#### <script type="module" src="https://unpkg.com/wired-elements@0.6.4/dist/wired-elements.bundled.js"></script>

- The DropDown does not work

#### <script src="https://unpkg.com/wired-elements@0.6.4/dist/wired-elements.bundled.js"></script>

- The DropDown does not work

i.e. the `type="module"` was not nescessary.

#### <script src="https://unpkg.com/wired-elements@1.0.0/dist/wired-elements.bundled.js"></script>

- Now the slider is fixed at 0. I can move it and the value changes but it's reset to 0.
    - I should investigate if this is something in our code

- Now the user can see and select the ComboBox values. But they are not shown in the "selected" attribute. Maybe it is because I taking the `selected` property and not the `selected` attribute. Yes. There is a difference. Best practice for web components is to keep them in sync. But they are seldom.

image

#### <script src="https://unpkg.com/wired-elements@2.0.0/dist/wired-elements.bundled.js"></script>

image

#### <script src="https://wiredjs.com/dist/showcase.min.js"></script>

- The user can see and select ComboBox values. But they are not shown in the "selected" attribute.
- The slider is fixed and cannot be moved at all

image

MarcSkovMadsen:
I can see that the <script src="https://unpkg.com/wired-elements@2.0.0/dist/wired-elements.bundled.js"></script> is because from 2.0.0 the there is no longer a bundle according to the wired github issues.

I will be using what @xavArtley described and what is working at https://wiredjs.com/showcase.html. 

js = """
<script src="https://unpkg.com/@webcomponents/webcomponentsjs@2.2.7/webcomponents-loader.js"></script>
<script src="https://wiredjs.com/dist/showcase.min.js"></script>
"""

I have tried setting attributes_to_watch={} for the slider and then I can move the slider. So I believe the problems are on our codes side. I have the same hypothesis for the ComboBox.

MarcSkovMadsen:
Arrgghhhh :-) Finally I realised it actually the wired-slider that does not work and not my code. See https://github.com/wiredjs/wired-elements/issues/121#issue-573516963

MarcSkovMadsen:
Having fun with the WebComponent pane. I'm trying out using the Google model-viewer web component and controlling the exposure using a parameter in Panel.

This is a reference example for WebComponents I've started creating.

web_component_poc

The example also reveals some of the weaknesses that has should be worked  on

- The src parameter cannot sync. My guess is it's because it has another name as a property.
- When changing the exposure using the slider you can see a glitch because it reloads and redraws everything. It's because I change the html of the model-viewer in the browser. We should think about this. Maybe we should not change it. Just change the `this.model.html` property instead.
- I would have liked a src ObjectSelector but I have not supported this yet.
- I would like to be able to control things like height and width of the model-viewer. I don't yet now how to.

MarcSkovMadsen:
I think there is a decision to be made implement.

a. HTML Webcomponent tag and attributes,
b. Javascript object and properties or
c. A mix both 

Right now I'm both.

Its really characterized by what I get and set in the .ts file. I'm

1. Adding the child element as the web component html tag via the `innerHTML` property.
2. Listening for changes to properties via `onchange` and getting the changed values
3.  Setting the attributes in the child element if it does not do this by it self.

The fundamental problem is really that the web components I've tried are conforming to best practices i.e. communicating via the html tag attributes. Many of them are changing their properties only and not syncing to the attributes. That's why I do 2. and 3.

I might do a pure web component implementation (a.) by using the `MutationObserver` ([link](https://stackoverflow.com/questions/41424989/javascript-listen-for-attribute-change)). Then only best practices web components would be supported.

I might a pure JavaScript object implementation (b.) I would then have to find out how to instantiate the class by it string name instead and then syncing .ts properties and python parameters via a dictionary.

I might stick with a mix if stop updating the attributes (3.) and do some other hacks/ workarounds I need to identify.

An alternative is also to implement pure versions of a. and b.

@philippjfr  and @xavArtley . Any ideas or preferences?

MarcSkovMadsen:
I suggest to keep a keep a mix but change to a clear distinction between web component attributes and js object properties.

I suggest doing this by adding a couple of parameters to the `WebComponent` pane and corresponding js model.

```python
properties_to_watch = param.Dict(doc="""
    The key is the name of the js property. The value is the name of the python parameter
    """
    )
    last_property_change = param.Dict(doc="""

    The key is the name of the property change. The value is the new value of the property
    """)
```

Referring to the above numbering I suggest

- Keeping 1. unchanged.
- Only doing 2. for `properties_to_watch`
- Implementing `MutationObserver` functionality for `attributes_to_watch`
- Dropping 3.

When I get the time I will try out this revised api.

As a side note I believe a JSClass pane could easily be developed from the ideas of the WebComponent. Instead of the html parameter there would be a `js_class` parameter (or similar) that contains information on how to initaliaze the js class.

If this is going to happen it might be an idea to rename and call the two JSPlugin and HTMLPlugin or something similar.

MarcSkovMadsen:
I can see that the problem with the `wired-combo` combobox is that it neither updates the `selection` attribute or runs the `onchange` function when the selection changes. Instead it fires a custom `selection` event.

I think it would be best practice for web components to behave as I expect so I've created a feature request here https://github.com/wiredjs/wired-elements/issues/122.

A workaround would be to extend the WebComponent api to be able to subscribe to custom events and then check if any of the properties_to_watch has changed.

MarcSkovMadsen:
I've added the `properties_to_watch` and `properties_last_changed` functionality to the WebComponent. Seems to work quite well

- wired radiobutton and checkboxes still work. The slider also works now.
- model-viewer now works without glitches

web_component_poc

web_component_poc

MarcSkovMadsen:
I've added functionality to watch events like click events. To test this I've also added a Wired Button

web_component_poc

I've implemented this using two new parameters on the pane

```python
events_to_watch = param.Dict(
        doc="""
    The key is the name of the js event. The value is the name of a python parameter to increment or None

    Use this if you want to support button clicks, mouseups etc.

    # Below is not implemented yet
    When the event is fired on the js side any changes to the html or properties_to_change will
    also be sent. Use this if you webcomponent don't fire the onchange event when your property changes
    """
    )
    events_count_last_change = param.Dict(
        doc="""
        Key is the name of the event, Value is the number of times it has fired in total
        """
    )
```

MarcSkovMadsen:
> This is amazing. Just for some background for me. Is there some automated way we could generate a set of components for some library including dynamic generation of parameters? For example if the underlying library ships some JSON spec of the components we could use that to auto-generate the components.

Maybe for some. Thats really worth investigating. 

But for setting up a new set of basic components like wired when the WebComponent api is done would probably take less than a day including testing and documentation.

And for setting up one web component with 5-15 attributes or parameters and some click events its so fast to configure that you wont notice.

I belive the above is true. Still to be tested though :-)

MarcSkovMadsen:
- Added functionality to check properties when events are triggered
- Added first iteration of wired notebook

web_component_poc_dropdown

web_component_poc_wired_ref

panel/models/web_component.ts
```diff
@@ -62,11 +64,32 @@ export class WebComponentView extends HTMLBoxView {
...
         this.model.eventsCountLastChange = eventsCountLastChanged;
+
+        this.checkIfPropertiesChanged()
+    }
+
+    checkIfPropertiesChanged(): void {
+        console.log("checkIfPropertiesChanged")
+        var propertiesChange: any = {};
+        for (let property in this.model.propertiesToWatch) {
+            var oldValue: any = this.propertyValues[property];
```
mattpap: If I may offer a piece of advice, don't use `var` unless you really know how it works. `var` has really weird semantics, especially if you're accustomed to lexical scoping (who isn't), and you may cause your self a lot of unnecessary debugging. Use `let` or `const` (preferably for immutable variables), they have normal semantics as one would expect. Let the compiler transpile this to code understood by all supported browsers, though these days `let` and `const` keywords are widely supported.

MarcSkovMadsen: Thanks. I will change. I don't use TypeScript very often. So I hoped someone like you would step in and guide me. 

Thanks. :-)

MarcSkovMadsen:
- All wired elements can now be disabled/ enabled
- Added an elevation parameter to selected wired elements
- Add Wired Calendar element

web_component_poc

MarcSkovMadsen:
Lots of bug fixes and nice improvements. Lots of new elements.

And support for nested properties !! I.e. since `webComponent.value` for the `wired-input` web component was not working I added support for nested properties, i.e. support for syncing `webComponent.input.value`.

web_component

panel/models/web_component.ts
MarcSkovMadsen: I've now changed var to let

MarcSkovMadsen:
Just a thought

One caveat of this methods is that I believe you cannot link/ embed them in in a static version. You need the server for communication between a WebComponent and something else.

The solution could be to create some code that could autogenerate the bokeh model .py and .ts files as well as the corresponding panel .py file from the WebComponent parameters.

MarcSkovMadsen:
I've added support for updating for example the inner part of the `html` parameter via a `parameters_to_watch` parameter and a `_get_html_from_parameters_to_watch` for a custom implementation.

An example is 

```python
class IconButton(WiredBase):
    html = param.String("<wired-icon-button><mwc-icon>favorite</mwc-icon><wired-icon-button>")
    parameters_to_watch = param.List(["icon"])

    icon = param.ObjectSelector("favorite", objects=MWC_ICONS, doc="""
    The name of an `mwc-icon <https://github.com/material-components/material-components-web-components/tree/master/packages/icon>`_
    """)
    def __init__(self, min_height=40, **params, ):
        super().__init__(min_height=min_height, **params)

    def _get_html_from_parameters_to_watch(self, **params) -> str:
        return f"<wired-icon-button><mwc-icon>{params['icon']}</mwc-icon></wired-icon-button>"
```

MarcSkovMadsen:
I've added a lot more functionality.

I've started testing If I can use the WebComponent for more than true web components. I can in fact. I've played around with the WebDataRocks pivot grid. It's pretty Amazing. 

I've added a feature request here. https://github.com/holoviz/panel/issues/1132#issue-577350903

If you run

```python
import panel as pn
import param
pn.config.sizing_mode="stretch_width"

pn.config.js_files["webdatarocks-toolbar"]="https://cdn.webdatarocks.com/latest/webdatarocks.toolbar.min.js"
pn.config.js_files["webdatarocks"]="https://cdn.webdatarocks.com/latest/webdatarocks.js"
pn.config.css_files.append("https://cdn.webdatarocks.com/latest/webdatarocks.min.css")


html = """
<div id="wdr-component"></div>
<script>
var pivot = new WebDataRocks({
    container: "#wdr-component",
    toolbar: true,
    report: {
        dataSource: {
            filename: "https://cdn.webdatarocks.com/data/data.csv"
        }
    }
});
</script>"""

class WebDataRocks(pn.pane.WebComponent):
    html = param.String(html)
    properties_to_watch = param.Dict({"toolbar": "toolbar"})

    datasource_filename = param.String()
    toolbar = param.Boolean(True)
    filename = param.String("https://cdn.webdatarocks.com/data/data.csv")

WebDataRocks().servable()
```

You get the below.

image
image
image

MarcSkovMadsen:
I have one important question based on my experiments with the WebDataRocks pivot grid. 

**Should the WebComponent inherit from `HTML` instead of `Widget`?** That would also change the `html` attribute to the `object` attribute of the HTML class.

@philippjfr . I would really like some input from you here because I can waste a lot of time if I don't get this right. Thanks.

## Background

In order to get the script in the `html` WebDataRocks string executed I have to add some functionality from the panel .ts implementation of the `HTML` pane. And it lead me to wonder whether the web component should inherit from `HTML` instead of `Widget` as I am currently doing.

I was actually so convinced that I did a lot of the refactoring. See this branch. https://github.com/MarcSkovMadsen/panel/tree/web-data-rocks. I got everything working except the communication from the front end to the backend. I simple could not get that working even though I can see that everything works as expected in the browser. I just don't know how to find out if the data is not sent to the bokeh model or not sent from the bokeh model to the panel model.

@philippjfr . Is it worth spending time on figuring out why this does not work?

philippjfr:
>Should the WebComponent inherit from HTML instead of Widget? That would also change the html attribute to the object attribute of the HTML class.

I think it's fine to inherit from Widget and perhaps factor out whatever code you need from the HTML class into a separate function.

MarcSkovMadsen:
I've started working on the PerspectiveViewer widget. It's not a pane that wraps the perspective package. The perspective package is not needed. It's only the front end viewer than I'm using.

The goal is to enable sending a dataframe to the frontend and displaying that using perspective-viewer + enabling interactivity.

```python
class Perspective(WebComponent):
    html = param.String(
        """
    <perspective-viewer id="view1" class='perspective-viewer-material-dark' style="height:100%;width:100%"></perspective-viewer>
    <script>
    console.log("start")

    var data = [
        { x: 1, y: "a", z: true },
        { x: 2, y: "b", z: false },
        { x: 3, y: "c", z: true },
        { x: 4, y: "d", z: false }
    ];

    console.log(document.currentScript);
    var viewer = document.getElementById("view1");
    viewer.load(data);

    console.log("end")
    </script>
    """
    )
    attributes_to_watch = param.Dict({
        "class": "theme",
        "plugin": "plugin",
    })

    theme = param.ObjectSelector("perspective-viewer-material-dark", objects=THEMES)
    plugin = param.ObjectSelector(Plugin.GRID.value, objects=Plugin.options())
```

perspective (1)

MarcSkovMadsen:
I've begun work on adding some general way of transfering data from the Panel WebComponent model to the client using the ColumnDataSource. I'm using the Perspective widget as my test case.

There are two things I need to handle. 

- The ColumnDataSource stores data as a dictionary of columns and the Perspective viewer load lists of dictionary rows. So I'm thinking that I need to be able to configure how the ColumnDataSource should be transformed in the client before it's being loaded. I think the `orient` parameter of the `DataFrame.to_dict` should be the inspiration here.
- How to load and update could differ between webcomponents. Some will have both load and update functionality. Some only load. Some will have properties that need to be set. Some will have functions that take the data as input.

I got the transfer of data and transformation to list of rows working.

perspective

MarcSkovMadsen:
Just for the record. I've added the .js MutationObserver to observe attributes changes of the WebComponent. I tried communicating the html string between client and server. I can see it does not work. I get infinite recursion.

Instead of solving the problem i will just implement another solution. Communicating the html string is not efficient anyways I believe. I should just communicate the attribute changes in the same way as I communicate property changes using the attribute_last_change parameter.

MarcSkovMadsen:
Just for the record. I showed my son https://github.com/DGLaurits Panel and how to create new WebComponents. We started working on the Material web components.

The work can be found in this form https://github.com/DGLaurits/panel.git in the `material` branch.

MarcSkovMadsen:
@philippjfr . I've added a lot of questions in the code. You can search for them via `@Philppfr`.

It would be really helpfull if you could guide me or maybe solve some of the problems. Any help is appreciated.

MarcSkovMadsen:
Added echarts/ pycharts as an example WebComponent.

Really easy to support communication from server to client. The other way is not supported.

```python
HTML = """
    <div class="echart" style="width:100%;height:100%;"></div>
    <script type="text/javascript">
        var myScript = document.currentScript;
        var myDiv = myScript.parentElement.firstElementChild;
        var myChart = echarts.init(myDiv);
        myDiv.eChart = myChart;
        Object.defineProperty(myDiv, 'option', {
           get: function() { return null; },
           set: function(val) { this.eChart.setOption(val); this.eChart.resize();}
        });
        Object.defineProperty(myDiv, 'resizes', {
           get: function() { return null; },
           set: function(val) { this.eChart.resize();}
        });
    </script>"""

class EChart(pn.pane.WebComponent):
    html = param.String(HTML)
    properties_to_watch = param.Dict({"option": "option", "resizes": "resizes"})

    echart = param.Parameter()
    option = param.Dict()
    resizes = param.Integer()


    def __init__(self, **params):
        if "echart" in params:
            params["option"] = self._to_echart_dict(params["echart"])
        super().__init__(**params)

    @classmethod
    def _to_echart_dict(cls, echart):
        if isinstance(echart, dict):
            return echart
        if 'pyecharts' in sys.modules:
            import pyecharts
            if isinstance(echart, pyecharts.charts.chart.Chart):
                return json.loads(echart.dump_options())

        return {}

    @param.depends("echart", watch=True)
    def update(self):
        self.option = self._to_echart_dict(self.echart)
```

This only thing I can't get working is `resizing`. Its like running the `eChart.resize()` has no effect when done during `render()` of the bokeh model. But if I run it later it works.

@philippjfr or @xavArtley  Any idea how to get this working?

echarts

xavArtley:
May be like for the vtk panel you have to override after_layout method and call the resize method in it

MarcSkovMadsen:
Thanks @xavArtley . That did the trick. 

echarts

MarcSkovMadsen:
I cannot find `resize` functionality on the Google Model Viewer so i've created a support/ feature request here https://github.com/google/model-viewer/issues/1088

modelviewr

MarcSkovMadsen:
Just for the record. The basic functionality for the `perspective-viewer` has been implemented and is working well. The things that I would expect not working are streaming and getting selected rows back. But the rest I would expect to work just fine.

perspective

MarcSkovMadsen:
We've discussed to refactor this pull request into such that 

- The WebComponent becomes a part of the panel package
- The Perspective, echarts/ pycharts, model-viewer, wiredjs implementations are moved to the examples gallery.

@philippjfr . Could you please confirm this is what you want? (It will require some work on my side, so please confirm).

philippjfr:
> Could you please confirm this is what you want? (It will require some work on my side, so please confirm).

Yes, I think this makes the most sense. As we refine the WebComponent model and the examples we can eventually consider migrating them into Panel itself or creating one or more panel extension packages.

philippjfr:
Superseded by https://github.com/holoviz/panel/pull/1122

ltalirz:
In case others stumble across this, it was meant to read "Superseded by #1252"

PR #1123, opened 2020-02-28
[VTKSlicer]fix bokeh hook function

xavArtley:
minor fix


philippjfr:
Assume this is ready? Need to document it somewhere but if you could use the status labels for PRs to indicate readiness that'd be helpful.

PR #1127, opened 2020-03-05
Added try/except for retrieving library version

julioasotodv:
In some weird environments, `param`'s way of retrieving version information has many bugs (since it opens a subprocess to call git, for instance, which gives problems when Python processes are created with systems such as Supervisor or Gunicorn). This commit adds the option for the library to keep on going, even if the version cannot be correctly retreived.

This is just the way it is also done in https://github.com/holoviz/param/blob/7d2d2a848dc79d9741709702f8b47f87497701b0/param/__init__.py#L46

BTW I have made basically the same PR in `pyviz_comms`, as I am facing the same problem using both libraries when the Panel server is launched from Gunicorn/Supervisor

### DIFF
--- a/panel/__init__.py
+++ b/panel/__init__.py
@@ -18,6 +18,9 @@ from .pane import panel, Pane # noqa
 from .param import Param # noqa
 from .template import Template # noqa
 
-__version__ = str(_param.version.Version(
-    fpath=__file__, archive_commit="$Format:%h$", reponame="panel"))
+try:
+    __version__ = str(_param.version.Version(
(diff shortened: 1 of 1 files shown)


jbednar:
See https://github.com/holoviz/pyviz_comms/pull/48 ; I think we should be fixing this in param.version.Version, e.g. by wrapping all of git_fetch in a try/except, not by requiring each package using Version to do this as in this PR.

julioasotodv:
Yeah, I should probably close this PR.

PR #1130, opened 2020-03-06
[VTKVolume] Use datasource to transfer data

xavArtley:
Hope to increase the speed of data transfer to the VTKVolume viewer using a ColumnDataSource

I need to make some timing

### DIFF
--- a/panel/models/vtk.py
+++ b/panel/models/vtk.py
@@ -6,7 +6,7 @@ from bokeh.core.properties import (String, Bool, Dict, Any, Override,
                                    Instance, Int, Float, PositiveInt, Enum)
 from bokeh.core.has_props import abstract
 from bokeh.core.enums import enumeration
-from bokeh.models import HTMLBox, Model
+from bokeh.models import HTMLBox, Model, ColumnDataSource
 
 vtk_cdn = "https://unpkg.com/vtk.js"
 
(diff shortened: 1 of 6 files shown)


xavArtley:
I noticed several limitations depending on the context
The largest data I was able to load without making chrome crash is an array of 400MB
However I was able to reach this array size only by saving an html file and loading it in chrome
If I try with a bokeh server (``show` method) I'm limited around 100-150MB

mattpap:
At least with a standalone HTML file, the syntax error is bogus, because the script failed much earlier when trying to load content of `<script>` element containing document's JSON. This silently fails and the content is empty resulting in an invalid JSON string. In my testing, 150 MB (ish) was the limit. Note that the serialized data is 25% larger due to base64 encoding, so 200 MB (ish) was the actual limit. This isn't unexpected, though seems very low. The server case is more concerning, because there should be no limit there, except for the available memory.

philippjfr:
So despite these limitations would you like to merge this @xavArtley, I'm very happy to defer to your judgement here.

xavArtley:
Can you wait for the moment? I'm in holidays for the next 2 weeks and I'd
like to be able to make some modificationif there is any bug.

Le lun. 9 mars 2020 à 21:19, Philipp Rudiger <notifications@github.com> a
écrit :

philippjfr:
I'll close for now. Let me know if you want to pick this up again.

PR #1157, opened 2020-03-16
App Template with Themes

MarcSkovMadsen:
I would like to contribute my ideas and code regarding a general App Template.

Must have
- Header
- Main Area
- Optional Sidebar.
    - Sidebar can be hidden.
- looks nice on mobile, tablet and desktop.
- Responsive
    - To window size and orientation changes
- White and Dark Theme
- Performant wrt. Bokeh layout engine.
- User can provide his own templates.
-  Support use of potential, upcoming Nav, Top etc. web components.
- Configurable
    - Sidebar width
    - Header height
    - Title name, url and icon
- Holoviews theme set to match selected app theme.
- Easy and intuitive interface for adding and removing items from the header and sidebar.
    - This will be used in a Single Page App where the content/ sub pages is added and removed dynamically. Just like awesome-panel.org does it currently.

Nice to have
- Optional footer
- Many themes like Default, Dark, Bootstrap, Material, Vaadin, Wired etc.
- User can change theme when he like. For example from Default to Dark.
- Examples to illustrate how to easily add
    - Navigation
    - Share on social links
    - Header links for example GitHub
    - Authentication
- Plotly, Altair, ECharts theme set to match selected app theme.

Here is an example of the look of the Bootstrap Template used by awesome-panel.org. This is already in the code.

image


MarcSkovMadsen:
I've realized that its probably best to seperate the app template and the theme.

I would like there to be default, default_dark, material and material_dark theme.

I must find those somewhere. For now there is a dark theme that looks like

image

MarcSkovMadsen:
I've experimented a lot with creating themes. So I've worked on a ThemeBuilder including ColorPalette, and ColorScheme components.

color_palette

theme_builder

There is so much functionality in this. I should be refactored into smaller, individual template and theme pull requests I believe.

MarcSkovMadsen:
Experimenting with the theme builder. I'm starting to understand the structure of the Material Design system. I can see that this could be a major task. In order to have motivation its important to have a concrete use case. I have a concrete use case for parts of this. But the rest would have to be done as things show up.

theme_builder

MarcSkovMadsen:
Looks like this now. I am sort of beginning to understand how to structure this. But it seems like a major undertaking.

If anybody has good ideas on how to speed the development up. Please let me know. (@philippjfr ?).

theme_builder

philippjfr:
Looking fabulous, let's discuss in person tomorrow I think.

MarcSkovMadsen:
Questions regarding Themes and Theme Builder

1. Should Panel come with premade themes? 
   a. A default light an alternative dark theme only?
   b. Multiple themes that might grow over time?
2. Should Panel support custom user themes?
    a. How can the user easily specify his theme?
        i. Would it be an idea that the user could put his `_styles` folder and they are automatically loaded on start up?
        ii. Would it be an idea if the `_styles` folder contains files with the same names as the panel `_styles` folder that they are loaded instead?
3. Should Panel support creating custom themes like illustrated with the Theme Builder?
    a. Or should that be left to some external project/ repository/ web site?
        i. A new repository under HoloViz at least as long as the development and release cycle could be rapid.
        ii. A new repository under marcskovmadsen for some time untill it stabilizes and matures. Could be available at awesome-panel.org. The user could customize and the download the css and json files created.
4. Is it ok to bet on and stick to the material design theme?
    a. Should the bootstrap theme be supported?
5. Should the ThemeBuilder create the full css files?
    a. Or just the changes to the original css files? (This is how its done now)
6. Should the ThemeBuilder use python and f-strings to create the style files? (This is how its done now).
    a. Should it alternatively use existing frameworks like less or sass and their build tools?
    b. Should it aim to update the Bokeh less files and make them more customizable? (As far as I can see they are not built from a general design model. They are hard coded in places. So my estimate is that this requires a lot of work).

Any comments @philippjfr ?

#### Additional Context

##### Use Cases

I believe that the aim is to support the following use cases

###### Use Case 1: User can configure a theme

Something like

```
pn.config.theme = "material-purple-green"
```

###### Use Case 2: User can easily use his own theme

Either by having his styles files in a `_style` folder in the root of his project. Or by loading them via `pn.config.css_files`. 

But the user should be able to replace the panel files (for example the dataframe.css file) so that they are not loaded twice.

###### Use Case 3: User can easily configure his own theme

Something like the theme builder.

Maybe the user can use set

```
theme = Theme(primary="purple", secondary="green", warning="red", fonts="roboto")
pn.config.theme=theme
```

Or maybe the user has access to a ThemeBuilder where he can download the style files and then use them via solutions for Use Case 2.

##### Perspectives

- I would hope to contribute or see some one contribute a set of [Material Web Components](https://github.com/material-components/material-components-web-components). I believe Material would be the most safe bet. There I would also bet on the material design system.

Alternative are the Vaadin or Bootstrap web components I think.

MarcSkovMadsen:
I believe there are basically two primary types of templates needed in Panel

1. A general App template with a menu and navigation
2. A general Dashboard template.

For the general Dashboard template you could get some inspiration from this general example

I don't think the layout/ style is so important. But the functionality is.

image

I see the following functionality

1. A navigation system where the user can navigate between dashboards and *sheets* in the dashboard.
2. There is functionality to 
    a. Pause any data load if the user wants to change a lot of settings.
    b. Manually update the view of data load is paused
    c. Opportunity to subscribe to dashboard
    d. Opportunity to easily share a link to the exact view and state of the dashboard.
    e. Functionality to download the (source) data. See this as one big table/ data frame that is then filtered/ aggregated in different ways by Tableau to show the different plots/ tables.
    f. Comment functionality (have never seen a user use this :-))
    g. Full screen option.
    h. Branding bar
    i. Functionality to search for dashboards.
    j. Functionality to star a dashboard.

Some functionality that I miss is a link to some documentation that explains the why, who and how of the dashboard. Its important for long term development, maintenance and usage.

Then there is some functionality else where where can get insights on how (much) the dashboard is used.

MarcSkovMadsen:
I also believe a *Gallery* template would be nice to have. I have one at awesome-panel.org. But a nicer looking one is here

image

https://training.talkpython.fm/courses/all

Could be an improvement to the template used for serving multiple apps.

FYI. @philippjfr . See this comment and the one above before we talk later today.

philippjfr:
This is incredibly difficult to review since it includes both the theming work and the templating work. I'm going to open a separate templating PR, providing some of the basic organization and architecture for templates and once that's merged I'll probably rebase this PR on top of that one.

MarcSkovMadsen:
I know. It was not originally the intention. It just grew.

philippjfr:
The default templating has been merged into the other PR now. We'll be tweaking that for a while and I'd love to get your feedback on the existing templates. That PR also introduced a theming component. I'll close this PR but keep the branch so we can pull out what we need later. Maybe you can ship the theme builder separately for now?

Issue #1118, opened 2020-02-23
Enable Panel to serve or appear to serve static assets


MarcSkovMadsen:
#### My Pain

I would like to be able to serve static assets to users of Panel.

- This can be assets that are developed before I start the Panel server.
- This can be assets that are generated by the user of the Panel application by clicking some button or something else.
- This can be assets that are generated by a batch job.

The assets can be text files, csv files, json documents, xlsx documents, images and in prinicple any file type.

#### Solution(s)

- Enable serving static assets from a static folder and its subfolders

#### Alternative Solution

Maybe this should not be a part of Panel. But if not a lot of users would still like to know how to obtain this functionality. So maybe some documentation describing one or more best practices and links to guides could be a solution. 

Maybe it is as easy as linking to some documents on how to use the already running Tornado server?

This should include a description on how to use a reverse proxy or Tornado (?) to enable using a common url like

example.com
example.com/static

for the Panel app and the static assets served.

MarcSkovMadsen:
@philippjfr . Just some words on your thoughts and how you would do this would be very valuable. Maybe me or some other user could try it out in our own apps, return the learnings and then one day this could be a part of Panel.

philippjfr:
This should be easy enough, I'd suggest having a configurable endpoint on `pn.state.static_directory` or similar.

MarcSkovMadsen:
Another use needs to be able to serve static assets with `pn.serve` and `pn.show()`.

https://discourse.holoviz.org/t/adding-local-js-files-to-my-app/430/6

lvankouwen:
I have been playing around with panel for a few weeks and I really like it; thanks for making such a nice framework available. The one thing I find missing most is being able to add static assets as described in this issue. So, a solution to this issue would be much appreciated!

philippjfr:
This is now supported and document in https://pyviz-dev.github.io/panel/user_guide/Deploy_and_Export.html

PR #1164, opened 2020-03-19
Allow serving REST APIs as part of panel serve

philippjfr:
Allows serving using `panel serve some_script.py some_notebook.ipynb --tranquilize` to serve REST APIs alongside the panel apps. Automatically discovers any tranquilized functions.

```python
import panel as pn

from tranquilizer import tranquilize

@tranquilize()
def order(cheese):
    '''I'd like to buy some cheese!'''
    return {'response':"I'm afraid we're fresh out of {}, Sir.".format(cheese)}

pn.panel("# A panel app").servable()
```

e.g. this will serve the Panel app on `http://localhost:5006/` and the REST API on `http://localhost:5006/rest/order`.


jlstevens:
This is great, thanks! I'll now check out this branch and test it...

MarcSkovMadsen:
If possible it would be so great if this supported

- filter parameters like ?type="goat" (for goat cheese :-))
- serving csv and xlsx files as well.
- get, post, update, delete requests
- Open API documentation

and the documentation hinted at how to use this in practice using serializers etc.

I know I'm hoping for a lot. I guess the alternative is to people (including me) in the direction of Django REST API, fast API, flask or something similar. But would nice to see this nicely integrate.

MarcSkovMadsen:
And one more thing. Would be nice if the documentation explained the word `tranquilize`. I don't understand the meaning of it in this context at least.

philippjfr:
I think for file serving we can easily add a simple static file request handler rather than building that on Tranquilizer or some other REST API.

mattpap:
> tranquilized functions

Is this a well defined term? Seeing just the title of this PR I didn't have any idea what to expect from its contents. I would use well known and defined terms like REST API across the board, and leave the name of the library as an implementation detail. So on the command like I would just use something along `--allow-rest-apis` or `--serve-rest`.

philippjfr:
Yeah, I hear you, we're still iterating on what we actually want to support and this PR was mostly a prototype for a client.

MarcSkovMadsen:
@philippjfr . I would like to serve some static files like sphinx documentation, some Angular .js files or similar. But the most important is actually to serve dynamic data in csv, xlsx and parquet format.

My primary use case is creating a dashboard and when the user gets to a dead end in the dashboard/ wants more she can get a link to the exact data in the plot, table or similar depending on the state of the app/ the value of sliders, dropdowns etc. She can use the link to download the data here and now. But also store or share the link. Or use PowerQuery in excel to setup a permanent link. Build her own small Panel app using Pandas and the link to get the data etc.

Dynamic, shareable links are powerfull.

philippjfr:
I understand that's a very different thing from a REST API though and is already covered by this issue of yours: https://github.com/holoviz/panel/issues/1118

JuliaS92:
Hi @philippjfr, as far as I understand this was part of the 0.10.0 release. Could you please provide an updated minimal example of how to use this? I tried the code snippet from the top of this thread and neither panel version 0.10.0, nor 0.10.3 accepted the --tranquilize option and it seems there are several --rest.. options added instead.
Thank you very much in advance.

PR #1165, opened 2020-03-19
Update signature of bokeh.io.export.export_png

Guillemdb:
The [bokeh.io.export.export.export_png](https://github.com/bokeh/bokeh/blob/12a3c0b7fbf36700b542acdefad15a518391d582/bokeh/io/export.py#L58) function has been updated, and now `filename` needs to be passes as a keyword argument. 

* This PR updates the filename parameter passed to the function to reflect the update of the `export_png` signature.

### DIFF
--- a/panel/io/save.py
+++ b/panel/io/save.py
@@ -45,7 +45,7 @@ def save_png(model, filename):
         state.webdriver = webdriver_control.create()
 
     webdriver = state.webdriver
-    export_png(model, filename, webdriver=webdriver)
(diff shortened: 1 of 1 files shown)


philippjfr:
Thanks @Guillemdb!

mattpap:
As an FYI, all APIs in bokeh will get keyword only args (if needed) in 3.0. I think it's a good idea to start writing code now that uses the APIs as if this was done already. This will simplify transition to the next major version.

philippjfr:
Thanks for the heads up.

Issue #1148, opened 2020-03-12
[bokeh] Closing connection: No msgid in header


maximlt:
With the following environment on Windows 10...
```
# packages in environment at C:\Users\maxim\Miniconda3\envs\pluies_poc:
#
# Name                    Version                   Build  Channel
altgraph                  0.16.1                     py_0  
asn1crypto                1.3.0                    py37_0  
atomicwrites              1.3.0                    py37_1  
attrs                     19.3.0                     py_0  
backcall                  0.1.0                    py37_0  
blas                      1.0                         mkl  
bleach                    3.1.0                    py37_0  
bokeh                     1.4.0                    py37_0  
ca-certificates           2020.1.1                      0  
certifi                   2019.11.28               py37_0  
cffi                      1.13.2           py37h7a1dbc1_0  
chardet                   3.0.4                 py37_1003  
colorama                  0.4.3                      py_0  
colorcet                  2.0.2                      py_0    pyviz
cryptography              2.8              py37h7a1dbc1_0  
cycler                    0.10.0                   py37_0  
decorator                 4.4.1                      py_0  
defusedxml                0.6.0                      py_0  
entrypoints               0.3                      py37_0  
freetype                  2.9.1                ha9979f8_1  
future                    0.18.2                   py37_0  
holoviews                 1.12.7                     py_0    pyviz
hvplot                    0.5.2                      py_0    pyviz
hypothesis                5.4.1                      py_0  
icc_rt                    2019.0.0             h0cc432a_1  
icu                       58.2                 ha66f8fd_1  
idna                      2.8                      py37_0  
importlib_metadata        1.4.0                    py37_0  
intel-openmp              2019.4                      245  
ipykernel                 5.1.4            py37h39e3cac_0  
ipython                   7.11.1           py37h39e3cac_0  
ipython_genutils          0.2.0                    py37_0  
ipywidgets                7.5.1                      py_0  
jedi                      0.16.0                   py37_0  
jinja2                    2.10.3                     py_0  
jpeg                      9b                   hb83a4c4_2  
jsonschema                3.2.0                    py37_0  
jupyter                   1.0.0                    py37_7  
jupyter_client            5.3.4                    py37_0  
jupyter_console           6.1.0                      py_0  
jupyter_core              4.6.1                    py37_0  
kiwisolver                1.1.0            py37ha925a31_0  
libpng                    1.6.37               h2a8f88b_0  
libsodium                 1.0.16               h9d3ae62_0  
[...]

PR #1170, opened 2020-03-20
Respect write-locks in synchronous Websocket events

philippjfr:
As described in #1148 we override the bokeh Websocket handlers asynchronous locking mechanisms to allow synchronous functions to trigger updates on the frontend immediately instead of blocking until they finish, e.g.:

```python
import panel as pn
import time

button = pn.widgets.Button(name='Run')
progress = pn.widgets.Progress(value=0)

def update_progress(event):
    for i in range(100):
        time.sleep(0.01)
        progress.value = i

button.on_click(update_progress)

pn.Row(button, progress).show()
```

If we didn't do this, this example wouldn't update the progress bar until the callback finishes. However in implementing this we are not respecting the Websocket handlers write lock. This PR ensures checks whether a write lock is enabled and if so locks the synchronous events out. This ensures most computations where updates aren't triggered in very quick succession can synchronously trigger events but may result in unexpected behavior where a callback suddenly locks in the middle. However once the callback finishes it'll all be made consistent so I think this is a worthwhile tradeoff until we handle async callbacks correctly. 

Fixes https://github.com/holoviz/panel/issues/1148

### DIFF
--- a/panel/io/server.py
+++ b/panel/io/server.py
@@ -76,9 +76,12 @@ def unlocked():
         events = []
         for conn in connections:
             socket = conn._socket
+            if hasattr(socket, 'write_lock') and socket.write_lock._block._value == 0:
+                state._locks.append(id(socket))
+            locked = id(socket) in state._locks
             for event in curdoc._held_events:
                 if (isinstance(event, ModelChangedEvent) and event not in old_events
-                    and hasattr(socket, 'write_message')):
+                    and hasattr(socket, 'write_message') and not locked):
                     msg = conn.protocol.create('PATCH-DOC', [event])
                     WebSocketHandler.write_message(socket, msg.header_json)
                     WebSocketHandler.write_message(socket, msg.metadata_json)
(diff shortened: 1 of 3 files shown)


panel/io/server.py
```diff
@@ -76,9 +76,12 @@ def unlocked():
         events = []
         for conn in connections:
             socket = conn._socket
+            if hasattr(socket, 'write_lock') and socket.write_lock._block._value == 0:
+                state._locks.append(id(socket))
+            locked = id(socket) in state._locks
```
mattpap: btw. This seems to me like a perfect case for using `WeakSet`.

philippjfr: 👍

mattpap:
@philippjfr, what's your policy on branch removal? I see that most (all?) are being kept after mege.

philippjfr:
>@philippjfr, what's your policy on branch removal? I see that most (all?) are being kept after mege.

There's an option to do this automatically right? Right now I manually delete them once in a while.

mattpap:
https://help.github.com/en/github/administering-a-repository/managing-the-automatic-deletion-of-branches

philippjfr:
Thanks, enabled.

PR #1181, opened 2020-03-23
Alert pane: Allows providing contextual feedback messages for typical user actions

MarcSkovMadsen:
One of the things i use a lot and see other users using in their analytics app are [Bootstrap inspired Alerts](https://getbootstrap.com/docs/4.0/components/alerts/). 

So I would like to contribute one.

The contributed code is **ready for review** by @philippjfr . 

I've chosen

- To implement one `Alert` with an `alert_type` parameter instead of a collection of `PrimaryAlert`, ..`InfoAlert`, ..., `DarkAlert` to keep the api in line with `Button`.
- To add an alert.css file instead of styling inline. That will enable easier custom styling.
- Not to implement a custom Bokeh model supporting the `alert_type`. Let me know if that is needed.
- To include it in `pn.pane` instead of `pn.widget` or `pn.components.bootstrap`. Those would be alternatives.
- To make the 'stretch_width` by default. I believe this is normal behaviour.
- To not implement the `alert-link` css class. Instead any `<a>` link tag will just be styled.
- To not implement the `alert-dismissible` js functionality. Would be nice to have though.

#### Alert notebook reference example

image

image

image

#### Server test example

`python -m panel serve 'panel\tests\pane\test_alert.py' --dev --show`

alert

#### Additional Context

This is an example of where the `css_classes` parameter does not work that well in Panel @philippjfr . In order to get the proper styling you need to define `.bk.alert-primary` instead of `.alert-primary` to get specific enough. That means you can't copy-paste/ or include one of the many existing style sheets out there. This is friction. :-)

### DIFF
--- a/panel/pane/alert.py
+++ b/panel/pane/alert.py
@@ -1,0 +1,55 @@
+"""Bootstrap inspired Alerts
+
+See https://getbootstrap.com/docs/4.0/components/alerts/
+"""
+import param
+
+from panel.pane.markup import Markdown
+
+ALERT_TYPES = ["primary", "secondary", "success", "danger", "warning", "info", "light", "dark"]
+
+
+class Alert(Markdown):
+    """
+    An Alert that renders Markdown
+
+    - CSS Styling is done via the classes `alert` and `alert-TYPE`, where TYPE is the alert_type.
+    - sizing_mode is set to `stretch_width` by default
+    """
+
+    alert_type = param.ObjectSelector("primary", objects=ALERT_TYPES)
+
+    def __init__(
+        self, text: str, **kwargs,
+    ):
+        """An Primary Alert that renders Markdown
+
+        - CSS Styling is done via the classes `alert` and `alert-primary`.
+        - sizing_mode is set to `stretch_width` by default
+
+        Arguments:
+            text {str} -- Some MarkDown text
+        """
+        if "margin" not in kwargs:
+            kwargs["margin"] = (0, 0, 25, 0)
+        if "sizing_mode" not in kwargs:
+            kwargs["sizing_mode"] = "stretch_width"
+
+        super().__init__(
+            text, **kwargs,
+        )
+
+        self._set_css_classes()
+
+    @param.depends("alert_type", watch=True)
+    def _set_css_classes(self):
+        css_classes = []
+        if self.css_classes:
+            for class_ in self.css_classes:
+                if class_ != "alert" and not class_.startswith("alert-"):
+                    css_classes.append(class_)
+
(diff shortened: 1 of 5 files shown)


philippjfr:
I've heard some hopeful things about eventually being able to remove the css prefixing but I doubt that'll happen in the near future.

philippjfr:
Thanks for the contribution though, very sensible addition.

MarcSkovMadsen:
A small step for man :-)

panel/pane/alert.py
```diff
@@ -0,0 +1,55 @@
...
+
+    - CSS Styling is done via the classes `alert` and `alert-TYPE`, where TYPE is the alert_type.
+    - sizing_mode is set to `stretch_width` by default
+    """
+
+    alert_type = param.ObjectSelector("primary", objects=ALERT_TYPES)
+
+    def __init__(
+        self, text: str, **kwargs,
+    ):
```
philippjfr: Sticking with this style for now:

```suggestion
    def __init__(self, text: str, **kwargs):
```

philippjfr:
Apart from my minor two comments this is mergeable. Could you maybe also generate a thumbnail for the gallery and upload it here? That way I can add it to the CDN and it'll find it during the next doc build.

MarcSkovMadsen:
One question @philippjfr is whether the `type` and colors of button and Alert should somehow be aligned?

image

image

philippjfr:
>One question @philippjfr is whether the type and colors of button and Alert should somehow be aligned?

Might be nice, maybe we should have a saturated=True/False option to toggle between the pastel and fully saturated colors?

MarcSkovMadsen:
> > One question @philippjfr is whether the type and colors of button and Alert should somehow be aligned?
> 
> Might be nice, maybe we should have a saturated=True/False option to toggle between the pastel and fully saturated colors?

...hmmmm.... I'm not an expert on colors. Try to stay out of it but always gets to spend a lot of time on it. Any experts around?

panel/pane/alert.py
```diff
@@ -0,0 +1,55 @@
...
+            text {str} -- Some MarkDown text
+        """
+        if "margin" not in kwargs:
+            kwargs["margin"] = (0, 0, 25, 0)
+        if "sizing_mode" not in kwargs:
+            kwargs["sizing_mode"] = "stretch_width"
+
+        super().__init__(
+            text, **kwargs,
+        )
```
philippjfr:
```suggestion
        super().__init__(text, **kwargs)
```

philippjfr:
Merging for now, can add the saturated colors later.

PR #1182, opened 2020-03-23
Card layout

MarcSkovMadsen:
I believe a `Card` `layout` is one of the things needed to help organize your app or dashboard. Currently Panel does not provide a Card layout..

I will try to contribute one. The one I have in mind is not based on a custom Bokeh model. Instead it is developed using existing functionality like Columns, Rows, Markdown and Button + CSS hacks.

image

FYI. @philippjfr . This is just a draft version. I will try to finish this. But let me know if this is not something that will fit into Panel or where you need it developed in some other way.

Tasks
- Enable changing header and body dynamically. Currently its done during construction only.
- Test
- Document.


MarcSkovMadsen:
The card now works as I want it to.

You can test it using `python -m panel serve 'panel\tests\pane\test_card.py' --dev --show`

card

Only thing missing is the notebook reference example.

MarcSkovMadsen:
I've added a reference example notebook. There are two issues in there marked with @philippjfr . 

One is the scroll which starts in the header. The other is the border which is not shown at the bottom of some of the examples. My understanding is that the bokeh layout engine does not take borders into account.

There is also the question of where this should reside currently its called as `pn.card.Card`. I think it would be better with `pn.layout.Card` but the problem is that it cannot in its current implementation be in the `layout` module as it also uses functionality like `HTML` and `Markdown` from the `pane` module.

The card is very usable besides these things I believe.

**Ready for review by @philippjfr** . Maybe it can be released. Maybe it needs a rewrite as a custom bokeh model. Maybe there are some bugs. Let me know.

card_ref_example

philippjfr:
Closing in favor of #1262

PR #1187, opened 2020-03-25
Compatibility with Bokeh 2 DatePicker

hyamanieu:
See https://github.com/bokeh/bokeh/pull/9509 , date is in ISO format.


philippjfr:
Thanks, we'll also have to update the existing tests which mocks a change in the date but still uses the old date format https://github.com/holoviz/panel/blob/ffcdd3cac26ee01d0e64dbdfb6afe52dd19681a0/panel/tests/widgets/test_input.py#L42

philippjfr:
Thanks again! Hugely appreciated.

PR #1189, opened 2020-03-25
Example of using Folium in Panel - Notebook for Gallery

MarcSkovMadsen:
Demonstrates the use of [`Folium`](https://python-visualization.github.io/folium/) in Panel.

This notebook was created as a response to the [Panel and Folium do they mix](https://discourse.holoviz.org/t/panel-and-folium-do-they-mix/342) question on [Discourse](https://discourse.holoviz.org/).

panel_folium_example


MarcSkovMadsen:
See also https://twitter.com/MarcSkovMadsen/status/1242896463635591168

Share if you like.

philippjfr:
Thanks, nice how easy this is. Could you either add folium to the test dependencies in setup.py or put a regex to skip testing the notebook at the end of the tox.ini?

MarcSkovMadsen:
I've added Folium to the test requirements @philippjfr 

image

MarcSkovMadsen:
I've shared a tweet. Feel free to share if you like https://twitter.com/MarcSkovMadsen/status/1243405883175911426?s=20

voltage_example

philippjfr:
>FileNotFoundError: [Errno 2] File sample_data.csv does not exist: 'sample_data.csv'

Seems to be erroring, I also haven't quite decided how to handle sample data files in Panel yet.

MarcSkovMadsen:
The idea is that the user generates the sample data by clicking the save button.

Will look for a workaround

PR #1193, opened 2020-03-26
Fileupload with Pandas and Plotly example

MarcSkovMadsen:
This notebook was contributed as an answer to the question [A button to upload csv file and plot it ](https://discourse.holoviz.org/t/a-button-to-upload-csv-file-and-plot-it/365) on [Discourse](https://discourse.holoviz.org/)


philippjfr:
Fabulous, but this should probably go in examples/gallery/param or we could rethink the categories. I should also probably provide descriptions on what the categories were meant to be, `external` was meant as any external component loaded into Panel using custom css and js_files and the HTML pane, e.g. your Folium example could go in there.

MarcSkovMadsen:
I have cleaned it up and moved it to param folder. I will not do more before for now. But if there are helpfull suggestions I will.

But there is one bug i cannot solve

https://github.com/holoviz/panel/issues/1195

The example looks like 

file_upload

philippjfr:
Please also clear the notebook before committing it.

MarcSkovMadsen:
I've cleared the output and committed it @philippjfr

philippjfr:
Same error as in the Folium example, can't find a sample_data.csv.

philippjfr:
Thanks!

PR #1215, opened 2020-04-01
Allow serving Flask apps

philippjfr:
Allows serving flask app using pn.serve, e.g.:

```python
import panel as pn
from flask import Flask

flask_app = Flask(__name__)

@flask_app.route('/app')
def hello_world():
    return 'Hello, World!'

def panel_app():
    return "# This Panel app runs alongside flask, access the flask app at [here](./flask/app)"

pn.serve({'/flask': flask_app, 'panel': panel_app}, port=5001)
```

This will serve the panel app on `localhost:5001/panel` and the flask app on `localhost:5001/flask/app`.

### DIFF
--- a/doc/releases.rst
+++ b/doc/releases.rst
@@ -6,11 +6,12 @@ Version 0.9.4
 
 Date: 2020-04-01
 
-This is a minor release fixing an issue with recent versions of Tornado. It also fixes issue with the packages built on the PyViz conda channel.
+This is a minor release fixing a number of regressions and compatibility issues which continue to crop up due to the upgrade to Bokeh 2.0 Additionally this release completely overhauls how communication in notebook environments are handled, eliminating the need to register custom callbacks with inlined JS callbacks to sync properties. Many thanks to the contributors to this release including @hyamanieu, @maximlt, @mattpap and the maintainer @philippjfr.
 
 Enhancements:
 
 - Switch to using CommManager in notebook hugely simplifying comms in notebooks and reducing the amount of inlined Javascript (`#1171 <https://github.com/holoviz/panel/pull/1171>`_)
+- Add ability to serve Flask apps directly using pn.serve (`#1215 <https://github.com/holoviz/panel/pull/1215>`_) 
 
 Bug fixes:
 
--- a/panel/io/server.py
+++ b/panel/io/server.py
@@ -5,6 +5,7 @@ from __future__ import absolute_import, division, unicode_literals
 
 import os
 import signal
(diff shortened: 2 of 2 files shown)


MarcSkovMadsen:
Wow. As a potential user of this I would like to learn what the best deployment strategi is for different use cases.

Panel on Bokeh server?
Panel on Django server?
Panel on Flask server?

And can I run Panel, Flask and the tranquilizer api on all those servers? How will they perform? What are the pros and cons?

This raises many questions. Can I use all the Flask functionality? For example can I build advanced rest apis using Flask extensions with  open api documentation? Can I serve dynamic files using Flask? Would it be a good idea to mix flask and panel to a large extent? Could I embed Panel apps in the Flask apps I serve? And how? And whats the architecture. I the flask server running. Or is the flask app running on Tornado?

Running on multiple servers as well as in notebook is a super power. But currently not documented and experiences not shared.

Request: Some kind of documentation and examples. :-)

philippjfr:
Yes, we have funded work to document deployment options, so it's definitely coming.

jbednar:
As one specific point among many that need documentation, if you used `@flask_app.route('/')` above, would the flask app be at `localhost:5001/flask`?  If so, that seems like a more suitable first example to me...

philippjfr:
>if you used @flask_app.route('/') above, would the flask app be at localhost:5001/flask? If so, that seems like a more suitable first example to me...

No that explicitly does not work, flask apps may only be served on a subpath.

philippjfr:
Well it was certainly beyond my ability to route things, might give that another go.

jbednar:
Ok, I now see that in the code "Flask apps must be served on a subpath.".  Definitely needs explaining, then...

pgierz:
This looks very very cool....looking forward to seeing some examples!

nitrocalcite:
This is really cool!  However, I also wanted to serve a Tornado app alongside `panel`; though I figured it was possible, I couldn't find any docs.  

So, for future comers to this thread, you can pass extra Tornado `RequestHandlers` to the Bokeh server underlying Panel using the (undocumented) `extra_patterns` kwarg in `pn.server` or `pane.show`.  You should set it to what you'd expect for a Tornado `Application` - a list of tuples that are `(path, handler)` or `(path: str, handler: RequestHandler, handler_init_kwargs: dict)`.

mvirus1996:
@philippjfr 
Can you explain how to redirect in this integration.
I tried "redirect()" function of flask but not working.

PR #1218, opened 2020-04-01
Update README.md

michaelaye:
add requirements section for the jupyter lab extension


philippjfr:
Thanks, won't wait on tests since it can't possible affect anything.

Issue #739, opened 2019-11-02
Should package-lock.json be include?


xavArtley:
Shouldn't we remove package-lock.json from the commited files?
@mattpap

mattpap:
Though it may be tempting due to its usually large size, `package-lock.json` is a useful tool, which allows to have reproducible environments. Otherwise `npm install` would result in always installing most recent packages, up to allowed dependency specification. Note however `npm install` can still change `packge-lock.json` and upgrade dependencies to some extent. To force npm to only use the lock file, one should use `npm ci`. This is the command that should be used in CI systems, as it name actually suggests. To fully update dependencies, the approach that works the best for me, is to remove the lock file and run `npm install` (`rm package-lock.json; npm install; git add package-lock.json`). I suppose there may be a more idiomatic way that I'm not aware of.

philippjfr:
We should add some of that to the developer docs.

PR #1220, opened 2020-04-02
Update guidance about package-lock.json

philippjfr:
Closes https://github.com/holoviz/panel/issues/739

### DIFF
--- a/panel/package-lock.json
+++ b/panel/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "@holoviz/panel",
-  "version": "0.9.4",
+  "version": "0.9.4-1",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
--- a/doc/developer_guide/index.rst
+++ b/doc/developer_guide/index.rst
@@ -110,10 +110,19 @@ performs a pip editable install of Panel.
 Developing custom models
 ------------------------
 
-If you are adding or editing a bokeh model, you can use `panel build
-panel` to build the javascript. The `build` command is just an alias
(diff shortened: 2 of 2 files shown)


panel/package-lock.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@holoviz/panel",
-  "version": "0.9.4",
+  "version": "0.9.4-1",
```
mattpap, line 3: I would suggest using a full version string like `0.9.4-dev.1` or `0.9.4-rc.1` and publishing with an appropriate tag, i.e. `npm publish --access=public --tag=dev` (or `--tag=rc` respectively). For production release you can skip `--tag` altogether or use `--tag=latest` (that's the default).

mattpap: Also version should be kept in `package.json`, then `npm install` will update `package-lock.json`.

philippjfr: So this was a build number generated using `npm version build` because I messed up the initial release of 0.9.4 and did `npm unpublish`, so I wouldn't ordinarily do this. I also realized too late that you can't republish after unpublishing, which means that the currently Python package won't be able to load the 0.9.4 release from the CDN :(

Will have to tag a 0.9.5 soon to fix it.

mattpap: > (...) can't republish after unpublishing (...)

This used to be possible on npm, but it's so unsafe that it was removed not that long ago.

philippjfr: Yeah, it's fair enough, although I wish it gave you a 5 minute window or something or until the first person downloaded it.

philippjfr:
Merging for now, will reset the version on the next release.

Issue #1236, opened 2020-04-07
Update the filename attribute of a FileDownload widget 


maximlt:
Hi!

I've been trying to use the ``FileDownload`` widget to allow my users to download data from a  dashboard. I wanted ``filename`` to be updated according to some parameters. While looking for a way to do that, I found a potential bug.

To reproduce the example below, you'll need to have a file called ``data.csv`` in your repo. It shows that when you create a widget and update its ``button_type``, it's correctly updated. However, after you've clicked once on the button, every time you update ``button_type`` again the *Download* window pops up. I observed the same behaviour trying to update ``filename``. Note that updating ``label`` doesn't have this unwanted side effect.

![image](https://user-images.githubusercontent.com/35924738/78615465-e1041480-7871-11ea-91eb-d04bc0ada696.png)

I believe the issue comes from the fact that the ``render`` function of *panel/models/file_download.ts* watches ``filename``, ``button_type`` (and ``data``). After ``render`` runs once, the attribute ``_initialized`` is set to ``True``, every new change of one of these attributes will trigger a click:

https://github.com/holoviz/panel/blob/1f1399fbbae2ef9a71e214d8da117c3d7124f14d/panel/models/file_download.ts#L62

philippjfr:
Thanks for the detailed issue and for tracking down the actual bug. Any chance you could send a quick PR our way? 😄

maximlt:
Ok I'm gonna give it a try! However, I don't expect my first attempt to be successful, I might need some guidance at some point ;) Let's see how it goes!

MarcSkovMadsen:
Hi @maximlt 

Thanks for taking the time to put in a good description. 

On thing that I find help full sometimes is to add a small video `.gif` to show whats going on. It's sometimes difficult sitting on the other side of an issue and trying to understand what the issue is or to reproduce it.

I use [SceenToGif](https://www.screentogif.com/) to do that.

If I was to take a look at this issue I think it would really have helped me to see the video.

A picture says more than a thousand words. A video `.gif` even more :-)

maximlt:
Fixed in #1246

Issue #1241, opened 2020-04-07
FileDownload widget produces empty file for dataframe xlsx / does not work for BytesIO object


MarcSkovMadsen:
#### Version

Panel 0.9.4

#### My Pain

The FileDownload widget produces empty `xlsx` files when clicking the button.

![image](https://user-images.githubusercontent.com/42288570/78681365-eeaab000-78ec-11ea-8991-6e6c1bffa981.png)

![image](https://user-images.githubusercontent.com/42288570/78681430-0124e980-78ed-11ea-8ebe-2801616fe843.png)

```python
import pandas as pd
import panel as pn
from io import BytesIO

data=pd.DataFrame({"a": [1]})

def get_file():
    output = BytesIO()
    writer = pd.ExcelWriter(output, engine='xlsxwriter')
    data.to_excel(writer, sheet_name="Data")
    writer.save()
    return output

file_download = pn.widgets.FileDownload(filename="data.xlsx", callback=get_file)
file_download
```

and there are no error messages in either the console or the jupyter server log.

#### Additional Context

But if you save the file manually it has contents

![image](https://user-images.githubusercontent.com/42288570/78685118-9924d200-78f1-11ea-86aa-e59872868bf8.png)

```python
with open("data_manual.xlsx", 'wb') as f:
    f.write(get_file().getvalue())
```

#### Solution

- Identify and fix bug
- or document how file download of Pandas XLSX files should be done.

MarcSkovMadsen:
My guess would be that the mime type is not supported (yet).

![image](https://user-images.githubusercontent.com/42288570/78685761-7e9f2880-78f2-11ea-845d-bbe7e63699a3.png)

MarcSkovMadsen:
According to https://filext.com/faq/office_mime_types.html the mime type should be

application/vnd.openxmlformats-officedocument.spreadsheetml.sheet

MarcSkovMadsen:
Having looked more at the code the problem is really that for non-string files the `read` method is used and for the `BytesIO` object created that is the empty string while the `getvalue` method is not.

![image](https://user-images.githubusercontent.com/42288570/78687243-44368b00-78f4-11ea-919f-b1bb2bbac524.png)


![image](https://user-images.githubusercontent.com/42288570/78687220-3da81380-78f4-11ea-87ff-cf1d8fdaa72e.png)

MarcSkovMadsen:
#### Workaround

A workaround for now is to set `output.read=output.getvalue`

```python
[...]

PR #1242, opened 2020-04-07
Filedownload example added

MarcSkovMadsen:
I've contributed an example for downloading a dataframe in xslx format in order to avoid spending time on bugs like https://github.com/holoviz/panel/issues/1241

For now only an xlsx is present. But I believe more examples like `csv`, `parquet` and `json` should be added by someone in the future.

Can be merged now I believe.

image


maximlt:
Hi @MarcSkovMadsen, cool example ;)

Just as a note, there are already examples with getting a csv file from a callback in the docs [here](https://panel.holoviz.org/reference/widgets/FileDownload.html#widgets-gallery-filedownload) and [here](https://panel.holoviz.org/gallery/simple/save_filtered_df.html#simple-gallery-save-filtered-df). Also note that I think that ``xlwt`` is only used to write ``.xls`` files (the old format), and that ``openpyxl`` is the default package used to write ``.xlsx`` files (if installed), see the [pandas docs](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.ExcelWriter.html). `XlsxWriter` works fine too, it's just not the default one.

MarcSkovMadsen:
Hi @maximlt 

Thanks for your valuable feedback.

The **purpose** of this notebook is to provide code examples and **code snippets** that enable you to quickly add FileDownload to your Panel dashboard or application.

This notebook was contributed because I had problems creating the right code to enable downloading DataFrames in `xlsx` format. See [FileDownload widget produces empty file for dataframe](https://github.com/holoviz/panel/issues/1241).

I just experienced multiple problems. 

- I did not have XlsxWriter installed
- I did not have xlwt installed
- I did not `writer.save`
- I did not `.seek(0)`

All that because the examples for producing a `BytesIO` from a `DataFrame` I could find did not describe theme in full. It's not described in the Pandas documentation either.

So I believe other users of Panel could struggle as well and start filing bug reports like I did because I thought the problem was the FileDownload widget in the beginning.

So an example like this could lower the friction of using Panel and supporting Panel.

MarcSkovMadsen:
I refactored the notebook to a general notebook for FileDownload examples.

It could grow over time with more examples when I or other encounter them.

maximlt:
Thanks @MarcSkovMadsen for your reply. Oh yes you're right, this example is gonna be really useful (and the list of examples should definitively grow, I'll try to add some after I fix #1236). I know it's gonna be useful because a few days ago I went through exactly the same process of finding how to download ``xlsx`` files, for sure this is a common pattern in a dashboard :)

I can't see your last commit, is that normal?

MarcSkovMadsen:
The commit is here https://github.com/holoviz/panel/pull/1242/commits/b59b5edc42e8dda14ab1dc39b3fe372621951563

philippjfr:
Some questions, why focus specifically on `xlsx`? It introduces additional dependencies for our builds and as far as I can tell does not demonstrate anything special about the format. I'd probably focus on simple `csv` export and then mention all the other formats that pandas can export to. Alternatively we could even have a dropdown to select which format to export to.

Separate comment, the current code can be simplified to:

```python
def get_file():
    output = BytesIO()
    data.to_excel(output, sheet_name="Data", engine='xlsxwriter')
    output.seek(0) # Important!
    return output
```

On another note, I very much appreciate you wanting to receive credit for your contributions but if that's something we will normalize in this repository I would much prefer if we had a standardized header portion with an author section over adding a whole section to the end of the notebook. Gallery examples should always end with the actual output so we can easily deploy it and then append an iframe of the running app.

maximlt:
Hi!

I tend to think that getting an Excel spreadsheet out a of a dashboard is quite common, that's why I think it's interesting that the docs provide some sample code showing how to to that, even if that code doesn't have much to do with `panel` itself.

I understand that you don't want to add another dependency as if every example needs one then things will go crazy. Could we then just wrap that code into something which would not break the tests:
```python
try:
     import openpyxl  # or xlsxwriter
     # panel/pandas/excel code here
except ImportError:
    raise ImportError(
    "In order to export Excel files from pandas you need to install "
    "an engine such as xlxt for .xls files or openpyxl or xlsxwriter for .xlsx files")
```

Or by simply commenting it?

jbednar:
I don't think a try/except like that would help us that much, as to build the docs that contain that example we'd still have to make sure that the dependency is available.

maximlt:
Yes sure it wouldn't help you much to build the docs but if this code is in the docs somewhere and is not what is ultimately served/displayed (the last cell basically), it could be useful to users, to see that it's possible with panel and that it's required to install another dependency (the docs could point to the pandas docs).

I use very often the search bar on panel.holoviz.org (it could be a little bigger) and it works fine. Dashboards are certainly used a lot by data/business analysts who have to/want to get their output as an Excel file. I'd expect them to look for Excel or xlsx/xls on the website, it's a bonus point for panel if they can find something relevant to them at this point ;) It's particularly important at the Proof of Concept stage, before they start to code anything they'll want to check whether panel has the buttons they want, the interactivity, the plotting capabilities and the **io** (xlsx in with the fileinput widget, xlsx out with the filedownload widget, great!)

Just to report my use case. I have an app that generates a dataframe. I could save that as a csv file, no problem. However I also wanted to trace which set of input parameters generated that data. A programmer would probably have saved that as a .json file or something similar. My users though will never open that accompanying .json file, they just don't know what that weird extension is. However they know that an Excel file can have several sheets, so the app exports a file with an ouput sheet and an input sheet.

After all, ``openpyxl`` and ``xlsxwriter`` seem to be pretty solid packages (3 millions download a month) and are certainly well tested with pandas. It seems to me that ``xlsxwriter`` even is a pure python package. I'll let you decide if you want to have one of them as an additional dependency to you examples.

philippjfr:
Merging for now, will clean up the header at a later stage.

PR #1248, opened 2020-04-09
VTK Synchronized

xavArtley:
supersede: https://github.com/holoviz/panel/pull/1070

Remaining tasks:
- [x] clean code (remove old serializer, refactor render function on js side, ...)
- [x] Split the old VTK pane in:
  - `VTKJS` (to load vtkjs files)
  - `VTKSynchronized` to handle `vtk`render window
- [x] to handle vtkScalarBarActor (colorbar https://github.com/holoviz/panel/pull/1270)
- [x] Documentation:
  - [x] split old VTK notebook
  - [x] update gallery interactive demo

Future tasks (in an other PR)
- [ ] discussion on the API we want to expose
- [ ] extend the serializer
  - to handle volumes (will need vtk-js modifications)
- [x] new Notebook to document capabilites of `VTKSynchronized` pane

@philippjfr  I think the PR is ready for review, everithing should almost work as before.
There are some small API breaks (for colorbars and mostly on javascript side)

I'll make more doc improvement in an other PR

ezgif com-video-to-gif (1)


panel/models/vtk/vtksynchronized.ts
```diff
@@ -0,0 +1,226 @@
...
+    PanelHTMLBoxView.prototype.connect_signals.call(this) //super.super.connec_signals
+    this.connect(this.model.properties.orientation_widget.change, () => {
+      this._orientation_widget_visibility(this.model.orientation_widget)
+    })
+    this.connect(this.model.properties.arrays.change, () => this._decode_arrays())
+    this.connect(this.model.properties.scene.change, () => {
+      this._plot()
+    })
+    this.connect(this.model.properties.one_time_reset.change, () => {
+      this._vtk_renwin.getRenderWindow().clearOneTimeUpdaters()
```
jourdain, line 160: You can now call that [method](https://github.com/Kitware/vtk-js/blob/master/Sources/Rendering/Misc/SynchronizableRenderWindow/vtkObjectManager.js#L538)

xavArtley:
> For the vtkLookupTable, I think you want to create `vtk.Rendering.Core.vtkColorTransferFunction` not the basic one that use the HueRange.
> 
> _Originally posted by @jourdain in https://github.com/holoviz/panel/pull/1070#issuecomment-611514817_
> 

Indeed it works. I thought vtkColortransferfunction was for volumes

image

panel/_package-lock.json
```diff
@@ -0,0 +1,359 @@
+{
```
philippjfr, line 1: Why check this in if regular old package-lock.json is now checked in?

xavArtley: actually I haven't succeed to use jszip and vtk in typescript for the moment
jszip I have an error with readable-stream
and vtkjs on importation.

philippjfr: I'm actually currently a -1 on bundling vtk.js with panel.js. Once bokeh allows multiple bundles in an extension we can add a panel-vtk.js bundle but for now it's too large to be included by default.

jourdain:
Python serializer:
I've updated our version [here](https://gitlab.kitware.com/vtk/vtk/-/merge_requests/6707). I still need to validate it inside PVW to make sure I didn't break anything.

In the mean time it should be pretty close to the one you had (except your last lut change). Could you test it? If we can I would like to have (if possible) the exact same file... ;-)

Thanks @xavArtley !

panel/_package-lock.json
xavArtley: this is the error I get when trying to add jszip and @types/jszip to node modules:
`Error: can't resolve 'stream' from 'C:\Users\XA235208\Workspace\PythonWorkspace\modules\panel\panel\node_modules\readable-stream\readable.js'`
may be @mattpap  has an idea to solve it

xavArtley:
> Python serializer:
> I've updated our version [here](https://gitlab.kitware.com/vtk/vtk/-/merge_requests/6707). I still need to validate it inside PVW to make sure I didn't break anything.
> 
> In the mean time it should be pretty close to the one you had (except your last lut change). Could you test it? If we can I would like to have (if possible) the exact same file... ;-)
> 
> Thanks @xavArtley !

@jourdain I think you should not keep this :
https://gitlab.kitware.com/vtk/vtk/-/merge_requests/6707/diffs#0ad8a8250f77efb2ca8213329f7c901a030a1477_314_494
in vtkjs there is no proper set function for the rotation matrix and it's just a (temporary?) workaround for the orientation not taking into account

jourdain:
I had to make some changes, but now it works as-is in my pvw legacy codes. @xavArtley  see if that works for you? 
https://gitlab.kitware.com/vtk/vtk/-/merge_requests/6707

xavArtley:
> Another, potentially lighter-weight option could be to subclass `vtkArchiver` in Python and reimplement `vtkPartitionedArchiver` and its compression routines entirely in Python. I have experience subclassing C++ classes in Python to be useable in C++ using [pybind11](https://pybind11.readthedocs.io/en/stable/), so it is definitely doable. This would provide the additional benefit of confining the temperamental logic of serializing a VTK scene into vtk-js to one place.

_Originally posted by @tjcorona in https://github.com/holoviz/panel/pull/1070#issuecomment-611493586_

I'll be interesting to learn how to do this this.
I have never tried to subclass a vtkclass
Do you have some reference code that I can inspired or any guidance to achieve it?

xavArtley:
@jourdain I had to modify the hashDataArray to use `md5 hexdigest` instead of digest otherwise I have some problems to decode zipped array on javascript side due to `/` special char

jourdain:
Thanks I've noticed and keep up with it. But I had to remove the numpy dependency for the LUT. But that could be achieved using a file on the side...

xavArtley:
I think I can remove numpy dependency if you want.
Was just more convenient to extract values of a vtkArray

jourdain:
If you want, to use the file from vtk 9.0 as-is then yes...

jourdain:
I'm ready to merge... So if you want to take the latest version of that file from my [MR](https://gitlab.kitware.com/vtk/vtk/-/merge_requests/6707) and add the LUT handling without numpy while making sure that everything works as expected for you that would be great.

With your latest edit/add-on, I'll trigger another test dashboard along with a pvw deployment to be sure...

And if everything is good to go, I'll merge it into VTK.

xavArtley:
> I'm ready to merge... So if you want to take the latest version of that file from my [MR](https://gitlab.kitware.com/vtk/vtk/-/merge_requests/6707) and add the LUT handling without numpy while making sure that everything works as expected for you that would be great.
> 
> With your latest edit/add-on, I'll trigger another test dashboard along with a pvw deployment to be sure...
> 
> And if everything is good to go, I'll merge it into VTK.

I have tested your last file it's seems working except for the mapper:
I have a vtkOpenGLPolyDataMapper but there is no alias for it in vtk.js:
https://github.com/Kitware/vtk-js/blob/c4459d8682107f241dec3c2e3d77f00a4d266d4e/Sources/Rendering/Misc/SynchronizableRenderWindow/vtkObjectManager.js#L413

jourdain:
I can add the mapping in vtkJS, it is easy enough. Do you want to do a PR?
I've already merged the py file in vtk/master+release but I can fix (remove) the unused import...

jourdain:
vtk 13.13.2 have the new mapping.

xavArtley:
@jourdain is there any sticking point to extent the synchronize renderer to handle vtkVolumes?

xavArtley:
Example of vtkActor animation and possibility to link and unlink camera beetween panels:
In the example after the first rotation of the both views shared camera and after the second rotation camera are unlinked
ezgif com-video-to-gif (3)

jourdain:
No specific reason for not handling volume, we just didn't had the need so far. 
In term of work we would just need new serializers on the Python side and the proper mapping on the client side. But as you may know by now, it is not too difficult or long to do.

xavArtley:
Thx good to know, I didn't wanted to start this and go in a dead end

xavArtley:
I think most of all fonctionalities of the old panel are functionals
ezgif com-video-to-gif (5)

Remaining tasks:
- [ ] clean code (remove old serializer, refactor render function on js side, ...)
- [ ] Split the old VTK pane in:
  - `VTKJS` (to load vtkjs files)
  - `VTKSynchronized` to handle `vtk`render window
- [ ] Documentation:
  - new Notebook to document capabilites of `VTKSynchronized` pane
  - update old VTK notebook
Future tasks (in an other PR)
- [ ] discussion on the API we want to expose
- [ ] extend the serializer
  - to handle volumes (will need vtk-js modifications)
  - to handle vtkScalarActor (colorbar)

xavArtley:
Je me heurte à un problème que je ne sais pas résoudre.
Il s'agit d'afficher plusieurs fois le même panel
Tant qu'on essaye pas d'afficher des éléments comme le widget d'orientation ou les axes, il n'y a pas de problème. Mais tous les éléments qui n'appartiennent pas à la fenetre de rendu côté python sont compliqués à gérer car la scène est juste mise à jour et non plus reconstruite de zéro.

jourdain:
Can you provide a concrete example of the issue? (feel free to write it in French if it is easier for you)

xavArtley:
Here is an example with the axes:
ezgif com-video-to-gif (6)
code is 
```python
import pyvista as pv
import panel as pn
pn.extension('vtk')

pl = pv.Plotter()
pl.add_mesh(pv.Cylinder())

geom_pane1 = pn.pane.vtk.vtk.VTKSynchronized(pl.ren_win, sizing_mode='stretch_width')
import numpy as np
axes = dict(
    origin = [-1,-1, -1],
    xticker = {'ticks': np.linspace(-1,1,5)},
    yticker = {'ticks': np.linspace(-1,1,5)},
    zticker = {'ticks': np.linspace(-1,1,5)},
    fontsize = 12,
    digits = 1,
    grid_opacity = 1,
    show_grid=True
)
geom_pane1.axes = axes
geom_pane1
```

jourdain:
My guess is that you are serializing some custom actors that don't have an equivalent on the JS side.
Can you share the json scene description for that example? Also do you know on what type of object it is trying to do a getWidth() on?

xavArtley:
"vtkOpenGLPixelSpaceCallbackMapper" actually it's objects that are added to the renderer only on the javascript side

xavArtley:
I found a way to solve the problem. I create a new context view on python side. 
I keep only one context on javascript side to allow camera share between views
ezgif com-video-to-gif (8)

philippjfr:
This is really, really great, thanks so much for your efforts here @xavArtley and thanks for helping out @jourdain. Just for my understanding what does the migration path here look like? Do we wait on a new vtk release? Do we ship both the old and new VTK models and switch them out depending on whether the appropriate VTK version is available or something else entirely?

xavArtley:
Actually this PR does not depend on any new version of vtk

tjcorona:
@xavArtley has removed any dependencies on the latest VTK. Once VTK v9.0 releases (within the next week or so), we may be able to use some of its features to lower the maintenance burden of Panel's VTK support.

jourdain:
@xavArtley is currently using his own Python module to extract the VTK scene. The new version of VTK will allow to rely on C++ for that part rather than Python. But normally switching from Python to C++ should not affect anything on the JS/Client side.

philippjfr:
Thanks for all the clarifications. I guess my main question now is whether we really need separate VTK and VTKSynchronized panes? From a user perspective it seems fairly confusing and Synchronized isn't really to informative.

xavArtley:
The synchronize mean the state between the vtkRenderWindow on python (server) is synchronized with the one on javascript side (client).

xavArtley:
VTK is too generic for me.

philippjfr:
>The synchronize mean the state between the vtkRenderWindow on python (server) is synchronized with the one on javascript side (client).

I gathered that much, I'm just saying that I would expect all panel objects to synchronize with the underlying Python object if possible (indeed most do), so the information value in appending that qualifier is pretty low. I'm just trying to understand if or why multiple different VTK objects have to exist. @jbednar Any opinions?

xavArtley:
We can rename it VTKRenderWindow

jbednar:
I'm confused as well! In Panel essentially every pane is just named for what the underlying object is (`pn.pane.Markdown`, `pn.pane.LaTeX`, etc.) and we would hope that every pane either already synchronizes or eventually will, so there would have to be a very good reason to name it anything other than `pn.pane.VTK` for a VTK object.  

Is there a distinction that needs to be made between this new object and the previous `pn.pane.VTK` object?  It sounds like Philipp is proposing replacing `pn.pane.VTK` with this new object, which seems like the right thing to do if this new object is a strict superset of the capabilities of the previous `pn.pane.VTK` object.  But if there are legitimate uses for having them be separate objects, then it would be good to list precisely what the two different objects can do so that we can choose appropriate names for them.

xavArtley:
In this case the panel VTKVolume should be VTK since it can render a VTK
object too

Le lun. 20 avr. 2020 à 22:28, James A. Bednar <notifications@github.com> a
écrit :

tjcorona:
> In this case the panel VTKVolume should be VTK since it can render a VTK object too

There are currently only three types of VTK windows (this one, VTKVolume and VTKjs), right? Why not have them all captured by a VTK class and differentiate between them at initialization by their arguments?

xavArtley:
I'm ok to rename VTKSynchronized to VTK but I wan't to keep VTKJS separate since it allow just to render a vtkjs file

xavArtley:
> > In this case the panel VTKVolume should be VTK since it can render a VTK object too
> 
> There are currently only three types of VTK windows (this one, VTKVolume and VTKjs), right? Why not have them all captured by a VTK class and differentiate between them at initialization by their arguments?

Since it does not expose the same fonctionalities at all I think we should keep different classes

xavArtley:
If I rename in VTK the panel, can I keep the typescript model as VTKSynchronizedPlot or should I rename it too

xavArtley:
Hello @jourdain ,
I began to work on volume serialization
For the moment I serialize juste the vtkvolume, the vtkvolumemapper and the vtkvolumeproperty
I register the deserializers on javascript side:
```
vtkObjectManager.setTypeMapping(
    'vtkVolume',
    vtkns['Volume'].newInstance,
  )
  vtkObjectManager.setTypeMapping(
    'vtkVolumeMapper',
    vtkns['VolumeMapper'].newInstance,
  )
  vtkObjectManager.setTypeMapping(
    'vtkVolumeProperty',
    vtkns['VolumeProperty'].newInstance,
  )
```

However I'm hurting to a bug I don't know how to solve
Here is a gif of the bug:
ezgif com-video-to-gif (9)

and the error:
```
index.js:1093 Uncaught (in promise) TypeError: Cannot read property 'getNumberOfComponents' of null
    at Object.e.buildBufferObjects (VM3331 vtk.js:282)
    at Object.e.updateBufferObjects (VM3331 vtk.js:282)
    at Object.e.renderPieceStart (VM3331 vtk.js:282)
    at Object.e.renderPiece (VM3331 vtk.js:282)
    at e.volumePass (VM3331 vtk.js:282)
    at Object.e.apply (VM3331 vtk.js:282)
    at Object.e.traverse (VM3331 vtk.js:282)
    at e.traverseVolumePass (VM3331 vtk.js:282)
    at Object.e.traverse (VM3331 vtk.js:282)
    at Object.e.traverse (VM3331 vtk.js:282)
```

if you want the scene serialized you can download it here:
https://drive.google.com/open?id=1lJmcrJ6wQArUNLMosLfnQWYELar9fJC4

jourdain:
The serialization of the vtkImageData seems wrong. We should have an `extent` but we have a `dimensions`. It does not look like the serialization I did for the input of the texture.

xavArtley:
> The serialization of the vtkImageData seems wrong. We should have an `extent` but we have a `dimensions`. It does not look like the serialization I did for the input of the texture.

I use the same code as in VTK source:
https://gitlab.kitware.com/vtk/vtk/-/blob/master/Web/Python/vtkmodules/web/render_window_serializer.py#L747-773

jourdain:
I guess that's should be fine then. As long as the setExtent get called internally which I may have fixed in vtk.js for the texture issue. 

Is it possible that you call render before the vtkImageData get fully created/filled and linked to the mapper?

jourdain:
Are we missing the piecewise function and lookup table?

jourdain:
Not related but I would skip the lights and let vtk.js create them.

xavArtley:
> I guess that's should be fine then. As long as the setExtent get called internally which I may have fixed in vtk.js for the texture issue.
> 
> Is it possible that you call render before the vtkImageData get fully created/filled and linked to the mapper?

I supposes it can be that but I have no idea how to correct it.


> Are we missing the piecewise function and lookup table?

For the moment I skipped this just to make tests quicker. The only serializer missing is the vtkPieceWise but it should not complcated to implement

xavArtley:
I will continue to post here to track my debugging steps and if someone find the trick before me.
> > Is it possible that you call render before the vtkImageData get fully created/filled and linked to the mapper?
> 
> I supposes it can be that but I have no idea how to correct it.
> 
Now I think this assumption is not the bug
I stopped in the chrome debugger just after the call to the synchronize function and I store the vtkVolume instance a global variable to inspect it
I tried to call the `getState` method on the vtkObjects here is the result:
image

My first thought is the bug is related to the fact I didn't serialize color and scalar opacity

xavArtley:
The error is indeed due to the call to render before the vtkImageData get fully created/filled
image

The error occur in this part of the code:
https://github.com/Kitware/vtk-js/blob/d53c83462fa6de08dfefad5642ca7a70acd7495f/Sources/Rendering/OpenGL/VolumeMapper/index.js#L1069-L1094

would make it sense to add a check if scalars are not null in vtk-js code?

xavArtley:
I have tested by adding this code to vtk-js
image

It seems to work:
ezgif com-video-to-gif (10)

However I'm not sur it's the best correction

jourdain:
We could provide some guard in vtk.js but I wanted to double check something with you first.
Are you only registering a render on camera modified once? Or each time you synch the view? I'm not sure to know why you have that callback as the synch call should indeed call render once done.

I'll have to check but maybe we can make sure we do not connect the actor until all its dependencies have been resolved. That way no issue if we have random render calls during a synch.

xavArtley:
I think I made a mistake in my second to last commit.
The correct behavior is I keep track of each camera callback added
I removed it before synchronization and add a new one at the end

xavArtley:
image

jourdain:
Do you still need the guard in vtk.js for the scalars?

xavArtley:
> Do you still need the guard in vtk.js for the scalars?

yes

xavArtley:
> Do you still need the guard in vtk.js for the scalars?

Finally I think I found a way by storing promises and :
```javascript
this._vtk_renwin.getRenderWindow().synchronize(this.model.scene)
    
    Promise.all(this._promises).then(() => {
      this._promises = []
      this.model.arrays_processed = [...this.model.arrays_processed]
      this._rendering_ok = true
      this._camera_callbacks.push(
        this._vtk_renwin
          .getRenderer()
          .getActiveCamera()
          .onModified(() => this._vtk_render())
      )
      this._vtk_renwin.resize()
      this._vtk_render()
      if (!this._orientationWidget)
        this._create_orientation_widget()
      if (!this._axes)
        this._set_axes()
    })
```

But I'm not 100% confident with this

jourdain:
It should be fine unless for what ever reason the scene report an array that will never resolve. In such case you will be waiting for ever.

xavArtley:
@jourdain 
hello, I'm working on exporting and importing the scene using a `.synch`  file (radom choice for the extension name). 
image

However to make things easier I change the way to serialize vtkLookupTable using a slight modification in the previous serializer, I export the table:
```python
    table = getArrayDescription(lookupTable.GetTable(), context)
```

However in javascript side I had to implement a specific lookupTableUpdater:
```typescript

function updateLookupTable(instance: any, state: any, context: any) {
  context.start()
  instance.set(state.properties)

  if (state.table) {
    const table: any = state.table
    context.getArray(table.hash, table.dataType, context).then(
      (values: any) => {
        const array = vtkns.DataArray.newInstance(
          Object.assign({values}, table)
        )
        instance.setTable(array)
        instance.modified()
        context.end()
      },
      (error: any) => {
        console.log("error fetching array", error)
      }
    )
  }
}
```
Do you think I can add it to vtkjs? 
I'm not sure if its induce some breaking change

jourdain:
Your change suggestion seems reasonable and we could update vtk.js to handle such LUT update.
But, I'm wondering if we should try to move more stuff to a more generic handling rather than using a custom `table` key for its state. I know I started the custom path for datasets too, but I'm wondering if we can achieve even generic handling for everything. 

I'm not asking you to deal with the datasets part but maybe, your work could be the seed for some refactoring (from my side when/if I get time to work on that).

Let me explain what I'm thinking. Right now we have the following structure:

```
{
  id: '234',
  type: 'vtkXXX',
  parent: '225',
  properties: { ... },
  dependencies: [ ... ],
  calls: [ ... ],
}
```

Via the dependencies and calls we can bind various vtkObjects together is a very generic ways. Currently the vtkDatasets have some custom handling because they are dealing with vtkDataArrays that require some binary fetching. I thought they were the only vtkObject dealing with vtkDataArrays and therefore, I didn't mind doing one custom handling for them. But now you are showing me that vtkLookupTable need the same thing. So maybe let's allow vtkDataArray to be handled in a generic manner too. Let's add another reserved key for `arrays` and we will use their registration field to bind them back to the instance.

So we should have something like that: 

```
{
  id: '234',
  type: 'vtkXXX',
  parent: '225',
  properties: { ... },
  dependencies: [ ... ],
  calls: [ ... ],
  arrays: [
      { hash, dataType, ..., registration: 'setTable' },
      ...,
  ]
}
```

So in the genericUpdate if `state.arrays` exist we process it kind of like you did but in a more generic fashion.

Would that sounds reasonable? Can you do a PR toward that direction? (or should I?)

xavArtley:
I'll be happy to increase genericity of the serializers.
I can try to work on it this week end since it's a 3 days off (at least for me)
The hard part on my side is, I have not a global picture of all possible breaking change it can induce in other libraries built on the synchonizable rendering.

xavArtley:
Most fetures on this PR are now implemented.
If we want to change the serializer later (I'll had a method to register a serializer, it's not a big work)
the only requirement is the possibility to generate unique ids for the vtk objects in the scene depending of the context.

I added this method to the context (the test on the vtkCamera is because it's the only object I want to share accross context to link model views)
```
    def getReferenceId(self, instance):
        if not self.idRoot or (hasattr(instance, 'IsA') and instance.IsA('vtkCamera')):
            return getReferenceId(instance)
        else:
            return self.idRoot + getReferenceId(instance)
```

philippjfr:
@xavArtley What's your feeling? Should I merge it as is or wait until you've been able to implement a pluggable serializer?

xavArtley:
I don't think we should wait for the pluggable serializer.
I'll do it in another PR but later.

philippjfr:
@xavArtley So I've now tried this PR, first of all I encountered a few errors running the examples, e.g.

```
index.js:1298 Uncaught TypeError: Cannot read property 'offset' of null
    at Object.e.setMapperShaderParameters (index.js:1298)
    at Object.e.updateShaders (index.js:1097)
    at Object.e.renderPieceDraw (index.js:1610)
    at Object.e.renderPiece (index.js:1687)
    at Object.e.render (index.js:89)
    at e.opaqueZBufferPass (index.js:68)
    at Object.NA.e.apply (index.js:44)
    at Object.NA.e.traverse (index.js:32)
    at Object.e.traverseOpaquePass (index.js:63)
    at e.traverseOpaqueZBufferPass (index.js:47)
```

```
index.js:49 Uncaught (in promise) TypeError: Cannot read property 'traverse' of undefined
    at FC.e.traverseVolumePass (index.js:49)
    at Object.NA.e.traverse (index.js:27)
    at Object.NA.e.traverse (index.js:35)
    at Object.e.traverse (index.js:83)
    at Object.e.traverseAllPasses (index.js:1066)
    at c (index.js:331)
    at Object.e.render (index.js:722)
    at Object.Ub.e.render (index.js:57)
    at Object.<anonymous> (index.js:257)
    at a (macro.js:825)
```

If these are to do with wrong versions on my end it would be good to state precisely what versions of various libraries are needed.

Secondly I'm not entirely convinced we want separate classes for the old and new VTK panes. My suggestions would be to make `VTK` a baseclass with a `__new__` method, which creates the appropriate subclass, which can be either `VTKJS` or `VTKRenderWindow`, does that sound reasonable? (Cc: @jbednar)

xavArtley:
> @xavArtley So I've now tried this PR, first of all I encountered a few errors running the examples, e.g.
> 
> ```
> index.js:1298 Uncaught TypeError: Cannot read property 'offset' of null
>     at Object.e.setMapperShaderParameters (index.js:1298)
>     at Object.e.updateShaders (index.js:1097)
>     at Object.e.renderPieceDraw (index.js:1610)
>     at Object.e.renderPiece (index.js:1687)
>     at Object.e.render (index.js:89)
>     at e.opaqueZBufferPass (index.js:68)
>     at Object.NA.e.apply (index.js:44)
>     at Object.NA.e.traverse (index.js:32)
>     at Object.e.traverseOpaquePass (index.js:63)
>     at e.traverseOpaqueZBufferPass (index.js:47)
> ```
> 
> ```
> index.js:49 Uncaught (in promise) TypeError: Cannot read property 'traverse' of undefined
>     at FC.e.traverseVolumePass (index.js:49)
>     at Object.NA.e.traverse (index.js:27)
>     at Object.NA.e.traverse (index.js:35)
>     at Object.e.traverse (index.js:83)
>     at Object.e.traverseAllPasses (index.js:1066)
>     at c (index.js:331)
>     at Object.e.render (index.js:722)
>     at Object.Ub.e.render (index.js:57)
>     at Object.<anonymous> (index.js:257)
>     at a (macro.js:825)
> ```
> 
> If these are to do with wrong versions on my end it would be good to state precisely what versions of various libraries are needed.
> 
> Secondly I'm not entirely convinced we want separate classes for the old and new VTK panes. My suggestions would be to make `VTK` a baseclass with a `__new__` method, which creates the appropriate subclass, which can be either `VTKJS` or `VTKRenderWindow`, does that sound reasonable? (Cc: @jbednar)

Can you retry with the last commit?
I set the version of vtkjs to https://unpkg.com/vtk.js@13.18.0/dist/vtk.js
There must have been some changes that makes it does not work anymore

philippjfr:
The offset error is now fixed but the traverse error is still there.

xavArtley:
I can't reproduce the traverse error

jbednar:
> My suggestions would be to make VTK a baseclass with a __new__ method, which creates the appropriate subclass, which can be either VTKJS or VTKRenderWindow, does that sound reasonable?

I very much hope this can be done, because otherwise users have to learn which object to select, a much more difficult task than learning which option on an object to select.  Figuring out options with `help(obj)` is easy once you have an `obj`; figuring out which object to use out of all the objects in Panel is difficult, and once they start with one of them many users won't realize that others are available.

xavArtley:
I have implemented it in last commit

Le mer. 13 mai 2020 à 16:29, James A. Bednar <notifications@github.com> a
écrit :

xavArtley:
@philippjfr I reproduced on linux

examples/reference/panes/VTK.ipynb
```diff
@@ -263,33 +148,119 @@
...
+    "filename = geo_pan_pv.export_scene(filename='geo_example')\n",
+    "print(\"The file is exported in :\", os.path.join(os.getcwd(), filename))\n",
+    "pn.pane.HTML('<iframe width=100% height=400 src=https://kitware.github.io/vtk-js/examples/SynchronizableRenderWindow/index.html></iframe>', sizing_mode='stretch_width')"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## Advance usage and interactivity\n",
```
jbednar, line 248:
```suggestion
    "## Advanced usage and interactivity\n",
```

examples/reference/panes/VTK.ipynb
```diff
@@ -263,33 +148,119 @@
...
     "\n",
-    "If no colorbar is found in the scene the method return `None`."
+    "This examples will show :\n",
+    " - where information about `scalars` used to color 3d objects can be found in the scene\n",
+    " - utilisation of the `orientation widget`\n",
+    " - how to add `axes` in the scene\n",
+    " - collaboration with the `player` widget to create an animation"
```
jbednar, line 254:
```suggestion
    " - where information about `scalars` used to color 3D objects can be found in the scene\n",
    " - utilization of the `orientation widget`\n",
    " - how to add `axes` in the scene\n",
    " - adding a `player` widget to create an animation"
```

examples/reference/panes/VTK.ipynb
```diff
@@ -263,33 +148,119 @@
...
    ]
   },
   {
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "We present an example of the use of `construct_colorbars`. Here we use `pyvista` module which simplifies the construction of VTK scenes"
+    "### Scene creation\n",
+    "\n",
+    "First we create scene with a nice spline inspired by this examples : [Creating a Spline](https://docs.pyvista.org/examples/00-load/create-spline.html#creating-a-spline)\n"
```
jbednar, line 263:
```suggestion
    "First we create a scene with a nice spline inspired by this examples : [Creating a Spline](https://docs.pyvista.org/examples/00-load/create-spline.html#creating-a-spline)\n"
```

examples/reference/panes/VTK.ipynb
```diff
@@ -325,24 +296,36 @@
...
+    "spline_pan = pn.panel(pl.ren_win, width=500, orientation_widget=True)\n",
+    "spline_pan"
    ]
   },
   {
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "## Adding axes to the VTK scene\n",
+    "By clicking on the three dots colobar can be displayed\n",
```
jbednar, line 313:
```suggestion
    "By clicking on the three dots, a colorbar can be displayed\n",
```

examples/reference/panes/VTK.ipynb
```diff
@@ -325,24 +296,36 @@
...
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "### Adding axes to the VTK scene\n",
+    "\n",
+    "Using the `axes` parameter, it is possible to add an axes into the 3d view.\n",
+    "\n",
+    "`Axes` are a like `orientation widget` and `colorbar`, a specific object built on javascript side.\n",
+    "\n",
+    "So they will not be synchronized with the python `vtkRenderWindow`"
```
jbednar, line 328:
```suggestion
    "The orientation widget on the bottom right of the panel is clickable too, allowing you to orient the camera in a specific direction."
   ]
  },
  {
   "cell_type": "markdown",
   "metadata": {},
   "source": [
    "### Adding axes to the VTK scene\n",
    "\n",
    "Using the `axes` parameter, it is possible to display axes in the 3D view.\n",
    "\n",
    "The `Axes` object (like `orientation widget` and `colorbar`) is a client-side object instantiated only on the JavaScript side, and is not synchronized back to the Python `vtkRenderWindow`."
```

examples/reference/panes/VTK.ipynb
```diff
@@ -361,14 +344,114 @@
...
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "### Working with Volumes\n",
+    "\n",
+    "All previous pane are examples using geometric examples.\n",
+    "\n",
+    "However `VTK` can work with both geomtric and volume objects"
```
jbednar, line 358:
```suggestion
    "All previous panes use geometric examples, but `VTK` can work with both geometric and volume objects:"
```

examples/reference/panes/VTK.ipynb
```diff
@@ -361,14 +344,114 @@
...
+    "volume_pan"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "As for other examples we can add easily new objects in the pane\n",
+    "\n",
+    "For examples we can create 3 orthogonal slices and add them to the scene"
```
jbednar, line 388:
```suggestion
    "As for other examples, we can add easily new objects in the pane.\n",
    "\n",
    "For example, we can create 3 orthogonal slices and add them to the scene:"
```

examples/reference/panes/VTK.ipynb
```diff
@@ -361,14 +344,114 @@
...
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "### Create an animation\n",
+    "\n",
+    "We will reuse the panel used for the spline\n",
+    "\n",
+    "For this animation we will rotate the actor and roll the scalars on the spline tube"
```
jbednar, line 409:
```suggestion
    "We will reuse the panel used for the spline.\n",
    "\n",
    "For this animation, we will rotate the actor and roll the scalars on the spline tube:"
```

panel/pane/vtk/vtk.py
```diff
@@ -28,16 +28,324 @@
...
+
+        Must contain at least ``xticker``, ``yticker`` and ``zticker``.
+
+        A ``ticker`` is a dictionary which contains:
+          - ``ticks`` (array of numbers) - required.
+              Positions in the scene coordinates of the corresponding
+              axis' ticks.
+          - ``labels`` (array of strings) - optional.
+              Label displayed respectively to the `ticks` positions.
+              If `labels` are not defined they are infered from the
```
jbednar, line 50:
```suggestion
              If `labels` are not defined they are inferred from the
```

panel/pane/vtk/vtk.py
```diff
@@ -287,78 +592,31 @@ def _subsample_array(self, array):
         return sub_array
 
 
-class VTK(PaneBase):
+class VTKJS(AbstractVTK):
     """
-    VTK panes allow rendering VTK objects.
+    VTK panes allow rendering vtk scene stored in a vtkjs.
```
jbednar, line 597:
```suggestion
    VTK panes allow rendering a VTK scene using vtkjs.
```

panel/models/vtk.py
```diff
@@ -26,99 +50,86 @@ class AbstractVTKPlot(HTMLBox):
...
 
+    scene = Dict(String, Any, help="""The serialized vtk.js scene on json format""")
 
 
-class VTKPlot(AbstractVTKPlot):
+class VTKJSPlot(AbstractVTKPlot):
     """
-    Bokeh model dedicated to plot a vtk render window with only 3D geometry objects
-    (Volumes are not suported)
+    Bokeh model dedicated to plot 3d scene saved under the `.vtk-js` format
```
jbednar, line 88:
```suggestion
    Bokeh model for plotting a 3D scene saved in the `.vtk-js` format
```

panel/models/vtk.py
```diff
@@ -26,99 +50,86 @@ class AbstractVTKPlot(HTMLBox):
...
 
-    color_mappers = List(Instance(ColorMapper))
+    width = Override(default=300)
 
 
-class VTKAxes(Model):
+class VTKSynchronizedPlot(AbstractVTKPlot):
     """
-    A Bokeh model for axes
+    Bokeh model dedicated to plot a vtk render window
```
jbednar, line 72:
```suggestion
    Bokeh model for plotting a VTK render window
```

examples/reference/panes/VTKJS.ipynb
```diff
@@ -0,0 +1,216 @@
...
+    "  - s: set representation of all actors to *surface*\n",
+    "  - w: set representation of all actors to *wireframe*\n",
+    "  - v: set representation of all actors to *vertex*\n",
+    "  - r: center the actors and move the camera so that all actors are visible\n",
+    "  \n",
+    "  The mouse must be over the pane to work\n",
+    "  <br>**Warning**: These keybindings may not work as expected in a notebook context, if they interact with already bound keys\n",
+    "* **``orientation_widget``** (bool): A boolean to activate/deactivate the orientation widget in the 3D pane. This widget is clickable and allows to rotate the scene in one of the orthographic projections.\n",
+    "* **``object``** (str or object): Can be a string pointing to a local or remote file with a `.vtkjs` extension.\n",
+    "___"
```
jbednar, line 48:
```suggestion
    "The ``VTK`` pane renders vtk.js files inside a panel, making it possible to load and interact with complex geometries in 3D.\n",
    "\n",
    "#### Parameters:\n",
    "\n",
    "For layout and styling related parameters see the [customization user guide](../../user_guide/Customization.ipynb).\n",
    "\n",
    "* **``axes``** (dict): A dictionary with the parameters of the axes to construct in the 3D view.\n",
    "    Must contain at least ``xticker``, ``yticker`` and ``zticker``.\n",
    "    - ``[x|y|z]ticker`` is a dictionary which contains:\n",
    "        - ``ticks`` (array of numbers) - required. Positions in the scene coordinates\n",
    "        of the corresponding axis ticks\n",
    "        - ``labels`` (array of strings) - optional. Label displayed respectively to\n",
    "        the `ticks` positions.\n",
    "\n",
    "        If `labels` are not defined they are inferred from the `ticks` array.\n",
    "    - ``digits``: number of decimal digits when `ticks` are converted to `labels`.\n",
    "    - ``fontsize``: size in pts of the ticks labels.\n",
    "    - ``show_grid``: boolean. If true (default) the axes grid is visible.\n",
    "    - ``grid_opacity``: float between 0-1. Defines the grid opacity.\n",
    "    - ``axes_opacity``: float between 0-1. Defines the axes lines opacity.\n",
    "* **``camera``** (dict): A dictionary reflecting the current state of the VTK camera\n",
    "* **``enable_keybindings``** (bool): A boolean to activate/deactivate keybindings. Bound keys are:\n",
    "  - s: set representation of all actors to *surface*\n",
    "  - w: set representation of all actors to *wireframe*\n",
    "  - v: set representation of all actors to *vertex*\n",
    "  - r: center the actors and move the camera so that all actors are visible\n",
    "  \n",
    "  The mouse must be over the pane to work.\n",
    "  <br>**Warning**: These keybindings may not work as expected in a notebook context, if they interact with already bound keys.\n",
    "* **``orientation_widget``** (bool): A boolean to activate/deactivate the orientation widget in the 3D pane. This widget is clickable and allows to rotate the scene in one of the orthographic projections.\n",
    "* **``object``** (str or object): Can be a string pointing to a local or remote file with a `.vtkjs` extension.\n",
    "___"
```

Not sure what those ending underscores do.

examples/reference/panes/VTKJS.ipynb
```diff
@@ -0,0 +1,216 @@
...
+    "vtk_pane.object = \"https://github.com/Kitware/vtk-js-datasets/raw/master/data/vtkjs/TBarAssembly.vtkjs\""
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "### Controls\n",
+    "\n",
+    "The ``VTK`` pane exposes a number of options which can be changed from both Python and Javascript. Try out the effect of these parameters interactively:"
```
jbednar, line 91:
```suggestion
    "The ``VTK`` pane exposes a number of options which can be changed from both Python and JavaScript. Try out the effect of these parameters interactively:"
```

examples/reference/panes/VTKJS.ipynb
```diff
@@ -0,0 +1,216 @@
...
+    "pn.Row(vtk_pane.controls(jslink=True), vtk_pane)"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## Camera control\n",
+    "\n",
+    "Once a VTKJS pane has been displayed it will automatically sync the camera state with the Pane object.\n",
```
jbednar, line 109:
```suggestion
    "Once a VTK pane has been displayed it will automatically sync the camera state with the Pane object.\n",
```

?

examples/reference/panes/VTKJS.ipynb
```diff
@@ -0,0 +1,216 @@
...
+    "The camera parameter is updated only at the end of the interaction\n",
```
jbednar, line 110:
```suggestion
    "The camera parameter is updated only at the end of the interaction.\n",
```

examples/reference/panes/VTKJS.ipynb
xavArtley: the ending underscore are for the horizontal separator;
image

jbednar: Looks nice!

examples/reference/panes/VTK.ipynb
```diff
@@ -44,144 +47,28 @@
...
+    "* **``set_background(r: float, g: float, b: float)``** : Set the color of the scene background as rgb color where `r`, `g`, and `b` are float in (0, 1)\n",
+    "\n",
+    "* **``synchronize()``** : Sychronize the modifications of the ``vtkRenderWindow``\n",
+    "\n",
+    "* **``unlink_camera()``** : Create a new vtkCamera object, allowing to the panel to have it's own camera.\n",
+    "\n",
+    "* **``link_camera(other: VTK)``** : Both panel will share the same camera, any change in one panel camera will be applied in the link panel\n",
+    "\n",
+    "**Warning** for applying any modification generated by these methods, the `sychronize()` method must be called\n",
+    "___"
```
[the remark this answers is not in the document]
xavArtley: It's for the horizontal separator
image

xavArtley:
> Great stuff! I've done a good bit of editing of the notebook text, and I _hope_ none of that breaks the JSON file format. It _shouldn't_, but it's worth loading the notebook into Jupyter after applying the changes just to make sure.

I have tested nothing breaks

xavArtley:
@philippjfr it should work now

philippjfr:
>@philippjfr it should work now

Can confirm, works now.

philippjfr:
Okay, I have one last question, what happened to `VTK.construct_colorbars()`? I take it this now happens client-side but if we want to remove it we should go through an actual deprecation cycle, i.e. start warning in this release and remove it in the next release.

philippjfr:
Apart from that this is all looks great, thanks for all your efforts here, they are hugely appreciated!

panel/pane/vtk/vtk.py
```diff
@@ -134,6 +134,29 @@ def get_color_mappers(self):
                     LinearColorMapper(low=low, high=high, name=name, palette=palette)
                 )
         return cmaps
+    
+    def construct_colorbars(self, orientation='horizontal'):
```
philippjfr, line 138: Did you want to deprecate this? I have no opinion but if you think it's a good idea we should do it now and add a warning using `self.param.warning()` which explains it's deprecated and what the replacement is.

xavArtley:
@philippjfr , @jbednar 
I have a concern that emerged this week,
it concerns the moment when the serialization of the object (vtkRenderWindow) takes place.
Currently, it only happens when _get_model or _update is called
Formerly, I had the serialize_on_instantiation parameter which allowed serializing the construction of the object.
This made it possible to avoid a problem in pyvista because the VTK pane was instantiated then the VTKRenderWindow was cleaned before displaying.
If serialize_on_instantiation was False, an empty pane was displayed. The solution was therefore to serialize at instantiation.
In this case we are no longer really on a synchronized pane, so I was wondering if I should not do two classes:
VTKRenderWindowSynchronized and VTKRenderWindow, where the switch was made depending on whether we are initializing on instantiation or not.

panel/pane/vtk/vtk.py
xavArtley: No need to deprecate it for the moment.

philippjfr:
>VTKRenderWindowSynchronized and VTKRenderWindow, where the switch was made depending on whether we are serializing on instantiation or not.

Is this still needed or did your latest changes address this?

xavArtley:
My last changed adress it

xavArtley:
Hello @jourdain,

I need your help on an issue.
When trying to solve a panel issue I encounter another issue with the convertion of scalars to colors.

This issue occur when I set the mapper to use the lookup table range `mapper.UseLookupTableScalarRangeOn()`
This is the result I obtained:
image

However if I set `UseLookupTableScalarRangeOff`, I get similar results as with VTK C++
image

You can download the zip file of the scene here:
https://drive.google.com/drive/folders/1kgxkS3VyvcZFcTuBL2ty_cTtdUr37wmg?usp=sharing
you can use it in https://kitware.github.io/vtk-js/examples/SynchronizableRenderWindow.html

The scene piece is a composition of 32 actors

jourdain:
Glad you figure it out... Sorry I could not provide much help. I'm off till the end of the week.

PR #1252, opened 2020-04-09
Adding a WebComponent model and component

MarcSkovMadsen:
As agreed I've started splitting the big Web Component PR into smaller PRs.

See https://github.com/holoviz/panel/pull/1122


MarcSkovMadsen:
Hi @philippjfr 

I see only one remaining issue to solve before this could be reviewed and published.

The problem is that the ECharts/ PyCharts example does not work in the notebook. It raises a `Bokeh: ERROR: Unable to run BokehJS code because BokehJS library is missing`.

image

### Reproduce

run the `examples/reference/panes/WebComponent.ipynb` notebook in this branch

or the code below in this branch

```python
import param
import panel as pn
pn.extension()
pn.config.sizing_mode="stretch_width"
from IPython.core.display import display, HTML
display(HTML("<style>.container { width:100% !important; }</style>"))


ECHARTS_JS = """
<script src="https://cdn.bootcss.com/echarts/3.7.2/echarts.min.js"></script>
"""

echarts_js_pane = pn.pane.HTML(ECHARTS_JS)

ECHARTS_HTML = """
    <div class="echart" style="width:100%;height:100%;"></div>
    <script type="text/javascript">
        var myScript = document.currentScript;
        var myDiv = myScript.parentElement.firstElementChild;
        var myChart = echarts.init(myDiv);
        myDiv.eChart = myChart;
        Object.defineProperty(myDiv, 'option', {
           get: function() { return null; },
           set: function(val) { this.eChart.setOption(val); this.eChart.resize();}
        });
        myDiv.after_layout = myChart.resize; // Resizes the chart after layout of parent element
    </script>"""

class EChart(pn.pane.WebComponent):
    html = param.String(ECHARTS_HTML)
    properties_to_watch = param.Dict({"option": "option"})

    echart = param.Parameter()
    option = param.Dict()

    def __init__(self, **params):
        if "echart" in params:
            params["option"] = self._to_echart_dict(params["echart"])
        super().__init__(**params)

    @classmethod
    def _to_echart_dict(cls, echart):
        if isinstance(echart, dict):
            return echart
        if 'pyecharts' in sys.modules:
            import pyecharts
            if isinstance(echart, pyecharts.charts.chart.Chart):
                return json.loads(echart.dump_options())

        return {}

    @param.depends("echart", watch=True)
    def update(self):
        self.option = self._to_echart_dict(self.echart)

pn.config.js_files["echart"]="https://cdn.bootcss.com/echarts/3.7.2/echarts.min.js"
ECHART_DICT = {
    "title": {"text": "ECharts entry example"},
    "tooltip": {},
    "legend": {"data": ["Sales"]},
    "xAxis": {"data": ["shirt", "cardign", "chiffon shirt", "pants", "heels", "socks"]},
    "yAxis": {},
    "series": [{"name": "Sales", "type": "bar", "data": [5, 20, 36, 10, 10, 20]}],
}
app = pn.Column(
    EChart(echart=ECHART_DICT, height=500),
)
app
```

### Error

```javascript
","margin":[5,10,5,10],"name":"","propertiesLastChange":{"option":{"legend":{"data":["Sales"]},"series":[{"data":[5,20,36,10,10,20],"name":"Sales","type":"bar"}],"title":{"text":"ECharts entry example"},"tooltip":{},"xAxis":{"data":["shirt","cardign","chiffon shirt","pants","heels","socks"]},"yAxis":{}}},"propertiesToWatch":{"option":"option"},"sizing_mode":"stretch_width"},"id":"1469","type":"panel.models.web_component.WebComponent"},{"attributes":{"client_comm_id":"83fd76e3550942b79a916c3c6473d429","comm_id":"ce9ba2f972204a4cbd41cb12fc6ac1e4","plot_id":"1468"},"id":"1470","type":"panel.models.comm_manager.CommManager"}],"root_ids":["1468","1470"]},"title":"Bokeh Application","version":"2.0.1"}}; var render_items = [{"docid":"4e17e63a-dda4-43f7-8857-6666494999aa","root_ids":["1468"],"roots":{"1468":"c50dc812-ff9a-4cd2-b52c-ef57f48fe819"}}]; root.Bokeh.embed.embed_items_notebook(docs_json, render_items); } if (root.Bokeh !== undefined) { embed_document(root); } else { var attempts = 0; var timer = setInterval(function(root) { if (root.Bokeh !== undefined) { clearInterval(timer); embed_document(root); } else if (document.readyState == "complete") { attempts++; if (attempts > 100) { clearInterval(timer); console.log("Bokeh: ERROR: Unable to run BokehJS code because BokehJS library is missing"); } } }, 10, root) } })(window);
```

### Additional Context

It works on the server using `app.show()`.

image

philippjfr:
@MarcSkovMadsen Thanks for splitting this PR out! I'll try to review tomorrow and can try to resolve remaining issues. Just to be an annoying nag, please do clear out notebooks before committing them.

MarcSkovMadsen:
Hi @philippjfr 

Should I do something else than shown below?  Because I have really been trying to clean my notebooks before committing?

image

philippjfr:
That's perfectly fine for now. We have a script to clear out all the metadata as well but it only works on *nix systems and I can do that later.

MarcSkovMadsen:
And @philippjfr regarding Echarts. I really think we should consider doing a proper bokeh model on that. We don't need to implement a like of functionality from client to server. Just being a being able to display this charts would be a big win for some people and for Panel.

These charts are different the all the other libraries already in. They are super fast, responsive and have transitions by default.

philippjfr:
>Bokeh: ERROR: Unable to run BokehJS code because BokehJS library is missing.

One of the JS dependencies must be failing to import or run. No other errors?

philippjfr:
Okay after having played around a bit I am very excited for this. One immediate comment I had is that the WebComponents reference guide should probably focus less on showing off all the possible libraries and instead focus in-depth on one particular example and explain how to wrap that component in a nice way. All the individual examples, e.g. echarts, model-viewer etc. can then become gallery notebooks in the external sections as you have already started to do.

MarcSkovMadsen:
Hi @philippjfr 

I have had the same thought. Makes sense. 

- I think the reference notebook should introduce things better - one thing at the time and lead up to some thing that uses the things together.
- Moving the echarts to a seperate notebook and providing more examples  would also be great.

MarcSkovMadsen:
I'm glad you like it after having played around with it. 

I think its really, really powerfull and will let users add a lot of custom things they need to Panel easily.

But I also think there can be so many use cases that users will find bugs, need help and maybe suggest changes to the api for some period of time. Just because this is a bit difficult to get right.

And eventually someone should build that `WebComponent` to Bokeh model converter. From a custom WebComponent read all the parameters and generate the `.ts` and 2x`.py` code. I think it can be done.

MarcSkovMadsen:
I've now read my way through the changes and I nothing catched my eye. Glad to see it improving. 

I will try running the code when I get the time as well to see if it still works. So maybe some comments to come.

MarcSkovMadsen:
FYI. @philippjfr 

I've rewritten the reference notebook almost from scratch. Hope you like it. Otherwise let me know.

It actually made me even more happy about the `WebComponent` and how power full it is because

- I could implement `mwc-button`, `mwc-slider` fast and without problems.
- I could implement `lit-datatable` as soon as I understood that I needed to 
    - set the `height`
    - provide data in the right format.

Furthermore I found a lot of new, awesome web component libraries that I've added to the resources list at the bottom of the notebook.

web_component_example

panel/pane/web_component.py
```diff
@@ -0,0 +1,424 @@
...
+        if 'properties_last_change' in msg:            
+            for prop, new_value in msg['properties_last_change'].items():
+                pname = self.properties_to_watch.get(prop)
+                if pname is None:
+                    continue
+                pobj = self.param[pname]
+                ptype = type(pobj)
+                if isinstance(new_value, str) and ptype in PARAMETER_TYPE:
+                    new_value = PARAMETER_TYPE[ptype](new_value)
+                msg[pname] = new_value
```
philippjfr, line 290: @MarcSkovMadsen I've refactored all the different methods you had introduced to make of the messaging protocols that Panel itself provides. This makes it much clearer what the differences in serialization and deserialization approaches are. Can you explain to me why properties, events and attributes have subtly different deserialization behaviors? Can these be unified? As example the `properties_last_change` deserialization explicitly checks:

```
if isinstance(new_value, str)
```

Is this necessary? 'events' and 'attributes' don't have this check.

philippjfr:
Some general review comments, I'll add more as I come across them:

* Please do not add typing yet, this should be decided on a project wide basis, and probably won't happen until we require 3.7 as the minimum version.
* You do not have to check if a parameter value or bokeh model property value is equal to the previous value. This happens automatically.

panel/pane/web_component.py
```diff
@@ -0,0 +1,424 @@
...
+    attr_dict = None
+
+    def handle_starttag(self, tag, attrs):
+        if self.first_tag:
+            for attr in attrs:
+                self.attr_dict[attr[0]] = attr[1]
+            self.first_tag = False
+
+
+class WebComponent(Widget):
```
philippjfr, line 47: Is there a reason this lives in `./pane` if it subclasses `Widget`?

panel/pane/web_component.py
MarcSkovMadsen: There are differences. For example for a boolean value on an attribute.

If True its stated as `attribute` or `attribute=""`. If False the attribute is just removed.

Maybe some things can be cleaned up. Hopefully the many tests will catch changes. The problem is that I don't know how to setup Typescript tests. So there are none.

I just remember it was complicated. But the many examples I created helped find all the corner cases.

panel/pane/web_component.py
MarcSkovMadsen: Yes. Try moving it into `./widgets`. For me it gave an error.

philippjfr:
Another general comment. All the files you are committing use DOS style line ending, these end up being really ugly in my editor and result in an inconsistent code base. I haven't quite worked out how it works yet but it would be good if we could standardize on Linux line endings. [Here](https://stackoverflow.com/questions/1967370/git-replacing-lf-with-crlf)'s a StackOverflow on the topic.

MarcSkovMadsen:
Hi @philippjfr 

Please let me know if there is something is should do here to get this accepted or to help you finalize it. For now I don't know of anything.

philippjfr:
>Please let me know if there is something I should do here to get this accepted or to help you finalize it. For now I don't know of anything.

I need to finish off my review here but have been swamped with other work. Hoping to finish by the end of the week.

MarcSkovMadsen:
I have now set `git config --global core.autocrlf true`. Let me know @philippjfr if you see problems again.

philippjfr:
Much appreciated, I'll try to finally finish reviewing this PR this weekend. Last I worked on it it was in a pretty decent shape.

philippjfr:
So I've been reviewing this and the main issue I see right now is loading of the Javascript modules. There should be a unified way of declaring these dependencies across notebooks and the server. Could you elaborate a little bit on how you currently load the JS modules in your server apps?

MarcSkovMadsen:
For now I don't have any nice way to load js modules because Panel does not support this. I load them via `panel.pane.HTML` (or in my Template). See the notebook examples for more info.

MarcSkovMadsen:
Similarly to the module imports. Importing the robot font and material icons is not really supported either.

```html
<link href="https://fonts.googleapis.com/css?family=Roboto:300,400,500" rel="stylesheet">
<link href="https://fonts.googleapis.com/css?family=Material+Icons&display=block" rel="stylesheet">
```

philippjfr:
I **think** you pushed over some of my commits yesterday, so there's some weirdness with the commit history now but that shouldn't matter. Anyway, I think I've solved those issues now and I can load all components by initializing the dependencies in `pn.config` or `pn.extension`. I'm not 100% happy with everything but it's in a state that I'm okay with merging and cleaning up as we move forward.

I'm very appreciative of the fact that you want credit for your **considerable** work here, I'm just not entirely happy inserting credit sections all over the core docs, i.e. the user guide and reference gallery (the actual gallery on the other hand should very explicitly call out contributors though). Several users have contributed significant work adding components and improving the docs and calling out specific users seems out of place to me (I'm not aware of any other projects that do this). That said I do want you to get the credit that you are due, so if you have any suggestions for calling out contributions by you (and other users) more prominently it would be great to hear them.

panel/_templates/autoload_panel_js.js
```diff
@@ -109,6 +109,10 @@ calls it with the rendered model.
         element.onerror = on_error;
         element.async = false;
         element.src = url;
+        if (url.endsWith('?module')) {
+          element.type = "module"
+        }
+		console.log(element)
```
philippjfr, line 115:
```suggestion
```

philippjfr:
@MarcSkovMadsen Please review my latest changes before I merge this.

MarcSkovMadsen:
> I'm just not entirely happy inserting credit sections all over the core docs, i.e. the user guide and reference gallery (the actual gallery on the other hand should very explicitly call out contributors though). 

Makes sense

MarcSkovMadsen:
My plan is to include ECharts, Wired and ModelViewer in awesome-panel.org using my version of the WebComponent. When that is done I would change to this version and try testing everything automatically and manually. The purpose is to be able to test it on a lot of different examples to make sure it works.

I would expect it to be this week.

Would that make sense @philippjfr ?

MarcSkovMadsen:
@philippjfr . Could you maybe list the things you would see improved? Then I/ you/ we could improve it as we get time and motivation? Thanks.

philippjfr:
That timeline sounds fine to me. I'd like to target a panel 0.10 release for mid June.

MarcSkovMadsen:
Feedback from review of WebComponent Notebook

- I've moved it from pane to widget folder
- The `<script src="./node_modules/@webcomponents/webcomponentsjs/webcomponents-loader.js"></script>` import raises an error. See below screenshot.
- The "raised" property of the MWCButton no longer works. See below video
    - My hypothesis is that now boolean attributes in general do not work.

webcomponents-loader_not_working

raised_not_working

MarcSkovMadsen:
Feedback from testing the WebComponent on the "Test Material Components" page of awesome-panel.

MWCButton (See screenshot below)
- Unelevated attribute no longer works. It's just fixed a True.
- Click event/ Clicks parameter no longer works.
    - It should increment everytime the button is clicked but does not.

MWCSelect (See screenshot below)
- Select event/ Selects parameter no longer works.
    - It should increment everytime the selection changes but does not.

image

image

#### Reproducable steps

1. `git clone https://github.com/MarcSkovMadsen/awesome-panel.git@panel-webcomponent`
    - alternatively `git clone https://github.com/MarcSkovMadsen/awesome-panel.git@0767cfb65919b7336417f284883ec987cb2061c3`
2. Follow the installation instructions here https://github.com/marcskovmadsen/awesome-panel#getting-started-with-the-awesome-panel-repository
3. Upgrade Panel to the branch with the web component: `pip install git+https://github.com/holoviz/panel.git@web_component_only`
4. Run the app: `python -m panel serve 'app.py' --dev --show`
5. Navigate to the `Test Materials Components' page and manually test the components
6. Try out the component

MarcSkovMadsen:
Feedback from review of rest of awesome-panel.org (cf. feedback on Material above).

The rest seems to work as before including Wired, ECharts, Perspective and ModelViewer Test Examples.

MarcSkovMadsen:
Feedback from review of WebComponent.ipynb notebook.

I cannot load the js files

image

MarcSkovMadsen:
I've added the material example that fails above to this branch

You can test it out with the command `python -m panel serve 'panel\tests\components\test_material.py' --dev --show`

In this context error messages for click the MWCButton and changing the value of the MWCSelect shows up in the console log. I cannot find information on why the `Unelevated` parameter does not work though.

image

MarcSkovMadsen:
Hi @philippjfr 

Could you take a look at the issues in the notebook and the material example? 

When done I will continue the review. 

Thanks.

nickderobertis:
Just want to say, I'm very excited for this feature. I am basically waiting for it to be merged to begin building a few different frontends in Panel. Since there are already several good web component libraries and it looks pretty easy to embed React components in web components, this really vastly expands the capabilities of Panel. Thanks @MarcSkovMadsen and @philippjfr for your work on this!

ltalirz:
Same here, would be fantastic to see this merged!

MarcSkovMadsen:
I addition to a simple python component I think there is also a lot of things that could be improved by documenting and creating a starter template for the existing components infrastructure using bokeh models. Inspiration can be found here https://twitter.com/streamlit/status/1283053687213096963?s=21

MarcSkovMadsen:
Hi @nickderobertis 

If you can already do React I think you should look at creating true Panel extensions by implementing Bokeh models. 

Normally they are written in Typescript only but I guess it should be possible to do in React also.

If you are up for it you could create a post "How to create Panel Extension/ Custom component" on discourse and point to the library/ libraries you would like to use.

I and others could help you get started. And if I have the time/ motivation I could create a panel-extensions-starter project on Github with

- Documentation
- Examples
- Starter code

that others could use? What do you think?

MarcSkovMadsen:
Hi @nickderobertis 

While we wait for the WebComponent to make it easier to extend Panel using Python primarely I've started a new project on how to create Panel Extensions. There should already be inspiration to get started.

You can check it out here https://github.com/marcskovmadsen/panel-extensions-template

[image](https://github.com/marcskovmadsen/panel-extensions-template)

nickderobertis:
@MarcSkovMadsen I don't know if I would go so far as saying I do React, I've done some toy examples and mainly got into it because before I found Panel I was playing around with Dash. Actually this WebComponent setup you have in this MR and the example Material components reminds me quite of bit of setting up a new custom React component in Dash. I found it difficult to scale Dash and I love the flexibility of Panel to work in Jupyter as well as a standalone server, so that is why I came to Panel and have been so excited for this MR! I just also mention React because there is already a rich ecosystem of open source components available.

While I'm not a React developer, I have a few years of experience with Typescript and Angular, so the idea of implementing Bokeh models doesn't scare me. But the whole reason I'm trying to use Panel is to spend more time writing Python and less time writing JS/TS as well as biolerplate code to transfer data between the backend and frontend. Actually after this is merged I was thinking about creating something analogous to this for React components, though I don't know enough about the architecture yet to say what the implementation would look like. Ultimately I'd like to be able to take existing React components and hook them up in Python with little work. I'm happy to write Bokeh models as necessary to support this and other extensions to Panel but I'd like to get to working nearly completely in Python when it comes time to actually build a frontend.

I'm happy to contribute to the panel extensions project once I go down this route, but I think that will be after this WebComponent is implemented.

jbednar:
> I found it difficult to scale Dash

@nickderobertis , it would be interesting to hear more about this, because there are few people with background in developing both for Dash and for Panel.  In some ways Dash scales better than Panel for the things it supports, because Dash lacks server-side session state, allowing it to handle more simultaneous users without overwhelming the server.  But maybe you mean in terms of complexity for the programmer as the application adds more components, in which case yes, I think Panel does scale better. Love to hear your thoughts on these or related topics so far!

MarcSkovMadsen:
The WebComponent should actually support configuring actions as well. A simple use case/ example could be wrapping

http://ricostacruz.com/nprogress/

image

MarcSkovMadsen:
FYI. I've plan to ship the WebComponent as a part of the awesome-panel-extensions package ASAP.

I will do improvements in that repo for a while. Either when it settles down or you @philippjfr tell me you would like to ship the WebComponent with panel I will sync the code.

MarcSkovMadsen:
The WebComponent is really just a new `HTML` pane that sets up bidirectional communication with the first html element in the html string. I believe the consequence is huge. You can easily wrap most existing .js libraries using the Web Component. Maybe not as the end product (use Bokeh models for that), but for quick exploration, prototypes, to test demand etc. It's really, really usefull.

Just to make sure that everybody understands this I've created the below example.

```python
spinjs_html = """
<div class="spinnerContainer" style="width: 100%; height:100px;">
       </div>
<script type="text/javascript">
    var spinnerContainer = document.getElementById('spinnerContainer');
    target = document.currentScript.parentElement.children[0];
    Object.defineProperty(target, 'lines', {
      get() {
        return this.spinner.lines;
      },
      set(value) {
        opts = {lines: value};
        this.innerHTML="";
        target.spinner=new Spinner(opts).spin(this);
      }
    });
    target.lines=13
</script>
"""

class Spinjs(WebComponent):
    html = param.String(spinjs_html)
    properties_to_watch = param.Dict({"lines": "lines"})
    
    lines = param.Integer(default=13, bounds=(1,20))
    height= param.Integer(default=100)

    
spinjs = Spinjs()
pn.Column(
    spinjs,
    spinjs.param.lines
)

# Note since the Spinner object does not support setting a new lines value, for example via target.spinner.lines=value
# I had to implement the lines Property on the target. It's a trick to learn.

# Note the implementation of WebComponent sets up bidirectional communication with the first element in the html string. 
# I.e. the `div` element and not the `script` element.
```

You will need to load the https://spin.js.org/spin.js file for it to work.

spin-js

MarcSkovMadsen:
The `WebComponent` pane is now available in the **awesome-panel-extensions** package.

You can study the [WebComponent Docs](https://awesome-panel.readthedocs.io/en/latest/packages/awesome-panel-extensions/index.html#webcomponent).

You can also learn more about building Panel Extensions and the Web Component extension in particular in the [Awesome Panel Extensions Guide](https://awesome-panel.readthedocs.io/en/latest/guides/awesome-panel-extensions-guide/index.html)

nickderobertis:
> > I found it difficult to scale Dash
> 
> @nickderobertis , it would be interesting to hear more about this, because there are few people with background in developing both for Dash and for Panel. In some ways Dash scales better than Panel for the things it supports, because Dash lacks server-side session state, allowing it to handle more simultaneous users without overwhelming the server. But maybe you mean in terms of complexity for the programmer as the application adds more components, in which case yes, I think Panel does scale better. Love to hear your thoughts on these or related topics so far!

@jbednar Yes, I should clarify because scalability can take a lot of different meanings. I'm talking about the latter,  the complexity of developing, maintaining, and adapting a larger application, rather than scalability from a user throughput perspective. 

Perhaps it will be helpful to explain my current use case a bit. I have been developing a few different open source Python packages which I think would be very nice to use with GUIs. These are primarily developer tools, and so my main use case is for only a single user (the developer) to be using the GUI at once, so supporting many simultaneous users is not a main priority. 

I have developed several packages separately but ultimately use them together in my work. So my vision for the GUIs is that each package should expose some main component. This main component can be used standalone to work with that package, or you can create a project dashboard by including the components for each of the packages being used in the project. This is where it seems like Panel is already well set up for this, as things are fairly well isolated.

Maybe it is just the way I am used to working with Angular, but to me it seems to have a maintainable frontend, components and services should be self-contained. Inputs and outputs (events) should be defined within a component then that component should be able to be plugged in anywhere. This is where I was struggling a bit with Dash. To link anything together in Dash, it must be done using callbacks. Callbacks must be registered on the main application instance, which already makes isolation difficult. 

Here is my project where I have built out a prototype Dash application: https://github.com/nickderobertis/py-file-conf-gui . Note that even the underlying package is currently unstable and not well documented so I'm not expecting you to understand what the package is trying to do, but you can see how I've laid things out in a Dash app.

I was able to figure out a workable solution to build components in Dash. You can see my `DashPythonComponent` here: https://github.com/nickderobertis/py-file-conf-gui/blob/master/pyfileconfgui/dash_ext/component.py

The purpose of that structure is that now callbacks can be defined in the same class which defines the layout and behavior of the component, rather than having to separately add them where the app instance is being created. After creating the app instance, I call the class method to register the app in the component base class. Then when components are initialized, they will each register their own callbacks on the main application instance. 

So I was feeling pretty good about moving forward with Dash after getting that working, but it turns out there are significant gotchas working with the callback structure: 

- Everything is linked together and updated using callbacks. I ran into a situation where class attributes were getting updated and I wanted the view to update in response. This is the main use case of Param within Panel and seems very straightforward there. But in Dash I had to add a callback with an Interval component as the input, and then the callback just returns the value of the class attribute. Then it will update at whatever frequency I set in the interval, but it does this by polling the backend at this interval, certainly not efficient and feels like a hacky setup for something very simple.
- Callbacks must have inputs and outputs, so it is a bit hacky if you just want to take some action on an event without displaying an output from it
- An attribute can only be the output of a single callback. So if you need to update the same attribute from two separate sources, the callback has to be defined for both together, and the function needs to determine which input it is getting, rather than being able to register them separately. This limits extending an existing component.
- All callbacks and components must be defined when the app starts. Not too much of an issue if using the `DashPythonComponent` I created, but it will increase load time for larger multi-page apps that could be lazy-loading the pages.

In Panel we have multiple options for connecting components (`@param.depends`, `.link`, `.watch`) so it seems much more flexible to handle more use cases. 

My other main criticism of Dash (which remains to be seen whether it will be an issue for Panel), is that it is supposedly easy to hook up your own React components, but I still found the process to be laborious. They have a cookiecutter repo which certainly made things easier. Maybe it is due to the specifics of the component I was trying to transfer over, but I ended up needing to write 270 lines of JS with substantial troubleshooting to get it to work... https://github.com/nickderobertis/dash-keyed-file-browser/blob/master/src/lib/components/KeyedFileBrowser.react.js  which is based on this React component: https://github.com/uptick/react-keyed-file-browser

Ultimately a lot of the pain in transferring the component was a difference in implementation philosophy. The only way to link things together in Dash is by attributes' values, while the React component was mainly exposing hooks for callback functions. So I had to add callback functions in JS to set properties which could be picked up in Dash.

Long-winded answer to a simple question, but hopefully that gives some context on the main reasons I'm trying to move over to Panel despite already learning Dash. 

@MarcSkovMadsen Thank you for putting together the extensions package. I've had a lot going on at work lately so haven't been able to play around with it yet, but I'm excited for when I get some time.

MarcSkovMadsen:
Hi @nickderobertis 

I have a feeling we are coming from the same place. We want to build our apps from small reusable components (like for example in Angular). And for that Param and Panel is very promising.

But sometimes I doubt if there are a lot of users wanting that. When I look at the success of Streamlit it's the ease of use that wins and with Dash it's the very nice looking smaller apps that wins.

But being able to build some like the Awesome Panel Designer in Panel is a sign of the power that can be unleashed from Panel one day. I think that it is very powerful for many python packages in the analytics and data science space to provide a UI on top. something that works both as a web app and in the Notebook. You can do as with Tensorboard a lot of advanced JS or you can use Panel if you are not a project with that kind of resources.

jbednar:
@nickderobertis , thanks for that background! I do think Panel is set up to make what you're trying to do easier, by explicitly being compositional in ways that let you focus on individual components that can then be pasted together into an overall application even when there are dependencies between them.

MarcSkovMadsen:
I will close this one as the ECharts example and the WebComponent have other lives now.

WillGeller:
Hello! Can you explain to me what was the end result here of the WebComponent Panel model. Has it made it's way into core Panel or is it still only a part of Awesome Panel Extensions. This feature is a key aspect of our team choosing to adopt Panel over native Bokeh the development of our Python SDK.

jbednar:
Is that maybe what is now called [ReactiveHTML](https://panel.holoviz.org/user_guide/Custom_Components.html#reactivehtml-components) in Panel?

WillGeller:
Oh that is possible but I didn't see a specific mention of it in this
thread. The reactiveHtml model seems to support bidirectional data binding.
Although the documentation for it doesn't specifically mention web
components there is an example with the material design component.

MarcSkovMadsen:
The ReactiveHTML component is the way to go.

PR #1262, opened 2020-04-12
Add Card and Accordion layout [doc-build]

philippjfr:
Supersedes https://github.com/holoviz/panel/pull/1182

Adds a Card model with configurable HTML tags and css classes and collapse button

Screen Shot 2020-04-12 at 2 34 58 PM

Here's what it will look like in the coming Material design template:

Screen Shot 2020-04-13 at 1 03 36 AM


maximlt:
This is really really cool!

MarcSkovMadsen:
Looks nice. Does it support 

- Markdown or HTML in the header?

For example I might like to put an icon in there, a link or set the "size"/ "type" of header?

philippjfr:
Here are the docs I'm about to add, you'll see an example with a logo: https://anaconda.org/philippjfr/card/notebook

MarcSkovMadsen:
Hi @philippjfr 

It's great. Here is a little feedback.

1. The toogle area is very small and difficult to hit (see video below). Can it be made larger?
2. Personally I would have liked to include something better than +/- for toggle in my example. I can see that a lot of times < and > (rotated 90 degrees) are used. And they are larger and bolder than what is there now.
3. I have not seen it but is there a divider or border between the objects? I think that gives a nice seperation of objects. Or would I just specify a divider my self? Then maybe that should be included as an example.
4. When clicking the toggle button a blue border appears while the +/- is clicked. This is a bit strange and should be removed if possible.
5. Personally I would like the scroll to be on the objects only and not include the title/ header.

#### Difficult to hit toggle

card (2)

#### Divider between objects

image

#### Blue border when clicking +/-

blue_border

#### Scroll includes title/ header

image

philippjfr:
This PR now also adds support for a new `Accordion` layout which behaves very similarly to a `Tabs` component:

accordion
Screen Shot 2020-04-13 at 7 38 13 PM

philippjfr:
Thanks for the feedback, a lot of what you're seeing is simply because anaconda.org defines various CSS which does not seem to work too well with the Card CSS. I'll try to make sure they don't interfere like that but there's only so much we can do.

>The toogle area is very small and difficult to hit (see video below). Can it be made larger?

It is actually considerably larger than it appears on anaconda.org, it defines various CSS tags that interfere with the actual look and feel. I'll still try to make it larger though.

> Personally I would have liked to include something better than +/- for toggle in my example. I can see that a lot of times < and > (rotated 90 degrees) are used. And they are larger and bolder than what is there now.

Personally I prefer +/- and again they appear much larger and bolder without anaconda.org's theme interferring. I'll try to make sure that external CSS doesn't interfere like that.

>I have not seen it but is there a divider or border between the objects? I think that gives a nice seperation of objects. Or would I just specify a divider my self? Then maybe that should be included as an example.

Happy to add an example, much less happy adding dividers by default.

>When clicking the toggle button a blue border appears while the +/- is clicked. This is a bit strange and should be removed if possible.

Again, this is anaconda.org CSS, not part of the default look.

>Personally I would like the scroll to be on the objects only and not include the title/ header.

Unlikely to be easy, but I'll give that a shot.

kcpevey:
I'm super excited to see this!! 🎉

philippjfr:
Merging for now. We can add additional tweaks later.

PR #1263, opened 2020-04-12
Wired JS Components Example

MarcSkovMadsen:
I've moved the wired code to a notebook. This PR builds on top of the `WebComponent` PR https://github.com/holoviz/panel/pull/1252.

This requires some work

- The app does not display nicely in notebook (see below).
- Some of the widgets no longer works.

#### Server

image

#### Notebook

image

@philippjfr . Do you know what goes wrong in the notebook?. I tried setting a fixed height but it did not change anything.


MarcSkovMadsen:
I've started to restructure the notebook. The purpose is to make in more understandable.

The result though is work for me, a length notebook and code that cannot just be copied into a users project.

image

philippjfr:
Closing for now as we discussed. Hope we can re-add it in some form later.

MarcSkovMadsen:
Please don't delete for now. Still need the notebook as a example of a WebComponent implementation.

PR #1266, opened 2020-04-13
Style fixes for Card and Accordion [doc-build]

philippjfr:
You can see the current builds here:

https://pyviz-dev.github.io/panel/reference/layouts/Accordion.html
https://pyviz-dev.github.io/panel/reference/layouts/Card.html

@MarcSkovMadsen @jbednar Further styling suggestions would be very welcome. Some questions I have:

* Should Card really have rounded corners? We use them elsewhere but I'm starting to think it looks cleaner without
* Should the Header title font be smaller by default?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -42,7 +42,7 @@ stages:
   - name: docs_dev
     if: tag =~ ^v(\d+|\.)+([a-z]|rc)\d+$
   - name: docs_daily
-    if: ((type = cron OR commit_message =~ /\[doc-build\]/) and branch = master)
+    if: ((type = cron and branch = master) OR (commit_message =~ /\[doc-build\]/))
 
 jobs:
   include:
--- a/doc/_static/nbsite.css
+++ b/doc/_static/nbsite.css
@@ -496,3 +496,11 @@ div.logo-block {
 div.bk-root {
   min-height: 50px
 }
+
+.bk.accordion-header {
+  color: black;
+}
+
+.bk.card-header {
+  color: black;
+}
--- a/examples/reference/layouts/Accordion.ipynb
+++ b/examples/reference/layouts/Accordion.ipynb
@@ -25,7 +25,7 @@
     "Styling-related parameters:\n",
     "\n",
     "* **``active_header_background``** (str): The background color of the header when the `Card` is expanded.\n",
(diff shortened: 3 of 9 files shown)


MarcSkovMadsen:
Hi @philippjfr 

I just love the accordion. 

- I also thinks it looks nice in the notebook without the rounded corners
- An i just thought wow that clicking the header toggles it open and closed.

There is alternative inspiration for accordions (material expansions) here https://material.angular.io/components/expansion/overview.

MarcSkovMadsen:
Regarding the according notebook. Could you make one of the example full width?

Just to illustrate that its possible and how to do it.

MarcSkovMadsen:
Regarding the Card. Three comments

#### Double border below.

I actually cannot say if it is good or bad. But it is different than bootstrap.

image

#### `button_color`

I guess it (should?) affect the color of the header text as well. Should it be called `header_color`?

image

#### Divider

Could you throw in a divider in the last example? 

Just to show case you can seperate things in the card and that lays out nicely.

image

philippjfr:
>I actually cannot say if it is good or bad. But it is different than bootstrap.

Fixed.

>I guess it (should?) affect the color of the header text as well. Should it be called header_color?

Fixed.

>Could you throw in a divider in the last example?

Done.

MarcSkovMadsen:
hmm. Something must have gone wrong. The accordion example looks like

image

philippjfr:
The daily doc build is built from master.

jbednar:
It would be nice if the docs for Card showed some cards side by side; right now it's not clear that it's expected you'd use them that way.

philippjfr:
>It would be nice if the docs for Card showed some cards side by side; right now it's not clear that it's expected you'd use them that way.

True and nested cards.

MarcSkovMadsen:
And maybe throw in an example with some text and an image. This is something that is a normal use case for cards as well.

MarcSkovMadsen:
Maybe an image like.

image

:-)

philippjfr:
😄 

This is the current status of cards in the material template btw:

Screen Shot 2020-04-14 at 7 02 45 PM

jbednar:
Nice!

PR #1285, opened 2020-04-19
Add support for ipywidgets on the server and in the notebook [doc-build]

philippjfr:
This demonstrates bi-directional communication of an ipywidget embedded in a Panel/Bokeh layout in the notebook.

ipywidget

- [x] Add docs
- [x] Release jupyterlab_pyviz 1.1
- [x] Try to support save and notebook exports somehow

Requires https://github.com/holoviz/pyviz_comms/pull/57/


MarcSkovMadsen:
Would users be able to specify IPyWidgets as widgets to `panel.Param`?

MarcSkovMadsen:
Does this mean that users can use ipyleaflet, bqplot etc. with Panel? Or is it only the base IPyWidgets?

philippjfr:
>Does this mean that users can use ipyleaflet, bqplot etc. with Panel? Or is it only the base IPyWidgets?

It does indeed.

>Would users be able to specify IPyWidgets as widgets to panel.Param?

Not yet, but possible in theory once this is merged.

Issue #1283, opened 2020-04-19
Standardize gallery examples


philippjfr:
As users contribute more examples to the gallery it's becoming clear that we need to standardize the format these examples take. I'd propose the following structure:

### Header

* Title
* Description
* Author
* Tags?
* License?
* ?

### Code

The actual code section of a gallery example can be a single cell with code comments or slowly building up the code across multiple cells. I think both have their place depending on the concepts being delivered.

### Output

The final output of the example should always be in the last cell of the notebook. This makes it possible for us to append an iframe of the running app in the website build.

Cc: @MarcSkovMadsen

philippjfr:
Would it make sense to add a template to the repo or even provide a CLI to generate the template?

MarcSkovMadsen:
I would be creating a cli one Day. At least in my own code and for my own use.

The example notebook template could be on thing.

Another would be for generering the 3 files involved in a new pane, widget or layout based on a bokeh model. Maybe even including test files. And basic tests.

A third thing is a new parameterized class. Maybe split in three: a model, a component and an app. And then with test files.

A fourth thing would be for converting a WebComponent into a true Panel Component based on a bokeh model.

philippjfr:
@MarcSkovMadsen I've seen your updated notebooks, thanks for making those changes but I'd like to make the suggestion of making the header more compact, i.e. having a single cell with roughly this formatting:

```
## ECharts Webcomponent example

**Description**: This example demonstrates blah, blah, blah, blah, ....
**Author**: Marc Skov Madsen
**Tags**: ECharts, WebComponent
**License**: BSD-3
```

@jbednar Any other suggestions?

MarcSkovMadsen:
I like the suggestions. But I would suggest a small change

## ECharts Webcomponent example

This example demonstrates blah, blah, blah, blah, ....

**Author**: Marc Skov Madsen
**Tags**: ECharts, WebComponent
**License**: BSD-3

MarcSkovMadsen:
And I think together with the *standard description* you should link to one or two example notebooks that are satisfying and can be used as inspiration.

philippjfr:
>But I would suggest a small change. I think you should give the WHY (i.e. description) a little bit more space.

Sounds good!

MarcSkovMadsen:
I have added in an image as a part of the description as it says more than a thousand words and works well if you are to communicate via demos, screenshots, videos etc.

I've actually considered if the image should be a small .gif video of the end product?

[...]

PR #1286, opened 2020-04-19
EChartsPyeCharts WebComponent Example

MarcSkovMadsen:
Builds on the WebComponent PR https://github.com/holoviz/panel/pull/1252

The more I use ECharts the more I realize how awesome the look and feel is to the end user.

@philippjfr . I've tried to use the standard template you provided here https://github.com/holoviz/panel/issues/1283. Please let me know if its ok.

This example is ready for review I believe.

echarts


philippjfr:
Closing, since we now have an ECharts model I don't think we'll need this.

MarcSkovMadsen:
Please don't delete for now. Still need the notebook as a example of a WebComponent implementation.

Issue #1128, opened 2020-03-05
Need a more flexible Number widget


maximlt:
**The problem**

When I create a Param-based app that declares a `Number` object with an open-ended bound, the default widget instantiated by panel is a `Spinner`. See the code to reproduce that behaviour.

```python
import param
import panel as pn

class Dummy(param.Parameterized):
    amount = param.Number(bounds=(0, None))

app = pn.panel(Dummy())
app.show()
```

This was an unexepected behaviour to me as the `Spinner`  widget isn't even introduced on the [widgets](https://panel.holoviz.org/user_guide/Widgets.html) page.

The instantiated `Spinner` has a `step` value of **1** which means that it's not possible to interactivaly set that widget value to a decimal number.
![spinner](https://user-images.githubusercontent.com/35924738/76013202-d9221f00-5f17-11ea-9bf9-07ab7cb1cb08.png)

It is however possible to programmatically set its value to a decimal number.
![spinner_setval](https://user-images.githubusercontent.com/35924738/76013452-31f1b780-5f18-11ea-87be-112b68ab09ec.PNG)
But now the widget doesn't display the exact value but the value rounded according to that step of 1.

**What would be really nice to have**

In the case of a number that doesn't have any hard-bounds, I'd rather not get a `Spinner` but a more flexible widget, possibly a `NumberInput`. This widget would be similar to the `TextInput` widget, it would however accept only numeric (int, float) as input.

Ideally, a parameter `format` would allow to control how the number is formatted by the widget.  The following code would instantiate a `NumberInput` widget that could display values such as `2.00 €` or `2.15 €`.
```python
pn.Param(Dummy.param, widgets={
    'amount': {'type': pn.widgets.NumberInput, 'format': '{.2f} €'},
)
```

In the end, the `Spinner` widget isn't that bad, and the `IntSlider` and `FloatSlider` widgets are also useful. I believe though that a more flexible widget would come in handy quite frequently, this is at least what I've observed with the few apps I've built so far.

**Something else I've tried**

I thought the `TextInput` widget would provide me with a simple workaround. Unfortunately, that failed.
![ti_error](https://user-images.githubusercontent.com/35924738/76015026-e096f780-5f1a-11ea-8a22-9036e8402996.PNG)

**Bonus**
Thanks a lot for `panel`, it's a great library! If I can be of any help in implementing this feature (Python only, sorry), please guide me through this work.

MarcSkovMadsen:
Hi @maximlt 

[...]

PR #1301, opened 2020-04-26
Do not override widget class if already specified by user

poplarShift:
Following the discussions in #1128, I discovered that providing the `widgets` dict currently did not allow specifying it for `param.Number`:

```
import param 
import panel as pn
pn.extension()

class T(param.Parameterized):
    a = param.Number(default=None)
t = T(a=1e-3)
    
pn.Pane(t.param, widgets={
    'a': pn.widgets.LiteralInput()
}) # falls back to Spinner no matter what
```

The reason was that `param.Number` was handled as a special case and actually overriden once again if no `bounds` are provided.

This PR simply adds another quick check and forces panel to roll with the user-supplied `widgets`, if provided, even for `param.Number`.

### DIFF
--- a/panel/param.py
+++ b/panel/param.py
@@ -310,7 +310,9 @@ class Param(PaneBase):
         p_obj = self.object.param[p_name]
         kw_widget = {}
 
+        widget_class_overridden = True
         if self.widgets is None or p_name not in self.widgets:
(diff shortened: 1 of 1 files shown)


philippjfr:
Thanks @poplarShift.

PR #1302, opened 2020-04-26
Fix is_number util function

poplarShift:
The current `is_number` function does not really do what it says on the tin:

```
from panel.util import is_number

# Works:

is_number(1.23) # True

# Does not work:

# is_number(1+1j) # throws TypeError
# is_number({}) # ditto

is_number('123') # Returns True which may be desired (?) but is not intuitive
```

I have not checked in detail every use of `is_number` in the code base so far, but I'm tentatively submitting this PR which should work for python 3.

### DIFF
--- a/panel/util.py
+++ b/panel/util.py
@@ -260,11 +260,7 @@ def value_as_date(value):
 
 
 def is_number(s):
-    try:
-        float(s)
-        return True
-    except ValueError:
-        return False
+    return isinstance(s, numbers.Number)
 
 
 def parse_query(query):


jbednar:
Hmm. The original code really doesn't seem to be testing whether something is a number, so maybe it should be renamed?  Depends how it's used.  I'm not sure for Panel purposes that imaginary numbers should be included in such a test, in that for most Panel purposes a real number is needed.

poplarShift:
Hah, you're completely right. I was so set on writing/finding some util function that would tell me whether something was a numeric type (for formatting the `LiteralInput` widget) that I didn't notice its sole purpose was the very specific `parse_query` function underneath. I'm closing but feel free to reopen if you see any merit in keeping it open.

PR #1304, opened 2020-04-27
Add support for VSCode

philippjfr:
It seems like VSCode does not yet like `_repr_mimebundle_` so we add an explicit `pn.config.comms = 'vscode'` that adds special handling to display ipywidgets output in VSCode. In practice I suspect that VSCode will have to fix this issue soon because ipywidgets 8.0 (currently in alpha) switches from `_ipython_display_` to `_repr_mimebundle_`.

- [ ] Add docs

### DIFF
--- a/panel/config.py
+++ b/panel/config.py
@@ -80,7 +80,7 @@ class _config(param.Parameterized):
         Specify the default sizing mode behavior of panels.""")
 
     _comms = param.ObjectSelector(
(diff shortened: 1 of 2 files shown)


MarcSkovMadsen:
I saw in the communication of the latest vs code release the “Ipywidgets are now supported”!

MarcSkovMadsen:
Looking so much forward to be able to develop using VS Code data science experience.

Issue #519, opened 2019-07-11
FileInput `name` keyword doesn't add name


ahuang11:
```
import panel as pn
pn.extension()

file_input = pn.widgets.FileInput(name='Select a file!')
file_input
```
![image](https://user-images.githubusercontent.com/15331990/61087256-fe42aa00-a3e9-11e9-8be1-4efe7c43274e.png)

maximlt:
Just had a look at this one. It seems that the bokeh `FileInput` widget does have a `name` attribute but that it's not used at all for UI purposes (it's an internal id used to select a bokeh model in a document).

There is a similar issue in bokeh https://github.com/bokeh/bokeh/issues/9552 dating from the beginning of this year. Apparently it's not a straightforward fix.

For the time being we could just remove ```name`` as a **Display** property in the reference docs [here](https://panel.holoviz.org/reference/widgets/FileInput.html) not to confuse people. @philippjfr  are you ok with that?

philippjfr:
Sure I suppose. What I would like to do is make a single issue listing all the widgets that don't have a way to set a title/name, rather than the multiple issues that we have currently.

maximlt:
> What I would like to do is make a single issue listing all the widgets that don't have a way to set a title/name, rather than the multiple issues that we have currently.

I can take care of that.

philippjfr:
Fabulous, I've made you a collaborator btw, so feel free to close issues once you've made the general one.

PR #1306, opened 2020-04-28
Improve the FileDownload widget

### LINKED ISSUE #1299: List of bugs related to the FileDownload widget
Fixing #1246 allowed me to dig into the code and better understand the behaviour of the ``FileDownload`` widget. I spotted some bugs that add up to this one #1289. Before submitting a PR, I wanted to list these bugs here.

The first one is quite bad, the last example in the Reference Gallery ([here](https://panel.holoviz.org/reference/widgets/FileDownload.html)) is actually broken. At first selecting a year (here 1970) and downloading the generated data works fine. Doing it a second time with another year seems to work fine but the data that is actually downloaded is the same data again, it hasn't been updated.
filedownload-broken-gallery

Here is another (minimal) example to reproduce this bug:
```python
from io import StringIO

opts = pn.widgets.Select(options=["a", "b", "c"])

@pn.depends(opts)
def cb(opts):
    return StringIO(opts)

fd = pn.widgets.FileDownload(callback=cb, filename="data.txt")
pn.Column(opts, fd).show()
```
The data available from the widget will always be the first one downloaded, it won't be updated.

Then there is a series of possibly related bugs. They occur after the widget is clicked and that its data/parameters are updated, the Save window pops up at each change. Also note that the ``disabled`` property isn't part of the documented API (see the [docs](https://panel.holoviz.org/reference/widgets/FileDownload.html)) but the widget inherits it. It doesn't actually disable the widget while this would be a useful feature for such a widget.
filedownload-broken-ui

Here is the code to reproduce that:
```python
from io import StringIO
fd = FD(file=StringIO("data"), filename="data.txt")
pn.Row(fd, fd.controls()).show()
```

I've been working on a PR that tries to tackle these bugs one by one (I wasn't aware of #1289 though). It's a rather slow process, mostly because of my inexistent html/typescript skills, as the root cause of all these bugs seem to arise from there.

@philippjfr  I'll submit what I have done so far at some point during the week-end so that you can check whether I went in the right direction.

maximlt:
Fixes #1299 

The intent of this PR is to fix various bugs with the FileDownload widget. I had to clearly define the behaviour of the widget, that to me follows three cases:
1. embed=True --> The widget is a link that looks like a button (right click to "Save as..." is possible)
2. auto=False --> There are two steps, first the widget is a button, after it's clicked it's a link (right click possible)
3. auto=True --> The widget is a button (right click to "Save as..." isn't possible).

To get there I had to make the following changes:
* On the python side: the biggest change is the new ``_transfers`` model attribute. It took me a long time before I dared to touch the python side but I found out it was the only solution that worked and allowed to fix all the bugs. When the widget is a button, a click will increment ``_clicks`` in typescript which will be reflected in python and trigger ``_transfer()``. This method will increment ``_transfers``, this signal will be caught on the typescript side, where a link will be created and clicked to download the data.
* On the typescript side: I've opted for an approach that is in favour of rendering first the widget and then updating it (that is not completely true as disabling/enabling the widget actually renders it again because of an inherited callback). It's quite a change compared to the first implementation, I believe it was required to cover all the cases supposedly offered by the widget.

Example fixed:
filedownload-fixed-example

Updating button type, disabled, filename fixed:
filedownload-fixed-ui

### DIFF
--- a/examples/reference/widgets/FileDownload.ipynb
+++ b/examples/reference/widgets/FileDownload.ipynb
@@ -14,7 +14,7 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "The ``FileDownload`` widget allows downloading a file on the frontend by sending the file data to the browser either on initialization (if ``embed=True``) or when the button is clicked. If the button\n",
+    "The ``FileDownload`` widget allows downloading a file on the frontend by sending the file data to the browser either on initialization (if ``embed=True``) or when the button is clicked.\n",
     "\n",
     "For more information about listening to widget events and laying out widgets refer to the [widgets user guide](../../user_guide/Widgets.ipynb). Alternatively you can learn how to build GUIs by declaring parameters independently of any specific widgets in the [param user guide](../../user_guide/Param.ipynb). To express interactivity entirely using Javascript without the need for a Python server take a look at the [links user guide](../../user_guide/Param.ipynb).\n",
     "\n",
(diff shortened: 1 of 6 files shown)


philippjfr:
Thanks for this, not an easy fix to make after all. The approach makes a lot of sense to me though so just let me know if you think it's ready.

maximlt:
Relieved that you're OK with all those changes!

Just spotted https://github.com/holoviz/panel/issues/519 so I'll see if I can also fix that and let you know when it's ready.

maximlt:
I think it's ready now.

philippjfr:
Looks ready to me, thanks so much for contributing!

MarcSkovMadsen:
So great to see this released. Just had feedback from one of my users that experienced this bug.

PR #1308, opened 2020-04-28
Add missing property in WidgetBox docs

maximlt:
The ``disabled`` property added to the ``WidgetBox`` widget by https://github.com/holoviz/panel/pull/532 wasn't documented in the reference example, fixes that.


philippjfr:
Thanks again!

Issue #509, opened 2019-07-09
Capture guidance about working with Matplotlib figures


philippjfr:
In https://github.com/pyviz/panel/pull/505 @tacaswell helpfully demonstrated the best way to work with matplotlib figures, we should capture that same guidance in the Matplotlib pane reference notebook.

tacaswell:
Ah, when I put in #505 I was looking for https://panel.pyviz.org/reference/panes/Matplotlib.html but could not find it...

philippjfr:
It lives in ``/examples/reference/panes/Matplotlib.ipynb``

PR #1309, opened 2020-04-28
Add matplotlib guidance to reference pane

maximlt:
Fixes https://github.com/holoviz/panel/issues/509

I found VSCode to be quite handy in working with notebooks and diffing them. It helped me to make sure I haven't added any metadata :)

### DIFF
--- a/examples/reference/panes/Matplotlib.ipynb
+++ b/examples/reference/panes/Matplotlib.ipynb
@@ -27,25 +27,34 @@
    ]
   },
   {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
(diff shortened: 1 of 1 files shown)


philippjfr:
Thanks, I think this makes a lot of sense. That said, I was still having issues with the inline backend popping up the figure if you don't set `matplotlib.use('agg')`. You're not seeing that.

maximlt:
Hi!
No I haven't noticed the inline backend popping up the figure in a notebook.
I've just tried again to test that and found something weird. The notebook runs fine in the classic notebook. In Lab however the figure isn't updated when the object is changed. I've tried with displaying the pane with `.servable()`, same bug. It works well only if I serve the pane with `.show()`. Would you like me to open an issue for that?

philippjfr:
Yes, let's open an issue.

maximlt:
Just tried again and it's working now in Lab on *master*. I don't know what went wrong but things seem to be alright now :)

Issue #1314, opened 2020-04-30
When Using Parameterized Class API, App Becomes Unresponsive After ListSelector/MultiChoice Has No Item Selected


pybokeh:
Hello,
I am fairly new to using Panel where I have been first learning to use the reactive API.  Recently, I am learning about Panel's parameterized class API.  Not sure if it is a bug or something I am doing wrong, but when I am using parameterized class API, my app becomes unresponsive or the chart won't update after my ListSelector widget or MultiChoice widget has no item selected.  My chart is no longer responsive or updates afterwards.  Here is my example [notebook](https://nbviewer.jupyter.org/github/pybokeh/jupyter_notebooks/blob/master/panel/parameterized/Altair_Covid19_Viewer.ipynb).  So in my example, "Ohio" is the only state selected.  So let's say I want to choose another state, I would deselect "Ohio", then select or choose another state.  But after doing so, my app becomes unresponsive or the chart no longer updates. I have to re-run code cell #4 for the chart to update again.  Everything else functionally works fine.

**Expected Behavior:** Even if no state is selected, I expect the app to remain responsive after I eventually select a state.

I have an equivalent example using reactive api, and my chart or dashboard will continue to work fine even after unselecting that singe item in the ListSelector/MultiChoice widget.

EDIT: Panel 0.9.5, bokeh 2.0.1

maximlt:
Hi @pybokeh,

I could reproduce that. At first I thought that `altair` didn't like so much when `state_province` was an empty list but it actually returns a plot in that case. You can check that with running `app.plotAltairLineChart()` below your dashboard .
![image](https://user-images.githubusercontent.com/35924738/80686243-b8fe7e80-8ac8-11ea-84e2-094a1b4163fd.png)
I guess it's an issue with `panel` then.

pybokeh:
Awesome!  Such a quick response!

PR #1315, opened 2020-04-30
Fix vega with empty datasets

xavArtley:
fix: https://github.com/holoviz/panel/issues/1314
ezgif com-video-to-gif


philippjfr:
Perfect, thanks!

PR #1319, opened 2020-05-02
Add support for serving static files

philippjfr:
This PR allows serving static files by providing a dictionary mapping from route to paths to `pn.serve` or `panel serve` on the command line.

On the commandline the syntax looks like this:

```bash
panel serve some_app.py --static-dirs custom_static="./examples"
``` 

while in Python it's specified as:

```python
pn.serve(some_pn_obj, static_dirs={'custom_static': './examples'})
```

In both cases all contents of './examples' will now be served on `http(s)://hostname/custom_static/` (note that the `/static` endpoint is reserved for bokeh's static files).

Implement https://github.com/holoviz/panel/issues/1118

- [x] Add docs
- [x] Add tests

Cc: @MarcSkovMadsen @lvankouwen Any chance you could give this a spin and see if it meets your requirements?

### DIFF
--- a/examples/user_guide/Deploy_and_Export.ipynb
+++ b/examples/user_guide/Deploy_and_Export.ipynb
@@ -191,7 +191,7 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "## Display in the Python REPL\n",
+    "### Display in the Python REPL\n",
     "\n",
     "Working from the command line will not automatically display rich representations inline as in a notebook, but you can still interact with your Panel components if you start a Bokeh server instance and open a separate browser window using the ``show`` method. The method has the following arguments:\n",
     "\n",
@@ -238,10 +238,14 @@
     ")\n",
     "```\n",
     "\n",
-    "The ``pn.serve`` function accepts the same arguments as the `show` method.\n",
-    "\n",
-    "\n",
-    "### Launching a server on the commandline\n",
+    "The ``pn.serve`` function accepts the same arguments as the `show` method."
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## Launching a server on the commandline\n",
     "\n",
     "Once the app is ready for deployment it can be served using the Bokeh server.  For a detailed breakdown of the design and functionality of Bokeh server, see the [Bokeh documentation](https://bokeh.pydata.org/en/latest/docs/user_guide/server.html). The most important thing to know is that Panel (and Bokeh) provide a CLI command to serve a Python script, app directory, or Jupyter notebook containing a Bokeh or Panel app. To launch a server using the CLI, simply run:\n",
     "\n",
@@ -298,6 +302,8 @@
     "                            Do not redirect to running app from root path\n",
     "      --num-procs N         Number of worker processes for an app. Using 0 will\n",
     "                            autodetect number of cores (defaults to 1)\n",
+    "      --static-dirs         Static directories to serve specified as key=value\n",
+    "                            pairs mapping from URL route to static file directory.\n",
     "      --websocket-max-message-size BYTES\n",
     "                            Set the Tornado websocket_max_message_size value\n",
     "                            (defaults to 20MB) NOTE: This setting has effect ONLY\n",
@@ -312,9 +318,31 @@
     "\n",
     "To turn a notebook into a deployable app simply append ``.servable()`` to one or more Panel objects, which will add the app to Bokeh's ``curdoc``, ensuring it can be discovered by Bokeh server on deployment. In this way it is trivial to build dashboards that can be used interactively in a notebook and then seamlessly deployed on Bokeh server.\n",
     "\n",
-    "When called on a notebook, `panel serve` first converts it to a python script using [`nbconvert.PythonExporter()`](https://nbconvert.readthedocs.io/en/stable/api/exporters.html), albeit with [IPython magics](https://ipython.readthedocs.io/en/stable/interactive/magics.html) stripped out. This means that non-code cells, such as raw cells, are entirely handled by `nbconvert` and [may modify the served app](https://nbsphinx.readthedocs.io/en/latest/raw-cells.html).\n",
+    "When called on a notebook, `panel serve` first converts it to a python script using [`nbconvert.PythonExporter()`](https://nbconvert.readthedocs.io/en/stable/api/exporters.html), albeit with [IPython magics](https://ipython.readthedocs.io/en/stable/interactive/magics.html) stripped out. This means that non-code cells, such as raw cells, are entirely handled by `nbconvert` and [may modify the served app](https://nbsphinx.readthedocs.io/en/latest/raw-cells.html)."
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## Static file hosting\n",
     "\n",
-    "### Accessing session state\n",
+    "Whether you're launching your application using `panel serve` from the commandline or using `pn.serve` in a script you can also serve static files. When using `panel serve` you can use the `--static-dirs` argument to specify a list of static directories to serve along with their routes, e.g.:\n",
+    "\n",
+    "    panel serve some_script.py --static-dirs assets=./assets\n",
+    "    \n",
+    "This will serve the `./assets` directory on the servers `/assets` route. Note however that the `/static` route is reserved internally by Panel.\n",
+    "\n",
+    "Similarly when using `pn.serve` or `panel_obj.show` the static routes may be defined as a dictionary, e.g. the equivalent to the example would be:\n",
+    "\n",
+    "    pn.serve(panel_obj, static_dirs={'assets': './assets'})"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## Accessing session state\n",
     "\n",
     "Whenever a Panel app is being served the ``panel.state`` object exposes some of the internal Bokeh server components to a user.\n",
     "\n",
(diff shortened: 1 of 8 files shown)


philippjfr:
What we might also consider is making it possible to serve a dictionary of filehandles on `pn.state.static_files`, e.g. you might allow users to do the following:

```python
pn.config.static_endpoint = 'static'
pn.state.static_files['custom.csv'] = StringIO(...)
```

which would then serve the `custom.csv` (and anything else that was added) on `http(s)://hostname/static/custom.csv`.

MarcSkovMadsen:
I would like to be using something like

```python
pn.serve(
    APP_ROUTES,
    static_dirs={
        'docs': './docs/_build/', # Sphinx documentation. Includes subfolders of html, css, js and images.
        'angular_app': './angular/src/build', # An older Angular App with a static subfolder.
        'static': './static' # includes './static/images/', './static/css/', 'static/js/' etc.
        },
    port=14033,
    dev=False,
    title="My Application"
    )
```

so I believe it would satisfy my use case.

philippjfr:
That should work fine, except for this:

>        'static': './static/', # includes './static/images/', './static/css/', 'static/js/'

The `static` endpoint is reserved for bokeh.

lvankouwen:
@philippjfr I think this feature is exactly what I need. I tried to test it, here is what I did:
- I followed the developers guidelines (https://panel.holoviz.org/developer_guide/index.html)
- I checked out to the serve_static branch
- I am working in the panel_dev environment
- When I try to import panel from the local repository I am missing packages
    - param
    - bokeh
    - pyviz_comms
    - markdown
- I have added these packages manually to the panel_dev enviroment. 
- Importing is now ok, but when I try to serve the simplest app I get the error below. I am not sure what this means so this is as far as I got. 

  File "C:\ProgramData\Anaconda3\envs\panel_dev\lib\site-packages\tornado\web.py", line 1703, in _execute
    result = await result
  File "C:\ProgramData\Anaconda3\envs\panel_dev\lib\site-packages\bokeh\server\views\doc_handler.py", line 58, in get
    template_variables=session.document.template_variables)
  File "C:\ProgramData\Anaconda3\envs\panel_dev\lib\site-packages\bokeh\embed\server.py", line 237, in server_html_page_for_session
    bundle = bundle_for_objs_and_resources(None, resources)
  File "C:\ProgramData\Anaconda3\envs\panel_dev\lib\site-packages\bokeh\embed\bundle.py", line 185, in bundle_for_objs_and_resources
    js_raw.extend([ Resources._inline(bundle.artifact_path) for bundle in extensions ])
  File "C:\ProgramData\Anaconda3\envs\panel_dev\lib\site-packages\bokeh\embed\bundle.py", line 185, in <listcomp>
    js_raw.extend([ Resources._inline(bundle.artifact_path) for bundle in extensions ])
  File "C:\ProgramData\Anaconda3\envs\panel_dev\lib\site-packages\bokeh\resources.py", line 372, in _inline
    with open(path, "rb") as f:
FileNotFoundError: [Errno 2] No such file or directory: 'C:\\src\\panel\\panel\\dist/panel.min.js'

maximlt:
Hi @lvankouwen, just replying here as I recently set up a dev environment for panel with similar settings I assume (Windows + conda). I didn't have the error you got with missing modules. Here is what I did:
1. ``conda create -n holoviz_dev -c pyviz "pyctdev>0.5.0"``
2. ``conda activate holoviz_dev``
3. ``doit ecosystem_setup``
4. ``git clone https://github.com/holoviz/panel.git``
5. ``cd panel``
6. ``doit env_create -c pyviz/label/dev -c conda-forge --name=panel_dev --python=3.7``
7. ``conda activate panel_dev``
8. ``doit develop_install -c pyviz/label/dev -c conda-forge -c bokeh -o build -o tests -o recommended -o examples``

That is slightly different from the docs:
* I created ``holoviz_dev`` because I didn't want to have ``pyctdev`` in my base environment
* I've added ``-o examples`` at the end of *8* to install the dependencies required to run the example notebooks (check ``setup.py`` to see which options are available)

If you run ``conda list`` after that, ``bokeh``, ``param``, etc. should be listed as installed. From your problems it seems that *8* didn't work well. In which directory were you when you  tried to import ``panel``?

philippjfr:
Missing that file means panel wasn't built. This should happen during the `python setup.py develop` or alternatively`pip install-e .` step but you can try to do it explicitly with `panel build panel`. An alternative explanation would be that you replaced the `static` endpoint, which I still have to disallow explicitly.

lvankouwen:
I did not install panel indeed. What I did was just add the panel directory as a sources root to pycharm. I understant now that panel needs to be build via the install script and can't operate from the files in the repo alone (I only have experience with developing very simple packages where this is possible). Anyway, I did `pip install -e panel`, which seemed to be successful. 

    > conda list panel

    # Name                    Version                   Build  Channel
    panel                     0.10.0a2.post9+g643033f.dirty           dev_0    <develop>


However, when I try to run my MWE I get again:

    FileNotFoundError: [Errno 2] No such file or directory: 'C:\\src\\panel\\panel\\dist/panel.min.js'

I have checked and the directory 'dist' is there but the file "panel.min.js" isn't. 

I also tried :  `doit develop_install -c pyviz/label/dev -c conda-forge -c bokeh -o build -o tests -o recommended -o examples`

But I get

    ERROR: File "setup.py" not found. Directory cannot be installed in editable mode: D:\
    TaskFailed - taskid:develop_install 
    Command failed: 'pip install --no-deps --no-build-isolation -e .' returned 1

This sounds like i am not in the right directory, but I surely was in C:\src\panel.

So I am kind of lost again :).

philippjfr:
Do you have bokeh 2.0.2? There was an issue with bokeh 2.0.1, which had issues serving extensions (like Panel) on Windows.

philippjfr:
>I have checked and the directory 'dist' is there but the file "panel.min.js" isn't.

Missed this part, I think you're missing nodejs which is required to compile panel.js. This is one way to install it:

```
doit develop_install -c pyviz/label/dev -c conda-forge -c bokeh -o build -o tests -o recommended -o examples
```

but you could just `conda install nodejs`. To be clear though, which directory were you running this from?

lvankouwen:
I did install nodejs via `conda install nodejs` before pip installing panel. the install got going but it gave me the missing 'panel.min.js' when running code. 

Next I tried running 
    doit develop_install -c pyviz/label/dev -c conda-forge -c bokeh -o build -o tests -o recommended -o examples

From the root of the panel repo (c:\src\panel in my case), which starts to install a bunch of packages, but ends with the message:

    ERROR: File "setup.py" not found. Directory cannot be installed in editable mode: D:\
    TaskFailed - taskid:develop_install
    Command failed: 'pip install --no-deps --no-build-isolation -e .' returned 1

Which really sounds like I am in the wrong directory, but when I try to run the command from some arbitrary dir I get `ERROR: Could not find dodo file 'C:\dodo.py'.` So that is a different error. 

Next thing I did is remove the repo and remove the environment. Cloned the repo again and created the panel_dev environment. From this clean slate I applied the doit command (no manual pip install). This gave the same error as stated above. 

So I am still kind of stuck. I am probably doing something stupid and maybe my testing effort is causing more trouble than it contributes...

philippjfr:
What happens when you run this yourself:

```
pip install --no-deps --no-build-isolation -e .
```

lvankouwen:
```
...
meshio-4.0.12        | 405 KB    | ########## | 100%
Preparing transaction: ...working... done
Verifying transaction: ...working... done
Executing transaction: ...working... b'Enabling notebook extension jupyter-js-widgets/extension...\n      - Validating: ok\n'
done
ERROR: File "setup.py" not found. Directory cannot be installed in editable mode: D:\
TaskFailed - taskid:develop_install
Command failed: 'pip install --no-deps --no-build-isolation -e .' returned 1

(panel_dev) C:\src\panel>pip install --no-deps --no-build-isolation -e .
Obtaining file:///C:/src/panel
    Preparing wheel metadata ... done
Installing collected packages: panel
  Attempting uninstall: panel
    Found existing installation: panel 0.10.0a2.post9+g643033f
    Uninstalling panel-0.10.0a2.post9+g643033f:
      Successfully uninstalled panel-0.10.0a2.post9+g643033f
  Running setup.py develop for panel
Successfully installed panel
```

So it does seem to work. But not really because panel still didn't build correctly (at least panel.min.js is still missing)

philippjfr:
Damn, `python setup.py develop` might provide more information about what's going wrong during the build.

lvankouwen:
Ok I found out what the problem was. I had my terminal configured to start at D:\ by default in stead of C:\. I navigate to "C:\" and start installing and all. Sounds fine right? Well for some reason above my level of understanding this gives a problem when building panel. The hint was given in the error message after running the `doit` command. 

     ERROR: File "setup.py" not found. Directory cannot be installed in editable mode: D:\

I have removed this default, restarted the terminal, and after that the installation just worked.

So, I got back to the original goal of testing the static asset inclusion and that works like a charm, terrific!

philippjfr:
Fabulous, thanks for going through all that!

ceball:
@lvankouwen You have some strange paths appearing, including two drives (c: and d:), plus missing modules you previously installed - so I would _suspect_ a confused (conda) environment, where the python/conda/pip you are running are not the ones you mean to be running, and are not necessarily even all from the same environment.

Without seeing exactly what commands you're running (in their original order, from scratch) along with corresponding full output, debugging is a bit tricky. But some things that might help me:
  * More info from the pip editable install command (so e.g. `pip install --verbose --no-deps --no-build-isolation -e .`
  * You're evidently using windows, but can you say more about your environment? E.g. where are you running the commands - the anaconda prompt, git bash, or what...? (And do you always use the same one, or do you switch between them?)
  * Are you using nested (and possibly stacked) conda environments?
  * The output of `where python`, `where pip` (and `pip --version`), `where conda` (plus `conda info` and `conda --list`).
  * The output of `set`

ceball:
Doh, @lvankouwen answered while I was writing my response. Github needs a "person a is writing..." notification :) You can ignore my post, then. Glad you found your problem.

> I had my terminal configured to start at D:\

Although I guess I know what you mean, please can you say exactly what you mean here (maybe with a screen shot)?

lvankouwen:
You can define in which directory your terminal starts by adding a autorun command in the registry (https://www.ghacks.net/2007/11/16/change-the-default-command-prompt-directory/). 
Don't do this! Or you'll get the kind of problems I was facing. I was using the Anaconda Prompt btw.

ceball:
I think you _ought_ to be able to do that kind of thing.  However, yes, it confuses pyctdev. (And I guess it might confuse some other tools, too.)

pyctdev defines a _doit task_ called "develop install". That doit task is made up of a series of _doit command actions_: "conda install [build deps]", "conda install [specified run deps]", "pip editable install". Each command action is run by doit using `subprocess.Popen()`; pyctdev is not passing a value for `cwd` and requests to `pip install .`, so it's going wrong in your case.

ceball:
Of course, this bug in pyctdev only explains why the default installation instructions did not work for you. The default instructions worked all the way up to the last command (pip editable install), which failed. What I don't get is why when you then ran the pip editable install command manually (which is what Philipp suggested you do), the js wasn't being produced. I guess we'll never know that (adding `--verbose` to that editable install command would probably have told us, though.)

PR #1322, opened 2020-05-05
Add a method to expose Param mapping

### LINKED ISSUE #620: param types vs panel widgets
The docs (https://panel.pyviz.org/user_guide/Param.html) tell us that
> internally Panel maintains a mapping between parameter types and widget types.

I think it would be helpful if we could explicitly list this mapping in the docs, i.e. which widget a given param type is translated to.

maximlt:
Fixes #620

### DIFF
--- a/examples/user_guide/Param.ipynb
+++ b/examples/user_guide/Param.ipynb
@@ -177,7 +177,7 @@
    "source": [
     "## Custom widgets\n",
     "\n",
-    "In the previous section we saw how parameters can automatically be turned into widgets. This is possible because internally Panel maintains a mapping between parameter types and widget types. However, sometimes the default widget does not provide the most convenient UI and we want to provide an explicit hint to Panel to tell it how to render a parameter. This is possible using the ``widgets`` argument to the `Param` pane. Using the ``widgets`` keyword we can declare a mapping between the parameter name and the type of widget that is desired (as long as the widget type supports the types of values held by the parameter type).\n",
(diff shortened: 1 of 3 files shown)


jbednar:
Looks good! Is the mapping that is returned modifiable in place, as an alternative to passing in a partial mapping in certain API calls?  If so I think that should be mentioned in its docstring and in the docs that were modified here.

philippjfr:
I'm also wondering this. IMO we should not make it modifiable directly returning some immutable view of the mapping and providing some API to override it, or we should just make the mapping attribute public.

jbednar:
I'd vote for making the mapping attribute a public Parameter, so that people can easily configure it how they like at the start of a body of code, then forget about it.

MarcSkovMadsen:
Yes. I would need to be able to update the mapping I believe when I have a library of mwc web components widgets that I would like to use instead. But as far as I remember not everything happens in the mapping. There are also code (if statements) determining what happens.

jbednar:
Good point.  I haven't looked at it, but there might be a way to map that logic into the mapping (no pun intended).  For instance, we could define some extra key, value pairs, where the key is a special value, not just a bare Parameter type.  That way the logic can select the special value when special cases are needed, and the static mapping can then statically map that value to a user-selectable type.

xavArtley:
If we make the mapping modifiable it would be nice to also be able to configure the default parameters of widgets in this mapping

maximlt:
Good to see this has generated some interesting discussion 😉 

The original intent was just to display the mapping so that users don't have to read the code to understand why *that* param type was displayed with *that* widget. I wasn't so sure that returning a view on the mapping was a good idea, obviously it was not!

It seems that making the mapping public is the general idea now. Before doing that the first step would be to change the code to remove any internal logic related to the current mapping.

philippjfr:
Superseded by #3173

Issue #33, opened 2018-09-07
Panel requires bounds to be set


kcpevey:
As discussed at our meeting...  
  
`minimum_size = param.Integer(default=10)`  
  
Will produce an un-editable widget (just some text on the screen). This also happens with `param.Number`. I believe it has something to do with None bounds.

jbednar:
When there are no bounds or softbounds, what *should* happen is to get an *editable* widget.  That's because the toolkit has no way of knowing what the left and right edge of the slider should correspond to.  We could make a convention for that case, e.g. to set the lower bound to 1/10th of the default value  (or 10x the default value if negative) and the higher bound to 10X the default value (or 1/10th the default value if negative), but to me that seems like just covering up the fact that we really don't know an appropriate range to show, and people won't realize that it's covering an inappropriate range.  So we *could* do that, but my vote is for simply adding some meaningful guess at the softbounds for the cases when there are no hard bounds.  Even a bad guess at a suitable softbounds by a human is still going to be more appropriate than anything that could be done automatically.

Having the value not be editable is a bug, though -- the value should always be editable for a non-constant parameter.

kcpevey:
>my vote is for simply adding some meaningful guess at the softbounds for the cases when there are no hard bounds. Even a bad guess at a suitable softbounds by a human is still going to be more appropriate than anything that could be done automatically.  
  
I don't think there is no way for us to make a meaningful guess without first knowing the ballpark. If we had a set guess that was outside of the specified default, that would throw an error. I think it will have to be tied to the default and would need to follow something like what you specified here:
> set the lower bound to 1/10th of the default value (or 10x the default value if negative) and the higher bound to 10X the default value (or 1/10th the default value if negative)  
  
Although different widgets will need slight variances of this (float vs Int vs Number).

jbednar:
> I don't think there is no way for us to make a meaningful guess without first knowing the ballpark. If we had a set guess that was outside of the specified default, that would throw an error. 

There shouldn't be an error in that case; that's what makes softbounds soft.  That said, I think Bokeh needs extending before bounds will truly be soft -- at least the last I checked, softbounds are soft for Param but not at the widget level, where there is no way to set the value outside the bounds (even if the hard bounds do allow it).

jbednar:
[...]

PR #1326, opened 2020-05-07
WIP: New SliderInput widget

xavArtley:
I open this PR mostly for discussion and improvements
The aim is to fix https://github.com/holoviz/panel/issues/1128 and https://github.com/holoviz/panel/issues/33

ezgif com-video-to-gif (3)


panel/models/widgets.py
```diff
@@ -127,3 +127,22 @@ class FileDownload(InputWidget):
...
+
+
+class SliderInput(Slider):
+    """
+    Slider associate with an input
+    """
+    tooltips = Override(default=False)
+
+    hard_start = Float(default=None, help="""
+    The minimum allowable value accept by the input.
```
jbednar, line 139:
```suggestion
    The minimum allowable value accepted as input, if any.
```

panel/models/widgets.py
```diff
@@ -127,3 +127,22 @@ class FileDownload(InputWidget):
...
+    """)
+
+    hard_end = Float(default=None, help="""
+    The maximum allowable value accept by the input.
```
jbednar, line 143:
```suggestion
    The maximum allowable value accepted as input, if any.
```

panel/models/widgets.py
```diff
@@ -127,3 +127,22 @@ class FileDownload(InputWidget):
...
+    """)
+
+    input_size = String(default=None, help="""
+    Size of the input in px or %
```
jbednar, line 147:
```suggestion
    Size of the widget in px or %
```
(Not sure about the name; seems like this is widget_size?  By input I'd expect to refer to the numeric value being put in, not the graphical widget.)

panel/widgets/slider.py
```diff
@@ -113,6 +113,24 @@ class FloatSlider(ContinuousSlider):
 
     step = param.Number(default=0.1)
 
+class SliderInput(FloatSlider):
+
+    show_value = param.Boolean(default=False, doc="""
+        Whether to show the widget value.""")
```
jbednar, line 119:
```suggestion
        Whether to show the current value numerically alongside the widget.""")
```

panel/models/widgets.py
xavArtley: Actualy it's really just the size of the numeric input
the size of the overall widget is define by width

xavArtley:
one of my concerns is the number of possibilities is the number of possibilities for the widget layout:
For example a small subset of possibilities

image

jbednar:
Good point!  Of those, the ones I'd personally find useful are:

image
image
image

panel/models/widgets.py
jbednar: Ah! I consider both the slider and the text box to be ways of providing input, and so referring to one part of it as input confused me. :-) 

Maybe `slider_size` (if needed) and `textbox_size`?

jbednar:
Also, I'd prefer the number to be "bare", as if it is merely reporting the value, unless and until someone clicks on it, as ipywidgets does:

image


image
 
Apps and dashboards tend to have a lot of sliders, only one of which is ever being messed with at a given time, and even then only rarely is it used numerically, so the numeric value is far more often used as a readout than as a widget.  So it would be great if the widgetness of it could fade into the background, and only be visible when someone actually tries to mess with it.

xavArtley:
> Also, I'd prefer the number to be "bare", as if it is merely reporting the value, unless and until someone clicks on it, as ipywidgets does:
> 
> image
> 
> image
> 
> Apps and dashboards tend to have a lot of sliders, only one of which is ever being messed with at a given time, and even then only rarely is it used numerically as an input, so the numeric value is far more often used as a readout than as a widget. So it would be great if the widgetness of it could fade into the background, and only be visible when someone actually tries to mess with it.

ezgif com-video-to-gif (5)

jbednar:
Beautiful!  You are a magical wish-granting genie! :-)

Jhsmit:
Would it be a good idea to extend this SliderInput widget to have two fields on either side and a range slider for fine control of ranges?

ahuang11:
Might be impossible in panel, but it'd be nice if the editable value box is directly above the slider:
image

https://github.com/bokeh/bokeh/issues/9943#issuecomment-718968158

PR #1339, opened 2020-05-12
Add a name parameter to the Param pane

### LINKED ISSUE #1332: Enable setting the panel.Param name manually.
#### My Pain

Sometimes I have a `param.Parameterized` class with a lot of parameters. For layout reasons I might split them up into multiple `panel.Param`s. I would like to give these seperate names. 

As far as I can see this is not supported by `panel.Param`.

#### Example

As you can see the name is MyComponent instead of the custom one specified.

image

```python
import param
import panel as pn

class MyComponent(param.Parameterized):
    selection1 = param.ObjectSelector()
    selection2 = param.ObjectSelector()

component = MyComponent()
app = pn.Column(
    pn.Param(component, parameters=["selection1"], name="Primary Selections"),
    pn.Param(component, parameters=["selection2"], name="Secondary Selections"),
)
app.show()
```

#### Solution

Support this.

#### Workaround

As you can see below the layout is not nice. There is too much space (margin I guess) between things.

image

```python
import param
import panel as pn

class MyComponent(param.Parameterized):
    selection1 = param.ObjectSelector()
    selection2 = param.ObjectSelector()

component = MyComponent()
app = pn.Column(
    pn.pane.Markdown("# Primary Selections"),
    pn.Param(component, parameters=["selection1"], show_name=False),
    pn.pane.Markdown("# Secondary Selections"),
    pn.Param(component, parameters=["selection2"], show_name=False),
)
app.show()
```

maximlt:
Fixes #1332 

I also fixed two non related bugs occurring when `.object` would be updated with `None` or with parameters `.param`.

image

### DIFF
--- a/examples/reference/panes/Param.ipynb
+++ b/examples/reference/panes/Param.ipynb
@@ -20,19 +20,20 @@
     "\n",
     "#### Parameters:\n",
     "\n",
-    "The basic parameters are\n",
+    "The basic parameters are:\n",
     "\n",
     "* **`object`** (param.parameterized.Parameters): The `param` attribute of a `param.Parameterized` Class\n",
-    "* **`parameters`** (None or List[str]): A list identifying the subset of parameters to include in the Pane.\n",
+    "* **`parameters`** (List[str]): A list identifying the subset of parameters to include in the Pane.\n",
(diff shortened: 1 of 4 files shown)


philippjfr:
This looks great, thanks for adding those tests as well!

PR #1342, opened 2020-05-13
Fix handling of the Toggle widget in a Param pane

### LINKED ISSUE #1331: pn.param mapping of boolean to toggle fails with param.depends

I am working on the branch "static_serve" on the panel_dev conda environment. 
panel                     0.10.0a2.post9+g643033f           dev_0                 <develop>
param                    1.10.0a2                                      py_0                    pyviz/label/dev
parameterized       0.7.4                                            pyh9f0ad1d_0    conda-forge
(Windows, Chrome browser)

I expect the toggle and the checkbox to work in the same way when rendered via pn.param. However, only the checkbox seems attach the param.depends method to itself. 
No error message occurs. The method "change_toggle" is just never called. 

```
import panel as pn
import param


class BugTester(param.Parameterized):

    bool_toggle = param.Boolean(default=False, label='toggle')
    bool_checkbox = param.Boolean(default=False, label='checkbox')

    def __init__(self, **params):
        super().__init__(**params)

    def view(self):
        widgets = {
            'bool_toggle': pn.widgets.Toggle,
            'bool_checkbox': pn.widgets.Checkbox}
        return pn.Param(self.param, widgets=widgets)

    @param.depends('bool_toggle', watch=True)
    def change_toggle(self):
        print('toggle changed')

    @param.depends('bool_checkbox', watch=True)
    def change_checkbox(self):
        print('checkbox changed')


if __name__ == "__main__":
    bug = BugTester()
    pn.serve(bug.view(), port=5050)
```

maximlt:
Fixes #1331 

The panel widget `Toggle` was not synced like the other widgets in a Param pane:
https://github.com/holoviz/panel/blob/b73ec774734ee7b88ebaf7749707309caeba66b2/panel/param.py#L378-L379
This special handling originated from #15 and was probably implemented this way because the widget that is created when one needs to expand a subparameter pane is a `Toggle`. I believe that now widgets created in a pane are given by `_ordered_params()` which do not expose that `Toggle` widget. So it's just fine to remove that special handling.

I also took advantage of this PR to fix a typo and improve the coverage of the param action test.

### DIFF
--- a/panel/param.py
+++ b/panel/param.py
@@ -375,86 +375,84 @@ class Param(PaneBase):
         widget._param_pane = self
 
         watchers = self._callbacks
-        if isinstance(widget, Toggle):
-            pass
+
+        def link_widget(change):
+            if p_name in self._updating:
+                return
+            try:
+                self._updating.append(p_name)
+                self.object.param.set_param(**{p_name: change.new})
+            finally:
+                self._updating.remove(p_name)
+
+        if isinstance(p_obj, param.Action):
+            def action(change):
+                value(self.object)
+            watcher = widget.param.watch(action, 'clicks')
         else:
-            def link_widget(change):
-                if p_name in self._updating:
-                    return
-                try:
-                    self._updating.append(p_name)
-                    self.object.param.set_param(**{p_name: change.new})
-                finally:
-                    self._updating.remove(p_name)
+            watcher = widget.param.watch(link_widget, 'value')
+        watchers.append(watcher)
 
-            if isinstance(p_obj, param.Action):
-                def action(change):
-                    value(self.object)
-                watcher = widget.param.watch(action, 'clicks')
+        def link(change, watchers=[watcher]):
+            updates = {}
+            if change.what == 'constant':
+                updates['disabled'] = change.new
+            elif change.what == 'precedence':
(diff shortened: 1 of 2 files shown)


philippjfr:
Thanks!

>This special handling originated from #15 and was probably implemented this way because the widget that is created when one needs to expand a subparameter pane is a Toggle.

That is correct.

>I believe that now widgets created in a pane are given by _ordered_params() which do not expose that Toggle widget. So it's just fine to remove that special handling.

That sounds right, but just to be sure do you know if this is currently under test and/or have you tested subobject toggling manually?

maximlt:
Hi @philippjfr and thanks for your feedback,

I've tested it manually and saw nothing wrong.

Some of the current tests emulate a click on the toggle button, see for example:
https://github.com/holoviz/panel/blob/b6cf95f1381c652b3ab60377cd51ff9886d2e029/panel/tests/test_param.py#L666
I didn't have to change them so I assumed everything was alright.

If you'd like me to add another test that's OK. I'd just need some guidance as I don't really know how to check that. Could it be as fine grained as checking that there is just one callback on the toggle button?

philippjfr:
I'm pretty sure it's fine too after looking at it so I'll merge. Thanks again!

PR #1343, opened 2020-05-13
Add TemplatedHTML model and panel layout

philippjfr:
This PR will make it much easier to do templating inside a Panel app by providing a component which allows laying out other components using custom HTML.

The model has an API that looks like this:

```python
TemplatedHTML(html="<div>{0}</div>", children=[Button(label='ABC')])
```

Here the template string `{0}` refers to the first of the children, i.e. the Button model will be inserted at that location of the HTML. This makes it possible to insert other components into arbitrary HTML nodes making it possible to template inside a Panel app. I haven't yet decided on the best Panel API for this component.

- [ ] Implement panel layout component
- [ ] Add tests
- [ ] Add reference gallery entry

Cc: @MarcSkovMadsen @sdc50


jbednar:
Nice! I wonder if the functionality here could be merged with pn.HTML, with templating only if `children` is not None.

It would be nice to have names rather than just numbers for expansion, so that it's more readable.

philippjfr:
>It would be nice to have names {button} rather than just numbers {0} for expansion, so that it's more readable and more explicit about the mapping.

Agreed, I didn't want to change the type of children at first because I was subclassing bokeh's `Box` baseclass but there's no reason to do that. That said we don't necessarily have to do that at the Bokeh level instead emulating it on the Panel object.

MarcSkovMadsen:
This is a game changer. Will be so powerfull. Just made my day.

MarcSkovMadsen:
Is that a way to use it with a parameterized class such that the parameters go into specific places?

philippjfr:
>Is that a way to use it with a parameterized class such that the parameters go into specific places?

Neat idea, could imagine a `html_form` (or similar) parameter on pn.Param that let's you do that.

MarcSkovMadsen:
I updated the suggestion above. Would be so nice. Think about the paramaterized class as a Django Model and the TemplatedHTML as the form of that model.

```python
TemplatedHTML(html="<div>{0}</div><div>{1}</div>", parent=my_param_class, children=["frequency", "update_plot"])
```

MarcSkovMadsen:
And I guess I should still be able to set the `widgets` as well.

```python
TemplatedHTML(html="<div>{0}</div><div>{1}</div>", parent=my_param_class, children=["frequency", "update_plot"], widgets={"update_plot": {"button_type": "primary"}})
```

MarcSkovMadsen:
Ahh. Yes @philippjfr . Setting the `html_form` on `pn.Param` would also be so nice.

philippjfr:
I also like your suggestion too though, maybe we have a completely separate `HTMLForm` layout as you described.

philippjfr:
>I wonder if the functionality here could be merged with pn.HTML, with templating only if children is not None.

My feeling is that this is too much overloading and blurs the distinction between Panes and Panels (i.e. layouts).

jbednar:
> My feeling is that this is too much overloading and blurs the distinction between Panes and Panels (i.e. layouts).

Hmm.  Interesting point.  pn.HTML can already look quite like a layout, in that you can put anything you want in there, including layout-like things.  Conversely, pn.TemplatedHTML can look like a pane, where what you are templating in could be a string title for a single object, just like configuring a pane, or a single block of parameters as proposed above.  So I don't see any clear-cut distinction between pn.HTML and pn.TemplatedHTML in terms of their actual capabilities that would make one be more pane-like and one be more layout-like.

That said, it seems like you want to convey that pn.HTML should be _thought_of_ as more like a pane, and pn.TemplatedHTML should be _thought_of_ as more like a layout, such that we should assume that in many cases what's being templated in is something like a pane?  If so that's a stance that may help position things for readers, though I'd have trouble defending it as something real rather than a rule of thumb.

philippjfr:
I think the distinction is much more clear cut than you make out. One lays out other components, the other does not. However, beyond that, like other layouts the TemplatedHTML component will have an `objects` parameter, can be iterated over, can be indexed, recurses over its contents internally (important for js-linking and other behaviors). If we did this for the HTML component we'd have to add tons of special casing so all these behaviors that are expected for layouts apply there too. So from both an API and internal implementation perspective blurring the distinction seems like a non-starter to me.

jbednar:
Hmm. I'm clearly missing something, because in the code I see `class TemplatedHTML(Box):`, where Box is from bokeh.models, and does not appear to inherit from the Panel class they way a layout would.  I thought you were just claiming that it is _like_ a layout, but it sounds like you are claiming it _is_ a layout? How does it get to be a layout without inheriting from Panel (i.e. where does it get its objects parameter, etc.?)?  If it does act just like a layout, then sure, very different from HTML; I'm just missing where that's happening.

philippjfr:
I haven't yet implemented the Panel component(s) wrapping the underlying bokeh model.

jbednar:
Ah, I see. In your head it's about to become a layout, but it's not in the codebase yet. Fair enough; carry on! :-)

philippjfr:
>But note that the title of this PR only promises a pane, not a layout...

Oops 😊

philippjfr:
Will be subsumed by #1894, closing.

PR #1344, opened 2020-05-14
Add Widget.from_param classmethod

philippjfr:
Simplifies the creation of a widget linked to and created from a particular parameter.

```python
import param

class A(param.Parameterized):
    
    value = param.Number()
    
a = A()

pn.widgets.FloatSlider.from_param(a.param.value, start=0, end=3.14)
```

@maximlt @MarcSkovMadsen Any thoughts on this? It's **a** solution to something I think you both brought up before. If you like the idea maybe we should also consider a shorter method name than `from_parameter`, suggestions welcome.

### DIFF
--- a/examples/user_guide/Param.ipynb
+++ b/examples/user_guide/Param.ipynb
@@ -226,6 +226,22 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
+    "However it is also possible to explicitly construct a widget from a parameter using the `.from_param` method, which makes it easy to override widget settings using keyword arguments:"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "pn.widgets.IntSlider.from_param(Example.param.unbounded_int, start=0, end=100)"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
     "## Custom name\n",
     "\n",
     "By default, a param Pane has a title that is derived from the class name of its `Parameterized` object. Using the ``name`` keyword we can set any title to the pane, e.g. to improve the user interface."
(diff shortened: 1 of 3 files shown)


jbednar:
> If you like the idea maybe we should also consider a shorter method name than from_parameter, suggestions welcome.

I got as far as the title on this page before my brain went (a) nice! and (b) can we have a shorter name? :-)  How about just `from`, currently supporting Parameters but e.g. supporting a traitlet or some other source potentially?  You could maybe also consider it another type of linking, and make `.link()` be either an instance or class method, where it does this when called on the class.

maximlt:
I don't remember if I ever brought this up but I'll still give you my 2 cents!

This
```python
pn.widgets.FloatSlider.from_parameter(a.param.value, start=0, end=3.14)
```

is better than
```python
pn.Param(a.param.value, widgets={"value": {"type": pn.widgets.FloatSlider, "start":0, "end":3.14}})
```

That's even better :-)
```python
pn.widgets.FloatSlider.from(a.param.value, start=0, end=3.14)
```

As for the implementation, is it intended that it returns a `Param` pane and not a widget? I guess I'd expect a widget. It'd make it clearer that this method is intended to create a single widget linked to a parameter and that it would not overlap (too much) with what can already be achieved with the `Param` pane. It can indeed be finely customized if one sets the `precedence` of the parameters and/or uses the keyword args `parameters` and/or `widgets`. I have sometimes the feeling that there are too many ways to achieve the same thing in `panel`, just trying to avoid that.

philippjfr:
>As for the implementation, is it intended that it returns a Param pane and not a widget?

No, I think it should return the widget as you point out.

philippjfr:
Oh and `.from` won't work since it's a reserved name in Python, any other suggestions?

maximlt:
Arf missed that. I like that the method starts with `from`, I'm sometimes confused with `.link` (that I don't use much TBH), by mistake I could write `a.param.value.link(pn.widgets.FloatSlider, start=0, end=3.14)`...
*Explicit is better than implicit*, `from_param` ? (+ `from_trait` later on?)

philippjfr:
Yeah, I'm also leaning towards `.from_param`,  not a huge fan of overly generic methods.

PR #1349, opened 2020-05-15
Fix a bug in the clone method of the layouts

### LINKED ISSUE #1329: GridBox ncols not honored when rendering from a callable
#297 ## ALL software version info
```
In [3]: param.__version__                                                                                                                                                                                                                                                                                                                                               
Out[3]: '1.10.0a2'

In [4]: pn.__version__                                                                                                                                                                                                                                                                                                                                                  
Out[4]: '0.10.0a2'
```

When a `GridBox` is first rendered dynamically (from a callable) then it first renders with only one column. After updating then the `ncols` parameter is respected. 

```
import random
import param
import panel as pn
pn.extension()

    ncols = param.Integer(default=6)
    
    @param.depends('ncols')
    def grid(self):
        rcolor = lambda: "#%06x" % random.randint(0, 0xFFFFFF)
        return pn.GridBox(*[pn.pane.HTML(background=rcolor(), width=50, height=50) for i in range(24)], ncols=self.ncols)
    
    def panel(self):
        return pn.Column(
            self.param.ncols,
            self.grid,
        )
t = Test()
t.panel()

```
![GridBox_ncols](https://user-images.githubusercontent.com/6454472/81433463-35790900-912a-11ea-8271-b2eefb53e2cc.gif)

maximlt:
Fixes #1329 

Which was caused by a tiny bug in the `clone` method.

### DIFF
--- a/panel/layout/base.py
+++ b/panel/layout/base.py
@@ -247,7 +247,7 @@ class ListLike(param.Parameterized):
                              % type(self).__name__)
         p = dict(self.param.get_param_values(), **params)
         del p['objects']
-        return type(self)(*objects, **params)
+        return type(self)(*objects, **p)
 
     def append(self, obj):
         """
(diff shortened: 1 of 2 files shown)


philippjfr:
Amazing, I was very confused by what could cause that issue.

PR #1353, opened 2020-05-18
Player visual indicator of button pressed

### LINKED ISSUE #1334: Visual indicator that Player is running
The Player widget should provide a visual affordance when the forward or back play buttons are active.

Currently the forward/back play buttons flash when you press them. Instead those buttons should probably act like toggles. Or perhaps the play and pause buttons should act as a radio button group.

xavArtley:
fix #1334
ezgif com-video-to-gif (6)


philippjfr:
Fabulous, thanks @xavArtley!

PR #1355, opened 2020-05-18
rename widget-box -> panel-widget-box

basnijholt:
Because it overwrites the style of ipywidgets.

https://github.com/jupyter-widgets/ipywidgets/blob/3bd75276370271e43ca5aa38d59a1848dfb65fb7/packages/controls/css/widgets-base.css#L99

Because of this code, running `holoviews.notebook_extension()` overwrites existing css that changes an `ipywidget` from:
image
to
Screenshot 2020-05-18 at 23 55 00

The annoying thing is that this change persists over kernel restarts *and* clearing of the notebook.

### DIFF
--- a/examples/gallery/demos/VTKSlicer.ipynb
+++ b/examples/gallery/demos/VTKSlicer.ipynb
@@ -156,7 +156,7 @@
     "    pn.Column(dataset_selection, toggle_parallel_proj, *volume_controls[1:], sizing_mode='fixed'),\n",
     "    pn.Param(smoother, parameters=['smooth_level', 'order']),\n",
     "    pn.layout.VSpacer(),\n",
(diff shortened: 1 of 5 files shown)


jbednar:
Sounds like a good idea, though I can't actually spot the difference between your two screenshots, unless you mean the size.

basnijholt:
The alternative gray-white rows in the first screenshot disappear and the widget also becomes larger.

jbednar:
Ah, I do see the gray changes (very subtle on my monitor), but I'll have to take your word on the size changes, given the large difference in the screenshot pixel widths and thus displayed sizes. :-)

philippjfr:
Seems fine, thanks for the fix @basnijholt!

Issue #1352, opened 2020-05-18
Add warning when Panel passed method/function with watch=True


philippjfr:
Probably the most common confusion I see among Panel users is that they add `watch=True` to `param.depends` when they don't need to. Since this is pretty much never a sensible thing to do when passing a method or function to Panel, I'm considering whether we should just warn when a user gives a method/function which has watch=True on it to Panel.

maximlt:
That's a great idea, it took me a while before I managed to understand that concept. When you see that `watch=True` option you just naturally want to use it because without it you may think the callback you set doesn't do anything at all. What made me click is when I read somewhere that `watch=True` was best suited for methods that have side effects. Do not hesitate to provide a rather long warning message that explains clearly why `watch=True` isn't needed when the method/function explicitly returns something.

jbednar:
I think the underlying source of confusion here is that Param only actually implements a subset of what it lets you declare. E.g. you can declare a `default` and `bounds` for Parameters, and Param will respect and enforce those directly. , but it won't do anything with `soft_bounds` or `step`; those are just hints left around for a GUI to use if it supports them.  

`@depends` is used both ways; by default it merely records dependency information (which Panel can then inspect and use), but if you also say `watch=True` Param actively re-runs the method itself, independently of Panel. That makes sense for something only used for its side effects, but if it's returning a value, Param has nowhere to put the value.  Panel does know what to do with a return value (i.e. display it), so if you provide such a method to Panel, it doesn't make sense for Param to also be invoking that method when a Parameter on which it depends changes; Panel will already be doing that, and moreover will be doing something with the output, making it be called twice with one of the return values discarded (surely an error!).

So  at least as far as Param and Panel alone are concerned, it is difficult to see any middle ground; `watch=True` makes sense for side-effecty methods with no return value, and the default `watch=False` makes sense for methods Panel calls to update a display.

So I do think it makes sense for Panel to warn when given a method or function with `watch=True` dependencies; that's almost surely (always?) an error.
 
What's less clear to me is whether `watch=True` _ever_ has valid uses for methods returning a value, ignoring Panel.  At the Param level, are there any cases where we want Param to invoke the method and then discard the output?  It's hard to see why that would ever be desirable. If not presumably the warning or error should be at the Param level, not the Panel level.

maximlt:
Thanks for having included that long warning message, it's gonna help a lot of people for sure!

MarcSkovMadsen:
[...]

PR #1356, opened 2020-05-19
Add warning if method or function with watch=True is provided

philippjfr:
Implements https://github.com/holoviz/panel/issues/1352

Cc: @jbednar @maximlt Please review the warning and chime in if you have clarifications.

### DIFF
--- a/panel/param.py
+++ b/panel/param.py
@@ -580,11 +580,13 @@ class ParamMethod(ReplacementPane):
                            f"a {fn_type} for panel to display you do not "
                            "have to set `watch=True` since panel will "
                            "automatically watch for changes in the "
-                           "dependencies. The use of `watch=True` "
-                           f"should be reserved for {fn_type}s which "
-                           "work via side-effects, e.g. by modifying"
-                           "internal state of a class or global state "
-                           "in an application's namespace.")
+                           "dependencies. If `watch=True` is set the "
+                           f"{fn_type} will be called twice for every "
+                           "change in one of the dependencies. The use of "
+                           f"`watch=True` should therefore be reserved "
+                           "for {fn_type}s which work via side-effects, "
+                           "e.g. by modifying internal state of a class "
+                           "or global state in an application's namespace.")
 
     #----------------------------------------------------------------
     # Callback API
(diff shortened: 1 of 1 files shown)


panel/param.py
```diff
@@ -566,8 +566,26 @@ def __init__(self, object=None, **params):
...
+        self.param.warning(f"The {fn_type} supplied to panel was found "
+                           "to have set `watch=True`. When providing "
+                           f"a {fn_type} for panel to display you do not "
+                           "have to set `watch=True` since panel will "
+                           "automatically watch for changes in the "
+                           "dependencies. The use of `watch=True` "
+                           f"should be reserved for {fn_type}s which "
+                           "work via side-effects, e.g. by modifying"
+                           "internal state of a class or global state "
+                           "in an application's namespace.")
```
jbednar, line 587: I think it should warn that the fn_type will get called twice each time the dependency changes. I don't think that's clear yet.

panel/param.py
```diff
@@ -566,8 +566,28 @@ def __init__(self, object=None, **params):
...
+                           "have to set `watch=True` since panel will "
+                           "automatically watch for changes in the "
+                           "dependencies. If `watch=True` is set the "
+                           f"{fn_type} will be called twice for every "
+                           "change in one of the dependencies. The use of "
+                           f"`watch=True` should therefore be reserved "
+                           "for {fn_type}s which work via side-effects, "
+                           "e.g. by modifying internal state of a class "
+                           "or global state in an application's namespace.")
+
```
jbednar, line 590:
```suggestion
        self.param.warning(f"The {fn_type} supplied for Panel to display "
                           "was declared with `watch=True`, which will "
                           f"cause the {fn_type} to be called twice for "
                           "any change in a dependent Parameter. "
                           "`watch` should be False when Panel is "
                           "responsible for displaying the result "
                           f"of the {fn_type} call, while `watch=True` "
                           f"should be reserved for {fn_type}s that work "
                           "via side-effects, e.g. by modifying internal  "
                           "state of a class or global state in an "
                           "application's namespace.")

```

PR #1374, opened 2020-05-29
Ensure non-existent models in Template do not cause errors

philippjfr:
Certain components do not render their models immediately instead preferring to do so lazily. The Template however expects all models to be available on initial render. This PR ensures that components that have not yet been instantiated are simply skipped avoiding any resultant errors.


MarcSkovMadsen:
Does that mean you can have a template with `{{ panel_10 }}` but not `add_panel` with name `panel_10`? That would be nice.

philippjfr:
No, totally unrelated to that but please open an issue about that.

Issue #1375, opened 2020-05-31
Improve panel.Param alignment of inputs and actions in row


MarcSkovMadsen:
#### My Pain

I think `panel.Param` is awesome. It makes it so easy layout parameters of `param.Pameterized` classes without having to spend a lot of time on it.

But for example when implementing the `DE:TE: Object Detection` app at awesome-panel.org I had to spend some time on find a workaround to align the below row.

![image](https://user-images.githubusercontent.com/42288570/83345934-7a9eee00-a318-11ea-8ade-61b83e2288b0.png)

#### Minimal Reproducing example

The problem is that using just one `panel.Param` as in the below code example looks like

![image](https://user-images.githubusercontent.com/42288570/83346004-306a3c80-a319-11ea-8a9b-79dee46cd036.png)

```python
import param
import panel as pn


class DETRApp(param.Parameterized):
    input_image_url = param.String("https://panel.holoviz.org/_static/logo_stacked.png", label="Input Image URL")
    run_detr = param.Action(label="Run DE:TR:")
    set_random_image = param.Action(label="Random Image")

    view = param.Parameter()

    def __init__(self, **params):
        params["view"] = self._get_view()
        super().__init__(**params)

    def _get_view(self):
        top_selections = pn.Row(
            pn.Param(
                self,
                parameters=["input_image_url", "run_detr", "set_random_image"],
                widgets={"set_random_image": {"button_type": "success"}},
                default_layout=pn.Row, show_name=False,
                width=900,
                )
        )
        app_view = pn.Column(
            top_selections,
        )
        return app_view

DETRApp().view.servable()
```

#### Solution

Improve the layout capabilities of `panel.Param` to align inputs with label with buttons.

#### Work Around

![image](https://user-images.githubusercontent.com/42288570/83345952-b2a63100-a318-11ea-8db6-6692c361979e.png)

```python
import param
import panel as pn


class DETRApp(param.Parameterized):
    input_image_url = param.String("https://panel.holoviz.org/_static/logo_stacked.png", label="Input Image URL")
    run_detr = param.Action(label="Run DE:TR:")
    set_random_image = param.Action(label="Random Image")

    view = param.Parameter()

    def __init__(self, **params):
        params["view"] = self._get_view()
        super().__init__(**params)

    def _get_view(self):
        top_selections = pn.Row(
            pn.Param(self, parameters=["input_image_url"], default_layout=pn.Row, show_name=False,),
            pn.Param(
                self,
[...]

PR #1376, opened 2020-05-31
End align widgets in Param Row layout

xavArtley:
fix: #1375

### DIFF
--- a/panel/param.py
+++ b/panel/param.py
@@ -344,10 +344,6 @@ class Param(PaneBase):
         # Update kwargs
         kw.update(kw_widget)
 
-        # Set widgets align to end in a Row layout
-        if self.default_layout == Row and 'align' not in kw:
-            kw.update({'align': 'end'})
-
         if hasattr(p_obj, 'get_range'):
             options = p_obj.get_range()
             if not options and value is not None:
--- a/panel/param.py
+++ b/panel/param.py
@@ -344,6 +344,10 @@ class Param(PaneBase):
         # Update kwargs
         kw.update(kw_widget)
 
+        # Set widgets align to end in a Row layout
+        if self.default_layout == Row and 'align' not in kw:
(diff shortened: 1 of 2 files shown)


xavArtley:
was wondering if we should not align all widgets to the end in a Row layout
image

philippjfr:
>was wondering if we should not align all widgets to the end in a Row layout

That's a good question and it seems like a good idea to me.

MarcSkovMadsen:
There is a similar problem as far as i remember if the default layout is a gridspec or gridbox.

xavArtley:
May be the default align of widgets should be 'end' no matter the layout?

philippjfr:
>May be the default align of widgets should be 'end' no matter the layout?

Could you add a commit to the `Widget` baseclass with, defining

```python
    align = param.ObjectSelector(default='end',
                                 objects=['start', 'end', 'center'], doc="""
        Whether the object should be aligned with the start, end or
        center of its container""")
```

and then append `[doc-build]` to the commit message? Then we can see what that would look like.

xavArtley:
Setting widgets align to end by default:
image

xavArtley:
The main issue of align is it aligns both horizontal and vertically the widget layout:

image

xavArtley:
To make a widget bottom left align we must define a custom class and some raw css:
image

MarcSkovMadsen:
You’re so cool. 

Thought this was just one of the friction points I would have to live with 👍

philippjfr:
>The main issue of align is it aligns both horizontally and vertically the widget layout:

Hmm, that's a good point, the CSS solution seems okay but possibly isn't ideal. Will have to think about it a bit.

philippjfr:
Closing for now in favor of https://github.com/holoviz/panel/pull/1504. May have to fall back to the CSS solution for grids though.

PR #1383, opened 2020-06-04
Add support for hierarchical multi-indexed DataFrame

philippjfr:
This PR improves the `DataFrame` widget in a two ways relating to support for DataFrame indexes:

1. It fixes a bug editing DataFrames with a non-default index name
2. It adds support for displaying hierarchical multi-indexed dataframes with aggregators

As an example here is a three-level DataFrame aggregated over the indexes:

```python
from bokeh.sampledata.population import data as population_data 

pop_df = population_data[population_data.Year == 2020].set_index(['Location', 'Sex', 'AgeGrp'])[['Value']]

pn.widgets.DataFrame(value=pop_df, hierarchical=True, aggregators={'Sex': 'sum', 'Location': 'sum'}, height=400) 
```

Screen Shot 2020-06-04 at 3 10 24 PM

### DIFF
--- a/examples/reference/widgets/DataFrame.ipynb
+++ b/examples/reference/widgets/DataFrame.ipynb
@@ -16,7 +16,7 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "The ``DataFrame`` widget allows displaying and editing a pandas DataFrame.\n",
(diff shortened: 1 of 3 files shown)


jbednar:
Cool!

PR #1389, opened 2020-06-09
Improvements for ipywidgets support

philippjfr:
Previously we had supported using ipywidgets in the notebook (using the regular Jupyter CommManager) and on the server using ipywidgets_bokeh. I had however overlooked the case where you start a server inside the notebook, e.g. when using `.show()`. This PR provides a custom IPyKernel to be used in such scenarios, when rendering a plot with `.show()` the kernel on each Comm is updated to point to this custom kernel allowing communication on the server even when it has been launched inside a notebook.

### DIFF
--- a/panel/io/ipywidget.py
+++ b/panel/io/ipywidget.py
@@ -1,6 +1,5 @@
 import logging
 
-from bokeh.io import curdoc
 from bokeh.document.events import MessageSentEvent
 from ipykernel.comm import CommManager
 from ipywidgets_bokeh.kernel import BokehKernel, SessionWebsocket, WebsocketStream
--- a/panel/io/ipywidget.py
+++ b/panel/io/ipywidget.py
@@ -1,0 +1,72 @@
+import logging
+
+from bokeh.io import curdoc
+from bokeh.document.events import MessageSentEvent
+from ipykernel.comm import CommManager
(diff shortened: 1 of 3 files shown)


panel/io/ipywidget.py
```diff
@@ -0,0 +1,71 @@
...
+    def __init__(self, *args, **kwargs):
+        self._document = kwargs.pop('document', None)
+        self._queue = []
+        super(PanelSessionWebsocket, self).__init__(*args, **kwargs)
+
+    def send(self, stream, msg_type, content=None, parent=None, ident=None, buffers=None, track=False, header=None, metadata=None):
+        msg = self.msg(msg_type, content=content, parent=parent, header=header, metadata=metadata)
+        msg['channel'] = stream.channel
+
+        doc = self._document
```
mattpap, line 19: I would add:
```py
@property
def document(self):
    return curdoc()
```
in `ipywidgets_bokeh` and override it here with `self._document`. Similar thing can be done with even trigger, and all this code duplication could be avoided.

philippjfr: Right I was going to make a PR on ipywidgets_bokeh to reduce code duplication.

philippjfr:
@mattpap I've pushed this PR: https://github.com/bokeh/ipywidgets_bokeh/pull/13/ which will allow me to reduce code duplication here. For now I'll merge and will remove the code duplication once a new ipywidgets_bokeh release is available.

PR #1395, opened 2020-06-10
Adding golden layout

kebowen730:
The goal of this PR is to add a Golden Layout template to panel. So far I have the content and input sections displaying and functioning, but the header is having issues displaying, particularly the sidebar button. I'm using the same css for that section as the bootstrap template, so I'm assuming that it's something to do with GoldenLayout. I haven't found a solution for this issue yet, this is what the dark them currently looks like:

Screen Shot 2020-06-10 at 12 40 02 AM

There might an issue related to the goldenlayout style sheet, I've been running into this error when I run it locally:

Screen Shot 2020-06-09 at 7 07 27 PM


philippjfr:
The problem is that you're using a bunch of CSS classes that come from either material or bootstrap and do not exist when those are not imported.

panel/template/goldenlayout/golden_layout.html
```diff
@@ -0,0 +1,67 @@
...
+{% block postamble %}
+<script type="text/javascript" src="//code.jquery.com/jquery-1.11.1.min.js"></script>
+<script type="text/javascript" src="//golden-layout.com/files/latest/js/goldenlayout.min.js"></script>
+<link type="text/css" rel="stylesheet" href="//golden-layout.com/files/latest/css/goldenlayout-base.css" />
+<link type="text/css" rel="stylesheet" href="//golden-layout.com/files/latest/css/goldenlayout-dark-theme.css" />
+{% endblock %}
+
+<!-- goes in body -->
+{% block contents %}
+<div class="container-fluid d-flex flex-column vh-100 overflow-hidden" id="container">
```
philippjfr, line 13: Each one of these CSS classes come from bootstrap and do nothing here, same with pretty much every class you use below.

Issue #1259, opened 2020-04-11
Add option to toggle throttled on widgets with interact and depends


Hoxbro:
#### Is your feature request related to a problem? Please describe.
I often want to use the `value_throttled` for my dashboard´s widgets and the only way to do it right now is the hard way by adding an explicit widget instance to `pn.interact` or `pn.depends`. 

#### Describe the solution you'd like
Either add a keyword like `throttled` to `pn.interact` and `pn.depends` or create a new function which use the `value_throttled` if it exist for the widget e.g. `pn.interact_throttled`.

#### Additional context
Here are examples of `pn.interact` and `pn.depends` with a `throttled` keyword:
![interact_example](https://user-images.githubusercontent.com/19758978/79048694-2e400780-7c1f-11ea-81ad-57a541e6a5f4.gif)

![ezgif com-video-to-gif](https://user-images.githubusercontent.com/19758978/79048762-a3abd800-7c1f-11ea-9a6c-9bced12f8156.gif)

Hoxbro:
I have got an working example of the feature request mentioned above by adding the keyword throttled to the functions `pn.interact` and `pn.depends`. I have changed the files panel/interact.py and panel/depends.py to get this to work. I also needed to change panel/widgets/sliders.py so it solves the issues mentioned in #1253 and #1256.  I have attached the [files](https://github.com/holoviz/panel/files/4470734/files.zip) changed.  Code used above: 

```
import panel as pn
pn.extension()

func = lambda x: x
pn.interact(func, x=1, throttled=True)
```

```
import panel as pn
pn.extension()

int_range_slider = pn.widgets.IntRangeSlider(end=10, value=(2, 10), value_throttled=(7, 9))

@pn.depends(int_range_slider, throttled=True)
def print_slider_value(slider_value):
    return slider_value

pn.Column(int_range_slider, print_slider_value)
```

Hoxbro:
Any updates on this enhancement?

philippjfr:
If there is any progress I'll record it here, if this is a priority for you I'd be happy to guide you through contributing a PR.

Issue #1256, opened 2020-04-10
Set `value_throttled` on instantiation of widgets


joelostblom:
From https://discourse.holoviz.org/t/how-to-throttle-values-on-mouseup/456/3?u=joelostblom

This works (no value throttling):

```python
int_range_slider = pn.widgets.IntRangeSlider()

@pn.depends(int_range_slider.param.value, watch=True)
def print_slider_value(slider_value):
    return slider_value[0]

pn.Column(int_range_slider, print_slider_value)
```

To get throttling I tried changing `int_range_slider.param.value` to `int_range_slider.param.value_throttled`, but this raises `TypeError: 'NoneType' object is not subscriptable`. This can be fixed by specifying `value_throttled=(3, 4)` (or any number) when initializing the slider (although it was not needed to set `value=` when using `int_range_slider.param.value`).

```
-----
panel       0.9.5
param       1.9.3
sinfo       0.3.1
-----
IPython             6.5.0
jupyter_client      5.2.3
jupyter_core        4.6.3
jupyterlab          2.1.0
notebook            5.6.0
-----
Python 3.7.6 | packaged by conda-forge | (default, Mar 23 2020, 23:03:20) [GCC 7.3.0]
Linux-5.6.2-arch1-2-x86_64-with-arch
4 logical CPU cores
-----
Session information updated at 2020-04-08 12:00
```

philippjfr:
This has been addressed.

Issue #1253, opened 2020-04-09
DiscreteSliders value_throttled does not work


Hoxbro:
#### ALL software version info
panel 0.9.5

#### Description of expected behavior and the observed behavior
DiscreteSlider has a value_throttled attribute but it does not work. I can see in your [documentation](
https://panel.holoviz.org/reference/widgets/DiscreteSlider.html#widgets-gallery-discreteslider) that DiscreteSlider does not have value_throttled. I don´t know if DiscreteSlider should have value_throttled or not, but if it should have it I would expect it to work like IntSlider.

#### Complete, minimal, self-contained example code that reproduces the issue

```
import panel as pn

func = lambda x: x

widget_1 = pn.widgets.DiscreteSlider(options={"x": 0, "y": 1, "z": 2})
widget_1.value_throttled = widget_1.value

widget_2 = pn.depends(widget_1)(func)
widget_3 = pn.depends(widget_1.param.value_throttled)(func)
widgets_discrete = pn.Column("### DiscreteSlider", widget_1, widget_2, widget_3)

widget_1 = pn.widgets.IntSlider(start=0, end=2, value=0)
widget_1.value_throttled = widget_1.value

widget_2 = pn.depends(widget_1)(func)
widget_3 = pn.depends(widget_1.param.value_throttled)(func)
widgets_int = pn.Column("### IntSlider", widget_1, widget_2, widget_3)

pn.Row(widgets_discrete, widgets_int).servable()
```

#### Screenshots or screencasts of the bug in action
![ezgif-1-2a6bdaaaa155](https://user-images.githubusercontent.com/19758978/78917256-eab69380-7a8e-11ea-9e54-92b9e2c749f1.gif)

PR #1400, opened 2020-06-11
Throttled argument for sliders in pn.depends and pn.interact

Hoxbro:
The goal of this PR is to discuss the option to add a parameter to `pn.depends` and `pn.interact` which throttles the slider widgets as mention in #1259. 

To get this to work I also needed to change panel/widgets/sliders.py so it solves the issues mentioned in #1253 and #1256. At the same time i have updated the documentation for sliders with value_throttled.

Missing tasks:
 - [ ] Discussion if this is the best implementation or the implementation is needed.
 - [ ] Keyword argument to be used to enable value_throttled on sliders. Right now it is `throttled` but i feel like `wait` is maybe a more clear.
 - [ ] Create tests for the chosen implementation.
 - [ ] Update documentation with the new implementation.
 - [ ] Other?


panel/interact.py
```diff
@@ -196,7 +197,13 @@ def update_pane(change):
                 self._pane = panel(new_object)
                 self._inner_layout[0] = self._pane
 
-            pname = 'clicks' if name == 'manual' else 'value'
+            if self.throttled is True and hasattr(widget, 'value_throttled'):
```
philippjfr, line 200:
```suggestion
            if self.throttled and hasattr(widget, 'value_throttled'):
```

panel/widgets/slider.py
```diff
@@ -241,6 +258,20 @@ def _update_value(self, event):
...
+            self.value_throttled = values[0]
+            return
+        index = self.values.index(self.value_throttled)
+        if self._syncing_throttled:
+            return
+        try:
+            self._syncing_throttled = True
+            self._slider.value = index
+        finally:
+            self._syncing_throttled = False
```
philippjfr, line 273: Can you explain the purpose behind this method?

panel/widgets/slider.py
```diff
@@ -270,6 +301,15 @@ def _sync_value(self, event):
...
             self._syncing = False
 
+    def _sync_value_throttled(self, event):
+        if self._syncing_throttled:
+            return
+        try:
+            self._syncing_throttled = True
+            self.value_throttled = self.values[event.new]
+        finally:
+            self._syncing_throttled = False
```
philippjfr, line 311: And this one?

panel/depends.py
```diff
@@ -25,8 +29,8 @@ def param_value_if_widget(arg):
     return arg
 
 
-def depends(*args, **kwargs):
-    updated_args = [param_value_if_widget(a) for a in  args]
-    updated_kwargs = {k: param_value_if_widget(v) for k, v in kwargs.items()}
+def depends(*args, throttled=False, **kwargs):
```
philippjfr, line 32: Little bit worried that this will clash with a user provided keyword called throttled, currently don't have a better suggestion though.

philippjfr:
Thanks for this! It mostly looks good but I'm not totally following the reasoning behind the update/sync value_throttled methods so at minimum they need some docstrings.

panel/widgets/slider.py
Hoxbro: I'm properly not understanding this correct, but the method is almost the same as the method above it `_update_values()` with no-throttled variables changed to throttled variables.

panel/depends.py
Hoxbro: I am not the biggest fan of it either. Maybe it could be changed to be sat globally by something like `pn.extension(slider_wait=True)`?

panel/widgets/slider.py
philippjfr: Ah, in that case it's my fault for not properly documenting that code :)

panel/depends.py
philippjfr: I do think you want control over this per instance and technically you can already achieve this with `pn.depends` by passing in the `widget.param.value_throttled` param explicitly, so maybe in that case it's just a matter of properly documenting that?

Hoxbro: You are right that their need to be proper documentation. What about a list of reserved keywords for `pn.interact` and `pn.depends`. I properly need a little help in finding where I should document it… I would think that it needs to be documented in a notebook file(s), can you help me here?

I was also looking at the error message for watch and I think there should be some overlap between the error messages for the keyword `watch` and a keyword like `throttled`. I have attached a pdf file with the different error messages for `watch` and `throttled` for `pn.interact` and `pn.depends`.

philippjfr:
Sorry this has been sitting for a while. It's definitely going to be included in the next minor release (0.10.0).

Hoxbro:
I have time to work with this PR, but I need some kind of guidance in how to move forward with it.

Hoxbro:
@philippjfr can you change the milestone to v0.11.0?

philippjfr:
Will do, and really sorry I didn't get back to this. It was on my list but at some point I had to give up. I'll figure out the way forward here and ping you.

philippjfr:
Also I want 0.11 to happen in about a month. No more 6 month periods between minor releases

Hoxbro:
@philippjfr I have done some bug fixes in this PR, do you see any problem in splitting this PR up into two? One for the bug fixes and one for the throttled argument.

philippjfr:
That sounds good to me.

PR #1424, opened 2020-06-18
Fixed compatibility with Bokeh 2.1

philippjfr:
There was a regression because bokeh added a `invalidate_cache` method on the internal layout class which we override for the Markup models.


mattpap:
> `this.layout = new CachedVariadicBox(this.el, this.model.sizing_mode, changed)`

It's interesting, because this line shouldn't compile. `MarkupView.layout` is defined as bokeh's `CachedVariadicBox`, which is structurally different from panel's.

PR #1450, opened 2020-06-22
Accept MethodType as panel.serve parameter

### LINKED ISSUE #1449: panel.serve only accepts functions, not methods
#### panel version info
0.9.6

```
server = MyServer()
panel.serve(server.gui_instance)
```
Should serve a template returned by `MyServer.gui_nstance()`, but it only produces a `>` int he browser and no error message.

Replaced by the following, it produces the expected behavior:
```
server = MyServer()
def instance():
    return server.gui_instance()
panel.serve(instance)
```

nritsche:
Closes #1449

### DIFF
--- a/panel/io/server.py
+++ b/panel/io/server.py
@@ -11,7 +11,7 @@ import uuid
 
 from contextlib import contextmanager
 from functools import partial
-from types import FunctionType
+from types import FunctionType, MethodType
 
 from bokeh.document.events import ModelChangedEvent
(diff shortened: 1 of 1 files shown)


nritsche:
I didn't find an easy way to add a test for this, but it makes my panel application work.

philippjfr:
Thanks! Don't worry about the test for now, I'll also have to set a higher threshold on CodeCov.

PR #1467, opened 2020-07-08
Matplotlib Link in Readme.md is broken, it is updated.

earik87:
Hi! 

I noticed that Matplotlib link was broken in readme.md. I updated it!


philippjfr:
Much appreciated, thank you!

PR #1469, opened 2020-07-08
Add support for interactive mode on Matplotlib

philippjfr:
Adds support for interactive Matplotlib plots based on ipympl.

Screen Shot 2020-07-08 at 7 05 52 PM

### DIFF
--- a/examples/reference/panes/Matplotlib.ipynb
+++ b/examples/reference/panes/Matplotlib.ipynb
@@ -20,6 +20,7 @@
     "#### Parameters:\n",
     "\n",
     "* **``dpi``** (int, default=144): The dots per inch of the exported png\n",
+    "* **``interactive``** (boolean, default=False): Whether to use the interactive ipympl backend\n",
     "* **``tight``** (bool, default=False): Automatically adjust the figure size to fit the subplots and other artist elements.\n",
     "* **``object``** (matplotlib.Figure): The Matplotlib Figure object to display\n",
     "\n",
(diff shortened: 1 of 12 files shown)


MarcSkovMadsen:
I thought the ipywidgets where costly to load and as such native bokeh/ panel implementations where preferenced?

jbednar:
Sure, but there are plenty of plots available from Matplotlib not available anywhere else, plus some people just prefer Matplotlib. Panel has no opinion about what plotting library you use, though of course each one comes with tradeoffs...

philippjfr:
Also this is entirely optional, I'm just giving users access to this capability.

MarcSkovMadsen:
Its not that i dont like. I do. I’m just trying to understand when a native implementation is required and when ipywidgets version Can be used. 

There Probably are a lot of things that Can be done via ipywidgets.

Ivuetify for example. 

Would it be ok to give users easy access to these in Panel via Param or would it be something you can do but really should not do? Instead bokeh models should be built?

jbednar:
Sure. I'd say that there is a large class of things that can very easily be made available to Panel users in some usable form. Some of those could be made much more performant if someone wants to invest time and energy into them, but if even the easily obtained version is already usable in many cases, why not?  I'd only invest in making a highly optimized special version if it's really needed; in the meantime it's great that more things just work!

mcrees:
Hi,
I have been experiencing issues with interactive Matplotlib panes.

First -  the interactive options are not displaying properly. All the functionality is working but the icons are not displaying correctly:
image
 
Second - Sizing to fit spaces does not seem to function well with interactive mode. When interactive mode is False, sizing works as expected (scaling or stretching).

Not sure if this is the best place to post.
Thanks for the good work

jbednar:
So that we can keep track of to-do items, please open two separate issues, each listing the relevant software versions and describing that specific problem. Comments on closed or merged PRs are very likely to get overlooked!

PR #1493, opened 2020-07-23
Add busy/loading indicators and enable on Template

philippjfr:
Adds two boolean indicators which can be used independently or in one of the default templates.

## BooleanStatus

Screen Shot 2020-07-23 at 12 20 29 PM

## LoadingSpinner

Screen Shot 2020-07-23 at 12 20 40 PM

## Template

Screen Shot 2020-07-23 at 12 21 46 PM

Here we can see the indicator on the right of the template header.

### DIFF
--- a/panel/_styles/widgets.css
+++ b/panel/_styles/widgets.css
@@ -139,3 +139,55 @@ progress:not([value])::before {
   from {background-position: 0%}
   to {background-position: 100%}
 }
+
+div.loader:after {
+  content: "";
+  background: conic-gradient(
+    #f3f3f3 0 100%
+  );
+  border-radius: 50%;
+  -webkit-mask-image: radial-gradient(transparent 50%, rgba(0, 0, 0, 1) 50%);
+  width: 100%;
+  height: 100%;
+  left: 0;
+  top: 0;
+  position: absolute;
+}
+
+div.loader.spin:after {
+  background: conic-gradient(
+    #3498db 25%,
+    #f3f3f3 0 50%
+  );
+  animation: spin 2s linear infinite;
+}
+
+/* Safari */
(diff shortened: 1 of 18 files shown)


MarcSkovMadsen:
Awesome.

Will it be possible to set easily set a custom color?

In general it would be nice to set a custom color for the different button_type, booleanstatus_type, loadingspinner_type. In an organisation there is normally some kind of branding guide you need to follow.

It would be nice to set that style in one place instead of many places. Setting it in CSS if it is easy would also be a good solution. In some of the material frameworks you can set global settings like primary-color in CSS and then everything just changes. That is nice. With Panel right now it's not easy.

philippjfr:
>Will it be possible to easily set a custom color?

No, this is currently not supported and can't be easily supported until we improve the way CSS is handled and built. I have plans to switch to something like [Sass](https://sass-lang.com/) and compile the CSS properly but for now defining preset colors is the best we can do.

philippjfr:
>In some of the material frameworks you can set :root settings like --primary-color in CSS and then everything just changes.

That is definitely going to be part of that CSS refactoring work.

panel/_styles/widgets.css
```diff
@@ -139,3 +139,203 @@ progress:not([value])::before {
...
+  height: 100%;
+  left: 0;
+  top: 0;
+  position: absolute;
+}
+
+div.loader.dark:after {
+  background: conic-gradient(
+    #0f0f0f 0 100%
+  ) !important;
```
mattpap, line 157: What's the reason for using `!important` here and in the other cases?

philippjfr: I'll have to look again, I **think** I can remove those now.

PR #1513, opened 2020-08-04
Implementation of (Int|Float)[Input|Spinner] (New Bokeh widgets)

xavArtley:
Int, Float separation or not separation?

- [x] Need to update documentation


philippjfr:
I think we should be consistent about it so I'm happy to make sure we introduce distinct Int/Float widgets throughout.

MarcSkovMadsen:
Please do a rename while you are at it to `IntInput` and `FloatInput` as the `Spinner` name is just confusing. I also thinks is something indicating progress. That's what it is in all other frameworks I know of.

xavArtley:
For me spinner is the right name for the i put with arrows https://en.m.wikipedia.org/wiki/Spinner_(computing) . In this PR there is already Int and Float Input widget (if no step is given) and an Int and Float Spinner if step is defined.

philippjfr:
I think `Spinner` on it's own is indeed confusing but `IntSpinner` and `FloatSpinner` actually resolve the ambiguity for me.

philippjfr:
I think we've now agreed that we should do the renaming from `Spinner` -> `NumberInput` in https://github.com/holoviz/panel/issues/1566. So could you start by renaming FloatSpinner and IntSpinner to `FloatInput` and `IntInput` in this PR? (If you're short on time I can do it).

xavArtley:
@philippjfr, I'm a bit short on time, I'll be glad if you could do it.
Thanks

Issue #1522, opened 2020-08-08
Disabled dates in DatePicker widget


ahuang11:
Seems like bokeh has those options
https://docs.bokeh.org/en/latest/_modules/bokeh/models/widgets/inputs.html#DatePicker

PR #1524, opened 2020-08-09
Enabled disabled dates

ahuang11:
https://github.com/holoviz/panel/issues/1522


philippjfr:
Great, thanks so much! Could we also document these parameters in examples/reference/widgets/DatePicker.ipynb so they show up here: https://panel.holoviz.org/reference/widgets/DatePicker.html

PR #1531, opened 2020-08-18
Add Tabulator model and widget

philippjfr:
Adds a new DataGrid/Table widget based on the [Tabulator](http://tabulator.info) library. This uses the implementation that @MarcSkovMadsen added to the awesome-panel-extension library but changes it in a number of fundamental ways:

1) Instead of a configuration dictionary (exposing all the options as a single dictionary) this exposes the various options in a way that is (partially) compatible with DataTable, e.g. columns are defined by providing a list of bokeh `TableColumn` classes, formatters are defined as bokeh `CellFormatter` types and editors are defined as bokeh `CellEditor` types. This reduces flexibility (although that can be readded), but provides the validation of actual models, compatibility with the bokeh DataTable and hopefully making it easier to eventually merge this work upstream into bokeh proper.

2) For now I've reverted back to supporting only DataFrame types as input. I'm happy to generalize this and support other input types but for now this allowed me to share basically the entire panel based implementation between the DataFrame and Tabulator classes.

## ToDo

- [x] Add tests
- [x] Add docs
- [x] Find better approach to selecting theme (currently must call `Tabulator.config()` to load CSS) 

### Features
  - [x] Match basic functionality of the DataFrame widget
  - [x] Expose additional [formatters](http://tabulator.info/docs/4.7/format)
  - [x] Add ability to [group rows](http://tabulator.info/examples/4.7?#grouping)
  - [x] Add ability to [group columns](http://tabulator.info/examples/4.7#column-groups)
  - [x] Expose [pagination](http://tabulator.info/examples/4.7#pagination)
  - [x] Add ability to [download data](http://tabulator.info/docs/4.7/download)
  - [x] Add [pagination support](http://tabulator.info/docs/4.7/page#local)
  - [x] Expose [frozen columns](http://tabulator.info/examples/4.7?#frozen-col)
  - [x] Expose [frozen rows](http://tabulator.info/examples/4.7?#frozen-row)
  - [ ] Support [nested data trees](http://tabulator.info/examples/4.7#tree)
  - [ ] Expose per row height, alignment and tooltips


MarcSkovMadsen:
What about stream and patch. How is that supported?

I created specific functions in order not to have to transfer the whole dataframe to the front end for every small change.

MarcSkovMadsen:
Another things is that the current configuration via a dictionary enables easy copy-paste of examples from the tabulator.info page. A site which has high quality documentation.

For Bokeh on the other hand I'm not able to find anything on the bokeh DataGrid except the reference api where you really need to concentrate to understand things. I guess that is the reason I don't understand the existing Bokeh datagrid because it is difficult to get started with and experiment with when there is no documentation.

philippjfr:
>What about stream and patch. How is that supported?

I added those to the new base class so they are supported both for the DataFrame and Tabulator widgets now.

>Another things is that the current configuration via a dictionary enables easy copy-paste of examples from the tabulator.info page. A site which has high quality documentation.

I'd be happy to add a `from_configuration` classmethod to construct directly from the Tabulator examples.

MarcSkovMadsen:
I think the from_configuration method or an _initial_configuration argument to the constructor would be very nice to have.

One thing I've also tried to find a solution for is how to support using javascript functions in the constructor. There are a lot of awesome examples where they sprinkle in a few small javascript functions to render, layout, style etc. Is it possible to transfer javascript functions from python in a configuration dictionary as well?

examples/reference/widgets/DataFrame.ipynb
```diff
@@ -16,7 +16,7 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "The ``DataFrame`` widget allows displaying and editing a pandas DataFrame. Note that for multi-indexed DataFrames editing is not possible and you will have to reduce the DataFrame to a single index.\n",
+    "The ``DataFrame`` widget allows displaying and editing a pandas DataFrame. Note that for multi-indexed DataFrames editing is not possible and you will have to reduce the DataFrame to a single index. Note that the `DataFrame` widget will eventually be replaced with the [`Tabulator`](./Tabulator.ipynb) widget.\n",
```
jbednar, line 19:
```suggestion
    "The ``DataFrame`` widget allows displaying and editing a pandas DataFrame. Note that editing is not possible for multi-indexed DataFrames, in which casey ou will need to reduce the DataFrame to a single index. Also note that the `DataFrame` widget will eventually be replaced with the [`Tabulator`](./Tabulator.ipynb) widget, and so new code should be written to use Tabulator instead.\n",
```

examples/reference/widgets/DataFrame.ipynb
```diff
@@ -258,6 +259,109 @@
     "pn.widgets.DataFrame(value=pop_df, hierarchical=True, aggregators={'Sex': 'sum', 'AgeGrp': 'sum'}, height=400)"
    ]
   },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## Streaming\n",
+    "\n",
+    "When we are monitoring some source of data which updates we may want to update the table with the newly arriving data, however we do not want to transmit the entire dataset each time. To handle efficient transfer of just the latest data we can use the `.stream` method on the `DataFrame` object:"
```
jbednar, line 268:
```suggestion
    "When we are monitoring a source of data that updates over time, we may want to update the table with the newly arriving data. However, we do not want to re-transmit the entire dataset each time. To handle efficient transfer of just the latest data, we can use the `.stream` method on the `DataFrame` object:"
```

panel/widgets/tables.py
```diff
@@ -197,6 +161,493 @@ def _get_columns(self):
...
+                    scalars and the filter will check if the values
+                    in the column match any of the items in the list.
+
+        Arguments
+        ---------
+        filter: Widget, param.Parameter or FunctionType
+            The filter which will provide the value to filter the
+            DataFrame along the declared column or a function which
+            will be passed the DataFrame and should return a filtered
+            copy of the DataFrame.
```
jbednar, line 261:
```suggestion
            The filter that will provide the value to filter the
            DataFrame along the declared column, or a function that
            will be passed the DataFrame and should return a filtered
            copy of the DataFrame.
```

Even slightly reworded I can't quite parse this sentence...

panel/widgets/tables.py
```diff
@@ -197,6 +161,493 @@ def _get_columns(self):
...
+        column: str or None
+            If the filter is a widget or parameter the column declares
+            the value will filter on.
```
jbednar, line 264:
```suggestion
            Column to which the filter will be applied, if the filter is a widget or parameter.
```

(Is that what it meant?)

jbednar, line 264:
```suggestion
            Column to which the filter will be applied, if the filter is a widget or parameter.
```

(Is that what it meant?)

panel/widgets/tables.py
```diff
@@ -197,6 +161,493 @@ def _get_columns(self):
...
+        if isinstance(filter, FunctionType):
+            deps = list(filter._dinfo['kw'].values()) if hasattr(filter, '_dinfo') else []
+        else:
+            filter = param_value_if_widget(filter)
+            if not isinstance(filter, param.Parameter):
+                raise ValueError(f'{type(self).__name__} filter must be '
+                                 'a parameter, widget or function.')
+            elif column is None:
+                raise ValueError('When filtering with a parameter or '
+                                 'widget a column to filter on must be '
```
jbednar, line 280:
```suggestion
                                 'widget, a column to filter on must be '
```

panel/widgets/tables.py
```diff
@@ -197,6 +161,493 @@ def _get_columns(self):
...
+        Streams (appends) the `stream_value` provided to the existing
+        value in an efficient manner.
+
+        Arguments
+        ---------
+        stream_value (Union[pd.DataFrame, pd.Series, Dict])
+          The new value(s) to append to the existing value.
+        reset_index (bool, default=True):
+          If the stream_value is a DataFrame and `reset_index` is True
+          then the index of it is reset if True. Helps to keep the
```
jbednar, line 394:
```suggestion
          If True and the stream_value is a DataFrame, 
          then its index is reset. Helps to keep the
```

panel/widgets/tables.py
```diff
@@ -197,6 +161,493 @@ def _get_columns(self):
...
+            self._stream(stream_value)
+        elif isinstance(stream_value, dict):
+            if stream_value:
+                try:
+                    stream_value = pd.DataFrame(stream_value)
+                except ValueError:
+                    stream_value = pd.Series(stream_value)
+                self.stream(stream_value)
+        else:
+            raise ValueError("The stream value provided is not a DataFrame, Series or Dict!")
```
jbednar, line 459: Not sure if the ! is really needed. :-)

panel/widgets/tables.py
```diff
@@ -197,6 +161,493 @@ def _get_columns(self):
...
+
+    def patch(self, patch_value):
+        """
+        Patches (updates) the existing value with the `patch_value` in
+        an efficient manner.
```
jbednar, line 464:
```suggestion
        Efficiently patches (updates) the existing value with the `patch_value`.
```

panel/widgets/tables.py
```diff
@@ -197,6 +161,493 @@ def _get_columns(self):
...
+
+        Patch a Dataframe with a Dictionary of Columns.
+        >>> value = pd.DataFrame({"x": [1, 2], "y": ["a", "b"]})
+        >>> tabulator = Tabulator(value=value)
+        >>> patch_value = {"x": [(slice(2), (3,4))], "y": [(1,'d')]}
+        >>> tabulator.patch(patch_value)
+        >>> tabulator.value.to_dict("list")
+        {'x': [3, 4], 'y': ['a', 'd']}
+
+        Patch a DataFrame with a Series. Please note the index is used in the update
```
jbednar, line 494:
```suggestion
        Patch a DataFrame with a Series. Please note the index is used in the update.
```

panel/widgets/tables.py
```diff
@@ -197,6 +161,493 @@ def _get_columns(self):
...
+        aggregators for different columns are required the dictionary
+        may be nested as `{index_name: {column_name: aggregator}}`""")
+
+    auto_edit = param.Boolean(default=False, doc="""
+        Whether clicking on a table cell automatically starts edit mode.""")
+
+    autosize_mode = param.ObjectSelector(default='force_fit', objects=[
+        "none", "fit_columns", "fit_viewport", "force_fit"], doc="""
+
+        Describes the column autosizing mode with one of the following options:
```
jbednar, line 576:
```suggestion
        Determines the column autosizing mode, as one of the following options:
```

panel/widgets/tables.py
```diff
@@ -197,6 +161,493 @@ def _get_columns(self):
...
+
+        ``"fit_columns"``
+          Compute columns widths based on cell contents but ensure the
```
jbednar, line 579:
```suggestion
          Compute column widths based on cell contents while ensuring the
```

panel/widgets/tables.py
```diff
@@ -197,6 +161,493 @@ def _get_columns(self):
...
+          table fits into the available viewport. This results in no
+          horizontal scrollbar showing up, but data can get unreadable
+          if there is not enough space available.
+
+        ``"fit_viewport"``
+          Adjust the viewport size after computing columns widths based
```
jbednar, line 585:
```suggestion
          Adjust the viewport size after computing column widths based
```

panel/widgets/tables.py
```diff
@@ -197,6 +161,493 @@ def _get_columns(self):
...
+        Whether to generate a hierachical index.""")
+
+    fit_columns = param.Boolean(default=None, doc="""
+        Whether columns should expand to the available width. This
+        results in no horizontal scrollbar showing up, but data can
+        get unreadable if there is no enough space available.""")
+
+    frozen_columns = param.Integer(default=None, doc="""
+        Integer indicating the number of columns to freeze. If set the
+        first N columns will be frozen which prevents them from
```
jbednar, line 607:
```suggestion
        Integer indicating the number of columns to freeze. If set, the
        first N columns will be frozen, which prevents them from
```

panel/widgets/tables.py
```diff
@@ -197,6 +161,493 @@ def _get_columns(self):
...
+        scrolling out of frame.""")
+
+    frozen_rows = param.Integer(default=None, doc="""
+       Integer indicating the number of rows to freeze. If set the
+       first N rows will be frozen which prevents them from scrolling
+       out of frame, if set to a negative value last N rows will be
```
jbednar, line 613:
```suggestion
       Integer indicating the number of rows to freeze. If set, the
       first N rows will be frozen, which prevents them from scrolling
       out of frame; if set to a negative value the last N rows will be
```

panel/widgets/tables.py
```diff
@@ -197,6 +161,493 @@ def _get_columns(self):
...
+
+    reorderable = param.Boolean(default=True, doc="""
+        Allows the reordering of a table's columns. To reorder a
+        column, click and drag a table's header to the desired
+        location in the table.  The columns on either side will remain
+        in their previous order.""")
+
+    sortable = param.Boolean(default=True, doc="""
+        Allows to sort table's contents. By default natural order is
+        preserved.  To sort a column, click on it's header. Clicking
```
jbednar, line 624:
```suggestion
        preserved.  To sort a column, click on its header. Clicking
```

panel/widgets/tables.py
```diff
@@ -260,77 +706,336 @@ def _process_param_change(self, msg):
...
+        for g in model.grouping:
+            group = self._renamed_cols.get(g.getter, g.getter)
+            index = self.indexes[self.indexes.index(group)+1]
+            g.aggregators = self._get_aggregators(index)
+
+
+class Tabulator(BaseTable):
+    """
+    The Tabulator Pane wraps the [Tabulator](http://tabulator.info/)
+    table to provide an awesome interactive table.
```
jbednar, line 719:
```suggestion
    table to provide a full-featured interactive table.
```

panel/widgets/tables.py
```diff
@@ -260,77 +706,336 @@ def _process_param_change(self, msg):
...
+
+    frozen_columns = param.List(default=[], doc="""
+        List indicating the columns to freeze. The column(s) may be
+        selected by name or index.""")
+
+    frozen_rows = param.List(default=[], doc="""
+        List indicating the rows to freeze. If set the
+        first N rows will be frozen which prevents them from scrolling
+        out of frame, if set to a negative value last N rows will be
+        frozen.""")
```
jbednar, line 730:
```suggestion
        List indicating the rows to freeze. If set, the
        first N rows will be frozen, which prevents them from scrolling
        out of frame; if set to a negative value the last N rows will be
        frozen.""")
```

Shouldn't these parameters that are shared with the DataFrame class be grouped into a base class so that the docstrings don't drift out of sync?

panel/widgets/tables.py
```diff
@@ -260,77 +706,336 @@ def _process_param_change(self, msg):
...
+        if self.groups and 'columns' in configuration:
+            raise ValueError("Groups must be defined either explicitly "
+                             "or via the configuration, not both.")
+        configuration['columns'] = self._config_columns(columns)
+        return configuration
+
+    @staticmethod
+    def config(css="default"):
+        """
+        Adds the specified css theme to pn.config.css_files
```
jbednar, line 1031:
```suggestion
        Adds the specified CSS theme to pn.config.css_files
```

examples/reference/widgets/Tabulator.ipynb
```diff
@@ -0,0 +1,724 @@
...
+    "\n",
+    "For layout and styling related parameters see the [customization user guide](../../user_guide/Customization.ipynb).\n",
+    "\n",
+    "##### Core\n",
+    "\n",
+    "* **``editors``** (``dict``):  A dictionary mapping from column name to a bokeh `CellEditor` instance or tabulator editor specification.\n",
+    "* **``formatters``** (``dict``): A dictionary mapping from column name to a bokeh `CellFormatter` instance or tabulator formatter specification.\n",
+    "* **``layout``** (str): Describes the column layout mode with one of the following options `'fit_columns'`, `'fit_data'`, `'fit_data_stretch'`, `'fit_data_fill'`, `'fit_data_table'`. \n",
+    "* **``frozen_columns``** (`list`): List of columns to freeze, preventing them from scrolling out of frame. Column can be specified by name or index.\n",
+    "* **``frozen_rows``**: (`list`): List of rows to freeze preventing them from scrolling out of frame. Rows can be specified by positive or negative index.\n",
```
jbednar, line 35:
```suggestion
    "* **``frozen_rows``**: (`list`): List of rows to freeze, preventing them from scrolling out of frame. Rows can be specified by positive or negative index.\n",
```

examples/reference/widgets/Tabulator.ipynb
```diff
@@ -0,0 +1,724 @@
...
+    "* **``page``** (``int``): Current page if pagination is enabled.\n",
+    "* **``page_size``** (``int``): Number of rows on each page.\n",
+    "* **``pagination``** (`str`, `default=None`):  Set to 'local' or 'remote' to enable pagination, by default pagination is disabled with the value set to `None`.\n",
```
jbednar, line 38:
```suggestion
    "* **``pagination``** (`str`, `default=None`):  Set to 'local' or 'remote' to enable pagination; by default pagination is disabled with the value set to `None`.\n",
```

examples/reference/widgets/Tabulator.ipynb
```diff
@@ -0,0 +1,724 @@
...
+    "* **``name``** (str): The title of the widget\n",
+    "\n",
+    "___"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "The ``Tabulator`` widget renders a DataFrame using an interactive grid, which allows directly editing the contents of the dataframe inplace, with any changes being synced with Python. The `Tabulator` will usually determine the appropriate formatter appropriately based on the type of the data:"
```
jbednar, line 58:
```suggestion
    "The ``Tabulator`` widget renders a DataFrame using an interactive grid, which allows directly editing the contents of the dataframe in place, with any changes being synced with Python. The `Tabulator` will usually determine the appropriate formatter appropriately based on the type of the data:"
```

examples/reference/widgets/Tabulator.ipynb
```diff
@@ -0,0 +1,724 @@
...
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "The list of valid Tabulator formatters can be found in the [Tabulator documentation](http://tabulator.info/docs/4.9/format#format-builtin).\n",
+    "\n",
+    "## Editors\n",
+    "\n",
+    "Just like the formatters the `Tabulator` will natively understand the Bokeh `Editor` types, however in the background it will replace most of them with equivalent editors natively supported by the tabulator library:"
```
jbednar, line 143:
```suggestion
    "Just like the formatters, the `Tabulator` will natively understand the Bokeh `Editor` types. However, in the background it will replace most of them with equivalent editors natively supported by the tabulator library:"
```

examples/reference/widgets/Tabulator.ipynb
```diff
@@ -0,0 +1,724 @@
...
+    "pn.widgets.Tabulator(df[['float', 'bool', 'str']], editors=bokeh_editors)"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "### Column layouts\n",
+    "\n",
+    "By default the DataFrame widget will adjust the sizes of both the columns and the table based on the contents, reflecting the default value of the parameter: `layout=\"fit_data_table\"`. Alternatively modes allow manually specifying the widths of the columns, giving each column equal widths or adjusting just the size of the columns.\n",
```
jbednar, line 194:
```suggestion
    "By default the DataFrame widget will adjust the sizes of both the columns and the table based on the contents, reflecting the default value of the parameter: `layout=\"fit_data_table\"`. Alternative modes allow manually specifying the widths of the columns, giving each column equal widths, or adjusting just the size of the columns.\n",
```

examples/reference/widgets/Tabulator.ipynb
```diff
@@ -0,0 +1,724 @@
...
+   "outputs": [],
+   "source": [
+    "pn.widgets.tables.Tabulator(custom_df, layout='fit_data', width=400)"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "To ensure that the table fits all the data but also stretches to fill all the available space set `layout='fit_data_stretch'`:"
```
jbednar, line 250:
```suggestion
    "To ensure that the table fits all the data but also stretches to fill all the available space, set `layout='fit_data_stretch'`:"
```

examples/reference/widgets/Tabulator.ipynb
```diff
@@ -0,0 +1,724 @@
...
+    "\n",
+    "select_table = pn.widgets.tables.Tabulator(sel_df, selection=[0, 3, 7])\n",
+    "select_table"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "Once initialized the ``selection`` parameter will return the integer indexes of the selected rows and the ``selected_dataframe`` property will return a new DataFrame containing just the selected rows:"
```
jbednar, line 392:
```suggestion
    "Once initialized, the ``selection`` parameter will return the integer indexes of the selected rows, while the ``selected_dataframe`` property will return a new DataFrame containing just the selected rows:"
```

examples/reference/widgets/Tabulator.ipynb
```diff
@@ -0,0 +1,724 @@
...
+    "select_table.selection = [1, 4, 9]\n",
+    "\n",
+    "select_table.selected_dataframe"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "The `selectable` parameter can also be used to control how the selection works, setting it to `False` will disable selecting entirely, while setting `selectable='checkbox'` will add checkboxes to perform the selection:"
```
jbednar, line 410:
```suggestion
    "The `selectable` parameter can also be used to control how the selection works. Setting it to `False` will disable selecting entirely, while setting `selectable='checkbox'` will add checkboxes to perform the selection:"
```

examples/reference/widgets/Tabulator.ipynb
```diff
@@ -0,0 +1,724 @@
...
+    "pn.widgets.tables.Tabulator(sel_df, selection=[0, 3, 7], selectable='checkbox')"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "### Freezing rows and columns\n",
+    "\n",
+    "Often times your table will be larger than can be displayed in a single viewport and scroll bars will be enabled. The issue with this is that you might want to make sure that certain information is always visible. This is where the `frozen_columns` and `frozen_rows` options come in.\n",
```
jbednar, line 428:
```suggestion
    "Sometimes your table will be larger than can be displayed in a single viewport, in which case scroll bars will be enabled. In such cases, you might want to make sure that certain information is always visible. This is where the `frozen_columns` and `frozen_rows` options come in.\n",
```

examples/reference/widgets/Tabulator.ipynb
```diff
@@ -0,0 +1,724 @@
...
+    "\n",
+    "#### Frozen columns\n",
+    "\n",
+    "When you have a large number of columns and can't fit them all on the screen you might still want to make sure that certain columns do not scroll out of view. The `frozen_columns` option makes this possible by specifying a list of columns that should be frozen, e.g. `frozen_columns=['index']` will freeze the index column:"
```
jbednar, line 432:
```suggestion
    "When you have a large number of columns and can't fit them all on the screen you might still want to make sure that certain columns do not scroll out of view. The `frozen_columns` option makes this possible by specifying a list of columns that should be frozen. E.g. `frozen_columns=['index']` will freeze the index column:"
```

examples/reference/widgets/Tabulator.ipynb
```diff
@@ -0,0 +1,724 @@
...
+    "pn.widgets.Tabulator(wide_df, frozen_columns=['index'], width=400)"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "#### Frozen rows\n",
+    "\n",
+    "Another common scenario is when you have certain rows with special meaning, e.g. aggregates which summarize the information in the rest of the table. In this case you may want to freeze those rows so they do not scroll out of view. You can achieve this by setting a list of `frozen_rows` by integer index (which can be positive or negative):"
```
jbednar, line 452:
```suggestion
    "Another common scenario is when you have certain rows with special meaning, e.g. aggregates that summarize the information in the rest of the table. In this case you may want to freeze those rows so they do not scroll out of view. You can achieve this by setting a list of `frozen_rows` by integer index (which can be positive or negative, where negative values are relative to the end of the table):"
```

examples/reference/widgets/Tabulator.ipynb
```diff
@@ -0,0 +1,724 @@
...
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## Filtering\n",
+    "\n",
+    "A very common scenario is that you want to attach a number of filters to a table in order to view just a subset of the data. You can achieve this through callbacks or other reactive approaches but the  `.add_filter` method makes it much easier.\n",
+    "\n",
+    "#### Widget filters\n",
+    "\n",
+    "The simplest approach to filtering is simply to declare one or more widgets and then using the `.add_filter` to add them as filters on specific columns. The filtering will depend on the type of the widget or parameters value:\n",
```
jbednar, line 517:
```suggestion
    "Perhaps the simplest approach to filtering is simply to declare one or more widgets and then using the `.add_filter` to add them as filters on specific columns. The filtering will depend on the type of the widget or parameter value:\n",
```

(Personally, I'd think the _simplest_ approach to filtering is to supply a static value, e.g. `make='Ford'`; that seems simpler than a widget!)

examples/reference/widgets/Tabulator.ipynb
```diff
@@ -0,0 +1,724 @@
...
+    "    pn.Column(slider, select),\n",
+    "    filter_table\n",
+    ")"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "For more complex filtering tasks you can supply a function, which should accept the DataFrame to be filtered as the first argued and must return a filtered copy of the data. By using the `pn.bind` function, which binds widget and parameter values to a function, complex filtering can be achieved, e.g. here we will add a filter function which uses `fnmatch` (similar to a regex) to filter on the 'Director' column of a listing of thousands of movies:"
```
jbednar, line 550:
```suggestion
    "For more complex filtering tasks you can supply a function, which should accept the DataFrame to be filtered as the first argument and must return a filtered copy of the data. By using the `pn.bind` function, which binds widget and parameter values to a function, complex filtering can be achieved. E.g. here we will add a filter function that uses `fnmatch` (similar to a regex) to filter on the 'Director' column of a listing of thousands of movies:"
```

examples/reference/widgets/Tabulator.ipynb
```diff
@@ -0,0 +1,724 @@
...
+    "pn.Row(director_filter, movies_table)"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## Streaming\n",
+    "\n",
+    "When we are monitoring some source of data which updates we may want to update the table with the newly arriving data, however we do not want to transmit the entire dataset each time. To handle efficient transfer of just the latest data we can use the `.stream` method on the `Tabulator` object:"
```
jbednar, line 589:
```suggestion
    "When we are monitoring some source of data that updates over time, we may want to update the table with the newly arriving data. However, we do not want to transmit the entire dataset each time. To handle efficient transfer of just the latest data, we can use the `.stream` method on the `Tabulator` object:"
```

examples/reference/widgets/Tabulator.ipynb
```diff
@@ -0,0 +1,724 @@
...
+    "\n",
+    "stream_table = pn.widgets.Tabulator(stream_df, layout='fit_columns', width=450)\n",
+    "stream_table"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "As example we will schedule a periodic callback which will stream new data every 1000ms (i.e. 1s) five times in a row:"
```
jbednar, line 608:
```suggestion
    "As example, we will schedule a periodic callback that streams new data every 1000ms (i.e. 1s) five times in a row:"
```

examples/reference/widgets/Tabulator.ipynb
```diff
@@ -0,0 +1,724 @@
...
+    "    stream_table.stream(stream_df, follow=follow)\n",
+    "\n",
+    "pn.state.add_periodic_callback(stream_data, period=1000, count=5)"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "If you are viewing this example in a live kernel you will be able to watch the table update and scroll along. If we want to disable the scrolling behavior we can set `follow=False`:"
```
jbednar, line 628:
```suggestion
    "If you are viewing this example with a live Python kernel you will be able to watch the table update and scroll along. If we want to disable the scrolling behavior, we can set `follow=False`:"
```

panel/widgets/tables.py
philippjfr: They are quite different unfortunately, one accepts a single int this one accepts a list.

panel/widgets/tables.py
philippjfr: Does this make sense?

            The value to filter DataFrame with along the declared
            column, or a function that should accept the DataFrame to
            be filtered as the first argument and should return a
            filtered copy of the DataFrame.

panel/widgets/tables.py
philippjfr: Yes.

examples/reference/widgets/Tabulator.ipynb
philippjfr: New text:

> The simplest approach to filtering is to select along a column with a constant or dynamic value. The `.add_filter` method allows passing in constant values, widgets and parameters. If a widget or parameter is provided the table will watch the object for changes in the value and update the data in response.

panel/widgets/tables.py
jbednar: The value by which to filter the DataFrame along the declared
column, or a function accepting the DataFrame to
be filtered and returning a filtered copy of the DataFrame.

examples/reference/widgets/Tabulator.ipynb
jbednar: Ok

PR #1535, opened 2020-08-20
Enhanced responsive grid template

philippjfr:
Simplified creation of responsive grid in ReactTemplate. This PR uses the GridSpec component to make it easy to populate the initial grid of the template and provides parameters to control the various breakpoints at which the grid reflows.

```python
responsive = pn.template.ReactTemplate(title='React Grid Template', row_height=300)

xs = np.linspace(0, np.pi)

def sine(freq, phase):
        responsive=True)

def cosine(freq, phase):
        responsive=True)

def tan(freq, phase):
    return hv.Curve((xs, np.tan(xs*freq+phase))).opts(
        responsive=True)

responsive.sidebar.append(freq)
responsive.sidebar.append(phase)

responsive.main[0:2, 0:4] = pn.Card(hv.DynamicMap(sine), title='Sine')
responsive.main[0:2, 4:8] = pn.Card(hv.DynamicMap(cosine), title='Cosine')
responsive.main[0:2, 8:12] = pn.Card(hv.DynamicMap(tan), title='Tan')
```

In other words there is a major difference in the way the main area of this template is populated compared to the other templates. The example above creates a grid of 12 columns and two rows which it populates with three cards.

ezgif-1-ca4734493c1b


nghenzi:
Great, it is much easier to use it !!!

MarcSkovMadsen:
Would it be possible to take the simplification one step further? What I would like to see is actually not a ReactTable but a

Vanille, Material, Bootstrap, Golden layout etc template with the react table grid layout or another grid layout like the material md layout or one of the other grid layouts I listed yesterday.

What I would like to be able to do is something like

```python
pn.template.XYZTemplate(title='Example Template', main_layout=ReactGridLayout)
```

```python
pn.template.XYZTemplate(title='Example Template', main_layout=GoldenLayout)
```

```python
pn.template.XYZTemplate(title='Example Template', main_layout=MDLayout)
```


and if `main_layout` is not specified the original layout (Column?) is just used. The we could add

MarcSkovMadsen:
One additional question. If my layout is to consist of two rows. How could I get sure they are maximized to the window height?

Most time users don't want the app in fixed height but responsive and fitting the window height.

**For example in the example above how would you make those plots adjust to the full height of the window?**

philippjfr:
>Most time users don't want the app in fixed height but responsive and fitting the window height.

I'd argue **most** is a considerable overstatement, but I get your point. Currently the grid implementation here is not responsive in height. You define a fixed height per row and then assign an arbitrary number of columns. I don't really see how you can have a grid that is responsive in both width and height unless you want to express it as "divide the available height between N rows". Will have to think about this but I don't recall ever coming across a grid implementation like this.

jbednar:
>  I don't really see how you can have a grid that is responsive in both width and height unless you want to express it as "divide the available height between N rows". 

Good question.  If I think about what I might want out of a responsive grid that's being used as a collection of cards, I think I might have a notion of an expected length and width per card (in pixels, I guess), and then I'd want the responsive grid to add as many columns are necessary to get close to that expected width, and as many rows as are necessary to get close to that expected length, and would then stretch the precise length and width to divide the space evenly amongst that number of rows and columns.  But it's easy to imagine people who would want responsiveness to work differently.

MarcSkovMadsen:
After second thought I think the responsive height would be nice to have. But I think it is not the normal behaviour. So "forget it" :-)

nghenzi:
I continue playing with this template. A problem that may appear is the overflow of the elements, for example you have a table (for example datatable) with two plots. In that case, you want that the card with the table has overflowY:'auto' property but that the plots don't. 

image


image

MarcSkovMadsen:
My first thought is that as a developer i would assumption no Overflow as default because the grid is meant to be reorganiser and resized. So the content would be responsive. 

Then you as a developer should have a possibilty fix the size and enable scrollbars? 

Just a thought.

philippjfr:
>My first thought is that as a developer i would assume no Overflow as default because the grid is meant to be reorganised and resized. So the content should be responsive.

Not all content can be made responsive effectively, especially tables rarely work well in a responsive scenario because the column get squashed. So I think we should add overflow and I don't anticipate any issues since responsively sized objects won't overflow, while contents that do overflow will.

nghenzi:
I could get a first PoC of the responsive datatable without overflow (there is some problems yet), but I had to change the resize funcion inside the template 


```
function resize_layout(obj, old_, new_, p, e, element) {
  // console.log('resize', obj, old_, new_, p, e, element);
  const card_height = new_['h']* {{rowHeight}} ;
  const tabla_jquery = document.getElementById('example_wrapper').clientHeight
  const body_tabla =  $('div.dataTables_scrollBody').height(); 

  if (card_height < tabla_jquery+15){
        $('div.dataTables_scrollBody').height( card_height - 10 - (tabla_jquery-body_tabla));
    } else 
    {$('div.dataTables_scrollBody').height(card_height);}
  window.dispatchEvent(new Event("resize"));
}

function resize_layout_stop(obj, old_, new_, p, e, element) {
  setTimeout(function(){
    
  const card_height = new_['h']* {{rowHeight}};
  const tabla_jquery = document.getElementById('example_wrapper').clientHeight
  const body_tabla =  $('div.dataTables_scrollBody').height(); 
  console.log('wwww', card_height, tabla_jquery, body_tabla)
  if (card_height < tabla_jquery+15){
        $('div.dataTables_scrollBody').height( card_height - 10 - (tabla_jquery-body_tabla));
    } else 
    {$('div.dataTables_scrollBody').height(card_height);}
  window.dispatchEvent(new Event("resize"));
  }, 100);
}

```

with the responsive add-on of the datatable framework

```

raw_css = """ div.dataTables_wrapper {
                    width: 100%;
                    height: 100%;
                    margin: 0 auto;
                } 
                div.dataTables_scrollBody {
  overflow-y: scroll !important;
   overflow-x:hidden !important;
}

     """ 


css = ['https://cdn.datatables.net/1.10.22/css/jquery.dataTables.min.css',
        'https://cdn.datatables.net/fixedheader/3.1.7/css/fixedHeader.dataTables.min.css',
      'https://cdn.datatables.net/responsive/2.2.6/css/responsive.dataTables.min.css']

js = {
    '$': 'https://code.jquery.com/jquery-3.5.1.js',
    'DataTable': 'https://cdn.datatables.net/1.10.22/js/jquery.dataTables.min.js',
    'respo1': 'https://cdn.datatables.net/fixedheader/3.1.7/js/dataTables.fixedHeader.min.js',
    'respo2': 'https://cdn.datatables.net/responsive/2.2.6/js/dataTables.responsive.min.js'
}

pn.extension(css_files=css, js_files=js, raw_css=[raw_css])

pn.config.sizing_mode = 'stretch_width'


script = """
<script>
if (document.readyState === "complete") {
var table = $('#example').DataTable( {
            "responsive": true,
            "scrollY":        "700px",
            "scrollCollapse": true,
            "paging":         true,
            "columnDefs": [{"className": "dt-left", "targets": "_all"}]
        });

fix_header = new $.fn.dataTable.FixedHeader( table );

} """

html = autompg.to_html().replace("""class="dataframe\"""",
            """ id="example" class="display nowrap" style="width:100%" """)

d = pn.pane.HTML(html+script, sizing_mode='stretch_width')

```
here it a gif showing it. 


responsive_table

If the idea of the template is to be easy to use, So I think to avoid all this tweaking, the overflow property is the best option. 


On the other side, I am working in the maximize button following the approach of this stackoverflow question

https://stackoverflow.com/questions/35689302/programmatically-change-height-width-of-items-in-react-grid-layout

I have something working, but not yet compatible with the last version of the grid template.

philippjfr:
@nghenzi Please open a new issue with your suggestions for overflow, I'm going to merge this shortly.

MarcSkovMadsen:
@nghenzi . If you like promoting your work and have the time I think it would be really awesome if you created some examples and  tweeted a bit about the examples.

The examples could go into the Panel Gallery. I would also like to host live versions at awesome-panel.org.

Just a thought.

nghenzi:
Hi @MarcSkovMadsen ! 

thanks for your comments !!!  I really aprecciate all what you contribute !!! 

> @nghenzi . If you like promoting your work and have the time I think it would be really awesome if you created some examples and tweeted a bit about the examples.

I dont like much to promote my work, and the social networks are really stressing and time consuming to fully participate in them. I like to contribute to this project due to all the learning adquired in the last year due to all of you. 
 
> The examples could go into the Panel Gallery. I would also like to host live versions at awesome-panel.org.

I am going to try to make some example with the width responsive version of the datatable to add.  Feel free to use what you like in the awesome-panel-org site. I learn a lot of you and the site :) !!!

> This template is really, really awesome and powerful. If it gets some users, feedback and improvements its a really strong thing.

I think too the template is really powerfull, and it can be really improved with new functionalities to have a fully functional easy configurable analytic app from python only.  

> Just a thought.

MarcSkovMadsen:
You are an inspiration too @nghenzi 👍 

I promoted the upcoming react template a bit on your and Panels behalf https://twitter.com/MarcSkovMadsen/status/1319680204785242114?s=20

PR #1548, opened 2020-08-25
Case-insensitive autocompletion

dwsideriusNIST:
Updated select.py to support case-insensitive AutocompletionInput, which was added to bokeh in v 2.1.0 (https://github.com/bokeh/bokeh/issues/9922).


philippjfr:
Thanks, test failure is unrelated.

PR #1570, opened 2020-09-08
Update comparisons

jbednar:
Now that ipywidgets_bokeh is available, some of the information in doc/Comparisons.rst was out of date, so that's been updated. Also clarified some other sections of that doc.


philippjfr (approved):
Thanks, the changes all look good to me and are even accurate 😄

PR #1577, opened 2020-09-13
Wait until JS dependency is loaded before rendering

philippjfr:


MarcSkovMadsen:
Is there any way this can be made more general @philippjfr ? 

I'm thinking about for example the awesome-panel-extensions and other users who have their own custom extensions or uses custom js libraries.

Issue #1578, opened 2020-09-14
When updating Param object the widget changes


MarcSkovMadsen:
I am working on an app where the `object` of a `pn.Param` is dynamically changed. But unfortunately it does not keep the original widget.

![param_issue](https://user-images.githubusercontent.com/42288570/93042363-1e55a500-f64f-11ea-933c-71e4bc4bd6e4.gif)

```python
import param
import panel as pn
from awesome_panel_extensions.frameworks import fast

class TextModel(param.Parameterized):
    text = param.String()

class TextView(param.Parameterized):
    text = param.ClassSelector(class_=TextModel)
    new = param.Action(label="new TextModel")
    text_pane = param.Parameter()
    view = param.Parameter()

    def __init__(self, **params):
        text=params.get("text", TextModel(text="Original Text"))
        params["text_pane"]=pn.Param(text, parameters=["text"], widgets={"text": {"type": fast.FastTextAreaInput, "width": 400}})
        params["new"]=self._new
        super().__init__(**params)

        self.view = pn.Column(
            self.param.new, self.text_pane
        )

    def _new(self, *_):
        self.text = TextModel(text="New Text")

    @param.depends("text", watch=True)
    def _update_text_pane(self, *_):
        self.text_pane.object = self.text

fast.FastTemplate(TextView().view).show(port=5007)
```

MarcSkovMadsen:
I have identified the problem and the solution

![image](https://user-images.githubusercontent.com/42288570/93042450-565ce800-f64f-11ea-86e1-b451762c29f1.png)

![param_no_issue](https://user-images.githubusercontent.com/42288570/93042529-92904880-f64f-11ea-95ef-cafc5fc9de59.gif)

MarcSkovMadsen:
And here is a test that fails

```python
import param
from panel import Param
from panel.widgets import TextAreaInput

def test_change_object_and_keep_widgets():
    """Test that https://github.com/holoviz/panel/issues/1578 is solved"""
    # Given
    class TextModel(param.Parameterized):
        text = param.String()

    class TextView(param.Parameterized):
        text = param.ClassSelector(class_=TextModel)
        new = param.Action(label="new TextModel")
        text_pane = param.Parameter()

        def __init__(self, **params):
            params["text"]=TextModel(text="Original Text")
            super().__init__(**params)

            self.text_pane = Param(self.text, parameters=["text"], widgets={"text": {"type": TextAreaInput}})
            self.new = self._new



[...]

PR #1579, opened 2020-09-14
fix issue 1578

MarcSkovMadsen:
This PR should fix https://github.com/holoviz/panel/issues/1578

The issue is that when changing the object of Param any widgets specified previously are not used because they have been *pop'ed*.

### DIFF
--- a/panel/param.py
+++ b/panel/param.py
@@ -324,7 +324,7 @@ class Param(PaneBase):
             widget_class = self.widget_type(p_obj)
         elif isinstance(self.widgets[p_name], dict):
             if 'type' in self.widgets[p_name]:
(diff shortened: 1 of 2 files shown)


MarcSkovMadsen:
I can see it looks like I have changed everything in the files. But I have not. I have only changed one line in Param and inserted a test in the end of the test_param file.

It must be some windows vs linux issue.

MarcSkovMadsen:
There is one test that fails. It's simply because the fix I've made reveals an error in another test.

I did not fix it because I believe you (@philippjfr ) should take a look and confirm that the test is problematic.

image

MarcSkovMadsen:
Closed in favor of https://github.com/holoviz/panel/pull/1583 which has correct line endings.

Issue #1574, opened 2020-09-11
type of LiteralInput does not work with pn.Param


MarcSkovMadsen:
I would like to use a LiteralInput as the widget in `pn.Param` of a Parameterized class but it is not possible.

As `type` provides the **type of widget** in the below `pn.Param` example the `type` of a LIteralInput cannot be provided.

```python
import param
import panel as pn
pn.extension()

WIDGETS = {
    "a_literal": {
        "type": pn.widgets.LiteralInput, "sizing_mode": "fixed", "width": 400, "placeholder": "Input a dictionary!"
    }
}

class ParameterizedApp(param.Parameterized):
    a_literal = param.Dict(default=None, label="A Dictionary")
    view = param.Parameter()
    
    
    def __init__(self, **params):
        super().__init__(**params)
        
        self.view = pn.Param(self, parameters=["a_literal"], widgets=WIDGETS)
        
parameterized_app = ParameterizedApp()
parameterized_app.view
```

So I can input a `list` even though the parameter is a `dict`.

![image](https://user-images.githubusercontent.com/42288570/92947917-ee44b100-f458-11ea-8efa-ccacfd44b2b6.png)

MarcSkovMadsen:
and ... this is of course not possible

```python
import param
import panel as pn
pn.extension()

WIDGETS = {
    "a_literal": {
        "type": pn.widgets.LiteralInput, "sizing_mode": "fixed", "width": 400, "placeholder": "Input a dictionary!", "type": dict
    }
}

class ParameterizedApp(param.Parameterized):
    a_literal = param.Dict(default=None, label="A Dictionary")
    view = param.Parameter()
    
    
    def __init__(self, **params):
        super().__init__(**params)
        
        self.view = pn.Param(self, parameters=["a_literal"], widgets=WIDGETS)
        
parameterized_app = ParameterizedApp()
parameterized_app.view
```

```bash
----------------------------------------------------------------
AttributeError                 Traceback (most recent call last)
<ipython-input-5-bbb4723617e4> in <module>
     19         self.view = pn.Param(self, parameters=["a_literal"], widgets=WIDGETS)
     20 
---> 21 parameterized_app = ParameterizedApp()
     22 parameterized_app.view

<ipython-input-5-bbb4723617e4> in __init__(self, **params)
     17         super().__init__(**params)
     18 
---> 19         self.view = pn.Param(self, parameters=["a_literal"], widgets=WIDGETS)
     20 
[...]

PR #1583, opened 2020-09-15
Fix issue 1578: Update Param.object does not keep widgets

MarcSkovMadsen:
Readed this with (hopefully) proper line endings.

### DIFF
--- a/panel/.gitattribute
+++ b/panel/.gitattribute
@@ -1,0 +1,2 @@
+# See https://docs.github.com/en/github/using-git/configuring-git-to-handle-line-endings
+* text=auto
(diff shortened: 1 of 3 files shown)


MarcSkovMadsen:
Verified that the line endings look ok.

Please note I work on a lot of projects where everbody is on Windows and we don't use Linux line endings. Thus I cannot change my global git line endings.

But I added a `.gitattribute` file which should handle that all contributors to Panel has the correct line endings. Hope this will work for Panel.

image

MarcSkovMadsen:
image

Please note that it's not possible to both set the `type` of widget and the `type` of literal :-) There is an open issue on this here https://github.com/holoviz/panel/issues/1574.

MarcSkovMadsen:
@philippjfr . If you tell me what else I should do I will do it. If it's fixing the failing test you need to tell me how.

Issue #1581, opened 2020-09-14
When updating Param object the parameters changes


MarcSkovMadsen:
I am working on an app where the object of a pn.Param is dynamically changed. But unfortunately it does not keep the original parameters.

Similar problem in https://github.com/holoviz/panel/issues/1578. But just regarding the widgets and not the parameters.

The below test shows the problem

```python
import param
from panel import Param
from panel.widgets import TextAreaInput

import param

def test_change_object_and_keep_widgets():
    """Test that https://github.com/holoviz/panel/issues/1579 is solved"""
    # Given
    class TextModel(param.Parameterized):
        text = param.String()
        param2 = param.String()

    class TextView(param.Parameterized):
        text = param.ClassSelector(class_=TextModel)
        text_pane = param.Parameter()

        def __init__(self, **params):
            params["text"] = TextModel(text="Original Text")
            super().__init__(**params)

            self.text_pane = Param(
                self.text, parameters=["text"], widgets={"text": {"type": TextAreaInput}}
            )

        @param.depends("text", watch=True)
        def _update_text_pane(self, *_):
            self.text_pane.object = self.text

    view = TextView()
    assert view.text_pane.parameters==["text"]

    # When
    view.text = TextModel(text="New TextModel")
    # Then
    assert view.text_pane.parameters==["text"]
```

```bash
$ pytest 'scripts\issue_param.py' -x -s --doctest-modules
============================================================== test session starts ===============================================================
platform win32 -- Python 3.7.6, pytest-5.4.3, py-1.9.0, pluggy-0.13.1
rootdir: C:\repos\private\panel-extensions-template, inifile: pytest.ini
plugins: cov-2.10.0
collected 1 item                                                                                                                                  

scripts\issue_param.py F

==================================================================== FAILURES ====================================================================
______________________________________________________ test_change_object_and_keep_widgets _______________________________________________________

    def test_change_object_and_keep_widgets():
        """Test that https://github.com/holoviz/panel/issues/1581 is solved"""
        # Given
        class TextModel(param.Parameterized):
            text = param.String()
            param2 = param.String()

        class TextView(param.Parameterized):
            text = param.ClassSelector(class_=TextModel)
            text_pane = param.Parameter()

            def __init__(self, **params):
                params["text"] = TextModel(text="Original Text")
                super().__init__(**params)

                self.text_pane = Param(
                    self.text, parameters=["text"], widgets={"text": {"type": TextAreaInput}}
                )

[...]

PR #1584, opened 2020-09-15
fixed issue 1581

MarcSkovMadsen:
This should fix https://github.com/holoviz/panel/issues/1581

I've added a test and a one line fix. For me all tests pass.

But I simply don't understand the logic of the function I've changed. Are we sure that all combinations of events wil work?

image

### DIFF
--- a/panel/param.py
+++ b/panel/param.py
@@ -227,7 +227,8 @@ class Param(PaneBase):
                 parameters = [] if event.new == [] else event.new
 
         if parameters != [] and parameters != self.parameters:
-            self.parameters = parameters
+            if not self.parameters:
+                self.parameters = parameters
             return
 
         for cb in list(self._callbacks):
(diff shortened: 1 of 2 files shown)


MarcSkovMadsen:
@philippjfr . Let me know if there is more I should do.

panel/param.py
```diff
@@ -227,7 +227,8 @@ def _update_widgets(self, *events):
                 parameters = [] if event.new == [] else event.new
 
         if parameters != [] and parameters != self.parameters:
-            self.parameters = parameters
+            if not self.parameters:
+                self.parameters = parameters
```
philippjfr, line 231: This change does not seem correct to me, but I can see how the previous logic was incorrect too. Will keep your test though.

MarcSkovMadsen: makes sense

philippjfr:
Okay, added what I think is the correct fix. That said I also can't explain the `return` there. Will have to think about that more.

philippjfr:
Okay I understand it now and left comments. The `return` statement was there because setting `self.parameters` would trigger the method to be called again, which would then lead to the appropriate changes being applied.

PR #1585, opened 2020-09-15
Unpin pytest now nbsmoke 0.5 is released

ceball:
I guess you meant `nbsmoke` in the comment? (Although I bet `nbval` has exactly the same problem. And I would like to replace `nbsmoke` with `nbval`. Maybe that's why you wrote nbval? :) )

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -108,7 +108,7 @@ _recommended = [
 _tests = [
     'flake8',
     'parameterized',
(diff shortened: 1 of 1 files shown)


ceball:
(The push build failed because it happened before nbsmoke 0.5 was available to the build. The PR build happened later.)

philippjfr:
Thanks!

PR #1588, opened 2020-09-16
AutocompleteInput: allow user-defined values

### LINKED ISSUE #685: AutocompleteInput that allows non-pre-defined options as their value
**Is your feature request related to a problem? Please describe.**

This is not related to a problem.

**Describe the solution you'd like**

I am putting in a feature request for `AutocompleteInput` that allows non-predefined values as its value, as well as a feature request for `AutocompleteInput` to permit substring matching.

The use case is as follows: We have a database field that is free text but short, representing human-defined categories of stuff. (In our case, it's plates attached to experiments, where the experiment names are defined by humans.) 

At the moment, we use two text boxes, and tell our users to use the correct one: AutocompleteInput if they are attaching plates to an existing experiment, and a plain text input box if they are creating a new experiment. 

The front-end might be simplified by allowing users to enter the experiment name into the text box directly, and then choosing an existing experiment name from there. Permitting substring matching would be helpful in this case, for example we might have an experiment name that is "cherry-picking on mutant 53", but I might only remember 53 and enter it first.

**Describe alternatives you've considered**

I understand this may break the API compatibility with "single value, option selection" widgets. Hence, if this were a different widget, that would be cool too.

**Additional context**

No other additional context that I can think of.

ltalirz:
fixes #685

bokeh 2.3.0 introduces a new `restrict` keyword (default: True).
Setting it to False allows users to enter values that are not present in
the list of completions.

### DIFF
--- a/examples/reference/widgets/AutocompleteInput.ipynb
+++ b/examples/reference/widgets/AutocompleteInput.ipynb
@@ -26,6 +26,7 @@
     "\n",
     "* **``options``** (list or dict): A list or dictionary of options to select from\n",
     "* **``value``** (object): The current value; must be one of the option values\n",
+    "* **``restrict``** (boolean): Set to False in order to allow users to enter text that is not present in the options list.\n",
     "\n",
     "##### Display\n",
     "\n",
--- a/panel/package-lock.json
+++ b/panel/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "@holoviz/panel",
-  "version": "0.10.0-a24",
+  "version": "0.10.0-a25",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
--- a/panel/widgets/select.py
+++ b/panel/widgets/select.py
@@ -19,7 +19,7 @@ from bokeh.models.widgets import (
 
 from ..layout import Column, VSpacer
 from ..models import SingleSelect as _BkSingleSelect
-from ..util import as_unicode, isIn, indexOf
+from ..util import as_unicode, isIn, indexOf, bokeh_version
 from .base import Widget, CompositeWidget
 from .button import _ButtonBase, Button
 from .input import TextInput, TextAreaInput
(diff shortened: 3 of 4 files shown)


ltalirz:
It's not really clear to me what caused the CI to fail here.

From local testing, however, it seems to me that there is some incompatibility with the panel master branch and bokeh 2.3 (?)

ltalirz:
@philippjfr Is there something I should be doing to fix this PR?

I'm also trying to set up a development version of bokeh and panel (using `python setup.py install`, tried both with bokeh 2.2 and 2.3.0.dev1) side-by-side but I'm getting issues like those in the browser console and am not sure how to continue:

image

philippjfr:
Hmm, that sounds like problem with bokeh 2.3.0dev1 rather than anything in Panel.

ltalirz:
@philippjfr Thanks, indeed this was an issue in my development setup.

The following example seems to work fine
```python
import panel as pn
import bokeh.models.widgets as bw
import panel.widgets as pw
from bokeh.models.widgets import PreText
from bokeh.layouts import column
from bokeh.plotting import curdoc

pn.extension()

txt = bw.PreText(text="")
input_box = pw.AutocompleteInput(restrict=False)
input_box.title = "Foo Bar"
input_box.value = "400"
input_box.completions = ["100001", "12344556", "3194567289", "209374209374"]

def update():
    txt.text = f"Autocomplete value is {input_box.value}"

input_box.param.watch(lambda new_val: update(), 'value')

col = pn.Column()
col.extend([input_box, txt])
col.servable()
```

It's still not clear to me why the CI on this PR failed and whether I need to make any modifications.

philippjfr:
That's not caused by you, I'll have to adjust the CI setup.

philippjfr:
Thanks again for this. I don't want to require bokeh 2.3 yet. So two favors, please add a version check which defines an entry in the `AutocompleteInput ._rename` which renames it to None when bokeh<2.3. Secondly please document the `restrict` keyword in the `AutocompleteInput` notebook in `examples/reference/widgets`.

ltalirz:
Done! 

Here just a check that the version selector does what it is supposed to do:
```python
>>> from distutils.version import LooseVersion
>>> LooseVersion('2.3.0dev1') < '2.3.0'
False
>>> LooseVersion('2.2.0') < '2.3.0'
True
>>> LooseVersion('2.3.0') < '2.3.0'
```

philippjfr:
Thanks!

PR #1596, opened 2020-09-21
add option to serialize all vtkDataset arrays

xavArtley:
A vtk dataset can contain several data arrays but only one is used to color the mesh, the "active scalars".
Until now only this array was serialized, it allow to minimize the size of the serialized scene.
However by allowing the serialization of all data arrays and using some custom javascript functions we could make some temporal 3d visualization easier.
This PR add a parameter `serialize_all_data_arrays` to allow this
Here is a comparison between both options:
vtk_all_arrays

### DIFF
--- a/panel/pane/vtk/synchronizable_serializer.py
+++ b/panel/pane/vtk/synchronizable_serializer.py
@@ -7,8 +7,7 @@ import time
 import zipfile
 
 from vtk.vtkCommonCore import vtkTypeUInt32Array, vtkTypeInt32Array
-from vtk.vtkFiltersGeometry import vtkCompositeDataGeometryFilter
-from vtk.vtkFiltersGeometry import vtkGeometryFilter
+from vtk.vtkFiltersGeometry import vtkCompositeDataGeometryFilter, vtkGeometryFilter
 from vtk.vtkRenderingCore import vtkColorTransferFunction
 
 # -----------------------------------------------------------------------------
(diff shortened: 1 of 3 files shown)


panel/pane/vtk/synchronizable_serializer.py
```diff
@@ -566,11 +630,24 @@ def genericProp3DSerializer(parent, prop3D, prop3DId, context, depth):
...
+    # backfaceProperties = actor.GetBackfaceProperty()
+
+    # if backfaceProperties:
+    #     backfacePropId = context.getReferenceId(backfaceProperties)
+    #     backPropertyInstance = serializeInstance(
+    #         actor, backfaceProperties, backfacePropId, context, depth + 1)
+    #     if backPropertyInstance:
+    #         instance['dependencies'].append(backPropertyInstance)
+    #         instance['calls'].append(['setBackfaceProperty', [wrapId(backfacePropId)]])
+
```
philippjfr, line 650: Keep this?

panel/pane/vtk/vtk.py
```diff
@@ -199,14 +199,23 @@ class BaseVTKRenderWindow(AbstractVTK):
...
          all the scene is rebuilt from scratch.
     """)
 
-    _rename = {'serialize_on_instantiation': None}
+    serialize_all_data_arrays = param.Boolean(default=False, constant=True, doc="""
+        If true, enable the serialization of all data arrays of vtkDataSets (point data, cell data and field data).
+        By default the value is False and only active scalars of each dataset are serialized and transfer to the
+        javascript side.
+
+        Enabling it will increase memory and data transfert usage but allows to create more reactive visualization
```
philippjfr, line 214:
```suggestion
        Enabling this option will increase memory and network transfer volume but enables results in more reactive visualizations
```

philippjfr:
Apart from a few comments this looks fine to me. However I do worry about continuing to expand the custom serializer so just checking where are we on using the native vtk serializer?

panel/pane/vtk/synchronizable_serializer.py
xavArtley: Yes we can keep this if vtkjs is going to support backface color we won't have to rethink about it

xavArtley:
> Apart from a few comments this looks fine to me. However I do worry about continuing to expand the custom serializer so just checking where are we on using the native vtk serializer?

using the function VTK.import_scene , we could use any serializer which create a file wich can be read in https://kitware.github.io/vtk-js/examples/SynchronizableRenderWindow.html

philippjfr:
Okay, merging for now.

PR #1615, opened 2020-10-07
Record session statistics on pn.state.session_info

philippjfr:
Stores various information about user sessions on `pn.state.session_info`. Currently this records the following fields per session:

- `started` (float): Timestamp when the session was started
- `rendered` (float): Timestamp when the session was fully rendered
- `ended` (float): Timestamp when the session was closed on the server
- `user_agent` (str): A string identifying the user agent

- [ ] Add tests
- [x] Add docs


jbednar:
Cool! Will the session info ever expire? If not, won't storing that data lead to ever-increasing memory usage over time?

philippjfr:
>Cool! Will the session info ever expire? If not, won't storing that data lead to ever-increasing memory usage over time?

This is true, that said it's about 1.6 KB per session so even at 100k sessions that's not a huge amount. What I might suggest is that it keeps only the most recent N sessions where N is configurable.

philippjfr:
Okay, I've now reduced it to 0.6 kB per session and added the `config.session_history` option.

jbednar:
> What I might suggest is that it keeps only the most recent N sessions where N is configurable.

Sounds good. It should probably also keep the number of sessions as a global value, so that we can tell how much data has been lost due to reaching the session_history limit.

panel/config.py
```diff
@@ -79,6 +79,12 @@ class _config(param.Parameterized):
         embedded. Useful when only partial updates are made in an
         app, e.g. when working with HoloViews.""")
 
+    session_history = param.Integer(default=0, bounds=(-1, None), doc="""
+        If set to a non-zero value this determines the maximum length
```
jbednar, line 83:
```suggestion
        If set to a non-negative value this determines the maximum length
```

(previously it was unclear what happens at zero, but I'm pretty sure zero just acts like positive numbers do, so I've reworded it to say so).

Issue #1598, opened 2020-09-22
FileDownloadWidget crashes with bokeh 2.3


ltalirz:
### ALL software version info
panel `master` branch
bokeh 2.3.0.dev1

#### Description of expected behavior and the observed behavior
Expected: Be able to create a `FileDownload` widget.

Observed: Fails with JS console error `panel.models.widgets.FileDownload(1002).data given invalid value: null`.

Workaround: Set `download.data = ""`.

#### Complete, minimal, self-contained example code that reproduces the issue

```python
import panel as pn
import panel.widgets as pw

pn.extension()

btn_download = pn.widgets.FileDownload(
            filename='data.json',
            button_type='primary',)
# uncomment line below to silence console error
#btn_download.data = ""

col = pn.Column()
col.extend([btn_download])
col.servable()

```

#### Screenshots or screencasts of the bug in action
![image](https://user-images.githubusercontent.com/1053245/93833960-840ee600-fc7a-11ea-9982-7d574d22315a.png)

PR #1622, opened 2020-10-12
Fixed bugs with None default on String params

philippjfr:
Fixes https://github.com/holoviz/panel/issues/1598


philippjfr:
@mattpap What should the property type be for a property that can be either null or a p.String?

philippjfr:
Nevermind, found the NullString type.

Issue #1257, opened 2020-04-10
Setting `value_throttled` does not set the initial widget value


joelostblom:
From https://discourse.holoviz.org/t/how-to-throttle-values-on-mouseup/456/7?u=joelostblom and the two following comments there.

`value_throttled` does not change the initial widget value. Take this example, using `value`, both the initial slider value and the and the param.value attribute are set to (7, 9).

```python
int_range_slider = pn.widgets.IntRangeSlider(end=10, value=(7, 9))

@pn.depends(int_range_slider.param.value, watch=True)
def print_slider_value(slider_value):
    return slider_value

pn.Column(int_range_slider, print_slider_value)
```

![image](https://user-images.githubusercontent.com/4560057/79005807-8f2aeb00-7b0c-11ea-9d08-75c7266f80e4.png)

On the other hand,`value_throttled` only sets `param.value_throttled`, not the slider value:

```python
int_range_slider = pn.widgets.IntRangeSlider(end=10, value_throttled=(7, 9))

@pn.depends(int_range_slider.param.value_throttled, watch=True)
def print_slider_value(slider_value):
    return slider_value

pn.Column(int_range_slider, print_slider_value)
```

![image](https://user-images.githubusercontent.com/4560057/79005852-a9fd5f80-7b0c-11ea-812c-8a617f24e55f.png)

So if I wanted to set the initial range of the slider, I can't do it with just `value_throttled`, I also need to set `value` to the same tuple, which seems redundant.

```
-----
panel       0.9.5
param       1.9.3
sinfo       0.3.1
-----
IPython             6.5.0
jupyter_client      5.2.3
jupyter_core        4.6.3
jupyterlab          2.1.0
notebook            5.6.0
-----
Python 3.7.6 | packaged by conda-forge | (default, Mar 23 2020, 23:03:20) [GCC 7.3.0]
Linux-5.6.2-arch1-2-x86_64-with-arch
```

philippjfr:
We should not allow setting `value_throttled` imo, it should be readonly.

On an unrelated note, none of your `pn.depends` decorators should have `watch=True`, that is only necessary for `pn.depends` which work via side-effects.

joelostblom:
> We should not allow setting `value_throttled` imo, it should be readonly.

:+1: Could it be set to the same as `value` upon initialization? Just so that it does not start as `None` or some other value than what is on the slider.

[...]

PR #1624, opened 2020-10-12
Initialize value throttled and set to constant

philippjfr:
Fix https://github.com/holoviz/panel/issues/1257


xavArtley:
should be changed here too
https://github.com/holoviz/panel/blob/b07ad9d3e4da578452d55a8c896d627bf6a2afb9/panel/widgets/input.py#L243-L254

Issue #1630, opened 2020-10-14
VTKVolume: Toggle option for dropdown UI in image


JeffreyWardman:
Problem: Dropdown UI widget (represented by the VTKjs favicon logo) in image is on by default, covering the image (easily reproducible with pn.Row(image, image, image)). 

Solution: Option to turn on/off the widget (with off being the default setting)

Additional: Would be useful to customize the dropdown UI widget itself as well.

philippjfr:
Implemented in #1631

JeffreyWardman:
@philippjfr Keeping this closed is fine. I think it's bad UX to have it open by default. When the image/window is small, the UI covers a large part of the object (the brain in #1631). For instance, a row consisting of the tabs, and three of these brain images results in a frustrating experience.

PR #1631, opened 2020-10-14
[VTKVolume] add option to expand the controller

xavArtley:
implements https://github.com/holoviz/panel/issues/1630

by default I let it to true 

controller_expand

Customizing the controller widget is doable but it should be more a vtk.js request since I use directly the VolumeController class defined in vtkjs and there are not options to customize it (https://kitware.github.io/vtk-js/api/Interaction_UI_VolumeController.html)


philippjfr:
Looks good, thanks!

Issue #1602, opened 2020-09-25
[VTK] Remove orientation widget interactivity to update vtkjs version


xavArtley:
For the moment I can't update vtkjs version mainly related to these issues:
https://github.com/Kitware/vtk-js/issues/1493
https://github.com/Kitware/vtk-js/issues/1381

However if I remove the interactivity of the orientation widget, it is not a problem anymore
The interactivity of orientation widget allow to set the XY, YZ or XZ view but we could provide other ways to achieve this behavior (trigger a callback with a variable change for example)

@jbednar, @philippjfr 
Do you think it's legitimate to remove the interactive behavior of the orientation for future panel version?
Being able to update the vtk.js version will simplify the integration of future improvements

philippjfr:
I'd be in favor of removing that interactivity, at least temporarily, if it allows us to upgrade vtkjs versions. Would be great to restore it later but for now I favor compatibility over this feature.

PR #1635, opened 2020-10-15
[VTK] optional interactive orientation widget

xavArtley:
following https://github.com/holoviz/panel/issues/1602 discussion I add a parameter to make the orientation widget interactive optional

By default the interactive_orientation_widget is True except for VTKRenderWindowSynchronized class where errors occur when we try to synchrnonize the view. I let the possibility to set it to true but in this case a warning is raised.

With this modification I updated the vtkjs to the latest version, it allows me to remove 2 files which were reimplementation of vtkjs corrections


xavArtley:
In the same time I added some corrections about the camera property state change

philippjfr:
Looks good thanks! Can you document the parameter in the reference notebook?

PR #1636, opened 2020-10-15
Add channels to extra dependencies

Hoxbro:
To be able to run the Django and panel together, [channels](https://channels.readthedocs.io/en/latest/) is needed. 

A reason why it was not added together with Django at first, is that it was not on conda-forge before April this year.

I have also updated the documentation for the Django Apps page.


philippjfr:
Thanks! Need to remove that `npm whoami` call to get external tests passing again but your change looks good so I'll merge.

PR #1637, opened 2020-10-15
[VTK] Synchronization demo

xavArtley:
I added a new VTK example to demonstrate the use of the VTKRenderWindowSynchronized pane
temp_demo

I already began some examples in https://github.com/holoviz/panel/pull/1287

Was wondering if I should merge all in one notebook


philippjfr:
Let's start by merging this one. Let me know about the status of the other one.

Issue #1641, opened 2020-10-18
Vanilla Template Issues


MarcSkovMadsen:
I have been trying the new Vanilla Template on the Master branch using the example from the reference guide.

I see some minor issues that I believe should be fixed.

**Issue - To many Scrollbars**

When I open it looks like this

![image](https://user-images.githubusercontent.com/42288570/96359002-d3c3be80-110d-11eb-819b-c5f01435f4a7.png)

I see 3 active scrollbars and one scrollbar (on sidebar) that is disabled. Visually this does not not look very well. It should be possible to get the template/ app to adjust and not need the 3 active scrollbars.

**Issue - Does not resize properly when "restored down" via square**

When I resize (by clicking square in upper right corner) it looks like below

![image](https://user-images.githubusercontent.com/42288570/96359028-2a30fd00-110e-11eb-8212-165c6b80ce13.png)

That is not satisfying either as the sidebar now is too small to contain the widgets and the plots don't resize to use the available width. I think the sidebar should have a user configurable fixed width.

**Issue - Dragging leads to lots of scrollbars**

When I resize by dragging the right side it adjusts again but I get the 3 active scrollbars again.

![image](https://user-images.githubusercontent.com/42288570/96359061-7c721e00-110e-11eb-90bd-3740232643c3.png)

**Issue - Header is not sticky**

Another thing is that the top bar is not "sticky". I would really expect that in a modern application.

![image](https://user-images.githubusercontent.com/42288570/96359076-9ca1dd00-110e-11eb-9e90-554d555fc2fa.png)

**Issue - Too little margin**

I would also like more margin as where marked below. Maybe it is a subjective thing?

![image](https://user-images.githubusercontent.com/42288570/96359095-eee2fe00-110e-11eb-88a9-af27941cb6ed.png)

MarcSkovMadsen:
**Issue - Cannot Use Header**

It's a normal thing to want to add drop down menus, texts and buttons to a header. And want to add some to the left side and some to the right side.

As there is just on `header` list I tried adding

```python
vanilla.header[:]=[
    pn.widgets.Button(name="left", sizing_mode="fixed", width=100),
    pn.Spacer(),
    pn.widgets.Button(name="Right", sizing_mode="fixed", width=100),
]
```

But that does not look nice.

![image](https://user-images.githubusercontent.com/42288570/96359244-e8558600-1110-11eb-86c8-de4211b4846b.png)

[...]

PR #1642, opened 2020-10-18
fix vanilla issues

MarcSkovMadsen:
I've tested out the Vanilla Template and could see it would not work for users. 

I've listed some issues here https://github.com/holoviz/panel/issues/1641 that this PR tries to solve.

You can try out the vanilla template example by running

```python
python -m panel serve 'panel\tests\template\test_vanilla_manual.py' --dev
```

vanilla (1)

Comments:

- I've fixed the sidebar width at `300px` as that in my experience is a good width. It's also the width used by Streamlit and I've never heard users propose any change to that. Maybe the width should be user customizable such that you can specify a string like `250px` or `16.67vw`?
- I've added a little bit more margin/ padding (~`10px`) in different places to make things look a little less busy. Maybe that should be user customizable as well.
- I've setup the template such that the sidebar button and title always have a width of 300px (same as sidebar). Then the header items will always be on top of the main area. In my experience that is a very nice looking layout. But if people have a very large app title it can cause problems.
- I've set `overflow-x: hidden` on both sidebar and main area. In my experience that is a very, very good solution. But if some people have very wide plots or images they would like to show it will not work. But in my experience there should be as few scrollbars as possible. There is no horizontal scrollbar in the Streamlit template and I've never heard users complain.

FYI @nghenzi and @philippjfr . This might have the implications for the other templates and themes like the React Template.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -121,3 +121,4 @@ builtdocs/
 discover/
 
 node_modules/
+tmp/
--- a/panel/template/vanilla/default.css
+++ b/panel/template/vanilla/default.css
@@ -3,7 +3,7 @@
 }
 
 #sidebar {
-  background-color: white;
+  background-color: 	#F8F8F8;
   box-shadow: 0px 0px 1px;
 }
 
--- a/panel/template/vanilla/vanilla.css
+++ b/panel/template/vanilla/vanilla.css
@@ -2,17 +2,32 @@ body {
   height: 100vh;
   margin: 0px;
   font-family: "Lato", sans-serif;
+  overflow-x: hidden;
+  overflow-y: hidden;
 }
 
 #container {
   padding:0px;
   width: 100vw;
+  max-width: 100vw;
 }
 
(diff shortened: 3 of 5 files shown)


philippjfr:
Thanks for these fixes, but it looks like you seem to have started with an older version of this template. Consequently this PR needs to be rebased.

MarcSkovMadsen:
How do I do that @philippjfr ? I did a git pull from the master branch before starting. Why does that not work?

philippjfr:
That should work just fine but I'm guessing you did that a while ago or made some mistake.

philippjfr:
Now you'll have to git merge or git rebase from master and then resolve the differences manually.

MarcSkovMadsen:
I must have made a mistake. Started working on it this morning. Damn :-)

MarcSkovMadsen:
Closed in favor of https://github.com/holoviz/panel/pull/1647

PR #1646, opened 2020-10-18
Configure Git to use Linux line endings for this repository.

MarcSkovMadsen:
I can see that when I contribute to Panel the line endings are changed. Thus it looks like the whole file is changed instead of a few lines.

I'm working on windows and normally with windows users. I cannot change my Git Global settings for line endings as that will change a lot of files in a lot of repositories.

But according to https://docs.github.com/en/free-pro-team@latest/github/using-git/configuring-git-to-handle-line-endings it's possible to add a `.gitattributes` to handle the problem for the local repository.

I've done that and then run `git add --renormalize .` to change any line endings that are not Linux line endings to Linux line endings.


xavArtley:
Hello, why don't you set it locally to your panel source directory?
`git config --local core.autocrlf true`

MarcSkovMadsen:
@xavArtley. Because I did not know that option and it was not described in the Git documentation I found. It's not described in the developer guide either https://panel.holoviz.org/developer_guide/index.html.

But I still think its a good idea to add the .gitattributes file as then the problem is solved once and for all. Philipp, you and others don't need to support Windows users like me on that question.

Issue #1644, opened 2020-10-18
Vanilla Template has margin-left of sidebar also when no sidebar


MarcSkovMadsen:
I've tried a small example using the Vanilla Template in the current branch of Panel. This example has no sidebar and it does not work well because it keeps the margin of the sidebar. See below

### Solution

The margin of the sidebar should be removed. Instead the user should be able to set a max-width of the main area. If no max-width is set the full width will be used (with a little bit of margin/ padding). If a max-width is set the main container is centered and has the specified main-width.

The reason for this is that if the app is primarely text/ image based then normally a max-width is set to enable a user to better read the content. For example in Bootstrap this is one of the first things you learn to do.

### Example

![image](https://user-images.githubusercontent.com/42288570/96364450-0e901b80-113b-11eb-85cb-d5b9613fcb23.png)

```python
import panel as pn
import numpy as np
import holoviews as hv


def test_template_with_no_sidebar():
    """Returns an app that uses the vanilla template in various ways.

Inspect the app and verify that the issues of [Issue 1641]\
(https://github.com/holoviz/panel/issues/1641) have been solved

- Navbar is "sticky"/ fixed to the top
- Navbar supports adding header items to left, center and right
- There is a nice padding/ margin everywhere
- Independent scroll for sidebar and main
- Only vertical scrollbars
"""
    vanilla = pn.template.VanillaTemplate(title="Vanilla Template")

    pn.config.sizing_mode = "stretch_width"

    xs = np.linspace(0, np.pi)
    freq = pn.widgets.FloatSlider(name="Frequency", start=0, end=10, value=2)
    phase = pn.widgets.FloatSlider(name="Phase", start=0, end=np.pi)

    @pn.depends(freq=freq, phase=phase)
    def sine(freq, phase):
        return hv.Curve((xs, np.sin(xs * freq + phase))).opts(responsive=True, min_height=400)

    @pn.depends(freq=freq, phase=phase)
    def cosine(freq, phase):
        return hv.Curve((xs, np.cos(xs * freq + phase))).opts(responsive=True, min_height=400)

    vanilla.main.append(freq)
    vanilla.main.append(phase)
    vanilla.main.append(pn.pane.Markdown(test_template_with_sidebar.__doc__))
    vanilla.main.append(
        pn.Row(
            pn.Card(hv.DynamicMap(sine), title="Sine"),
            pn.Card(hv.DynamicMap(cosine), title="Cosine"),
        )
    )
    vanilla.main.append(
[...]

PR #1647, opened 2020-10-18
Fix Vanilla Template v.2

MarcSkovMadsen:
Supersedes https://github.com/holoviz/panel/pull/1642. I had to fix some line endings and Git base branch issues.

----------

I've tested out the Vanilla Template and could see it would not work for users. 

I've listed some issues here https://github.com/holoviz/panel/issues/1641 that this PR tries to solve.

You can try out the vanilla template example by running

```python
python -m panel serve 'panel\tests\template\test_vanilla_manual.py' --dev
```

vanilla2

Comments:

- I've fixed the sidebar width at `300px` as that in my experience is a good width. It's also the width used by Streamlit and I've never heard users propose any change to that. Maybe the width should be user customizable such that you can specify a string like `250px` or `16.67vw`?
- I've added a little bit more margin/ padding (~`10px`) in different places to make things look a little less busy. Maybe that should be user customizable as well.
- I've set `overflow-x: hidden` on both sidebar and main area. In my experience that is a very, very good solution. But if some people have very wide plots or images they would like to show it will not work. But in my experience there should be as few scrollbars as possible. There is no horizontal scrollbar in the Streamlit template and I've never heard users complain.
- I've given the sidebar just a little bit of gray to more clearly separate it from the rest.

FYI @nghenzi and @philippjfr . This might have the implications for the other templates and themes like the React Template.


MarcSkovMadsen:
I believe this template will be useable on mobile to some extent. The only thing that would be nice would be if it was possible to assign a size grid of 12 columns like the Bootstrap or Material Grid such that the main items would be layed out differently on a widescreen and on mobile.

vanilla_mobile

MarcSkovMadsen:
I've also fixed the issue with the 300px margin when there is no sidebar.

image

https://github.com/holoviz/panel/issues/1644

MarcSkovMadsen:
I've added a new parameter called `main_max_width` to the template. The reason is that one of the basic things in a lot of web pages is that they center the main content and give it a max size in order to make sure it's not too wide to read/ work with.

I've set the `main_max_width` to '' (empty string) by default. And it will not be applied.

The screenshots below show examples with main_max_width of `600px`.

image

image

MarcSkovMadsen:
FYI @philippjfr . 

I've "fixed" the issues I have seen and tried to improve it based on the experiences I have had with the awesome-panel.org template and site for almost a year.

I don't plan to do further improvements except based on your review.

philippjfr:
Really appreciate your work here. I expect a rapid series of 0.10.x releases just for fixes like this at this point. Will review tomorrow and then I'm pretty much ready for the 0.10.0 release by about Tuesday. The blog post is finally written and I should have some time to address final issues tomorrow.

MarcSkovMadsen:
I've done some further improvements. I know from other communities that being able to use custom favicons and support search engine optimization is badly requested.

So I've added support for favicon, meta and base data. I also changed the code a bit so the title is also shown in the browser tab instead of always 'Panel Application'.

image

image

panel/template/base.py
```diff
@@ -37,6 +37,9 @@
     'https://localhost:{port}</a>'
 )
 
+# Should be served from somewhere else!
+FAVICON_URL = "https://raw.githubusercontent.com/MarcSkovMadsen/awesome-panel/2781d86d4ed141889d633748879a120d7d8e777a/assets/images/favicon.ico"
```
MarcSkovMadsen, line 41: You should serve this from somewhere else @philippjfr

panel/template/base.py
```diff
@@ -393,6 +430,7 @@ def _css_files(self):
         return css_files
 
     def _init_doc(self, doc=None, comm=None, title=None, notebook=False, location=True):
+        title = title or self.title
```
MarcSkovMadsen, line 433: Import to be able to show user specified title in browser tab.

philippjfr:
First comment, would you mind removing unrelated changes to the CHANGELOG, alert and .git files?

MarcSkovMadsen:
Hi @philippjfr . As stated above Please review (and merge) https://github.com/holoviz/panel/pull/1646 first. Then those CHANGELOG, alert and .git files changes should go away AND THE CHANGES SHOULD BE CLEAR AS THERE IS NO WINDOWS/ LINUX line endings discrepancies.

Let me know if this is not possible.

philippjfr:
Going to merge this and follow up in another PR.

PR #1650, opened 2020-10-19
Improvements for template favicon and metadata handling

philippjfr:
Cleaning up recent PRs adding favicon and metadata handling for the basic templates.

- Bundles the favicon.ico into the panel.js distribution so it's available as a server resource and will be uploaded to NPM
- Removes default metadata

To Do:

- [ ] Update docs

@MarcSkovMadsen


MarcSkovMadsen:
Hi @philippjfr 

Looks nice. If I can find time I will try to contribute the changes to the docs. But right now I don't if it will be today or within a week. So feel free to do it first.

The changes you have made looks nice. 

I can see you have set keywords, description etc. to "". I was in doubt here. The reason why I added something is that have Panel, HoloViz etc. show up in the description, keywords etc. on several sites will make it more know by Google Search Engine as far as I understand. Thus people who use default will be advertising a bit for HoloViz and Panel which is a small price to pay for a great product. But I also understand if you don't want users to "pay that price".

philippjfr:
Yeah, I get your reasoning there. Maybe I should promote more but feel like it'd be a little bit rude to insert that everywhere. Can reconsider that later. For now I'll merge.

PR #1651, opened 2020-10-19
Bundle external JS dependencies for custom models

philippjfr:
During build time this PR would pull all external resources specified on `Model.__javascript__` definitions and bundle them into the `dist` folder, which means when `resources='server'` they could be served locally.

The downside to this approach is that this grows the Python distribution by about 9 MB and this is only set to grow as more models are added. However this is the only approach that will allow these models to work in an air-gapped environment.

- [x] Switch `__javascript__` dynamically depending on the current resources setting

@mattpap @MarcSkovMadsen @jbednar @xavArtley Opinions welcome here


jlstevens:
I think at this stage of the project it is a good idea. That said, I think we should decide on a size limit and put it in the tests...e.g a 15MB limit for now and fail if we grow past that. Whether automated or not, I think we need to keep an eye on the total size to make sure it doesn't explode.

philippjfr:
So this will not work for all external dependencies, in particular if a dependency expects other files to be present alongside the main js/css files. As an example katex expects various woff font files to be present. So for now I'd only enable this for a subset of models.

Issue #1654, opened 2020-10-20
Nb templates is not used in templates


xavArtley:
I checked the VTKSlicer notebook and I found the notebook template was not applied

With further investigation I noticed:
the constructor signature of `BaseTemplate` is the followinf one:
https://github.com/holoviz/panel/blob/3961571aecf125e642bdfd092e99189a18ba551d/panel/template/base.py#L60-L62
and the Template constructor signature is:
https://github.com/holoviz/panel/blob/3961571aecf125e642bdfd092e99189a18ba551d/panel/template/base.py#L600-L602
as we can notice `nb_template` in the `Template` constructor is passed as the second argument of `BaseTemplate` parent class which is `items` and `nb_template` remain to None

PR #1660, opened 2020-10-20
fix nb_template in templates

xavArtley:
fix https://github.com/holoviz/panel/issues/1654


philippjfr:
Thanks, appreciate it!

PR #1662, opened 2020-10-20
Material Template Fixes

MarcSkovMadsen:
I have tested the Material Template and see some issues that I would like to fix

- The header items overflow the header on the right
- The header has no box shadow (as the Vanilla template has)
- There are a lot of scrollbars.
- A little bit more padding/ margin is need.
- Cannot set the main max-width. 
- Add a little bit of gray to the sidebar.

image

I will develop the fixes ASAP. Will change the Material Template to behave as the Vanilla Template.

### DIFF
--- a/panel/template/theme.py
+++ b/panel/template/theme.py
@@ -7,7 +7,7 @@ class Theme(param.Parameterized):
     """
     A Theme customizes the look and feel of a Template by providing
     custom CSS and bokeh Theme class.
-   
+
     When adding a new theme a generic Theme class should be created,
     which is what users will important and set on the Template.theme
     parameter. Additionally a concrete implementation of the Theme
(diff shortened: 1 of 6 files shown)


philippjfr:
See https://github.com/holoviz/panel/pull/1661

philippjfr:
Sorry misread, thought this was Golden layout as well.

philippjfr:
>There are a lot of scrollbars.

So I don't know what's up with that, I do not see scrollbars, only when I actually try to scroll.

philippjfr:
That's probably specific to my particular browser tbh.

philippjfr:
Actually I can't any browser to show all the scrollbars all the time, is this a windows thing?

nghenzi:
I see the scrollbars in windows too with chrome

philippjfr:
Okay, clearly needs fixing but really I blame windows 😜

MarcSkovMadsen:
The only thing I know is that `overflow: auto` is handle differently by Chrome and Firefox. As far as I remember Chrome shows disabled scrollbars while Firefox only shows the scrollbars when needed. Therefore I always set them to `scroll` instead of `auto` because then they behave the same and I don't need to handle margin/ padding/ width differences between browsers.

MarcSkovMadsen:
image

MarcSkovMadsen:
FYI. I will start working on this.

MarcSkovMadsen:
I've now implemented the fixes needed and aligned with the VanillaTemplate.

I have some ideas on providing a MaterialTheme for Bokeh. I will try to add that as well.

MarcSkovMadsen:
I've now added a Bokeh MaterialTheme and MaterialDarkTheme but then 

- the default.css is no longer imported (and it gets the purple header as shown below)
- and I don't think the bokeh theme is actually changed. The font in the plots does not look bigger and Roboto like

image

panel/template/theme.py
```diff
@@ -59,3 +59,113 @@ class DarkTheme(Theme):
 
     bokeh_theme = param.ClassSelector(class_=(_BkTheme, str),
                                       default=_BkTheme(json=BOKEH_DARK))
+
+MATERIAL_FONT = "Roboto, sans-serif, Verdana"
+MATERIAL_THEME = {
```
philippjfr, line 64: This should live in panel/template/material/__init__.py

philippjfr:
Again, in the interest of making quick progress here I'll merge and make a new PR. You have access rights for this repo so please push directly to holoviz/panel rather than your own fork so we can collaborate more easily.

panel/template/theme.py
MarcSkovMadsen: That's awesome. I just did not know that one. Will move it.

MarcSkovMadsen:
@philippjfr . Did you have any ideas on fixing the below

-----

I've now added a Bokeh MaterialTheme and MaterialDarkTheme but then

- the default.css is no longer imported (and it gets the purple header as shown below)
a- nd I don't think the bokeh theme is actually changed. The font in the plots does not look bigger and Roboto like

MarcSkovMadsen:
Hi @ceball 

I've have heard you have an interest in applications using Material Theme. If you have any ideas, guidance, requirements feel free to comment. Thanks.

philippjfr:
>the default.css is no longer imported (and it gets the purple header as shown below)

That is deliberate, the default.css corresponds to the default theme. I'm still trying to understand why you created the MaterialTheme instead of extending the DefaultTheme but maybe it's just not well documented and that was your intention?

MarcSkovMadsen:
The intention is to make the Bokeh/ HoloViews plot look consistent with the bootstrap theme. This includes the "Roboto" font, some font size changes, some color changes etc.

Maybe I implemented it wrongly. I believe the Bokeh JSON THEME is correct. But maybe I need some guidance on how to move on from there @philippjfr ?

MarcSkovMadsen:
@philippjfr . One temporary solution for getting material themed widgets could be using a small script to replace the bokeh classes with the material classes. Maybe it's as simple as defining the conversions in a dictionary.

What do you think?

The thing I don't know is if I can get the script to execute every time the css class changes. But maybe I can using a MUTATIONOBSERVER.

If it works it could be a really easy and powerful thing.

### POC

image

image

MarcSkovMadsen:
One problem right now is I don't know how to find documentation for MASTER. I can only find documentation for 0.9.7.

Maybe a link to the MASTER documentation in the GITHUB README or the DEVELOPER GUIDE would be nice to have.

philippjfr:
Here's where those builds go: https://pyviz-dev.github.io/panel/

xavArtley:
@philippjfr
when is updated this site?
For what I see the version is  Panel 0.10.0a24.post1+g2b9c98f documentation

philippjfr:
The builds have been broken for a while, trying to fix that. It's usually rebuilt on each dev release or when a commit ending in `[doc-build]` is merged.

PR #1677, opened 2020-10-22
Template Fixes

MarcSkovMadsen:
So far I've

- Vanilla Template apply `main_max_width`.
- Set `max-with: 100%` for img. It's a hazzle using images in Markdown without this.

Such that this looks nice

image


MarcSkovMadsen:
Same fixes for MaterialTemplate such that this is ok.

material_working

MarcSkovMadsen:
Ready for review @philippjfr

PR #1689, opened 2020-10-23
Correction resize of Golden Template

xavArtley:
Current behavior:
Peek 23-10-2020 20-53

After fix:
Peek 23-10-2020 21-01


philippjfr:
Looks like in both there's a bit of margin missing on the right.

xavArtley:
I have no personal opinion on this.

Issue #1133, opened 2020-03-09
Add support for echarts and pycharts


MarcSkovMadsen:
Plotting libraries. Can always use more. This looks super cool and mature

https://github.com/pyecharts/pyecharts

It's based on e-charts which is an apache incubator project.

https://github.com/apache/incubator-echarts

Looks super easy to support as it can do a json dump that just needs to be transfered to the client.

![image](https://user-images.githubusercontent.com/42288570/76235406-5527aa00-622b-11ea-98bc-fafbf4362a72.png)

MarcSkovMadsen:
I took a look of what is required to implement this.

A basic example of the usage is

https://www.echartsjs.com/en/tutorial.html#Get%20Started%20with%20ECharts%20in%205%20minutes

I thought about whether we should just use WebComponent to provide this. But I would think that there could be some special requirements and I guess we would also like to transfer each data series efficiently as a ColumnDataSource.

I would say the implementation should be very similar to the Altair/ Vega implementation. So that that is where we should start.

The basic format of the plot is

```json
{
            title: {
                text: 'ECharts entry example'
            },
            tooltip: {},
            legend: {
                data:['Sales']
            },
            xAxis: {
                data: ["shirt","cardign","chiffon shirt","pants","heels","socks"]
            },
            yAxis: {},
            series: [{
                name: 'Sales',
                type: 'bar',
                data: [5, 20, 36, 10, 10, 20]
            }]
        };
```

Its the `data` of each series we would like to transfer as a `ColumnDataSource`.

MarcSkovMadsen:
I started adding support using the `WebComponent`pane.

See https://github.com/holoviz/panel/pull/1122#issuecomment-599179666

MarcSkovMadsen:
Done with Panel 0.10

PR #1691, opened 2020-10-24
Add support for Pyecharts

MarcSkovMadsen:
Finalizes the request for Echarts **and** Pyecharts https://github.com/holoviz/panel/issues/1133.

I believe it's important to support pyecharts as this is **THE** Python wrapper for ECharts and it's highly popular with ~10.000 stars on Github.

I've now added support for Pyecharts. I've tried to see how other plotting panes support dictionary specification and python wrapper specification.

I've added a test for ECharts and a manual test for pyecharts. I've tried to define the manual test for pyecharts such that its not disturbing the normal test suite, but can be run manually on demand.

I've added a comment to the ECharts notebook about suppport for pyecharts. But I've not provided an example as I was not sure you would like to add that dependency. But there is a simple example in the `test_echart.py` file.

You can test the manual example via `python -m panel serve 'panel\tests\pane\test_echart.py' --dev`

image

FYI @philippjfr . It's ready for review

### Note

I cannot get BokehJS to load in the Notebook. I don't believe this has anything to do with the `pyecharts` support as I could not before as well.

image

### DIFF
--- a/examples/reference/panes/ECharts.ipynb
+++ b/examples/reference/panes/ECharts.ipynb
@@ -14,13 +14,13 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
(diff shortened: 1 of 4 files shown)


philippjfr:
Thanks, all looks good. Got lucky this goes in before the release since I got held up on releasing param yesterday.

Issue #1693, opened 2020-10-25
Support providing items for sidebar and main in Template Constructor


MarcSkovMadsen:
I believe it's a natural thing to provide one item or a list items to the Template constructor in the header, sidebar or main. But it is currently not supported

```python
from panel import template
from panel.pane import Markdown

ABOUT = """
# About the Awesome Panel Project
"""

def view() -> template.BaseTemplate:
    """The about view of awesome-panel.org"""
    main = [Markdown(ABOUT, sizing_mode="stretch_width")]
    return template.GoldenTemplate(title="About", main=main)

if __name__.startswith("bokeh"):
    view().servable()
```

raises

```bash
$ 2020-10-25 04:19:49,381 Starting Bokeh server version 2.2.3 (running on Tornado 6.0.4)
2020-10-25 04:19:49,383 User authentication hooks NOT provided (default user enabled)
2020-10-25 04:19:49,386 Bokeh app running at: http://localhost:5006/issue_template
2020-10-25 04:19:49,386 Starting Bokeh server with process id: 2500
2020-10-25 04:21:35,461 Error running application handler <bokeh.application.handlers.script.ScriptHandler object at 0x00000197CD7090C8>: Parameter 'main' value must be an instance of ListLike, not '[Markdown(str, sizing_mode='stretch_width')]'
File "__init__.py", line 1320, in _validate:
(self.name, class_name, val)) Traceback (most recent call last):
  File "C:\repos\private\awesome-panel\.venv\lib\site-packages\bokeh\application\handlers\code_runner.py", line 197, in run
    exec(self._code, module.__dict__)
  File "C:\repos\private\awesome-panel\scripts\issue_template.py", line 37, in <module>
  File "C:\repos\private\awesome-panel\scripts\issue_template.py", line 34, in view
  File "C:\repos\private\awesome-panel\.venv\lib\site-packages\panel\template\base.py", line 409, in __init__
    super(BasicTemplate, self).__init__(template=template, **params)
  File "C:\repos\private\awesome-panel\.venv\lib\site-packages\panel\template\base.py", line 61, in __init__
    super(BaseTemplate, self).__init__(**params)
  File "C:\repos\private\awesome-panel\.venv\lib\site-packages\param\parameterized.py", line 2339, in __init__
    self.param._setup_params(**params)
  File "C:\repos\private\awesome-panel\.venv\lib\site-packages\param\parameterized.py", line 991, in override_initialization
    fn(parameterized_instance,*args,**kw)
[...]

PR #1694, opened 2020-10-25
Enable specifying items on Template construction

MarcSkovMadsen:
This PR implements the FR https://github.com/holoviz/panel/issues/1693.

I.e. it provides support for providing header, sidebar, main and modal item(s) on construction.

The implementation is ready for review @philippjfr . The PR includes the relevant tests. But you can also manually play around with the example below.

## Example

You can manually test via

```python
from panel.template import ReactTemplate, VanillaTemplate
from panel.pane import Markdown

# item1 = Markdown("Hello Header")
# item2 = Markdown("Hello Sidebar")
# item3 = Markdown("Hello Main")

item1 = [Markdown("Hello Header"), Markdown("Hello2")]
item2 = [Markdown("Hello Sidebar"), Markdown("Hello2")]
item3 = [Markdown("Hello Main"), Markdown("Hello2")]
# VanillaTemplate(title="VanillaTemplate", header=item1, sidebar=item2, main=item3).servable()
ReactTemplate(title="ReactTemplate", header=item1, sidebar=item2, main=item3).servable()
```

which looks like

image


panel/template/base.py
```diff
@@ -5,6 +5,7 @@
 from __future__ import absolute_import, division, unicode_literals
 
 import os
+from panel.layout.grid import GridSpec
```
philippjfr, line 8: Imports should be relative and be ordered correctly, i.e. you should use `from ..layout.grid import GridSpec` and then move this import into the list of internal imports.

panel/template/base.py
```diff
@@ -570,6 +579,30 @@ def _get_favicon_type(favicon):
...
+        if class_==GridSpec:
+            grid = GridSpec(ncols=12)
+            for index, item in enumerate(value):
+                grid[index, :]=item
+            return grid
+
+        return value
+
+
+
```
philippjfr, line 605: Too much space.

panel/template/base.py
```diff
@@ -570,6 +579,30 @@ def _get_favicon_type(favicon):
...
+            return value
+        if isinstance(value, tuple):
+            value = [*value]
+        elif not isinstance(value, list):
+            value = [value]
+
+        # Important to fx. convert @param.depends functions
+        value = [_panel(item) for item in value]
+
+        if class_==ListLike:
```
philippjfr, line 594:
```suggestion
        if class_ is ListLike:
```

panel/template/base.py
```diff
@@ -570,6 +579,30 @@ def _get_favicon_type(favicon):
...
+            return ListLike(objects=value)
+        if class_==GridSpec:
```
philippjfr, line 596:
```suggestion
        if class_ is GridSpec:
```

panel/template/base.py
```diff
@@ -570,6 +579,30 @@ def _get_favicon_type(favicon):
...
+            grid = GridSpec(ncols=12)
+            for index, item in enumerate(value):
+                grid[index, :]=item
```
philippjfr, line 599: Suppose this is reasonable, but it is *a bit* weird.

philippjfr:
Mostly nitpicky comments. If you fix it up this by about noon this can be the last commit before a release.

panel/template/base.py
MarcSkovMadsen: This comment I don't understand. What is weird?

If you are referring to the `ncols=12` I've taken that from the ReactTemplate constructor

image

MarcSkovMadsen:
Thanks for the review @philippjfr . You are a man of **style** :-) That is great. 

I've pushed fixes based on your comments. Except the *weird* comment which I do not understand.

As a side note I need to get a good QA flow for working with Panel. My normal flow is different. It's automated based on isort, autoflake8, black, pylint, mypy and pytest in that order. So most things are automated away and I don't have to spend mental band width on it. So I need to learn to do more manual things here.

panel/template/base.py
philippjfr: Iterating over the list and assigning each to a row is okay behavior but it's certainly not the only behavior that makes sense. Anyway I'm fine with it.

philippjfr:
I think we should definitely enforce a few more code quality checks but for now let's release 🥳

Issue #1629, opened 2020-10-13
Widget-binding API


jbednar:
The [API guide](https://panel.holoviz.org/user_guide/APIs.html) lays out four different APIs for using Panel, which one might think is already confusing enough, but even so I'd like to propose one more.

## Parameterized param.depends API

Right now, one can write a Parameterized class using `@depends("val")` to indicate that a certain method should be called when parameter `val` changes:

```python
import param

class C(param.Parameterized):
    val = param.Selector(dict(v1=1, v2=2))
    
    @param.depends("val")
    def output(self):
        return self.val

c=C(name="")

import panel as pn
pn.extension()
pn.Column(c.param, c.output)
```
![image](https://user-images.githubusercontent.com/1695496/95907966-f6e31c80-0d61-11eb-8f54-070e3819e79e.png)

Pros of this approach:
- Concise
- Separates Panel-specific (and thus JS, browser, Jupyter, etc.-specific) code from the domain code (`C`), allowing domain code to be defined in separate modules that don't import or otherwise require Panel

Cons:
- Requires classes, which many Python programmers aren't familiar with
- Requires understanding Parameters
- Widget types are difficult to specify if you don't want the default widget for a given Parameter type

## Widgets pn.depends API

A similar syntax is supported for bare functions, in this case depending on parameters stored on widgets:

```python
import panel as pn
pn.extension()

valw = pn.widgets.Select(options=dict(v1=1, v2=2))

@pn.depends(valw)
def output(val): 
    return val

pn.Column(valw, output)
```
![image](https://user-images.githubusercontent.com/1695496/95908005-0498a200-0d62-11eb-9f05-9534daf5df34.png)

Pros of this approach:
- Concise
- Avoids having to understand classes or Parameters
- Explicit about what type of widget you'll get
- Conceptually related to the Parameterized API, making it simpler to switch between them

Cons:
- Intertwines Panel and domain code, requiring widgets to be instantiated before code with dependencies can be defined
- Confusing for users who want to understand `output` all by itself, but have to look up what `depends` does and understand how it might modify the function (e.g. wondering if it can still be called with an explicit argument for `val`, or has that argument been bound to `valw` permanently?)
- Not clear how the same function could be used in multiple different Panel apps, each with their own choice of widgets, without redefining the function
- If making a Panel for existing code where Panel and domain logic can't be intertwined in this way, requires writing a lot of wrapper functions just to express the binding from widgets to function arguments, then call the real code.

## Widget-binding API

[...]

PR #1702, opened 2020-10-27
Add pn.bind function to bind parameters to a function

philippjfr:
Implements https://github.com/holoviz/panel/issues/1629

### DIFF
--- a/panel/depends.py
+++ b/panel/depends.py
@@ -26,7 +26,78 @@ def param_value_if_widget(arg):
 
 
 def depends(*args, **kwargs):
+    """
+    Annotates a function or Parameterized method to express its
+    dependencies on a set of parameters. This allows libraries
+    such as Panel to watch those dependencies for changes and update
+    the output of the annotated function when one of the dependencies
+    changes.
+
+    The specified dependencies can be either be Parameter instances,
+    widgets, or, if a method is supplied, they can be defined
+    as strings referring to Parameters of the class, or Parameters of
+    subobjects (Parameterized objects that are values of this object's
+    parameters).
+    """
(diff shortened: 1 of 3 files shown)


panel/depends.py
```diff
@@ -26,7 +26,78 @@ def param_value_if_widget(arg):
...
+    dependencies on a set of parameters. This allows libraries
+    such as Panel to watch those dependencies for changes and update
+    the output of the annotated function when one of the dependencies
+    changes.
+
+    The specified dependencies can be either be Parameter instances,
+    widgets, or, if a method is supplied, they can be defined
+    as strings referring to Parameters of the class, or Parameters of
+    subobjects (Parameterized objects that are values of this object's
+    parameters).
```
jbednar, line 40:
```suggestion
    Python decorator annotating a function or Parameterized method to 
    express its dependencies on a set of Parameters. This declaration
    allows libraries like Panel to watch those dependencies for changes
    and invoke the function or method to get an updated output when one
    of the dependencies changes.

    This function is the same as the corresponding param.depends 
    decorator, but extended so that if widgets are provided, the underlying
    value parameter of the widget is extracted as a dependency. This 
    extension is solely for syntactic convenience, allowing the widget to be
    passed in as a synonym for the underlying parameter, leaving this 
    decorator otherwise behaving the same as the underlying Param
    depends decorator. 
    
    For the Panel version of the decorator, the specified dependencies can
    either be Parameter instances, Panel or ipywidgets widgets, or, if a 
    Parameterized method is supplied rather than a function, they can be
    defined as strings referring to Parameters of that class, or Parameter 
    objects of that object's subobjects (i.e., Parameterized objects that are
    values of this object's parameters).
```

PR #1716, opened 2020-10-28
Add seek(0) to pane.image

### LINKED ISSUE #1711: BytesIO/StringIO break pn.pane.image if called second time
Panel: 0.10.0.post7+gb5cf6928

When having a BytesIO/StringIO object as input to pn.pane.image it will crash the second time a widget is called.
I don´t know if it a feature or a bug... I have made a simple implementation of a fix, if it is a bug [here](https://github.com/Hoxbro/panel/commit/af03e001a23d1e1cea065f85abfb4e0176d0d688).

```python
import os

import param
import panel as pn


class ImageExample(param.Parameterized):
    def __init__(self, **params):
        self.button = pn.widgets.Button(name="Create Animation", button_type="primary")
        self.break_ = pn.widgets.Button(name="Break Animation", button_type="primary")
        super().__init__(**params)
        self.gif = None

    @param.depends(pn.state.param.busy)
    def animation(self, busy):
        return None if busy else self.gif

    @param.depends("button.clicks")
    def _create_animation(self):
        # it could be any gif... :)
        self.gif = pn.pane.GIF(open(os.path.join(os.path.dirname(__file__),'index.gif'), 'rb'))

    @param.depends("break_.clicks")
    def _break_animation(self):
        pass


ie = ImageExample()
pn.Column(
    ie.button,
    ie.break_,
    ie.animation,
    ie._create_animation,
).servable()


```

bug

Hoxbro:
Attempt to fix #1711 by forcing attribute with read to set seek(0). 

I can't think of a good reason not to start the read at 0, but if there is this PR can be closed.

### DIFF
--- a/panel/pane/image.py
+++ b/panel/pane/image.py
@@ -78,6 +78,8 @@ class ImageBase(DivPaneBase):
                 with open(self.object, 'rb') as f:
                     return f.read()
         if hasattr(self.object, 'read'):
+            if hasattr(self.object, 'seek'):
+                self.object.seek(0)
             return self.object.read()
         if isurl(self.object, None):
             import requests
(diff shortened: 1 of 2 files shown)


philippjfr:
Looks good to me.

Issue #1722, opened 2020-10-30
Hovering over Plotly plot raises DeserializationError


MarcSkovMadsen:
#### ALL software version info

Panel 0.10.1
Bokek 2.2.3
Plotly 4.8.2 and 4.12.0
Windows 10
Chrome

#### Description of expected behavior and the observed behavior

I hover over Plotly plot. I would not expect exceptions raised but I do.

#### Additional Context

scoodood has also noted this on Discourse https://discourse.holoviz.org/t/this-panel-plotly-example-is-throwing-error/1363 for the Plotly Reference example.

#### Complete, minimal, self-contained example code that reproduces the issue

```python
"""
Scoodood is asking on [Discourse](https://discourse.holoviz.org/t/how-to-capture-the-click-event-on-plotly-plot-with-panel/1360)

How to capture the click event on Plotly plot with Panel?
"""

import numpy as np
from panel.template.react import ReactTemplate
import plotly.graph_objs as go
import panel as pn

pn.extension("plotly")
pn.config.sizing_mode = "stretch_width"


def create_plot():
    t = np.linspace(0, 10, 50)
    x, y, z = np.cos(t), np.sin(t), t
    fig = go.Figure(
        data=go.Scatter3d(x=x, y=y, z=z, mode="markers"), layout=dict(title="3D Scatter Plot")
    )
    fig.layout.autosize = True
    return fig


def create_layout(plot):
    description_panel = pn.layout.Card(
        __doc__, header="# How to capture Plotly Click Events?", sizing_mode="stretch_both"
    )
    plot_panel = pn.pane.Plotly(plot, config={"responsive": True}, sizing_mode="stretch_both")
    settings_panel = plot_panel.controls(jslink=True)

    template = ReactTemplate(title="Awesome Panel - Plotly App")
    template.sidebar.append(settings_panel)
    template.main[0, :] = description_panel
    template.main[1:4, :] = plot_panel
    return template


def create_app():
    plot = create_plot()
    return create_layout(plot)


app = create_app()
app.servable()
```

#### Stack traceback and/or browser JavaScript console output

```bash
$ python -m panel serve 'scripts\discourse_1360_plotly_clickevents.py' --dev
2020-10-30 19:25:22,434 Starting Bokeh server version 2.2.3 (running on Tornado 6.0.4)
2020-10-30 19:25:22,436 User authentication hooks NOT provided (default user enabled)
[...]

PR #1735, opened 2020-11-01
Fix None serialization on LiteralInput

philippjfr:
Fixes https://github.com/holoviz/panel/issues/1722

### DIFF
--- a/panel/widgets/input.py
+++ b/panel/widgets/input.py
@@ -349,13 +349,13 @@ class LiteralInput(Widget):
     def _process_param_change(self, msg):
         msg = super(LiteralInput, self)._process_param_change(msg)
         if 'value' in msg:
-            value = '' if msg['value'] is None else msg['value']
+            value = msg['value']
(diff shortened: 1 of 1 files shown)


panel/widgets/input.py
```diff
@@ -349,13 +349,13 @@ def _process_property_change(self, msg):
     def _process_param_change(self, msg):
         msg = super(LiteralInput, self)._process_param_change(msg)
         if 'value' in msg:
-            value = '' if msg['value'] is None else msg['value']
+            value = msg['value']
             if isinstance(value, string_types):
                 value = repr(value)
             elif self.serializer == 'json':
                 value = json.dumps(value, sort_keys=True)
```
xavArtley, line 356: Don't know if you can deactivate sort_key=True since it seems to generate a bug when a point is hover and unhover quickly:
Bug with sort_key=True
sort_key_true

without sort_key:
sort_key_false

Issue #1255, opened 2020-04-10
Update docs to be consistent on `value_throttled` and `callback_policy`


joelostblom:
From https://discourse.holoviz.org/t/how-to-throttle-values-on-mouseup/456/2

I thought `callback_policy` was deprecated, but if I look at the [documentation of IntRangeSlider](https://panel.holoviz.org/reference/widgets/IntRangeSlider.html), it is still there. However, in the [docs for RangeSlider](https://panel.holoviz.org/reference/widgets/RangeSlider.html), there is no `callback_policy` and instead there is `value_throttled` which is referencing `callback_throttle`, which in turn does not exist on this doc page but it does exist on the IntRangeSlider doc page where `value_throttled` does not exist.

Related to #1253

joelostblom:
This doc update could also include clarifying what `value_throttled` actually does. The current docstring reads

> Tuple of upper and lower bounds of selected range where events are throttled by callback_throttle value.

To me this implied that throttling will only occur in the range specified in the tuple. However, the current behavior of the parameter is to toggle throttling for the entire range of the widget and set the initial value to what is specified in the tuple. And as @philippjfr [pointed out here](https://discourse.holoviz.org/t/how-to-throttle-values-on-mouseup/456/6?u=joelostblom), the throttling means that the widget fires on "mouseup", it seems like the reference to `callback_throttle` is obsolete, is that correct? If so, I would rewrite like this for clarity:

> Tuple of upper and lower bounds for the selected value range. In contrast to setting `value`, events are throttled so that they only update on mouseup rather than continuously. 

---

Personally, I think the most intuitive behavior would be to make `value_throttled` a boolean parameter and rely exclusively on `value` to specify the value range, and not have the user switch between using either `param.value` or `param.value_throttled`, but instead always be using `param.value` in `@pn.depends`. This would avoid confusion stemming from not realizing that you need to change both the initialization and the decoration to throttle a value.

This parameter name could be changed to `throttle_value` to avoid confusion with how they are named in bokeh and under the hood it could set the `param.value` attribute to the bokeh `value_throttled` valule, but the user would not have to change the name of the attribute they are using. So they code would look like:

```python
int_range_slider = pn.widgets.IntRangeSlider(end=10, throttled_value=True)

@pn.depends(int_range_slider.param.value, watch=True)
def print_slider_value(slider_value):
    return slider_value
```

[...]

PR #1739, opened 2020-11-02
Add value throttled to DiscreteSlider

Hoxbro:
Supersede the bug fixes and documentation update in #1400

There has been some changes to the codebase since the previously PR, most notably it seems like #1256 has been fixed. 

So this PR updates the DiscreteSlider with the ability to use value_throttled as an argument #1253. I could not get value_throttled to work with the `constant=True`, which is why it is disabled. The PR also updates unit test for sliders with `value_throttled`.

Lastly the sliders examples/documentation with value_throttled has been updated #1255.


Hoxbro:
@philippjfr do you have any comments or suggestion? And do you have a way to get `value_throttled = param.Parameter(constant=True)` to work if it is needed.

panel/widgets/slider.py
```diff
@@ -175,7 +175,7 @@ class DiscreteSlider(CompositeWidget, _SliderBase):
 
     value = param.Parameter()
 
-    value_throttled = param.Parameter(constant=True)
+    value_throttled = param.Parameter()
```
philippjfr, line 178: Setting these to constant was intentional. Users should generally only set `value` and only use `value_throttled` to watch it. So use the `edit_constant` property to change it internally.

philippjfr:
I had one comment about keeping `value_throttled` as constant, otherwise this looks good to me. Thanks for reviving this work!

matsvandecavey:
@philippjfr With the `value_throttled` as constant, is there a way to change its value programmatically?
My app has a *reset* button, but I can't change a constant value:
```python
myslider.value_throttled = default_value
```
This throws a `TypeError` 
> TypeError: Constant parameter 'value_throttled' cannot be modified

philippjfr:
@matsvandecavey Change the `.value` and `.value_throttled` will update.

matsvandecavey:
Hi Philipp, This is not working for me.

value_throttled

Issue #1742, opened 2020-11-04
Varius Template Style Issues


MarcSkovMadsen:
I plan to contribute PR to fix or improve the below ASAP. Screenshots are from current MASTER branch. 

This Bug Report is created to collect the bugs.

#### Material Template

![image](https://user-images.githubusercontent.com/42288570/98129196-24cb0500-1eb9-11eb-8f60-55811057e186.png)

#### Bootstrap

Does not obey `main_max_width`

![image](https://user-images.githubusercontent.com/42288570/98129333-4d52ff00-1eb9-11eb-9569-e8b36fcc101a.png)

![image](https://user-images.githubusercontent.com/42288570/98130928-167de880-1ebb-11eb-99c7-d46758d37ae7.png)

#### React

![image](https://user-images.githubusercontent.com/42288570/98129614-9c992f80-1eb9-11eb-9938-145a7469c896.png)

#### Vanilla

![image](https://user-images.githubusercontent.com/42288570/98129703-b9cdfe00-1eb9-11eb-8fb2-c083a684b0a0.png)

MarcSkovMadsen:
Upps. The tests above are not fully valid. I at awesome-panel.org I have some css fixes for Panel 0.10. The above is for MASTER.

I will do the PR based on master and the test_manual.py file and only fix issues on that.

philippjfr:
Are these all fixed now? Please reopen if not.

PR #1743, opened 2020-11-04
Template Fixes - Mostly sidebar or main heights

MarcSkovMadsen:
In https://github.com/holoviz/panel/issues/1742 I noted some template issues. This PR fixes the issues I see when using the `test_manual.py` app to test. This is mainly main height fixes. And then the Bootstrap Template did not support the `main_max_width` parameter.

I have checked both in Chrome and Firefox.

I could not fix the Bootstrap sidebar at `300px` so I did not change that.

I will not do more now. I think this is the most important to fix.

### DIFF
--- a/panel/template/bootstrap/bootstrap.css
+++ b/panel/template/bootstrap/bootstrap.css
@@ -12,6 +12,7 @@ body {
 }
 
 #sidebar {
+  padding: 15px;
   transition: all 0.2s cubic-bezier(0.945, 0.020, 0.270, 0.665);
   transform-origin: center left; /* Set the transformed position of sidebar to center left side. */
   overflow-y: scroll;
--- a/panel/template/bootstrap/bootstrap.html
+++ b/panel/template/bootstrap/bootstrap.html
@@ -71,7 +71,7 @@
           {% for doc in docs %}
           {% for root in doc.roots %}
           {% if "main" in root.tags %}
-          {{ embed(root) | indent(4) }}
+          {{ embed(root) | indent(6) }}
           {% endif %}
           {% endfor %}
           {% endfor %}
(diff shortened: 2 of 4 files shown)


panel/template/bootstrap/bootstrap.html
```diff
@@ -65,13 +65,19 @@
...
-      {% for root in doc.roots %}
-      {% if "main" in root.tags %}
-      {{ embed(root) | indent(4) }}
+      {% if main_max_width %}
+        <div style="margin-left: auto; margin-right: auto;max-width: {{main_max_width}}">
+      {% endif %}
+          {% for doc in docs %}
+          {% for root in doc.roots %}
+          {% if "main" in root.tags %}
+          {{ embed(root) | indent(4) }}
```
philippjfr, line 74:
```suggestion
          {{ embed(root) | indent(6) }}
```

panel/template/react/react.css
```diff
@@ -36,7 +36,7 @@ a.navbar-brand {
   text-decoration-style: initial;
   text-decoration-color: initial;
   font-weight: 400;
-  font-size: 1.5em;
+  font-size: 2.0em;
```
philippjfr, line 39: Can you show me what that looks like?

panel/template/bootstrap/bootstrap.css
```diff
@@ -12,6 +12,7 @@ body {
 }
 
 #sidebar {
+  padding: 15px;
```
philippjfr, line 15: Did bootstrap have no padding unlike other templates?

MarcSkovMadsen: Yes

PR #1754, opened 2020-11-09
Add a throttled argument to depends, interact and bind

Hoxbro:
Supersede #1400 

The goal of this PR is to discuss the option to add a argument to `pn.depends` , `pn.interact` and `pn.bind` which throttles the slider widgets as mention in #1259. 

This is a work in progress and await input from @philippjfr. 

For testing purpose I use this script:
``` python

import panel as pn

pn.extension()

slider_dict = {
    "DateSlider": dict(start=date(2018, 9, 1), end=date(2018, 9, 10)),
    "DateRangeSlider": dict(
        start=date(2018, 9, 1),
        end=date(2018, 9, 10),
        value=(date(2018, 9, 2), date(2018, 9, 4)),
    ),
    "DiscreteSlider": dict(
        options=[0.1, 1, 10, 100],
        value=1,
    ),
    "FloatSlider": dict(start=1, end=10, value=5),
    "IntSlider": dict(start=1, end=10, value=5),
    "IntRangeSlider": dict(start=1, end=10, value=(2, 5)),
    "RangeSlider": dict(start=1, end=10, value=(2, 5)),
}

func = lambda x: repr(x)
throttled = True

dashboard = []
for slider, kwargs in slider_dict.items():
    widget = getattr(pn.widgets, slider)(**kwargs)
    bind = pn.bind(func, widget, throttled=throttled)
    depends = pn.depends(widget, throttled=throttled)(func)
    interact = pn.interact(func, x=widget, throttled=throttled)
    slider = pn.Column("## " + slider, widget, bind, depends, interact)
    dashboard.append(slider)

pn.Row(*dashboard).show()
```

Missing tasks:
 - [ ] Discussion if this is the best implementation or the implementation is needed.
 - [ ] Keyword argument to be used to enable value_throttled on sliders, right now it is `throttled`.
 - [ ] Create tests for the chosen implementation.
 - [ ] Update documentation with the new implementation.
 - [ ] Other?

### DIFF
--- a/examples/user_guide/Interact.ipynb
+++ b/examples/user_guide/Interact.ipynb
@@ -393,6 +393,25 @@
    "source": [
     "interact(f, x=dict([('one', 10), ('two', 20)]))"
    ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## Disabling continuous updates for slider widgets\n",
+    "When interacting with functions which takes a long time to run, realtime feedback can be a burden instead of being helpful.\n",
+    "\n",
+    "If you are using slider widgets and you have a function which takes long time to calculate, you can set the keyword argument `throttled` to `True` in `interact`. This will then first run the function after the release of the mouse button."
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "interact(f, x=(0.0, 20.0, None, 5.5), throttled=True)"
+   ]
   }
  ],
  "metadata": {
--- a/panel/depends.py
+++ b/panel/depends.py
@@ -7,12 +7,9 @@ from .widgets import Widget
 ipywidget_classes = {}
 
 
-def param_value_if_widget(arg, throttled=False):
+def param_value_if_widget(arg):
     if isinstance(arg, Widget):
-        if throttled is True and hasattr(arg, 'value_throttled'):
-            return arg.param.value_throttled
-        else:
-            return arg.param.value
+        return arg.param.value
 
     from .pane.ipywidget import IPyWidget
     if IPyWidget.applies(arg) and hasattr(arg, 'value'):
@@ -28,7 +25,7 @@ def param_value_if_widget(arg, throttled=False):
     return arg
 
 
-def depends(*args, throttled=False, **kwargs):
+def depends(*args, **kwargs):
     """
     Python decorator annotating a function or Parameterized method to
     express its dependencies on a set of Parameters. This declaration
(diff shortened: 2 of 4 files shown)


philippjfr:
I'm a -1 on the changes to bind and depends since you can very easily use `widget.param.value_throttled`.

philippjfr:
So what I'd like to see here is removing it from the depends/bind code, adding an example to the `interact` docs and finally a test.

philippjfr:
Also happy to see an example showing depends/bind usage with `value_throttled`, don't have any suggestions on where that should go though.

Hoxbro:
Thanks for the comments, I will do what you have suggested later this week. 

I would just like to mention that I have had a lot of problem finding out the correct way to apply `value_throttled` argument in depends/interact, when I started learning Panel and I still find myself look at old code to get it to work. I can see I'm properly not the only one which have these problem if I look at the discourse [1416](https://discourse.holoviz.org/t/how-to-use-callback-policy-intrange-slider/1416) and [1404](https://discourse.holoviz.org/t/how-to-disable-continuous-update-on-a-slider-widget/1404), so maybe a independent documentation page with examples of `value_throttled` could be a good option. This page could then be referenced at the slider widgets page.

MarcSkovMadsen:
@Hoxbro. Why do you use `value_throtled` at all and what does it do? (I've never used that.)

xavArtley:
Value throttle is updated only when you end interaction with the widget.
For exemple slider send value_throttle only when you release mouse button.
It avoid to call too often some functions wich can be time consuming

Le ven. 13 nov. 2020 à 11:51, Marc Skov Madsen <notifications@github.com> a
écrit :

Hoxbro:
As @xavArtley explained it is for calling a function after the mouse is released from the slider and not continually doing the sliding. When the function takes a long time to calculate it becomes very frustrating because it feels like it takes even longer than it should, see example below.  Also if the function has to open a file or look up values from a database it will do it a unnecessary amount of times.

``` python
import panel as pn
import time

pn.extension()

def func(x):
    time.sleep(1)
    return repr(x)

widget1 = pn.widgets.IntSlider(start=1, end=100)
widget2 = pn.widgets.IntSlider(start=1, end=100)

slider1 = pn.bind(func, widget1)
slider2 = pn.bind(func, widget2.param.value_throttled)

pn.Column(widget1, slider1, widget2, slider2).servable()
```
ezgif-4-e1a154739680

Right now the only options is to add the `widget.param.value_throttled` as an input instead of just `widget`. When you have a lot of sliders it becomes more confusing and therefore it would have been nice if there were an option to just use `value_throttled` instead of `value` if a widget had it. 

Another problem with `value_throttled` (to my knowledge) is that it is hard to incorporate into a parametrized class, what I do is either what I showed [here](https://discourse.holoviz.org/t/how-to-use-callback-policy-intrange-slider/1416/4) or just create a "simple" widget.

philippjfr:
Thanks so much! I'll merge this for now, and further improvements for the docs can come later.

philippjfr:
The Parameterized class problem is a good point, I think we should allow configuring that in the Param pane `widgets` argument. Also for a later PR though.

Issue #1715, opened 2020-10-28
Redirect URL goes to http not https


konradsitarz:
Hi, I'm trying to deploy Panel dashboard as azure app service with AAD authorization. I have a problem with the redirect url which is being changed from **https**://myapp.azurewebsites.net changed to **http**://myapp.azurewebsites.net in the redirect URI parameter when trying to log in with azure active directory.

So the url looks like
https://login.microsoftonline.com/common/oauth2/authorize?redirect_uri=http%3A%2F%myapp.azurewebsites.net&client_id=myid&response_type=code&state=mystate

and I suppose it should be 

https://login.microsoftonline.com/common/oauth2/authorize?redirect_uri=https%3A%2F%myapp.azurewebsites.net&client_id=myid&response_type=code&state=mystate

Panel version: 0.10.0

Error from MS page: AADSTS50011: The reply URL specified in the request does not match the reply URLs configured for the application

philippjfr:
The redirect URI is determined by the protocol you are using when you visit the app or the explicit config value you've configured:

```python
        if config.oauth_redirect_uri:
            redirect_uri = config.oauth_redirect_uri
        else:
            redirect_uri = "{0}://{1}".format(
                self.request.protocol,
                self.request.host
            )
```

So this is entirely controlled by what you do.

konradsitarz:
Ok, I found that it can be changed with environmental variable PANEL_OAUTH_REDIRECT_URI.
I think it could be added to the documentation because It's not there I found it directly in the code.

philippjfr:
I'm still not seeing what is wrong with the default of:

```python
redirect_uri = "{0}://{1}".format(
                self.request.protocol,
                self.request.host
            )
```

maximlt:
Hey @antallnguyen, could you please open a new issue? The maintainers are working hard on preparing Panel 1.0 and your message here could easily get lost.

antallnguyen:
> Hey @antallnguyen, could you please open a new issue? The maintainers are working hard on preparing Panel 1.0 and your message here could easily get lost.

Turns out it was an error on my part.. the URL format I did was incorrect. Should have been "http://foo.bar" instead of "http://foo.bar/"

philippjfr:
That's an easy mistake to make, a better error message would be good.

el-abcd:
Hi,

[...]

PR #1759, opened 2020-11-10
Document OAuth redirect URI

philippjfr:
Closes https://github.com/holoviz/panel/issues/1715

### DIFF
--- a/examples/user_guide/Authentication.ipynb
+++ b/examples/user_guide/Authentication.ipynb
@@ -103,6 +103,18 @@
     "panel serve oauth_example.py --oauth-encryption-key=...\n",
     "\n",
     "PANEL_OAUTH_ENCRYPTION=... panel serve oauth_example.py ...\n",
+    "```\n",
+    "\n",
+    "### Redirect URI\n",
+    "\n",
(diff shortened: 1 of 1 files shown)


el-abcd:
Hi,

I am a fairly new panel user, and am trying to deploy an app on Google Cloud Run (using google oauth for authentication) and also saw the app using http:// in the redirect URI (even though I used https:// when accessing the app).  

I was able to workaround this using the --oauth-redirect-uri="https://..." parameter in the dockerfile, but wanted to share what I **think** is happening...

I suspect (from https://stackoverflow.com/a/67445611, etc) that google cloud run is using https from my browser to a GFE (Google Front End).  But, then from the GFE to the cloud run container I think it is using plain http (perhaps to avoid the extra https work in the container to keep them light-weight?).  
So, the panel/tornado app might be seeing a plain http:// request, even though the user uses https:// in their browser.  

Then, the google oauth server is unhappy when it see's this mismatch in the url.  

At first I worked around it by adding the http:// url to the google oauth config.  This DID work, but it had a warning that this would prevent me from moving the app out of TESTING state.  i.e. they want to see https:// in the uri, per https://developers.google.com/identity/protocols/oauth2/web-server#uri-validation

The "--oauth-redirect-uri=" CLI parameter in the dockerfile DID work (as mentioned), and is what I'm using now.  

I'm guessing that perhaps if I edited the "--port 8080" in the dockerfile to "--port 443" that it might cause it to send https requests to the docker container?  But I didn't try it and could be wrong, maybe the GFE would send http still. 

Anyways, I just wanted to note what might be going on behind the scenes, and why this might be happening (at least for users on google cloud run that try to enable oauth).   If other users are also seeing this, perhaps a note in the docs would be useful.  

Thanks for this project/tool! 

Regards,
Eric

PR #1760, opened 2020-11-10
Add docs and functionality to work in various notebook environments

philippjfr:
@jbednar Please review

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,7 +52,6 @@ If used in JupyterLab, Panel requires the pyviz labextension to be installed:
 jupyter labextension install @pyviz/jupyterlab_pyviz
 ```
 
-
 ## Usage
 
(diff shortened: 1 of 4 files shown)


jbednar (approved):
Looks great, and very exciting! I've pushed various updates directly to the branch.

Can you make a statement about other notebook technologies not yet mentioned, e.g. nteract, zeppelin, etc. -- should users typically expect things to work, expect they can be made to work with some effort, expect them not to be supported without miracles, or..?

philippjfr:
I'd expect nteract to work using `pn.extension(comms='ipywidgets')` or even ``pn.extension(comms='vscode')` but haven't tested it. No idea about zepellin or anything else. Basically depends on the extent to which they either support classic notebook comms or ipywidgets.

PR #1763, opened 2020-11-10
Updated environments docs

jbednar:
- The "Command line" environment previously listed for Panel actually only covers server usage, so this PR changes it to say "Editor + Server" instead.
- What about exporting static HTML? Seems like that ought to be in the list, but I don't know how to tell Panel users how to do it. I myself only do it using nbconvert, but presumably there's a more direct way not involving Jupyter?
- I also put in a section about what to do if your desired environment is not on this list. That may be too much detail, so please pare it down if you think it should say less.
- Presumably these changes, once finalized, should also be reflected into the readme.

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,11 +54,11 @@ jupyter labextension install @pyviz/jupyterlab_pyviz
 
 ## Usage
 
-Panel can be used in a wide range of environments:
+Panel can be used in a wide range of development environments:
 
-### Command line
+### Editor + Server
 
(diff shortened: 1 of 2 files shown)


philippjfr:
The edits look good so please replicate them in the readme.

>What about exporting static HTML? Seems like that ought to be in the list, but I don't know how to tell Panel users how to do it. I myself only do it using nbconvert, but presumably there's a more direct way not involving Jupyter?

This isn't really in the same category as the others and I would not mention it here.

Issue #1776, opened 2020-11-12
Should panel_obj.save have `progress` keyword?


ahuang11:
Since embed has progress, and you can set embed=True in save, should save also has progress?

philippjfr:
Yes, it should.

PR #1778, opened 2020-11-13
Add progress=True

ahuang11:
https://github.com/holoviz/panel/issues/1776

```
import hvplot.xarray
import xarray as xr
import panel as pn
pn.extension()

ds = xr.tutorial.open_dataset('air_temperature')
pane = pn.panel(ds.hvplot('lon', 'lat'))
pane.save('test.html', embed=True, progress=True)
```
image

### DIFF
--- a/panel/viewable.py
+++ b/panel/viewable.py
@@ -713,10 +713,10 @@ class Viewable(Renderable, Layoutable, ServableMixin):
            The path to save json files to
         load_path: str (default=None)
            The path or URL the json files will be loaded from.
-        embed_states: dict (default={})
-          A dictionary specifying the widget values to embed for each widget
         progress: boolean (default=True)
           Whether to report progress
+        embed_states: dict (default={})
+          A dictionary specifying the widget values to embed for each widget
         """
         return save(self, filename, title, resources, template,
                     template_variables, embed, max_states, max_opts,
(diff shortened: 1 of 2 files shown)


panel/viewable.py
```diff
@@ -715,11 +715,13 @@ def save(self, filename, title=None, resources=None, template=None,
...
         embed_states: dict (default={})
           A dictionary specifying the widget values to embed for each widget
+        progress: boolean (default=True)
+          Whether to report progress
         """
         return save(self, filename, title, resources, template,
                     template_variables, embed, max_states, max_opts,
                     embed_json, json_prefix, save_path, load_path,
-                    embed_states)
+                    progress, embed_states)
```
philippjfr, line 724: Looks like this was broken before. Really should have used keywords here.

panel/viewable.py
```diff
@@ -715,11 +715,13 @@ def save(self, filename, title=None, resources=None, template=None,
            The path or URL the json files will be loaded from.
         embed_states: dict (default={})
           A dictionary specifying the widget values to embed for each widget
+        progress: boolean (default=True)
+          Whether to report progress
```
philippjfr, line 719:
```suggestion
        progress: boolean (default=True)
          Whether to report progress
        embed_states: dict (default={})
          A dictionary specifying the widget values to embed for each widget
```

philippjfr:
Thanks @ahuang11!

Issue #1770, opened 2020-11-11
Plotly plot in Plotly Pane inside Card does not work well.


MarcSkovMadsen:
Panel 0.10.1

I'm trying to create a demo streaming app with the different plotting libraries supported by Panel. I can see that the Plotly plot does not (re-)size well inside a card. And furthermore does not work responsively in the ReactTemplate.

![plotly_card](https://user-images.githubusercontent.com/42288570/98850094-87da1000-2454-11eb-9c74-be10e8e5914c.gif)

#### Minimum reproducible example

![image](https://user-images.githubusercontent.com/42288570/98849777-071b1400-2454-11eb-95ef-021e9a6f9648.png)

![image](https://user-images.githubusercontent.com/42288570/98849825-1b5f1100-2454-11eb-9e34-7755e199cba0.png)

![image](https://user-images.githubusercontent.com/42288570/98849878-303ba480-2454-11eb-8488-b534e80c80b2.png)

```python
import numpy as np
import altair as alt
import pandas as pd
import panel as pn
from streamz.dataframe import DataFrame as sDataFrame
import hvplot.pandas
import plotly.express as px

plotly_pane = pn.pane.Plotly(sizing_mode="stretch_both", config={'responsive': True})

def update_plotly(data):
    data = pd.concat(data).reset_index()
    plot = px.line(data, y="y", title='Plotly', template="plotly_dark")
    plot.layout.autosize=True
    plotly_pane.object = plot

df = sDataFrame(example=pd.DataFrame({"y": []}, index=pd.DatetimeIndex([])))
window_stream = df.cumsum().stream.sliding_window(50)

plotly_sink = window_stream.sink(update_plotly)

def emit():
    df.emit(pd.DataFrame({"y": [np.random.randn()]}, index=pd.DatetimeIndex([pd.datetime.now()])))
emit()
pn.state.add_periodic_callback(emit, period=100, count=500)

layout = pn.template.ReactTemplate(
    site="Awesome Panel",
    title="Streaming w. Echarts",
    theme=pn.template.react.DarkTheme,
    row_height=200,
)
layout.main[0:2, 0:6]=pn.layout.Card(plotly_pane, header="PLOTLY", sizing_mode="stretch_both")
layout.servable()
```

xavArtley:
From what I see it's more a plotly problem. It seems like plotly is not height responsive
https://plotly.com/javascript/responsive-fluid-layout/
[...]

PR #1780, opened 2020-11-13
Plotly reactive height

xavArtley:
should fix https://github.com/holoviz/panel/issues/1770
Based on @jonmmease comment I wrapped ploty plot inside a div with height and width set to 100%
I had to call PlotlyPlots.resize on the plotly plot creation for the first resize to work
```python
import pandas as pd
import plotly.express as px

data = pd.DataFrame([
    ('Monday', 7), ('Tuesday', 4), ('Wednesday', 9), ('Thursday', 4),
    ('Friday', 4), ('Saturday', 4), ('Sunay', 4)], columns=['Day', 'Orders']
)

fig = px.line(data, x="Day", y="Orders")
fig.update_traces(mode="lines+markers", marker=dict(size=10), line=dict(width=4))
fig.layout.autosize = True

responsive = pn.pane.Plotly(fig, config={'responsive': True})

pn.Column('# A responsive plot', responsive, sizing_mode='stretch_both').show()
```
plotly_reactive
plotly_reactive

### DIFF
--- a/panel/models/plotly.ts
+++ b/panel/models/plotly.ts
@@ -1,4 +1,6 @@
 import * as p from "@bokehjs/core/properties"
+
+import {div} from "@bokehjs/core/dom"
 import {clone} from "@bokehjs/core/util/object"
 import {isEqual} from "@bokehjs/core/util/eq"
 import {HTMLBox} from "@bokehjs/models/layouts/html_box"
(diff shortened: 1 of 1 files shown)


MarcSkovMadsen:
Thanks @xavArtley . I believe this is very important for Panel as Plotly is a great and very popular plotting library that users would expect to work.

MarcSkovMadsen:
@xavArtley . Do you understand why the card width changes When clicked?

None of the other plotting libraries causes this.

xavArtley:
> @xavArtley . Do you understand why the card width changes When clicked?
> 
> None of the other plotting libraries causes this.

I don't know but the behavior is the same as:
`pn.layout.Card('1'*100, header='Plotly', sizing_mode="stretch_both")`

philippjfr:
This is great, thanks!

Issue #1784, opened 2020-11-13
Location sync broken for dates, ints


johnzeringue:
#### ALL software version info

```
Python 3.9.0

bokeh==2.2.3
notebook==5.7.9
panel==0.10.1

macOS Catalina 10.15.7

Chrome 85.0.4183.121
```

##### Expected

When syncing params with the browser location, refreshing the page or sharing the URL should restore those param values.

##### Observed

Refreshing the page or sharing the URL does not restore param values for dates and ints. It does seem to work for strings and floats.

Editing the URL by changing `?bday="2000-01-01"` to `?bday=2000-01-01` lets you share the URL (though the sync on page load immediately changes the URL back).

#### Complete, minimal, self-contained example code that reproduces the issue

```python
import panel as pn
import panel.widgets as pnw

bday = pnw.DatePicker()
pn.state.location.sync(bday, {"value": "bday"})

@pn.depends(bday)
def summary(bday):
    return f"My birthday is {bday}"

app = pn.Column(bday, summary)
app.servable()
```

#### Stack traceback and/or browser JavaScript console output

```
2020-11-13 16:12:06,453 Exception in callback functools.partial(<bound method IOLoop._discard_future_result of <tornado.platform.asyncio.AsyncIOMainLoop object at 0x142be9c10>>, <Task finished name='Task-369' coro=<_needs_document_lock.<locals>._needs_document_lock_wrapper() done, defined at /Users/johnzeringue/.pyenv/versions/3.9.0/envs/pydata-panel/lib/python3.9/site-packages/bokeh/server/session.py:51> exception=ValueError("CalendarDate 'value' only takes datetime types.")>)
Traceback (most recent call last):
  File "/Users/johnzeringue/.pyenv/versions/3.9.0/envs/pydata-panel/lib/python3.9/site-packages/tornado/ioloop.py", line 741, in _run_callback
    ret = callback()
  File "/Users/johnzeringue/.pyenv/versions/3.9.0/envs/pydata-panel/lib/python3.9/site-packages/tornado/ioloop.py", line 765, in _discard_future_result
    future.result()
  File "/Users/johnzeringue/.pyenv/versions/3.9.0/envs/pydata-panel/lib/python3.9/site-packages/bokeh/server/session.py", line 71, in _needs_document_lock_wrapper
    result = await result
[...]

PR #1788, opened 2020-11-15
Fix location sync

Hoxbro:
This is a fix to the location sync mentioned in #1784. 

I have not done this for any other widgets than `DatePicker`, because I don´t know if there is a smarter way to do it, if there is let me know. Also need to create unit test for it. 

animated


philippjfr:
The correct way to do this will be to get this into the param serialization framework.

philippjfr:
In fact this is already correctly implemented in param, so not quite sure why the deserialize call shouldn't work.

PR #1789, opened 2020-11-15
Fixes image not scaling dynamically 

### LINKED ISSUE #1777: Image not scaling dynamically

```
Python 3.9.0

bokeh==2.2.3
notebook==5.7.9
panel==0.10.1

macOS Catalina 10.15.7

Chrome 85.0.4183.121
```

##### Expected

I should be able to scale image up and down dynamically in Jupyter Notebook and using the standalone server.

##### Observed

In the notebook, I'm able to scale up and down <= 300 width. I can't make the image larger than 300 pixels wide.

Using the standalone server, it looks like it scales just once (either up or down) and then gets stuck.

```python
import panel as pn

def panel_logo(width=300):
    # also happens with .jpg
    return pn.panel("https://panel.holoviz.org/_static/logo_stacked.png", width=width)

pn.interact(panel_logo)
```

Hoxbro:
Fixes #1777

animated

### DIFF
--- a/panel/pane/image.py
+++ b/panel/pane/image.py
@@ -42,7 +42,7 @@ class ImageBase(DivPaneBase):
 
     imgtype = 'None'
 
-    _rerender_params = ['alt_text', 'link_url', 'embed', 'object', 'style']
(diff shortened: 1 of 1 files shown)


philippjfr:
Thanks @Hoxbro, appreciate it!

PR #1799, opened 2020-11-21
added harware_automation.ipynb streaming example

kaseyrussell:
Here's a streaming example (as suggested by @MarcSkovMadsen) from my post on [discourse.holoviz](https://discourse.holoviz.org/t/how-to-gui-for-hardware-automation-scientific-instrument-control/1445/3). Please let me know what you think.

### DIFF
--- a/examples/gallery/streaming/hardware_automation.ipynb
+++ b/examples/gallery/streaming/hardware_automation.ipynb
@@ -1,0 +1,210 @@
+{
+ "cells": [
+  {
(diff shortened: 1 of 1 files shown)


philippjfr:
Thanks @kaseyrussell, looks great! I'll merge and may make some small edits later. I think adding a streaming section is a good idea but we'll want to add more examples so this one doesn't look quite so lonely :)

PR #1800, opened 2020-11-22
Add throttled argument to pn.param widgets parameter

Hoxbro:
The reason for this PR it is hard to incorporate `value_throttled` for sliders into a parametrized class. 
This will add the possibility to add a `throttled` argument to pn.param `widgets` argument. So the following is possible:

``` python
import param
import panel as pn


class A(param.Parameterized):
    value_ = param.Range(
        default=(100, 250),
        bounds=(0, 250),
    )
    
    value_throttled_ = param.Range(
        default=(100, 250),
        bounds=(0, 250),
    )
  
    def __init__(self, **params):
        super().__init__(**params)
        widgets = {
            "value_": pn.widgets.IntRangeSlider,
            "value_throttled_": {
                "type": pn.widgets.IntRangeSlider,
                "throttled": True
            },
        }
        self.controls = pn.Param(self, widgets=widgets)

    @param.depends("controls")
    def calculation(self):
        return self.value_, self.value_throttled_


a = A()
pn.Column(a.controls, a.calculation)
```

screen

To get this to work I'm changing the param object from linking with `value` to linking with `value_throttled` of the slider and vice versa. This is only done if `throttled` is enabled and the widget contains the attribute `value_throttled`. 

This is the reason why the unit test without `throttled` enabled is on `value` and with `throttled` enabled is on `value_throttled`.

I hope that the configuration of `throttled` is done as mentioned [here](https://github.com/holoviz/panel/pull/1754#issuecomment-729053796). 

If there is a better way to implement this, this PR can be closed. 

TODO:
- [ ] I need some guidance if I need to add more to the unit test. 
- [x] Add some documentation for this feature. I think it properly should be located [here](https://panel.holoviz.org/reference/panes/Param.html#panes-gallery-param).

### DIFF
--- a/examples/reference/panes/Param.ipynb
+++ b/examples/reference/panes/Param.ipynb
@@ -47,6 +47,7 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
+    "## Model building\n",
     "Let's build a model of a cycling Athlete and her PowerCurve. \n",
     "\n",
     "The PowerCurve is a recording of her maximum power output in Watt per kg for fixed durations of time."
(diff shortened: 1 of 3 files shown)


philippjfr:
This looks fabulous, thanks for your efforts around this!

PR #1837, opened 2020-12-14
Fix numeric widget

### LINKED ISSUE #1836: Numeric Param Widgets, initialised with None, displayed as active "0"
Panel 10.1

Widgets or parameters initialised as "None" has value of 'None', but still display an active '0' 

```
class Test(param.Parameterized) :
    w_none_but_0 = param.Number(default=None, allow_None=True)
    w_none_forced = param.Number(default=None, allow_None=True)

    
    # Since default None, was not displayed correctly, tried to override with None,
    # to clear the widget, but had no effect..
    def __init__(self, **params):
        super().__init__(**params)
        self.param.set_param("w_none_forced", 450) 
        self.param.set_param("w_none_forced", None)    # didn't improve anything
          
    
t = Test()
pn.Row(t)

```

image

Hoxbro:
Fixes #1836

 I'm a bit uncertain if the change in `panel/param.py` is okay, since I don't know if every widgets accepts None as a value. If there is a better way to implement this, this PR can be closed.

Example of fix:
``` python
import param
import panel as pn


class Test(param.Parameterized) :
    number = param.Number(default=0, allow_None=True)
    none = param.Number(default=None, allow_None=True)

    def view(self):
        return self.number, self.none


t = Test()
pn.Column(t, t.view).servable()
```

### DIFF
--- a/panel/widgets/tables.py
+++ b/panel/widgets/tables.py
@@ -633,7 +633,7 @@ class DataFrame(BaseTable):
     hierarchical = param.Boolean(default=False, constant=True, doc="""
         Whether to generate a hierachical index.""")
 
-    fit_columns = param.Boolean(default=None, doc="""
+    fit_columns = param.Boolean(default=True, doc="""
         Whether columns should expand to the available width. This
         results in no horizontal scrollbar showing up, but data can
         get unreadable if there is no enough space available.""")
--- a/panel/param.py
+++ b/panel/param.py
@@ -359,7 +359,7 @@ class Param(PaneBase):
         kw = dict(disabled=p_obj.constant, name=label)
 
         value = getattr(self.object, p_name)
-        if value is not None:
+        if value is not None or getattr(p_obj, 'allow_None', False):
             kw['value'] = value
 
         # Update kwargs
--- a/panel/tests/widgets/test_input.py
+++ b/panel/tests/widgets/test_input.py
@@ -3,6 +3,8 @@ from __future__ import absolute_import, division, unicode_literals
 import pytest
 from datetime import datetime, date
 
+import param
+import panel as pn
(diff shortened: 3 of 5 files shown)


panel/widgets/tables.py
```diff
@@ -633,7 +633,7 @@ class DataFrame(BaseTable):
     hierarchical = param.Boolean(default=False, constant=True, doc="""
         Whether to generate a hierachical index.""")
 
-    fit_columns = param.Boolean(default=None, doc="""
```
philippjfr, line 636: This seems unrelated.

philippjfr:
I didn't think you could pass None to a slider. What happens?

panel/widgets/tables.py
Hoxbro: The changes i made gave two type of errors, though both related to `controls(jslink=True)`.  One of the error was related to `pn.indicators.Progress` and the other `pn.widgets.DataFrame`. The one for DataFrame can be reproduced by:
```python
df_widget = pn.widgets.DataFrame()
df_widget.controls(jslink=True)
# ValueError: Boolean 'value' only takes a Boolean value.
```

I could see in the [documentation](https://panel.holoviz.org/reference/widgets/DataFrame.html) that it should be True by default.

Hoxbro:
It works but it is in no way perfect.
example

The reason for the changes to the sliders is that `pn.indicators.Progress` `value` need to be able to give a non integer value. I get an error by running the following without setting `allow_None=True`. 

```
indeterminate = pn.indicators.Progress()
indeterminate.controls(jslink=True)
# ValueError: Parameter 'value' must be an integer.
```

If there is a better way to get it working, I would definitely like to here it.

philippjfr:
Revived this in #3174.

PR #1859, opened 2020-12-19
Bug fix: make max as upper bound for value

hyamanieu:
Currently defining `max` when creating the instance does not change the upper `value` bound. Which leads to a ValueError.
Simply calling the method `_update_value_bounds` in `__init__` fixes the problem.


philippjfr:
Much appreciated!

PR #1869, opened 2020-12-21
Allow using asynchronous callbacks

philippjfr:
```python
import asyncio
import panel as pn

button = pn.widgets.Button(name='Click to launch')
text = pn.widgets.StaticText()

async def on_click(event):
    text.value = f'Running {event.new}' # Indicate task is running
    await asyncio.sleep(wait) # Simulate long running IO task
    text.value = f'Finished {event.new}!' # Indicate task finished

button.on_click(on_click)

pn.Row(button, text).servable()
```

async

Depends on https://github.com/holoviz/param/pull/449 and Python 3.8+

- [x] Add tests
- [x] Add docs

### DIFF
--- a/doc/user_guide/index.rst
+++ b/doc/user_guide/index.rst
@@ -57,6 +57,9 @@ when needed.
 `Server Deployment <Server_Deployment.html>`_
  Step-by-step guides for deploying Panel apps locally, on a web server or on common cloud providers.
 
+`Asynchronous and Concurrent Processing <Async_and_Concurrency.html>`_
+ Learn how leverage asynchronous and concurrent processing to make your app more responsive.
+
 `Authentication <Authentication.html>`_
  Learn how to add an authentication component in front of your application.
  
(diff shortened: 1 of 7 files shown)


MarcSkovMadsen:
Gamechanger 👍

MarcSkovMadsen:
Can i safely update update pane objects and widget values in the async callback?

philippjfr:
Yes, you can. Only when you update bokeh models directly do you have to decorate with `pn.io.with_lock`.

MarcSkovMadsen:
Nice start to learn about async

philippjfr:
Note I'll have to release Param before this works.

MarcSkovMadsen:
I would really like to see an example where you

- Load data from external database/ service (async use case)
- Transform data using pandas (not async use case. But we would still like to understand how to do that efficiently, in background, using threads or multiple processes etc.)
- Load data to external database/ service (async use case)
- Update `object` of pane or `value of widget. Just to show it is ok to do.

Issue #1874, opened 2020-12-29
Echart pane doesn't fully work with pyechart


jbogaardt:
Echart pane doesn't behave correctly when wrapping a pyechart object + interacting wigh widgets.  Refer to this [discourse discussion](https://discourse.holoviz.org/t/pyecharts-working-example/1590).

#### Description of expected behavior and the observed behavior
I expect this to work, and it initializes, but upon update it returns an error.
```python
from pyecharts.charts import Bar
import panel as pn
pn.extension('echarts')


bar1 = pn.widgets.IntSlider(start=1, end=100, value=50)
bar2 = pn.widgets.IntSlider(start=1, end=100, value=50)

@pn.depends(bar1.param.value, bar2.param.value)
def plot(bar1, bar2):
    my_plot= (Bar()
        .add_xaxis(['Bar1', 'Bar2'])
        .add_yaxis('Values', [bar1, bar2])
    )
    return pn.pane.ECharts(my_plot, width=500, height=250)
pn.Row(pn.Column(bar1, bar2), plot)
```

#### Stack traceback and/or browser JavaScript console output
ValueError: expected an element of Dict(String, Any), got <pyecharts.charts.basic_charts.bar.Bar object at 0x7ffada0b5910>

PR #1875, opened 2020-12-29
ECharts depends on sliders fix

MarcSkovMadsen:
Fixes https://github.com/holoviz/panel/issues/1874

### DIFF
--- a/panel/pane/echarts.py
+++ b/panel/pane/echarts.py
@@ -11,11 +11,11 @@ from ..viewable import Layoutable
 from .base import PaneBase
 
 
-
 class ECharts(PaneBase):
     """
     ECharts panes allow rendering echarts.js plots.
     """
+
     object = param.Parameter(default=None, doc="""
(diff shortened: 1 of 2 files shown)


MarcSkovMadsen:
# Test 1

This works because a `ValueError` is no longer raised.

pyecharts1

```python
from pyecharts.charts import Bar
import panel as pn
pn.extension('echarts')


bar1 = pn.widgets.IntSlider(start=1, end=100, value=50)
bar2 = pn.widgets.IntSlider(start=1, end=100, value=50)

@pn.depends(bar1.param.value, bar2.param.value)
def plot(bar1, bar2):
    my_plot= (Bar()
        .add_xaxis(['Bar1', 'Bar2'])
        .add_yaxis('Values', [bar1, bar2])
    )
    return pn.pane.ECharts(my_plot, width=500, height=250)
pn.Row(pn.Column(bar1, bar2), plot).servable()
```

philippjfr:
Thanks @MarcSkovMadsen!

PR #1878, opened 2020-12-30
TextToSpeech Widget

MarcSkovMadsen:
I think this could be a nice addition to Panel and provide some nice use cases. I hope you (@philippjfr ) will think so as well. 

It might not be within "Data Viz" but for exploring and testing out ideas and working with machine learning or just developing small tools it can be really nice. And for me Panel is about data, models and analytics including AI and DL.

It is READY for review. I can't find any more bugs or things to improve for my use cases.

You can checkout how it works in the video on discourse https://discourse.holoviz.org/t/speech-synthesis/1631

[image](https://discourse.holoviz.org/t/speech-synthesis/1631)


philippjfr:
Would be happy to merge this, but let's call it `TextToSpeech` or something similar.

MarcSkovMadsen:
I like the TextToSpeech name. Have been thinking the same. The reason why I did not do it is that 1) the HTML version is called speechSynthesis and 2) there is also SpeechSynthesisUtterance and SpeechSynthesisVoice objects defined and used.

Would you you rename those additional objects to Utterance and Voice as well? I like that for simplicity and specificity. But we loose the reference to the original objects. Like renaming Tabulator to Table. 🙂 

@philippjfr

MarcSkovMadsen:
I've renamed to TextToSpeech, Utterance and Voice respectively.

MarcSkovMadsen:
Note: I'm currently refactoring as requested. This includes changing the `speak` function to a `value` parameter. This has the consequence that the same text can no longer be spoken twice in a row. I have a feeling that for a lot of testing/ learning this is actually not what you want. 

So I believe I will also add a speak action.

MarcSkovMadsen:
I believe I have implemented review feedback. Next step is on you @philippjfr . Let me know if there is more I should do. Thx.

philippjfr:
A few more comments:

- Please follow the exact template of other reference notebooks, i.e. no titles, keep imports at the top, keep the heading levels the same
- Do not leave pylint annotations in the code, Panel uses flake8
- Keep spaces between assignments (i.e. `a = b` not `a=b`) and equalities (i.e. `a == b` not `a==b`)
- Do not add brackets for single line blocks in JS:

```javascript
if (a==b)
  doSomething()
```

not 

```javascript
if (a==b) {
  doSomething()
}
```

- Do not use semi-colons unless necessary
- Do not use `this_ = this` patterns, instead use arrow functions (`() => { }`) wherever possible.

Issue #1821, opened 2020-11-30
Template assets not loaded when running panel application via jupyter-server-proxy


AjayThorve:
#### Description of expected behavior and the observed behavior
Running the react template app example, but via jupyter-server-proxy fails to load assets (css and js) resulting in a black screen:

![image](https://user-images.githubusercontent.com/20476096/100673466-5c8e6680-3318-11eb-8919-0abc903ac2ba.png)

#### Complete, minimal, self-contained example code that reproduces the issue
```python
import panel as pn
import numpy as np
import holoviews as hv
from panel.template import ReactTemplate

pn.extension()
react = pn.template.ReactTemplate(title='React Template')

pn.config.sizing_mode = 'stretch_both'

xs = np.linspace(0, np.pi)
freq = pn.widgets.FloatSlider(name="Frequency", start=0, end=10, value=2)
phase = pn.widgets.FloatSlider(name="Phase", start=0, end=np.pi)

@pn.depends(freq=freq, phase=phase)
def sine(freq, phase):
    return hv.Curve((xs, np.sin(xs*freq+phase))).opts(
        responsive=True, min_height=400)

@pn.depends(freq=freq, phase=phase)
def cosine(freq, phase):
    return hv.Curve((xs, np.cos(xs*freq+phase))).opts(
        responsive=True, min_height=400)

react.sidebar.append(freq)
react.sidebar.append(phase)

# Unlike other templates the `ReactTemplate.main` area acts like a GridSpec 
react.main[:4, :6] = pn.Card(hv.DynamicMap(sine), title='Sine')
react.main[:4, 6:] = pn.Card(hv.DynamicMap(cosine), title='Cosine')
react.show(open=False)
```

visit url `localhost:8888/proxy/{port}`

#### Stack traceback and/or browser JavaScript console output
![image](https://user-images.githubusercontent.com/20476096/100673755-dde5f900-3318-11eb-807b-3e65b391acf1.png)
If you notice the url bootstrap.css is trying to load, its `localhost:8888/static` instead of `localhost:8888/proxy/{port}/static`, which infact is the correct url.

Also, other resources like `bokeh` and `panel.min.js` are infact loaded correctly using `localhost:8888/proxy/{port}` as the base

**Expected Behavior**
All assets are loaded similar to the scenario when we run it without jupyter-server-proxy.

#### ALL software version info
bokeh: 2.2.3
panel: 0.10.2

cc @bryevdv @philippjfr

[...]

PR #1891, opened 2021-01-08
fix local bundled paths for template resources

AjayThorve:
While #1821  was partially resolved thanks to #1830 , the template resources were still not loading due to the same path issues. This PR aims to resolve those. Testing in a jupyter_server_proxy env and jupyterlab without proxy seems to work for the changes proposed. 

This allows the following example to work:

```python
import panel as pn
import numpy as np
import holoviews as hv

pn.extension()
react = pn.template.ReactTemplate(title='React Template')

xs = np.linspace(0, np.pi)

def sine(freq, phase):
        responsive=True, min_height=400)

def cosine(freq, phase):
        responsive=True, min_height=400)

react.sidebar.append(freq)
react.sidebar.append(phase)

# Unlike other templates the `ReactTemplate.main` area acts like a GridSpec 
react.main[:4, :6] = pn.Card(hv.DynamicMap(sine), title='Sine')
react.main[:4, 6:] = pn.Card(hv.DynamicMap(cosine), title='Cosine')

react.show(port=5000) # go to localhost:5000
# OR
react.show(websocket_origin="localhost:8888", port=5000) # and then go to localhost:8888/proxy/5000/
```
cc @philippjfr

### DIFF
--- a/panel/template/base.py
+++ b/panel/template/base.py
@@ -449,7 +449,7 @@ class BasicTemplate(BaseTemplate):
         name = type(self).__name__.lower()
         resources = _settings.resources(default="server")
         if resources == 'server':
-            dist_path = urljoin(state.base_url, LOCAL_DIST)
+            dist_path = urljoin(state.app_url, LOCAL_DIST)
         else:
             dist_path = CDN_DIST
 
--- a/panel/template/base.py
+++ b/panel/template/base.py
@@ -448,8 +448,9 @@ class BasicTemplate(BaseTemplate):
     def _template_resources(self):
         name = type(self).__name__.lower()
         resources = _settings.resources(default="server")
+        base_url = state.base_url[1:] if state.base_url.startswith('/') else state.base_url        
         if resources == 'server':
-            dist_path = urljoin(state.app_url, LOCAL_DIST)
+            dist_path = urljoin(base_url, LOCAL_DIST)
         else:
             dist_path = CDN_DIST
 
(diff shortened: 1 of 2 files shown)


panel/template/base.py
```diff
@@ -449,7 +449,7 @@ def _template_resources(self):
         name = type(self).__name__.lower()
         resources = _settings.resources(default="server")
         if resources == 'server':
-            dist_path = urljoin(state.base_url, LOCAL_DIST)
+            dist_path = urljoin(state.app_url, LOCAL_DIST)
```
philippjfr, line 452: Bit confused why this would be necessary (or why it would even work). `app_url` points to the URL of the specific app, e.g. `https://localhost:5006/my_app` and the dist dir is not present at that route.

AjayThorve: Hey @philippjfr , thanks for pointing it out, I wasn't aware of the `app_url` usage and have updated it to use state.base_url again, with a check for `startswith("/")`. 

`base_url = state.base_url[1:] if state.base_url.startswith('/') else state.base_url `

Is this fix okay?

philippjfr: Looks okay. Does it work for your use cases? I'll try this out shortly but if you have already tested the jupyter proxy scenarios let me know.

AjayThorve: jupyter proxy scenarios work for me, and so do the examples provided in the description. Haven't tested the app based env you mentioned - `https://localhost:5006/my_app`, if there is any example to test that you can point me to, I'd be happy to test it out as well.

philippjfr:
Thanks @AjayThorve. Will get a dev release out tomorrow and may release 0.10.3 with some backported fixes including this one.

PR #1892, opened 2021-01-08
Update charts example

kcpevey:
Adds `pyechart` example discussed in [this issue](https://github.com/holoviz/panel/issues/1874). 

I suspect this will also need `pyechart` added to the ci environment but I'm not sure where exactly to do that.

### DIFF
--- a/.github/workflows/docs.yaml
+++ b/.github/workflows/docs.yaml
@@ -50,7 +50,7 @@ jobs:
           conda activate test-environment
           conda list
           doit develop_install ${{ env.CHANS_DEV}} -o doc -o examples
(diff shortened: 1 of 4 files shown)


philippjfr:
Thanks @kcpevey!

kcpevey:
I wasn't sure if the optional dep should be added to setup.py or not. Thanks for taking it the rest of the way!

philippjfr:
Thanks for contributing! I tried, didn't work since setup.py is used as the source of the conda env and it's not yet available on conda. So manual pip install is all we could do for now.

kcpevey:
ah ok. That makes sense!

PR #1895, opened 2021-01-10
Add Trend indicator

MarcSkovMadsen:
The `Trend` indicator enables the user to display a Dashboard KPI Card with

- Title: The name or a short title
- Value: Primary Value to display. For example an absolute value.
- Value_change: A secondary value. For example a percentage change.
- Plot: A plot. One of line", "step", "area", "bar

The card can be layout out as

- a column (text and plot on top of each other) or
- a row (text and plot after each other)

You can run the streaming test example below via 

```bash
 python -m panel serve 'panel\tests\widgets\test_stats_plot_card.py' --dev --show
```

image

**Must do**

- [x] support `row` layout
- [x] document via notebook
- [x] Identify and fix small responsive, resize quirks.
- [x] verify implementation is efficient, i.e. can stream to 30 of these widgets every 500ms without memory or performance problems.
- [x] Make sure all properties are connected as they should.
- [ ] Make it customizable whether y-axis whether contains 0.

**Nice to do**

- [ ] Support card `description` to provide more info and context of the card. Maybe shown as tooltip or via click on card.
- [ ] Add tooltip to plot if it makes sense for streaming.
- [ ] change text size/ layout calculations. Right now the text and plot uses 50% of the available height or width each. This sometimes leaves a lot of empty space in the `text` div.


MarcSkovMadsen:
#### Additional Context

There is a video here https://discourse.holoviz.org/t/statsplotcard-pr/1689

After having studied the different "stats cards" out there I would say there are 3 archetypes

- StatsTextCard: Text Only
- StatsProgressCard: Text and a Progress Indicator
- StatsPlotCard: Text and Plot

They can the be layed out either in a column or row format.

panel/widgets/stats_plot_card.py
```diff
@@ -0,0 +1,61 @@
...
+The card enables performant streaming.
+The text section is responsive and resizes on window resize.
+"""
+import param
+from bokeh.models.sources import ColumnDataSource
+from panel.models.stats_plot_card import StatsPlotCard as _BkStatsPlotCard
+from panel.widgets.base import Widget
+
+
+class StatsPlotCard(Widget):
```
philippjfr, line 22: Let's make this an indicator and move it to `/widgets/indicators.py`

panel/widgets/stats_plot_card.py
```diff
@@ -0,0 +1,61 @@
...
+    """
+    _widget_type = _BkStatsPlotCard
+
+    _rename = {}
+
+    title = param.String(doc="""The title or a short description of the card""")
+    layout = param.ObjectSelector(default="column", objects=["column"])
+    value = param.String(doc="""The primary value to be displayed""")
+    value2 = param.String(doc="""A secondary value to display together with the primary value""")
+    plot_data = param.ClassSelector(
```
philippjfr, line 46: Would prefer to support dictionaries of arrays and dataframes here.

panel/tests/widgets/test_stats_plot_card.py
```diff
@@ -0,0 +1,68 @@
...
+import panel as pn
+from bokeh.models.sources import ColumnDataSource
+from panel.widgets.stats_plot_card import StatsPlotCard
+
+
+def test_constructor():
+    return StatsPlotCard(title="Test")
+
+
+def test_app():
```
philippjfr, line 12: I'd like to organize all your manual tests into one module or in fact make them into the examples and then add some explicit example tests. Doesn't have to be in this PR though.

panel/widgets/stats_plot_card.py
MarcSkovMadsen: I sort of do as well. But I would like to support efficient patching and streaming. That is the main goal.  Do you have an existing API or implementation i could clone @philippjfr ?

MarcSkovMadsen:
One question is also if you @philippjfr would father merge in the stackline (plot) as an indicator and let the user compose it into the card he needs?

panel/widgets/stats_plot_card.py
philippjfr: Could try to reuse what we implemented (I refactored your implementation) on the `BaseTable` class.

philippjfr: The main reason why using Bokeh models directly is not a good idea is that they generally should only be rendered once and you can end up with odd behavior if you reuse them across multiple views.

panel/widgets/stats_plot_card.py
MarcSkovMadsen: done

panel/widgets/stats_plot_card.py
MarcSkovMadsen: I have started looking at this. But it leaves me with a lot of questions

1. If I cannot instantiate one `ColumnDataSource` and use it to `stream` and `patch` data, then I don't understand how I can efficiently `stream` and `patch` data in Panel. The efficiency of the ColumnDataSource is essential for streaming (and in general). Resending the whole series every time it updates is too slow and requires too much cpu.
2. I've looked at the `BaseTable` Class. But it contains a lot of functionality specific to tables. It would not be relevant for the `StatsPlotCard`, `PerspectiveViewer` widget or any other widget where you just want to transfer tabular data efficiently. Should I/ someone refactor the essential parts into a `BaseData` or `BaseTabularData` widget?

panel/tests/widgets/test_stats_plot_card.py
MarcSkovMadsen: Ok. The first step here would be you @philippjfr determining how and where you want those examples organized. A first reference example would be a good idea and maybe a comment in the developer docs about it.

MarcSkovMadsen:
I've added an indicator of the change

image

panel/widgets/stats_plot_card.py
philippjfr: Yes, I'll factor out a TabularData Mixin of some kind. Will look at this asap.

MarcSkovMadsen:
Right now the values I stream are strings because I don't know how the user would like to round, format and add units. That is flexible but also puts more work on the user. Is that how the api should be @philippjfr ?

image

MarcSkovMadsen:
I have implemented the row layout

image

jbednar:
This looks really great; thanks @MarcSkovMadsen !

philippjfr:
@jbednar Any opinions on naming this component?

jbednar:
KPI? KPIPlot? KPICard? MetricCard? StatSummary? Too many possible names to pick just one! :-)

panel/widgets/stats_plot_card.py
philippjfr: Okay, this is done now and I've rewritten the component on top of the ReactiveData baseclass.

philippjfr:
Tbh now looking at this I think I am now consider whether it wouldn't be better split this into distinct indicators for the value and change text and the plot.

jbednar:
Having them be separate is good, since each makes sense on its own, but having it be easy to instantiate as a complete unit is also convenient.

panel/models/stats_plot_card.ts
```diff
@@ -0,0 +1,263 @@
...
+    this.plot = new Plot({
+      background_fill_color: null,
+      border_fill_color: null,
+      outline_line_color: null,
+      min_border: 0,
+      sizing_mode: "stretch_both",
+      toolbar_location: null,
+    });
+
+    console.log(this.plot)
```
philippjfr, line 77:
```suggestion
```

jlstevens:
In the recent HoloViz meeting I think we discussed ideas for better names than `StatsPlotCard`. If I remember right, I suggest something along the lines of `Trend`...I think the last suggestion was `TrendIndicator`.

jbednar:
Hmm. `Dial`, `Gauge`, and `Spinner` all convey "indicator" inherently, so they aren't comparable to something like `Trend` which denotes the underlying numerical quantities.  `Number` and `Progress` are similarly about the underlying quantity, yet somehow they seem more able to be used as indicators without stating that they are so, while `Trend` for me begs to have some qualifier that conveys indicator directly, e.g. `TrendView` or `TrendDisplay` or `Trendline` or `Trendgraph` or `TrendStats` or `TrendSummary` or `TrendIndicator`.

jlstevens:
I quite like `TrendStats`.

jbednar:
If we want to be cute, `Trendicator`. :-)

philippjfr:
Performance indeed pretty decent:

indicator

philippjfr:
I agree with the problem with `TrendIndicator` vs the simpler names for all the others. `TrendStats` I don't like much because it's really not doing much "stats" apart from displaying the change.

PR #1896, opened 2021-01-12
A Lottie Files player example using the bodymovin .js library and the ReactiveHTML component

MarcSkovMadsen:
Hi @philippjfr 

This can be used to discuss, test and improve the api of the `ReactiveHTML` component. And maybe it could end up as a reference example in the Gallery.

The current notebook is preliminary and contains some questions on how I should do different things that I currently cannot do?

A short review and useful comments would be appreciated. Then I will continue making test and reference examples. Thanks.

image


MarcSkovMadsen:
Improved based on todays discussions and api improvements

image

FYI. @philippjfr

philippjfr:
Closing since we have other examples now and tons has changed here.

PR #1897, opened 2021-01-12
Lottie webcomponent player example using the ReactiveHTML component

MarcSkovMadsen:
This can be used to discuss and test the `ReactiveHTML` component. And maybe grow into a reference example for the Gallery.

@philippjfr . There are some things that do not work or that I cannot get working. The questions are in the notebook. I would appreciated some hints or suggestions.  Then I will continue with the example.

image


philippjfr:
This is appreciated, the more examples we can experiment with the better. I'll take this for a spin today.

MarcSkovMadsen:
I was thinking that maybe when the ReactiveHTML gallery examples are ready I or somebody could create similar examples as bokeh extensions for comparison and strength and weaknesses of the approach. And just in general to document how to create custom bokeh extensions.

MarcSkovMadsen:
FYI. @philippjfr

Got some more things working. I could also use the `load` function in the console to load other lottie files. But when I add it to the `_scripts` I get an error. 

There is something about the `_scripts`. As far as I can tell they run initially. I'm in doubt whether that is the right behavior or whether only the `render` script should be triggered????? 

image

philippjfr:
Closing as we have other examples now and tons has changed here.

PR #1898, opened 2021-01-12
Jsme molecule editor example using the ReactiveHTML component

MarcSkovMadsen:
This is an example for discussion and testing of the new `ReactiveHTML` component. This could end up going into the Gallery as a reference example.

@philippjfr . Feel free to explore this as an example. This example contains the complication that the parameter values needs to be extracted from functions on the js side. There is an alternative `jsme-react` component that might be better to use (link in notebook)? But I don't know how to use react with Panel. :-)

image


MarcSkovMadsen:
FYI @philippjfr 

I tried to use the `_script` `render` api of yesterday. But I still cannot get the JSME editor to show. The `render` script is never run.

PR #1911, opened 2021-01-17
Multiple files in FileInput

miliante:
Update input.py to add 'multiple' parameter to FileInput
paramenter filename can be a string or a list
parameter mime_type can be a string or a list
_process_property_change accounts for either a string or a list as input for 'msg'

### DIFF
--- a/panel/widgets/input.py
+++ b/panel/widgets/input.py
@@ -58,11 +58,15 @@ class FileInput(Widget):
 
     accept = param.String(default=None)
 
-    filename = param.String(default=None)
+    filename = param.ClassSelector(default=None, class_=(str, list),
+                               is_instance=True)
 
-    mime_type = param.String(default=None)
+    mime_type = param.ClassSelector(default=None, class_=(str, list),
+                               is_instance=True)
 
     value = param.Parameter(default=None)
+    
+    multiple = param.Boolean(default=True)
 
     _widget_type = _BkFileInput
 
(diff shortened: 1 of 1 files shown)


panel/widgets/input.py
```diff
@@ -58,11 +58,15 @@ class FileInput(Widget):
...
+    filename = param.ClassSelector(default=None, class_=(str, list),
+                               is_instance=True)
 
-    mime_type = param.String(default=None)
+    mime_type = param.ClassSelector(default=None, class_=(str, list),
+                               is_instance=True)
 
     value = param.Parameter(default=None)
+    
+    multiple = param.Boolean(default=True)
```
philippjfr, line 69: Mind sorting the parameters alphabetically?

philippjfr:
Thanks so much! If you could, would you add some documentation for the multiple parameter to `examples/reference/widgets/FileInput.ipynb`?

miliante:
I was thinking, perhaps default should be false to avoid breaking compatibility with what everyone has already.

philippjfr:
>I was thinking, perhaps default should be false to avoid breaking compatibility with what everyone has already.

Definitely.

philippjfr:
Thanks for your contributions @miliante. I'll make the documentation updates in a new PR.

miliante:
Thank you @philippjfr. Sorry for late replies. I'll keep the documentation in mind for next time.

PR #1948, opened 2021-01-28
Fix issues with nullable properties in Bokeh 2.3

philippjfr:


panel/models/layout.py
```diff
@@ -1,20 +1,20 @@
...
+    active_header_background = Nullable(String(help="Background color of active Card header."))
 
     button_css_classes = List(String, help="CSS classes to add to the Card collapse button.")
 
     collapsed = Bool(True, help="Whether the Card is collapsed.")
 
     collapsible = Bool(True, help="Whether the Card should have a button to collapse it.")
 
-    header_background = String(help="Background color of the Card header.")
+    header_background = Nullable(String(help="Background color of the Card header."))
```
mattpap, line 15: You need to unwrap this, so `Nullable(String, help=..."))` or `Nullable(String(), help="...")`. I'm not sure what happens if the wrapped version is used. I guess it's a no-op. If that's the case, then we need to at last notify API creators about this. In the long term, I think, we will have to redesign properties similarly to what happened in bokehjs, where property types (kinds) are detached from properties and their descriptors. This would align bokeh's models implementation with dataclasses (broadly speaking, not necessarily std dataclasses).

philippjfr: Thanks!

PR #1971, opened 2021-02-10
Add --warm option to panel serve

philippjfr:
Allows executing scripts on startup to ensure the caches are warmed up

### DIFF
--- a/panel/command/serve.py
+++ b/panel/command/serve.py
@@ -113,9 +113,9 @@ class Serve(_BkServe):
             help    = "The length of the session history to record.",
             default = 0
         )),
-        ('--warm-cache', dict(
+        ('--warm', dict(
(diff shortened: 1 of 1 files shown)


MarcSkovMadsen:
Remember to include a warm option to `panel.serve` @philippjfr  :-)

philippjfr:
I guess I could but if you're already using `panel.serve` then you are already running a script which can much more effectively "warm up" the session by populating the `pn.state.cache` and running the desired imports. Do you still think it's useful?

MarcSkovMadsen:
In my experience having the same options from panel serve and panel.serve is what is by some users (me) expected. It’s confusing Wien there is a discrepancy.

but i can see here that it might not be needed.

PR #1983, opened 2021-02-16
Add autoreload option

philippjfr:
Adds an `--autoreload` option to `panel serve` which reloads the application when a change in a script/notebook is detected.

https://user-images.githubusercontent.com/1550771/108083929-e9bd9d80-7073-11eb-930a-7543ed413dab.mp4


MarcSkovMadsen:
I want this so badly as it will speed up my workflow a lot.

Does it only reload one module or all modules that change?

philippjfr:
>Does it only reload one module or all modules that change?

Currently reloads the scripts you are serving but I'll be working on making that configurable.

MarcSkovMadsen:
You consider it. I know from Streamlit that reloading nested modules can be difficult. They never got it working 100% even though it was one of the first bugs I reported.

philippjfr:
Can you clarify what you mean by "nested modules"?

MarcSkovMadsen:
You can see a description of some of the problems here https://github.com/streamlit/streamlit/issues/358

philippjfr:
I've now borrowed some code from streamlit after trying a couple of different approaches but I suspect we'll have the same limitations as they do.

panel/io/server.py
```diff
@@ -174,6 +176,52 @@ async def get(self, *args, **kwargs):
...
+    try:
+        def post_check():
+            newdoc = curdoc()
+            # script is supposed to edit the doc not replace it
+            if newdoc is not doc:
+                raise RuntimeError("%s at '%s' replaced the output document" % (self._origin, self._runner.path))
+        self._runner.run(module, post_check)
+    finally:
+        self._unmonkeypatch_io(old_io)
+        set_curdoc(old_doc)
```
philippjfr, line 221:
```suggestion
        bk_set_curdoc(old_doc)
```

MarcSkovMadsen:
Just testing out the `autoreload` feature. It seems to work really, really well. This will definitely speed up my development process.

And seeing errors show up in the browser instead of terminal is also just so nice.

image

PR #1989, opened 2021-02-19
Correctly switch resource mode on models

philippjfr:
Ensures that the model resource mode is handled correctly.

### DIFF
--- a/panel/io/resources.py
+++ b/panel/io/resources.py
@@ -59,7 +59,7 @@ def bundled_files(model, file_type='javascript'):
     for url in getattr(model, f"__{file_type}_raw__", []):
         filepath = url_path(url)
         test_filepath = filepath.split('?')[0]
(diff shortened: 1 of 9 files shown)


panel/io/resources.py
```diff
@@ -36,18 +39,37 @@ def conffilter(value):
...
+
+def bundled_files(model, file_type='javascript'):
+    bdir = os.path.join(PANEL_DIR, 'dist', 'bundled', model.__name__.lower())
+    name = model.__name__.lower()
+    
+    files = []
+    for url in getattr(model, f"__{file_type}_raw__", []):
+        filepath = url_path(url)
+        test_filepath = filepath.split('?')[0]
+        if RESOURCE_MODE0 == 'server' and os.path.isfile(os.path.join(bdir, test_filepath)):
```
philippjfr, line 62:
```suggestion
        if RESOURCE_MODE == 'server' and os.path.isfile(os.path.join(bdir, test_filepath)):
```

Issue #1999, opened 2021-02-23
pn.bind does not support the watch argument


MarcSkovMadsen:
I was just trying out `pn.bind` and I can see that it does not support setting `watch=True`. This limits the use cases. Then it's just easier to stick with `@pn.depends` as it supports all use cases.

In the below example the slider does not work.

```python
import numpy as np
import panel as pn
import pandas as pd
import plotly.graph_objects as go

pn.extension('plotly')

data = {}
for i in range(10):
    xo = 5 * np.random.rand()
    xf = 5 + 5 * np.random.rand()
    x = np.linspace(xo, xf, int(50 + 50*np.random.rand()))
    e = np.random.rand()
    y = x**(1+e) * np.cos(x) + 4*(np.random.rand(len(x)) - 0.5)
    data[i] = {
        'x': x.tolist(),
        'y': y.tolist()
    }

nlines = pn.widgets.IntSlider(name='Number of lines', value=2, start=1, end=9)
timeseries = pn.pane.Plotly()

def get_plot(nlines):
    print("get_plot", nlines)
    fig = go.Figure()
    for i in range(nlines):
        fig.add_trace(
            go.Scatter(x=data[i]['x'], y=data[i]['y'], mode='lines', name=f'line-{i}'))
    timeseries.object = fig
get_plot(nlines.value)
pn.bind(get_plot, nlines=nlines, watch=True)

pn.Row(nlines, timeseries).servable()
```

![image](https://user-images.githubusercontent.com/42288570/108864048-1e8ca000-75f2-11eb-9273-19e8f0d4d2f8.png)

jbednar:
Here I assume you meant `pn.bind(get_plot, nlines=nlines, watch=True)`? That's an interesting suggestion, addressing something @jlstevens was worried about, i.e. that `pn.bind` only generates updates when the output of it is embedded in a Panel layout. I guess with `watch=True` that semantics would change, and it would be registered for updates regardless of whether it's in a layout?

MarcSkovMadsen:
Exactly

jlstevens:
This was indeed something that bothered me about `bind`. 

Like Marc, I would like the option to `watch` it (can be off by default) but the problem I have with `watch=True` is that it means there is a namespace clash with any potential parameter called `watch`. I guess an actual watch parameter would take precedence (in which case you wouldn't be able to watch it?!).

jbednar:
[...]

PR #2000, opened 2021-02-23
Adds the watch argument to pn.bind

MarcSkovMadsen:
Fixes https://github.com/holoviz/panel/issues/1999

### DIFF
--- a/panel/depends.py
+++ b/panel/depends.py
@@ -116,7 +116,8 @@ def _param_bind(function, *args, **kwargs):
         Positional arguments to bind to the function.
     kwargs: object, param.Parameter
         Keyword arguments to bind to the function. The keyword watch is special. If set to True the
-        function is always run.
+        function is always run when a parameter changes. If not specified (default) or set to False
+        the function is only run automatically when embedded in a panel.
 
     Returns
     -------
--- a/panel/depends.py
+++ b/panel/depends.py
@@ -54,7 +54,7 @@ def depends(*args, **kwargs):
     return param.depends(*updated_args, **updated_kwargs)
 
 
-def bind(function, *args, **kwargs):
+def bind(function, *args, watch=False, **kwargs):
     """
     Given a function, returns a wrapper function that binds the values
     of some or all arguments to Parameter values and expresses Param
@@ -82,6 +82,9 @@ def bind(function, *args, **kwargs):
         The function to bind constant or dynamic args and kwargs to.
     args: object, param.Parameter, panel.widget.Widget, or ipywidget
         Positional arguments to bind to the function.
+    watch: boolean
+        Whether the function will be automatically evaluated when one
+        of the parameter dependencies change.
     kwargs: object, param.Parameter, panel.widget.Widget, or ipywidget
         Keyword arguments to bind to the function.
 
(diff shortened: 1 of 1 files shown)


MarcSkovMadsen:
Manual Test Is below

pn_bind_works

```python
import numpy as np
import panel as pn
import pandas as pd

pn.extension('plotly')

data = {}
for i in range(10):
    xo = 5 * np.random.rand()
    xf = 5 + 5 * np.random.rand()
    x = np.linspace(xo, xf, int(50 + 50*np.random.rand()))
    e = np.random.rand()
    y = x**(1+e) * np.cos(x) + 4*(np.random.rand(len(x)) - 0.5)
    data[i] = {
        'x': x.tolist(),
        'y': y.tolist()
    }

nlines = pn.widgets.IntSlider(name='Number of lines', value=2, start=1, end=9)
timeseries = pn.pane.Plotly()

def get_plot(nlines):
    print("get_plot", nlines)
    fig = go.Figure()
    for i in range(nlines):
        fig.add_trace(
            go.Scatter(x=data[i]['x'], y=data[i]['y'], mode='lines', name=f'line-{i}'))
    timeseries.object = fig
get_plot(nlines.value)
pn.bind(get_plot, nlines=nlines, watch=True)

pn.Row(nlines, timeseries).servable()
```

panel/depends.py
```diff
@@ -115,13 +115,20 @@ def _param_bind(function, *args, **kwargs):
     args: object, param.Parameter
         Positional arguments to bind to the function.
     kwargs: object, param.Parameter
-        Keyword arguments to bind to the function.
+        Keyword arguments to bind to the function. The keyword watch is special. If set to True the
+        function is always run when a parameter changes. If not specified (default) or set to False
+        the function is only run automatically when embedded in a panel.
```
jbednar, line 120: Hmm. Would it be clearer to do this as `depends(*args, watch=False, **kwargs)`, and then document `watch` as a separate argument (reverting the change to the `kwargs` docs)?  I'm not sure if that would behave any differently, but it seems more explicit to me.

MarcSkovMadsen: Hmm :-) I thought about that. But does it not mean `watch` has to come prior to any other named argument? 

I did not implement it because I believe today on `@pn.depends` you can throw in `watch` any where. I would like it to behave as `@pn.depends`.

jbednar: Declaring `watch` explicitly should not affect how you call it:

image

philippjfr: +1

philippjfr (approved):
Looks good, thanks!

MarcSkovMadsen:
Thanks. Great to get it in. Now I will start to learn to use it.

I think that for many users explicitly declaring widgets and binding them to a function is easier to understand than parameterized classes, Param and annotations.

Let me know if there is anything I should do.

panel/depends.py
```diff
@@ -82,6 +82,9 @@ def bind(function, *args, **kwargs):
         The function to bind constant or dynamic args and kwargs to.
     args: object, param.Parameter, panel.widget.Widget, or ipywidget
         Positional arguments to bind to the function.
+    watch: boolean
+        Whether the function will be automatically evaluated when one
+        of the parameter dependencies change.
```
jbednar, line 87:
```suggestion
        Whether to evaluate the function automatically whenever one of the
        bound parameters changes.
```

panel/depends.py
```diff
@@ -114,6 +117,9 @@ def _param_bind(function, *args, **kwargs):
         The function to bind constant or dynamic args and kwargs to.
     args: object, param.Parameter
         Positional arguments to bind to the function.
+    watch: boolean
+        Whether the function will be automatically evaluated when one
+        of the parameter dependencies change.
```
jbednar, line 122:
```suggestion
        Whether to evaluate the function automatically whenever one of the
        bound parameters changes.
```

PR #2004, opened 2021-02-24
Add IDOM pane

philippjfr:
This now works as well as I would hope both in the notebook and when deployed. One thing I had to do since I'm using preact rather than React is swap out the build paths and provide my own package.json. This also means that external JS components have to be served either by the Panel `jupyter_server_extension` or by the Bokeh server. I'll have to do some more testing that the Jupyter server extension is installed properly in all cases but that's technically separate from this PR. Another side-effect of this is that users have to go via `pn.pane.IDOM.install` (rather than `idom.install`) to install external packages to ensure that all the paths are set correctly.

One nicety I've provided for integration with Panel/Param is `pn.pane.IDOM.use_param`, which can be used like so to get a state value which mirrors a parameter or widget value:

```python
aw = pn.widgets.IntSlider(name='a', start=0, end=20, value=1)
bw = pn.widgets.IntSlider(name='b', start=0, end=20, value=1)

@idom.component
def view():
    a = pn.pane.IDOM.use_param(aw)
    b = pn.pane.IDOM.use_param(bw.param.value) # equivalent to passing in the widget
    return idom.html.div({}, f'{a}+{b}={a+b}')

pn.Row(aw, bw, view)
```

Screen Shot 2021-02-25 at 12 34 56 PM

In the long-term I would still like to have a much simpler implementation which leverages `idom-client-react` and is shipped as a standalone bokeh extension in http://github.com/idom-team/idom-bokeh.


MarcSkovMadsen:
Can you also set a parameter? So implementing a Material Slider in Idom but use the value of the slider outside Idom?

philippjfr:
>Can you also set a parameter? So implementing a Material Slider in Idom but use the value of the slider outside Idom?

You can do whatever you like in the idom callbacks.

rmorshea:
@philippjfr I started work on [idom-bokeh](https://github.com/idom-team/idom-bokeh), but haven't been able to get it to work unfortunately. The docs are a little thin when it comes to pre-built extensions so I haven't been able to make much more progress. I've written up [an issue](https://github.com/idom-team/idom-bokeh/issues/1) with some info on the errors I'm getting.

PR #2021, opened 2021-02-26
Remove change of theme_toggle

MarcSkovMadsen:
When testing Panel 0.11.0rc3 I see an unintended behaviour from the FastListTemplate. When setting the theme via a string it sets `theme_toggle` to False. This was never intended. And it will make it difficult for my users of awesome-panel.org and at work to switch the theme.

So I removed the line of code.

### DIFF
--- a/panel/template/fast/base.py
+++ b/panel/template/fast/base.py
@@ -1,8 +1,8 @@
 import pathlib
 
-import panel as pn
 import param
 
+from ...io.state import state
 from ..base import BasicTemplate
 from ..react import ReactTemplate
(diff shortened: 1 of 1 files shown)


philippjfr:
This change does not actually work, on reload it will never hit the `if "theme" not in params:` branch and therefore continue to default to the hardcoded theme.

philippjfr (changes requested):
This change does not actually work, on reload it will never hit the `if "theme" not in params:` branch and therefore continue to default to the hardcoded theme.

PR #1690, opened 2020-10-24
Adds better Tables and interactive analysis via Perspective, Tabulator and PivotTable

MarcSkovMadsen:
The support for DataFrame Tables in Panel is strong if you look at the underlying functionality. But the look, feel and user experience could be better. See

- https://github.com/holoviz/panel/issues/1525
- https://github.com/holoviz/panel/issues/1486
- https://github.com/holoviz/panel/issues/1223
- https://github.com/holoviz/panel/issues/1178
- https://github.com/holoviz/panel/issues/1177
- https://github.com/holoviz/panel/issues/1108

This PR addresses this by adding the awesome `PerspectiveViewer`, `Tabulator` and `PivotTable` widgets.

## Background Info

**I believe the PerspectiveViewer and Tabulator widgets are ready for deployment**. I would expect no issues from normal usage. I have used them. There are a lot of pytests in the PR and there are working Jupyter Notebook examples included.

In some advanced streaming example there might be some issues. But for that we need the use cases and request from users.

But I strongly believe these widgets should get out there and be available to the Panel community. They are (also) game changers for Panel.
 
The implementation of the PivotTable widget is very basic. But the functionality that is there works.

I contribute these together in one PR as **they share some common code for "Dataframe" widgets**. So actually this PR also contributes some code that should make it easier to contribute Dataframe based widgets in the future. An example from awesome-panel-extensions is the

### PerspectiveViewer

The tests are in the `test_perspective.py` file. 

It also contains a test app you can run with `python -m panel serve 'panel\tests\widgets\test_perspective.py' --dev` or `python 'panel\tests\widgets\test_perspective.py'`. There is also a reference guide notebook called `PerspectiveViewer.ipynb`

image

### Roadmap

- Support other Dataframe types than Pandas.
- Improve implementation of streaming and patching. Currently the perspective-viewer webcomponent loads all data from scratch.
- Eventually integrate with the full Perspective package and apache arrow

FYI. @ceball. This might interest you.

### Tabulator

The tests are in the `test_tabulator` folder. 

It also contains a test app you can run with `python -m panel serve 'panel\tests\widgets\test_tabulator\test_tabulator_examples.py' --dev`. There is also a reference guide notebook called `Tabulator.ipynb`.

As you can see below the Tabulator Widget will support the dark themed Templates pretty well using the `midnight` theme. It also contains other themes like `bootstrap` and `material`.

image

[...]

PR #2034, opened 2021-02-28
Implement Perspective pane

philippjfr:
Supersedes https://github.com/holoviz/panel/pull/1690

```python
df = pd.DataFrame(np.random.randn(10, 4), columns=list('ABCD')).cumsum()
rollover = pn.widgets.IntInput(name='Rollover', value=15)
perspective = pn.pane.Perspective(
    df, sizing_mode='stretch_width', height=400, theme='material-dark'
)
def stream():
    data = df.iloc[-1] + np.random.randn(4)
    perspective.stream(data, rollover=rollover.value)
cb = pn.state.add_periodic_callback(stream, 50)
pn.Column(
    pn.Row(cb.param.period, rollover, perspective.param.theme),
    perspective
)
```

https://user-images.githubusercontent.com/1550771/109423086-76eee380-79de-11eb-8274-f55b6a45f5ea.mp4


MarcSkovMadsen:
So grateful. For me this will be so useful. Thanks.

Like Christmas Eve.

Issue #581, opened 2019-08-08
allow_None in a param parameter doesn't allow empty String widgets


hyamanieu:
Software Versions:
```
bokeh 1.3.4
panel 0.6.1
jupyter core 4.5.0
jupyter-notebook 6.0.0
```

I would expect setting a `param.Integer` with `allow_None = True` to return a widget that accepts empty values and returns None. Alternatively, I would expect using a `LiteralInput `with `placeholder="0" `and `type=int` to set "0" when entering no value (as both None & 0 are considered as False in `if` statements)
Instead, `(invalid)` is appended to the widget name and no callback happens.

Example 1:
```python
class CustomExample(param.Parameterized):
    """An example Parameterized class"""

    select_string = param.Selector(objects=["red", "yellow", "green"])
    select_number = param.Selector(objects=[0, 1, 10, 100])
    some_number = param.Integer(default = None, allow_None=True)
    
pn.Param(CustomExample.param, widgets={
    'select_string': {'type': pn.widgets.RadioButtonGroup, 'button_type': 'success'},
    'select_number': pn.widgets.DiscretePlayer
}
)
```

Example 2:
```python
class CustomExample(param.Parameterized):
    """An example Parameterized class"""

    select_string = param.Selector(objects=["red", "yellow", "green"])
    select_number = param.Selector(objects=[0, 1, 10, 100])
    some_number = param.Integer(default = None, allow_None=True)
    
pn.Param(CustomExample.param, widgets={
    'select_string': {'type': pn.widgets.RadioButtonGroup, 'button_type': 'success'},
    'select_number': pn.widgets.DiscretePlayer,
    'some_number': {'type': pn.widgets.LiteralInput, 'placeholder':0,type:int}
}
)
```

jbednar:
I agree with your first point; there should be a way to select nothing and to end up with the value set to None.  I don't know how that would actually be achieved with the widgets available, but I do think that's what we should strive for.

For LiteralInput, the placeholder is meant for things like "Type your number here", so I would _not_ have an expectation that the contents of that string would be processed by Python.  If you want the value to actually be zero, shouldn't you just set the initial value to zero (rather than setting the placeholder to the string "0")?

hyamanieu:
My bad, I misunderstood the "placeholder" function. However, whatever I write into placeholder, it never shows up.

For now, I guess I have to explicitly tell the user to enter 0 when she/he doesn't anything.

jbednar:
[...]

PR #2038, opened 2021-03-01
allow None as default value for Numeric widgets

xavArtley:
Partially solve https://github.com/holoviz/panel/issues/581

image

```python
class CustomExample(param.Parameterized):
    """An example Parameterized class"""

    select_string = param.Selector(objects=["red", "yellow", "green"])
    select_number = param.Selector(objects=[0, 1, 10, 100])
    some_number = param.Integer(default = None, allow_None=True)
    
pn.Param(CustomExample.param, widgets={
    'select_string': {'type': pn.widgets.RadioButtonGroup, 'button_type': 'success'},
    'select_number': pn.widgets.DiscretePlayer
}
)
```
still produce
image
 we need to declare explicitly `'some_number': {"placeholder": "", "value":None},`
to produce
image

this part of the code prevent fully working but I don't know the global consequences to remove it
https://github.com/holoviz/panel/blob/aec90cfca5da1e4e9a639fb594407efc61f3be08/panel/param.py#L388-L391

### DIFF
--- a/panel/widgets/input.py
+++ b/panel/widgets/input.py
@@ -237,12 +237,13 @@ class _SpinnerBase(_NumericInputBase):
     __abstract = True
 
     def __init__(self, **params):
-        if params.get('value') is None:
-            value = params.get('start', self.value)
+        value = params.get('value', self.value)
+        if value is None:
+            value = params.get('start', value)
             if value is not None:
                 params['value'] = value
-        if 'value' in params and 'value_throttled' in self.param:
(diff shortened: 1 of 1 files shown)


Hoxbro:
Seems to be a lot of overlap with my PR #1837 and with the same conclusions.

xavArtley:
Indeed since we haven't solved the same issue number I haven't seen your PR.

Issue #2040, opened 2021-03-02
Panel 0.11.05rc: Fast Template uses Vanilla Theme


MarcSkovMadsen:
### Panel 0.11.05rc

![image](https://user-images.githubusercontent.com/42288570/109599891-5d53b600-7b1c-11eb-91fb-5811759ffd99.png)

### awesome-panel-extensions for Panel 0.10.3

I've learned you should remove unnecessary information. And I don't why the side of the plot cannot be transparent/ same color as background. Looks much more calm and modern to me.

![image](https://user-images.githubusercontent.com/42288570/109599910-680e4b00-7b1c-11eb-9b7a-bb5a5617223d.png)

MarcSkovMadsen:
I can see the problem comes when specifying `theme='dark'` then the Vanilla Dark theme is used and not the Fast Dark theme

```bash
$ pytest 'scripts\test_fast.py'
Test session starts (platform: win32, Python 3.7.6, pytest 5.4.2, pytest-sugar 0.9.3)
rootdir: C:\repos\private\awesome-panel, inifile: pytest.ini
plugins: anyio-2.0.2, dash-1.17.0, cov-2.8.1, mock-3.1.0, sugar-0.9.3
collecting ...

―――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――― test_template_theme_parameter ―――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――― 

def test_template_theme_parameter():
        template = pn.template.FastGridTemplate(title="Fast", theme="dark")
        # Not '#3f3f3f' which is for the Vanilla theme
>       assert template.theme.bokeh_theme._json['attrs']['Figure']['background_fill_color']=="#181818"
E       AssertionError: assert '#3f3f3f' == '#181818'
E         - #181818
E         + #3f3f3f

scripts\test_fast.py:6: AssertionError

 scripts/test_fast.py ⨯                                                                                                                                                                                                                          100% ██████████
=================================================================================================================== short test summary info =================================================================================================================== 
FAILED scripts/test_fast.py::test_template_theme_parameter - AssertionError: assert '#3f3f3f' == '#181818'

Results (1.79s):
       1 failed
         - scripts/test_fast.py:3 test_template_theme_parameter
```

```python
import panel as pn

def test_template_theme_parameter():
[...]

PR #2041, opened 2021-03-02
Fix: Fast Template to use Fast Theme instead of Vanilla Theme

MarcSkovMadsen:
Fixes https://github.com/holoviz/panel/issues/2040

### DIFF
--- a/panel/template/fast/base.py
+++ b/panel/template/fast/base.py
@@ -5,12 +5,14 @@ import param
 from ...io.state import state
 from ..base import BasicTemplate
 from ..react import ReactTemplate
-from .theme import THEMES, DefaultTheme
+from .theme import THEMES, FastDefaultTheme
 
 _ROOT = pathlib.Path(__file__).parent
 
 
 class FastBaseTemplate(BasicTemplate):
+    _THEMES = THEMES
+    _DEFAULT_THEME = FastDefaultTheme
 
     theme_toggle = param.Boolean(default=True, doc="""
         If True a switch to toggle the Theme is shown.""")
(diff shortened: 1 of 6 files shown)


MarcSkovMadsen:
Would it be possible to get this one in @philippjfr . It's important as the whole look and feel of the Fast Templates with HoloViews plots is different.

MarcSkovMadsen:
After reviewing this I can see my first fix only fixed problems on the surface. I got it to use the FastDarkTheme. But I should actually get it to use the FastListDarkTheme and FastGridDarkTheme respectively.

I've now refactored the tests and implementation. I believe I have solved it.

But I still have problems getting the panel `dev` environment running. So I cannot thoroughly test it. Sorry.

MarcSkovMadsen:
@philippjfr . The two failing tests I simply don't understand. I think it has something todo with how templates and themes are related. I have tried to understand that for a long time but not yet been able to. I've fix the problem but as I simply don't understand the flow its not possible for me.

I can see though that where it fails if the theme is instantiated `()` there is no error. But the theme parameter is a class and not an instance. 

image

I have tried to understand what a property descriptor is without luck.

`ValueError: Parameter 'theme' value must be an instance of (Theme, str), not '<property object at 0x7fb96f5fc5f0>'`

**Can you help?**

xavArtley:
I think you make a wrong usage of the property decorator:
If you don't instantiate your class you are just returning the declaration of your property
image

philippjfr:
This is not how themes work. They get resolved internally to the actual theme class that will be applied but the user and the parameter hold the base class.

philippjfr:
Anyway I'll check into why the bokeh theme isn't correctly applied.

philippjfr:
Basically a user should never see or know about the fact that there are template specific themes. They import and provide the theme base class and in the background it resolves it to the concrete theme object for that template. Closing this PR and investigating.

Issue #1977, opened 2021-02-15
Parse problems when mapping param.Array() to a widget


LeonvanKouwen:
#### ALL software version info
panel                     0.10.3             pyhd3eb1b0_0
param                   1.10.1             pyhd3eb1b0_0

#### Description of expected behavior and the observed behavior
When running the MWE below I get 
* Input Widget: [1 2 3]
* Output String: [2 4 6]

When try to change the value of one of the array elements , I get an '(invalid)' input.
When I change the input to something like
* Input Widget: [1, 2, 4]

The js error is gone, but python gives
`ValueError: Parameter 'array' value must be an instance of ndarray, not '[1, 2, 4]'`

So it seems that two things go wrong:
1. The comma's are lost when showing the array in the input field
2. The input isn't parsed with `np.array()`

#### Complete, minimal, self-contained example code that reproduces the issue

```
import panel as pn
import param
import numpy as np



class ArrayDemo(param.Parameterized):

    array  = param.Array(np.array((1, 2, 3)))

    @param.depends('array')
    def show(self):
        return str(2 * self.array)


demo = ArrayDemo()
pn.serve(pn.Column(demo.param, demo.show))
```

PR #2047, opened 2021-03-02
Implement array input

xavArtley:
fix https://github.com/holoviz/panel/issues/1977

array_input

### DIFF
--- a/panel/widgets/input.py
+++ b/panel/widgets/input.py
@@ -370,6 +370,21 @@ class LiteralInput(Widget):
         return msg
 
 
+class ArrayInput(LiteralInput):
+
+    def _process_property_change(self, msg):
+        msg = super()._process_property_change(msg)
+        if 'value' in msg and isinstance(msg['value'], list):
+            import numpy as np
+            msg['value'] = np.asarray(msg['value'])
+        return msg
+
+    def _process_param_change(self, msg):
+        if 'value' in msg:
+            msg['value'] = msg['value'].tolist()
+        return super()._process_param_change(msg)
+
+
 class DatetimeInput(LiteralInput):
     """
     DatetimeInput allows declaring Python literals using a text
@@ -451,9 +466,6 @@ class DatetimeRangeInput(CompositeWidget):
 
     end = param.Date(default=None)
 
-    format = param.String(default='%Y-%m-%d %H:%M:%S', doc="""
-        Datetime format used for parsing and formatting the datetime.""")
-
     _composite_type = Column
 
     def __init__(self, **params):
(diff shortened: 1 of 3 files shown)


panel/widgets/input.py
```diff
@@ -370,6 +370,21 @@ def _process_param_change(self, msg):
         return msg
 
 
+class ArrayInput(LiteralInput):
+
+    def _process_property_change(self, msg):
+        msg = super()._process_property_change(msg)
+        if 'value' in msg and isinstance(msg['value'], list):
+            import numpy as np
```
philippjfr, line 378: NumPy is a core Bokeh requirement so you can safely import this above.

xavArtley:
I made a mistake I remove the format correction

philippjfr:
I'm a little worried about merging this, what happens when you put in a 1000x1000 array?

philippjfr:
Having the `ArrayInput` is fine but not guarding for the size of the array when doing the widget mapping seems like it might be problematic.

xavArtley:
Indeed it can only be used to edit small arrays

panel/widgets/input.py
```diff
@@ -370,6 +372,33 @@ def _process_param_change(self, msg):
...
+            msg['value'] = np.asarray(msg['value'])
+        return msg
+
+    def _process_param_change(self, msg):
+        if 'value' in msg and msg['value'] is not None:
+            if msg['value'].size <= self.max_array_size:
+                msg['value'] = msg['value'].tolist()
+        msg = super()._process_param_change(msg)
+        if 'value' in msg:
+            msg['value'] = msg['value'].replace(' ', ', ')
```
jbednar, line 398: Can you add a comment above this line explaining why it's needed? Will it cause any issues for arrays of strings that might have spaces or commas?

panel/widgets/input.py
```diff
@@ -370,6 +372,33 @@ def _process_param_change(self, msg):
         return msg
 
 
+class ArrayInput(LiteralInput):
+
+    max_array_size = param.Number(default=1000, doc="""
+        maximum number of elements in the array to allow full serialization
+        else only the string representation will be passed to the javascript
+        side and edition capabilities through the widget are lost.
```
jbednar, line 380:
```suggestion
        Arrays larger than this limit will be allowed in Python but will not
        be serialized into JavaScript. Although such large arrays will thus
        not be editable in the widget, such a restriction helps avoid 
        overwhelming the browser and lets other widgets remain usable.
```

Also, what happens in the case of a large array? It doesn't look like the user will see any message, and I don't know what happens if the user does try to edit it in the browser. Ideally I'd think the widget should show the first few values along with a message indicating that it's truncated and not editable, and somehow any updating messages from the browser would be dropped if the user does edit it (and ideally they would not be allowed to edit it at all).

panel/widgets/input.py
philippjfr: Also a bit confused here, shouldn't the `.tolist()` above deal with this since only arrays have a repr that drops the commas?

xavArtley: I forgot to drop the replace, I previously used it on the array repr, but with the tolist it is not needed anymore

xavArtley:
array_inputt

PR #2053, opened 2021-03-03
Fix visual look of markdown reference example

Jacob-Barhak:
The current visual html on the panel web site is broken with one of the examples - the conversion to html failed to convert a code cell and it looks like an ugly string. This version of the notebook fixes it and it visually looks ok.


philippjfr:
Thanks @Jacob-Barhak! Could you please clear the notebook before committing it?

Jacob-Barhak:
Sorry Philipp,

You should have it fixed momentarily.

Jacob

PR #2056, opened 2021-03-05
Allow bind to extract dependencies and evaluate other dynamic functions

philippjfr:
Allows binding another dynamic function:

```python
slider = pn.widgets.IntSlider(value=2, start=0, end=10)
slider2 = pn.widgets.IntSlider(value=4, start=0, end=10)
slider3 = pn.widgets.IntSlider(value=2, start=0, end=10)

def add(a, b):
    return a + b

def divide(a, b):
     return a / b

bound = pn.bind(divide, pn.bind(add, slider, slider2), slider3)
assert bound() == 3
```


jbednar (approved):
Cool!

Issue #2052, opened 2021-03-03
Expose the visible attribute on widgets and indicators


sdc50:
I often want to create widgets that are initially not visible. The underlying Bokeh widgets have a `visible` attribute, but this is not exposed in the Panel API. 

The following example demonstrates a pattern that I often have a need for:

```python
delete_btn = pn.widgets.Button(name='Delete', button_type='danger', width=100)
confirm_btn = pn.widgets.Button(name='Confirm', button_type='danger', width=100)
cancel_btn = pn.widgets.Button(name='Cancel', button_type='success', width=100)
spn = pn.widgets.indicators.LoadingSpinner(value=True, color='primary', aspect_ratio=1, width=0)

args = {'del': delete_btn, 'conf': confirm_btn, 'can': cancel_btn, 'spn': spn}
delete_btn.js_on_click(args=args, code='del.visible=false; conf.visible=true; can.visible=true;')
cancel_btn.js_on_click(args=args, code='del.visible=true; conf.visible=false; can.visible=false;')
confirm_btn.js_on_click(args=args, code='conf.visible=false; can.visible=false; spn.width=50;')


pn.Row(delete_btn, confirm_btn, cancel_btn, spn)
```

![panel_widgets_visible](https://user-images.githubusercontent.com/6454472/109832683-3671ae80-7c06-11eb-8349-5cf3adeeeb47.gif)

In this example I want the `confirm_btn` and `cancel_btn` to initially be hidden. I am able to hide them with JavaScript (by clicking the cancel button), but there is no way to declare them as not visible on instantiation. 

To get around this for the time being I'm using the `css_classes` attribute, but this requires adding additional CSS styles and more cumbersome JavaScript:

```python
pn.config.raw_css.append('.hidden{visibility: hidden;}')

delete_btn = pn.widgets.Button(name='Delete', button_type='danger', width=100)
confirm_btn = pn.widgets.Button(name='Confirm', button_type='danger', width=100, css_classes=['hidden'])
cancel_btn = pn.widgets.Button(name='Cancel', button_type='success', width=100, css_classes=['hidden'])
spn = pn.widgets.indicators.LoadingSpinner(value=True, color='primary', aspect_ratio=1, width=0) 

args = {'del': delete_btn, 'conf': confirm_btn, 'can': cancel_btn, 'spn': spn}
[...]

PR #2065, opened 2021-03-09
Expose widgets visible property

xavArtley:
fix https://github.com/holoviz/panel/issues/2052


philippjfr:
Thanks! Do you know if all widgets support this properly?

MarcSkovMadsen:
One big, but invisible step for man :-) Thanks @xavArtley

PR #2067, opened 2021-03-10
Fix loading of template resources from relative paths

### LINKED ISSUE #2066: Deployment of Templated Apps to Binder using jupypter-panel-proxy not working.
Panel 0.11

## My Pain

I would like to demonstrate to colleagues and the Panel community how easy it is to share Panel apps on binder. It's really just adding `jupyter-panel-proxy` as described in the Panel docs and your notebooks in the root directory.

But no. The world is not so easy and the FastGridTemplate does not work

## Reproducing Steps

Open https://mybinder.org/v2/gh/MarcSkovMadsen/load-forecasts/HEAD?urlpath=/panel/app

## Screenshot of bug

image

## Repository

https://github.com/marcskovmadsen/load-forecasts

## Additional Context

I would also in the near future have access to a JupyterHub and would like to run my panel apps there. And it would be the same problem on JupyterHub. 

So there is a large community of users on Universities, Govt. institutions and in Enterprises that do not have easy access to share their Panel apps via JupyterHub.

FYI. @WesleyTheGeolien

philippjfr:
Fixes #2066

### DIFF
--- a/panel/template/base.py
+++ b/panel/template/base.py
@@ -462,9 +462,9 @@ class BasicTemplate(BaseTemplate):
     def _template_resources(self):
         name = type(self).__name__.lower()
         resources = _settings.resources(default="server")
-        base_url = state.base_url[1:] if state.base_url.startswith('/') else state.base_url        
         if resources == 'server':
-            dist_path = urljoin(base_url, LOCAL_DIST)
+            base_url = state.base_url[1:] if state.base_url.startswith('/') else state.base_url        
+            dist_path = '/' + urljoin(base_url, LOCAL_DIST)
         else:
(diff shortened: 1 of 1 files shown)


philippjfr:
@xavArtley Do you know what this error is about?

>ValueError: VTKVolume pane does not support objects of type 'UniformGrid'.

xavArtley:
May be but I don't understand the link with this issue

philippjfr:
There isn't one, all PRs are currently failing due to this error.

xavArtley:
May be a change in pyvista version?

xavArtley:
It's strange I installed latest pyvista version but it should not raise error
image

Issue #2054, opened 2021-03-05
Perspective raises difficult to understand ValueError if data column is of type int


MarcSkovMadsen:
Panel 0.11

If you run the below with `panel serve` you get a `ValueError`

```python
import panel as pn
pn.extension("perspective")

data = {0: ['1981 01 01 00   161    28 10173   270    21     0     0     0',
  '1981 01 01 01   111    33 10175   270    21     0     0 -9999',
  '1981 01 01 02   111    39 10183   330    26     0     0 -9999',
  '1981 01 01 03    94    39 10192 -9999     0     0     0 -9999',
  '1981 01 01 04    72    39 10196 -9999     0     0     0 -9999'],
 'year': [1981, 1981, 1981, 1981, 1981],
 }

pn.pane.Perspective(data).servable()
```

```bash
ValueError: failed to validate ColumnDataSource(id='1021', ...).data: expected an element of ColumnData(String, Seq(Any)), got {0: ['1981 01 01 00 161 28 10173 270 21 0 0 0', '1981 01 01 01 111 33 10175 270 21 0 0 -9999', '1981 01 01 02 111 39 10183 330 26 0 0 -9999', '1981 01 01 03 94 39 10192 -9999 0 0 0 -9999', '1981 01 01 04 72 39 10196 -9999 0 0 0 -9999'], 'year': [1981, 1981, 1981, 1981, 1981]}

Traceback (most recent call last):
  File "c:\repos\load-forecasts\.venv\lib\site-packages\bokeh\application\handlers\code_runner.py", line 197, in run
    exec(self._code, module.__dict__)
  File "C:\repos\load-forecasts\test_error.py", line 12, in <module>
    pn.pane.Perspective(data).servable()
  File "c:\repos\load-forecasts\.venv\lib\site-packages\panel\viewable.py", line 314, in servable
    self.server_doc(title=title, location=True)
  File "c:\repos\load-forecasts\.venv\lib\site-packages\panel\viewable.py", line 765, in server_doc
    model = self.get_root(doc)
  File "c:\repos\load-forecasts\.venv\lib\site-packages\panel\pane\base.py", line 254, in get_root
    root = self._get_model(doc, comm=comm)
  File "c:\repos\load-forecasts\.venv\lib\site-packages\panel\pane\perspective.py", line 135, in _get_model
    properties = self._process_param_change(self._init_params())
[...]

PR #2069, opened 2021-03-10
Handle integer column names on Perspective widget

philippjfr:
Fixes https://github.com/holoviz/panel/issues/2054

### DIFF
--- a/panel/pane/perspective.py
+++ b/panel/pane/perspective.py
@@ -134,9 +134,9 @@ class Perspective(PaneBase, ReactiveData):
 
     def _process_param_change(self, msg):
         msg = super()._process_param_change(msg)
-        for param in ('columns', 'row_pivots', 'column_pivots'):
-            if msg.get(param):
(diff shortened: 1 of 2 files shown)


philippjfr:
Yes, we should check there's no overlap.

panel/pane/perspective.py
```diff
@@ -130,6 +132,41 @@ def _init_params(self):
         props['source'] = ColumnDataSource(data=self._data)
         return props
 
+    def _process_param_change(self, msg):
+        msg = super()._process_param_change(msg)
+        for param in ('columns', 'row_pivots', 'column_pivots'):
+            if msg.get(param):
+                msg[param] = [str(col) for col in msg[param]]
```
Hoxbro, line 139:
```suggestion
        for par in ('columns', 'row_pivots', 'column_pivots'):
            if msg.get(par):
                msg[par] = [str(col) for col in msg[par]]
```
I think this is the reason for the flake error.

PR #2071, opened 2021-03-10
Check vtkmodules and vtk in sys.modules

xavArtley:
Fix CI errors relative to VTKVolume


philippjfr:
Thanks, really appreciate the quick turnaround.

PR #2079, opened 2021-03-13
Tqdm Indicator

MarcSkovMadsen:
One very powerful tool for progress reporting is tqdm. It gives very nice reporting for loops and also for a lot of numpy/ Pandas operations.

I would like to have the same capabilities inside my Panel apps whether its on the server or in a notebook. But especially in combination with the new ` --auto` flag that makes the panel server really suited for exploratory work also it would be really nice to display tqdm in the app.

So this is a PR to contribute the Tqdm Progress Indicator

tqdm-progress2

## Must Do:

[x] Notebook Reference Guide
[x] Consider renaming `TQDMProgress` to `TQDM`.
[x] Rename to `Tqdm`.
[x] Remove initial `not active` indication.
[x] Move test to test file.
[x] Fix problems that ipywidgets are required but not used in notebook

## Nice to Have

[] Set `value`, `max` and `text` parameters to constant
    - When I try I get `TypeError: Constant parameter 'max' cannot be modified` even though I use `param.edit_constant`. DON'T KNOW HOW TO FIX THIS.
[] Fix layout/ margin issues for "row" layout in notebook (see pictures in discussion below). DON'T KNOW HOW TO FIX THIS.
[x] Consider whether adding `__call__` is good or bad for the user.
    - With `call` it's easier to use but it also might confuse some users as `Tqdm.tqdm` now contains the `pandas` method. (C.f. notebook). I WILL KEEP IT. I THINK ITS OK.
[x] Consider converting to Bokeh Extension instead of `Viewer`. WONT DO.
[x] Move to indicators.py file (throws errors if I try).
[x] Inherit from Indicator (throws errors if I try). WON'T DO. Tqdm is a `Viewer` and not a `Widget`.
[] Verify that nested tqdm works.


MarcSkovMadsen:
#### Alternative API.

Originally I would have liked the api to be, i.e. `tqdm` instead of `tqdm.tqdm` in the loop. But I could not figure out how to implement it.

```python
 tqdm = TQDMProgress(width=800, layout="column")
    def run(*events):
        for index in tqdm.tqdm(range(0,10)):
            time.sleep(0.2)
```

But I ended up with.

```python
 tqdm = TQDMProgress(width=800, layout="column")
    def run(*events):
        for index in tqdm.tqdm(range(0,10)):
            time.sleep(0.2)
```

I don't think it will make a difference. Maybe the api I ended up with makes more sense.

Hoxbro:
This looks pretty cool!

To get it to works with `tqdm` you could add a `__call__` to your `TQDMProgress` class like this:
``` python
    def __call__(self, *args, **kwargs):
        return self.tqdm(*args, **kwargs)
```

MarcSkovMadsen:
Thanks  @Hoxbro . It was actually `__call__` I tried but could not get working. But powered by the inspiration of your kind words I will try again :-)

MarcSkovMadsen:
Got it working @Hoxbro . It was easy. Maybe I forgot `self` in the first place.

I think the simplification is worth it.

tqdm-progress3

MarcSkovMadsen:
Added notebook documentation.


Uploading tqdm-reference.mp4…

panel/widgets/tqdm_indicator.py
```diff
@@ -105,19 +130,31 @@ def close(self):
                 super().close()
                 indicator.reset()
                 return _tqdm
+
         return ptqdm
 
 
 if __name__.startswith("bokeh"):
```
philippjfr, line 137: Presumably for testing?

casperdcl:
as a minor point, `tqdm` is used as a word rather than an acronym (so tqdm or Tqdm but never TQDM)

> `tqdm` means "progress" in Arabic (*taqadum*, تقدّم) and is an abbreviation for "I love you so much" in Spanish (*te quiero demasiado*).

:stuck_out_tongue_winking_eye:

panel/widgets/tqdm_indicator.py
MarcSkovMadsen: Yes. Should (re-)moved

MarcSkovMadsen:
Ahh. Thanks @casperdcl

MarcSkovMadsen:
One issue I don't know how to solve in the current implementation is the "row" alignment in the notebook. It looks like

image

but on server it looks like

image

MarcSkovMadsen:
Hi @philippjfr 

The PR is now at a stage where it would be helpful with a first review and some feedback on 

- whether you would like the Tqdm indicator in Panel
- what "must" be done before it is ready for acceptance.

You can see the list of outstanding things that would be nice to solve at the top. I am stuck there solving those things and would need some hints, guidance or help there.

I don't plan to do more work before I get the feedback :-) Thanks.

panel/widgets/tqdm_indicator.py
MarcSkovMadsen: removed now.

panel/widgets/tqdm_indicator.py
```diff
@@ -0,0 +1,139 @@
+# I did not put it into the indicators.py file yet as it raises errors due to circular imports I
+# believe
+import panel as pn
+import param
+from tqdm.asyncio import tqdm as _tqdm
```
philippjfr, line 5: Panel uses relative imports, never us an absolute import like this (this is probably also the source of your circular import issues).

```suggestion
import param

from tqdm.asyncio import tqdm as _tqdm

from ..viewable import Viewer
```

panel/widgets/tqdm_indicator.py
```diff
@@ -0,0 +1,139 @@
...
+# I cannot have it inherit from pn.indicators.BaseIndicator as that raises an exception
+class Tqdm(pn.viewable.Viewer):
+    value = param.Integer(
+        default=0,
+        bounds=(0, None),
+        doc="""
+        The current value of the progress bar. If set to None the progress
+        bar will be indeterminate and animate depending on the active
+        parameter.""",
+    )
```
philippjfr, line 24: Please be consistent about the formatting.

```suggestion
    value = param.Integer(default=0, bounds=(0, None), doc="""
        The current value of the progress bar. If set to None the progress
        bar will be indeterminate and animate depending on the active
        parameter.""")
```

examples/reference/indicators/Tqdm.ipynb
```diff
@@ -0,0 +1,177 @@
...
+    "from panel.widgets.tqdm_indicator import Tqdm\n",
+    "\n",
+    "pn.extension()"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "# Tqdm Indicator\n",
```
philippjfr, line 24: No titles in reference notebooks.

MarcSkovMadsen: Ok. I have implemented it. 

But have you thought about how Screenshots or Videos of a Reference Guide without a title communicate? It can be a little bit difficult for the casual viewer to see what this is about?

panel/widgets/tqdm_indicator.py
MarcSkovMadsen: Thanks using this and an extended relative import `..pane.markup` I was able to move `Tqdm` to the indicators file.

panel/widgets/tqdm_indicator.py
MarcSkovMadsen: Black :-)

examples/reference/indicators/Tqdm.ipynb
philippjfr: Reference Guides are primarily documentation so they should be optimized for rendering on the website which adds the title and a user that opens the notebook file can see the notebook title. For a screenshot or video it's easy enough to temporarily insert a title.

panel/widgets/tqdm_indicator.py
philippjfr: No 😄

MarcSkovMadsen:
FYI. @philippjfr I have implemented your suggestions. 

There are two minor, nice to have items (see todo at top) that I don't know how to fix. I don't think they are important. 

Let me know if there is more I should do. I don't plan to do more on this. I think its fine.

panel/widgets/tqdm_indicator.py
philippjfr: Well maybe eventually.

PR #2084, opened 2021-03-15
Custom Models - Make it easy please

MarcSkovMadsen:
The purpose of this PR is to make it easier for the Community to develop and contribute Custom Panel Models based on Bokeh Models. We do this by providing a Developer Guide Notebook. 

More could be added like a Template Repository that makes it very easy to develop, test and deploy to pypi+conda one or more Custom Models.

## Inspiration

Inspiration for developing such a guide can be found here

- [Extending Bokeh](https://docs.bokeh.org/en/latest/docs/user_guide/extensions.html)
- [Writing our own Dash Components](https://dash.plotly.com/plugins)
- [Custom IpyWidgets](https://ipywidgets.readthedocs.io/en/stable/examples/Widget%20Custom.html)
- [Authoring Custom Jupyter Widgets](https://blog.jupyter.org/authoring-custom-jupyter-widgets-2884a462e724)
- [Streamlit Component Guide](https://docs.streamlit.io/en/stable/streamlit_components.html)

## Must Do

[] Describe difference between widget and pane
[] Describe when to use `__javascript_raw__` and `__javascript__`. Similar for css.


philippjfr:
Merging this as something is better than nothing.

PR #2088, opened 2021-03-15
Add empty progress bar

### LINKED ISSUE #2080: Cannot give Progress initial look of value=0, max=100
Panel 0.11

I am trying to build the `Tqdm` indicator with an initial look of 

image

But the only initial look I can get is

image

This is produced by

```python
import panel as pn

progress = pn.widgets.indicators.Progress(active=False)
progress.servable()
```

## Background

I think the "non-active" look is not nice. And it is something that will catch the eye of the user instead of the button, plot or slider she should initially interact with.

## Hack

This code will give the desired initial look

```python
import panel as pn

progress = pn.widgets.indicators.Progress(value=1, max=1000000, active=False)
progress.servable()
```

## Solution

I don't know. Maybe it's just the hack.

Hoxbro:
Fixes #2080

By setting the value to -1 the progress bar is empty. 

```python
import panel as pn

progress1 = pn.widgets.indicators.Progress(value=-1, active=False)
progress2 = pn.widgets.indicators.Progress(value=-1, active=True)
progress3 = pn.widgets.indicators.Progress(value=0, active=False)
progress4 = pn.widgets.indicators.Progress(value=0, active=True)
progress = pn.Column(progress1, progress2, progress3, progress4)
progress.servable()
```

https://user-images.githubusercontent.com/19758978/111211562-d5b38000-85ce-11eb-9da4-e0ec5c179216.mp4

### DIFF
--- a/panel/tests/widgets/test_misc.py
+++ b/panel/tests/widgets/test_misc.py
@@ -41,7 +41,7 @@ def test_audio_url(document, comm):
 def test_progress_bounds():
     progress = Progress()
     progress.max = 200
-    assert progress.param.value.bounds == (0, 200)
+    assert progress.param.value.bounds == (-1, 200)
     progress.value = 120
     assert progress.value == 120
 
--- a/panel/widgets/indicators.py
+++ b/panel/widgets/indicators.py
@@ -122,10 +122,10 @@ class Progress(ValueIndicator):
 
     max = param.Integer(default=100, doc="The maximum value of the progress bar.")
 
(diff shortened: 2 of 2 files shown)


philippjfr:
Interesting, I guess I don't quite understand why this works.

Hoxbro:
I think this could be a case of "It's not a bug, it's a feature".

philippjfr:
Probably, the HTML spec can be...weird. Certainly would have expected it to treat `0` differently from `null`/unset.

Issue #1925, opened 2021-01-19
Terminal component


MarcSkovMadsen:
A Terminal Component for Panel would be nice to have.

![image](https://user-images.githubusercontent.com/42288570/104992394-3f753a80-5a21-11eb-9232-639b67bfcea9.png)

There are many terminal components that could be used. See for example

- [Top 5 Terminal Components](https://ourcodeworld.com/articles/read/292/top-5-best-terminal-console-emulator-plugins-for-javascript-and-jquery)
- [React Terminal Component](https://github.com/rohanchandra/react-terminal-component)
- [XTerm.js](https://xtermjs.org/)
- [TerminalJS](http://www.erikosterberg.com/terminaljs/)

@philippjfr stated he would like an XTerm implementation for Panel and asked me to create this FR based on discussions in https://github.com/holoviz/panel/issues/1914

MarcSkovMadsen:
## Related Issues

- https://github.com/holoviz/panel/issues/1978
- https://github.com/holoviz/panel/issues/1904

## Use Cases

- Capturing stdout/ stderr from your Panel Application
- Running processes in the terminal
    - For example I am trying to create a pipeline for my "daily routine operations". This includes running `kubectl` (kubernetes cli) in the command line. Some of the output I just need to inspect. Some of the output I need to respond to. Like saying yes/ no. Or opening a hyperlink to enter a username and password. Some output I need to transform into a color styled table that indicates where the problems are (if any).
- Creating true terminal experiences where you are connecting `bash`, `python` cli or `ipython cli` to the front end.
- Interactive Bots or Games.

philippjfr:
An xterm.js based Terminal widget has been merged.

PR #2090, opened 2021-03-16
Terminal Widget based on xtermjs

MarcSkovMadsen:
https://github.com/holoviz/panel/issues/1925

Implements a Terminal Component based on [xtermjs](https://github.com/xtermjs/xterm.js/)

## Must Do

[x] Make it work in Jupyter Hub (See image with issue in post below).
[x] Create Reference Guide Notebook
[x] Design the api
[x] Support input
[x] Support output including links, emojis and danish characters.
[x] Fix size issues (using add on [fit](https://github.com/xtermjs/xterm.js/tree/master/addons/xterm-addon-fit))
[x] Define `.js` and `.css` assets to include on bokeh model.
[x] Make it Work on Server
[x] Make it Work in notebook

# Nice to Have

[] Support running subprocesses like bash, python and ipython in pty
    - [] On windows (winpty?)
    - [] Improve on `add_periodic_callback`. Should this be seperate thread or other period length?
    - [x] On Linux
    - [x] Handle user exit from terminal.
[] Support Addons
    - [] [Attach](https://github.com/xtermjs/xterm.js/tree/master/addons/xterm-addon-attach)
    - [x] [fit](https://github.com/xtermjs/xterm.js/tree/master/addons/xterm-addon-fit)
    - [] [Search](https://github.com/xtermjs/xterm.js/tree/master/addons/xterm-addon-search)
    - [x] [Web-Links](https://github.com/xtermjs/xterm.js/tree/master/addons/xterm-addon-web-links)
[] Make `.fit` calculation in `.ts` model 100% bullet proof and also support `sizing_mode="stretch_height"`.
[] Determine whether to inherit from `StreamIO` or not. c.f. Philipp comment below.
[] Fix hacks for handling repeated input/ output
[] Enable copy-paste
[] Implement nice kill in notebook (see image in posts below).
[x] Example of using subprocess.Popen
[x] Example using streaming data from some longer running process
[x] Example of redirecting python log


MarcSkovMadsen:
I just added the basic "Button Extension" described in the [Custom Extension guide](https://github.com/holoviz/panel/pull/2084). It made is SO EASY to get started with a setup. It took a few minutes to get a basic custom extension working. 

Maybe a `panel create-extension` cli command doing the same could make it even more simple?

MarcSkovMadsen:
A minimum POC is there

https://user-images.githubusercontent.com/42288570/111262373-78e2b480-8624-11eb-8e10-92dc3fa8c7d1.mp4

MarcSkovMadsen:
@philippjfr . If you have good ideas about the api we want feel free to provide an example.

I will have to explore and learn because I have not working with terminals before. And not so much with redirecting logs, stdout etc. or using pty.

philippjfr:
It should probably just implement the file protocol, i.e. at minimum implement `write` but maybe the entire interface including `read`, `seek`, `readline`, `readlines` `readable`, `writable`, `seekable`, `tell`, `close`, `closeable`, `flushed` etc. Probably just start with `write` and then do:

```python
import sys

term = Terminal()
sys.stdout = term
sys.stderr = term
```

MarcSkovMadsen:
Got the basics working. The Terminal is now defined as

```python
class Terminal(StringIO, Widget):
    # Parameters to be mapped to Bokeh model properties
    value = param.String(doc="""
        User input from the terminal""")
    object = param.String(doc="""
        System output to the terminal""")
```

- I was in doubt. But I believe the `Terminal` is a widget with a `value` (~user input) and an `object` (output) parameter.
- I will have to learn whether to inherit from StringIO is a good thing or not. For now it works.
- For now I have defined the value and object as `String`. I don't know if it should have been `Bytes`. I have a feeling that it is needed for some uses cases. For example `term.write('\x1bc')` does not clear the screen as described here https://github.com/xtermjs/xterm.js/issues/950

philippjfr:
>I was in doubt. But I believe the Terminal is a widget with a value (~user input) and an object (output) parameter.

Definitely straddling the line but if it's accepting user input, it's a widget.

>I will have to learn whether to inherit from StringIO is a good thing or not. For now it works.

Don't think this is a good idea. Just satisfy the file-protocol.

>For now I have defined the value and object as String parameters. I don't know if it should have been Bytes. I have a feeling that it is needed for some uses cases. For example term.write('\x1bc') does not clear the screen as described here xtermjs/xterm.js#950

Not sure about separating the `value`/`object` naming. Also what does `object` return? The entire history? The last output?

panel/models/terminal.ts
```diff
@@ -0,0 +1,74 @@
...
+// See https://docs.bokeh.org/en/latest/docs/reference/core/properties.html
+import * as p from "@bokehjs/core/properties"
+import { div } from "@bokehjs/core/dom";
+
+
+// The view of the Bokeh extension/ HTML element
+// Here you can define how to render the model as well as react to model changes or View events.
+export class TerminalView extends HTMLBoxView {
+    model: Terminal
+    term: any // Element
```
Tyriar, line 13: Not sure about your build/module system but you might be able to do this to bring the type in without the implementation:

```ts
import type { Terminal } from 'xterm';
```

panel/tests/widgets/test_terminal.py
```diff
@@ -0,0 +1,60 @@
+import panel as pn
+import sys
+import subprocess
+import shlex
+
+pn.config.js_files["xtermjs"]="https://unpkg.com/xterm@4.11.0/lib/xterm.js"
+pn.config.css_files.append("https://unpkg.com/xterm@4.11.0/css/xterm.css")
```
philippjfr, line 7: Assume this is for testing? These should be declared on the model.

MarcSkovMadsen: Sure. One step at a time :-)

panel/models/terminal.ts
MarcSkovMadsen: Thanks. I will try it out.

MarcSkovMadsen:
The object returns the last output.

The value returns the last input from the user

MarcSkovMadsen:
Status: Output works, layout works (with a few hacks)
Next Step: User Input

Uploading panel_terminal3.mp4…

MarcSkovMadsen:
Got the `.js` and `.css` imports working in the notebook. And started writing the reference guide.

image

MarcSkovMadsen:
When "streaming to the terminal" in the notebook I experienced

MarcSkovMadsen:
When I kill the process in the notebook it looks like

image

MarcSkovMadsen:
Importing the `terminal` extension in Jupyterlab does not work

image

philippjfr:
I would seriously discourage displaying anything in the same cell where you are loading the extension.

philippjfr:
That seems to be a bug though, it should wait on bokeh to be loaded instead of erroring. But give it a few seconds before you display something to let Bokeh load. If that works then it's a distinct bug and you shouldn't worry about it.

MarcSkovMadsen:
I just used one cell to make it as "minimal" as possible. It's the same problem when using multiple cells @philippjfr 

image

MarcSkovMadsen:
When I don't import the `terminal` extension it works. 

image

The problem for me is that importing the terminal works in the notebook.

philippjfr:
I'll take a look when I get a chance.

MarcSkovMadsen:
Thanks @philippjfr . It's not urgent as I will just finalize the rest and then you can take a look as a part of the review/ rewrite you will be doing anyways. Thanks.

PR #2091, opened 2021-03-16
Redesign ReactiveHTML._scripts

philippjfr:
Fixes some issues raised by @xavArtley in https://github.com/holoviz/panel/pull/1894


xavArtley:
I have a new interesting bug:
This is not working
reactive_bug

But if I put a text with the value in the title it works correctly
reactive_ok

philippjfr:
The problem here is that `value` is incremented server side but also linked to the content of `button.children.innerHTML` on the client. Not sure how to resolve that to be quite honest. The fact that `Button ${value}` "works" is the actual bug here from my perspective.

philippjfr:
Okay, so now the two cases you described behave the same. But now we need to figure out a nice syntax to declare that we do not want the `${value}` to be synced bi-directionally.

philippjfr:
Okay so the series of events is:

1. Button click event is sent to python
2. `${clicks}` is extracted from the DOM children and sent as a model change event (`clicks = 0`)
3. `self.clicks += 1` increments clicks value (`clicks=1`)
4. ModelChange Event generated in 2. arrives (`clicks=0`)

philippjfr:
Only way of solving this would be to trigger syncing after the DOM is rendered.

philippjfr:
Perhaps it's worth making a distinction in `_dom_events`. Currently they serve a dual purpose, they exist to attach events to them but they also exist to trigger syncing of DOM properties.

philippjfr:
Tbh that's what inline callbacks are for, this works for example:

```python
class Button(ReactiveHTML):
    
    value = param.Integer(default=0)
    
    title = param.String()

    _html = '<button id="button" onclick="${_button_click}">${title} ${value}</button>'
    
    def _button_click(self, event):
        self.value = self.value + 1

button = Button(title='Blah')
button
```

philippjfr:
@xavArtley What would you think about renaming `_dom_events` to `_sync_events` or similar to make the purpose clearer? Then inline callbacks and explicit `on_event` callbacks can be reserved for real DOM event callbacks.

philippjfr:
Merging for now, can make decisions about `_dom_events` in a separate PR.

Issue #1975, opened 2021-02-11
(Calendar + clock) widget


jbednar:
It's not currently easy for a user to select a datetime value in Panel. Current options:

- **Datepicker** and **Dateslider**: convenient for selecting a date, but can't select a time
- **Datetimeinput**: returns a datetime, but not very friendly to the user; requires them to type a string without much feedback on what's accepted
- **Datetimeslider**: If #1972 is implemented, would make it easy to scroll through datetimes, and if https://github.com/bokeh/bokeh/issues/9943 is implemented would even make it feasible for a user to input a very specific datetime when needed, but would still not be that friendly for users who are thinking in terms of calendars and clocks

I think it would be valuable to have a datetime widget that is a drop-in replacement for the Datetimeinput widget, returning a single datetime64 value, but with user-friendly controls for people thinking of a specific time on a specific date: a calendar to select the date, and hh:mm:ss text boxes to select the time.  Obviously we can build such functionality already out of Bokeh widgets, but then you get date, hour, minute, and seconds values separately rather than a single datetime value useful for feeding to something like Pandas. Needing to select a datetime seems like a common and typical use case, and so I think there ought to be a straightforward drop-in widget like this to handle it.

Hoxbro:
I have been playing around with the existing Datepicker in panel, the bokeh model and flatpickr. flatpickr has a option to enable time, and I got a working code of it in action (maybe with some small hacks...)   

https://user-images.githubusercontent.com/19758978/111658738-e8b39380-880c-11eb-801c-fe24b14dab00.mp4

nghenzi:
this is awesome  !!!

I was using a JS library, but it would be nice if something like this is incorporated to panel

you can check the JS library here

https://discourse.holoviz.org/t/elegant-date-and-time-picker/1262/3?u=nghenzi2019

MarcSkovMadsen:
Awesome both you @hoxbro and the work

nghenzi:
flatpickr support a range date selection too. 

![image](https://user-images.githubusercontent.com/46707985/111695266-a9d40c80-8811-11eb-9e9c-a9e25a9bd5f9.png)

jbednar:
Both of those look fabulously useful!

philippjfr:
>I have been playing around with the existing Datepicker in panel, the bokeh model and flatpickr. flatpickr has a option to enable time, and I got a working code of it in action (maybe with some small hacks...)

Can you submit a PR? Would love to do this.

[...]

PR #2107, opened 2021-03-21
Adding a DatetimePicker widget

Hoxbro:
As mentioned in #1975 I have been working on a `DatetimePicker`. This is based on the existing `DatePicker` already in Panel and Bokeh, so a lot of the code is reused from it. 

I could not get the value to be datetime in the Bokeh model, so I went with a simple string to communicate with the model. I also tried to inherit a lot of the code from the existing `DatePicker` but in the end it was easier to not do it.

The following code can be used to start the widget. 
``` python

import panel as pn
pn.extension()


a = pn.widgets.DatetimePicker(
    name="Datetime2",
    start=date(2019, 2, 1),
    end=date(2022, 1, 1),
    value=datetime.today(),
    enable_time=True,
    enable_seconds=True,
    military_time=True,
).servable()
```
Example of the widget in action:
https://user-images.githubusercontent.com/19758978/111658738-e8b39380-880c-11eb-801c-fe24b14dab00.mp4

**TODO** 
- [ ] Adding documentation
- [ ] Adding unit test
- [ ] Properly some refactoring of the code
- [ ] Other things?


philippjfr:
This is awesome, thank you! Is there an easy way to refactor this in such a way that we can add a DatetimeRange widget without duplicating too much code?

jbednar:
Right; very often what I need to do is to choose a datetime range for which data to plot, so in practice I'm more often in need of a range than of a single datetime.

Hoxbro:
This is definitely possible. I have a small example with it. Although it is not finished yet as I have not implemented it on the python/panel side, but it works in the browser/bokeh model. It should be pretty straight forward to implement it on the panel side, but this will be another day.

I have attached a video showcasing it, where the console logs a change in the widget.

https://user-images.githubusercontent.com/19758978/111920516-bb2a4c80-8a8f-11eb-94db-63ac8edc956c.mp4

TvieiraB:
This looks great, can't wait to see it in Lumen.

philippjfr:
This is great, thank you @Hoxbro. Will merge and do any cleanup tasks in another PR if I find anything.

Issue #675, opened 2016-05-11
Allow editing slider values as text, at least for DynamicMap


jbednar:
For a bounded DynamicMap, e.g. this cell from the DynamicMap tutorial:

![image](https://cloud.githubusercontent.com/assets/1695496/15184207/de23dd68-1759-11e6-8ad2-2bbd93bbf214.png)

the slider values can be set only to the precision supported by the slider, which is a couple of digits when dragging the slider, and (for this example at least) a precision of 0.001 when using the left and right arrow keys to adjust the value more precisely.  It would be great if we could click in the text box and type an arbitrarily precise value, since the underlying object supports any such value.

Obviously, such a feature would require error checking, such as cropping the value to the bounds of the range, but that doesn't seem onerous to implement.  It makes the most sense for a DynamicMap, because it's the only way I can see to provide arbitrarily precise values, but it could also be implemented for a regular HoloMap by snapping the value to the nearest defined value.

MarcSkovMadsen:
I will close this one. The initial request of editing slider values have been solved with https://blog.holoviz.org/panel_0.12.0.html#New-widgets

For example we can now do

![image](https://user-images.githubusercontent.com/42288570/132971988-ca1ccae9-eba1-4412-9781-05de991a3f08.png)

```python
import param
import panel as pn
import holoviews as hv
import numpy as np

slider = pn.widgets.EditableFloatSlider(name='EditableFloatSlider', start=0, end=3.14)

def plot(value):
    xs = np.linspace(-5,5,100)
    return hv.Curve((xs,-(xs-2)**2))  * hv.Text(0, -30, str(value))
plot=pn.bind(plot, value=slider)

pn.Column(
    slider,
    pn.pane.HoloViews(hv.DynamicMap(plot))
).servable()
```

philippjfr:
Thanks @MarcSkovMadsen but that's not really what the request was. For this issue to be fully addressed the HoloViews pane would have to support this by default for float/int based widgets.

philippjfr:
This was implemented.

PR #2111, opened 2021-03-22
Add editable sliders

philippjfr:
```python
float_slider = pn.widgets.EditableFloatSlider(name='EditableFloatSlider', editable=False) 
int_slider = pn.widgets.EditableIntSlider(name='EditableIntSlider', end=10)
range_slider = pn.widgets.EditableRangeSlider(name='EditableRangeSlider', editable=(False, True))
pn.Column(float_slider, int_slider, range_slider)
```

Screen Shot 2021-03-22 at 3 30 22 PM

- [ ] Add tests
- [x] Add docs

### DIFF
--- a/examples/reference/widgets/EditableFloatSlider.ipynb
+++ b/examples/reference/widgets/EditableFloatSlider.ipynb
@@ -1,0 +1,123 @@
+{
+ "cells": [
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "import panel as pn\n",
+    "\n",
+    "pn.extension()"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "The ``EditableFloatSlider`` widget allows selecting selecting a numeric floating-point value within a set bounds using a slider and for more precise control offers an editable number input box.\n",
+    "\n",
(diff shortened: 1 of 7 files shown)


jbednar:
Thanks so much! Fixes https://github.com/holoviz/panel/issues/675 and (most of?) https://github.com/holoviz/panel/issues/1128 , plus can help address https://github.com/bokeh/bokeh/issues/9943 for Bokeh users willing to include Panel in their dependencies.

Is it possible to tweak the CSS involved here?  It would be great if we could have editable sliders that are approximately the same cost (in space and visual complexity) as the non-editable sliders, which seems doable to me. Can we have anything like the following options?

1. Simply removing the box around the spinner, which seems doable though a bit funny looking (and maybe no worse than before?): 

2. Reserving only a single digit for the numerical value, automatically expanding it (and shrinking the slider width to compensate) as the size expands (which seems unlikely to be doable when dragging the slider; likely to have the text overwrite the slider without doing major magic):

3. Getting rid of the spinner's box and moving it to where the slider value is normally shown, eliminating any border space taken by the spinner box:

jbednar:
For comparison, the non-editable versions would look roughly like this, i.e. the same as option 3 above but without any gray text and without the spinner widgets:

philippjfr:
Latest version after visual tweaks:

Screen Shot 2021-03-29 at 2 28 31 PM

jbednar:
Very cool, thanks!  Let's try this out for a while, and then maybe at some point consider whether to make it the default for a Param.Number mapping.

PR #2120, opened 2021-03-26
Update Param.ipynb

### LINKED ISSUE #2064: Doc: Explain the exact scenario where `watch=True` needs to be used in depends
In the documentation for Parameterized classes (https://panel.holoviz.org/user_guide/Param.html) the first place where `watch=True` is used is in the example with maps of countries and continents. 

It will be helpful to add a line or two here to say why `watch=True` needs to be used with some functions (`_update_countries`) and not some others (`view`). 

Is it because view is part of the panel, and so panel automatically makes sure the calls are made when parameters are updated? while `_update_countries` is never part of the panel, so it needs to be made to watch explicitly?

dhruvbalwada:
Added little more description of watch=True. Should close #2064.

### DIFF
--- a/examples/user_guide/Param.ipynb
+++ b/examples/user_guide/Param.ipynb
@@ -360,7 +360,7 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "Whenever the continent changes it will now eagerly execute the ``_update_countries`` method to change the list of countries that is displayed, which in turn triggers an update in the view method updating the map."
+    "Whenever the continent changes it will now eagerly execute the ``_update_countries`` method to change the list of countries that is displayed, which in turn triggers an update in the view method updating the map. Note that the ``watch=True`` only needs to be added to the decorators of methods that are not being passed into Panel's layout object (e.g. ``pn.Row``), to ensure that they are executed when the value of parameter changes."
    ]
   },
   {
--- a/examples/user_guide/Param.ipynb
+++ b/examples/user_guide/Param.ipynb
@@ -360,7 +360,7 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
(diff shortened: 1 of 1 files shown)


examples/user_guide/Param.ipynb
```diff
@@ -360,7 +360,7 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "Whenever the continent changes it will now eagerly execute the ``_update_countries`` method to change the list of countries that is displayed, which in turn triggers an update in the view method updating the map."
+    "Whenever the continent changes it will now eagerly execute the ``_update_countries`` method to change the list of countries that is displayed, which in turn triggers an update in the view method updating the map. Note that the ``watch=True`` only needs to be added to the decorators of methods that are not being passed into Panel's layout object (e.g. ``pn.Row``), to ensure that they are executed when the value of parameter changes."
```
jbednar, line 363:
```suggestion
    "Whenever the continent changes Param will now eagerly execute the ``_update_countries`` method to change the list of countries that is displayed, which in turn triggers an update in the view method updating the map. Note that there is no need to add ``watch=True`` to decorators of methods that are passed to a Panel layout (e.g. ``viewer.View`` being passed to ``pn.Row`` here), because Panel will already handle dependencies on those methods, executing the method automatically when the dependent parameters change. Indeed, if you specify ``watch=True`` for such a method, the method will get invoked twice each time a dependency changes (once by Param internally and once by Panel), so you should reserve ``watch=True`` only for methods that aren't otherwise being monitored for dependencies."
```

jbednar (commented):
Thanks; this is a good place to clarify watch=True. @philippjfr , can you please verify my suggested wording, and merge if you think it's ok?

examples/user_guide/Param.ipynb
philippjfr: Looks good

Issue #2116, opened 2021-03-24
Streaming to Tabulator table inside Template raises ValueError


MarcSkovMadsen:
panel: 0.11.1
os: windows

#### `Tabulator.stream` not working inside Template

If I click the "Stream" button I get a `ValueError` in the console

![image](https://user-images.githubusercontent.com/42288570/112262473-ff734380-8c6d-11eb-8375-0cd67916cc35.png)

```python
import numpy as np
import pandas as pd
import panel as pn
from panel.widgets import Tabulator

pn.config.sizing_mode="stretch_width"

DATA = pd.DataFrame(np.random.randn(10, 5), columns=list('ABCDE'))
tabulator = Tabulator(DATA)
stream = pn.widgets.Button(name="Stream")
def stream_data(self, *events):
    tabulator.stream(pd.DataFrame(np.random.randn(10, 5), columns=list('ABCDE')))
stream.on_click(stream_data)

template = pn.template.FastListTemplate(title="Tabulator")
template.main[:]=[tabulator]
template.sidebar[:]=[stream]
template.servable()
```

```bash
$ panel serve 'application\pages\tabulator_app\__init__.py' --auto --show
2021-03-24 06:54:32,518 Starting Bokeh server version 2.3.0 (running on Tornado 6.1)
2021-03-24 06:54:32,519 User authentication hooks NOT provided (default user enabled)
2021-03-24 06:54:32,521 Bokeh app running at: http://localhost:5006/__init__
2021-03-24 06:54:32,521 Starting Bokeh server with process id: 27352
2021-03-24 06:54:35,135 404 GET /favicon.ico (::1) 0.00ms
2021-03-24 06:54:35,145 WebSocket connection opened
2021-03-24 06:54:35,146 ServerConnection created   
2021-03-24 06:55:16,886 WebSocket connection closed: code=1001, reason=None
2021-03-24 06:55:19,622 WebSocket connection opened
2021-03-24 06:55:19,623 ServerConnection created
2021-03-24 06:55:36,672 Exception in callback functools.partial(<bound method IOLoop._discard_future_result of <tornado.platform.asyncio.AsyncIOMainLoop object at 0x000001C3914B2288>>, <Task finished coro=<_needs_document_lock.<locals>._needs_document_lock_wrapper() done, defined at c:\repos\private\awesome-panel\.venv\lib\site-packages\bokeh\server\session.py:51> exception=ValueError('Length of values (30) does not match length of index (20)')>)
[...]

PR #2134, opened 2021-03-31
Ensure changes on ReactiveData source are scheduled correctly

philippjfr:
Should never have applied model changes directly. This PR ensures all model changes are correctly scheduled.

Fixes https://github.com/holoviz/panel/issues/2116

### DIFF
--- a/panel/reactive.py
+++ b/panel/reactive.py
@@ -22,6 +22,7 @@ from tornado import gen
 from .config import config
 from .io.callbacks import PeriodicCallback
 from .io.model import hold
+from .io.notebook import push
 from .io.server import unlocked
 from .io.state import state
 from .models.reactive_html import (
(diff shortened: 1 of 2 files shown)


panel/reactive.py
```diff
@@ -1372,6 +1376,5 @@ def on_event(self, node, event, callback):
                              f"nodes include: {self._parser.nodes}.")
         self._event_callbacks[node][event].append(callback)
         events = self._get_events()
-        for ref, (m, _) in self._models.items():
-            m.events = events
-            push_on_root(ref)
+        for ref, (model, _) in self._models.items():
+            self._apply_updates([], {'events': events}, model, ref)
```
philippjfr, line 1380:
```suggestion
            self._apply_update([], {'events': events}, model, ref)
```

Issue #2136, opened 2021-04-01
jupyter_bokeh3.0  import error in ipywidget function


fleming79:
The jupyter_bokeh 3.0 release does not have BokehModel at the top level. So `panel.io.notebook.ipywidget` causes an import error when called.

```
  File "...lib\site-packages\panel\io\notebook.py", ... , in ipywidget
    from jupyter_bokeh import BokehModel
ImportError: cannot import name 'BokehModel' from 'jupyter_bokeh' (...\lib\site-packages\jupyter_bokeh\__init__.py)
```

changing to 
```
from jupyter_bokeh.widgets import BokehModel
```
resolves the import error.

Another less import issue in that function is that the line
```obj._cleanup(current[0])```
emits an error when cleaning up objects that don't have a `_cleanup` method such as for a panel Table.

**Comment by philippjfr** [comment]
Thanks for discovering that, would you submit a PR fixing the import?

>emits an error when cleaning up objects that don't have a _cleanup method such as for a panel Table.

That's odd, all Panel objects have a `_cleanup` method.

**Comment by fleming79** [comment]
The following code produces the error for me in JupyterLab:
```
import hvplot.pandas
import panel as pn
import pandas as pd
from ipywidgets import Box
b = Box([pn.ipywidget(pd.DataFrame({'a':[1,2,3,4]}).hvplot.table(height=100))])
display(b)
b.children=[]
```

**Comment by MarcSkovMadsen** [comment]
I just experienced this bug as well in vs code in a longer notebook

![image](https://user-images.githubusercontent.com/42288570/113488883-26791300-94c1-11eb-88e6-a318c1d215f6.png)

**Comment by MarcSkovMadsen** [comment]
After changing the Panel code to

![image](https://user-images.githubusercontent.com/42288570/113488999-cf277280-94c1-11eb-860f-7ca9bb57c265.png)

I get the below error in VS Code

![image](https://user-images.githubusercontent.com/42288570/113488982-bd45cf80-94c1-11eb-9c5a-ca4c22666266.png)

**Comment by fleming79** [comment]
I got the same bug but think it is something to do with VS code. This error is also emitted ([using code cells](https://code.visualstudio.com/docs/python/jupyter-support-py))
```
Error: Module @bokeh/jupyter_bokeh, semver range ^3.0.0 is not registered as a widget module 
```

MarcSkovMadsen:
FYI. @philippjfr . See VS Code bugs above.

PR #2138, opened 2021-04-01
Update notebook.py

fleming79:
Fix import error in ipywidget for jupyter bokeh  as mentioned in #2136. Also catches the error for obj missing _cleanup method. 
Note: This is my first attempt at a pull request so hope it is what you're after.

### DIFF
--- a/panel/io/notebook.py
+++ b/panel/io/notebook.py
@@ -346,7 +346,7 @@ def ipywidget(obj, **kwargs):
     -------
     Returns an ipywidget model which renders the Panel object.
     """
-    from jupyter_bokeh import BokehModel
+    from jupyter_bokeh.widgets   import BokehModel
     from ..pane import panel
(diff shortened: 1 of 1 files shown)


panel/io/notebook.py
```diff
@@ -355,7 +355,10 @@ def ipywidget(obj, **kwargs):
         def view_count_changed(change, current=[model]):
             new_model = None
             if change['old'] > 0 and change['new'] == 0 and current:
-                obj._cleanup(current[0])
+                try:
+                    obj._cleanup(current[0])
+                except:
```
philippjfr, line 360:
```suggestion
                except Exception:
```

philippjfr (changes requested):
One small fix, otherwise looks great.

philippjfr (approved):

philippjfr:
Thanks @fleming79! Much appreciated.

PR #2143, opened 2021-04-04
Updated json pane to accept single quote

### LINKED ISSUE #1798: Quote conversion in JSON pane breaks valid JSON content
The Javascript code for the JSON pane blindly replaces all single quotes with double quotes, presumably to convert Python single quoted strings to standard JSON double quoted strings. 

https://github.com/holoviz/panel/blob/3fdfa3248d1bd838fee88af6a277bf94eeaad87a/panel/models/json.ts#L17

However, it is perfectly legal to have single quotes inside of JSON string values, so this will turn correctly encoded JSON string into invalid JSON.

Screen Shot 2020-11-20 at 7 02 53 PM

The simplest fix would simply be to remove the quote conversion hack. Otherwise, you are going to need to be smart enough to only convert quotes that are not inside of strings.

Hoxbro:
Fixes #1798

Trying to use `literal_eval` for the `object` and then `json.dumps`. This PR also makes it possible to be more flexible with the `object`. 

The only thing I'm not completely sure about is if overlook something with the removal of `.replace(/(\r\n|\n|\r)/gm, "")`. 

```python
import panel as pn

json1 = pn.pane.JSON(object={"test": "can't show this"})
json2 = pn.pane.JSON(object=["can't show this"])
json3 = pn.pane.JSON(object="can't show this")
json4 = pn.pane.JSON(object="can show this")
json5 = pn.pane.JSON(object='{"b": 3}')
json6 = pn.pane.JSON(object="{'b': 3}")

pn.Column(json1, json2, json3, json4, json5, json6).servable()
```

| Without the PR | With the PR (`literal_eval`)|With the PR (`json.loads`)
|--------|-------|-------|
||||

### DIFF
--- a/panel/models/json.ts
+++ b/panel/models/json.ts
@@ -15,7 +15,7 @@ export class JSONView extends PanelMarkupView {
 
   render(): void {
     super.render();
-    const text = this.model.text
+    const text = this.model.text.replace(/(\r\n|\n|\r)/gm, "")
     let json;
     try {
       json = window.JSON.parse(text)
--- a/panel/models/json.ts
+++ b/panel/models/json.ts
@@ -15,7 +15,7 @@ export class JSONView extends PanelMarkupView {
 
   render(): void {
     super.render();
-    const text = this.model.text.replace(/(\r\n|\n|\r)/gm, "").replace("'", '"')
+    const text = this.model.text
     let json;
     try {
       json = window.JSON.parse(text)
--- a/panel/pane/markup.py
+++ b/panel/pane/markup.py
@@ -4,6 +4,7 @@ Markdown, and also regular strings.
 """
 import json
 import textwrap
+from ast import literal_eval
 
 from six import string_types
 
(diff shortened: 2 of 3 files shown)


panel/pane/markup.py
```diff
@@ -338,10 +339,11 @@ def applies(cls, obj, **params):
 
     def _get_properties(self):
         properties = super()._get_properties()
-        if isinstance(self.object, string_types):
-            text = self.object
-        else:
-            text = json.dumps(self.object or {}, cls=self.encoder)
+        try:
+            data = literal_eval(self.object)
```
Hoxbro, line 343: Should I use `json.loads` here instead of `literal_eval`?

philippjfr: `json.loads` probably makes more sense.

panel/models/json.ts
```diff
@@ -15,7 +15,7 @@ export class JSONView extends PanelMarkupView {
 
   render(): void {
     super.render();
-    const text = this.model.text.replace(/(\r\n|\n|\r)/gm, "").replace("'", '"')
```
philippjfr, line 18: Is there much harm in leaving this? Would be good to keep as much functionality in pure JS as possible.

Hoxbro: I have added back `.replace(/(\r\n|\n|\r)/gm, "")`. If I add back `.replace("'", '"')` it breaks the examples 1, 2, 3 and 6.

philippjfr: 👍

Issue #2142, opened 2021-04-03
Use of pytest.yield_fixture is deprecated


Hoxbro:
When running the unit tests the following warning is given:

``` log
 PytestDeprecationWarning: @pytest.yield_fixture is deprecated.
  Use @pytest.fixture instead; they are the same.
```

The warning conveniently states the fix.

I have labeled the issue as a good first issue, I hope this is ok.

PR #2144, opened 2021-04-05
Fix Issue #2142: Use of pytest.yield_fixture is deprecated

OBITORASU:
Replaced all instances of ```@pytest.yield_fixture``` which is deprecated with ```@pytest.fixture``` in this PR.

Fixes Issue #2142


philippjfr:
Thanks, very much appreciated!

PR #2154, opened 2021-04-08
Update Datatables Gallery Example

MarcSkovMadsen:
Updates the example based on user questions https://discourse.holoviz.org/t/adding-buttons-to-panel-datatable/2079

- js and css was updated and extended
- added example with export buttons

https://user-images.githubusercontent.com/42288570/114079747-daf0ab80-98aa-11eb-95a9-54065fb591ec.mp4


philippjfr:
Thanks @MarcSkovMadsen

PR #2164, opened 2021-04-12
Add support for bokeh theme

brl0:


panel/pane/plot.py
```diff
@@ -46,6 +47,12 @@ class Bokeh(PaneBase):
     Bokeh panes allow including any Bokeh model in a panel.
     """
 
+    theme = param.ClassSelector(default=None,
+                                class_=(Theme, str),
+                                allow_None=True,
+                                doc="""
```
philippjfr, line 53: While not as explicit, `default=None` already forces `allow_None=True`:

```suggestion
    theme = param.ClassSelector(default=None, class_=(Theme, str), doc="""
```

philippjfr:
Thanks, small comment, otherwise happy with this. If you could also update the documentation in `examples/references/panes/Bokeh.ipynb` that would be super but I won't let it hold up merging.

Issue #2147, opened 2021-04-08
Events support for Panel Echarts


JackDapid:
As a big fan of https://awesome-panel.org/echarts by @MarcSkovMadsen and https://panel.holoviz.org/reference/panes/ECharts.html I am using echarts a lot and got frustrated that they seemed to have dropped events / JS callback support from pyecharts? - see https://github.com/pyecharts/pyecharts/issues/1671
Without 'interactivity" the nice echarts are only half useful.

Clearly forking an pyecharts with bringing in that support is a n option, but I am looking for a discussion if there are some other ideas that could benefit other Panel echarts users.

MarcSkovMadsen:
One thing is that you could turn this into a feature request for events support for the Panel Echarts pane :-)

## Additional Context

Regarding the api for adding support for events inspiration could be drawn from the [panel-highcharts](https://github.com/marcskovmadsen/panel-highcharts) package I recently made. There are lots of event but I setup a "clever" mechanism to subscribe to the ones I need.

For example for the [Network example](https://mybinder.org/v2/gh/marcskovmadsen/panel-highcharts/HEAD?urlpath=lab/tree/examples/Network.ipynb) you subscribe using

![image](https://user-images.githubusercontent.com/42288570/114022669-4cf8ce80-9872-11eb-97f6-c110b8787186.png)

and get something like shown here

https://twitter.com/i/status/1378418244978106374

JackDapid:
Hi @MarcSkovMadsen thank you. panel-highcharts is really nice, I didn't see it on awesome-panel webpage before, but learned now to follow you on twitter for such news :-). 
Ok I see I was misinterpreting: https://github.com/holoviz/panel/blob/cc1548550b6375f69d0c9bc7e58207c0dd2af84c/panel/pane/echarts.py#L18
that Panel Echarts pane is using pyecharts, but now (awake!) I see it is not and I really like your subscription mechanism. WIll have some time hopefully over the weekend to look in details to panel-highcharts  and Panel Echarts pane to see if I can build a minimal demo following the mechanism.

JackDapid:
I checked panel-highcharts - really nice @MarcSkovMadsen - can't use it for the current project, but sure I will find a good use for one of the next projects.

For events in panel.Echarts happy to try to implement something similar, but I am a liitle stuck with panel model logic (see below), so currently I use a workaround inspired by @nghenzi from: https://github.com/holoviz/panel/issues/1883

## Workaround

[...]

PR #2174, opened 2021-04-14
Adding Echarts events

JackDapid:
Adding Echarts events support to Echarts modal #2147  
- create initial JS events - done
- min. Echarts test app - done
- next step: sending back params to python
- next step: improved tests (from app to autom. test as in tests folder)
- next step: some doc


panel/tests/test_echarts.py
```diff
@@ -0,0 +1,70 @@
...
+    "responsive": True
+}
+
+event_config = {
+    # 'click': 'series.tree', # Tests for Echarts event names with query
+    # 'click': None, # Tests for Echarts event names without query
+    # 'click': {'name': 'Child Clickable'}, # Tests for Echarts event names without query
+    # 'click': {'query': 'series.tree', 'base_url': 'https://www.TEST.de/AssetDetail.aspx?AssetId=',
+    #     'identifier': 'id'}, # Tests for new browser tab Echarts event
+    'click': {'query': 'series.tree', 'handler': 'e => console.log("I got an event:", e)'}, # Tests for handler Echarts event
```
MarcSkovMadsen, line 47: The js `handler` is very clever and useful. It is something that could be valuable in general every time a JS library is wrapped in Panel. It would be nice to have a common "api" or naming convention for that. So I would suggest @philippjfr also gives his comments and suggestions on this.

And now you use it to handle events. In my implementation of `panel-highcharts` I have a convention for specifying js functions as the values of the json/ dict that constructs the chart. That convention is that a value that looks like `'function ( ) { SCRIPT }'` is a javascript function and will be converted to a such before being passed to the constructor of the chart. A similar cross panel convention would be nice to have.

JackDapid: +1 and as you guys know much better how this fits in best to the overall logic and convention happy to implement as soon you agreed :-)

JackDapid:
@MarcSkovMadsen & @philippjfr I just pushed a first 'fully' functional version of echarts events - creating JS events as well as getting the subscribed events back as python params works - missing points:

- agreement of the event-config convention, e.g. `'function ( ) { SCRIPT }'` together with query
- more thinking and agreement of event filtering as different echarts series return useful param info in different ways
- understanding why pytest fail - I started that process
- proposing some doc
- implementing a pytest for events

JackDapid:
I added custom select and filters:
click': {'query': 'series.tree', 'select': 'data'}, # Tests for Echarts event with custom select
'click': {'query': 'series.tree', 'filters': ['children']}, # Tests for Echarts event with custom filters
'click': {'query': 'series.tree', 'select': 'data', 'filters': ['children']}, # Tests for Echarts event with custom select and filters

JackDapid:
not sure I understand the signature pytest errors, some help is appreciated.

One problem with event readonly?:
`FAILED panel/tests/pane/test_base.py::test_pane_clone[ECharts] - TypeError: Read-only parameter 'event' cannot be modified` under investigation.

MarcSkovMadsen:
Hi @JackDapid 

My argument for having `event` `readonly` is that if you show the parameter in a notebook or app then it is nice that the `TextInput` is disabled. I think @philippjfr is the one who can answer if `readonly=True` should be removed during testing or something else should be done.

JackDapid:
Thanks @MarcSkovMadsen und Danke @philippjfr 

yes, that was also my reason and happy to either leave it out or propose a modification of the test_pane_clone to leave out read-onlies. Any insight on the signatures issues?, looks to me like something general effecting the count of all model params, but will start to look into more details now, but on a first look I haven't found the same to be an issue in other PRs.

For a more general discussion on how I (and we @gluonnet, adding @kjp) could contribute to Panel I would like to point you to a PR (Press Release and not Pull Request :-)) http://www.caakz.com/aviation-administration-of-kazakhstan-launches-digital-regulator-sunflower/ and video, where we showcase some of our heavy use of Holoviews, Panel. ....
https://youtu.be/tUaAj3u_vvA?t=952
If intersting for you I would love a discussion with both of you @philippjfr and @MarcSkovMadsen how we can contribute to Panel?, awesome-panel?, ...?, maybe showcase some use-cases .... We were 'dragged' into Panel by awesome-panel, use our own modified react template, ... so only through this PR (Pull Request again ;-) ) realised the @philippjfr Holoviews, Datashader to @MarcSkovMadsen Panel-awesome connection :-)

philippjfr:
I'd really like to provide some detailed feedback here because the functionality is definitely great but I'm missing a bit of context. A lot of that could be answered by some documentation that describes the scope and implementation of the event system.

So for now a couple of general points:

# Events vs Parameters

I would not require coupling events to parameters, we should leverage Bokeh events more and provide ways for users to subscribe to those using an `on_event` method. This is more general tasks as in theory a lot of Panel components could allow subscribing to such events. The new `ReactiveHTML` component does this by providing the `on_event` method which allows users to subscribe to arbitrary DOM events on a specific DOM node.
 
By using a real event system rather than abusing parameter/properties for this we won't have to merge all the events into a single event property and eliminate the need to name specific events.

# A real specification

I'd love of the events proposed here, I see things like `query`, `select` and `filters` in the event config but don't quite know what they mean so if you could write that up that would be very helpful. Also instead of a `handler` for JS callbacks I'd prefer adding a separate `js_on_event` method which allows registering custom callbacks, which we can then internally wrap with bokeh `CustomJS` calls so they can reference other Bokeh models.

Happy to expand on any of this, but some references that might help to see what I'm talking about are the Python implementation of `on_event` (https://github.com/holoviz/panel/blob/master/panel/reactive.py#L1362), the definition of the custom Bokeh `DOMEvent` (https://github.com/holoviz/panel/blob/master/panel/models/reactive_html.py#L144), the DOM event listener setup (https://github.com/holoviz/panel/blob/master/panel/models/reactive_html.ts#L333) and the dispatching of the Bokeh events (https://github.com/holoviz/panel/blob/master/panel/models/reactive_html.ts#L183)

MarcSkovMadsen:
Hi @JackDapid 

I'm just blown away by the presentation. You have really put together a strong application based on HoloViz and Panel. 

I am also really interested in connecting and discussing.

My guess is that you both a lot of Ideas on where Panel should be improved, you already have some improvements that could improve Panel in general and showcasing and explaining some of your work would be truly amazing.

I'm including @nghenzi here who created the React Template. @nghenzi checkout the videos in the links above to see how your Template has been taken to the next level.

philippjfr:
Agreed, absolutely incredible work on that application. Thanks for sharing and as @MarcSkovMadsen indicated we'd be very grateful for your feedback.

JackDapid:
@philippjfr and @MarcSkovMadsen - not forgotten, just delivery deadline this Friday and will come back to it either later this week or during the weekend - agree to almost all written looking forward to next steps

JackDapid:
@philippjfr here some context for the events. The original idea was 'just' to get some client side JS event handling for echarts, so for example clicking on a node in the echarts - while discussing with @MarcSkovMadsen it turned out that we could easily add both server side and client side event support for echarts, which we than implemented - not aiming at a general Panel events support. But open for ideas, clearly. 
So currently possible to subscribe to events (selection possible by query, filter and select settings - which for sure need improved doc and also some examples) and all that for echarts only. Does that help with the context or did I misunderstood the question?

philippjfr:
I'm reviving this, but do have one question. Are the query and filter fields in the event config really needed? Sending these events is not expensive so I'd rather have a simpler configuration and make the whole event available in Python.

philippjfr:
pre-commit.ci autofix

philippjfr:
@JackDapid Will merge shortly but if you have suggestions or issues with the approach please file an issue.

Issue #1500, opened 2020-07-27
Add Binder links to all examples and user guides


philippjfr:
As the title says, all pages built from notebooks should have links to binder.

ceball:
Should be supported now by nbsite - https://github.com/pyviz-dev/nbsite/pull/188

However, might look ugly or take up too much vertical space - haven't seen how it looks yet with our theme. Will be happy to make that better when we see an example...

MarcSkovMadsen:
Where can I check this out @ceball . I tried to look at references example at for example https://pyviz-dev.github.io/panel/reference/widgets/DatetimeInput.html#widgets-gallery-datetimeinput. But they don't have any links.

ceball:
@MarcSkovMadsen the website build hasn't been updated to use this new feature of nbsite yet. I'll have a go at that now (no promises...I haven't built the panel docs myself for 3 years or so :) ).

ceball:
Currently looks like this:

![Screenshot from 2020-10-21 17-35-16](https://user-images.githubusercontent.com/1929/96750574-fda60b00-13c3-11eb-8cb2-0feaf950b338.png)

It's a sphinx note (https://docutils.sourceforge.io/docs/ref/rst/directives.html#note). I think we need to style it better and remove some text!

Some thoughts/questions...

Is the button well known enough that no text at all is required?

I think a binder link might not be useful on every notebook (though if just a small button, doesn't matter...although if just a small button, will people find it on notebooks where it does matter...).

I also think it might be worth combining with the "download from github" link.

And might be worth adding binder badge to notes about non-interactive notebooks:

![Screenshot from 2020-10-21 17-41-59](https://user-images.githubusercontent.com/1929/96751302-ec113300-13c4-11eb-92d6-b44d3b9929ed.png)

ceball:
For comparison, here's how dask does it:

![Screenshot from 2020-10-21 17-38-36](https://user-images.githubusercontent.com/1929/96751600-4dd19d00-13c5-11eb-8f20-5bc640cdf1ae.png)

jbednar:
The Dask way to to it looks fine, though I don't know if it needs the bar saying "Live Notebook" at all.

philippjfr:
Agreed, the dask one looks fine.

ceball:
As far as nbsite goes, it's the same as dask, except "note" (nbsite) vs "admonition" (dask):

dask:
```
.. admonition:: Live Notebook
[...]

Issue #1369, opened 2020-05-26
Suggestions for improving the website


maximlt:
Hi,

I've spent quite a lot of time on the website and I  wanted to suggest some potential improvements:
-  [ ] Change the color of the text in the search bar. It's green right now and is difficult to read.
![image](https://user-images.githubusercontent.com/35924738/82906036-effb6f00-9f64-11ea-8254-c16c2fa0547f.png)
- [ ] Move the search bar to a place where it's easier to find and make it bigger
![image](https://user-images.githubusercontent.com/35924738/82906886-36050280-9f66-11ea-82f1-eceee2f11fe5.png)
- [ ] Replace static images by GIFs on the index page to demonstrate the power of Panel and the kind of interactivity you can get (applies to the readme too)
![image](https://user-images.githubusercontent.com/35924738/82906949-47e6a580-9f66-11ea-8bc3-cfdd9fecd66c.png)
- [ ] Headers and sub-headers should be permalink for easier referencing to specific content in a page
![image](https://user-images.githubusercontent.com/35924738/82907179-8ed49b00-9f66-11ea-8bba-0a32916087c2.png)
- [ ] Headers and sub-headers should appear in a table of content always available to the user to more easily browse through the content of a page
- [ ] There should be a link to a **binder** environment on every notebook page
- [ ] The Reference Gallery should be above the Gallery in the table of content (I use it a lot more)
- [ ] The Reference Gallery should have a table of content at the top to quickly find a widget/layout/pane (it's pretty hard now, specially for the widgets, I usually search through the page)
- [ ] The first line describing a widget/layout/pane in the reference gallery should be highlighted in its own box. It's the most important piece of information of the page.
![image](https://user-images.githubusercontent.com/35924738/82907900-83ce3a80-9f67-11ea-9582-b1ede89c4016.png)
- [ ] The API reference manual isn't very useful as of now, clicking on a subpackage removes completely the API table of content which is then difficult to explore
![image](https://user-images.githubusercontent.com/35924738/82908489-57ff8480-9f68-11ea-8cd0-6eab26328757.png)
- [ ] It doesn't look very good when set to a small size (same as on mobile if I'm not mistaken)
![image](https://user-images.githubusercontent.com/35924738/82908651-9137f480-9f68-11ea-9b78-6cab3162054b.png)

These are just suggestions that I believe would make the website look better and easier to use. Feel free to make some more suggestions that I would then add to this list.

maximlt:
[...]

PR #2198, opened 2021-04-17
Enable Binder

MarcSkovMadsen:
Address https://github.com/holoviz/panel/issues/1500, https://github.com/holoviz/panel/issues/1369#issuecomment-635944381 and https://github.com/holoviz/panel/issues/2218. 

This will enable opening the Panel examples on Binder. Whether it will also be possible to add a binder link to all examples is unknown and "nice to have" for this PR.

image

For now you can access binder via https://mybinder.org/v2/gh/holoviz/panel/binder?urlpath=lab/tree/examples

I would like a suggestion from @philippjfr or @jbednar which link we should provide when released. I see the following options

- Link to `binder` branch. 
    - Pros: We can control what is available on Binder
    - Cons: We will have to remember and spend time on updating the `binder` branch
- Link to `master` branch.
    - Pros: Everything is always the latest
    - Cons: There might be things is master that changes the behaviour of the current version of Panel released.
- Link to *release branch* of Panel. I.e. for example the version for the Panel 0.11.3 release
    - Pros: The current version of Panel is available on Binder
    - Cons:
        - Will have to remember to update links across the repository.
        - Users having bookmarked the link would stay on that version of Panel on Binder


MarcSkovMadsen:
Status: Things are starting to work. I'm finding bugs in the Reference Examples. I have to update the gallery notebooks in order to be able to serve them as apps. I would like to configure VS Code to enable showcasing how powerful the combination of Panel and VS Code is.... Also on Binder and in a Jupyter Hub.

https://user-images.githubusercontent.com/42288570/115115998-a9a56900-9f97-11eb-9aa7-41bedbb17a21.mp4

So I would really, really like Panel to work in VS Code. But it requires fixing https://github.com/bokeh/jupyter_bokeh/issues/131

MarcSkovMadsen:
Hi @philippjfr 

As I see it 1) Binder is working 2) It now comes with a nice setup like `panel-jupyter-proxy`, `panel serve` of all gallery examples, vs code integrated and a new, improved index page.

In principle this could be all for this PR except adding a link to binder in different places. For example the README.

BUT all the notebooks need to be walked through and prettified in order to be useable as apps. This could be a lot of work and a lot of minor changes.

Here are my questions.

- Do you want the **gallery notebooks** updated in general to be responsive and end up in a nice app template?
- Do you want the **reference examples** updated in general to be responsive and end up in a nice app template?`
- If yes to some of these questions. Do you want seperate PRs for each notebook, one large Notebook PR or just this PR with everything?
- Could you answer the question about which link to provide to users posted in the first post? Thanks.

philippjfr:
>Do you want the gallery notebooks updated in general to be responsive and end up in a nice app template?

No, some of them sure, but definitely not all. The focus should be the feature being demonstrated not making all the examples look like.

>Do you want the reference examples updated in general to be responsive and end up in a nice app template?`

Definitely not, in fact I'd argue there's no reason to make these available as apps at all.

>Could you answer the question about which link to provide to users posted in the first post? Thanks.

I'll make sure to configure this once you're done here. The link will go in the interactivity warning that floats on the right side.

philippjfr:
>If yes to some of these questions. Do you want separate PRs for each notebook, one large Notebook PR or just this PR with everything?

Just make a single PR for any coordinated changes like this.

MarcSkovMadsen:
Thanks. @philippjfr 

I think it is OK that the Reference Guides do not end up as apps. But for the Gallery Examples I would argue that they should end up as a nice app. I assume users would like to end up with something that looks nice and can be distributed as an end product. And it would help communicate that it is now possible and easy to do now with Panel.

**If not how would I determine which examples it makes sense to serve as apps?**

----------------------

Regarding the link. It should be in more places than where it says *download notebook*. It should be on the github README, it should be on the front page of the Panel web site. Probably even with some gif/ video illustrating what is possible.

To me binder access is a game changer. All the users who think Panel is not working due to the static nature of the documentation now has a quick route to actually test out Panel and try to work with it. And the integrated app means you can actually try out the end product. Not just learn about how to build some of the smaller components.

You can arrange workshops, hackathons etc. without the hazzle of having users configure stuff on their local laptop. In conferences that is always a hazzle. etc. etc.

philippjfr:
>But for the Gallery Examples I would argue that they should end up as a nice app. I assume users would like to end up with something that looks nice and can be distributed as an end product.

We should definitely have many more nice polished apps up there but many apps up there specifically focus on a relatively small feature and may just be a few lines of code. In those cases adding a template around it would be quite distracting.

>Regarding the link. It should be in more places than where it says download notebook. It should be on the github README, it should be on the front page of the Panel web site. Probably even with some gif/ video illustrating what is possible.

Agreed, it should be in as many places as possible.

MarcSkovMadsen:
Ok. My final attempt to convince you it is not very disturbing but actually quite powerful and an eye opener. :-)

Lets take `stocks_altair.ipynb` as an example. The only changes are

1. An extra, very small App section

image

2. A tiny change to the code before the App section to make plots responsive.

I think that is a very small price to pay compared to providing users with a large gallery of apps they can really get inspiration from and get and understanding of the look and feel. And the gallery provides search functionality as well. They can search for `altair` and find it no matter if it is a small or large example.

https://user-images.githubusercontent.com/42288570/115422401-29803d00-a1fd-11eb-991e-89bfac3140bf.mp4

philippjfr:
Okay, you've convinced me 😆

MarcSkovMadsen:
Thanks for listening @philippjfr . You won't regret it. Except for all the bugs I will find when trying to support and app also 😄

MarcSkovMadsen:
Final question @philippjfr . A lot of the examples plots that have a fixed and **smal** width. Is that on purpose? What is the reason for not making them wider or responsive?

image

philippjfr:
>A lot of the examples plots have a fixed and small width. Is that on purpose? What is the reason for not making them wider or responsive?

In many cases they are just copy-pasted from the documentation of that project.

jbednar:
> Do you want the gallery notebooks updated in general to be responsive and end up in a nice app template?

I would want things to be responsive by default regardless of whether they are in an explicit app template. I'd want our examples to show good practice, so that each of them are a starting point that's not shooting themselves in the foot. People should't copy from things we know are bad, and if we put a lot of content up that isn't the best way to do things, it's a disservice.

But is it really necessary to include the pn.Template line? It's fairly distracting; can we not have a standard template always accesible (a la Streamlit) or just responsive behavior by default?

examples/gallery/simple/clifford_interact.ipynb
```diff
@@ -15,7 +15,8 @@
     "from datashader import transfer_functions as tf\n",
     "from colorcet import palette_n\n",
     "\n",
-    "pn.extension()"
+    "pn.extension()\n",
+    "pn.config.sizing_mode=\"stretch_width\""
```
philippjfr, line 19: It's a matter of preference but config options can be passed to the extension:

```suggestion
    "pn.extension(sizing_mode='stretch_width')"
```

MarcSkovMadsen: Thanks. I like that preference :-) I will update.

MarcSkovMadsen:
@jbednar . Some things the "default" template would have to know is a lot of configuration.

- Which template?
- Title?
- What goes into sidebar?
- What goes into main?
- How wide should main be?

etc. I don't know how that could be done automatic and removed from the notebooks.

I would actually argue that 1) In practice it is not distracting 2) It's actually showing what the workflow to get to "production" with your dashboard or app is. For many use cases eventually you would need to wrap it into a nice template. 3) It's really valuable as now we have a to share the results of the examples as something that really looks nice and communicates.

So I don't see it as a pain. But a gain :-)

jbednar:
>  Which template?
>  Title?
>  What goes into sidebar?
>  What goes into main?
>  How wide should main be?

Here I'd suggest following what existing less-configurable tools like shiny and streamlit do, to give users a gentle introduction to making a Panel app that is easy to get into and may be all they every need to do.  E.g. `pn.sidebar(panel_object)`, `pn.main(panel_object`), `pn.title(text)`, etc., unless someone has a better suggestion. Defaults like how big main should be have to be configurable in a template anyway, so those don't seem like they would affect whether there is a default template.

philippjfr:
>Here I'd suggest following what existing less-configurable tools like shiny and streamlit do, to give users a gentle introduction to making a Panel app that is easy to get into and may be all they every need to do. E.g. pn.sidebar(panel_object), pn.main(panel_object), pn.title(text), etc., unless someone has a better suggestion

Really not excited about this, I don't want yet another API. I'd be okay with something like:

```python
pn.extension(template='material')

a_component.servable(area='header')
another_component.servable(area='main')
yet_another_component.servable(area='sidebar')
```

jbednar:
That sounds pretty clean to me!

MarcSkovMadsen:
Status:

- All gallery examples wrapped in a nice template and is running as web apps on Binder.

The only things outstanding is 

- Enable customizing the site_url of A Template. Currently clicking the site url in an app on binder sends you to `https://hub.gke2.mybinder.org` instead of `https://hub.gke2.mybinder.org/user/holoviz-panel-4gfpeow4/panel`. The user should be able to set this as what is needed depends on the deployment scenario.

- Getting VS Code working again. C.f. https://github.com/bokeh/jupyter_bokeh/issues/131. I cannot do this but @philippjfr said he would do it.

More can always be done. But that should be done in other PRs I believe.

examples/gallery/simple/clifford_interact.ipynb
MarcSkovMadsen: @philippjfr . I have updated all Panel notebooks and test examples to use this terminology. It's nice and simple.

MarcSkovMadsen:
Binder is **now working** as I originally intended it to.

- You can open jupyter lab in Binder via https://mybinder.org/v2/gh/holoviz/panel/binder?urlpath=lab/tree/examples
- You can open the panel app gallery in Binder via https://mybinder.org/v2/gh/holoviz/panel/binder?urlpath=panel
    - Panel has an improved index/ gallery page
    - You can click the `site_title` in an app and navigate back to the gallery.
    - Apps are "responsive"
- You can open VS Code on Binder as well.
- Everything is installed and configured to work with all gallery and reference notebooks.
    - Gallery and reference notebooks work. Except a few bugs below.

Some things are not in scope for me

- Adding links from every example at panel.holoviz.org to the notebook and app on Binder. I hope @philippjfr can/ will do this.

Some things are not working and I am not capable of getting it working

- C.f. https://github.com/bokeh/jupyter_bokeh/issues/131
- Serving notebooks as live "docs" with static code cells and dynamic output. For example via Voila. See https://github.com/holoviz/panel/issues/2241.
- And lots more that I created individual bug reports for.

## Working Video

https://user-images.githubusercontent.com/42288570/116806896-45a2a900-ab30-11eb-9ab5-8aa36a745f6e.mp4

## Not Working

Watch this video on Youtube https://www.youtube.com/watch?v=gI3qQRVdIIE

@philippjfr Feel free to review and start creating value from this :-) I don't plan to do more on this before getting a review and feedback.

Thanks.

binder/start
```diff
@@ -0,0 +1,14 @@
...
+echo 'BasicTemplate.param.site_url.default="./"' >> panel/template/base.py
+export DISPLAY=:99.0
+export PYVISTA_OFF_SCREEN=true
+export PYVISTA_USE_IPYVTK=true
+export PYVISTA_PLOT_THEME=document
+which Xvfb
+Xvfb :99 -screen 0 1280x1024x24 > /dev/null 2>&1 &
+sleep 3
+jupyter trust examples/**/*.ipynb
+jupyter trust examples/**/**/*.ipynb
```
jbednar, line 13: I'm not a bash expert, but doesn't `**` indicate recursive directory searching nowadays? If so, isn't the second trust call redundant?

examples/gallery/apis/stocks_altair.ipynb
```diff
@@ -165,14 +165,51 @@
...
     "In practice, different projects will be suited to one or the other of these APIs, and most of Panel's functionality should be available from any API."
    ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## App\n",
+    "\n",
+    "Lets wrap it into nice template that can be served via `panel serve stocks_altair.ipynb`"
```
jbednar, line 175:
```suggestion
    "Let's wrap it into nice template that can be served via `panel serve stocks_altair.ipynb`"
```

jbednar (commented):
Nice! This will make things look and feel much better. Should we promote the approach from  https://github.com/holoviz/panel/pull/2271 or is it better to put the template separately like it is here?

MarcSkovMadsen:
> Nice! This will make things look and feel much better. Should we promote the approach from #2271 or is it better to put the template separately like it is here?

When the new approach #2271 is ready and tested I think it would make a lot of sense to switch. That syntax is nice and shorter.

The only downside is that then not a lot of examples will show how to build apps with templates for non-notebook users. There the natural workflow might still be to explicitly declare a template ??

jbednar:
Good point. We should still have an explicit example of building such a template explicitly somewhere, and while it's true people may not see that example, having people use servable unnecessarily in a standalone file is not that terrible. I.e., I agree it doesn't make as much sense outside a notebook, but it's still going to be fine for many people, and so I'm ok with more advanced non-notebook users having to do a bit more work to find that there is this other way.

setup.py
```diff
@@ -149,7 +149,12 @@ def run(self):
         'ipywidgets_bokeh',
         'ipyvolume',
         'ipyleaflet',
-        'xarray <0.17'
+        'xarray <0.17',
+        'pydeck',
+        'pyecharts',
+        'idom',
+        'xlsxwriter',
+        'pyarrow',
```
philippjfr, line 157: Most of these cannot be added to setup.py since that is also the ground truth for conda installs. If I remove these here will this break the binder setup?

philippjfr, line 157:
```suggestion
```

binder/start
MarcSkovMadsen: I thought so. But it did not work like that for me. So I added it twice. That works.

setup.py
MarcSkovMadsen: Yes. But we can add them to one of the files in the binder subfolder. Can you point out which should not be included?

philippjfr: Already removed all the ones you added. Anything that is not conda installable cannot appear here, so for now I've removed pydeck, pyecharts, idom, xlsxwriter and pyarrow. You could readd pyarrow since that is available.

MarcSkovMadsen:
There is one fundamental thing to consider. Which branch should be promoted and shared with users? In principle any branch will be available on Binder. But it's important to agree on one that is promoted.

My recommendation would be to promote the `binder` branch. The it can be updated and fixed whenever there is a new release. The `master` branch for example will not always be working is my experience. For example the `.ts` code cannot build right now I believe.

philippjfr:
>For example the .ts code cannot build right now I believe.

There are a few errors but they are not preventing the .ts from being built and don't cause any real issues. However I should add a CI check to ensure there are zero TS errors or warnings. Not entirely excited about a binder branch since that's an additional release step but given all the alternatives that still seems like the best approach.

MarcSkovMadsen:
The alternative is just using the `master` branch. But then users would see functionality that is not yet released. That might be confusing. But maybe also just fine.

jbednar:
Sound to me like it ought to be using a tag indicating the latest release, rather than a separate branch?  Our release workflow would then need to update `latest-release` or `released` whenever we make a full (non-dev) release.  Seeing changes post-release would not only be confusing, it would likely not work, if it's anything like when I try Panel outside the release process (due to the JS build/publish process involved).

panel/tests/widgets/test_tables.py
```diff
@@ -5,7 +5,7 @@
 
 try:
     import pandas as pd
-    from pandas.util.testing import (
+    from pandas.testing import (
```
philippjfr, line 8:
```suggestion
    from pandas._testing import (
```

philippjfr:
Thanks again for this extraordinary effort @MarcSkovMadsen. I'll be adding binder links to all pages in a separate PR.

PR #2208, opened 2021-04-18
Fix of value_throttled in editable sliders

### LINKED ISSUE #2204: EditableFloatSlider, EditableIntSlider, EditableRangeSlider in reference example working strangely and raising errors
I'm on the current master branch of Panel and the FloatSlider is not working as expected.

https://mybinder.org/v2/gh/holoviz/panel/binder?urlpath=lab/tree/examples/reference/widgets/EditableFloatSlider.ipynb

If I 

- drag the slider
- keep the mouse pressed 
- release the mouse

I see sometimes see the slider jumping around when mouse is pressed and always an error message when released.

https://user-images.githubusercontent.com/42288570/115104174-03d20a00-9f57-11eb-9750-675741011be6.mp4

```bash
  File "/srv/conda/envs/notebook/lib/python3.7/site-packages/pyviz_comms/__init__.py", line 325, in _handle_msg
    self._on_msg(msg)
  File "/home/jovyan/panel/viewable.py", line 261, in _on_msg
    doc.unhold()
  File "/srv/conda/envs/notebook/lib/python3.7/site-packages/bokeh/document/document.py", line 669, in unhold
    self._trigger_on_change(event)
  File "/srv/conda/envs/notebook/lib/python3.7/site-packages/bokeh/document/document.py", line 1180, in _trigger_on_change
    self._with_self_as_curdoc(event.callback_invoker)
  File "/srv/conda/envs/notebook/lib/python3.7/site-packages/bokeh/document/document.py", line 1198, in _with_self_as_curdoc
    return f()
  File "/srv/conda/envs/notebook/lib/python3.7/site-packages/bokeh/util/callback_manager.py", line 161, in invoke
    callback(attr, old, new)
  File "/home/jovyan/panel/reactive.py", line 278, in _comm_change
    self._process_events({attr: new})
  File "/home/jovyan/panel/reactive.py", line 250, in _process_events
    self.param.set_param(**self._process_property_change(events))
  File "/srv/conda/envs/notebook/lib/python3.7/site-packages/param/parameterized.py", line 1472, in set_param
    self_._batch_call_watchers()
  File "/srv/conda/envs/notebook/lib/python3.7/site-packages/param/parameterized.py", line 1611, in _batch_call_watchers
  File "/srv/conda/envs/notebook/lib/python3.7/site-packages/param/parameterized.py", line 1573, in _execute_watcher
    watcher.fn(*args, **kwargs)
  File "/home/jovyan/panel/widgets/slider.py", line 538, in _sync_value
    self.param.set_param(**{event.name: event.new})
  File "/srv/conda/envs/notebook/lib/python3.7/site-packages/param/parameterized.py", line 1465, in set_param
    setattr(self_or_cls, k, v)
  File "/srv/conda/envs/notebook/lib/python3.7/site-packages/param/parameterized.py", line 316, in _f
    instance_param.__set__(obj, val)
  File "/srv/conda/envs/notebook/lib/python3.7/site-packages/param/parameterized.py", line 318, in _f
    return f(self, obj, val)
  File "/srv/conda/envs/notebook/lib/python3.7/site-packages/param/__init__.py", line 623, in __set__
    super(Dynamic,self).__set__(obj,val)
  File "/srv/conda/envs/notebook/lib/python3.7/site-packages/param/parameterized.py", line 318, in _f
    return f(self, obj, val)
  File "/srv/conda/envs/notebook/lib/python3.7/site-packages/param/parameterized.py", line 909, in __set__
    raise TypeError("Constant parameter '%s' cannot be modified"%self.name)
TypeError: Constant parameter 'value_throttled' cannot be modified
```

Hoxbro:
Fixes #2204

The error in #2204 was related to a missing `edit_constant` and the possibility of multiple events in `_sync_value`, which should now be fixed for the three editable sliders.

When looking / testing the editable sliders I could not get `value_throttled` to work with the `_input_widget`. To fix this, I changed the communication from `jscallback` to `param.watch`. Furthermore, I don't completely understand why in the current design it is possible to exceed the limit of the bounds with the _input_widget like
| Current | This PR |
| ----| ----|
|before| after|

I would expect the bounds to set to newer be exceeded, but it seems to be possible by design. Therefore, I have only implemented it for `EditableIntSlider` and `EditableFloatSlider`.

### DIFF
--- a/panel/widgets/slider.py
+++ b/panel/widgets/slider.py
@@ -457,9 +457,6 @@ class _EditableContinuousSlider(CompositeWidget):
     show_value = param.Boolean(default=False, readonly=True, precedence=-1, doc="""
         Whether to show the widget value.""")
 
-    exceed_bounds = param.Boolean(default=False, doc="""
-        Whether is it possible to exceed the bounds.""")
-
     _composite_type = Column
     _slider_widget = None
     _input_widget = None
@@ -518,8 +515,7 @@ class _EditableContinuousSlider(CompositeWidget):
         self._label.param.set_param(**{'margin': margin, 'value': label})
 
     @param.depends('start', 'end', 'step', 'bar_color', 'direction',
-                   'show_value', 'tooltips', 'format', 'exceed_bounds',
-                   watch=True)
+                   'show_value', 'tooltips', 'format', watch=True)
     def _update_slider(self):
         self._slider.param.set_param(**{
             'format': self.format,
@@ -531,13 +527,6 @@ class _EditableContinuousSlider(CompositeWidget):
             'show_value': self.show_value,
             'tooltips': self.tooltips
         })
-
-        if self.exceed_bounds:
-            self._value_edit.start = None
-            self._value_edit.end = None
-        else:
-            self._value_edit.start = self.start
-            self._value_edit.end = self.end
         self._value_edit.step = self.step
 
     @param.depends('value', watch=True)
--- a/panel/widgets/slider.py
+++ b/panel/widgets/slider.py
@@ -472,22 +472,15 @@ class _EditableContinuousSlider(CompositeWidget):
         )
         self._slider.param.watch(self._sync_value, 'value')
         self._slider.param.watch(self._sync_value, 'value_throttled')
+
         self._value_edit = self._input_widget(
             margin=0, align='end', css_classes=['slider-edit']
         )
+        self._value_edit.param.watch(self._sync_value, 'value')
+        self._value_edit.param.watch(self._sync_value, 'value_throttled')
+
         label = Row(self._label, self._value_edit)
         self._composite.extend([label, self._slider])
-        self._slider.jscallback(args={'value': self._value_edit}, value="""
-        value.value = cb_obj.value
-        """)
-        self._value_edit.jscallback(args={'slider': self._slider}, value="""
-        if (cb_obj.value < slider.start) {
-          slider.start = cb_obj.value
-        } else if (cb_obj.value > slider.end) {
-          slider.end = cb_obj.value
-        }
-        slider.value = cb_obj.value
-        """)
         self._update_editable()
         self._update_layout()
         self._update_name()
@@ -528,6 +521,8 @@ class _EditableContinuousSlider(CompositeWidget):
             'show_value': self.show_value,
             'tooltips': self.tooltips
         })
+        self._value_edit.start = self.start
+        self._value_edit.end = self.end
         self._value_edit.step = self.step
 
     @param.depends('value', watch=True)
--- a/panel/widgets/slider.py
+++ b/panel/widgets/slider.py
@@ -468,7 +468,7 @@ class _EditableContinuousSlider(CompositeWidget):
         super().__init__(**params)
         self._label = StaticText(margin=0, align='end')
         self._slider = self._slider_widget(
-            margin=(0, 0, 5, 0), sizing_mode='stretch_width'
+            value=self.value, margin=(0, 0, 5, 0), sizing_mode='stretch_width'
         )
         self._slider.param.watch(self._sync_value, 'value')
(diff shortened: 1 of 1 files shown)


philippjfr:
>Furthermore, I don't completely understand why in the current design it is possible to exceed the limit of the bounds with the _input_widget like

Very much by design indeed, there's many scenarios where this is useful. That said there's an argument to be made that this behavior should be togglable and/or it should restore the original bounds if you exceed the bounds and then restore them. @jbednar Any thoughts on that?

Hoxbro:
Added the functionality to be togglable: 

after2

jbednar:
Thanks, but a single "exceedable" switch is not sufficent, because a widget needs to accept both exceedable and non-exceedable (hard and soft) bounds on both the lower and upper sides. A hard bound is an absolute limit -- it is an error to set the value outside those bounds. A soft bound or suggested range helps set the slider up to be useful, but can totally be exceeded if needed. Think about it as the soft bounds being for the user, helping the user have a slider that covers a useful range, and the hard bounds are for the program, making sure that the value never exceeds an allowed range so that the program doesn't crash (e.g. by avoiding negative numbers where they are not allowed).  

So here, we don't want an "exceed bounds" flag, because (a) that would apply to both the high and low which many have different characteristics, and (b) just because we want people to be able to go outside the soft bounds doesn't mean we want them to be able to go outside all conceivable bounds; there are still often hard bounds to respect.

E.g. many physical quantities like length or mass or area never allow negative numbers and allow arbitrarily large values in theory, but for a given app there's some finite range that's typical (e.g. 10 to 100). So we'd set a hard bounds of 0,None and a soft bounds of 10,100.  The slider would go between 10 and 100, while the spinner would allow any positive number.  Without the "100" we couldn't get a slider (as we can't have a slider ranging up to infinity), but it's arbitrary; we just have to pick something, and then let the user override it. Without the 0 we'll have to have assertions or exception handling in our code to catch negative numbers. Only with both full hard and soft bound support can we handle everything.

See https://github.com/bokeh/bokeh/issues/9943 for the full discussion.

Hoxbro:
I was not aware of the discussion behind this. I only looked at the [development documentation](https://pyviz-dev.github.io/panel/reference/widgets/EditableFloatSlider.html), where the feature of exceeding start and end is not mentioned. So thank you for the detailed explanation. 

I will remove the `exceed_bounds` asap and focus this PR on getting `value_throttled` working for the editable sliders.

Hoxbro:
Value throttled should now work for the editable sliders. 
<details>
  <summary>Code</summary>

  ``` python
import panel as pn


edit_float_slider = pn.widgets.EditableFloatSlider(name="Editable FloatSlider")


@pn.depends(edit_float_slider.param.value_throttled, watch=True)
def test_throttled_10(v1):
    print(v1)


edit_int_slider = pn.widgets.EditableIntSlider(name="Editable IntSlider", start=0, end=8, step=2, value=4)


@pn.depends(edit_int_slider.param.value_throttled, watch=True)
def test_throttled_11(v1):
    print(v1)


edit_range_slider = pn.widgets.EditableRangeSlider(name="Editable RangeSlider", start=0, end=8, step=2, value=(1,4))


@pn.depends(edit_range_slider.param.value_throttled, watch=True)
def test_throttled_12(v1):
    print(v1)


pn.Column(edit_float_slider, edit_int_slider, edit_range_slider,).servable()

  ```

</details>


after2



_Some notes / comments_ : 
1. If I don't add value to `_slider_widget`, value_throttled do not work for `edit_int_slider` when going to zero:
  - with value in `slider_widget`
  with
  - without value in `slider_widget`
  without
2. `EditableFloatSlider`´s FloatInput properly need a format like `EditableRangeSlider`.
3. The `EditableRangeSlider`´s  `_end_edit` should always be higher than `_start_edit` and the other way around
  limit

Issue #2207, opened 2021-04-17
Bokeh Theme is lost on update


MarcSkovMadsen:
When using a template I know from experience I need to create the template before I create other objects. The reason is that the Bokeh Theme does not work well otherwise. It's getting dropped when the user starts interacting with the app.

An example is the stocks_hvplot.ipynb notebook in the gallery. If you add 

```python
pn.template.FastListTemplate(site="Panel", title="hvPlot Stock Explorer", sidebar=[row[0]], main=[row[1]]).servable();
```

As an additional cell and serve it you will experience the theme is applied when the app is started and then dropped.

https://user-images.githubusercontent.com/42288570/115113758-27637780-9f8c-11eb-989c-7443fe8d0aab.mp4

## Additional Context

You should be able to experiment with it via

Notebook: https://mybinder.org/v2/gh/holoviz/panel/binder?urlpath=lab/tree/examples/gallery/apis/stocks_hvplot.ipynb
App: https://mybinder.org/v2/gh/holoviz/panel/binder?urlpath=panel/stocks_hvplot.ipynb

Issue #1816, opened 2020-11-27
Bokeh dark theme lost when removing and adding HoloViews pane


MarcSkovMadsen:
Panel 0.10.2
Bokeh 2.2.3

I would like to demo a nice styling of all Panel components in the Fast design style by enabling them to select and navigate to any Panel component from a drop down.

I can see if you switch away from the HoloViews pane and back the dark theme is lost. I would expect the theme to be applied consistently.

![bokeh_theme_lost](https://user-images.githubusercontent.com/42288570/100476037-dc33ef80-30e4-11eb-9628-31221292a3cc.gif)

```python
import numpy as np
import holoviews as hv
import panel as pn
import param

hv.extension("bokeh")
pn.extension("echarts")

ACCENT_REST = "#DF3874"


def _create_hvplot():
    # Generate some data
    cl1 = np.random.normal(loc=2, scale=0.2, size=(200, 200))
    cl2x = np.random.normal(loc=-2, scale=0.6, size=200)
    cl2y = np.random.normal(loc=-2, scale=0.1, size=200)
    cl3 = np.random.normal(loc=0, scale=1.5, size=(400, 400))
    # Create an overlay of points and ellipses
    clusters = (
        hv.Points(cl1).opts(color="blue").opts(tools=["hover"])
        * hv.Points((cl2x, cl2y)).opts(color="green")
        * hv.Points(cl3).opts(color="#FDDC22")
    )
    plot = (
        clusters
        * hv.Ellipse(2, 2, 2).opts(line_width=3, color=ACCENT_REST)
        * hv.Ellipse(-2, -2, (4, 2)).opts(line_width=3, color=ACCENT_REST)
    )
    return plot


def _create_echarts_plot():
    echart = {
        "tooltip": {},
        "legend": {"data": ["Sales"]},
        "xAxis": {
            "data": ["shirt", "cardign", "chiffon shirt", "pants", "heels", "socks"],
            "axisLine": {"lineStyle": {"color": "#ccc"}},
        },
        "yAxis": {
            "axisLine": {"lineStyle": {"color": "#ccc"}},
        },
        "series": [
            {
                "name": "Sales",
                "type": "bar",
                "data": [
                    1.0,
                    1.2,
                    1.4,
                    1.6,
                    1.8,
                    2.0,
                ],
                "itemStyle": {"color": "#DF3874"},
            }
        ],
        "responsive": True,
    }
    text_style = {"color": "#ccc"}
[...]

PR #2209, opened 2021-04-18
Fix theming on HoloViews plot updates

philippjfr:
Fixes https://github.com/holoviz/panel/issues/1816
Fixes https://github.com/holoviz/panel/issues/2207


Hoxbro:
I just tested, and this does also fix #1816.

philippjfr:
It does indeed.

PR #2269, opened 2021-04-30
Enhancements for Template modals

philippjfr:
Replaces custom modal implementations with the native implementations for Fast, Bootstrap and Material and ensures they are sized correctly.

@MarcSkovMadsen I've also removed the footer that says "Made with". It's nice to attribute Panel/Fast etc. but in many cases not appropriate. I'd be open to adding a footer area and perhaps even adding that by default but having no way to disable it just won't work.

Lastly this PR replaces the "hacky" HTML pane which is used to send executable scripts with a `TemplateActions` ReactiveHTML component which can trigger specific (pre-defined) actions simply by updating a single variable.


panel/template/material/material.html
```diff
@@ -135,21 +136,19 @@
   drawer.open = true;
   {% endif %}
 
-  var modal = document.getElementById("pn-Modal");
+  var modal_el = document.getElementById("pn-Modal");
   var span = document.getElementById("pn-closeModal");
```
philippjfr, line 140:
```suggestion
```

PR #2296, opened 2021-05-07
Add save_layout parameter to ReactTemplate

syamajala:
Add a parameter to save layout changes to localStorage for ReactTemplate.


MarcSkovMadsen:
It would be really nice to have the `FastGridTemplate` updated as well. They are normally kept in sync. At some stages they should be refactored and share the same code if possible.

philippjfr:
Thanks @syamajala. The main thing I'm wondering about is how we can save the layout per dashboard, e.g. if I deploy two different dashboards using the ReactTemplate will the layout settings for one persist for the other? Do we need to work out some mechanism to hash each application to allow multiple layout persistence. The simplest would be to simply hash it on the application title but other suggestions welcome.

philippjfr:
@syamajala Anything I can do to help here?

syamajala:
This should use the url for saving layouts now.

Issue #2297, opened 2021-05-08
DiscreteSlider formats numpy.int64 options as %.3g


MarcSkovMadsen:
Panel 0.11.3

When providing the `DiscreteSlider` with `numpy.int64` options the `formatter` is set to `%3g`

```python
from panel.widgets import DiscreteSlider
import numpy as np

def test_discrete_slider_with_numpy64_options():
    options = [np.int64(1), np.int64(1)]
    slider = DiscreteSlider(value=options[-1], options=options)
    assert slider.formatter != '%.3g'
```

```bash
$ pytest 'panel\tests\widgets\test_slider.py::test_discrete_slider_with_numpy64_options'
============================================================================================= test session starts ============================================================================================== 
platform win32 -- Python 3.8.4, pytest-6.2.3, py-1.10.0, pluggy-0.13.1 -- c:\repos\private\panel_docker\panel\.venv\scripts\python.exe
cachedir: .pytest_cache
rootdir: C:\repos\private\panel_docker\panel, configfile: tox.ini
plugins: anyio-2.2.0, nbsmoke-0.5.0, cov-2.11.1
collected 1 item                                                                                                                                                                                                 

panel/tests/widgets/test_slider.py::test_discrete_slider_with_numpy64_options FAILED                                                                                                                      [100%] 

=================================================================================================== FAILURES =================================================================================================== 
__________________________________________________________________________________ test_discrete_slider_with_numpy64_options ___________________________________________________________________________________ 

    def test_discrete_slider_with_numpy64_options():
        options = [np.int64(1), np.int64(1)]
        slider = DiscreteSlider(value=options[-1], options=options)
>       assert slider.formatter != '%.3g'
E       AssertionError: assert '%.3g' != '%.3g'
E        +  where '%.3g' = DiscreteSlider(options=[1, 1], value=1).formatter

panel\tests\widgets\test_slider.py:243: AssertionError
=========================================================================================== short test summary info ============================================================================================ 
FAILED panel/tests/widgets/test_slider.py::test_discrete_slider_with_numpy64_options - AssertionError: assert '%.3g' != '%.3g'
============================================================================================== 1 failed in 0.55s ===============================================================================================
```

## Solution

The formatter should have no decimals and use `,` as thousands separator if needed.

PR #2298, opened 2021-05-08
Fix DiscreteSlider uses '%.3g' formatter for numpy64 ints

MarcSkovMadsen:
This Addresses #2297

I've changed from old school python 2.6 and below format string to modern format string also. See https://mkaz.blog/code/python-string-format-cookbook/

You could consider renaming `formatter` to `format` also to align with other widgets. C.f. below

image

### DIFF
--- a/examples/reference/widgets/DiscreteSlider.ipynb
+++ b/examples/reference/widgets/DiscreteSlider.ipynb
@@ -27,6 +27,7 @@
     "* **``options``** (list or dict): A list or dictionary of options to select from\n",
     "* **``value``** (object): The current value; must be one of the option values\n",
     "* **``value_throttled ``** (object): The current value; must be one of the option values, throttled until mouseup\n",
+    "* **``formatter``** (str): Allows setting a string formatter, default is '.3g'. If the options specified are integers this is changed to '0,.0f'.\n",
     "\n",
     "##### Display\n",
     "\n",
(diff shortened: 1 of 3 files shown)


panel/widgets/slider.py
```diff
@@ -338,14 +350,19 @@ def _get_embed_state(self, root, values=None, max_opts=3):
...
     def labels(self):
         title = (self.name + ': ' if self.name else '')
-        if isinstance(self.options, dict):
-            return [title + ('<b>%s</b>' % o) for o in self.options]
-        else:
-            return [title + ('<b>%s</b>' % (o if isinstance(o, string_types) else (self.formatter % o)))
-                    for o in self.options]
+        return [title + ('<b>%s</b>' % (o if isinstance(o, string_types)
+            else self._format_value(o, self.formatter)))
+            for o in self.options]
```
philippjfr, line 364: This does not seem correct to me, if you're providing an `options` dictionary then the keys are the formatted values and don't need to have further formatting applied.

philippjfr: Also what happens if the values are neither strings or floats/ints? You can have any object type in the `options`.

examples/reference/widgets/DiscreteSlider.ipynb
```diff
@@ -27,6 +27,7 @@
     "* **``options``** (list or dict): A list or dictionary of options to select from\n",
     "* **``value``** (object): The current value; must be one of the option values\n",
     "* **``value_throttled ``** (object): The current value; must be one of the option values, throttled until mouseup\n",
+    "* **``formatter``** (str): Allows setting a string formatter, default is '.3g'. If the options specified are integers this is changed to '0,.0f'.\n",
```
philippjfr, line 30: This should be alphabetically sorted.

philippjfr:
Ping @MarcSkovMadsen, left you a few comments here.

PR #2301, opened 2021-05-09
Fix sidebar width issues and make it customizable

### LINKED ISSUE #2223: Make default width of sidebar match default width of components
Panel: 0.11.3

Going through the Panel gallery and wrapping things in Templates I can see that the default sidebar width of `300px` of Templates does not fit the default widget of Panel components.

An example is the `stocks_plotly.ipynb` example that looks like

image

when adding

`pn.template.FastListTemplate(site="Panel", title="Plotly Stock Explorer", sidebar=[ticker, window], main=[row[1]]).servable();`

The consequence is friction and more work for users. They have to specify a fixed width of `250px` or set `sizing_mode='responsive'`.

## Solution

Increase the default sidebar width of all templates to wrap fixed width widgets.

MarcSkovMadsen:
Fixes #2223

The problem is that the sidebar width does not match the default fixed width of buttons and dropdowns. Furthermore users are also requesting the ability to configure the sidebar width. The problem is shown below.

image


MarcSkovMadsen:
Only outstanding thing to do is to document it across notebooks.

jbednar:
Thanks! Very useful!

MarcSkovMadsen:
I have not documented and also implemented the `GoldenTemplate`. I believe this is ready.

When you have time @philippjfr . Please review and let me know if there is more to do. Thanks.

philippjfr:
Happy with this, thanks!

Issue #2175, opened 2021-04-15
Fasttemplate: Allow changing colors


rbps1:
### Current situation
The fast template is an excellent way of building dashboards without spending too much time for html optimisations. However the default colors can be an issues somethings (corporate environment with specific brand colors).

The underlying mechanisms of the fast template do not allow to overwrite color variables using additional css, instead colors have to be passed when creating the template. Currently, this does only support the header colors.

[https://panel.holoviz.org/reference/templates/FastGridTemplate.html#templates-gallery-fastgridtemplate](https://panel.holoviz.org/reference/templates/FastGridTemplate.html#templates-gallery-fastgridtemplate)

![grafik](https://user-images.githubusercontent.com/82584491/114861779-01a06c00-9dee-11eb-85ae-d69f64a415d4.png)

### Feature request: 
Allow to customize all color variables that the fast design template uses when creating a new instance of the fasttemplate.

```
:root {
  --accent-fill-active: #A01346;
  --accent-fill-hover: #DA1A5F;
  --accent-fill-rest: #C01754;
  --accent-foreground-active: #DF3874;
  --accent-foreground-cut-rest: #FFFFFF;
  --accent-foreground-hover: #E55E8E;
  --accent-foreground-rest: #E1477E;
  --background-color: #181818;
  --container-background: #3B3B3B;
  --neutral-divider-rest: #2E2E2E;
  --neutral-fill-active: #262626;
  --neutral-fill-hover: #333333;
  --neutral-fill-input-active: #181818;
  --neutral-fill-input-hover: #181818;
  --neutral-fill-input-rest: #181818;
  --neutral-fill-rest: #2B2B2B;
  --neutral-fill-stealth-active: #212121;
  --neutral-fill-stealth-hover: #262626;
  --neutral-fill-stealth-rest: #181818;
  --neutral-focus: #717171;
  --neutral-focus-inner-accent: #350617;
  --neutral-foreground-active: #E5E5E5;
  --neutral-foreground-focus: #E5E5E5;
  --neutral-foreground-hover: #E5E5E5;
  --neutral-foreground-rest: #E5E5E5;
  --neutral-layer-floating: #4A4A4A;
  --neutral-outline-active: #424242;
  --neutral-outline-hover: #808080;
  --neutral-outline-rest: #5A5A5A;
}
```

Applies to Grid and List version of the template.

### Discourse discussion
[...]

PR #2303, opened 2021-05-10
Custom Colors in Fast Templates

MarcSkovMadsen:
Addresses #2175

I would like to enable specifying a custom background, neutral and accent color for the FastListTemplate and FastGridTemplate. Both for the main area and for the header.

If possible I would like to implement it in a way that we user can also experiment with different colors, fonts and more like shown below.

https://user-images.githubusercontent.com/42288570/117608968-646ff380-b15f-11eb-8b64-25743387e3c0.mp4


MarcSkovMadsen:
Please note that I cannot get the `--container-background` css variable added so I've left that hard coded for now. I have asked on the Fast Discourse how to add this one. See https://discord.com/channels/449251231706251264/482296787021856768/841167712217858058

philippjfr:
What I'd love to do is expose it via a ReactiveHTML component so you can dynamically update the colors. Would be happy to help with a implementation.

MarcSkovMadsen:
But after having thought about it. I think that the scope of this PR will be getting the custom branding/ color working on Template construction. And a separate PR will address the dynamic configuration. Because the first part is a "blocker" for some users. The second part is the "show off" part.

MarcSkovMadsen:
Ready for review.

It would be nice to additionally

- Validation of color parameters. The Fast .js libraries only accept hex or rgb values
- Find some way to also set the `accent` color of the `index.html` template also, so that it is also "branded"

https://user-images.githubusercontent.com/42288570/118087111-56231100-b3c5-11eb-8fab-e96fb368a49f.mp4

philippjfr:
Tests are failing:

https://github.com/holoviz/panel/pull/2303/checks?check_run_id=2572708149#step:12:1230

The problem is that `param.Color` only accepts hex colors right now. I really think it should accept valid HTML colors though.

MarcSkovMadsen:
I will take a look @philippjfr . 

Please note the fast js only accepts hex or rgb. It does not accept html colors. And I need to use that to specify the color palette to use.

philippjfr:
>I have just been used to them always failing to reasons outside my PR

We have a very different experience then, they have been working pretty consistently from my perspective.

MarcSkovMadsen:
Maybe some of the approaches athttps://stackoverflow.com/questions/1573053/javascript-function-to-convert-color-names-to-hex-codes could be used to convert color name to hex or rgb on js side.

mattpap:
> (...) could be used to convert color name to hex or rgb on js side.

bokehjs has color management utilities in `core/util/color`, which may be useful here.

MarcSkovMadsen:
How would I use that in plain js in a `script` tag inside the `.html` template @mattpap ?

MarcSkovMadsen:
You can now use named colors. For the Fast templates. So now things should work.

When working on the next PR (config widget to change style on the fly) some refactoring might have to be introduced. But that is to be determined.

MarcSkovMadsen:
I don't plan to do more for now. I don't know of anything that needs improvement @philippjfr . When you have time please review and let me know. Thanks.

philippjfr:
Sorry I let this sit for so long. All the changes look good but there appear to be a bunch of merge conflicts.

MarcSkovMadsen:
Merge conflicts solved @philippjfr . I have tested that it still works manually and visually also.

Issue #2315, opened 2021-05-14
Support serving a Site of Apps from a configuration file


MarcSkovMadsen:
As soon as your collection of apps and scripts starts to grow, questions starts coming in

**And how do my users easily navigate the when working on the site?**
**Where do I find the code, documentation, ....?**
**Who developed or owns this particular app?**
**How can I easily manage which files to serve and where?**
**How can I customize the url of a specific app. I.e. not use `/spread_app` but `/spread-analysis` or even `/trading/spread-analysis` for my file `spread_app.py`.

In my experience a configuration file is needed. Currently no easy way to configure a site (via a configuration file) exists.

## Additional Requirements

- By default the `/` root page should show a nice Index/ Gallery page with images. Related to https://github.com/holoviz/panel/issues/2313.
- But it should also be able to serve a custom landing page at "/" and specify an alternative location like "/gallery" for the Gallery page.
- By default apps should include a nice menu to enable easy navigation across the site. Related to https://github.com/holoviz/panel/issues/2314

## Additional Context

The ability to easily serve a site from a list or glob of scripts is currently something special to Panel and should be treasured. Streamlit and Dash cannot do it. Bokeh and Voila don't have the nice gallery page that Panel now has.

- Deployments can be hard. So if you can keep to one deployment and add more apps via a configuration file it can make your time to market much faster.
- In my experience a single, micro app is not enough. Things quickly grow into multi-page apps/ larger sites.
- This can save energy resources and help make the world greener. For example deploying apps individually often means allocating a minimum of resources per app. In a Trading Environment where I am we can quickly create 10s, 100s or 1000s of apps that might not be used that often. But still needs to be available. Thus we are running a lot of resources idle.

I would like to specify something like `panel serve site.toml` and then the site is created from that with menu and nice gallery file. My current attempt to make it easy for Panel users to do this is here https://github.com/MarcSkovMadsen/awesome-analytics-apps-template

philippjfr:
This makes a lot of sense to me. The configuration file should also support all the commandline options that you'd usually supply to `panel serve`. We should come up with a full specification of what the configuration file would look like and all the options it would support.

jbednar:
BTW, there's a current proposal to add .ini support to Param: https://github.com/holoviz/param/issues/470 . I'm meeting with the author in a few days to plan how to incorporate those changes.

philippjfr:
[...]

Issue #2314, opened 2021-05-14
Make it easy to add a nested menu


MarcSkovMadsen:
As soon as you are creating a multipage app the question arrives. How do I add a menu?

![image](https://user-images.githubusercontent.com/42288570/118217015-ebc9a980-b474-11eb-973a-57afcbc60203.png)

Currently there is no easy to use solution.

- Using a `HTML` pane does not work very well if the menu is nested and expandable as the bokeh layout engine does not relayout when the menu is expanded.
- The Fast templates have been prepared for a menu based on a html string via the `sidebar_footer` parameter. That is what I am using at awesome-panel.org.
- It is really convenient to be able to add additional links to `non-Panel` apps into the menu. For example to documentation and other resources. As well as other apps built using other frameworks like Bokeh, Dash, Django, Flask, Streamlit, Voila to make a larger Analytics Apps deployment work together.
- It is really convenient to enable nesting (to at least one level) the menu.
- It is really convenient to enable the user to specify a custom sort method.
- It is really convenient to be able to specify this configuration via a `.ini` or `.toml` file with the same structure as requested in https://github.com/holoviz/panel/issues/2313

## Additional Context

There is some code here that can be used as a starting point for the FastTemplates.

https://github.com/MarcSkovMadsen/awesome-panel-extensions/blob/7ee861cf4b1af79dcf083c13f024ea9d2be85a86/awesome_panel_extensions/frameworks/fast/fast_menu.py

nghenzi:
thanks @MarcSkovMadsen  for tagging me. 

I will watch the fast_menu. I am exploring to add a nested menu prepending an element to the container templates with JS like this

```
<script>
  let container = document.getElementById('container')
  let p = document.getElementsByClassName("topnav")[0];
  container.prepend(p)
</script>
``` 

This together with the a row contrusted with the reactiveHTML template gives like result something like the following. 


https://user-images.githubusercontent.com/46707985/118735401-c2af6d00-b816-11eb-8bb7-e70494ae842b.mp4

It needs some css layout refinement css but I think it goes in the correct way. Besides, it should be possible to add it to the right of the page (not to the left due to the way the sidebar hide/show option is managed). 

For the api I think having a nested div of links would make the way to construct the nested menu.

nghenzi:
I continue working with the top Nav Bar.  To define the nested structure of the links it is only need a nested dict. 

[...]

Issue #2313, opened 2021-05-14
Enable Adding Meta Data link images to the index/ gallery page


MarcSkovMadsen:
After having announced the new index/ gallery page the first question is if images can be added. Currently they cannot.

My experience is that users and stakeholders in a business would like an easy to navigate index/ gallery page with images and more meta-information. Like author, code_url, documentation_url, owner etc.

## Solution

Add a GalleryTemplate to Panel.

PR #2320, opened 2021-05-15
FastGalleryTemplate. Includes User and Application model.

MarcSkovMadsen:
This PR enables a user to easily create a gallery of applications and/ or other resources from a `.yaml` file.

- Provides most of #2313. The rest can be provided when #2315 has been implemented.
- The User and Application Model developed in this PR is also needed for #2314

Todo:

[x] Add User and Application Models
[x] Add FastGalleryTemplate
[x] Add functionality to read configuration of Users and Applications from `.yaml` file
[] Document FastGalleryTemplate


MarcSkovMadsen:
You can test the gallery using a command like `panel serve panel/tests/test_fast_gallery_template.py --auto --show`

philippjfr:
Thanks for making a start on this. I would strongly urge you to split up the PR though, especially the config file will need significant discussion and should not be conflated with the gallery and menu.

panel/tests/template/fast/test_fast_list_template.py
```diff
@@ -21,9 +21,8 @@ def test_template_theme_parameter():
 
 def test_app():
     app = FastListTemplate(
-        title="FastListTemplate",
         site="Panel",
-        site_url="https://panel.holoviz.org",
```
MarcSkovMadsen, line 26: Not related to this PR. But improves the manual test.

panel/tests/test_template.py
```diff
@@ -18,16 +19,16 @@
 from panel.layout import Row
 from panel.pane import HoloViews, Markdown
 from panel.template import (
-    BootstrapTemplate, GoldenTemplate, MaterialTemplate, ReactTemplate,
+    BootstrapTemplate, FastListTemplate, FastGridTemplate, GoldenTemplate, MaterialTemplate, ReactTemplate,
```
MarcSkovMadsen, line 22: Not relevant to this PR. But I discovered these templates where not tested in same way as the other templates.

philippjfr:
As discussed I'll close but would love to see this in a separate repo.

PR #2343, opened 2021-06-02
Simplified styling of Card layout

philippjfr:
The Card css was somewhat brittle and I did not much like the old +/- indicators for uncollapsed/collapsed states respectively. This PR changes the styling to use arrows as indicators instead and makes the CSS less brittle:

### Accordion

Screen Shot 2021-06-02 at 6 06 50 PM

### Cards

Screen Shot 2021-06-02 at 6 07 35 PM

### DIFF
--- a/panel/dist/css/card.css
+++ b/panel/dist/css/card.css
@@ -24,7 +24,7 @@
 .bk.card-button {
   background-color: transparent;
   position: absolute;
-  left: 6px;
+  left: 0.5em;
 }
 .bk.card-header-row {
   position: relative !important;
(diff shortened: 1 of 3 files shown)


jbednar:
The down arrow when it's expanded makes good sense, given that it points to all the expanded content, but I think I've more often seen a right arrow than an up arrow for the unexpanded one, I presume because clicking on it feels like moving forward. Your call, of course!

philippjfr:
No you're right, feels more natural. Will follow up tomorrow.

PR #2368, opened 2021-06-10
Add FastTemplateSettings component

MarcSkovMadsen:
In order to quickly iterate on the style of your (Fast-)Template you need to be able to quickly experiment with different settings.

To facilitate this I propose adding a `FastTemplateSettings` component.

You can run the example via ` python 'panel\tests\template\fast\test_fast_settings.py'`.

https://user-images.githubusercontent.com/42288570/121462144-71774100-c9b0-11eb-92c2-423094a0ffa0.mp4

## Must Do

- [x] Create POC
- [x] Add `corner_radius` parameter to FastTemplates
- [x] Align FastTemplate and FastStyle functionality
    - [x] Fix header accent problem
. [x] Update `FastListTemplate` and `FastGridTemplate` documentation
- [x] Add standalone example to Gallery and set it up as an app on Binder
- [x] Refactor, Document, Clean up

## Nice to Do

- [] Also update Bokeh Theme if possible.
- [x] Fix bottom area not controlled by `body-design-provider`
- [] Enable adding the rest of the parameters like `site`, `logo`, `title`, .....
- [] Simplify `template_constructor` string to include only non-default values.


MarcSkovMadsen:
Only polish, examples and docs remain

image

MarcSkovMadsen:
@philippjfr . I have started to that think that maybe the FastStyler, FastSettingsEditor or what it’s called is better off in its own repository. Together with the FastGallery.

 Someday I probably would be creating the panel-fast repository anyways with the widgets I already have and plan to support in the future.

What do you think @philippjfr ?

philippjfr:
As long as the Fast templates are in Panel itself I think this has its place as well. I'd be very happy to eventually move the templates to Panel-fast as well and then use a plugin mechanism to support backward compatibility (i.e. to keep `pn.templates.FastListTemplate` working).

MarcSkovMadsen:
Another questions is changing the Template Editing should be provided. I see the following options.

1. **Standalone**: Standalone `FastTemplateEditor` component that the user can instantiate and add to the template how he/ she likes.
2. **Parameter**: The `FastListTemplate` and `FastGridTemplate` contains a parameter `editor` that can be used toggle whether to add the editor to the sidebar or not on construction.
3. **Always included**: The templates always contains the `FastTempleteEditor` and we add a `settings` or `configuration` button to toggle it on/ off in the header.

----------------

1. Would give the most flexibility.
3. Would we be simplest for the end user, it would give the Fast Templates even more `wow` effect. But could slow template down cause there I add to models extra to the template that will communicate a bit back and forth.

@philippjfr, @jbednar @nghenzi . Do you have any suggestions on which alternative I should go with?


https://user-images.githubusercontent.com/42288570/121785209-bb327800-cbb8-11eb-9d65-10051a61833a.mp4

MarcSkovMadsen:
Ready for review @philippjfr . The big question to me is the api. What should be the names/ apis of this? See question 1.-3. above.

You can check out the example notebooks and apps on Binder

## Notebooks

- https://mybinder.org/v2/gh/holoviz/panel/fast-styler?urlpath=lab/tree/examples/reference/templates/FastTemplateSettings.ipynb
- https://mybinder.org/v2/gh/holoviz/panel/fast-styler?urlpath=lab/tree/examples/gallery/templates/FastListTemplateDesigner.ipynb
- https://mybinder.org/v2/gh/holoviz/panel/fast-styler?urlpath=lab/tree/examples/gallery/templates/FastGridTemplateDesigner.ipynb

## App

- https://mybinder.org/v2/gh/holoviz/panel/fast-styler?urlpath=panel/FastListTemplateDesigner
- https://mybinder.org/v2/gh/holoviz/panel/fast-styler?urlpath=panel/FastGridTemplateDesigner

FastGridTemplateDesigner

MarcSkovMadsen:
From this PR could be extracted just the `FastListTemplate` and `FastGridTempate` improvements. If it will make it easier to review that first then let me know.

philippjfr:
Please do not commit gifs that big to the repo.

philippjfr:
Also please make sure the indentation of your TS contributions is consistent with the rest of the repo (2 spaces throughout).

philippjfr:
Same with your parameter formatting. I realize there is no style guide that declares an explicit style but I point this out on almost every PR, so please do follow the prevailing style for now.

MarcSkovMadsen:
Thanks for the comments. What would I do about assets in the future? Where can I contribute them?

philippjfr:
>Thanks for the comments. What would I do about assets in the future? Where can I contribute them?

At minimum they should be minimized and optimized, hacky and there are better solutions but I often use https://ezgif.com/optimize. However for longer gifs/videos I think I just have to upload them somewhere.

>In everything else I do Black just formats my code.

Giving you slack on Python because it's my stubbornness keeping us away from Black but that doesn't cover TypeScript.

panel/template/fast/settings/fast_template_settings.py
```diff
@@ -0,0 +1,347 @@
...
+    settings of the Template.
+
+    If you need more control you can add the `comms` and `editor`
+    components individually instead.
+    """
+
+    comms = param.ClassSelector(class_=Column, constant=True, doc="""
+        The component that provides bidirectional communication to set
+        and get the style values. Include this in your app if you want
+        to enable changing Template settings dynamically.""")
```
philippjfr, line 52: Can you explain this to me? This seems like a pretty big hack and makes it difficult for me to decide to merge this.

panel/template/fast/list/default.css
```diff
@@ -1,9 +1 @@
...
-#header {
-  box-shadow: 2px 2px 10px silver;
-}
-
-#sidebar {
-  background-color: white;
-  box-shadow: 2px 2px 10px silver;
-}
\ No newline at end of file
+/* fast default.css */
```
philippjfr, line 1: Empty file?!

philippjfr (commented):
The functionality here is great but in the current state I'm not comfortable merging this, particularly so close to a release (that's on me sorry!). Is there any chance you can split general improvements to the Fast template(s) into a separate PR and leave the designer for later?

panel/template/fast/settings/fast_template_settings.py
philippjfr: I guess the idea is to include the designers in the layout since they need to be added somewhere. Generally I really dislike this and instead these objects should be inserted as roots on the document so that they are entirely invisible to a user. I can do that refactoring but probably not before a 0.12 release.

MarcSkovMadsen: I was in doubt on whether to always include them or not. My guess is that 

- setting template colors etc. dynamically in a live app will not be used often. And then they will just add extra complexity/ slow down the template. The implication is to not include them in the template.
- users would like this to be as easy to use as possible. The implication is to always include them in the template.

So what to do?

panel/template/fast/list/default.css
MarcSkovMadsen: The file has been there for a long time. But more and more functionality has been moved to the template to make it dynamic.

I thought it should stay there as all template have these files ??

What should I do?

MarcSkovMadsen:
Yes I can split it. I will do it in the weekends.

panel/template/fast/settings/fast_template_settings.py
philippjfr: >I was in doubt on whether to always include them or not. My guess is that

I'm not suggesting that they should always be added just that the mechanism by which they are added is not what I'd want to see. Ideally users would toggle an option on the template which then ensures these objects are added as roots for the template.

panel/template/fast/list/default.css
philippjfr: Let's leave it for now.

philippjfr:
What's your feeling about the designer/settings component now? I still feel this might be out of scope. I'll close but if you want to include some of this in panel let's consider how we can do that in a clean way.

MarcSkovMadsen:
My hunch that some day we should refactor templates to be dynamic. I.e. you should be able to change site, title, favicon, header_background etc. dynamically. It could be done for example via ReactiveHTML.

philippjfr:
That seems like a good approach.

MarcSkovMadsen:
And probably some of the functionality should be carved out into separate components. For example the golden layout and grid layout could be dynamic components. The maximize screen could be carved out to a component. The header could be a component. The sidebar could be a component.

Issue #2173, opened 2021-04-14
Update the Panel Roadmap on the web site


MarcSkovMadsen:
It's from 11/2019 :-)

https://panel.holoviz.org/about/roadmap.html

![image](https://user-images.githubusercontent.com/42288570/114733811-f3961100-9d43-11eb-9f69-3bd4afa3e405.png)

PR #2382, opened 2021-06-14
Remove roadmap

philippjfr:
While we might want to readd a Roadmap the current items in the roadmap are either no longer priorities or have already been implemented.

Fixes https://github.com/holoviz/panel/issues/2173


gdementen:
FWIW, the FAQ at https://panel.holoviz.org/FAQ.html still point to it (leading to a 404 error).

Issue #2390, opened 2021-06-15
Add ability to specify default application


philippjfr:
When you launch multiple apps at once there should be a way to specify one application as the default, either by making it the root (`/`) or by somehow forwarding from the root rather than serving up the index page. Could also be implemented in bokeh.

MarcSkovMadsen:
I've been needing this for a long time as well. Just did not get to write the FR.

PR #2391, opened 2021-06-15
Add --default-app commandline argument

philippjfr:
Implements https://github.com/holoviz/panel/issues/2390


MarcSkovMadsen:
That was easy. 

My implementation would have been to reuse the `--index` argument. And if it was a `.py` or `.ipynb` file serve that. But I could not see through the consequences of that change. But it would have kept the arguments lower and maybe been a bit more intuitive? @philippjfr 

This is great for serving for example the FastGallery at "/".

philippjfr:
Not a bad idea, will play with that.

Issue #2416, opened 2021-06-23
Vertical Scrollbar in Fast Templates almost unusable due to small height


MarcSkovMadsen:
Panel: 0.11.3

![image](https://user-images.githubusercontent.com/42288570/123043202-3d603f00-d3f8-11eb-9c86-c773595fd8f6.png)

```python
import panel as pn

pn.extension(sizing_mode="stretch_width")

spacer = pn.Spacer(background="#FFEEFF", height=500, width=2500, sizing_mode="fixed")
scroll_container = pn.Row(spacer, scroll=True)
pn.template.FastListTemplate(title="Scroll Issue",main=[scroll_container]).servable()
```

PR #2417, opened 2021-06-23
fix vertical scrollbar in fast templates

MarcSkovMadsen:
Fixes https://github.com/holoviz/panel/issues/2416 by setting a reasonable height for the horizontal scrollbar.

image

```python
import panel as pn

spacer = pn.Spacer(background="#FFEEFF", height=500, width=2500, sizing_mode="fixed")
scroll_container = pn.Row(spacer, scroll=True)
pn.template.FastListTemplate(title="Scroll Issue",main=[scroll_container]).servable()
```

### DIFF
--- a/panel/template/fast/grid/fast_grid_template.css
+++ b/panel/template/fast/grid/fast_grid_template.css
@@ -187,7 +187,7 @@ html {
   scrollbar-shadow-color: var( --neutral-fill-rest);
 }
 
-::-webkit-scrollbar { width: 12px; height: 3px;}
+::-webkit-scrollbar { width: 12px; height: 12px;}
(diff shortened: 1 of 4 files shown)


panel/template/fast/grid/fast_grid_template.css
```diff
@@ -187,7 +187,7 @@ html {
   scrollbar-shadow-color: var( --neutral-fill-rest);
 }
 
-::-webkit-scrollbar { width: 12px; height: 3px;}
```
philippjfr, line 190: Is there a reason why this definition isn't in `panel/template/fast/css/fast_root.css`?

MarcSkovMadsen: Legacy reasons and laziness :-)

I've just pushed a change to move it to fast_root.css.

Issue #2424, opened 2021-06-24
Fast Tabulator Styling


MarcSkovMadsen:
I'm on the current master branch (pre 0.12 release) of Panel.

I'm a fan of the `Tabulator` widget and the `Fast` templates. 

I would like them to work nicely together. More specifically I would like the `site` and `midnight` themes to work better with the Fast templates.

## Site

![image](https://user-images.githubusercontent.com/42288570/123200955-0ea69f00-d4b2-11eb-9f1f-e9920073671e.png)

## Midnight

The midnight theme works relative ok. But could probably be improved further.

![image](https://user-images.githubusercontent.com/42288570/123200993-267e2300-d4b2-11eb-9ce1-60d4606133f9.png)

## Code

```python
import panel as pn
import pandas as pd

pn.extension(sizing_mode="stretch_width")
ACCENT_COLOR = "#FDAC53"
data = pd.DataFrame(
    {
        "x": [
            1,
        ]
        * 10,
        "y": [
            2,
        ]
        * 10,
    }
)
tabulator = pn.widgets.Tabulator(data, theme="site")

pn.template.FastListTemplate(
    title="Fast Tabulator",
    main=[tabulator],
    sidebar=[tabulator.param.theme],
    header_background=ACCENT_COLOR,
    header_accent_base_color="white",
    accent_base_color=ACCENT_COLOR,
).servable()
```

PR #2425, opened 2021-06-24
improve Tabulator style for Fast Templates

MarcSkovMadsen:
Fixes https://github.com/holoviz/panel/issues/2424

During the implementation of this I realized

- It's difficult to customize the style of one theme from `fast_root.css` without affecting all Tabulator themes.
- When you change the Tabulator theme dynamically in the app, a new css `link` element is appended to the bottom of the `head` section which makes it difficult to override styles from `pn.config.raw_css`, `template.config.raw_css`, `fast_root.css` etc.
- When changing the Tabulator theme dynamically the Tabulator widget did not resize nicely on the "first change to a new theme". 

As a consequence I only made one very minor change to `fast_root.css`. Furthermore I updated the `tabulator.ts` code to update the `link` element instead of replacing it. It works in my tests. But I am thinking that maybe there was a reason for not doing this in the first place? I also added a small timeout before resizing (to wait for css to load) which worked out nicely in my tests.

Further improvements to Tabulator for Fast Templates should probably be done by contributing a Fast theme to Tabulator it self.

image

```python
import panel as pn
import pandas as pd

ACCENT_COLOR = "#FDAC53"
data = pd.DataFrame(
    {
        "x": [
            1,
        ]
        * 10,
        "y": [
            2,
        ]
        * 10,
    }
)
tabulator = pn.widgets.Tabulator(data, theme="site")

template=pn.template.FastListTemplate(
    title="Fast Tabulator",
    main=[tabulator],
    sidebar=[tabulator.param.theme],
    header_background=ACCENT_COLOR,
    header_accent_base_color="white",
    accent_base_color=ACCENT_COLOR,
)
template.servable()
```


MarcSkovMadsen:
This is the affect of the change across Tabulator themes

https://user-images.githubusercontent.com/42288570/123207381-97770800-d4bd-11eb-8663-206b59e178db.mp4

panel/models/tabulator.ts
```diff
@@ -353,13 +353,19 @@ export class DataTabulatorView extends PanelHTMLBoxView {
       }
     }
 
-    let parent_node = document.getElementsByTagName("head")[0]
     if (old_node != null) {
-      if (old_node.href == css)
+      if (old_node.href == css){
```
philippjfr, line 357:
```suggestion
      if (old_node.href == css)
```

panel/models/tabulator.ts
```diff
@@ -353,13 +353,19 @@ export class DataTabulatorView extends PanelHTMLBoxView {
...
         return false
-      if (old_node.parentNode != null)
-        parent_node = old_node.parentNode
+      } else {
```
philippjfr, line 359:
```suggestion
      else {
```

philippjfr: Should be consistent about formatting single line blocks.

philippjfr:
Will need to look at this closely, getting the dynamic CSS loading working in all cases was quite difficult, so I'd be surprised if this wasn't causing some regression.

MarcSkovMadsen:
FYI @philippjfr . I will give this a second iteration including creating a seperate `Fast` css stylesheet. Can see it's needed. I will try to give it as a PR to Tabulator and ask if Fast design folks are interested in reviewing. 

But probably we will have to host it our selves for Panel 0.12.

image

MarcSkovMadsen:
I now have a nice stylesheet. See https://discourse.holoviz.org/t/tabulator-style-for-fast-templates/2478

fast-tabulator-stylesheet-x2

MarcSkovMadsen:
Having a stylesheet changes the situation a bit. Eventually I would like it to be as easy as possible to use Tabulator with the Fast Templates. And actually not only Tabulator but the most used panes and widgets including matplotlib, plotly etc. I see the following options.

1. Add Fast theme to Tabulator widget based on 
    - i. self hosted css stylesheet
    - ii. Tabulator hosted css stylesheet
2. Just document the link to the self hosted css stylesheet. 

3. Could be hosted at
    - i. Panel github
    - ii. awesome-panel-assets (currently)
    - iii. panel-fast github (not created yet)

Eventually I would like the Fast templates to select the right themes and/ or includes the right css stylesheet automatically in order to make it as easy as possible to use.

What do you recommend @philippjfr ?

philippjfr:
Bundled the theme into Panel for now, which means as soon as Panel 0.12 is published it'll be available from the CDN.

MarcSkovMadsen:
Thanks. I noticed the modifier option. Would it be possible to add a modification based on the theme?

That would make it possible to set the Plotly theme automatically etc. Removing a lot of friction.

philippjfr:
>Thanks. I noticed the modifier option. Would it be possible to add a modification based on the theme?

In theory yes, I think we should just put modifiers on the theme as well, but probably not for 0.12.

PR #2457, opened 2021-06-29
Add separate RangeSlider value_start and value_end parameters

philippjfr:
Can make some filtering operations easier.

### DIFF
--- a/panel/widgets/slider.py
+++ b/panel/widgets/slider.py
@@ -16,10 +16,12 @@ from bokeh.models.widgets import (
 
 from ..config import config
 from ..io import state
-from ..util import param_reprs, unicode_repr, value_as_datetime, value_as_date
+from ..util import (
+    edit_readonly, param_reprs, unicode_repr, value_as_datetime, value_as_date
+)
 from ..viewable import Layoutable
-from .base import Widget, CompositeWidget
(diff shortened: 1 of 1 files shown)


panel/widgets/slider.py
```diff
@@ -351,13 +353,50 @@ def values(self):
...
+            params['value'] = (params.get('start', self.start),
+                               params.get('end', self.end))
+        params['value_start'], params['value_end'] = params['value']
+        super().__init__(**params)
+
+    @param.depends('value', watch=True)
+    def _sync_values(self):
+        vs, ve = self.value
+        with edit_readonly(self):
+            self.param.set_param(value_start=vs, value_end=ve)
```
jbednar, line 378: Were value_start and value_end meant to be readonly, then?

philippjfr: Hmm, yeah this is wrong. Thanks for catching it, if I can I will make it read-only again.

jbednar: I'm not sure it's feasible for them to be read only if you want them to be independently controlled by widgets. Seems like it would be good to see if param.Composite can be updated to work with events and dependencies, as it was meant for precisely this use case...

Issue #2062, opened 2021-03-08
Include a debug/info decorator


hyamanieu:
### Current status

In its current status, interactive Panel objects raising errors will either create the traceback in the jupyter cell (depending on `panel.config.console_output`) or inside the console in a server context.

This is inconvenient in a production context as a raised error may just look like an unresponsive application.

### Short description
In a production context as well as during analytical work, quickly knowing why an error is raised is very helpful.

I would like to suggest a decorator to add to any function that is used to render a dynamic pane, e.g. using `pn.depends`, `param.depends`, or `pn.interact`.

This solution could be either included in the standard library, or if it is too niche inside a template or in the gallery.

### Proposed solution

add the following decorator in the standard library

```python
import sys
import traceback
import functools
import panel as pn

info_layout = pn.Tabs(closable=True)

def status_decorator(f: callable =None, behavior: str='raise', info_layout: pn.layout.base.ListPanel=info_layout) -> callable: 
    """   
    Parameters
    ----------
    f : callable, optional
        The function to decorate.
    behavior : str, optional
        If 'replace', the object which should have been rendered will be replaced by the error message and the error will not be raised. 
        If 'raise', the error will be raised, meaning the target object will look frozen. 
        If 'erase', the function will return None and nothing will be rendered in place of the target object. 
        The default is 'raise'.
        
        In all cases, the error message will be appended to the list-like panel object 'info_layout'
    info_layout : pn.layout.base.ListPanel, optional
        A list-like panel object where each element is an error status. The default is info_layout.


    Returns
    -------
    callable
        The decorated function.

    """
    
    def mydec(f):
        @functools.wraps(f)
        def wrapper(*args,**kwargs):
            try:
                #info_layout.clear()
                return f(*args,**kwargs)
            except Exception as e:
                v, e, tb = sys.exc_info()
                stack = traceback.extract_tb(tb)
                (filename, line, procname, text) = stack[-1]
                info = pn.pane.Alert(object=repr(stack[-1]), height=100)
                info_layout.append((procname,info))
                if behavior == 'replace':
                    return info
                elif behavior == 'raise':
                    raise e
                elif behavior == 'erase':
                    return
                    
        return wrapper
            
    if f is None:
        return mydec
    else:
        return mydec(f)
```

The `ListPanel` object I am using is Tabs, so that the user can close it once they acknowledge they read it. Whenever https://github.com/holoviz/panel/issues/1555 is closed, any type of list type layout would work well.

[...]

PR #2548, opened 2021-07-20
Debugger widget

hyamanieu:
Added debugger widget. Fixes https://github.com/holoviz/panel/issues/2062
NB: only interesting change in reactive.py is within `_process_events`
Sample code to try out:

```python
import panel as pn
import param
pn.extension()
class A(param.Parameterized):

    with_throttled_enabled = param.Range(
        default=(100, 250),
        bounds=(0, 250),
    )

    def __init__(self, **params):
        super().__init__(**params)
        widgets = {
            "with_throttled_enabled": {
                "type": pn.widgets.IntRangeSlider,
                "throttled": False,
            },
        }
        self.controls = pn.Param(self, widgets=widgets)

        
    @param.depends("controls")
    #@status_decorator(behavior='erase')
    def calculation(self):
        if self.with_throttled_enabled[0] > 130:
            raise ValueError('too low')
        return pn.Pane((self.with_throttled_enabled), min_width=200)

    @param.depends("controls")
    #@status_decorator(behavior='erase')
    def buggy_calculation(self):
        return pn.Pane(self.with_throttled_enabled[1]/self.with_throttled_enabled[0], min_width=200)

a = A()


debug = pn.widgets.Debugger(name='Debugger')


a = pn.Column(a.controls, pn.pane.Markdown('### will bug above 130'), 
          a.calculation, 
          pn.pane.Markdown('### division by 0'), 
          a.buggy_calculation, 
          debug,
          )
a.servable()
```

## To-do:
- [x] Correct visualization defaults
- [x] Correct button locations
- [x] Add example notebooks for the doc
- [x] Add test
- [x] make sure the console output works as usual in a notebook.


panel/reactive.py
```diff
@@ -253,22 +249,20 @@ def _param_change(self, *events):
...
-                self.param.set_param(**self_events)
-            for k, v in self_events.items():
-                if '.' not in k:
-                    continue
-                *subpath, p = k.split('.')
-                obj = self
-                for sp in subpath:
-                    obj = getattr(obj, sp)
-                with edit_readonly(obj):
-                    obj.param.set_param(**{p: v})
```
philippjfr, line 271: You undid a bunch of changes here.

panel/reactive.py
```diff
@@ -290,19 +284,15 @@ def _change_event(self, doc=None):
             state.curdoc = None
             state._thread_id = None
 
-    def _comm_change(self, doc, ref, comm, subpath, attr, old, new):
-        if subpath:
-            attr = f'{subpath}.{attr}'
```
philippjfr, line 295: And here.

panel/reactive.py
```diff
@@ -290,19 +284,15 @@ def _change_event(self, doc=None):
...
         if attr in self._changing.get(ref, []):
             self._changing[ref].remove(attr)
             return
 
         with hold(doc, comm=comm):
             self._process_events({attr: new})
 
-    def _server_change(self, doc, ref, subpath, attr, old, new):
-        if subpath:
-            attr = f'{subpath}.{attr}'
```
philippjfr, line 305: And here.

philippjfr:
This looks great! However it looks like you started with an old(er) checkout and simply undid a bunch of recent changes when you rebased.

hyamanieu:
Yes I have noticed the mistake but didn't have time to fix. I added a
comment about it, I'll fix the merge conflict next time.

Le mar. 20 juil. 2021 à 16:10, Philipp Rudiger ***@***.***> a
écrit :

hyamanieu:
I can't get to have an exception accumulated in a notebook cell, even with master branch & 0.11.3. Not sure how `ServableMixing._on_error` gets triggered.

philippjfr:
Would be great to take this opportunity to add a bunch more log statements throughout the codebase as. The places you identified are definitely good but there's probably a bunch more that would help.

hyamanieu:
> Would be great to take this opportunity to add a bunch more log statements throughout the codebase as. The places you identified are definitely good but there's probably a bunch more that would help.

I identified the functions called whenever a callback is fired. 

I would use `panel.callbacks` logger only when the end user is the target, not the Panel developers. Or what is your intent?

panel/reactive.py
hyamanieu: outdated

philippjfr:
>I would use panel.callbacks logger only when the end user is the target, not the Panel developers. Or what is your intent?

No that makes sense. Just wondering if we can have a few log levels. I'm also thinking of adding an `/admin` endpoint to a server if requested, which provides details about the active sessions and optionally provides access to detailed logs as well.

hyamanieu:
> > I would use panel.callbacks logger only when the end user is the target, not the Panel developers. Or what is your intent?
> 
> No that makes sense. Just wondering if we can have a few log levels. I'm also thinking of adding an `/admin` endpoint to a server if requested, which provides details about the active sessions and optionally provides access to detailed logs as well.

A bit like in jupyterhub, that could be helpful indeed. But that could be left at the discretion of the developing user, all the tools are already there to do such a thing.

philippjfr:
Sorry I left this so late to review, I've cleaned this up a bit and changed it to subscribe to the main panel logger. I'm still a little bit concerned about the brittleness of the width/height/sizing_mode handling which was pretty broken when I first looked at this. I've had to make some changes to terminal sizing but I don't think this handles every combination of min/max_width/height options and different sizing modes correctly. I do want to get a 0.13.0 RC release out today. Will try to clean this up more and get it in.

hyamanieu:
Philipp, thanks for your review & corrections.
I have a bettered version of this widget in a separate project, I still didn't have the time to add it within this branche of Panel and won't have time for the next two weeks. 

Some features that my experience with my user case made me add:

- I have fixed some sizing problems between card & terminal, although I may have not thoroughly tested it like you as I simply fixed it.
- I have added the possibility to optionally subscribe to additional loggers
- I have added a third button to send an email with logs attached to a specified address (basically a `mailto:` URI)
- I have added an option to have a flying debugger if embedded within a template (it requires changes in the template css however :s )

To do:
- [ ] Make it work in jupyter. After pulling from master yesterday, it's not working in jupyter (trad and lab) anymore. I have unticked it from my first message.
- [ ] (optionally) find a way to auto-test widgets within notebook if it's possible ; so that next time we don't need to manual test it.


There are still some problems with terminal itself where opening and closing the card. The text sometimes appear and sometimes disappear. Interestingly, exporting the log by clicking on the floppy disk button shows the full output even if the terminal is completely black. After testing a bit more, I will open a separate issue as it's not related to this widget but the interaction of `Terminal `& `Card`.

Besides the js console keeps complaining about the missing buttons when the card is closed. Not sure if a callback can be "deactivated" whenever the Card is closed so that we don't have the warning in the console.

hyamanieu:
@philippjfr I have tested your changes. Very nice. I could not make sizing errors and the debugger now works in a jupyter context.

I have added a couple options (logging formatting + decide which logger to prompt) and fixed documentation.

I allowed myself to ping you for a review. There is a left over `print`, not sure this was wanted.

Besides the latter, to me it is ready to merge.

Thanks :)
ps: I'll add the "flying" effect and the mailing option at some point later. Current version works already well for end users to communicate clearly with dashboard developers.

PR #2555, opened 2021-07-21
Fix bug updating Tqdm indicator

### LINKED ISSUE #2554: TQDM.ipynb reference raises exception: AttributeError: Column.value property descriptor does not exist
I'm on Panel 0.12.0, Param 1.11.1, Bokeh: 2.3.3

If I run the TQDM.ipynb reference notebook top to bottom and click one of the buttons I get a `AttributeError: Column.value property descriptor does not exist` error.

image

philippjfr:
Fixes #2554

### DIFF
--- a/panel/widgets/indicators.py
+++ b/panel/widgets/indicators.py
@@ -749,6 +749,8 @@ class Tqdm(Indicator):
 
     _layouts = {Row: 'row', Column: 'column'}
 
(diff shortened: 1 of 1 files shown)


MarcSkovMadsen:
Ahhh. Simple solution.

philippjfr:
Yeah, just unfortunate I didn't catch this. I did test it extensively but then something must have changed.

PR #2572, opened 2021-07-24
support packages in config of _autoload_js cf. #2571

### LINKED ISSUE #2571: "packages" no longer supported in __js_require__
I just upgraded the `panel-highchart` extension to Panel 0.12.0. But I can see something changed.

image

You can replicate this on Binder: https://mybinder.org/v2/gh/marcskovmadsen/panel-highcharts/6c10727?urlpath=lab/tree/examples/HighChart.ipynb

I believe the issue is that I "mock" Panel to be able to load the extension via `pn.extension('highchart')`. The mocking is done in https://github.com/MarcSkovMadsen/panel-highcharts/blob/6c107271a4902f7a8df0b87f3b7a9c81d58c1149/panel_highcharts/config.py

image

This worked in Panel 0.11 but appearently not in Panel 0.12.

## Solution

Describe and support a way to develop and deploy Panel Extensions. Currently I have to investigate, do some mocking and then fix when mocking breaks in new versions.

MarcSkovMadsen:
Fixes #2571

@philippjfr . If you want to support `None` exports values. C.f. https://github.com/holoviz/panel/issues/2571#issuecomment-886083634 then I or you would have to change line 93 below and maybe more

image

But this fix is ok for me.

### DIFF
--- a/panel/io/notebook.py
+++ b/panel/io/notebook.py
@@ -96,7 +96,7 @@ AUTOLOAD_NB_JS = _env.get_template("autoload_panel_js.js")
 NB_TEMPLATE_BASE = _env.get_template('nb_template.html')
 
 def _autoload_js(bundle, configs, requirements, exports, skip_imports, ipywidget, load_timeout=5000):
-    config = {'paths': {}, 'shim': {}}
+    config = {'packages': {}, 'paths': {}, 'shim': {}}
     for conf in configs:
         for key, c in conf.items():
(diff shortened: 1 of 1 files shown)


philippjfr:
Tbh we should a) document `__js_require__` and b) add some tests so I don't introduce regressions again.

philippjfr:
I haven't used the packages config myself so didn't anticipate anyone else using it.

Issue #2365, opened 2021-06-09
Add event support to Vega/ Altair


MarcSkovMadsen:
I would like to be able to react to clicks, hover etc on Altair and Vega Charts.

You can find the different events here https://vega.github.io/vega/docs/event-streams/.

joelostblom:
There is a POC for this in streamlit which might be helpful https://github.com/streamlit/streamlit/issues/2263. There is also some similar work in b2 https://github.com/yifanwu/b2

PR #2592, opened 2021-07-30
Implement support for Vega events

philippjfr:
vega_events

Implements https://github.com/holoviz/panel/issues/2365


maximlt:
Really liking this API! 👍 

The description of `debounce` may need a few more words, or a link to a source explaining what that is, I expect more users not to be acquainted with this sort of concept. And it's pretty important here, not to have an expensive function being called to many times.

Also wondering if this API could in the future be adapted to the other interactive plotting panes, like Bokeh/Matplotlib...

MarcSkovMadsen:
+1 for a Commons event API, especially Across plot panes.

Issue #2608, opened 2021-08-04
Provide a square logo that will work in a dark theme


MarcSkovMadsen:
The square logo include in the panel repo and used at the Panel site only works for default/ light themed templates. I need the logo to either work for both themes or have one for each.

The logo file is here https://raw.githubusercontent.com/holoviz/panel/b220588de764e7a8aae92b8d5d764c21d7374c4a/doc/_static/logo_stacked.svg and here https://panel.holoviz.org/_static/logo_stacked.png

![image](https://user-images.githubusercontent.com/42288570/128118288-29127dca-04d6-4e9f-a09b-17622b7b82fb.png)

PR #2609, opened 2021-08-04
Logos for dark theme

MarcSkovMadsen:
Adds #2608

## The logo_stacked_dark_theme.png is working for me

image


jbednar:
The official font is Ubuntu, and you may need that installed to view the logo properly (in which case the logo should probably be exported with different options). The kerning has been tightened as well.

MarcSkovMadsen:
Ahh ok. Then the svgs actually work (if the user has Ubuntu installed) and there is not more to do in this PR.

PR #2625, opened 2021-08-10
Do not fix line-height in Fast template

### LINKED ISSUE #2600: Number Indicator overlaps with Fast Templates
Hello,

when using the Number indicator on a fast template, the sizing is broken.
```python
import panel as pn
pn.extension()

pn.template.FastListTemplate(
    main=[pn.indicators.Number(name="MyNumber", value=42)]
).show()
```
fast_broken

Other templates show it fine, e.g. bootstrap
```python
import panel as pn
pn.extension()

pn.template.BootstrapTemplate(
    main=[pn.indicators.Number(name="MyNumber", value=42)]
).show()
```

bootstrap_fine

philippjfr:
@MarcSkovMadsen Could you test this a bit. Fixing the line-height like this seems like an exceptionally bad idea but we have to see if this causes issues.

Fixes #2600

### DIFF
--- a/panel/template/fast/css/fast_bokeh.css
+++ b/panel/template/fast/css/fast_bokeh.css
@@ -23,7 +23,6 @@
   background-image: none;
   font-family: var(--body-font);
   font-size: var(--type-ramp-base-font-size);
-  line-height: var(--type-ramp-base-line-height);
 }
 .bk-root pre.bk {
   font-family: Courier, monospace;


philippjfr:
Going to merge, cut an RC release and update the gallery examples to see how this affects things.

MarcSkovMadsen:
Ok. I don't know how the Fast Templates will take this. It's not me inventing that the line-height should be "fixed" to a variable in the Fast Design system. It's what the design system does.

philippjfr:
Are you sure that it sets the line-height to an absolute pixel value?

philippjfr:
In all the apps I've tested it seems fine. I'll be releasing 0.12.1 in the next couple of hours so if you get a chance to test as well that would be appreciated.

MarcSkovMadsen:
The apps I have tested also look fine. But YES I am sure it is what fast.design does :-)

image

philippjfr:
Yeah, but as you can see it defines a bunch of variable line heights depending on the exact component. You just applied the `--type-ramp-base-line-height` globally for all `.bk` components, which just isn't going to work.

PR #2632, opened 2021-08-13
Fix #2631 - handle RuntimeError: dictionary changed size during iteration

### LINKED ISSUE #2631: Syncable._param_change not robust
I'm on Panel 0.11.3

I have a complicated app where I show the currently running `Job`. The job runs in background thread. When the job is finished its no longer shown in the app.

I can see that sometimes (random) `Syncable._param_change` fails with a

```bash
  File "/opt/conda/lib/python3.8/site-packages/panel/reactive.py", line 237, in _param_change
    for ref, (model, parent) in self._models.items():
RuntimeError: dictionary changed size during iteration
```

The code of `Syncable._param_change` is

```python
 def _param_change(self, *events):
        msgs = []
        for event in events:
            msg = self._process_param_change({event.name: event.new})
            if msg:
                msgs.append(msg)

        events = {event.name: event for event in events}
        msg = {k: v for msg in msgs for k, v in msg.items()}
        if not msg:
            return

        for ref, (model, parent) in self._models.items():
                self._apply_update(events, msg, model, ref)
```

I can see that when it fails the value of  `model` is `CheckboxGroup(id='7669', ...)`. I can see that the value of `self._models.items()` is ` dict_items([])`. My understanding is what triggered `_param_change` was that I set `job.running=False`. But during `_param_change` the job is removed from the application leading to `self._models_items()` changing.

My guess is that this should be supported and somehow the iteration should be done on a copy of `self._models.items()`. Or alternatively error handling should be introduced???

MarcSkovMadsen:
Fixes #2631. Probably also needed for #2597 to work in practice.

I don't know if

- There are any side effects. For example is it always good to `_apply_update` if  `self._models.items()` has changed?
- It would be more efficient to use `.copy` instead of `list`? But according to https://www.geeksforgeeks.org/python-cloning-copying-list/ list seems like an efficient solution. And using `list` is also what is mostly recommended here https://stackoverflow.com/questions/11941817/how-to-avoid-runtimeerror-dictionary-changed-size-during-iteration-error.

### DIFF
--- a/panel/reactive.py
+++ b/panel/reactive.py
@@ -248,7 +248,7 @@ class Syncable(Renderable):
         if not msg:
             return
 
-        for ref, (model, parent) in self._models.items():
+        for ref, (model, _) in list(self._models.items()):
             self._apply_update(events, msg, model, ref)
 
     def _process_events(self, events):
@@ -939,7 +939,7 @@ class ReactiveData(SyncableData):
                 finally:
                     self._updating = False
                 # Ensure that if the data was changed in a user
-                # callback, we still send the updated data 
+                # callback, we still send the updated data
                 if old_data is not self.value:
                     self._update_cds()
         if 'indices' in events:
--- a/panel/reactive.py
+++ b/panel/reactive.py
@@ -248,7 +248,7 @@ class Syncable(Renderable):
         if not msg:
             return
 
-        for ref, (model, _) in list(self._models.items()):
+        for ref, (model, _) in self._models.copy().items():
             self._apply_update(events, msg, model, ref)
 
     def _process_events(self, events):


Hoxbro:
I would definitely use `copy` for the reasons mentioned [here](https://mobile.twitter.com/raymondh/status/1251953119556104192).

philippjfr:
Good advice @Hoxbro!

MarcSkovMadsen:
Our threading expert @Hoxbro to the rescue :-) Thanks.

MarcSkovMadsen:
There is a longer discussion here to learn from https://bugs.python.org/issue40327

panel/reactive.py
```diff
@@ -248,7 +248,7 @@ def _param_change(self, *events):
         if not msg:
             return
 
-        for ref, (model, parent) in self._models.items():
+        for ref, (model, _) in list(self._models.items()):
```
philippjfr, line 251:
```suggestion
        for ref, (model, _) in self._models.copy().items():
```

PR #2644, opened 2021-08-18
Compatibility for bokeh 2.4

philippjfr:


panel/io/server.py
```diff
@@ -317,7 +327,11 @@ def handle_exception(handler, e):
 
             # Clean up
             del sys.modules[module.__name__]
-            doc._modules.remove(module)
+
+            if hasattr(doc, 'modules'):
+                doc.modules._modules.remove(module)
```
MarcSkovMadsen, line 332: Please test if module in the list or add try catch. I believe I’ve added two PRs so far to catch and handle similar problems. 😉

PR #2645, opened 2021-08-18
Implement profiling page for Panel

philippjfr:
Adds an optional `/admin` endpoint to Panel with profiling output. This will expand to a number of visualization and views to monitor a live Panel server, including seeing currently active sessions, logging output, callbacks, and much more.

Screen Shot 2021-08-18 at 7 59 23 PM


MarcSkovMadsen:
Do I see the terminal in action?

MarcSkovMadsen:
For inspiration check also scalene out @philippjfr . https://github.com/plasma-umass/scalene

philippjfr:
@MarcSkovMadsen scalene indeed looks great but I see no way to actually run it from within an existing Python session. The only way I found to invoke it required forking the process and running a code snippet in the forked process.

doc/user_guide/index.rst
```diff
@@ -65,6 +65,10 @@ when needed.
 
 `Authentication <Authentication.html>`_
  Learn how to add an authentication component in front of your application.
+
+`Profiling and Debugging <Profiling_and_Debugging.html>`_
```
philippjfr, line 69:
```suggestion
`Profiling and Debugging <Performance_and_Debugging.html>`_
```

MarcSkovMadsen:
For me the main take away is also the look and frem which looks great.

Issue #2642, opened 2021-08-18
Tabulator makes wrong selection when combined with a filter


fogo5:
#### Description of expected behavior and the observed behavior

The tabulator widget makes the wrong selection when combined with a filter.

If the table get filted, the selected row is based on the row number selected and not the actual row selected. See screenshots.

The expected behavior: The filtered table, when selected should provide selection and selected_dataframe based on the shown filtered table.

#### Complete, minimal, self-contained example code that reproduces the issue

```
import panel as pn
import pandas as pd
pn.extension('tabulator')


df = pd.DataFrame([['a', 'b', 'c', 'd', 'e'],[1,2,3,5,6]], index = ['name', 'num1']).T
table_widget = pn.widgets.Tabulator(df, disabled = True)
num1_slider = pn.widgets.IntRangeSlider(start=1, end=6, name='num1')
table_widget.add_filter(num1_slider, 'num1')
pn.Row(num1_slider, table_widget)


table_widget.selected_dataframe
table_widget.selection
```

#### Screenshots or screencasts of the bug in action

Working example with NO filtering:
![Before filtering](https://user-images.githubusercontent.com/49350937/129894304-4bb3fb41-35fe-4166-83e5-59457d0a90f5.png)

After filtering, it no longer works:
![After Filtering](https://user-images.githubusercontent.com/49350937/129894364-2966aa31-4c9c-4fed-a6ff-927a9507d47c.png)

#### Software version info
Mozilla Firefox -  91.0 (64-bit)
Google Chrome - Version 92.0.4515.131 (Official Build) (64-bit)

bokeh                     2.3.3            py39h06a4308_0  
conda                     4.10.3           py39hf3d152e_0    conda-forge
ipykernel                 6.1.0            py39hef51801_0    conda-forge
ipython                   7.26.0           py39hef51801_0    conda-forge
jupyter_client            6.1.12             pyhd8ed1ab_0    conda-forge
jupyter_core              4.7.1            py39hf3d152e_0    conda-forge
jupyter_server            1.10.2             pyhd8ed1ab_0    conda-forge
jupyterlab                3.1.6              pyhd8ed1ab_0    conda-forge
jupyterlab_server         2.7.0              pyhd8ed1ab_0    conda-forge
numpy                     1.20.3           py39hf144106_0  
pandas                    1.3.1            py39h8c16a72_0  
panel                     0.12.1                     py_0    pyviz
param                     1.11.1                     py_0    pyviz
[...]

PR #2676, opened 2021-08-27
correctly accessing the filtered dataframe for selection of tabulator…

Stubatiger:
Accessing now ._processed in the selected_dataframe property as discussed in #2642


philippjfr:
Thanks @Stubatiger!

Issue #2682, opened 2021-08-28
ReactiveHTML loop not working if option tag not included


MarcSkovMadsen:
I'm on Panel 0.12.1.

While trying to create an [Accordion Layout from the ReactiveHTML](https://discourse.holoviz.org/t/how-do-i-create-a-layout-accordion-from-reactivehtml/2730) I could not get the jinja loop working without including the `<option id="option">` tag. 

For example when creating a layout my list of panels are not options. They are just a list of panels to be inserted.

## Works - Using option tag

I'm using

```python
    _template = """
{% for obj in objects %}<option id="option">${obj}</option>{% endfor %}
"""
```

and it works

![image](https://user-images.githubusercontent.com/42288570/131208677-bd65f624-4dd9-48b3-9e46-5bafcabb7441.png)

```python
import panel as pn
import param

pn.extension("vega", sizing_mode="stretch_width")

class FastAccordion(pn.reactive.ReactiveHTML):
    objects = param.List()


    _template = html ="""
{% for obj in objects %}<option id="option">${obj}</option>{% endfor %}
"""

    def __init__(self, *objects, **params):
        params["objects"]=[pn.panel(obj) for obj in objects]
        super().__init__(**params)

component1 = pn.Column("a", pn.Spacer(background="blue"),
    background="yellow", name="Component 1")
vegalite = {
  "$schema": "https://vega.github.io/schema/vega-lite/v4.json",
  "data": {"url": "https://raw.githubusercontent.com/vega/vega/master/docs/data/barley.json"},
  "mark": "bar",
  "encoding": {
    "x": {"aggregate": "sum", "field": "yield", "type": "quantitative"},
    "y": {"field": "variety", "type": "nominal"},
    "color": {"field": "site", "type": "nominal"},
  },
  "width": 'container',
}
vgl_pane = pn.panel(vegalite, height=240)

accordion = FastAccordion(component1, vgl_pane, "d")

pn.template.FastListTemplate(
    title="Awesome Panel",
    main = [accordion, vgl_pane],
    main_layout="",
).servable()
```

## Does not work - Not Using option tag

If I use the template

```pyton
_template = """
{% for obj in objects %}${obj}{% endfor %}
"""
```

I see

```bash
2021-08-28 08:29:33,155 error handling message
 message: Message 'PULL-DOC-REQ' content: {}
 error: TypeError('Object of type Column is not JSON serializable')
Traceback (most recent call last):
[...]

PR #2683, opened 2021-08-28
Adds a comment about the ReactiveHTMLL loop wrapper element

MarcSkovMadsen:
Addresses https://github.com/holoviz/panel/issues/2682

### DIFF
--- a/examples/user_guide/Custom_Components.ipynb
+++ b/examples/user_guide/Custom_Components.ipynb
@@ -153,7 +153,9 @@
     "  <option id=\"option\">${obj}</option>\n",
     "  {% endfor %}\n",
     "</select>\n",
-    "```"
+    "```\n",
+    "\n",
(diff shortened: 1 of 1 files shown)


examples/user_guide/Custom_Components.ipynb
```diff
@@ -153,7 +153,9 @@
     "  <option id=\"option\">${obj}</option>\n",
     "  {% endfor %}\n",
     "</select>\n",
-    "```"
+    "```\n",
+    "\n",
+    "Here we wrap the `${obj}` in the `<option id=\"option\">` element. But depending on your use case, you can use other tag on the form `<name-of-tag id=\"option\">`."
```
philippjfr, line 158: The important thing to note here is that the wrapping element **must** have an `id`. It's also worth noting that the only reason to use an explicit loop (rather that just inserting `${children_parameter}`) is to be able to wrap each component in a custom tag.

philippjfr:
Clarified this in https://github.com/holoviz/panel/pull/2689

Issue #2685, opened 2021-08-28
Progress indicator inconsistencies


hyamanieu:
Panel: master ac04ed1c & 0.12
param: 1.10.2a13

The Progress indicator is a bit frustrating: 

- the documentation indicates indeterminate state shall be set by setting `value` to None, however `param.value` doesn't allow None.
I know that setting default to None will make it allow None. however if setting from an int back to None, it throws an error (separate issue within `param` or bokeh).
- setting from an int to None throws an error. If setting a second time, no error is thrown but it doesn't turn to indeterminate mode.
- When instantiating, setting `value` to 0 will render an indeterminate bar rather than an empty one. This lead to the tricks of having a value of -1 or having a very large max like here: https://github.com/holoviz/panel/blob/ac04ed1c08a993cd678fb162deb5fa567184ec40/panel/widgets/indicators.py#L789-L792 .

I would like to submit the following changes:

- -1 shall be used for indeterminate state rather than empty.
- 0 and None (null) should make an empty bar
- None should throw a FutureWarning as deprecated then Default value should be -1

I'll submit a PR.

jbednar:
Can you be more specific about the issues with None, with some reproducible code? The default is None and therefore allow_None should be true already. Is there some bug here being worked around?

hyamanieu:
@jbednar please see here below. Yes it's a bug being worked around, and it seems to happen at bokeh level, not param now I'm taking a closer look to the stack trace further below.

browsers: firefox, chromium, brave
os: Ubuntu 18.04
panel: 0.12.1
param: 1.11.1
python: 3.8

The bug only happens if the progress bar is displayed (either with `display` in a notebook like in my example or externally with `progress.show()` ). It also happens with the current master branch.

That bug or not, it's not possible to reset the bar in an indeterminate mode in the current branch anyways. It could be solved by 
using `this.progressEl.removeAttribute("value")` within `progress.ts`, but that won't fix the issue with "0" interpreted as a null by javascript. For this I couldn't find a fix. I would vote for note using None at all.

```
import panel as pn
pn.extension()

import param


print('panel version:', pn.__version__)

print('param version:', param.__version__)



progress = pn.widgets.Progress(value=None, bar_color='warning')
display(progress)

progress.value = 18

print('Is None allowed? ', progress.param.value.allow_None)

progress.value = -1

progress.value = 0

progress.value = None
```

Output:

```
panel version: 0.12.1
param version: 1.11.1

[bar]
[...]

Issue #2196, opened 2021-04-16
`warn` inaccurately documented for bar_color in Progress


ahuang11:
```
bar_color (str): The color of the bar, one of ‘primary’, ‘secondary’, ‘success’, ‘info’, ‘warn’, ‘danger’, ‘light’, ‘dark’
```

ValueError: warn not in Parameter bar_color's list of possible objects, valid options include [primary, secondary, success, info, danger, warning, light, dark]

hyamanieu:
Can be closed.

PR #2686, opened 2021-08-28
Make progress indicator more consistent

hyamanieu:
Fixes https://github.com/holoviz/panel/issues/2685

### DIFF
--- a/panel/models/progress.ts
+++ b/panel/models/progress.ts
@@ -49,8 +49,15 @@ export class ProgressView extends HTMLBoxView {
   }
 
   setValue(): void {
-    if (this.model.value != null)
+    console.log(this.model.value)
+    console.log(this.progressEl.value)
+    if (this.model.value == null)
+      this.progressEl.value = 0
+    else if (this.model.value >= 0)
       this.progressEl.value = this.model.value
+    else if (this.model.value < 0)
+      this.progressEl.removeAttribute("value")
+    console.log(this.progressEl)
   }
 
   setMax(): void {
--- a/panel/models/progress.ts
+++ b/panel/models/progress.ts
@@ -49,8 +49,6 @@ export class ProgressView extends HTMLBoxView {
   }
 
   setValue(): void {
-    console.log(this.model.value)
-    console.log(this.progressEl.value)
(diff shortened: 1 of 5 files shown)


hyamanieu:
Wrong issue mentioned in my commit, it actually fixes #2196

panel/models/progress.ts
```diff
@@ -49,8 +49,15 @@ export class ProgressView extends HTMLBoxView {
   }
 
   setValue(): void {
-    if (this.model.value != null)
+    console.log(this.model.value)
+    console.log(this.progressEl.value)
```
philippjfr, line 53: Can you remove the log statements?

panel/widgets/indicators.py
```diff
@@ -786,13 +796,8 @@ def __init__(self, **params):
 
         self.param.watch(self._update_layout, list(Viewable.param))
 
-        if self.value == 0:
-            # Hack: to give progress the initial look
```
philippjfr, line 790: Definitely didn't like this hack but is this condition handled, i.e. under what configuration do you get an indeterminate bar and under what condition do you get an empty bar now?

hyamanieu: In the current state, Tqdm cannot have a value of None, so no indeterminate bar, nor -1, so no empty bar (explaining the trick).
I am proposing in my last commit to simply copy the params `value` and `max` from Progress and set default to 0 for value. So now Tqdm can have an indeterminate state and no hack is programmed.

panel/models/progress.ts
hyamanieu: Removed in last commit

panel/widgets/indicators.py
hyamanieu: To repond to your question: now Tqdm can be indeterminate with a value of -1 (not happening in the gallery examples) and be empty with a value of 0.

philippjfr: Okay, will check out the PR and play around. The Progress indicator can be used independently of Tqdm so I want to make sure that still works as expected.

philippjfr:
Thanks @hyamanieu, appreciate it!

Issue #2702, opened 2021-09-02
Cannot panel serve on Master branch - I.e. cannot contribute


MarcSkovMadsen:
For the last couple of weeks when I have been trying to start a contribution from the MASTER branch it has not been possible.

Also today when I `git pull`, `pip install -e . -U` and then `panel serve script.py` I see 

```python
$ panel serve 'script.py' --autoreload --show
ERROR: 'Document' object has no attribute '_callbacks'
```

It is highly problematic that the master branch does not work because 1) Nobody can start contributing 2) We have lots of binder links linking to the MASTER branch. I.e. our documentation does not work

#### Solution

1) Get Master branch working
2) Replace binder links with links to the latest release. It does not make sense to link binder to something that is under development and might not reflect the documentation.

#### Example

```python
import plotly.express as px

import panel as pn

pn.sizing_mode="stretch_width"


def get_plotly_carshare_data():
    return px.data.carshare()


def plotly_carshare_plot(carshare):
    fig = px.scatter_mapbox(
        carshare,
        lat="centroid_lat",
        lon="centroid_lon",
        color="peak_hour",
        size="car_hours",
        color_continuous_scale=px.colors.cyclical.Phase,
        size_max=15,
        zoom=10,
        mapbox_style="carto-positron",
    )
    fig.layout.autosize = True
    return fig


data = get_plotly_carshare_data()
component = plotly_carshare_plot(data)
carshare_panel = pn.pane.Plotly(
    component,
    config={"responsive": True},
    background="yellow",
)

pn.Column(
    component,
    background="lightgray",
    sizing_mode="stretch_both",
).servable()
```

philippjfr:
Which Bokeh version are you running, as I can't reproduce this.

philippjfr:
Yeah, absolutely can't reproduce even after I created a new environment and ran `pip install -e . -U`

philippjfr:
Will close but please reopen if you figure out what was causing this or how you created your environment.

philippjfr:
Also just tried the master deployment on Binder without isssue.

philippjfr:
Still would like to know what happened here, any ideas @MarcSkovMadsen ?

MarcSkovMadsen:
Nope

philippjfr:
Anyway, appreciate the issues and glad they resolved themselves. Transitioning between Bokeh versions is the likely culprit, master will switch to being >=2.4 very soon.

PR #2703, opened 2021-09-02
Change binder link from master to binder branch for robustness

MarcSkovMadsen:
Addresses suggestion 2 of https://github.com/holoviz/panel/issues/2702
-----

The master branch is always changing which means

- its not consistent with the docs for the current release
- Starting Binder takes a long time because it has to build a new image
- might not work because the master branch does not work.

I suggest changing to a branch called `binder`.

I've already created the `binder` branch from the `v0.12.1` (`git checkout tags/v0.12.1`) and tested that it works.

You can test binder via https://mybinder.org/v2/gh/holoviz/panel/binder?urlpath=lab/tree/examples

## Release Procedure

Please note that when ever a new version of Panel is released a step in that process is to merge the release into the `binder` branch and manually test that binder can build the image and that things works. But it would be a good test anyways that Panel indeed works in a fresh environment and that all the gallery apps works etc.

## Improvements to binder branch

Any improvements to the binder branch during releases should start from the binder branch and not from master.


philippjfr:
Shouldn't we point it to the last release instead?

README.md
```diff
@@ -11,7 +11,7 @@
...
 | Docs | [![gh-pages](https://img.shields.io/github/last-commit/holoviz/panel/gh-pages.svg)](https://github.com/holoviz/panel/tree/gh-pages) [![site](https://img.shields.io/website-up-down-green-red/https/panel.holoviz.org.svg)](https://panel.holoviz.org) |
-| Binder | [![Binder](https://mybinder.org/badge_logo.svg)](https://mybinder.org/v2/gh/holoviz/panel/master?urlpath=lab/tree/examples) |
+| Binder | [![Binder](https://mybinder.org/badge_logo.svg)](https://mybinder.org/v2/gh/holoviz/panel/binder?urlpath=lab/tree/examples) |
```
philippjfr, line 14:
```suggestion
| Binder | [![Binder](https://mybinder.org/badge_logo.svg)](https://mybinder.org/v2/gh/holoviz/panel/v0.12.1?urlpath=lab/tree/examples) |
```

Issue #1514, opened 2020-08-05
Responsive mode for Plotly plots not working in practice


MarcSkovMadsen:
Panel: 0.9.7
Plotly: 4.9.0

#### My Pain

One of the promises of Panel and what sets is aside from for example Dash is that it supports a wide range of Python plotting libraries. But I don't think Plotly is truly supported as responsive mode is not well supported. And responsive mode is very important for modern apps.

Concretely I'm trying to show case the power of the awesome panel designer. Not only to design Panel components and applications. But in general as an interactive, exploratory environment for developing python Data Viz using your favorite editor.

But if I want to develop plotly plots they look like below.

![image](https://user-images.githubusercontent.com/42288570/89371132-c0927c80-d6e2-11ea-918c-5bf6a6bc3082.png)

If I select another component in the drowdown on the left and select the plotly_carshare_plot component it looks like.

![image](https://user-images.githubusercontent.com/42288570/89371356-6c3bcc80-d6e3-11ea-835f-afc46690787c.png)

But as you can see there are still problems. This time the plotly plot expands outside of its container both downwards and to the right (I've added a border to the container, so that you can see it.).

If I look at the console I see error messages from Plotly.

![image](https://user-images.githubusercontent.com/42288570/89371311-46162c80-d6e3-11ea-8e0b-0468cc225a03.png)

#### Minimum Reproducible Example

I have not been able to reproduce the behaviour shown in the first image in a small example. But I can show that it outgrows the container its in.

![image](https://user-images.githubusercontent.com/42288570/89371449-b45aef00-d6e3-11ea-97d3-63d1f5659329.png)

Furthermore if I change the size of the browser window the Plotly plot does not resize. I.e. it is not truly responsive.

![image](https://user-images.githubusercontent.com/42288570/89371483-cd63a000-d6e3-11ea-9fc5-ef1dbae1fd10.png)

![image](https://user-images.githubusercontent.com/42288570/89371493-d48aae00-d6e3-11ea-869f-9cb69672990b.png)

here I don't get errors in the console

![image](https://user-images.githubusercontent.com/42288570/89371576-14ea2c00-d6e4-11ea-8502-fd650e9edd50.png)

```python
import plotly.express as px

import panel as pn

STYLE = """
.bk.designer-centered-component {
    color: black;
    border-color: #9E9E9E;
    border-width: 1px;
    border-style: solid;
}
"""
pn.config.raw_css.append(STYLE)


def get_plotly_carshare_data():
[...]

PR #2704, opened 2021-09-02
Make Plotly pane resize when window resizes

MarcSkovMadsen:
Fixes https://github.com/holoviz/panel/issues/1514

I've implemented the fix and tested that it works

https://user-images.githubusercontent.com/42288570/131785863-c23d5c29-0eea-4292-b594-e38965f58220.mp4


philippjfr:
Great!

PR #2711, opened 2021-09-03
Add the version number in the binder badge

maximlt:
Which should help us remember that the link has to be updated after a new release.


philippjfr:
Nice, thanks!

Issue #89, opened 2018-10-08
Add examples of Graph/Network rendering


jbednar:
As of PRs #85 and #88, dot-based graph plots are now supported in Panel, either as SVG representations of GraphViz Graph/DiGraph objects, or as PNG representations of NetworkX objects created by `nxpd.draw()`:

```
from graphviz import Graph
hw = Graph('hello', format='svg')
hw.edge('Hello', 'World')

import networkx as nx, nxpd
G = nx.cycle_graph(4, create_using=nx.DiGraph())
d = nxpd.draw(G, show='ipynb')

import panel as pn
pn.Row(hw, "Text", d, "Text")
```
![image](https://user-images.githubusercontent.com/1695496/46636996-db642c00-cb1f-11e8-92db-af247895fbbc.png)

Remaining issues:

- [x] Spacing for these objects is not currently automatically determined
- [ ] It would be nice to provide graph examples in the User Guide

philippjfr:
nxpd seems like a pretty niche library to consider here. I'd just draw the networkx graph using pygraphviz:

```
from graphviz import Graph
hw = Graph('hello', format='svg')
hw.edge('Hello', 'World')

import networkx as nx
G = nx.complete_graph(5)
B = nx.nx_agraph.to_agraph(G)

import panel as pn
pn.Row(hw, "Text", A.draw(prog='dot', format='svg').decode('utf-8'), "Text")

**Comment by jbednar** [comment]
A.draw?

**Comment by philippjfr** [comment]
Fixed.

**Comment by MarcSkovMadsen** [comment]
The above example does not work for me. The output of `A.draw` needs to be clean by removing things in front of `<svg`.

This works

```python
from graphviz import Graph
import panel as pn
import networkx as nx

graphviz_graph = Graph('hello', format='svg')
graphviz_graph.edge('Hello', 'World')

graph = nx.complete_graph(5)
pyviz_graph = nx.nx_agraph.to_agraph(graph)

def clean_svg(svg):
    svg.find("<svg")
    svg_start = svg.find("<svg")
    return svg[svg_start:]

networkx_svg = clean_svg(pyviz_graph.draw(prog='dot', format='svg').decode('utf-8'))

pn.Row(
    graphviz_graph,
    "Text",
    networkx_svg,
    "Text").servable()
```

MarcSkovMadsen:
#2732 Should close this one.

PR #2732, opened 2021-09-12
Gallery Examples for Graphviz and NetworkX

MarcSkovMadsen:
Addresses https://github.com/holoviz/panel/issues/89 by adding GraphViz and NetworkX gallery examples.

I've added a new `viz` folder. The vision would be to add examples for all the *viz tools you know and love* that are not already covered by the User Guide and Reference Guides. The purpose is to 1) Demonstrate the power of Panel 2) To make it very easy for users to get started with the tool they know and love. 

An example could be seaborn. And you might say "but there is already a Matplotlib reference guide". But I don't think that is enough. Some users might not know or think of that if Matplotlib is supported then Seaborn is as well. Furthermore there is a bit of friction to figure out how to get the Figure object in Seaborn etc. I also have code with styling for `default` and `dark` theme. And having that in the example would just make it so much easier for users to get a working application.

## Todo

[x] Check that it works on Binder: https://mybinder.org/v2/gh/holoviz/panel/graphviz-gallery-example?urlpath=lab/tree/examples

### DIFF
--- a/examples/gallery/viz/GraphViz.ipynb
+++ b/examples/gallery/viz/GraphViz.ipynb
@@ -1,0 +1,127 @@
+{
+ "cells": [
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "fc69e05f-be77-4c75-9157-60c60dcba6a5",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "from graphviz import Graph\n",
+    "import panel as pn\n",
+    "\n",
+    "pn.extension(sizing_mode=\"stretch_width\")"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "fe92851c-51a2-4684-97a5-a1603da30ff2",
+   "metadata": {},
+   "source": [
+    "# Panel and GraphViz\n",
+    "\n",
+    "The purpose of this example is to show how easy it is to use [GraphViz](https://graphviz.readthedocs.io/en/stable/manual.html#) with Panel."
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "15f4789d-68d2-43d9-9791-8b7659f85be4",
+   "metadata": {},
+   "source": [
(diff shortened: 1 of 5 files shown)


MarcSkovMadsen:
Still need to find out how to get graphviz installed with conda for use on binder

image

MarcSkovMadsen:
networkx-speedup

graphviz-example-speedup

MarcSkovMadsen:
Works on Binder

image

MarcSkovMadsen:
I would like to know

- Would it be an ok vision/ ambition to create a folder showcasing the "tools you know and love" with Panel?
- Should we add recommendations on when to use the tools? Or which alternatives to check out?

philippjfr:
Not ready to merge so I won't get this into 0.12.2.

examples/gallery/viz/GraphViz.ipynb
```diff
@@ -0,0 +1,128 @@
+{
+ "cells": [
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "fc69e05f-be77-4c75-9157-60c60dcba6a5",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "from graphviz import Graph\n",
```
philippjfr, line 10: This requires `python-graphviz` but you are installing `pygraphviz`, they are distinct libraries.

MarcSkovMadsen: I must admit I am confused by the graphviz/ pygraphviz/ pygraphviz and pip vs conda story. 

But at least it works on Binder **without** conda installing `python-graphviz` and by pip installing `graphviz`+`pygraphviz`. So I would claim `python-graphviz` is not absolute required ??? The problem with `python-graphviz` is that its a conda only thing and wont be available for use in the `setup.py` file.

I am simple in doubt on what to do here because of all that confusion. Can you help @philippjfr?

PR #2752, opened 2021-09-17
WIP: Add ability to define custom tornado handler plugins

philippjfr:
Allows defining custom Tornado handler plugins via the `panel serve` CLI.

As a demo this also declares a `TileHandler` which allows users to ship a file alongside an app which renders tiles when given a request.

### nyc_taxi_tiles.py

```python
import datashader as ds

import dask.dataframe as dd

ddf = dd.read_parquet('~/development/datashader/examples/data/nyc_taxi.parq/').persist()

x_range = dd.compute(ddf.pickup_x.min(), ddf.pickup_x.max())
y_range = dd.compute(ddf.pickup_y.min(), ddf.pickup_y.max())

def aggregate(x_range, y_range, width, height):
    cvs = ds.Canvas(
        width, height, x_range, y_range, 
    )
    agg = cvs.points(ddf, 'pickup_x', 'pickup_y')
    img = ds.tf.dynspread(ds.tf.shade(agg))
    return img 
```

### nyc_tile_viewer.py

```python
import holoviews as hv
import panel as pn

hv.extension('bokeh')

pn.Row(
    hv.Tiles('http://localhost:5006/tiles/nyc_taxi_tiles/{X}/{Y}/{Z}.png').opts(
        responsive=True,
        xlim=(-20037508.34, 20037508.34),
        ylim=(-20037508.34, 20037508.34)
    ),
    sizing_mode='stretch_both'
).servable()
```

Now this can be run with:

```
panel serve nyc_tile_viewer.py --plugins="{'/tiles/(.*)':'panel.plugins.tiles.TileHandler'}"
```

Screen Shot 2021-09-17 at 11 16 33 PM

### DIFF
--- a/panel/command/serve.py
+++ b/panel/command/serve.py
@@ -5,6 +5,7 @@ ways.
 
 import ast
 import base64
+import importlib
 import logging # isort:skip
 import os
 
@@ -168,6 +169,10 @@ class Serve(_BkServe):
         ('--autoreload', dict(
             action  = 'store_true',
             help    = "Whether to autoreload source when script changes."
+        )),
+        ('--plugins', dict(
+            action  = 'store',
+            type    = str
         ))
     )
 
@@ -274,6 +279,15 @@ class Serve(_BkServe):
             pattern = REST_PROVIDERS['param'](files, 'rest')
             patterns.extend(pattern)
             state.publish('session_info', state, ['session_info'])
+
+        if args.plugins:
+            plugins = ast.literal_eval(args.plugins)
+            for slug, plugin in plugins.items():
+                plugin_parts = plugin.split('.')
+                mod, cls = '.'.join(plugin_parts[:-1]), plugin_parts[-1]
+                handler = getattr(importlib.import_module(mod), cls)
+                print(slug, handler)
+                patterns.append((slug, handler))
 
         if args.oauth_provider:
             config.oauth_provider = args.oauth_provider
--- a/panel/plugins/tiles.py
+++ b/panel/plugins/tiles.py
@@ -1,0 +1,67 @@
+import io
+import importlib
+
+from collections import Counter
+from pathlib import Path
+
+from datashader import Canvas
+from datashader.tiles import TileRenderer, MercatorTileDefinition
+
+from tornado.web import RequestHandler, HTTPError
+
+CACHE = Path('tile_cache')
+
+class TileProvider:
+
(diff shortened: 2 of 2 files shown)


philippjfr:
pre-commit.ci autofix

MarcSkovMadsen:
Is it possible to add and remove the handlers dynamically?

How does this compare to the REST API already supported?

Can a user use Async? How? I would probably prefer that.

How to get this to scale? Now you have Panel on web socket where you probable use num-procs and/ or num-threads?

What about oauth? Jwt tokens? Swagger api docs?

Would it be possible and better to build on FastApi?

philippjfr:
>Is it possible to add and remove the handlers dynamically?

Not with this PR no.

>How does this compare to the REST API already supported?

This will let you build generic endpoints to do whatever you want unlike the REST functionality which is much more limited.

>Can a user use Async? How? I would probably prefer that.

Tornado handler HTTP methods like `.get` can be synchronous or asynchronous, it's entirely up to you.

>How to get this to scale? Now you have Panel on web socket where you probable use num-procs and/ or num-threads?

When using num-procs this endpoint will be duplicated across processes and the requests will be routed just like any other request.

>What about oauth? Jwt tokens? Swagger api docs?

If you've enabled authentication adding the @authenticated decorator will put the endpoint behind OAuth.

>Would it be possible and better to build on FastApi?

If someone decides to reimplement the entire Bokeh server functionality on top of FastAPI I'd be very happy but I'd estimate that would take me 1-2 months.

panel/command/serve.py
```diff
@@ -371,6 +390,28 @@ def customize_kwargs(self, args, server_kwargs):
...
                     "or via explicit argument, not both."
                 )
+
+        for plugin in (args.plugins or []):
+            try:
+                with add_sys_path('./'):
+                    plugin_module = importlib.import_module(plugin)
+            except ModuleNotFoundError:
+                import pdb
+                pdb.set_trace()
```
philippjfr, line 400:
```suggestion
```

panel/command/serve.py
```diff
@@ -264,6 +278,10 @@ class Serve(_BkServe):
             help    = "The endpoint for the liveness API.",
             default = "liveness"
         )),
+        ('--plugins', dict(
+            action  = 'append',
+            type    = str
+        ))
```
hoxbro, line 284:
```suggestion
        )),
```

PR #2753, opened 2021-09-19
Correct supported python versions

MarcSkovMadsen:
I don't know exactly which python version we support. But at least we don't support python 2.7 anymore. I figured we support the ones that we test so it 3.7, 3.8, 3.9.

I saw this because I'm reinstalling panel. Cannot `panel serve`. Something about missing `doc` `_callbacks` again.


philippjfr:
Latest panel requires Bokeh 2.4, but that is pinned so I don't know why you're getting the document callbacks error since it should have forced you to Upgrade.

PR #2757, opened 2021-09-19
Adding ICO and pathlib support for image

Hoxbro:
Added a `ICO` class to `pane.image` and `pathlib` conversion to pane.image and favicon in templates.

### DIFF
--- a/panel/pane/__init__.py
+++ b/panel/pane/__init__.py
@@ -13,7 +13,7 @@ from .echarts import ECharts # noqa
 from .holoviews import HoloViews, Interactive # noqa
 from .idom import IDOM # noqa0
 from .ipywidget import IPyWidget # noqa
-from .image import GIF, JPG, PDF, PNG, SVG # noqa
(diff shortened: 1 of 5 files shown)


philippjfr:
Not sure it's worth it to make a reference gallery item for ICO but we need to exclude it from the unit test that checks such an example exists.

Hoxbro:
I don't either see the need for a reference gallery item. 

Should I also make the logo in BasicTemplate accept pathlib?

philippjfr:
>Should I also make the logo in BasicTemplate accept pathlib?

Yes, but maybe in a separate PR.

MarcSkovMadsen:
Without a reference example, how would users know its supported? And what potential arguments are available?

Hoxbro:
I'm not opposed to adding a reference gallery of ICO. 

Though, I don't think anybody will ever use an ICO pane directly but will rather pass a path or URL of an `.ico` file into a theme which will then just work. E.g., like what you did with awesome-panel.

PR #2759, opened 2021-09-19
Added check for bool in parse_query

### LINKED ISSUE #2758: Sync location does not work with checkbox
(this library, plus any other relevant software, e.g. bokeh, python, notebook, OS, browser, etc)
``` yaml
panel	0.13.0a5.post8+g80a66c3f
bokeh	2.4.0
OS	Windows	10
```

The sync location does not work when using a checkbox. 

I would gladly help solving the problem but I need a small hint of where in the code to look. 

``` python
import panel as pn

box = pn.widgets.Checkbox(value=False)
pn.state.location.sync(box)
box.servable()

```

image

Hoxbro:
Fixes #2758

TODO:
- [x] Write/update a unittest

### DIFF
--- a/panel/tests/test_util.py
+++ b/panel/tests/test_util.py
@@ -4,7 +4,7 @@ from bokeh.models import Div
 
 from panel.io.notebook import render_mimebundle
 from panel.pane import PaneBase
-from panel.util import get_method_owner, abbreviated_repr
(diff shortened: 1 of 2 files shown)


philippjfr:
Thanks @Hoxbro !

Issue #2755, opened 2021-09-19
ERROR: 'Document' object has no attribute '_callbacks'


MarcSkovMadsen:
I've just installed Panel from scratch according to the developer guide https://panel.holoviz.org/developer_guide/index.html.

When I `panel serve 'script.py' --autoreload --show` I get

```bash
ERROR: 'Document' object has no attribute '_callbacks'
```

(I installed this from scratch as I got the same error in my existing installation).

## script.py

```python
import panel as pn
import pandas as pd
import param

pn.extension()

data = pd.DataFrame()

class MyDataClass(param.Parameterized):
    data = param.DataFrame()

my_data = MyDataClass()

button = pn.widgets.Button(name="UPDATE")

@param.depends(button.param.value)
def update(value):
    my_data.data = pd.DataFrame({"x": [1]*10})

pn.Column("# Test", pn.Param(my_data), button).servable()
```

maximlt:
What version of bokeh do you have?

maximlt:
Without https://github.com/holoviz/panel/pull/2741 being merged it's likely that you end up with a 2.3 version instead of the actually required 2.4 one.

Hoxbro:
I'm having the same problem and I'm running on master branch of panel and bokeh version 2.4.0. The problem only happens when running `panel serve` with `--autoreload`. The error is coming from: https://github.com/holoviz/panel/blob/80a66c3f03b255c417dc0dc246e41b360fe5d9c4/panel/command/serve.py#L39

philippjfr:
Damn, looks like we indeed missed a compatibility change with Bokeh 2.4 and will likely have to release **another** patch release (i.e. 0.12.4).

PR #2760, opened 2021-09-20
Ensure cleanup happens when enabling --warm or --autoreload with Bokeh 2.4

philippjfr:
Fixes https://github.com/holoviz/panel/issues/2755

### DIFF
--- a/panel/tests/command/test_serve.py
+++ b/panel/tests/command/test_serve.py
@@ -37,6 +37,8 @@ def run_panel_serve(args):
 class NBSR:
     def __init__(self, stream) -> None:
         '''
+        NonBlockingStreamReader
+
         stream: the stream to read from.
                 Usually a process' stdout or stderr.
         '''
@@ -46,7 +48,7 @@ class NBSR:
 
         def _populateQueue(stream, queue):
             '''
-            Collect lines from 'stream' and put them in 'quque'.
+            Collect lines from 'stream' and put them in 'queue'.
             '''
 
             while True:
@@ -74,6 +76,7 @@ def test_autoreload_app(py_file):
     app2 = "import panel as pn; pn.Row('# Example 2').servable(title='B')"
     py_file.file.write(app)
     py_file.file.flush()
+    os.fsync(py_file.file)
 
     py_file.file.seek(0)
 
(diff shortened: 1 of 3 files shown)


panel/tests/command/test_serve.py
```diff
@@ -0,0 +1,108 @@
...
+            pytest.fail("no matching log line in process output")
+        port = int(m.group(1))
+
+        r = requests.get(f"http://localhost:{port}/{app_name}")
+        assert r.status_code == 200
+        assert "<title>A</title>" in r.content.decode('utf-8')
+
+        py_file.file.write(app2)
+        py_file.file.flush()
+
```
maximlt, line 104: Shouldn't it wait here for a little while?

panel/tests/command/test_serve.py
```diff
@@ -0,0 +1,108 @@
...
+    try:
+        yield p
+    except Exception as e:
+        p.terminate()
+        p.wait()
+        print("An error occurred: %s", e)
+        try:
+            out = p.stdout.read().decode()
+            print("\n---- subprocess stdout follows ----\n")
+            print(out)
```
maximlt, line 27: Are these print statements captured by pytest and displayed when running the tests?

philippjfr: They are if the test fails.

panel/tests/command/test_serve.py
philippjfr: Flushing the write buffer should really be sufficient but who knows what it's doing on some platforms.

Hoxbro:
I'm seeing some (new) problems with this fix.

I have this code 
``` python
import panel as pn
pn.panel("Test").servable()
```
Which I run with `panel serve race_condition.py --autoreload`, but when I reload the page multiple times I get this error:

``` bash
2021-09-20 22:06:31,569 DocumentLifeCycleHandler on_session_destroyed callback <bound method PeriodicCallback._cleanup of PeriodicCallback(callback=functools.partial(<function _reload_on_update at 0x00000250A9488AF8>, {'race_condition.py': 1632167947.1386704}), count=None, name='PeriodicCallback00196', period=500, running=False, timeout=None)> failed with following error: 'DocumentCallbackManager' object has no attribute '_change_callbacks'
```

**Screenshot**


**Screencast**
https://user-images.githubusercontent.com/19758978/134068656-65f0a758-39b8-47d6-8952-3e358cf988e7.mp4

PR #2772, opened 2021-09-22
Link an assets image instead of an embedded one for the Markdown reference pane

### LINKED ISSUE #2734: Markdown Reference Example image url is not valid
https://panel.holoviz.org/reference/panes/Markdown.html

image

maximlt:
Fixes #2734 

It seems like `attachments` stored within notebooks, such as in this case an image, aren't supported by the build tools. The workaround here was to just save that image in `/assets/` along with the other images and link it in the notebook.


philippjfr:
Oh wow, should have caught that attachments thing.

PR #2797, opened 2021-09-30
Comm channel docs

jbednar:
Added a new user guide describing comms channels that can be used with Panel.

I'm not sure what to call this info, nor if every bit of it is accurate, but I think we need something like this. I'm not sure whether to name it "Communication channels" or "Comms for Panel" or the current "Jupyter and Servers"; suggestions appreciated. Whatever it's called, it documents which libraries or modules supply the Python/JS communications in lots of different contexts. There are a lot of such modules, and they fail regularly, so it's important to keep track of them all!

It would be great if someone could proofread it carefully and make sure I didn't make any mistakes or capture any misunderstandings of mine. Additional related content to add here would also be very helpful!

The rendered notebook can be viewed at https://github.com/holoviz/panel/blob/cdf6fdfb4c29100b9117513d543925a31ae3fc3c/examples/user_guide/Jupyter_and_Servers.ipynb


jbednar:
- [ ] Should probably move much of the Usage section of the main doc/indext.rst guide here, and point to it from the homepage. The content is related, though on the homepage the question is about "how do I develop using panel" in the context I want, and here the emphasis is on "what works where, and whose job is it to make it work".

examples/user_guide/Jupyter_and_Servers.ipynb
```diff
@@ -2,10 +2,10 @@
  "cells": [
   {
    "cell_type": "markdown",
-   "id": "9258841f",
+   "id": "8d78f011",
```
philippjfr, line 5: NOOOOOOO!

Issue #2800, opened 2021-10-04
Updating a filtered Tabulator with "remote" pagination


samuelyeewl:
There is a bug while editing a Tabulator widget, when pagination is set to "remote" and filters are applied.
When editing a cell, although the correct cell is updated in the browser, the dataframe is updated based on the indices of the _unfiltered_ dataframe. This overwrites the entire column's worth of values.
This only occurs when "remote" pagination is enabled, but works when the pagination mode is "local".

#### ALL software version info
Panel 0.12.4, Python 3.8.10

#### Description of expected behavior and the observed behavior
Expected behavior: Correct cell in the dataframe is modified when the tabulator widget is modified.
Observed behavior: The dataframe row that is modified is the one with the same index in the unfiltered table, when it should be the filtered index. 

#### Complete, minimal, self-contained example code that reproduces the issue

```
import pandas as pd
import panel as pn
import panel.widgets as pnw
pn.extension('tabulator')

# Second column is "A" for even indices, "B" for odd indices.
test_df = pd.DataFrame(data={'col_one': list(range(10)),
                             'col_two': ['A', 'B']*5})
table = pnw.Tabulator(test_df, page_size=5, pagination='remote', height=200)

# Add a filter
col_two_selector = pnw.TextInput(value='A')
table.add_filter(col_two_selector, column='col_two')
pn.Column(col_two_selector, table)
```

#### Screenshots or screencasts of the bug in action
![Initial state of Tabulator widget](https://user-images.githubusercontent.com/7497009/135929042-f5644677-4f09-4a8d-a74b-2be7deecfd46.png)
Initial state of Tabulator widget

![Modified Tabulator widget](https://user-images.githubusercontent.com/7497009/135929087-b4a0be62-3cf2-4f91-917d-215afe5c6d2b.png)
Third row of filtered dataframe, corresponding to index 4, is modified.

![Final Dataframe state](https://user-images.githubusercontent.com/7497009/135929181-ee7870a6-8b57-4872-8a0d-43ba5ee4a38b.png)
Final dataframe state has second column "AACAA", when it should be ABCBA for the first five rows. Those rows were overwritten by the values in the filtered view.

samuelyeewl:
I've discovered that the problem arises because of [these lines](https://github.com/holoviz/panel/blob/3624b5f5074e5366ab82945878db6d22b370aec6/panel/widgets/tables.py#L987-L991) in the code:
```
        if self.sorters:
[...]

PR #2801, opened 2021-10-05
Fixed tabulator remote pagination bug.

samuelyeewl:
This is a fix for bug #2800.


philippjfr:
This looks fine to me, I guess the previous approach may have been  slightly more efficient if there were no sorters but I'm definitely not convinced that matters (and also it was incorrect).

PR #2804, opened 2021-10-06
Daterangepicker

samuelyeewl:
I added a DateRangePicker widget, for when you only want a datetime.date range rather than a datetime.datetime range.

### DIFF
--- a/examples/reference/widgets/DateRangePicker.ipynb
+++ b/examples/reference/widgets/DateRangePicker.ipynb
@@ -1,0 +1,121 @@
+{
+ "cells": [
+  {
+   "cell_type": "code",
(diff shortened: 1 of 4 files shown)


philippjfr:
@samuelyeewl Do you still want to add this? I think it's reasonable to add although we might consider making `DatetimerRangePicker` more configurable.

MarcSkovMadsen:
I need this to to support `param.DateRange` parameters.

MarcSkovMadsen:
Closes #5844 5844

philippjfr:
Superseded by https://github.com/holoviz/panel/pull/6027

Issue #2817, opened 2021-10-14
Wrong dimensions order in VtkVolume panel


xavArtley:
I had to use VtkVolume recently and I noticed dimensions are not handle correcty:
If we take the doc example:
![image](https://user-images.githubusercontent.com/18531147/137278582-6e5a0b9c-0eee-4819-95eb-6e2b73e1ab57.png)

the spacing is 3 on x axis, 2 on y axis and 1 on z axis however the default coordinate system is the following:
![image](https://user-images.githubusercontent.com/18531147/137278851-53438d32-3f47-47c3-994f-8b5d024d1038.png)
and it appears the z and x axis are swaped

The correct result should be:
![image](https://user-images.githubusercontent.com/18531147/137279298-735c6c8b-36c2-49d0-bef1-7cb544b2f815.png)

PR #2818, opened 2021-10-14
Vtkvolume correction of dimensions order

xavArtley:
fix https://github.com/holoviz/panel/issues/2817

tests if data serialization is the same when numpy array is fortran or c order or when an vtkImageData is used


philippjfr:
Thanks @xavArtley, will include in 0.12.5

PR #2826, opened 2021-10-15
Vtk/vtk improvements

xavArtley:
I need to use VTK panel at work so in the same time I had some small enhancements

Now nans are handle for VTKVolumes and nan_opacity in slices can be set
nan_opacity

I updated model with the new bokeh syntax in static init


philippjfr:
Ready to merge?

PR #2842, opened 2021-10-21
Proposal to fix #2834

### LINKED ISSUE #2834: Panel objects should mirror bokehjs convention
panel 0.12.4
bokeh 2.4.1

I am expecting that the `name` parameter of a Panel object corresponds to the name attribute of the equivalent bokeh model, but it is renamed in

https://github.com/holoviz/panel/blob/9966691fe18f0a492566f540e2661792e3e75fa1/panel/widgets/base.py#L33

It becomes impossible within `jscallback`s to use `document.getElementsbyName`.

same thing with `tags`, which isn't set at all.

IMHO, this is a bug more than a lack of feature.

```
import panel as pn
pn.extension()

ti = pn.widgets.TextInput(name='foo')

ti.jscallback(value='console.log(cb_obj.name)')

ti

```

open the console log and you'll see that `null` is returned.

```javascript
null
```

#### proposed solution

I would let the parameters `name` be its former self, add `tags` and `title` ; then let `title`  & `name` parameters sync with each other unless they are explicitly set upon instantiating the widget

However my knowledge of the library is not deep enough to understand what would be the effects downstream.

hyamanieu:
Added a tags and "model_name" as parameters to all widgets.


hyamanieu:
@philippjfr could you help me understanding `test_embed.py` ? 

The returned callbacks doing `list(model.select({'type': CustomJS}))` have an empty `args` dictionnary, and the code looks like this:

```javascript
var state = null
for (var root of cb_obj.document.roots()) {
  if (root.id == '1007') {
    state = root;
    break;
  }
}
if (!state) { return; }
state.set_state(cb_obj, String(cb_obj.active.indexOf(0) >= 0))
```

rather than how it is tested in the dictionary. Where do these callbacks are supposed to come from?

I modified select as follows:
- added `title` attribute
- not renaming `name` to `title` in the `_rename` dictionary.

hyamanieu:
@philippjfr sorry to ping you again - I faced again a case where I needed the "tags" or "names" attribute and had to go around it with a custom solution.
I'll have some time after NYE to work on this issue. If you can point me how I can improve my contributions that would help.
Thanks :) .

philippjfr:
Closing, I do want to consider this for v1.0 but this PR is too stale to revive at this point. Really appreciate the work and hope we can revive it in a new branch/PR.

hyamanieu:
dear @philippjfr ,
thanks for your consideration. Feel free to ping me on #2834 when you believe it's a good time to restart and I'll work on a new PR.

PR #2870, opened 2021-11-01
Added FastApi in the user guide for embedding apps

t-houssian:
I created a tutorial to embed panel apps in fastAPI, I created the tutorial in jupyter lab and formatted similarly to the django tutorial. I also added the examples in `panel/examples/apps/fastApi`. Let me know if there is anything else I need to update, add, or change and I'll get on it.


philippjfr:
This is sooo exciting. Thanks @t-houssian! I'll probably merge now and then follow up subsequently and quiz you on that PR.

MarcSkovMadsen:
When will we see it in the docs :-)  Can't wait. Thanks @t-houssian.

One idea could be to add an example of an REST API endpoint and showing the Open API docs. 

For me that is really powerful.

And if in general there is some key of advantage (performance) of running on Fast API vs. Bokeh/ Panel server or the opposite it would be nice to know.

khannaum:
fastAPI  makes remote browser loading fast

https://panel.holoviz.org/user_guide/FastApi_Apps.html?highlight=fastapi

**no HTML link provided  in the main page**
https://panel.holoviz.org/user_guide/

image

philippjfr:
Thanks for catching that!

Issue #1460, opened 2020-07-02
AutocompleteInput does not work with param


MarcSkovMadsen:
Panel: 0.9.5
bokeh: 2.0.2

I was trying to help a user on Discourse get the `AutocompleteInput` working with `panel.Param` and `param.ObjectSelector` on discourse https://discourse.holoviz.org/t/how-to-map-param-string-to-pn-widgets-autocompleteinput/914/2 but could not.

I believe there is a bug

#### Code

```python
import panel as pn
import param

OBJECTS = ["China", "Thailand", "Japan"]


class App(param.Parameterized):
    string_value = param.ObjectSelector(default="China", objects=OBJECTS)

    view = param.Parameter()

    def __init__(self, **params):
        params["view"] = pn.Param(
            self,
            parameters=["string_value"],
            widgets={"string_value": pn.widgets.AutocompleteInput,},
        )
        super().__init__(**params)


App().view.servable()
```

#### Stack Trace

```bash
$ python -m panel serve 'scripts\issues\issue_autocomplete.py' --dev --show
2020-07-02 05:25:54,690 Starting Bokeh server version 2.0.2 (running on Tornado 6.0.4)
2020-07-02 05:25:54,691 User authentication hooks NOT provided (default user enabled)
2020-07-02 05:25:54,694 Bokeh app running at: http://localhost:5006/issue_autocomplete
2020-07-02 05:25:54,694 Starting Bokeh server with process id: 22920
2020-07-02 05:25:58,453 Error running application handler <bokeh.application.handlers.script.ScriptHandler object at 0x000001E1F7842648>: List 'options' must be a list.
File "__init__.py", line 1376, in _validate:
raise ValueError("List '%s' must be a list."%(self.name)) Traceback (most recent
call last):
  File "C:\repos\private\awesome-panel\.venv\lib\site-packages\bokeh\application\handlers\code_runner.py", line 197, in run
    exec(self._code, module.__dict__)
  File "C:\repos\private\awesome-panel\scripts\issues\issue_autocomplete.py", line 21, in <module>
    App().view.servable()
  File "C:\repos\private\awesome-panel\scripts\issues\issue_autocomplete.py", line 16, in __init__
    widgets={"string_value": pn.widgets.AutocompleteInput,},
  File "C:\repos\private\awesome-panel\.venv\lib\site-packages\panel\param.py", line 177, in __init__
    self._update_widgets()
[...]

PR #2874, opened 2021-11-02
Allow to set AutoCompleteInput in a Param pane

maximlt:
Fixes https://github.com/holoviz/panel/issues/1460

I did not integrate `AutoCompleteInput` into the *Select* model as suggested in the issue since its model is actually halfway through a text input (the value doesn't have to be one of the options with `restrict=False`) and a select widget.


philippjfr:
Thanks, test failures all spurious, merging.

PR #2876, opened 2021-11-02
Add a groups (aka optgroup) parameter to Select

maximlt:
The Select widget of Bokeh supports option groups (optgroup) that allow to visually group the elements of a dropdown menu.  The options of a Bokeh Select widget are grouped when they're declared in a dictionary, the keys of the dictionary being the groups labels. Panel already allows users to declare the options of its Select widget with a dictionary (the keys becoming the labels, and the values the values). Because of that this functionality is exposed in Panel by adding a `groups` parameter to the Select widget that is `None` by default (i.e. no groups) and accepts a dictionary that defines how the options should be grouped.

*When options is a list*
image

*When options is a dict*
image

*Error raised when the values of groups don't match exactly with the options*
image

* [x] Add tests
* [x] Add docs

### DIFF
--- a/panel/widgets/select.py
+++ b/panel/widgets/select.py
@@ -161,7 +161,7 @@ class Select(SingleSelectBase):
             if (all(isinstance(values, dict) for values in groups.values()) is False
                and  all(isinstance(values, list) for values in groups.values()) is False):
                 raise ValueError(
-                    'The values of the groups dictionnary must be all of '
+                    'The values of the groups dictionary must be all of '
                     'the dictionary or the list type.'
                 )
             labels, values = self.labels, self.values
--- a/panel/widgets/select.py
+++ b/panel/widgets/select.py
@@ -132,10 +132,9 @@ class Select(SingleSelectBase):
         scrollable area.""")
     
     groups = param.Dict(default=None, doc="""
-        Dictionary whose keys are used to visually group the options. All
-        put together the values of this dictionary must be equal to the
-        list of options if options is a list, or the keys of options if
-        options is a dictionary.""")
+        Dictionary whose keys are used to visually group the options
+        and whose values are either a list or a dictionary of options
+        to select from. Mutually exclusive with ``options``.""")
 
     _source_transforms = {'size': None, 'groups': None}
 
@@ -147,6 +146,12 @@ class Select(SingleSelectBase):
         super().__init__(**params)
         if self.size == 1:
             self.param.size.constant = True
+        self._validate_options_groups()
+
+    @param.depends('options', 'groups', watch=True)
+    def _validate_options_groups(self):
+        if self.options and self.groups:
+            raise ValueError('options and groups are mutually exclusive.')
 
     def _process_param_change(self, msg):
         msg = super()._process_param_change(msg)
@@ -154,40 +159,32 @@ class Select(SingleSelectBase):
             msg.pop('size')
         groups = msg.pop('groups', None)
         if groups is not None:
+            if (all(isinstance(values, dict) for values in groups.values()) is False
+               and  all(isinstance(values, list) for values in groups.values()) is False):
+                raise ValueError(
+                    'The values of the groups dictionnary must be all of '
+                    'the dictionnary or the list type.'
+                )
             labels, values = self.labels, self.values
             unique = len(set(self.unicode_values)) == len(labels)
-            all_group_values = list(itertools.chain(*groups.values()))
-            all_group_values_unicode = list(map(as_unicode, all_group_values))
-            if isinstance(self.options, dict):
-                if not set(self.labels) == set(all_group_values_unicode):
-                    raise ValueError(
-                        f'The values of the `groups` dictionary ({all_group_values_unicode}) '
-                        f'do not match with the options labels ({self.labels})'
-                    )
-                if unique:
-                    options = {
-                        group: [
-                            (as_unicode(self.options[label]), label)
-                            for label in labels
-                        ]
-                        for group, labels in groups.items()
-                    }
+            if groups:
+                if isinstance(next(iter(self.groups.values())), dict):
+                    if unique:
+                        options = {
+                            group: [(as_unicode(value), label) for label, value in subd.items()]
+                            for group, subd in groups.items()
+                        }
+                    else:
+                        options = {
+                            group: [as_unicode(v) for v in self.groups[group]]
+                            for group in groups.keys()
+                        }
+                    msg['options'] = options
                 else:
-                    options = {
-                        group: [as_unicode(v) for v in groups[group]]
-                        for group in groups.keys()
+                    msg['options'] = {
+                        group: [(as_unicode(value), as_unicode(value)) for value in values]
+                        for group, values in groups.items()
                     }
-                msg['options'] = options
-            else:
-                if not set(self.values) == set(all_group_values):
-                    raise ValueError(
-                        f'The values of the `groups` dictionary ({all_group_values}) '
-                        f'do not match with the options ({self.values})'
-                    )
-                msg['options'] = {
-                    group: [(as_unicode(value), as_unicode(value)) for value in values]
-                    for group, values in groups.items()
-                }
             val = self.value
             if values:
                 if not isIn(val, values):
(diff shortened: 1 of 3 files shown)


philippjfr:
Thanks @maximlt! Not saying we should do this but wouldn't it be simpler if `options` simply accepted the dictionary format you now implemented for `groups`?

maximlt:
Options already accept a dictionary that's for instance `dict(a=[1,2,3], b=[4,5,6])` and will happily display the str representation of the lists as the items that can be selected. Some people might rely on that :/ 

Another implementation I thought about was that `groups` would be a dict of indexes, instead of labels, i.e. `options=[1,2,3]` and `groups=dict(first=[0, 1], second=[0])`. But that's pretty close to what I've done.

philippjfr:
Ah okay, yeah I missed that possibility. Kind of unfortunate. Alternatively I could imagine setting **either** `options` **or** `groups`, setting both just seems a little awkward and unnecessary.

maximlt:
Yes that sounds like a good idea, I'll play around with that.

panel/widgets/select.py
```diff
@@ -147,47 +146,45 @@ def __init__(self, **params):
...
         msg = super()._process_param_change(msg)
         if msg.get('size') == 1:
             msg.pop('size')
         groups = msg.pop('groups', None)
         if groups is not None:
+            if (all(isinstance(values, dict) for values in groups.values()) is False
+               and  all(isinstance(values, list) for values in groups.values()) is False):
+                raise ValueError(
+                    'The values of the groups dictionnary must be all of '
+                    'the dictionnary or the list type.'
```
philippjfr, line 166:
```suggestion
                    'the dictionary or the list type.'
```

panel/widgets/select.py
```diff
@@ -147,47 +146,45 @@ def __init__(self, **params):
         super().__init__(**params)
         if self.size == 1:
             self.param.size.constant = True
+        self._validate_options_groups()
+
+    @param.depends('options', 'groups', watch=True)
```
philippjfr, line 151: If we require param 1.12 we could start using `on_init`:

```suggestion

    @param.depends('options', 'groups', watch=True, on_init=True)
```

hyamanieu:
If I may add my two cents, albeit a bit late: I would opt for either
- reproducing the API of the original bokeh model, I believe it's easier to maintain and there'll be less fluffy private functions
-  or making an entirely new model to fit the new API. In that case, the API of the "group" parameter of Tabulator could have been copied for consistency, meaning both "options" & "groups" could be used at the same time, just the former one is a translation table (like Tabulator's `titles`) and the latter a grouping (like Tabulator's `groups`). So if someone needs both at once, it'll work...

philippjfr:
>reproducing the API of the original bokeh model, i.e. tuples as translation tables and dictionnary for grouping. I believe it's easier to maintain and there'll be less fluffy private functions. Downside is that we need to warn the API changed, it won't be back-compatible.

Definitely won't be able to make such a fundamental backwards incompatible change for such a crucial widget at this point. In general the idea of mirroring bokeh is a worthy goal but in practice there's a fair amount of inconsistency in the bokeh APIs. We just have to make sure we add sufficient tests for the "fluffy private functions" that exist to paper over the differences.

>meaning both "options" & "groups" could be used at the same time, just the former one is a translation table (like Tabulator's titles) and the latter a grouping (like Tabulator's groups). So if someone needs both at once, it'll work... Downside is it's more work.

Right that's what @maximlt implemented initially but it seemed pretty redundant. In the case of `Tabulator` the list of `columns` is known, here you'd have to specify both which just seems awkward. If we want both grouping and relabeling I think that can be achieved by a dict-of-dicts format: `groups={'group_a': {'label': 'value', ...}, ...}`.

I have no strong opinion about whether to introduce a new `GroupedSelect` widget for this or overload the existing widget.

hyamanieu:
> Definitely won't be able to make such a fundamental backwards incompatible change for such a crucial widget at this point. In general the idea of mirroring bokeh is a worthy goal but in practice there's a fair amount of inconsistency in the bokeh APIs. We just have to make sure we add sufficient tests for the "fluffy private functions" that exist to paper over the differences.

Understood.

> 
> I think that can be achieved by a dict-of-dicts format: `groups={'group_a': {'label': 'value', ...}, ...}`.
> 

That sounds like a good idea.  It could even be directly done within `options`, without needing a `groups` param. Upon initialization it can detect whether it's a dict of list or a dict of dict.


> I have no strong opinion about whether to introduce a new `GroupedSelect` widget for this or overload the existing widget.

This is what I did for our current application, but having a dict of dict seems simpler.

Here's the class definition if it's selected. It's quite simple and the property methods are not even necessary, I made them for consistency with the original `Select`.

```python
class BokehSelect(Widget):
    options = param.ClassSelector(default=[''], class_=(dict, list))
    value = param.String(default=None)
    
    _widget_type = bk.models.Select
    
    
    def __init__(self, **params):
        super().__init__(**params)
        values = self.values
        if self.value is None and None not in values and values:
            self.value = values[0]
    
    @property
    def labels(self):
        if isinstance(self.options, dict):            
            return [v[1] if isinstance(v,tuple) else v for v in sum(self.options.values(),start=[])]
        else:
            return [v[1] if isinstance(v,tuple) else v for v in self.options]

    @property
    def values(self):
        if isinstance(self.options, dict):            
            return [v[0] if isinstance(v,tuple) else v for v in sum(self.options.values(),start=[])]
        else:
            return [v[0] if isinstance(v,tuple) else v for v in self.options]

    @property
    def groups(self):
        if isinstance(self.options, dict):
            return list(self.options.keys())
        else:
            return None

    @property
    def _items(self):
        return dict(zip(self.labels, self.values))
```

philippjfr:
>That sounds like a good idea. It could even be directly done within options, without needing a groups param. Upon initialization it can detect whether it's a dict of list or a dict of dict.

That was also my initial inclination but as @maximlt pointed out above our `Select` widget allows **any** value as an option and that includes `dict` and `list` types. So there is no way to distinguish a list/dict as a value and a list/dict as a group.

hyamanieu:
OK now I understand, thanks for taking the time to explain.

panel/widgets/select.py
```diff
@@ -163,7 +162,7 @@ def _process_param_change(self, msg):
                and  all(isinstance(values, list) for values in groups.values()) is False):
                 raise ValueError(
                     'The values of the groups dictionnary must be all of '
```
philippjfr, line 164:
```suggestion
                    'The values of the groups dictionary must be all of '
```

panel/widgets/select.py
```diff
@@ -161,7 +161,7 @@ def _process_param_change(self, msg):
             if (all(isinstance(values, dict) for values in groups.values()) is False
                and  all(isinstance(values, list) for values in groups.values()) is False):
                 raise ValueError(
-                    'The values of the groups dictionnary must be all of '
+                    'The values of the groups dictionary must be all of '
                     'the dictionary or the list type.'
```
philippjfr: Sorry should have actually read this. What does "all of the dictionary or list type" mean?

maximlt:
@philippjfr here's some more info about the failing tests.

* `test_embed_param_jslink`

The current test expects the model to have two `CustomJS` callbacks while now it has an additional one. Its `id` is `1007`, it has no `args` (empty dict) and its `code` is:
```js
var state = null
for (var root of cb_obj.document.roots()) {
  if (root.id == '1007') {
    state = root;
    break;
  }
}
if (!state) { return; }
```

* `test_embed_select_str_jslink`

The current test expects the document to have one `roots` (a `Row`) while now it has an another one. It's a `State` with the following properties:

```python
  document:<bokeh.document.document.Document object at 0x7fc8e9149950>
  id:'1005'
  js_event_callbacks:{}
  js_property_callbacks:{}
  json:False
  model_class_reverse_map:{'DataModel': <class 'bokeh.model....ataModel'>, 'Callback': <class 'bokeh.models...Callback'>, 'OpenURL': <class 'bokeh.models....OpenURL'>, 'CustomJS': <class 'bokeh.models...CustomJS'>, 'Filter': <class 'bokeh.models...s.Filter'>, 'IndexFilter': <class 'bokeh.models...exFilter'>, 'BooleanFilter': <class 'bokeh.models...anFilter'>, 'GroupFilter': <class 'bokeh.models...upFilter'>, 'CustomJSFilter': <class 'bokeh.models...JSFilter'>, 'Selection': <class 'bokeh.models...election'>, 'SelectionPolicy': <class 'bokeh.models...onPolicy'>, 'IntersectRenderers': <class 'bokeh.models...enderers'>, 'UnionRenderers': <class 'bokeh.models...enderers'>, 'DataSource': <class 'bokeh.models...taSource'>, ...}
  name:None
  ref:{'id': '1005'}
  state:{'C': defaultdict(<functio...s": []}'}), 'B': defaultdict(<functio...s": []}'}), 'A': defaultdict(<functio...s": []}'})}
  struct:{'id': '1005', 'type': 'panel.models.state.State', 'attributes': {}}
  subscribed_events:[]
  syncable:True
  tags:[]
  values:['A']
  widgets:{'1003': 0}
```

PR #2880, opened 2021-11-04
Fixed UTF-8 decoding in Terminal widget

jlstevens:
Thanks @philippjfr for the recent fixes to the Terminal widget! And here is another one...

When the subprocess output is in ASCII range, the unicode characters are one byte which means you can read any length of bytes from the file descriptor and decode. When you are out the range, the characters can be two bytes long. This resulted in unicode decode errors for any subprocess using more of UTF8 than just ASCII.

This PR fixes this by trying to decode and reading an extra byte on failure and trying to decode that (and one more time). I no longer get any decode errors without needing to use 'ignore' to drop bytes that can't be decoded.

### DIFF
--- a/panel/widgets/terminal.py
+++ b/panel/widgets/terminal.py
@@ -163,6 +163,20 @@ class TerminalSubprocess(param.Parameterized):
     def _remove_last_line_from_string(value):
         return value[: value.rfind("CompletedProcess")]
 
+    def _decode_utf8_on_boundary(self, fd, max_read_bytes):
+        "UTF-8 characters can be multi-byte so need to decode on correct boundary"
+        raw = os.read(fd, max_read_bytes)
+        try:
+            return raw.decode('utf-8')
+        except UnicodeDecodeError:
+            extra_byte = os.read(fd, 1)
+            data = (raw + extra_byte)
+            try:
+                return data.decode('utf-8')
+            except UnicodeDecodeError:
+                another_extra_byte = os.read(fd, 1)
+                return (data + another_extra_byte).decode('utf-8')
+
     def _forward_subprocess_output_to_terminal(self):
         if not self._fd:
             return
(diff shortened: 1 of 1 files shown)


jlstevens:
I'm not entirely clear why I need to try adding one or *two* bytes but I still see decode errors if I only try appending a single byte.

panel/widgets/terminal.py
```diff
@@ -163,25 +163,28 @@ def _reset(self):
...
+        try:
+            return raw.decode('utf-8')
+        except UnicodeDecodeError:
+            extra_byte = os.read(fd, 1)
+            data = (raw + extra_byte)
+            try:
+                return data.decode('utf-8')
+            except UnicodeDecodeError:
+                another_extra_byte = os.read(fd, 1)
+                return (data + another_extra_byte).decode('utf-8')
```
philippjfr, line 178:
```suggestion
    def _decode_utf8_on_boundary(self, fd, max_read_bytes, max_retries=2):
        "UTF-8 characters can be multi-byte so need to decode on correct boundary"
        raw = os.read(fd, max_read_bytes)
        while max_retries > 0:
            try:
                decoded = raw.decode('utf-8')
            except UnicodeDecodeError:
                raw = (raw + os.read(fd, 1))
                max_retries -= 1
                continue
            return decoded
        raise UnicodeDecodeError('Could not utf-8 decode subprocess output.')
```

jlstevens: I've implemented this differently already (and the `UnicodeDecodeError` at the end has a different signature, it doesn't behave like a normal exception)

jlstevens:
Ready for review. 

Maybe instead of raising an `Exception` when a boundary can't be found it could be a `RuntimeError`?

philippjfr:
Thanks @jlstevens!

PR #2895, opened 2021-11-08
Allow to update the completions options from a parameter

maximlt:
I took care of allowing an `AutocompleteInput` widget to be created from a Parameter in https://github.com/holoviz/panel/pull/2874 but missed allowing to update its options/completions.

In this PR I also reset the widget's value to `''` if the current one is not in the new options/completions.

### DIFF
--- a/panel/param.py
+++ b/panel/param.py
@@ -485,7 +485,11 @@ class Param(PaneBase):
                     self._rerender()
                 return
             elif change.what == 'objects':
-                updates['options'] = p_obj.get_range()
+                options = p_obj.get_range()
+                if ('options' in widget.param and
+                    isinstance(widget.param['options'], param.List)):
+                    options = list(options)
+                updates['options'] = options
             elif change.what == 'bounds':
                 start, end = p_obj.get_soft_bounds()
                 supports_bounds = hasattr(widget, 'start')
--- a/panel/tests/test_param.py
+++ b/panel/tests/test_param.py
@@ -5,7 +5,9 @@ import param
 from bokeh.models import (
     Div, Slider, Select, RangeSlider as BkRangeSlider, MultiSelect,
     Row as BkRow, CheckboxGroup, Toggle, Button, TextInput as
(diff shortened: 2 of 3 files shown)


jbednar:
> I also reset the widget's value to '' if the current one is not in the new options/completions.

Is '' always a legal value for this widget or the corresponding Parameter?

maximlt:
The underlying Bokeh widget accepts `value` to be outside of the completions options:
```python
# Runs fine
BkAutocompleteInput(value='ccc', completions=['aaa, 'bbb'])
```
So any `value` is legal for the Bokeh widget, and this applies to the Panel widget too.

When the widget is created from a `Selector` Parameter, `""` is legal only if the user has declared it:
```python
    choice1 = param.Selector(default='"", objects=['aaa, 'bbb'], check_on_set=False)
    # or
    choice2 = param.Selector(objects=['""', 'aaa', 'bbb'])
```

I've [documented](https://pyviz-dev.github.io/panel/user_guide/Param.html#custom-widgets) `choice1`, which seems to me like what people would most likely want, i.e. an empty autocomplete input to start things off.

It seems that there's one corner case though:
```python
    choice = param.Selector(objects=['aaa', 'bbb'])

test = Test()
test_pane = pn.Param(test, widgets={'choice': {'widget_type': pn.widgets.AutocompleteInput, 'value': ''}})
test_pane
```
The above code runs without any error and displays an empty autocomplete input, `test.choice` is equal to `'aaa'` and the widget's value is `""`. This happens because the default value of the widget class AutocompleteInput (`""`) doesn't change so the update mechanism isn't triggered.

jbednar:
> I've documented choice1, which seems to me like what people would most likely want, i.e. an empty autocomplete input to start things off.

I think that covers the case of a free-text input that's guided into one of a few common choices but allows any choice. But another common case is simply having a very large but finite list of categories, where autocomplete helps the user avoid having to page through many screens full of options. It sounds like we are not supporting that case well. In that case I'd want the display to start out with 'aaa' and always be synced.

PR #2904, opened 2021-11-11
Cleaned up and refactored Tabulator

philippjfr:
- [x] Clean up and standardize method naming on frontend
- [x] Autosize Tabulator height
- [x] Ensure that styling is applied correctly in paginated views (previously rules were applied for a single page at a time which is incorrect)
- [x] Optimize paginated styling (do not want to recompute styling each time page changes)
- [ ] Fix `frozen_rows` now that height is auto-computed

### DIFF
--- a/panel/models/tabulator.ts
+++ b/panel/models/tabulator.ts
@@ -268,7 +268,7 @@ export class DataTabulatorView extends PanelHTMLBoxView {
         this.setStyles()
         this.relayout()
         this.setFrozen()
-        this._initializing = true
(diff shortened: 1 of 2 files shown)


maximlt:
I can't select rows on a table defined like `pn.widgets.Tabulator(sel_df, selectable=True)`.

philippjfr:
>I can't select rows on a table defined like pn.widgets.Tabulator(sel_df, selectable=True).

Thanks, fixed.

Issue #2894, opened 2021-11-07
Add format parameter to DateRangeSlider


MarcSkovMadsen:
Heischo asked how he could format the label of the `DateRangeSlider`. @Hoxbro answered it was not supported but a solution would be

```python
import panel as pn
import param
from datetime import datetime

pn.extension()


class FormatDateRangeSlider(pn.widgets.DateRangeSlider):
    format = param.String(r"%m%Y")

    def _process_property_change(self, msg):
        msg = super()._process_property_change(msg)
        if "value" in msg:
            v1, v2 = msg["value"]
            v1 = datetime.strptime(v1.strftime(self.format), self.format)
            v2 = datetime.strptime(v2.strftime(self.format), self.format)
            msg["value"] = (v1, v2)

        if "value_throttled" in msg:
            v1, v2 = msg["value_throttled"]
            v1 = datetime.strptime(v1.strftime(self.format), self.format)
            v2 = datetime.strptime(v2.strftime(self.format), self.format)
            msg["value_throttled"] = (v1, v2)

        return msg


date_range_slider = FormatDateRangeSlider(
    name="Date Range Slider",
    start=datetime(2021, 1, 1),
    end=datetime(2021, 12, 31),
    value=(datetime(2021, 5, 1), datetime(2021, 9, 1)),
    format="%m/%Y",
)

date_range_slider
```

![image](https://user-images.githubusercontent.com/42288570/140647669-31423d22-2ad9-44a4-b786-ceacd30943f5.png)

**Please support `format` parameter**

PR #2917, opened 2021-11-14
Added format parameter to DateSlider and DateRangeSlider

Hiescho:
Referring to feature request https://github.com/holoviz/panel/issues/2894 I have added a 'format' parameter to DateSlider and DateRangeSlider.

### DIFF
--- a/panel/widgets/slider.py
+++ b/panel/widgets/slider.py
@@ -172,6 +172,9 @@ class DateSlider(_SliderBase):
 
     end = param.Date(default=None)
 
+    format = param.String(r"%m%Y", doc="""
+        Allows defining a custom date format string. Default is "%m%Y".""")
+
     _rename = {'name': 'title'}
 
     _source_transforms = {'value': None, 'value_throttled': None, 'start': None, 'end': None}
@@ -182,15 +185,14 @@ class DateSlider(_SliderBase):
         if 'value' not in params:
             params['value'] = params.get('start', self.start)
         super().__init__(**params)
-
+        
     def _process_property_change(self, msg):
         msg = super()._process_property_change(msg)
-        if 'value' in msg:
-            msg['value'] = value_as_date(msg['value'])
-        if 'value_throttled' in msg:
-            msg['value_throttled'] = value_as_date(msg['value_throttled'])
+        if "value" in msg:
+            msg["value"] = datetime.strptime(msg['value'].strftime(self.format), self.format)
+        if "value_throttled" in msg:
+            msg["value_throttled"] = datetime.strptime(msg['value'].strftime(self.format), self.format)
         return msg
-
 
 class DiscreteSlider(CompositeWidget, _SliderBase):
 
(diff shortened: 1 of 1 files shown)


panel/widgets/slider.py
```diff
@@ -452,6 +454,9 @@ class DateRangeSlider(_RangeSliderBase):
     end = param.Date(default=None)
 
     step = param.Number(default=1)
+    
+    format = param.String(r"%m%Y",
```
philippjfr, line 458: Is this the right default? Shouldn't it include the day?

panel/widgets/slider.py
```diff
@@ -182,16 +185,15 @@ def __init__(self, **params):
...
     def _process_property_change(self, msg):
         msg = super()._process_property_change(msg)
-        if 'value' in msg:
-            msg['value'] = value_as_date(msg['value'])
-        if 'value_throttled' in msg:
-            msg['value_throttled'] = value_as_date(msg['value_throttled'])
+        if "value" in msg:
+            msg["value"] = datetime.strptime(msg['value'].strftime(self.format), self.format)
+        if "value_throttled" in msg:
+            msg["value_throttled"] = datetime.strptime(msg['value'].strftime(self.format), self.format)
```
philippjfr, line 194: This logic is not correct. The `_process_property_change` method transforms the value sent from the frontend (usually an integer) into a datetime object. In the `_process_param_change` method you should do something like this.

philippjfr:
Thanks for the contribution @Hiescho. As it is right now this PR needs a little bit of work. I'd be happy to work with you to get it fixed up, just let me know what your bandwidth is on getting this done.

panel/widgets/slider.py
Hoxbro: I put the functionality in `_process_property_change`  in my answer to @Hiescho, to always return the exact datetime(s) shown in the widget. This is to avoid something like this:
https://user-images.githubusercontent.com/19758978/141821143-20b23956-bcae-4ce6-9765-d0be2e8665d3.mp4

Is there a better way to do this? Or am I misunderstanding what is expected of the widget?

(The function should still use `value_as_date`)

panel/widgets/slider.py
Hiescho: You are right. The default should be 

r"%d %b %Y"

to get the same formatting as with the original DateRangeSlider.

panel/widgets/slider.py
Hiescho: I am plotting monthly data and want to select start and end month of the range to be displayed with a DateRangeSlider. The DateRangeSlider does have a 'step' parameter, but it seems to be in units of milliseconds. So, as different months of the year have different lengths in time, it is not possible to make the slider handles move in steps of full calendar months by simply specifying a fixed step length. The next best solution I came up with is formatting the DateRangeSlider labels to display calendar months and year only, and discard the smaller timeframe parts of the slider values elsewhere in my code where necessary.

Hiescho: So, it seems that _process_property_change is the right place to put the change in.

philippjfr:
@Hiescho If you do ever get some time to work on this let me know. For now I'll close.

PR #2942, opened 2021-11-22
Ensure curdoc is available to async callbacks

philippjfr:
Ensures that asynchronous callbacks have access to `pn.state.curdoc`.

### DIFF
--- a/panel/io/server.py
+++ b/panel/io/server.py
@@ -105,13 +105,13 @@ def async_execute(func):
         unlock = not func.func.lock
     else:
         unlock = not getattr(func, 'lock', False)
-    if unlock:
-        @wraps(func)
-        async def wrapper(*args, **kw):
+    curdoc = state.curdoc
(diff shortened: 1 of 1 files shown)


Jhsmit:
If you're scheduling multiple coroutines with `async_execute` with this wrapper (such as in [this](https://discourse.holoviz.org/t/understanding-pn-io-with-lock-with-coroutines-and-buffers/3084) example), as the first coroutine completes, you'll no longer have access to `curdoc` because the `set_curdoc` contextmanager reverts it to `None`. 

A workaround is to get a local reference to doc (`doc=pn.state.curdoc`) in the coroutine before awaiting anything. However, that  might not be 100% fail save.
Is there a reason why the set_curdoc has to revert `pn.state.curdoc` to `None`?

philippjfr:
It definitely shouldn't revert to None. It should restore the previous state. Sounds like a bug.

PR #2963, opened 2021-11-30
Support bytes on Image

philippjfr:
Since we can validate the struct of an image there's no reason not to support this.


MarcSkovMadsen:
But what is an image? Is it the bytes object returned by the `FileInput`, is it a data_url? Is it a Pillow `Image`? Is it something similar from opencv? Is it a numpy array. I've been working a bit with an `ImageInputArea` and use cases from ML. And I see different assumptions across use cases and ML/ DL frameworks.

PR #2964, opened 2021-11-30
Support async functions on ParamMethod/ParamFunction

philippjfr:
Allows an approach like this:

```python
import aiohttp

widget = pn.widgets.IntSlider(start=0, end=10)

async def get_img(index):
    async with aiohttp.ClientSession() as session:
        async with session.get(f"https://picsum.photos/800/300?image={index}") as resp:
            return pn.pane.JPG(await resp.read())
            
pn.Column(widget, pn.bind(get_img, widget))
```

### DIFF
--- a/panel/param.py
+++ b/panel/param.py
@@ -2,15 +2,16 @@
 Defines the Param pane which converts Parameterized classes into a
 set of widgets.
 """
-import os
-import sys
-import json
-import types
 import inspect
 import itertools
+import json
+import os
+import sys
+import types
 
 from collections import OrderedDict, defaultdict, namedtuple
 from contextlib import contextmanager
+from functools import partial
 from six import string_types
 
 import param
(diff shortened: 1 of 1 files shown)


philippjfr:
Linux 3.6 test failure is spurious.

MarcSkovMadsen:
How would it work with `pn.panel(pn.bind(get_img, widget), loading_indicator=True`? Would the loading indicator be shown while the image is being loading async? That would be awesome.

philippjfr:
Correct! Considering whether to just enable that by default.

MarcSkovMadsen:
The problem is that if a function is fast. Then trying to show the loading indicator actually slows down your application. Or at least it feels like that to the user. 

Somehow the loading indicator should first be started after maybe 0.5 seconds or so ??

philippjfr:
Good point, I think because of that it's probably best left up to the user. Don't want to be in the business of timing user callbacks.

Issue #2983, opened 2021-12-07
Change Accent and Header Background Color for Fast Templates


MarcSkovMadsen:
The Fast templates are pretty nice - especially if you change their colors. This I always do when demoing Panel and creating videos. I.e. I always have to add code to change the default color which might overwhelm the user. 

There are two challenges

- The default colors are not as nice as they can be
- I have to change two colors (accent+header background)

## Solution

1) Change to 

```python
ACCENT_COLOR = "#0072B5"
```

2) add `accent` argument to `__init__`. If provide it should set the `accent_base_color` and `header_background` parameters.

## Example

![image](https://user-images.githubusercontent.com/42288570/144964569-c9ee3e3e-cb60-4185-83e5-7a7e2f8ecbce.png)

![image](https://user-images.githubusercontent.com/42288570/144964585-c95829a4-8abf-44a0-837f-9979b37959f7.png)

```python
import panel as pn
import numpy as np
import holoviews as hv

pn.extension(sizing_mode="stretch_width")

ACCENT_COLOR = "#0072B5"

def box():
    box = hv.BoxWhisker((np.random.randint(0, 10, 100), np.random.randn(100)), 'Group').sort()
    box.opts(box_color=ACCENT_COLOR)
    return box

pn.template.FastListTemplate(
    site="Awesome Panel", title="New Accent Color",
    main=[box,box],
    accent_base_color=ACCENT_COLOR, header_background=ACCENT_COLOR
).servable()
```

PR #2984, opened 2021-12-07
Change Accent and Header Background Color for Fast Templates

MarcSkovMadsen:
Addresses #2983

Looks like

FastGridTemplate
FastGridTemplateDark
FastListTemplate
FastListTemplateDark


panel/template/fast/base.py
```diff
@@ -12,7 +12,7 @@
 
 class FastBaseTemplate(BasicTemplate):
 
-    accent_base_color = param.String(doc="""
+    accent_base_color = param.String(default="#0072B5", doc="""
```
MarcSkovMadsen, line 15: I wanted a way for the user to get the color for their app to make it easier to use the color in plots etc.

philippjfr:
I like the color but really we need to be way more consistent about theming and styling. If we make this change it needs to apply to all templates and we should define a fixed color palette for both light (default) and dark themes which all the template specific themes can reference somehow.

MarcSkovMadsen:
I don't think vanilla, bootstrap, material and fast templates can be alike. They are different design systems with very different properties.

And the Fast Templates are so far ahead of the rest of the templates. Who will ever take the time to develop the required css files for the other templates?

philippjfr:
>I don't think vanilla, bootstrap, material and fast templates can be alike. They are different design systems with very different properties.

I'm not saying the need to be alike but I think there should a central definition of the default color palette, which is then reused by most of the themes. 

```
default_palette = {
    'primary': ...,
    'accent': ...,
    'secondary': ...,
    ...
}
```

and then we can have a templated CSS file like this:

```
:root {
  --pn-primary-color: {{ primary }};
  --pn-accent-color: {{ accent }};
  --pn-secondary-color: {{ secondary }};
}
```

then finally each template can use those variables to control its color scheme, e.g. the fast template CSS can then just use: 

```
:root {
    --container-background: var(--pn-primary-color);
}
```

MarcSkovMadsen:
I'm not quite sure what the status on this is. Would it be possible to get into 0.13? Is there anything I should do to get this in?

(It would make my "developer advocate" work much easier to have this in. Then I would not have to always pick some more beautiful colors to make examples look nice. And the new users would have less code to worry about).

MarcSkovMadsen:
> > I don't think vanilla, bootstrap, material and fast templates can be alike. They are different design systems with very different properties.
> 
> I'm not saying the need to be alike but I think there should a central definition of the default color palette, which is then reused by most of the templates.
> 
> ```
> default_palette = {
>     'primary': ...,
>     'accent': ...,
>     'secondary': ...,
>     ...
> }
> ```
> 
> and then we can have a templated CSS file like this:
> 
> ```
> :root {
>   --pn-primary-color: {{ primary }};
>   --pn-accent-color: {{ accent }};
>   --pn-secondary-color: {{ secondary }};
> }
> ```
> 
> then finally each template can use those variables to control its color scheme, e.g. the fast template CSS can then just use:
> 
> ```
> :root {
>     --container-background: var(--pn-primary-color);
> }
> ```

This problem here is that the `accent color` in Fast corresponds to the `accent` color in Bootstrap. We need some decision on whether to only use `primary` color terminology only. Or we use different terminologies for different frameworks, so that it is easier to relate to their documentation. In my head I've been implementing, describing everything as the latter.

MarcSkovMadsen:
If you don't like BLUE, we might be able to find a better green than the suggested one :-)

philippjfr:
Will merge and probably follow up separately.

Issue #2985, opened 2021-12-07
Enable Param to return an alphabetically sorted list of parameters


MarcSkovMadsen:
## My Pain

I sometimes use Param for educational purposes. For example to enable a user to explore the parameters of a class. If the class has lots of parameters they often end of being a mess because they are sorted by some order of precedence that does not make a lot of sense to a new user.

Finding something alphabetically is in many cases the natural, easiest and expected way.

## Solution: `sorted` parameter

Add `sorted` bool parameter  to Param.

## Example

Try to find something here 😄 

![image](https://user-images.githubusercontent.com/42288570/144975816-de7031b6-e7fa-48be-a38c-544901be3f94.png)

```python
import panel as pn
import numpy as np
import holoviews as hv

pn.extension(sizing_mode="stretch_width")

template = pn.template.FastListTemplate(
    site="Awesome Panel", title="New Accent Color",
    main=[pn.widgets.Button(name="Click Me", button_type="primary")],
    accent="#A01346"
)
pn.Param(template).servable()
```

## Additional Context

It would also be nice to be able to be able to sort `.controls()` alphabetically on each tab.

MarcSkovMadsen:
This has been merged. So I will close.

PR #2986, opened 2021-12-07
Add alphabetical and custom sort to Param

MarcSkovMadsen:
Addresses #2985

image


panel/param.py
```diff
@@ -138,6 +139,11 @@ class Param(PaneBase):
     show_name = param.Boolean(default=True, doc="""
         Whether to show the parameterized object's name""")
 
+    sorted = param.Parameter(default=False, doc="""
```
MarcSkovMadsen, line 142: Could also be called `sort`.

I would like to use `param.ClassSelector` instead of `param.Parameter` to get validation. But I did not know how to set `class_=(bool, ?callable)`.

MarcSkovMadsen:
## Example from issue

Much easier to find something

image

```python
import panel as pn
import numpy as np
import holoviews as hv

template = pn.template.FastListTemplate(
    site="Awesome Panel", title="New Accent Color",
    main=[pn.widgets.Button(name="Click Me", button_type="primary")],
)
pn.Param(template, sorted=True).servable()
```

panel/param.py
philippjfr: I vote for `sort` and the parameter can be declared like this:

```python
from collections.abc import Callable
    
    sort = param.ClassSelector(class_=(bool, Callable))
```

panel/param.py
```diff
@@ -138,6 +139,11 @@ class Param(PaneBase):
...
+    If True the widgets will be sorted alphabetically by label. If a callable is provided
+    it will be used to sort the Parameters, for example lambda x: x[1].label[::-1] will sort by the
+    reversed label.""")
```
philippjfr, line 145: Four space indent please :)

panel/param.py
```diff
@@ -573,6 +579,15 @@ def action(event):
     def _ordered_params(self):
         params = [(p, pobj) for p, pobj in self.object.param.objects('existing').items()
                   if p in self.parameters or p == 'name']
+        if self.sorted is True or callable(self.sorted):
```
philippjfr, line 582:
```python
if self.sort:
```

Should be sufficient.

panel/param.py
```diff
@@ -6,6 +6,7 @@
 import itertools
 import json
 import os
+from re import S
```
philippjfr, line 9: Doesn't appear to be used.

philippjfr (changes requested):
Looks great, I requested a few changes but overall I'll be very happy with this.

panel/param.py
MarcSkovMadsen: It depends. If we can change to `param.ClassSelector` to get validation then yes. Otherwise if user provided `sorted="no"` I implemented it to ignore it and use the default sorting. With your suggestion it would instead sort alphabetically. :-)

philippjfr: See above :-)

philippjfr: You already did it, ignore me :-)

MarcSkovMadsen:
Thanks for your comments @philippjfr . I've fixed and pushed.

There is one test failure. But I cannot see it should be related to these changes.

image

Please let know if I should do more. Thanks.

jbednar (approved):
Very cool, thanks!

maximlt:
The failed test was due to a corrupted dev release of holoviews (my bad!). This had been fixed very quickly. @MarcSkovMadsen feel free to push an empty commit to rerun the test suite.

philippjfr:
I've just restarted the test suite.

maximlt:
Thanks @philippjfr, I always forget that the CI can be rerun (there should be a more obvious button but maybe that's intentional).

Issue #2988, opened 2021-12-07
Tabulator does not handle multi-index correctly


philippjfr:
Seems like there's some issue with the rendering of a multi-index on Tabulator:

```python
from bokeh.sampledata.autompg import autompg

pn.widgets.Tabulator(autompg.set_index(['cyl','name']), pagination='remote', page_size=15)
```

<img width="579" alt="Screen Shot 2021-12-07 at 2 12 09 PM" src="https://user-images.githubusercontent.com/1550771/145035367-cf76caed-c407-4c5d-867e-eb20eec73c7c.png">

PR #2989, opened 2021-12-07
Fix Tabulator with multi-index and pagination

philippjfr:
Fixes https://github.com/holoviz/panel/issues/2988


Hoxbro:
Were writing this in the issue, but then I saw the PR. 

This is related to `MultiIndex`  where a name can be `None` which errors Tabulator, see https://discourse.holoviz.org/t/tabulator-on-pandas-groupby-describe-stack-failing/3136/5.

philippjfr:
Related but definitely slightly different, will look at that in a separate PR.

PR #2999, opened 2021-12-08
Update ButtonGroup value when changing options 

### LINKED ISSUE #2998: Error displaying CheckBoxGroup after modifying it in a callback
Hello,

I am trying to modify the options of a CheckBoxGroup based on an other widget, while keeping the currently selected options if they still make sense. This causes the checked boxes of the CheckBoxGroup to be erroneously displayed.

versions:
panel==0.12.5
param==1.12.0

To reproduce:

```
import panel as pn
pn.extension()
    
sets = {'set_0': ['A', 'B'], 
        'set_1': ['B', 'C']}

set_select = pn.widgets.Select(options=list(sets.keys()))
elements_select = pn.widgets.CheckBoxGroup(value=[], options=sets[set_select.value])

def update_element_options(*events):
    for event in events:
        elements_select.options = sets[event.new]
        elements_select.value = sorted(set(elements_select.value) & set(elements_select.options))
set_select.param.watch(update_element_options, ['value'])

pn.Column(set_select, elements_select)
```

Run this, check 'B' in the CheckBoxGroup, then select 'set_1' in set_select. The options of the CheckBoxGroup change, but it seems that it is the 'C' option that is selected.

Hoxbro:
Fixes #2998

### DIFF
--- a/panel/tests/widgets/test_select.py
+++ b/panel/tests/widgets/test_select.py
@@ -320,6 +320,11 @@ def test_toggle_group_check(document, comm):
         select._process_events({'active': []})
         assert select.value == []
 
+        select.value = ["A", "B"]
+        select.options = ["B", "C"]
+        select.options = ["A", "B"]
+        assert widget.labels[widget.active[0]] == "B"
+
 
 def test_toggle_group_radio(document, comm):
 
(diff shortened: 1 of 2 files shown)


Hoxbro:
I have an idea of how to fix the problem but don't know when I have time to take a look at it. So you can update/close this PR if necessary @philippjfr.

philippjfr:
Fabulous, thank you!

PR #3011, opened 2021-12-13
Add home to TOC of the home page

maximlt:
Attempt to get the table of content in the left side bar on the home page. Trying that because it's the only difference I could spot with `holoviews`.

### DIFF
--- a/doc/index.rst
+++ b/doc/index.rst
@@ -96,6 +96,7 @@ The Panel project is grateful for the sponsorship by the organizations and compa
    :hidden:
    :maxdepth: 2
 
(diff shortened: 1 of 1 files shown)


doc/index.rst
```diff
@@ -96,6 +96,7 @@ The Panel project is grateful for the sponsorship by the organizations and compa
    :hidden:
    :maxdepth: 2
 
+   Home <self>
```
philippjfr, line 99: I was also hoping this was the issue but as far as I could tell this basically has no effect. We'll see in a dev build.

Issue #3020, opened 2021-12-17
A Label Indicator


MarcSkovMadsen:
A very common visualization (in machine learning) is Labelling/ Classification.

I would propose adding an indicator that looks something like

![image](https://user-images.githubusercontent.com/42288570/146484046-12b1cba8-c1ac-432c-99ce-aabdd8f67b37.png)

and with an api something like

```python
class Label(pn.reactive.ReactiveHTML):
    """The ClassificationPlot provides plots of the output of a classification, i.e. the *labels*
    and their *score*."""

    object = param.Dict(
        doc="""
    The output of a classification""",)
    color = param.Color(
        ACCENT_COLOR,
        """
    The color of the bars of the plot""",
    )
    theme = param.Selector(
        default="default",
        objects=["default", "dark"],
        constant=True,
        doc="""
    The theme of the plot. Either 'default' or 'dark'""",
    )
    top = param.Integer(
        5,
        bounds=(1, None),
        doc="""
    Displays the top number of elements
    """,
    )
```

Inspiration: https://gradio.app/docs/#o_label

PR #3021, opened 2021-12-17
Adds a Label indicator

MarcSkovMadsen:
Adds a `Label` as requested in #3020.

# Todo

- [] Create Notebook reference example
- [] Clarify if it is ok with Apex dependency
- [] Get Review.


MarcSkovMadsen:
I will close and create a new one as this unfortunately included the BeforeAfterSlider work.

PR #3022, opened 2021-12-17
Adds Scores indicator

MarcSkovMadsen:
Adds a `Label`/ `Scores` indicator as requested in #3020.

## Todo

- [x] Create Notebook reference example
- [x] Rename from `Label` to `Scores`
- [x] Get `layout` parameters working.
- [ ] Get Review.
  - [ ] Clarify if it is ok with Apex dependency

## Nice to have

- [ ] Fix error in Notebook console.

### DIFF
--- a/panel/widgets/indicators.py
+++ b/panel/widgets/indicators.py
@@ -907,7 +907,7 @@ class Label(ReactiveHTML):
 
     value = param.Dict(
         doc="""
-        The output of a classification. The key is the label and the value the score. 
+        A dictionary of labels (key) and their scores (value).
         For example  `{"egyptian": 0.22, "tabby cat": 0.18, "tiger cat": 0.13, "lynx": 0.09, "Siamese cat": 0.04}`""",
         precedence=-1, # Will raise error if not set to -1 because its not used on js side
     )
@@ -915,13 +915,13 @@ class Label(ReactiveHTML):
         5,
         bounds=(1, 10),
         doc="""
-        The number of labels to plots
+        The maximum number of labels to plots
     """,
     )
     color = param.Color(
         "#0072B5",
         """
-        The color of the bars of the plot""",
+        The color of the bars. Default is the same as the Fast default accent_base_color.""",
     )
     theme = param.Selector(
         default="default",
--- a/panel/widgets/indicators.py
+++ b/panel/widgets/indicators.py
@@ -899,10 +899,11 @@ def _get_theme() -> str:
         return "dark"
     return "default"
 
-class Label(ReactiveHTML):
-    """The Label visualizes the result of a classification, i.e. the *labels* and their *score*.
+class Scores(ReactiveHTML):
+    """The Scores indicator visualizes a dictionary of *labels* and their *score*. 
     
-    For example a result like `{"egyptian": 0.22, "tabby cat": 0.18, "tiger cat": 0.13, "lynx": 0.09, "Siamese cat": 0.04}`.
+    For example the results of a ML classification like 
+    `{"egyptian": 0.22, "tabby cat": 0.18, "tiger cat": 0.13, "lynx": 0.09, "Siamese cat": 0.04}`.
     """
 
     value = param.Dict(
@@ -931,7 +932,7 @@ class Label(ReactiveHTML):
     )
     
 
-    label = param.String("ORANGUTAN", constant=True)
+    label = param.String("", constant=True)
     top_value = param.List(constant=True)
 
     _base_options = param.Dict(CONFIG, constant=True)
--- a/panel/widgets/indicators.py
+++ b/panel/widgets/indicators.py
@@ -850,3 +850,145 @@ class Tqdm(Indicator):
         """Resets the parameters"""
         self.value = self.param.value.default
         self.text = self.param.text.default
+
+"""Module of classification tools"""
+
+import panel as pn
+import param
+
+CONFIG = {
+    "series": [
+        {
+            "name": "Score",
+            "data": [
+                {"x": "Egyptian cat", "y": 1},
+                {"x": "tabby, tabby cat", "y": 21.0},
+                {"x": "tiger cat", "y": 15.0},
+                {"x": "lynx, catamount", "y": 10.0},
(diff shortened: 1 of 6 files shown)


MarcSkovMadsen:
You can explore the `Label` indicator via `panel serve 'panel/tests/widgets/test_label.py' --auto --show`



https://user-images.githubusercontent.com/42288570/146491477-d7090c7e-3116-4ccf-9fc3-94fcaf5c7a8e.mp4

MarcSkovMadsen:
I've added a notebook example at `examples/reference/indicators/Label.ipynb`. You can serve it as an app via `panel serve examples/reference/indicators/Label.ipynb --autoreload --show`.

Unfortunately some errors show in the browser console.

image

jbednar:
This looks useful, but it seems much more specialized than I'd expect from the name Label. Maybe `ClassLabel` or `LabelScores`?

MarcSkovMadsen:
Or Classification or Labels

jbednar:
I did think about `Classification`, but from that word I wouldn't expect such an object to have scores or multiple possible classifications, just a single class label. Really this is a probability of classification per possible class, which is quite specific. `Labels` we're already using elsewhere in HoloViz to mean a collection of text tags, which I think is a more literal and accurate use than the specialized one here.

MarcSkovMadsen:
The purpose is to visualize the output of a classification. That output is a label. Thus the label takes 50% of the height. But very often you don't get only one label, but multiple labels with scores. The other 50% is given to visualize this level of detail.

Hope this helps naming.

jbednar:
Right; I know the ML context; it's just that Panel is general, not tied to any particular domain, while gradio is quite ML focused. I think it's great to have ML-focused components available in Panel, as long as there's something in the name to indicate that they aren't trying to be something basic or general purpose.

MarcSkovMadsen:
LabelScores or simpler Scores.

philippjfr:
I like `Scores`.

MarcSkovMadsen:
I've renamed from `Label` to `Scores`.

panel/widgets/indicators.py
```diff
@@ -850,3 +850,155 @@ def reset(self):
...
+    __javascript__ = ["https://cdn.jsdelivr.net/npm/apexcharts"]
+
+    def __init__(self, **params):
+        params["theme"] = params.get("theme", _get_theme())
+
+        super().__init__(**params)
+
+        self._handle_change()
+
+    @param.depends("value", "top", watch=True)
```
philippjfr, line 990:
```suggestion

    @param.depends("value", "top", watch=True, on_init=True)
```

philippjfr (commented):
This is really nice, so thanks @MarcSkovMadsen! I'm going to perform some of my own refactoring but some high-level comments:

- I realize there's still no code style guide but parameters have always followed a standardized indentation, so please, please just follow it for now.
- I like the fact that all notebooks you contribute are servable as nice templated apps, but I'm not a huge fan of the `App` section you add at the end. We now have mechanism to enable a template in the `pn.extension`, e.g. `pn.extension(template='fast-list')` call so please just use that. Will demonstrate how in the changes I will push. We probably have to convert a bunch of other notebooks as well.
- Instead of the `App` section all other indicators (and widgets) have a controls section, so this component should have one too. I will add that.
- This component requires an external library (`apexcharts`) and since the component is imported by default this custom library will always be included. That doesn't seem so great, will consider what to do about this.
- The `_get_theme` function is great and we should use it consistently if possible. I've created a `state.theme` property here and we should work at making most if not all components support it.

MarcSkovMadsen:
Thanks. 

But how do I easily set parameters on the template @philippjfr . `pn.extension(template='fast-list')` works ok if I don't have to set parameters. But if I do then it gets convoluted.

MarcSkovMadsen:
What does the `on_init=True` in `param.depends` do @philippjfr . I tried it once. I expected it to run the function, but it did not. I have never seen any documentation for it.

MarcSkovMadsen:
ahhh. `on_init` only makes sense on a Class? I tried using it on a standalone function.

MarcSkovMadsen:
Regarding the code formatting. I don't like to forget it. I simply forget it because it's not a part of my workflow in any other project. We just apply isort, black and other auto formatting.

philippjfr:
>But how do I easily set parameters on the template @philippjfr . pn.extension(template='fast-list') works ok if I don't have to set parameters. But if I do then it gets convoluted.

`pn.state.template` exposes the template. That said I do not think default examples should use a lot of custom configuration of the template. It should use the defaults in almost all cases. If those defaults aren't appropriate they should probably be changed.

jbednar:
>ahhh. on_init only makes sense on a Class? I tried using it on a standalone function.

Only a class has a constructor to do initialization.

maximlt:
`on_init` is documented [here](https://param.holoviz.org/user_guide/Dependencies_and_Watchers.html#dependencies) in the Param's docs (you need a good detective eye to find it :D)

MarcSkovMadsen:
> > ahhh. on_init only makes sense on a Class? I tried using it on a standalone function.
> 
> Only a class has a constructor to do initialization.

You. I imaged the standalone function would just be run once. Instead of me having to do run the function via an extra line `update()`. Its the same need whether the function is a method on a class or not.

jbednar:
Interesting idea; seems like it would work and makes sense!

philippjfr:
>I imaged the standalone function would just be run once. Instead of me having to run the function via an extra line update(). Its the same need whether the function is a method on a class or not.

Would be okay with me, at least it's consistent. I regret supporting functions on `depends` in the first place, but back then we didn't have `bind`.

philippjfr:
Sorry to have this sitting around and getting stale for so long. I'm still not entirely convinced that adding an entire charting library (apexcharts) for an indicator is sensible. So I'd probably want to revive this in some extension package (like panel-apexcharts). I'll close but if you really strongly feel this should be included I'll revive it.

MarcSkovMadsen:
I don't care about apex chart. But I still think Panel needs an easy to use Scores indicator.

PR #3024, opened 2021-12-19
Ensure CachedVariadicBox only caches layout for finite height and width

### LINKED ISSUE #2971: JSON pane does not adjust height when object changes
Panel: 0.12.5

I would expect the "Something else" text to be moved downwards when the JSON pane expands. But it does not.

https://user-images.githubusercontent.com/42288570/144559504-d71edd15-9faf-477a-9752-20564bdb6b44.mp4

```python
from typing import Dict
import panel as pn
from typing import Dict

def data(n):
    return {index: index for index in range(n)}

slider = pn.widgets.IntSlider(value=0, start=0, end=100)
idata = pn.bind(data, slider)
pn.Column(
    slider,
    idata, "Something else"
).servable()
```

Hoxbro:
Fixes #2971 

Only cache layout if the width and height is finite.

### DIFF
--- a/panel/models/layout.ts
+++ b/panel/models/layout.ts
@@ -73,7 +73,7 @@ export class CachedVariadicBox extends VariadicBox {
     const min_count = (!this.changed || (this.sizing_mode == 'fixed') || (this.sizing_mode == null)) ? 0 : 1;
     const cached = this._cache.get(key_str);
     const cache_count = this._cache_count.get(key_str)
-    if (cached != null && cache_count != null && (cache_count >= min_count)) {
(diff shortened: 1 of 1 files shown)


philippjfr:
Much appreciated @Hoxbro! I've renamed the PR since it should fix more than just the `JSON` pane.

philippjfr:
So I realize this change is actually correct, nonetheless I have to revert for now. I'm seeing 10-20x slowdowns in layout computations for production apps. I'm not sure we'll be able to find a solution here that doesn't cause serious issues. The good news is that the layout engine is going away over the next few months and this should no longer be an issue.

PR #3025, opened 2021-12-21
Remove usage of "six.string_types"

rahulporuri:
This PR removes the usage of `six.string_types` in the codebase and replaces them with just `str`. This is doable now that panel doesnt support Python < 3.6. This was done by simply searching for and replacing the usage of `six.string_types` in the codebase using an editor.


panel/pane/media.py
```diff
@@ -107,7 +106,7 @@ def _process_param_change(self, msg):
 
 class Audio(_MediaBase):
 
-    object = param.ClassSelector(default='', class_=(string_types + (np.ndarray,)),
+    object = param.ClassSelector(default='', class_=(str, np.ndarray,)),
```
philippjfr, line 109:
```suggestion
    object = param.ClassSelector(default='', class_=(str, np.ndarray),
```

philippjfr:
Branch needs to be rebased and I commented on one small fix. Thanks for all this cleanup!

rahulporuri:
@philippjfr thanks for the recommendation. I've addressed your comment and rebased the branch. Can you take another look?

philippjfr:
Thanks @rahulporuri, looks good! Will merge when tests pass.

PR #3026, opened 2021-12-21
Remove usage of "sys.version_info" to handle Python 2 code

rahulporuri:
This PR removes the usage of `sys.version_info` and derivatives in the codebase to handle Python < 3 code. The derivatives are - 

- `@py3_only` decorator used to mark tests that are meant to be run on Python > 2.
- `as_unicode` and `unicode_repr` utility functions
- `unicode` alias to `str` on Python 3

These changes can be done now that the package supports only Python >= 3.6

### DIFF
--- a/panel/config.py
+++ b/panel/config.py
@@ -433,7 +433,7 @@ class panel_extension(_pyviz_extension):
             else:
                 setattr(config, k, v)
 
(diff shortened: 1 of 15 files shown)


philippjfr:
Appreciate the cleanup @rahulporuri! All looks good.

PR #3027, opened 2021-12-21
Remove "utf-8" coding cookies from modules

rahulporuri:
The default encoding on Python 3 is `"utf-8"` so we can remove the coding cookies from python modules in the package.


philippjfr:
Thanks @rahulporuri!

PR #3029, opened 2021-12-21
DOC: add note on enabling panel widget on Jupyter Lab

raybellwaves:
PR motivated by https://discourse.holoviz.org/t/what-does-the-render-with-panel-jupyter-lab-button-do/3039/5


philippjfr:
Thanks @raybellwaves!

PR #3040, opened 2021-12-24
fixes #3039: tqdm style color reset

### LINKED ISSUE #3039: TQDM removes text color set by user via style parameter
See https://discourse.holoviz.org/t/tdqm-text-color-change/3193

I will put in a fix now.

MarcSkovMadsen:
Fixes #3039 discussed here https://discourse.holoviz.org/t/tdqm-text-color-change/3193


philippjfr:
Thanks @MarcSkovMadsen!

PR #3046, opened 2021-12-27
fix deckgl tooltips

### LINKED ISSUE #3015: Tooltips can not be disabled in the DeckGL pane
Tested in `panel=0.12.6` with `pydeck==0.7.1` but pretty sure I saw the issue in previous versions.

The [docs for DeckGL pane](https://panel.holoviz.org/reference/panes/DeckGL.html) indicate there are two tooltip settings:

- `pydeck.Deck.tooltip`; and 
- `pn.pane.DeckGL.tooltips`

Both of these accept `Boolean` type, so I would expect that setting either (or both) of these to `False` should disable tooltips, however this is not the case.

This can be easily reproduced in the [PyDeck notebook example from the docs](https://hub.gke2.mybinder.org/user/holoviz-panel-lyvsn525/lab/tree/examples/reference/panes/DeckGL.ipynb#PyDeck):

- pass `tooltip=False` to `pydeck.Deck`; and/or
- pass `tooltips=False` to  `pn.pane.DeckGL`

Any combination I try, I still get a tooltip showing up on the map:

Screen Shot 2021-12-14 at 13 38 24

Cheers,
Dan

MarcSkovMadsen:
Fixes #3015

I don't know why this simple change makes it work. But it does.

https://user-images.githubusercontent.com/42288570/147452675-330cf3bd-e0df-43eb-abd0-8327846fdb2c.mp4

### DIFF
--- a/panel/models/deckgl.ts
+++ b/panel/models/deckgl.ts
@@ -226,7 +226,7 @@ export class DeckGLPlot extends HTMLBox {
       initialViewState: [ Any,                          {} ],
       layers:           [ Array(Any),                   [] ],
       mapbox_api_key:   [ String,                       '' ],
-      tooltip:          [ Any,                          {} ],
+      tooltip:          [ Any,                          false ],
       viewState:        [ Any,                          {} ],
     }))
 


philippjfr:
Superseded by https://github.com/holoviz/panel/pull/3158. Thanks for the fix though!

PR #3057, opened 2022-01-01
fix: use provided theme.css

### LINKED ISSUE #3056: Custom theme is not not applied
Version 0.12.6. Problem can also be reproed using latest code available in the repo.

When using a css theme that is not located the BUNDLE_DIR, the BasicTemplate ignores the provided value, and uses the base_css instead.

This is closely following the example from https://panel.holoviz.org/user_guide/Templates.html:
``` python
import param
import pathlib
from panel.template import BootstrapTemplate
from panel.template.theme import DefaultTheme, Theme

class BootstrapAcfTheme(DefaultTheme):
     # To repro the issue, the css provided here should not be in the BUNDLE_DIR
    css = param.Filename(default=pathlib.Path(__file__).parent / 'bootstrap-acf.css')
    _template = BootstrapTemplate

#Providing my own theme to the BootstrapTemplate
bootstrap_theme = BootstrapTemplate(title='Autocorrelation function', theme=BootstrapAcfTheme)
bootstrap_theme.show()
```
Upon rendering the page, the `{{ raw_css }}` tag from the template is replaced by the base_css instead of the content of `bootstrap-acf.css`. The served web page contains the following snippet, which comes from `\panel\template\theme\default.css`
```css
...
<style type="text/css">
#header {
  background-color: #00aa41;
}
</style>
```

#### Code
This issue seems to be due to a cut and paste error. Line 620 uses `theme.base_css`, I am quite certain that it should use `theme.css` instead. Line 620 is an exact copy of line 613. 
https://github.com/holoviz/panel/blob/007475dbb7de30373fe7111348d4fa020d446ac9/panel/template/base.py#L615-L621

I'll provide a PR that corrects this.

blelem:
Fixes #3056

### DIFF
--- a/panel/template/base.py
+++ b/panel/template/base.py
@@ -617,7 +617,7 @@ class BasicTemplate(BaseTemplate):
                     if (BUNDLE_DIR / name / basename).is_file():
                         css_files['theme'] = dist_path + f'bundled/{name}/{basename}'
                     else:
-                        with open(theme.base_css, encoding='utf-8') as f:
+                        with open(theme.css, encoding='utf-8') as f:
                             raw_css.append(f.read())
 
         return {


philippjfr:
Thanks for the fix @blelem!

PR #3067, opened 2022-01-06
Add a `disabled_options` parameter to a custom Select widget

maximlt:
This PR adds a `disabled_options` parameter to the `Select` widget which allows to disable some options, which in practice makes them unusable and un-clickable.

image

I've extended the base `Select` widget of Bokeh to implement this option on the client-side. I've called the model `CustomSelect` instead of something more specific, but I'm open for a better name.

`disabled_options` is a List Parameter (default to `[]`) that accepts a subset of the `options` **values**, i.e. if `options` is a dictionary then `disabled_options` must list some dictionary values, not keys. I've had the feeling this would be more intuitive, but since dictionary options have been confusing people I'd also be fine implementing the "labels" alternative if that's clearer.

I've added some validation in a dedicated callback but I observed that it's called after `_process_param_change` which means that the bokeh model and then the UI still gets updated before an error is raised on the Python side. I wasn't sure implementing the validation/error part in `_process_param_change` was the right thing to do either.


philippjfr (changes requested):
Congrats on your first custom Bokeh model! Only have one comment which is that we use 2-space indentation for TypeScript files. If you could resolve that I'll merge.

panel/models/customselect.ts
```diff
@@ -0,0 +1,67 @@
...
+                element.setAttribute('disabled', 'true')
+            }
+        })
+        return opts
+    }
+
+    protected _update_disabled_options(): void {
+        console.log(this.input_el)
+        console.log('UPDATE DISABLED OPTIONS HERE')
+        console.log(this.model.disabled_options)
```
philippjfr, line 26: Oh and please remove these.

maximlt: Fixed in https://github.com/holoviz/panel/pull/3067/commits/f713f504357aa423184ed860211e67173987f925

maximlt:
@philippjfr I've made the changes you requested. Sorry for the indentation, I had noticed the difference but then forgot about it. VSCode wasn't also playing nicely (which is quite rare), I did set the indentation to be of `2` spaces by default for typescript files but this setting was overridden by an indentation auto-detection feature.

I noticed that this PR was actually breaking the `SingleSelect` widget by passing it `disabled_options` while this was not a property of the Bokeh model. So I've implemented the possibility to disable options of this widget too (`Select`, when `size` is set to a value greater than 1).

And made three other changes (plus a few minor ones):
* Raise an error when `groups` is set and `size` is greater than 1, since the `SingleSelect` widget doesn't support options grouping
* Raise an error when on initialization the value is one the disabled options (shouldn't have missed this one!)
* Refactored the tests

Issue #3060, opened 2022-01-04
Default value for TextInput in `widgets.Tabulator.download_menu` does not happen to be default value for filename


xblahoud:
#### ALL software version info
(this library, plus any other relevant software, e.g. bokeh, python, notebook, OS, browser, etc)

panel==0.12.5

bokeh==2.4.2
holoviews==1.14.6
hvplot==0.6.0
param==1.12.0

#### Description of expected behavior and the observed behavior

This code should produce a button that downloads the Tabular into file `changed.csv` by default. However, it downloads `table.csv` instead (unless the value was changed there and back).
```python
filename, button = t.download_menu(
            text_kwargs={'name': 'Enter filename', 'value': 'changed.csv'},
            button_kwargs={'name': 'Download table'}
        )
```

#### Complete, minimal, self-contained example code that reproduces the issue

```
import numpy as np
import pandas as pd
import panel as pn

pn.extension()

download_df = pd.DataFrame(np.random.randn(10, 5), columns=list('ABCDE'))

download_table = pn.widgets.Tabulator(download_df)

filename, button = download_table.download_menu(
    text_kwargs={'name': 'Enter filename', 'value': 'default.csv'},
    button_kwargs={'name': 'Download'}
)

pn.Row(
    pn.Column(filename, button),
    download_table
)
```

PR #3068, opened 2022-01-07
Respect the default filename set by `download_menu`

maximlt:
Fixes https://github.com/holoviz/panel/issues/3060

The JS callback set on the TextInput widget wasn't being initially triggered so the Tabulator `filename` property still had its original default value (`'table.csv`). I found no way to either trigger the JS callback or set the `filename` Tabulator property (it's not exposed to Panel users, it's just a property of the model). Instead I've removed that JS callback entirely and have set `filename` in the button JS callback.


philippjfr:
Can you pin pyvista back to whatever was the previous version for now so the tests pass?

Issue #3058, opened 2022-01-02
`Color` parameter on ReactiveHTML


Jhsmit:
I'm getting an error with a `param.Color` parameter on a `ReactiveHTML`

code:

```python
import param
from panel.reactive import ReactiveHTML
import panel as pn

class TestReactive(ReactiveHTML):

    bg_color = param.Color(
        default=None,
        doc="Color of the background"
    )

    _template = """
    <div id="container" style="width:100%; height:100%;"><p>test</p></div>
    """

if __name__ == '__main__' or __name__.startswith('bokeh_app'):
    test = TestReactive(
        height=500,
        sizing_mode='stretch_width',
    )

    tmpl = pn.template.FastListTemplate()
    tmpl.main.append(test)

    tmpl.servable()

```

Error:

```
  File "C:\Users\jhsmi\Miniconda3\envs\py38_panel\lib\site-packages\bokeh\core\property\bases.py", line 452, in _validate_type_param
    raise ValueError("setting 'help' on type parameters doesn't make sense")
ValueError: setting 'help' on type parameters doesn't make sense
```

When I change `Color` to, for example, `Number`,  I don't have any problems.

#### ALL software version info
```
panel==0.12.6
bokeh==2.4.2
Windows 10
```

Hoxbro:
As a workaround you can change None to a color like red. 

``` python 
import param
from panel.reactive import ReactiveHTML
import panel as pn


class TestReactive(ReactiveHTML):

    bg_color = param.Color(default="red", doc="Color of the background")

    _template = """
        <div id="container" style="background-color: ${bg_color}; width:100%; height:100%;"><p>${bg_color}</p></div>
    """


if __name__ == "__main__" or __name__.startswith("bokeh_app"):
    test = TestReactive(
        height=500,
        sizing_mode="stretch_width",
    )

    tmpl = pn.template.FastListTemplate()
    tmpl.main.append(test.param.bg_color)
    tmpl.main.append(test)

    tmpl.servable()
```

jbednar:
Thanks, @Hoxbro. Sounds like the title of this issue could be made more specific to focus on the use of None in such a case.

maximlt:
Running this snippet is enough to get the error:
```python
import param
from panel.reactive import ReactiveHTML

class TestReactive(ReactiveHTML):

    bg_color = param.Color(default=None)

    _template = """
    <div id="container" style="width:100%; height:100%;"><p>test</p></div>
    """
```

[...]

PR #3070, opened 2022-01-07
Allow to convert a Color Parameter with `allow_None=True` to a Bokeh property

maximlt:
Fixes https://github.com/holoviz/panel/issues/3058

```python
# The Bokeh Color property has `_default_help` set which causes
# an error to be raise when Nullable is called on it. This converter
# overrides the Bokeh _help to set it to None and avoid the error.
```

While this fixes the issue linked I'm not sure it's a good long term solution:
* It might well be a bug in Bokeh
* The only two properties that have `_default_help` set are `Color` and `Alpha` (`Alpha` causes no problem since it has no Parameter type associated). But if a mapped property (e.g. `String`) gets its `_default_help` set then we'll likely encounter the same error.

### DIFF
--- a/panel/links.py
+++ b/panel/links.py
@@ -50,7 +50,6 @@ PARAM_MAPPING = {
         if isinstance(p.class_, type) and issubclass(p.class_, param.Parameterized) else
         bp.Any(**kwargs)
     ),
-    pm.Color: lambda p, kwargs: bp.Color(**kwargs),
     pm.DataFrame: lambda p, kwargs: (
         bp.ColumnData(bp.Any, bp.Seq(bp.Any), **kwargs),
         [(bp.PandasDataFrame, lambda x: ColumnDataSource._data_from_df(x))]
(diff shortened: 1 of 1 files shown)


philippjfr:
@mattpap Any idea what best to do here? Seems to me like this either shouldn't be an error condition. All properties should let you wrap it in `Nullable` without erroring.

mattpap:
> Any idea what best to do here? Seems to me like this either shouldn't be an error condition. All properties should let you wrap it in Nullable without erroring.

The error condition is there to help one realize something is off with their properties' setup, e.g. having:
```py
prop = Nullable(Color(help="help"))
```
instead of
```py
prop = Nullable(Color, help=help)
```
The issue comes from different treatment of top-level properties, which become basis for descriptors, and inner properties (type parameters), which only serve type-related purpose (validation, serialization).

mattpap:
At least in `branch-3.0`, one can wrap `Color` in `Nullable`, unless explicit `help` is given:
```py
In [1]: from bokeh.models import *

In [2]: from bokeh.core.properties import *

In [3]: class T0(Model):
   ...:     p0 = Color(help="p0")
   ...: 

In [4]: class T1(Model):
   ...:     p1 = Nullable(Color, help="p1")
   ...: 

In [5]: class T2(Model):
   ...:     p2 = Nullable(Color(help="p2"))
   ...:    
ValueError: setting 'help' on type parameters doesn't make sense
```

mattpap:
One thing we may consider doing, is merging help, but that may have limited utility.

philippjfr:
Thanks @maximlt. Will merge with the workaround.

PR #3072, opened 2022-01-07
Compatibility with Bokeh 3.0 branch

philippjfr:
As work on the Bokeh 3.0 branch is picking up we need to make sure that we keep up to date with it to report back on compatibility concerns particularly where the shadow DOM and layout changes are concerned.


philippjfr:
Note @MarcSkovMadsen that with the shadow DOM changes in Bokeh all the nicely crafted CSS changes, e.g. to adapt the CSS styling for the fast templates, will be completely broken. We will have to have some discussion with @mattpap to work out a transition plan as this will affect a lot of users. I'll invite you to our weekly check-in next week so we can discuss what to do here.

PR #3083, opened 2022-01-09
Expose the `orientation` parameter of the button group widgets

maximlt:
This PR simply exposes a Bokeh property not exposed by Panel.

image

Is the CSS of some templates to adapt for this kind of change or is this good to go?


philippjfr:
Should be good to go. Thanks @maximlt !

Issue #2865, opened 2021-10-29
DynamicMap is not updated when embedded in dynamic Tabs


pat-schmitt:
It seems like holoviews `DynamicMap` is not updated when embedded in panel `Tabs` when `dynamic` is set to `True`.

See example below with two videos (first with `dynamic=False`, second with `dynamic=True`). Only with `dynamic=False` the figure is updated as expected. 

I am not sure if it is a holoviews or a panel problem.

#### ALL software version info
panel=0.12.4
holoviews=1.14.6
bokeh=2.4.0

#### Complete, minimal, self-contained example code that reproduces the issue (with `dynamic=True`)

```
import numpy as np
import holoviews as hv
from holoviews import opts
from holoviews.streams import Stream, param
import panel as pn

hv.extension('bokeh')
pn.extension()

lin = np.linspace(-np.pi,np.pi,300)

def lissajous(t, a=3, b=5, delta=np.pi/2.):
    return (np.sin(a * t + delta), np.sin(b * t))

def lissajous_crosshair(t, a=3, b=5, delta=np.pi/2):
    (x,y) = lissajous(t,a,b,delta)
    return hv.VLine(x) * hv.HLine(y)

path = hv.Path(lissajous(lin))

Time = Stream.define('Time', t=param.Number(default=0.0, doc='A time parameter'))
time = Time(t=np.pi/4)
dmap = hv.DynamicMap(lissajous_crosshair, streams=[time])

def button_click(event):
    dmap.event(t=button.clicks)

button = pn.widgets.Button(name='change dmap')
button.on_click(button_click)

app = pn.Tabs(('Plot', path * dmap),
              ('Button', button),
              dynamic=True)

app.show()
```

#### Video with `dynamic=False`

https://user-images.githubusercontent.com/50843444/139404906-da30d2f4-0467-43d5-b14a-172060ea9fe1.mp4

#### Video with `dynamic=True`

https://user-images.githubusercontent.com/50843444/139404985-893f7fd2-71b1-4d71-bc48-b01933eaf967.mp4

fmaussion:
Just to add some context here: this bug currently prevents us to use dynamic plots for a new app we have just released (https://bokeh.oggm.org/mb_simulator/app) which is quite expensive in plots hidden by tabs. Setting new parameters in the app therefore makes a quite bumpy user experience, which is sad (we cant rule out that we are doing things wrongly though)

fmaussion:
[...]

PR #3091, opened 2022-01-11
Fix Tabs reuse of rendered model in dynamic mode

philippjfr:
The logic for reusing previously rendered models in `Tabs` dynamic mode was wrong. Here we ensure the rendered model is correctly reused.

As a side effect this also happens to fix https://github.com/holoviz/panel/issues/2865


philippjfr:
Windows issues unrelated will fix CI in a separate PR.

PR #3092, opened 2022-01-11
Use https instead of the git:// protocol to clone in GHA

maximlt:
Github has [improved its Git protocol security](https://github.blog/2021-09-01-improving-git-protocol-security-github/) which means that usage of `git://` has been disabled. This broke the CI on Windows.


philippjfr:
Thanks @maximlt !

Issue #2802, opened 2021-10-06
Toast Widget (pop up notifications)


MarcSkovMadsen:
Often times you would like to inform the user of progress, events, errors etc. One way to do this is via a `Toast`.

## Potential Requirements

- success+error message
  - Other message types, for example corresponding to `button_type`
- string message
- Markdown string message
- Supports icons
- Can trigger event/ action when clicked
- Supports buttons
- General pane message like `HoloViews` plot or `Markdown` pane
- Prevents duplicate messages/ Allows duplicate messages
- Can provides list of historical messages and maybe even show them to the user
- Can send Web Notifications

## Potential JS Libraries

- [Notyf](https://github.com/caroso1222/notyf)
- [Toastr](https://codeseven.github.io/toastr/demo.html)
- [Sweet Alert](https://github.com/sweetalert2/sweetalert2) (Very advanced features)
- [Notistack](https://iamhosseindhv.com/notistack/demos) (react based)
- [Toastify](https://github.com/apvarun/toastify-js)
- [Notie](https://github.com/jaredreich/notie) (Works like an Alert).
- [PNotify](https://sciactive.com/pnotify/) (Supports showing lots of toast using a system called *modalish*).

You can find a very long list here https://openbase.com/categories/js/best-javascript-toast-libraries

## Panel Example - Notyf

See https://discourse.holoviz.org/t/toast-a-pop-up-notification/2903

![toast](https://user-images.githubusercontent.com/42288570/136136746-b2ba3b84-daef-462e-b2e1-e5c034141a47.gif)

maximlt:
@MarcSkovMadsen can we close this issue now we have `Notfications`? :)

MarcSkovMadsen:
Yes

PR #3093, opened 2022-01-11
Implement Notifications API for templates

philippjfr:
Adds `Template.notifications` which provides an API to display notifications/toasts on the frontend. Each template can provide its own `NotificationArea` implementation based on whatever CSS/JS framework you want. Currently we ship three `NotificationArea` implementations:

1. `BootstrapNotificationArea` builds on Bootstrap's [built-in support for toasts](https://getbootstrap.com/docs/4.3/components/toasts/)
2. `MaterialNotificationArea` builds on the Material component [built-in "snackbar" support](https://material.io/components/snackbars/web)
3. `ToastedNotificationArea` builds on [toasted.js](https://github.com/shakee93/toastedjs) a very light-weight toast implementation

## Implementation

The Bootstrap and Material implementations both rely on adding `Notification` instances as children of the `NotificationArea` while the `Toasted` implementation simply creates a bokeh `DataModel` which is linked to the notification via callbacks in JS.

## API

Currently I envision the following API:

- `NotificationArea.send(body: str, header=None: str, duration=3000: int, autohide=True: boolean): Notification`: Displays a notification and returns a Notification instance
- `NotificationArea.position = 'left' | 'right'`: Controls whether to display notification on bottom-left or bottom-right
- `NotificationArea.clear()`: Clears all notifications
- `Notification.destroy()`: Destroys the notification

## Styling

Of course styling depends a little bit on the specific implementation however most support some variation of `type`: 'info' | 'warning' | 'error' borrowing on the bootstrap API. So at minimum this is what I would expose.

## Demo

notifications

## ToDo

- [ ] Finalize API after feedback
- [ ] Decide on styling API
- [ ] Add docs
- [ ] Add tests

### DIFF
--- a/panel/links.py
+++ b/panel/links.py
@@ -37,36 +37,26 @@ class Parameterized(bokeh.core.property.bases.Property):
         raise ValueError(msg)
 
 
+class ParameterizedList(bokeh.core.property.bases.Property):
+    """ Accept a list of Parameterized objects.
+
+    This property only exists to support type validation, e.g. for "accepts"
+    clauses. It is not serializable itself, and is not useful to add to
+    Bokeh models directly.
+
+    """
+
+    def validate(self, value, detail=True):
+        super().validate(value, detail)
+
+        if isinstance(value, list) and all(isinstance(v, param.Parameterized) for v in value):
+            return
+
+        msg = "" if not detail else f"expected list of param.Parameterized, got {value!r}"
+        raise ValueError(msg)
+
 
 _DATA_MODELS = weakref.WeakKeyDictionary()
-
-PARAM_MAPPING = {
-    pm.Array: lambda p, kwargs: bp.Array(bp.Any, **kwargs),
-    pm.Boolean: lambda p, kwargs: bp.Bool(**kwargs),
-    pm.CalendarDate: lambda p, kwargs: bp.Date(**kwargs),
-    pm.CalendarDateRange: lambda p, kwargs: bp.Tuple(bp.Date, bp.Date, **kwargs),
-    pm.ClassSelector: lambda p, kwargs: (
-        (bp.Instance(DataModel, **kwargs), [(Parameterized, create_linked_datamodel)])
-        if isinstance(p.class_, type) and issubclass(p.class_, param.Parameterized) else
-        bp.Any(**kwargs)
-    ),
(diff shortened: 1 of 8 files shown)


philippjfr:
@maximlt @MarcSkovMadsen @jbednar This is very much a work in progress but if you have comments/suggestions about the API that would be very helpful.

MarcSkovMadsen:
My thought was that I would have added the toast as a new component. I made an example on discourse a while back that shows the concept. I think it's more flexible not to tie it to a template, but provide the toast as a basic building block.

jbednar:
Looks good to me!

Like Marc, I do think it might not be necessary to tie it to a template. E.g. couldn't we have the same app in a notebook, but then displaying the notification some other way?

MarcSkovMadsen:
The request for a standalone toast is here https://github.com/holoviz/panel/issues/2802

The basic implementation is here https://discourse.holoviz.org/t/toast-a-pop-up-notification/2903

MarcSkovMadsen:
One question I have is whether Toasts should be layouts that can display Panel components? The toast I proposed can only display text. That is my need.

philippjfr:
The reason why I think it makes sense as part of a template or as a global object is that it doesn't respect any of the layout placement you might otherwise do with it, e.g. placing it inside a Row/Column effectively makes no sense, because the actual toasts will appear in the fixed absolute position you specify. Adding it as a root with .servable does make sense I suppose and you can still do that with my version.

If we came up with some UI that worked in notebooks I'd probably just move the notifications object onto pn.state.

philippjfr:
Okay, so now I'm leaning towards actually making it global, i.e. providing a `pn.config.notifications` option to enable the notification area and then store the object on `pn.state.notifications`.

philippjfr:
Okay based on the feedback I'm now leaning towards the `pn.config.notifications` and `pn.state.notifications` option. Enabling the option will enable notifications both in the notebook and in the served app:

notifications_nb

MarcSkovMadsen:
Seems like a good solution to me? Should the modal one day be similarly refactored into a global option?

philippjfr:
>Seems like a good solution to me? Should the modal one day be similarly refactored into a global option?

If we can make a modal that works in the notebook I'd be very open to that.

PR #3096, opened 2022-01-13
Handle authentification errors

philippjfr:
Avoids redirect loop when the auth provider errors out or denies access. The Auth error template can be customized using the `--oauth-error-template`CLI argument but by default an access denied error will look something like this:

Screen Shot 2022-01-13 at 20 37 55
.


MarcSkovMadsen:
You once said that the word "Fast" should be removed from the Template. But looks good.

PR #3100, opened 2022-01-14
add document argument to django.py autoload_js_script call

### LINKED ISSUE #3099: Django autoload_handle broken
Panel =  0.13.0a25
Bokeh  = 2.4.2
Django = 2.2.14

When loading a Panel app embedded in Django, the `AutoloadJsConsumer` call just hangs. After stepping through the code it 
appears there is an error, which causes it to enter an eternal loop:

```python
> /Users/rditlsc9/miniconda/envs/tethys-vtime/lib/python3.7/site-packages/panel/io/django.py(37)autoload_handle()
-> js = autoload_js_script(resources, session.token, element_id, app_path, absolute_url)

TypeError: autoload_js_script() missing 1 required positional argument: 'absolute_url'
```

It appears that #2919 changed the signature of `autoload_js_script`, but the call to it in `panel/io/django.py:autoload_handle` wasn't updated accordingly.

As a side note - is there a better way to get this type of error to log? I wasn't able to see any indication of an error until I stepped through the code in a debugger.

sdc50:
adds missing argument in call to `autoload_js_script`

fixes #3099

### DIFF
--- a/panel/io/django.py
+++ b/panel/io/django.py
@@ -34,7 +34,7 @@ async def autoload_handle(self, body):
         server_url = None
 
     resources = self.resources(server_url)
-    js = autoload_js_script(resources, session.token, element_id, app_path, absolute_url)
+    js = autoload_js_script(session.document, resources, session.token, element_id, app_path, absolute_url)
 
     headers = [
         (b"Access-Control-Allow-Headers", b"*"),


philippjfr:
Thanks @sdc50!

PR #3111, opened 2022-01-18
Implement Tabulator buttons

philippjfr:
If you want to trigger custom actions by clicking on a table cell you may declare a set of `buttons` that are rendered in columns after all the data columns. To respond to button clicks you can register a callback using the `on_button_click` method:

button_click

### DIFF
--- a/examples/reference/widgets/Tabulator.ipynb
+++ b/examples/reference/widgets/Tabulator.ipynb
@@ -11,7 +11,7 @@
     "import pandas as pd\n",
     "import panel as pn\n",
     "\n",
-    "pn.extension('tabulator')"
+    "pn.extension('tabulator', css_files=[pn.io.resources.CSS_URLS['font-awesome']])"
    ]
   },
   {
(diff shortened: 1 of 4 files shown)


MarcSkovMadsen:
Wow. Some one thinks tables are important. I do too. Love it.

MarcSkovMadsen:
I wondered about the use cases. I would say some very popular ones are "download file" and "open url". Are they easy to implement with these buttons @philippjfr ?

maximlt:
I've just tried it and the event returns the row number, not the row index. Which means that to get a value one has to write something like `df.loc[df.index[event.row], 'URL']`. Returning the row index might be a little simpler for users, or the whole row data.

philippjfr:
That's what we do for selections and the expanded parameter as well, you should just be able to use `iloc` like this:  `df.iloc[event.row]`

maximlt:
I just find working with indexes usually more natural than working with row numbers but if this is what is already done elsewhere then let's favor being consistent. (This is concise enough anyway: `df.iloc[event.row]['URL']`)

jbednar:
Sounds like that iloc usage ought to be illustrated in the docs?

contang0:
is there a way to move button position to the left of the table?

PR #3116, opened 2022-01-19
Fix Django resource handling

philippjfr:
@sdc50 Before this PR Django wasn't working at all for me since it couldn't load some resources. Did you have similar issues?


sdc50:
> Did you have similar issues?

I haven't been able to reproduce this issue. For me using both `inline` and `cdn` worked at least sense v0.13.0a28.

Note: I haven't been able to get the bokeh resources `server` option to work for me because of the URL it tries to use.

Issue #3086, opened 2022-01-10
Add template kwargs to servable


MarcSkovMadsen:
I've been told that Panel would like to promote a code flow where the template is specified in `pn.extension`.

For me the current way it works is cumbersome. Everytime I use that flow I would need to extract the template and set its parameters. Something like `site`, `accent_base_color`, `header_background` almost always need to be set. And sometimes `sidebar_width` etc. also needs to be set.

Please support providing these arguments either in `pn.extension` or `.servable`.

I would prefer to add **kwargs** to `.servable` or a dict argument `template_kwargs` to `pn.extension`.

## Example

How could I easily set the above template parameters in the example below?

```python
import plotly.express as px


def plot(count):
    x = [str(i) for i in range(0, count)]
    y = list(range(0, count))
    return px.bar(x=x, y=y)


import panel as pn

pn.extension("plotly", sizing_mode="stretch_width", template="fast")

count = pn.widgets.IntSlider(name="Count", start=10, end=50, value=10).servable(
    location=True, area="sidebar"
)
icount = pn.bind(plot, count)  # Makes function rerun when the widget value is changed
pn.panel(icount).servable(title="Slider Example", area="main")
```

philippjfr:
+1 on adding `template_kwargs` to `pn.extension` and by extension `pn.config`.

philippjfr:
It sounds like we decided not to do this and instead use `pn.state.template.param.update`.

PR #3120, opened 2022-01-20
Adds template_kwargs to config

MarcSkovMadsen:
Addresses https://github.com/holoviz/panel/issues/3086

Makes it easy to configure the template.

```python
import panel as pn

pn.extension(sizing_mode="stretch_width", template="fast", template_kwargs=dict(header_background="red"))

pn.panel("hello").servable()
```

image

Without this you would have to do something like 

```python
import panel as pn

pn.extension(sizing_mode="stretch_width", template="fast")
pn.state.template.header_background = "red"
# ...

pn.panel("hello").servable()
```

which is not very concise if you want to set many attributes.

I could not find a natural place to document this @philippjfr . I would be happy to update some docs if you could point out where. You are welcome to consider whether you want this PR. I can live with setting attributes of `pn.state.template`. But this api would be "nice to have". Thanks.

### DIFF
--- a/panel/config.py
+++ b/panel/config.py
@@ -126,6 +126,9 @@ class _config(_base_config):
     template = param.ObjectSelector(default=None, doc="""
         The default template to render served applications into.""")
 
+    template_kwargs = param.Dict(default={}, instantiate=True, doc="""
+        A dictionary of arguments to apply to the template specified.""")
+
     theme = param.ObjectSelector(default='default', objects=['default', 'dark'], doc="""
         The theme to apply to the selected global template.""")
 
--- a/panel/io/state.py
+++ b/panel/io/state.py
@@ -500,7 +500,7 @@ class _state(param.Parameterized):
             return self._templates[self.curdoc]
         elif self.curdoc is None and self._template:
             return self._template
-        template = config.template(theme=config.theme)
+        template = config.template(theme=config.theme, **config.template_kwargs)
         if self.curdoc is None:
             self._template = template
         else:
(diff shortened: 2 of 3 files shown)


philippjfr:
I'm in favor of this but note that you can set multiple parameters on any object with:

```python
pn.state.template.param.update(
    header_background = "red",
    ...
)
```

panel/config.py
```diff
@@ -126,6 +126,9 @@ class _config(_base_config):
     template = param.ObjectSelector(default=None, doc="""
         The default template to render served applications into.""")
 
+    template_kwargs = param.Dict(default={}, instantiate=True, doc="""
```
philippjfr, line 129: If we allow this to be changed then it should sync the changes with the template when updated for consistency. If we don't want to allow that we should set this to `constant`.

MarcSkovMadsen:
Lets think a bit more about maybe just investing more in

```python
pn.state.template.param.update(
    header_background = "red",
    ...
)
```

If this was widespread knowledge then it would be more valuable.

philippjfr:
>If this was widespread knowledge then it would be more valuable.

We should really write a quick primer on the most important param features.

MarcSkovMadsen:
After thinking about it. I would recommend closing this PR in favor if me starting to use

```python
pn.state.template.param.update(
    header_background = "red",
    ...
)
```

On discourse and in notebook guides when needed. We would avoid more complexity in the api and share knowledge about a useful `param` feature.

**Agree?**

philippjfr:
Sounds good! I've started an issue about writing a primer/cheatsheet for Param which I'd like to reproduce in Panel in some form: https://github.com/holoviz/param/issues/593

Issue #3085, opened 2022-01-10
Enable intellisense for arguments to widgets, panes, layouts etc.


MarcSkovMadsen:
One challenge when using Panel and the rest of the HoloViz ecosystem is the lack of useful intellisense in a modern editor and IDE.

## Panel

No help on input arguments. Means you will have to remember or look up things. This slows you down and adds mental strain

![image](https://user-images.githubusercontent.com/42288570/148723105-f72ec81f-683f-4da7-9190-a58fb4b04e2b.png)

## Streamlit

Provides much more help when you need it.

![image](https://user-images.githubusercontent.com/42288570/148723192-f580381c-7a31-4001-a3d1-01fd980a7bd7.png)

## Additional Context

I believe there are ways to **inherit** docs/ arguments from parent classes. I believe some frameworks have code we could copy.

Alternatively we would have to add custom constructors all over the Panel code.

maximlt:
I'm also using VSCode and I agree that this should be improved!

philippjfr:
Kind of frustrated by this, we explicitly enabled [signatures](https://www.python.org/dev/peps/pep-0362/) on Parameterized class constructors which should be the accepted way to declare the accepted kwargs. Anyone know what Intellisense uses to extract the signature?

brettcannon:
Assuming you're using Pylance as your language server, it's primarily based on type hints. If you believe you implemented them appropriately and it still isn't working then please open an issue at https://github.com/microsoft/pylance-release so the team can look into it.

If you're using Jedi then please open an issue at https://github.com/davidhalter/jedi .

patrick91:
Maybe adding a `py.typed` file might help https://blog.whtsky.me/tech/2021/dont-forget-py.typed-for-your-typed-python-package/

MarcSkovMadsen:
Maybe we should look into what, Django models, attrs, data classes, traits, pydantic etc do? They must have solved this problem?

brettcannon:
For attrs at least, they implemented https://github.com/microsoft/pyright/blob/main/specs/dataclass_transforms.md which I believe is in the process of becoming a proposed PEP.

MarcSkovMadsen:
I've created a Feature Request with [Param](https://param.holoviz.org/) because it has to be solved there https://github.com/holoviz/param/issues/587. I would suggest continuing the discussion there. Thanks.

MarcSkovMadsen:
[...]

PR #3132, opened 2022-01-22
Stub experiments

MarcSkovMadsen:
Addresses https://github.com/holoviz/panel/issues/3085

So far this is an experiment to be able to automatically create stub (`.pyi`) files to provide information in modern editors and IDEs.

I would like to make some general functionality that can take any project with `param.Parameterized` classes and add `.pyi` stubs. Either directly to the project or to some alternative folder or package. What is best is to be determined.

But I would like to be able to point it the Panel code as well as all the project I have for work and the Panel extensions I'm maintaining. I would like other `param` users to be able to do the same.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -123,3 +123,4 @@ discover/
 node_modules/*
 /tmp
 .devcontainer/*
+out
--- a/panel/__init__.py
+++ b/panel/__init__.py
@@ -1,3 +1,95 @@
+"""# Panel is a high level app and dashboarding framework
+
+Works with the tools you know and ❤️. Welcome to our community 👪
+
+- Get Started: https://panel.holoviz.org/
+- Join the Community: https://discourse.holoviz.org/
+- Contribute: https://github.com/holoviz/panel
+
+Follow us on [Twitter](https://twitter.com/Panel_org) or [LinkedIn](https://www.linkedin.com/company/79754450).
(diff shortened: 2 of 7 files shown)


philippjfr:
Thanks for all your experiments here. Agree with all your points, here are my expectations for this:

1. Ideally the code to generate the stubs moves to Param
2. We will not check the .pyi files into the repo but generate them at build time and include them in the package.
3. Adding docstrings is **always** welcome and we should decide asap whether we should convert them to markdown

MarcSkovMadsen:
Totally agree with your points. 

There are additional options 

I. Stubs can be distributed as a separate package?

II. The functionality could live in a separate repo/ for a while until tried and tested on a multitude of projects?

MarcSkovMadsen:
Regarding markdown. I prefer Markdown. Can this be Myst flavored Markdown now default for Sphinx? It would support better Sphinx generated docs as Panel matures I hope.

philippjfr:
Highly appreciate the attempt here and won't delete the branch so we can come back to it but the stubgen definitely shouldn't live in Panel and I've started experimenting with other approaches.

MarcSkovMadsen:
Please note this branch contains code for converting all Parameter types used in Panel to type annotations. Also handles default values and Allow_None arguments.

PR #3155, opened 2022-01-30
Expose save_png in save

### LINKED ISSUE #3005: Enable saving to BytesIO png.
I've often wanted to make it possible for users to download a `.png` of a specific panel object. And it is a painful process as the `.save` method only supports saving to `.png` if the filename provided is a string. But for filedownload you need to save to a `BytesIO`, i.e. you would be providing a `BytesIO` as `filename`.

Here is a user also having the problem https://discourse.holoviz.org/t/how-to-save-panel-object-to-png-without-file-extension-when-using-bytesio/3152

## Solution

Expose the `as_png` variable in the `.save` method to the user.

## Workaround

The example on discourse also contains a workaround using a temporary file.

Hoxbro:
Fixes #3005 

https://user-images.githubusercontent.com/19758978/151700567-3a338759-4d8a-4064-a5a3-a1bcf5e7e448.mp4
<details>
  <summary> Code </summary>

``` python
from io import BytesIO

import hvplot.pandas
import panel as pn
import pandas as pd
import param

ACCENT_COLOR = "#0072B5"


class Explorer(param.Parameterized):
    df = pd.DataFrame({"y": range(10)})

    def __init__(self, **params):
        super().__init__(**params)
        self.plot = pn.Column(
            self.df.hvplot().opts(color=ACCENT_COLOR, line_width=6, responsive=True),
            self.df.hvplot().opts(color=ACCENT_COLOR, line_width=6, responsive=True),
        )
        self.save_to_png = pn.widgets.FileDownload(
            filename="timeseries.png",
            label="Download .png",
            button_type="primary",
            callback=self.download_column,
        )

    def download_column(self):
        self.save_to_png.loading = True

        b = BytesIO()
        self.plot.save(b, as_png=True)
        b.seek(0)

        self.save_to_png.loading = False
        return b


explorer = Explorer()

pn.template.FastListTemplate(
    site="Awesome Panel",
    title="Application with .png download",
    sidebar=[explorer.save_to_png],
    main=[explorer.plot],
    accent_base_color=ACCENT_COLOR,
    header_background=ACCENT_COLOR,
).servable()
```

</details>

I don't know if this should also be made available for saving the template, right here: 
https://github.com/holoviz/panel/blob/5bd9fc8187bb3a663d29ce6ff200d88b72991123/panel/template/base.py#L247-L249

I have observed that sometimes the saved plot is cropped, but I don't think this is related to the changes in the PR.
**Cropped**
cropped

**Not cropped**
not cropped

### DIFF
--- a/panel/io/save.py
+++ b/panel/io/save.py
@@ -98,7 +98,7 @@ def save_png(model, filename, resources=CDN, template=None, template_variables=N
         if img.width == 0 or img.height == 0:
             raise ValueError("unable to save an empty image")
 
-        img.save(filename)
+        img.save(filename, format="png")
     except Exception:
         raise
     finally:
@@ -148,7 +148,8 @@ def file_html(models, resources, title=None, template=BASE_TEMPLATE,
 def save(panel, filename, title=None, resources=None, template=None,
          template_variables=None, embed=False, max_states=1000,
          max_opts=3, embed_json=False, json_prefix='', save_path='./',
-         load_path=None, progress=True, embed_states={}, **kwargs):
+         load_path=None, progress=True, embed_states={}, as_png=None,
+         **kwargs):
     """
     Saves Panel objects to file.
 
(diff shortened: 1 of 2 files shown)


philippjfr:
Thanks @Hoxbro!

PR #3160, opened 2022-01-31
CI: fix Macos tests workflow

maximlt:


maximlt:
@philippjfr this is ready to me. It's now the Github action that sets up `conda` that defines from which channels the packages should be pulled. And I've added `nodefaults` and the strict option to attempt to prevent pyctev (conda actually) to update conda while running a task with multiple `conda install` commands, which is what was breaking the task on mac-os.

(I also tried quickly to see what would happen on Python 3.10 and hit solve conflicts, so this is going to be for another time!)

PR #3168, opened 2022-02-02
Added explicit comparison to JS

jbednar:
Added a section to comparisons.rst comparing and contrasting using Panel with developing directly in JavaScript/HTML/CSS.


doc/about/comparisons.rst
```diff
@@ -59,24 +60,46 @@ Voila is a technology for deploying Jupyter notebooks (with or without Panel cod
...
 
 Overall, Panel can be used in a much wider range of applications than streamlit, including exploratory data analysis and capturing a reproducible workflow in a Jupyter notebook, developing a simple streamlit-like app, or developing complex, multi-page responsive apps, all without having to switch frameworks or learn a new set of tools. Panel thus supports the entire life cycle of data science, engineering, or scientific artifacts, not just a narrow task of developing a specific type of simple app.
+
+
+Comparing Panel and JavaScript
+------------------------------
+
+Whenever you evaluate any Python framework for building JavaScript/HTML/CSS web apps, it's important to consider the baseline alternative of just writing JavaScript, HTML, and CSS directly. Writing JS/HTML/CSS will give you full control over all aspects of the resulting applications, allowing you to tailor the complete look and feel and behavior to match even the most precise requirements. Moreover, JS/HTML/CSS apps can be deployed on any web server without needing any special consideration to running a Python process, which makes deployment and scaling much more straightforward.
```
philippjfr, line 89: I'm not sure this is a sensible comparison really. Comparing against a completely static website seems odd since it's they really do not compete on functionality in any standard use case. Most websites that have some level of interactive features will be built using HTML/JS/CSS yes, but in almost all cases they'll have some server in the background whether that is a node.js server, flask or some framework in yet another language.

philippjfr: >We could add an additional section comparing against HTML/JS/CSS + a generic server like flask, where the considerations would be a bit different

Definitely not an additional section, I don't think it's worth comparing static HTML/JS/CSS vs Panel at all. They just don't compete sufficiently to warrant such a comparison imo.

jbednar: I have worked closely with multiple organizations that have a different standard for "production" apps and R&D work, with production apps expected to be JS. My intention (perhaps not quite realized!) with this section is to emphasize to people considering such an arrangement to be explicit about considering the pros and cons of translating a Python app to native JS/HTML/CSS (with a Python backend if needed, sure), or even against developing the app natively in JS/HTML/CSS. Specifically, such a translation ossifies and pins the app in place because of the complexity of the code involved and the unlikelihood that the domain experts speak JS natively, and that's fine if requirements are known or unchanging. But if requirements evolve either through discovery or through changes in the subject area (both of which are very common), people need to decide what's more important: easily tweaking style and response time of a certain button press? or being able to adapt the overall app to achieve the actual goal? I don't think people always approach the problem this way, and I want to make that an explicit part of the docs. That's also feedback I got at the PyData Dashboarding shootout -- shouldn't we compare against no Python web framework at all?

jbednar: And note that it's not necessarily a competition for the same applications, it's a matter of helping people make the right choice for their application.

doc/background/comparisons.md
```diff
@@ -51,23 +52,43 @@ Voila is a technology for deploying Jupyter notebooks (with or without Panel cod
...
+------------------------------
+
+Whenever you evaluate any Python framework for building JavaScript/HTML/CSS web apps, it's important to consider the baseline alternative of just writing JavaScript, HTML, and CSS directly. Writing JS/HTML/CSS will give you full control over all aspects of the resulting applications, allowing you to tailor the complete look and feel and behavior to match even the most precise requirements. Moreover, JS/HTML/CSS apps can be deployed on any web server without needing any special consideration to running a Python process, which makes deployment and scaling much more straightforward.
+
+Given those advantages, here's why you would want to use a high-level Python framework instead:
+
+- Does the app depend on sources of data or other tools already available in Python? If so it will normally be much easier to build in a Python framework like Panel.
+
+- Do the people building the app normally work in Python? If so, they are likely to be much more productive in Python, both directly for writing the app's code and indirectly by being able to use all of their usual support tools and infrastructure (and from not having to learn JS).
```
MarcSkovMadsen, line 86: In trading the users (i.e. traders) often have the requirement that they need to be able to read and review the app. They are actually responsible for what ever happens in the app whether its a model providing insights or prescriptions or whether its a button that triggers an algo to start trading by sending data to a REST API. So Python is a good choice for the entire app.

Issue #3170, opened 2022-02-03
Render with Panel button breaks with Jupyterlab root_dir


govinda18:
#### ALL software version info
Jupyterlab 3.2.8
Panel master

#### Description of expected behavior and the observed behavior
Hi team, jupyterlab allows you to set a custom root_dir and preferred_dir config for your server app. You can set it using the following:
```bash
$ jupyter notebook --generate-config
```

Then in your `~/.jupyter` folder you should find `jupyter_notebook_config.py` and there we can set a configuration like:
```python
## The directory to use for notebooks and kernels.
#  Default: ''
c.ServerApp.root_dir = '/'
## Preferred starting directory to use for notebooks and kernels.
#  Default: ''
import os
c.ServerApp.preferred_dir = os.getcwd()
```

Now when you launch jupyter, the jupyter extensions get file path relative to the `root_dir`. So suppose you launch your lab from `/home/gt` and root dir is set to `/` and we try to render a panel preview for a file located at `/home/gt/notebook.ipynb`, this will break as the path passed to the [PanelHandler.get](https://github.com/holoviz/panel/blob/master/panel/io/jupyter_server_extension.py#L83) is actually `home/gt/notebook.ipynb` but the bokeh utility [here](https://github.com/bokeh/bokeh/blob/branch-3.0/bokeh/command/util.py#L117) tries to get abspath which is relative to cwd (which is /home/gt) and it turns out to be `/home/gt/home/gt/notebook.ipynb`.

Here, what the PanelHandler should be doing is setting the path to bokeh as `root_dir + path`. A good example is the [FileContentsManager](https://github.com/jupyter-server/jupyter_server/blob/7615f32497df4f2a133eea451a11cb16b51dd8b1/jupyter_server/services/contents/filemanager.py#L46) of jupyter extension. As you can see [here](https://github.com/jupyter-server/jupyter_server/blob/7615f32497df4f2a133eea451a11cb16b51dd8b1/jupyter_server/services/contents/filemanager.py#L51), for all the utility it evaluates the path relative to the root dir.

#### Complete, minimal, self-contained example code that reproduces the issue

```
import panel as pn
pn.extension()
pn.pane.Markdown("# Test").servable()
```

This simple code throws 500 Internal server errors if the following configuration is done - 

```python
c.ServerApp.root_dir = '/'
import os
c.ServerApp.preferred_dir = os.getcwd()
```

[...]

PR #3172, opened 2022-02-03
Support jupyter server root_dir with lab extension

govinda18:
Jupyter sends path relative to the root_dir which can be different from the cwd. This commit fixes the lab extension for panel preview to account for root_dir.

Reference: #3170


philippjfr:
Amazing, thanks for contributing! Will merge when tests pass.

Issue #3131, opened 2022-01-22
Add docstrings and parameter docstrings where missing


MarcSkovMadsen:
Looking at the code while trying to autogenerate type stubs I realized that docstring are missing in a lot of places.

For example in in panel.widgets.slider

![image](https://user-images.githubusercontent.com/42288570/150634457-7bae5797-cdd2-4f09-863d-01f126e4a8a2.png)

This makes using Panel more difficult than it has to be.

When adding the docs I would recommand adding a link in the class docstring to the reference guide if it exists. This will make navigation and learning so much easier.

## Example

Adding the class string 

![image](https://user-images.githubusercontent.com/42288570/150634501-da83a0c9-586b-4207-986d-fcf646aa023a.png)

would enable me to navigate the docs much, much quicker

![image](https://user-images.githubusercontent.com/42288570/150634552-90bce52d-807a-4169-9dab-d7b99c096c6d.png)

than today

![image](https://user-images.githubusercontent.com/42288570/150634623-981f46bb-b9b9-408b-82a6-d48b18aceb8e.png)

@maximlt . I believe you would agree here :-)

jbednar:
Adding docstrings would be an excellent task for someone to take on; indeed, they are sorely missed! But I don't think it would be appropriate to add links to the reference guide into the docstrings as above, because for one thing the reference guide is _generated_ from the docstrings (and inspection). The reference guide would then be linking to itself when it is generated. Plus tying each class to a particular URL like that seems unlikely to be something we could keep up to date and consistent. But I _think_ we could provide some mechanism where a Param-based library registers a base URL for a superclass and then the appropriate URL could be generated and appended to the docstring as the class is created, which would ensure that it's done consistently.

I'd also like each of our Reference Gallery items to include the output of `help(obj)` for that obj, since that's typically complementary information that too few people know about.

maximlt:
Also +1 on the docstrings of course!

And as for the URL, my first reaction was no because I don't think this is
very common and maybe not so easy to maintain. But giving it more thoughts
it could be a great idea. Panel is not a traditional Python library, it's
very visual. When I write code that uses Panel I spend a lot of time in the
reference gallery, to get a better impression of how a component works and
what I can do with it. Take the example of the Tabulator widget, there's no
way its docstring could describe all the things you can do with it.

@jbednar Marc added a like to the reference gallery, that is manually
crafted. Not the API reference which is indeed automatically generated (I
never have a look at this page, there's usually enough information in the
[...]

PR #3176, opened 2022-02-06
Add docstrings to sliders module

MarcSkovMadsen:
A first step into solving https://github.com/holoviz/panel/issues/3131

An isolated PR for adding docstrings to the sliders module. I can use the feedback from this review if/ when I add more docstrings.

Please note some basic concepts

- I've used Markdown in docstrings because they are simple, known to lots of people, render nicely in VS Code.
- For each slider I've added a link to the reference gallery notebook. It will speed up navigation and learning so much in VS Code.
- For each slider I've added a small, working code example in a fenced code block. Again being able to see and copy-paste such an example easily in VS Code will just speed up development so much I believe.

But this is my opinion. Please review and provide yours. Thanks.

## Example

https://user-images.githubusercontent.com/42288570/152671609-7145c3ce-1316-46c0-a123-a81b4fed5d55.mp4


panel/widgets/slider.py
```diff
@@ -72,10 +75,10 @@ def _update_model(self, events, msg, root, model, doc, comm):
         return super()._update_model(events, msg, root, model, doc, comm)
 
 
-class ContinuousSlider(_SliderBase):
+class _ContinuousSlider(_SliderBase):
```
maximlt, line 78: That's an API breaking change which I believe isn't required for this documentation work.

panel/widgets/slider.py
```diff
@@ -121,32 +124,73 @@ def _get_embed_state(self, root, values=None, max_opts=3):
...
 
-    end = param.Number(default=1.0)
+    ## Example
 
-    value = param.Number(default=0.0)
+    ```python
+    import panel as pn
 
-    value_throttled = param.Number(default=None, constant=True)
+    pn.extension()
```
maximlt, line 138: I think I wouldn't include the import and the call to `.extension()`. I've had a look at the pandas' docs and they don't include any import: https://pandas.pydata.org/docs/reference/api/pandas.merge.html

panel/widgets/slider.py
```diff
@@ -121,32 +124,73 @@ def _get_embed_state(self, root, values=None, max_opts=3):
...
 
-    step = param.Number(default=0.1)
+    slider=pn.widgets.FloatSlider(value=0.5,start=0.0,end=1.0,step=0.1, name="Float Value")
```
maximlt, line 140: I'd rather format this with spaces and without assigning to a variable.
```
pn.widgets.FloatSlider(value=0.5, start=0.0, end=1.0, step=0.1, name="Float Value")
```

panel/widgets/slider.py
```diff
@@ -191,14 +264,38 @@ def _process_property_change(self, msg):
...
 
-    value = param.Parameter()
+    ```python
+    import panel as pn
+
+    pn.extension()
+
+    slider = pn.widgets.DiscreteSlider(
+        value="love",
+        options=list(["We", "love", "data", "apps"]),
```
maximlt, line 281: Maybe use numbers instead of strings in this example, it feels like a more natural usage of a discrete slider to me.

maximlt:
Thanks Marc for this! Do you have any clue on how the autodocumented API would render with these docstrings including Markdown? Do you know if there are any changes to make to the doc build to render Markdown (sphinx settings?)?

panel/widgets/slider.py
Hoxbro: Should either use `list` or `[]`

panel/widgets/slider.py
```diff
@@ -191,14 +264,38 @@ def _process_property_change(self, msg):
...
 
-    value_throttled = param.Parameter(constant=True)
+    options = param.ClassSelector(default=[], class_=(dict, list), doc="""
+        A list or dictionary of valid options.""")
 
-    formatter = param.String(default='%.3g')
+    value_throttled = param.Parameter(constant=True, doc="""The value of the slider. Updated when
+        the handle is released""")
+
+    # Why do we have both a format and formatter parameter?
```
Hoxbro, line 296: Just adding a comment here for now so this is not going to be lost.

panel/widgets/slider.py
```diff
@@ -573,41 +751,90 @@ def _sync_value(self, event):
...
     show_value = param.Boolean(default=False, readonly=True, precedence=-1, doc="""
         Whether to show the widget value.""")
 
-    start = param.Number(default=0., doc="Lower bound of the range.")
-
-    step = param.Number(default=0.1, doc="Slider and number input step.")
-
-    value = param.Range(default=(0, 1), doc="Current range value.")
-
     value_throttled = param.Range(default=None, constant=True)
```
Hoxbro, line 838: Missing doc

panel/widgets/slider.py
```diff
@@ -436,20 +583,51 @@ def _process_property_change(self, msg):
...
+        step=2,
+        name="A tuple of datetimes",
+    )
+    ```
+    """
 
-    value = param.Tuple(default=(None, None), length=2)
+    value = param.Tuple(default=(None, None), length=2, doc=
+        """The selected range as a tuple of values. Updated when one of the handles is
+        dragged. Supports datetime.datetime, datetime.date, np.datetime64 ranges.""")
```
Hoxbro, line 611: Will properly add and in the last sentence.

panel/widgets/slider.py
```diff
@@ -161,14 +205,43 @@ def _process_property_change(self, msg):
...
+        name="A datetime Value",
+    )
+    ```
+    """
+
+    value = param.Date(default=None, doc="""
+        The selected date value of the slider. Updated when the slider handle is dragged. Supports
+        datetime.datetime, datetime.date or np.datetime64 types.""")
+
+
```
Hoxbro, line 236: Remove extra space

panel/widgets/slider.py
```diff
@@ -121,32 +124,73 @@ def _get_embed_state(self, root, values=None, max_opts=3):
...
-class FloatSlider(ContinuousSlider):
+class FloatSlider(_ContinuousSlider):
+    """The FloatSlider widget allows selecting a floating-point value within a
+    set of bounds using a slider.
 
-    start = param.Number(default=0.0)
+    Reference: https://panel.holoviz.org/reference/widgets/FloatSlider.html
 
-    end = param.Number(default=1.0)
+    ## Example
```
Hoxbro, line 133: When looking at the GIF this seem to be a very big font. Maybe add another #?

panel/widgets/slider.py
MarcSkovMadsen: I will carve it out and make a separate PR. I did it because my understanding is that the ContinousSlider is used as a Parent class for other sliders and not exposed to users. There is no Gallery Reference Guide either.

I just did it because looking at the sliders module I looked a bit unfinished.

panel/widgets/slider.py
MarcSkovMadsen: I looked at the Streamlit documentation before hand. And they don't either. I chose to add it because then you get a working example. I will remove it to be in line with other libraries.

panel/widgets/slider.py
MarcSkovMadsen: I did assign to slider (and include `import panel` + `pn.extension` because then it was easy to copy code to use. I've removed it.

The space is there because that is how Black formats python code. I would argue we should keep the space 😄

MarcSkovMadsen: I've removed the `slider =`

MarcSkovMadsen:
No. I just no @jbednar stated he support use of Markdown.

I don't believe its documented anywhere how to build the documentation locally to be able to check it out.

panel/widgets/slider.py
MarcSkovMadsen: Thanks

MarcSkovMadsen:
Thanks. I've updated based on the suggestions from @maximlt and @Hoxbro

panel/widgets/slider.py
maximlt: Yes you're right I don't think this class should directly be used but since it's part of the public API (in Python the line between public and private API is somewhat blurry) as its name doesn't start with an underscore, we're less in our right to rename it as we'd like (some users may create their own widgets based on that class, who knows!). So we'd have to deprecate it first and then remove it in a future version.

panel/widgets/slider.py
maximlt: Yes I agree with you there's pros and cons between the two approaches. Seems like the one without the imports is the preferred one around (the easier to maintain too I guess).

panel/widgets/slider.py
maximlt: @MarcSkovMadsen unless I'm wrong in your code there were no spaces around the equal sign and after most commas, I'd be very surprised to see black format it this way since I assume it's still very close to PEP8. What black may do is to break the code into multiple lines, not sure that would be a good idea here.

maximlt:
> @maximlt . Regarding your question on Sphinx docs.
> 
> No. I just know @jbednar stated he support use of Markdown.
> 
> I don't believe its documented anywhere how to build the documentation locally to be able to check it out.

I believe it would be necessary to see how the embedded Markdown renders in the autodocumented API reference. I think the last time I checked the integration wasn't great, but I may be wrong so it's worth trying. The reference to build Panel's documentation is in the relevant Github Action workflow (I've never tried to build it locally!).

I know that personally I never look at the API page on the site since the Reference gallery already details the signature of Panel's components (and I have easy access to Panel's source code). Yet with a little bit more love the API page could be made more useful, so I'd really like to make sure the Markdown you intend to add integrates well and doesn't prevent us from using nice Sphinx directives like `versionadded` or `deprecated`.

maximlt:
Something worth noting for the links we plan to embed in the docstrings is that Panel's website is the documentation of the latest version and we don't yet have a system to keep the website of previous versions. So someone using let's say Panel 0.13 in the future maybe be directed to the site that will be then at version 2.6. There isn't much that can be done in this PR against that though!

panel/widgets/slider.py
Hoxbro: I don't know what black you are running @MarcSkovMadsen but my black gives the same output as @maximlt with spaces after comma and not after equal.

mattkram:
I was going to recommend looking into the MyST project regarding rendering Markdown do strings, but it is unfortunately not currently possible.  Here is a GitHub issue for reference: https://github.com/executablebooks/MyST-Parser/issues/228

It may be a good option should that become available in the future.

PR #3177, opened 2022-02-06
Add docstring to panel module

MarcSkovMadsen:
One of the first steps into addressing https://github.com/holoviz/panel/issues/3131

I've added a docstring to the panel module. This will help the user a lot when working in VS Code. 

I believe it will also help in Jupyter (when using `?panel`) and in autogenerated Sphinx documentation.

I can use the feedback from this review when/ if I add more documentation.

Please note I've

- Used Markdown
- Added urls for easy navigation
- Added two "core" examples with gifs
- Added emojis

I believe the will leave a good first impression and be very helpful whenever you work with Panel.

I added added a full featured docstring just to demonstrate what is possible. It might be better (?) to simplify and just include the intro section with the short text and hyperlink?

## VS Code Experience

https://user-images.githubusercontent.com/42288570/152672440-8a131ec7-93ba-419a-bda8-d23e41ae34cb.mp4

### DIFF
--- a/panel/__init__.py
+++ b/panel/__init__.py
@@ -1,13 +1,16 @@
-"""[Panel site](https://panel.holoviz.org/), [Getting started guide](https://panel.holoviz.org/getting_started/index.html), [Community forum](https://discourse.holoviz.org/), [Github](https://github.com/holoviz/panel), [Twitter](https://twitter.com/Panel_org), and [LinkedIn](https://www.linkedin.com/company/79754450)
+"""Panel is a high level app and dashboarding framework
+====================================================
 
-# Panel is a high level app and dashboarding framework
-
-Panel is an open-source Python library that lets you create custom interactive web apps and 
-dashboards by connecting user-defined widgets to plots, images, tables, or text. 
+Panel is an open-source Python library that lets you create custom
+interactive web apps and dashboards by connecting user-defined widgets
+to plots, images, tables, or text.
 
 Panel works with the tools you know and ❤️.
 
-![Panel Dashboard](https://assets.holoviews.org/panel/thumbnails/docstrings/panel.gif)
+.. figure:: https://user-images.githubusercontent.com/42288570/152672367-6c239073-0ea0-4a2b-a4c0-817e8090e877.gif
+   :alt: Panel Dashboard
+
+   Panel Dashboard
 """
 from . import layout # noqa
 from . import links # noqa
--- a/panel/__init__.py
+++ b/panel/__init__.py
@@ -5,9 +5,11 @@ Works with the tools you know and ❤️.
 
 [Check out](https://panel.holoviz.org/) | [Get started](https://panel.holoviz.org/getting_started/index.html) | [Join the community](https://discourse.holoviz.org/) | [Github](https://github.com/holoviz/panel) | [Twitter](https://twitter.com/Panel_org) | [LinkedIn](https://www.linkedin.com/company/79754450)
 
+![Panel Dashboard](https://user-images.githubusercontent.com/42288570/152672367-6c239073-0ea0-4a2b-a4c0-817e8090e877.gif)
+
 ## Interactive models with `.bind`
 
-![Interactive Model App](https://user-images.githubusercontent.com/42288570/150686594-21b03e55-79ef-406b-9e61-1764c6b493c3.gif)
+![Interactive Model App](https://user-images.githubusercontent.com/42288570/152672364-582016c1-3085-4895-9884-541b896c34a6.gif)
 
 You can use Panels `.bind` to bind your models to widgets.
 
@@ -55,7 +57,7 @@ This includes
 
 ## Interactive dataframes with `.interactive`
 
-![Interactive DataFrame App](https://user-images.githubusercontent.com/42288570/150683991-9cece6a1-3751-42d2-8256-505f5deb12be.gif)
+![Interactive DataFrame App](https://user-images.githubusercontent.com/42288570/152672366-39b8b964-5627-4448-9788-5faf86095c3c.gif)
 
 You can use [hvplot .interactive](https://hvplot.holoviz.org/user_guide/Interactive.html) and 
 Panels widgets to make your dataframes interactive.
--- a/panel/__init__.py
+++ b/panel/__init__.py
@@ -7,7 +7,7 @@ dashboards by connecting user-defined widgets to plots, images, tables, or text.
 
 Panel works with the tools you know and ❤️.
 
-![Panel Dashboard](https://user-images.githubusercontent.com/42288570/152672367-6c239073-0ea0-4a2b-a4c0-817e8090e877.gif)
+![Panel Dashboard](https://assets.holoviews.org/panel/thumbnails/docstrings/panel.gif)
 """
 from . import layout # noqa
 from . import links # noqa
--- a/panel/__init__.py
+++ b/panel/__init__.py
@@ -93,6 +93,8 @@ from . import links # noqa
 from . import pane # noqa
 from . import param # noqa
 from . import pipeline # noqa
+from . import reactive # noqa
+from . import viewable # noqa
 from . import widgets # noqa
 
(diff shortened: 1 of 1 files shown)


MarcSkovMadsen:
Gifs used in the documentation

interactive-model
interactive-dataframe
panel-dashboard

maximlt:
I think this is a little too much for main module docstring. I'd rather not have a long example like that to maintain in here (it's easy to forget updating this kind of code that is never executed, even more so for the GIFs), instead I'd rather just point to the documentation as you did in the beginning.

I've had the look at the output of `module?` for pandas, numpy, bokeh and xarray:
* xarray has nothing
* bokeh has a very short introduction and a link to its doc
* pandas has an introduction to what pandas is and its main features
* numpy has a short introduction, and some details about how to use to doc, the list of available subpackages and utilities and a few more things

Based on that, I'd rather just have a short introduction to Panel and point to the documentation. The content of the landing page could be used to fill in the docstring: https://pyviz-dev.github.io/panel/index.html

panel/__init__.py
```diff
@@ -1,10 +1,98 @@
...
+
+For more about hvplot `.interactive` check out this [blog post](https://towardsdatascience.com/the-easiest-way-to-create-an-interactive-dashboard-in-python-77440f2511d1).
+"""
 from . import layout # noqa
 from . import links # noqa
 from . import pane # noqa
 from . import param # noqa
 from . import pipeline # noqa
-from . import reactive # noqa
-from . import viewable # noqa
```
Hoxbro, line 7: This is properly a mistake after a rebase or similar.

Hoxbro:
I agree with the points made by @maximlt. Though, I don't think there is anything wrong with adding some more flavor to the documentation for a more visual library like panel than computational libraries like xarray, numpy, and pandas.  

I feel like the GIF's for the two examples are moving too fast for a new user and should properly be slowed down. I will properly also limit them to only do one thing, like show the change when clicking on buttons or the slider, and not show the shift to the dark theme. 

I will properly also split the creation of the widgets/plot and the theme up, again for simplicity so each line only does one thing.  

A personal preference is that I will use the Oxford comma when listing items.

MarcSkovMadsen:
Thanks @maximlt and @Hoxbro 

My reflections are 1) I also looked at the libraries you mentioned @maximlt and I was surprised. My thought is that we are front runners here. It seems python packages have not really started providing powerful information to users in VS code. 2) I can have a short or long version. But I would like to include the links. I see a lot of users that does not know our discourse, that has a hard time navigating our documentation etc.

panel/__init__.py
MarcSkovMadsen: Thanks for noticing

MarcSkovMadsen:
Based on the feedback I've pushed a "simplified" version.

https://user-images.githubusercontent.com/42288570/152727847-926f52c6-75ee-4ca1-93ef-1f864f037f35.mp4

maximlt:
Thanks for adapting it to make it shorter, it's going to make our life easier and the goal really is to help users to reach the documentation and Discourse. Because, after all, if you've already installed Panel and intend to use it, it's very likely that you already know what Panel can do (to a certain extent of course).

panel/__init__.py
```diff
@@ -1,3 +1,14 @@
+"""[Panel site](https://panel.holoviz.org/), [Getting started guide](https://panel.holoviz.org/getting_started/index.html), [Community forum](https://discourse.holoviz.org/), [Github](https://github.com/holoviz/panel), [Twitter](https://twitter.com/Panel_org), and [LinkedIn](https://www.linkedin.com/company/79754450)
+
+# Panel is a high level app and dashboarding framework
+
+Panel is an open-source Python library that lets you create custom interactive web apps and 
+dashboards by connecting user-defined widgets to plots, images, tables, or text. 
+
+Panel works with the tools you know and ❤️.
+
+![Panel Dashboard](https://user-images.githubusercontent.com/42288570/152672367-6c239073-0ea0-4a2b-a4c0-817e8090e877.gif)
```
philippjfr, line 10:
```suggestion
![Panel Dashboard](https://assets.holoviews.org/panel/thumbnails/docstrings/panel.gif)
```

jbednar:
> My thought is that we are front runners here. It seems python packages have not really started providing powerful information to users in VS code.

I would guess that other packages haven't found VSCode docstring-rendering support to be a priority because VSCode hasn't supported the docstring formats typical in the SciPy/PyData community, namely NumpyDoc or Google styles:

https://stackoverflow.com/questions/57017994/what-is-the-python-docstring-format-supported-by-visual-studio-code

That link is a bit old, and I can't tell if it's still true; hard to follow in discussions like https://github.com/microsoft/vscode-python/issues/675 . If VSCode does support NumpyDoc nowadays, that's a reason to bite the bullet and commit to using NumpyDoc throughout HoloViz tools.

On the other hand, I personally don't like the numpy conventions (or the RST syntax on which they are based) at all, and I'd much prefer Markdown, but I think there are still a couple blockers:

1. We need Sphinx to understand the docstrings, so that we can generate Reference (API) documentation. Is there an extension that will do this, e.g. MyST?
2. Markdown itself doesn't support crossreferencing, and extensions to it are non-standard. Personally, I'd be ok with the Markdown flavor provided by MyST, if that can be supported well.

If those blockers can be addressed, then I think it comes down to a judgement call about whether it's better to use a clean and clear syntax (MyST Markdown :-) or the syntax that the rest of the community is already using (NumpyDoc). If the blockers can't be addressed, it seems to me that NumpyDoc is the clear winner.

philippjfr:
Thanks @MarcSkovMadsen for pushing on this. Adding docstrings everywhere is a really important goal. Before I get into various conventions I just want to state the current state of things:

Panel officially uses NumpyDoc and therefore reST syntax throughout. This is because it is the de-facto standard in the Python and particularly the PyData ecosystem. I also prefer it over Google's standard because it is more explicit about supported types. One might say types should be in the code but descriptions of the types in the docstring are always going to be more readable than the weird typing constructs you have to add to support complex unions of types using Python typing. HoloViews went a different route mainly because Jean-Luc objected to NumpyDoc for being too verbose.

We generally do not use any advanced reST features such as cross-linking or other advanced markup at present but those are important features that we should leverage and would be missing out on if we adopted vanilla Markdown. Therefore from my perspective we only have to options:

1. Stick with the reST+NumPyDoc style we use currently
2. Adopt a MyST flavored Markdown NumpyDoc style

While I agree VSCode rendering is very nice it is afaik the only editor that supports it and it seems a very odd choice on their part not to support the prevailing standard of reST based numpydoc. In particular even MyST does not yet officially support autodoc to parse Markdown in docstrings (see https://github.com/executablebooks/MyST-Parser/issues/228). My opinion therefore that we have no choice but to stick with the prevailing standard until there is a push from the community to standardize on some Markdown based numpydoc style.

philippjfr:
After some more investigation it also looks like VSCode **should** support some limited conversion of reST -> Markdown. In the pyright test suite you can see what should be supported: https://github.com/microsoft/pyright/blob/main/packages/pyright-internal/src/tests/docStringConversion.test.ts

philippjfr:
Actually it does look like there is now a replacement for autodoc that uses MyST: https://github.com/john-hen/demo-MyST-docstring/blob/main/docs/extensions/myst_docstring.py

It may be a little while until that is turned into a well supported Sphinx extension though.

MarcSkovMadsen:
I would understand if we say NumpyDoc and reST for now as the safe choice. Its important to have a decision such that docstrings can be contributed.

I will try to push microsoft pylance to support a bit more reST and then hope to get the support that way.

Would that be the way to go @philippjfr ?

philippjfr:
>Would that be the way to go @philippjfr ?

I think for that does seem like the safest choice. If we ever decide that the ecosystem has caught up I'm sure we could write a script to convert all the existing reST docstrings to Markdown.

MarcSkovMadsen:
I've changed to docstring to restructuredText. Would you review again @philippjfr ? Thanks.

PR #3198, opened 2022-02-15
Add thread safety and TTL to pn.state.as_cached

philippjfr:
Now that we have the ability to set `--num-threads` I expect many more users will leverage threads to speed up their application. However not all functionality in Panel was written with thread-safety in mind. This PR ensures that `pn.state.as_cached` maintains thread locks for each cache key ensuring that if multiple threads are trying to access the same cache key at the same time only one of the executes the function that is being cached on.

Additionally this PR also adds a TTL (time-to-live) for cache items allowing users to declare when a cached value expires and has to be refreshed.

### DIFF
--- a/panel/io/state.py
+++ b/panel/io/state.py
@@ -5,6 +5,7 @@ import datetime as dt
 import logging
 import json
 import threading
+import time
 
 from collections import OrderedDict, defaultdict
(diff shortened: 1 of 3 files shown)


MarcSkovMadsen:
Ttl is really important. Thanks

Issue #3107, opened 2022-01-17
DatetimePicker raises CalendarDate parameter 'start' only takes date types


MarcSkovMadsen:
Panel 0.12.6

## Issue

```python
import pandas as pd
import panel as pn

start_date = pd.Timestamp('2022-01-17 00:00:00')

pn.widgets.DatetimePicker(
    value=start_date,
    enable_seconds=False,
    name="datetime",
    start=start_date,
    end=start_date,
).servable()
```

```bash
$ python 'script.py'
Traceback (most recent call last):
  File "script.py", line 6, in <module>
    pn.widgets.DatetimePicker(
  File "C:\repos\trading_analytics\fumo-engine\.venv\lib\site-packages\panel\widgets\input.py", line 191, in __init__
    super().__init__(**params)
  File "C:\repos\trading_analytics\fumo-engine\.venv\lib\site-packages\panel\widgets\base.py", line 52, in __init__
    super().__init__(**params)
  File "C:\repos\trading_analytics\fumo-engine\.venv\lib\site-packages\panel\reactive.py", line 79, in __init__
    super().__init__(**params)
  File "C:\repos\trading_analytics\fumo-engine\.venv\lib\site-packages\panel\viewable.py", line 538, in __init__
    super().__init__(**params)
  File "C:\repos\trading_analytics\fumo-engine\.venv\lib\site-packages\panel\viewable.py", line 399, in __init__
    super().__init__(**params)
  File "C:\repos\trading_analytics\fumo-engine\.venv\lib\site-packages\panel\viewable.py", line 230, in __init__
    super().__init__(**params)
  File "C:\repos\trading_analytics\fumo-engine\.venv\lib\site-packages\param\parameterized.py", line 3146, in __init__
    self.param._setup_params(**params)
  File "C:\repos\trading_analytics\fumo-engine\.venv\lib\site-packages\param\parameterized.py", line 1363, in override_initialization
    fn(parameterized_instance, *args, **kw)
  File "C:\repos\trading_analytics\fumo-engine\.venv\lib\site-packages\param\parameterized.py", line 1616, in _setup_params
    setattr(self, name, val)
  File "C:\repos\trading_analytics\fumo-engine\.venv\lib\site-packages\param\parameterized.py", line 353, in _f
    return f(self, obj, val)
  File "C:\repos\trading_analytics\fumo-engine\.venv\lib\site-packages\param\__init__.py", line 621, in __set__
    super(Dynamic,self).__set__(obj,val)
[...]

PR #3202, opened 2022-02-16
Switch DatetimePicker start/end to param.Date

philippjfr:
A DatetimePicker should allow datetimes for the start and end values.


Hoxbro:
Just for reference see #3107 why it was not done in the first place.

PR #3206, opened 2022-02-20
fix: wrong offset when memoryview format is non-byte

maartenbreddels:
the offset should be in bytes, not elements. See also https://github.com/bokeh/ipywidgets_bokeh/issues/46#issuecomment-1046292704


jbednar:
Thanks! That looks good to me.

philippjfr:
Thanks @maartenbreddels, really appreciate it!

PR #3207, opened 2022-02-21
CI: add Python 3.10 in the matrix

maximlt:
At least an attempt to do so, now that param, hvplot, holoviews, datashader and colorcet have it in their test matrix.


maximlt:
@philippjfr I've added 3.10 in the text matrix. Except that I've had to avoid running the examples on 3.10, conda/mamba couldn't solve with the examples options. vtk/pyvista are already pinned (for some other reason) but looking at their GH repo and infos it seems that they don't yet fully support 3.10 anyway.

The tests are displayed as *cancelled* on a couple of mac-os runs, this is due to an error that occurs in the `test_subprocess` test in `test_terminal.py`. It's not related to this PR (I've seen this failure in previous CI runs weeks ago).

maximlt:
How about merging this one @philippjfr ? The `test_examples` task isn't executed on Python 3.10 due to VTK, but besides that the tests run fine on Python 3.10.

philippjfr:
Will re-attempt at a later date.

PR #3211, opened 2022-02-21
Make only one call to publish_display_data

maximlt:
`publish_display_data` is used to load come HTML/CSS/JS in a notebook when `pn.extension()` is called. The function it's part of is used by holoviews too (and then geoviews and hvplot). Unfortunately in JupyterLab each call to `publish_display_data` adds an empty line. This PR attempts to reduce the number of calls to `publish_display_data` from 3 to 1. @philippjfr I've merged all the JS into one string (without being quite sure if this is valid). If it's fine, before merging, I'd like to test it by running some notebooks with Jupyter Classic and Lab.

### DIFF
--- a/panel/io/notebook.py
+++ b/panel/io/notebook.py
@@ -252,17 +252,19 @@ def load_notebook(inline=True, load_timeout=5000):
         else:
             settings.resources.unset_value()
 
-    publish_display_data({
-        'application/javascript': bokeh_js,
-        LOAD_MIME: bokeh_js,
-    })
-    bokeh.io.notebook.curstate().output_notebook()
-
-    # Publish comm manager
+    # CSS and JS for the comm manager
     CSS = (PANEL_DIR / '_templates' / 'jupyter.css').read_text()
     JS = '\n'.join([PYVIZ_PROXY, _JupyterCommManager.js_manager, nb_mime_js])
-    publish_display_data(data={LOAD_MIME: JS, 'application/javascript': JS})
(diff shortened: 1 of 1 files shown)


panel/io/notebook.py
```diff
@@ -252,17 +252,19 @@ def load_notebook(inline=True, load_timeout=5000):
...
-    publish_display_data(data={LOAD_MIME: JS, 'application/javascript': JS})
-    publish_display_data(data={'text/html': f'<style>{CSS}</style>'})
+
+    # Call publish_display_data once to avoid displaying more than one empty
+    # line in JupyterLab.
+    merged_js = '\n'.join([bokeh_js, JS])
+    publish_display_data(data={
+        LOAD_MIME: merged_js,
+        'application/javascript': merged_js,
+        'text/html': f'<style>{CSS}</style>',
```
philippjfr, line 265: You can't do this sadly. The frontend only renders one of the MIME type that you send it, i.e. they're treated as alternative representations of the same thing and it's up to the frontend to render the "richest" representation.

maximlt: Ok thanks for the clarification, I see. I've separated it again but included the CSS in the second call to `publish_display_data` to make 2 calls instead of 3.

philippjfr: This still has the same issue, since some frontends will decide 'text/html' is the richest MIME type it can display that's the only thing it will actually render resulting in the JS not to be rendered at all.

philippjfr:
Closing, unfortunately I do not think there's a way to publish all the contents in one call.

PR #3212, opened 2022-02-22
Add pyodide support to ReactiveHTML

philippjfr:
Mostly an experiment for now but this PR explores what it would be like if we allowed users to write `ReactiveHTML` callbacks in Python and execute them using pyiodide.

```python
class ClickCounter(pn.reactive.ReactiveHTML):
    
    clicks = param.Integer()
    
    _template = """
    <div id='counter' onclick="${click}"><h1>Clicked 0 times</h1></div>
    """

    @pyscript
    def click(self):
        self.clicks += 1
        counter.innerHTML = f"<h1>Clicked {self.clicks} times</h1>"
```

pyiodide

### DIFF
--- a/panel/config.py
+++ b/panel/config.py
@@ -441,7 +441,8 @@ class panel_extension(_pyviz_extension):
         'tabulator': 'panel.models.tabulator',
         'gridstack': 'panel.layout.gridstack',
         'texteditor': 'panel.models.quill',
-        'jsoneditor': 'panel.models.json_editor'
+        'jsoneditor': 'panel.models.json_editor',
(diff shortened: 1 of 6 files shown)


MarcSkovMadsen:
Its a quite nice experiment. I have experimented with it here https://github.com/MarcSkovMadsen/panel-sketch.

Some examples of really using this end to end by 1) Extracting some data 2) Transforming it using pandas  etc 2) outputting it to some plot would be really nice. Also seeing how the data.model can be updated directly on the js side would be nice.

philippjfr:
Interesting experiment but also insane so closing.

PR #3222, opened 2022-03-03
Add LinearGauge indicator

philippjfr:
A `LinearGauge` provides a much more compact value indicator than `Dial` or `Gauge`. 

Screen Shot 2022-04-13 at 15 53 15
Screen Shot 2022-04-13 at 15 53 00

- [ ] Add tests
- [x] Make `colors` handling consistent with `Gauge` and `Dial`
- [x] Improve `horizontal` support

### DIFF
--- a/examples/reference/indicators/LinearGauge.ipynb
+++ b/examples/reference/indicators/LinearGauge.ipynb
@@ -15,24 +15,22 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "The ``LinearGauge`` is a value indicator providing a visual representation of a value in a certain range as a simple linear gauge.\n",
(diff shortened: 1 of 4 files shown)


jbednar:
That's useful, but I'd think that in a bank of them the current value of it won't stand out much, while I'd want higher values to be very visible in many cases.  The usual way that's done is to dim the alpha of the background for the areas above the current value, potentially to zero. 

In general since there is really only one value being displayed, one will often want this to be very small, so it would be good if the defaults weren't assuming this needs font and tick sizes like a curve or scatterplot...

Also, the colors here seem backward if that's a failure rate. :-)

philippjfr:
Agree with everything you said. Total work in progress for now.

PR #3224, opened 2022-03-04
add azure oauth v2.0 endpoint support

j-r77:
Added a (very simple) additional authentication class that uses Azure AD's v2.0 oauth endpoint. This one uses the standard openid, email, and profile scopes instead of Azure-specific ones.

In particular, I couldn't get the default azuread auth to work with accounts that require MFA.

Azure's own docs suggest using V2.0 for all new applications:
https://docs.microsoft.com/en-us/azure/active-directory/azuread-dev/v1-overview


philippjfr:
Love it, thanks for contributing @j-r77!

Issue #3225, opened 2022-03-06
RuntimeError: Patched curdoc has been previously destroyed


MarcSkovMadsen:
Panel=0.12.6

I've started seeing errors like the below for one of my Panel apps. What the user sees is error 500. Normally the Panel/ Bokeh server is robust to exceptions. But not here.

The problem is potentially related to getting files for an azure blob storage using an azure sdk.

I will be investigating, but record it here so that it can be helpful for others experiencing something similar. So far I cannot create a reproducible example.

Any hints at potential causes or solutions would be nice. Thanks.

```bash
2022-03-04 09:04:02,108;INFO;Request URL: '[https://<REMOVED BY ME>.blob.core.windows.net/runs?restype=REDACTED&comp=REDACTED&marker=REDACTED&include=REDACTED'/nRequest](https://<REMOVED BY ME>.blob.core.windows.net/runs?restype=REDACTED&comp=REDACTED&marker=REDACTED&include=REDACTED%27/nRequest) method: 'GET'/nRequest headers:/n    'x-ms-version': 'REDACTED'/n    'Accept': 'application/xml'/n    'User-Agent': 'azsdk-python-storage-blob/12.9.0 Python/3.7.2 (Linux-5.4.0-1070-azure-x86_64-with-debian-9.8)'/n    'x-ms-date': 'REDACTED'/n    'x-ms-client-request-id': '0911dc44-9b9a-11ec-89ae-028d771633e0'/n    'Authorization': 'REDACTED'/nNo body was attached to the request
2022-03-04 09:04:02,338 Failed to run session creation hooks RuntimeError('Patched curdoc has been previously destroyed')

Traceback (most recent call last):

  File "/usr/local/lib/python3.7/site-packages/bokeh/server/contexts.py", line 239, in create_session_if_needed

    await self._application.on_session_created(session_context)

  File "/usr/local/lib/python3.7/site-packages/panel/io/server.py", line 169, in on_session_created

    cb(session_context)

  File "/usr/local/lib/python3.7/site-packages/panel/io/server.py", line 118, in _initialize_session_info

    if config.session_history == 0 or session_id in sessions:

  File "/usr/local/lib/python3.7/site-packages/panel/config.py", line 237, in __getattribute__

    params = super().__getattribute__('param')

  File "/usr/local/lib/python3.7/site-packages/param/parameterized.py", line 2623, in param

    return Parameters(self.__class__, self=self)

[...]

PR #3228, opened 2022-03-08
Guard pn.state.curdoc access

philippjfr:
When a document has been destroyed the session_context is not longer accessible but it may still be accessed.

@MarcSkovMadsen Could you test?

Fixes https://github.com/holoviz/panel/issues/3225

### DIFF
--- a/panel/io/state.py
+++ b/panel/io/state.py
@@ -469,8 +469,11 @@ class _state(param.Parameterized):
             doc = _curdoc()
         except Exception:
             return None
-        if doc.session_context:
-            return doc
+        try:
+            if doc.session_context:
+                return doc
+        except Exception:
+            return None
 
     @curdoc.setter
     def curdoc(self, doc):


ncvangilse:
I'm on panel 0.13 and I am still experiencing #3225 
However I changed [curdoc() in bokeh/io/doc.py](https://github.com/bokeh/bokeh/blob/branch-3.0/bokeh/io/doc.py#:~:text=if%20doc%20is,return%20doc) so it doesn't raise the RuntimeError if doc is None.
My replacement code is then:
```
        if doc is not None:
            return doc
```
But perhaps the `PATCHED_CURDOC[-1]` should be deleted as well?

Now my panel apps have been running for 24 hrs and counting...

philippjfr:
0.13 does not yet include this PR

maximlt:
Hi @ncvangilse, could you open a new issue please?

PR #3252, opened 2022-03-18
Remove all tornado imports

philippjfr:
This PR makes it so that in a pyiodide context the `panel.io.server` module and all tornado based imports are not run. Additionally this sets the default comm to be `'ipywidgets'` so that all Panel objects are rendered using the `jupyter_bokeh.BokehModel` ipywidget wrapper that also works in jupyterlite.


MarcSkovMadsen:
Jupyterlite is one client environment to run in. Another one would be pyodide directly. Image creating interactive apps using Panel and the rest of the ecosystem without have to run a server in the background. That would be amazing.

PR #3268, opened 2022-03-27
Various updates to the docs and its build

maximlt:
The main goal of this PR is to fix the docs build that is currently broken and prevented the site update after the release of 0.12.7. This may take a few iterations.


maximlt:
I actually need to merge this PR first to get the UI from Github to manually trigger the docs workflow.

PR #3270, opened 2022-03-28
Added fullpath to jupyter_server_extension

Hoxbro:
I noticed I couldn't get the Panel widget in jupyter lab to work in a non-base conda environment.
Screenshot 2022-03-28_13 25 03

With the output in my terminal:
Screenshot 2022-03-28_13 25 14

The error is because `~` is present in the path and when `os.path.abspath` is used it gives a wrong result.
``` python
path = "~/hello_there.mp4"
a_path = os.path.abspath(path)
ae_path = os.path.abspath(os.path.expanduser(path))
path, a_path, ae_path
```
image

The `abspath` command is called from [bokeh](https://github.com/bokeh/bokeh/blob/560a57e166a1f54319df57127502d48ee4ecc72e/bokeh/command/util.py#L117) and a fix should be implemented there too.

### DIFF
--- a/doc/generate_modules.py
+++ b/doc/generate_modules.py
@@ -160,7 +160,7 @@ def recurse_tree(path, excludes, opts):
     """
     # use absolute path for root, as relative paths like '../../foo' cause
     # 'if "/." in root ...' to filter out *all* modules otherwise
-    path = os.path.abspath(path)
+    path = os.path.abspath(os.path.expanduser(path))
     # check if the base directory is a package and get is name
     if INIT in os.listdir(path):
         package_name = path.split(os.path.sep)[-1]
--- a/panel/command/serve.py
+++ b/panel/command/serve.py
@@ -23,6 +23,7 @@ from tornado.web import StaticFileHandler
 
 from ..auth import OAuthProvider
 from ..config import config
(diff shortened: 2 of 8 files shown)


philippjfr:
This is great thank you! I notice you added a utility but aren't using it consistently yet.

Hoxbro:
I only have it in `panel/io/jupyter_server_extension.py` to avoid long lines. 

Then I just searched for `abspath` and updated it with `expanduser` because it was the easiest thing to do. 

If you want, I can move the functions into a `utils` file and call it consistently.

philippjfr:
>If you want, I can move the functions into a utils file and call it consistently.

Seems sensible.

doc/generate_modules.py
```diff
@@ -160,7 +160,7 @@ def recurse_tree(path, excludes, opts):
     """
     # use absolute path for root, as relative paths like '../../foo' cause
     # 'if "/." in root ...' to filter out *all* modules otherwise
-    path = os.path.abspath(path)
+    path = os.path.abspath(os.path.expanduser(path))
```
Hoxbro, line 163: This is outside the panel module so I chose to keep it.

PR #3283, opened 2022-03-30
Do not call an Action with an argument

pepijndevos:
I'm not sure if this is a documentation error or a Panel error, but Param basically says

> `param.Action`: A callable with no arguments, ready to invoke
https://param.holoviz.org/user_guide/Parameter_Types.html#invocations

And Panel clearly invokes the Action with an argument here.

### DIFF
--- a/panel/param.py
+++ b/panel/param.py
@@ -471,7 +471,7 @@ class Param(PaneBase):
             watcher = widget.param.watch(event, 'clicks')
         elif isinstance(p_obj, param.Action):
             def action(change):
-                value(self.object)
+                value()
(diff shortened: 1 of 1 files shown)


philippjfr:
Thanks for the PR and thanks for catching this. However this change would break backward compatibility in a pretty bad way and indeed passing in the object that is doing the invoking was for a long time the only way you could actually achieve certain forms of interactivity. Now that we have `param.Event` this is not a huge issue anymore but I would rather update param documentation than make this breaking change in Panel.

philippjfr:
Closing. We should figure out what docs need updating to point out the inconsistency here.

PR #3288, opened 2022-03-31
Scroll to center selected Tabulator row

douglas-raillard-arm:
When updating row selection, scroll in the widget so that the selected
row is centered. This fixes the current behavior that leaves the
selected row just passed the end of the viewed range.

The original intended behavior was to place the selected row at the top
of the viewed range, but due to a Tabulator bug, we needed to request
"bottom" for that. Now that Tabulator seems to be fixed, use "center" so
that hopefully this issue will not come back.

Signed-off-by: Douglas Raillard <douglas.raillard@arm.com>


philippjfr:
Aiming for an 0.13.0 release candidate tomorrow.

PR #3326, opened 2022-04-06
Fixes cross selection disabled

### LINKED ISSUE #3325: CrossSelector does not respect disabled=True
panel = "0.13.0a46"

When setting `disabled=True`, the CrossSelector widget remains editable.

The expected behavior should be readonly.

```
import panel as pn

pn.extension()

cross_selector = pn.widgets.CrossSelector(
    name="Fruits",
    value=["Apple", "Pear"],
    options=["Apple", "Banana", "Pear", "Strawberry"],
    disabled=True,
)


pn.Row(cross_selector).show()


```

Hoxbro:
Fixes #3325 

This is done by calling self._update_disabled at the end of `__init__`

image

### DIFF
--- a/panel/widgets/select.py
+++ b/panel/widgets/select.py
@@ -636,6 +636,8 @@ class CrossSelector(CompositeWidget, MultiSelect):
         self._selections = {False: [], True: []}
         self._query = {False: '', True: ''}
 
+        self._update_disabled()
+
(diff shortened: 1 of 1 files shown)


philippjfr:
Ah got it, need to be created before the constructor. This is fine, will merge.

PR #3348, opened 2022-04-09
Handle empty vega selections

joelostblom:
This handles both `None` and an empty dict which is what is returned after selecting and then unselecting


philippjfr:
Thanks!

PR #3349, opened 2022-04-09
Show how to filter Altair charts with a panel selection

joelostblom:
I have included an example of how we can filter an Altair chart via a selection in another chart as well as documented the advantages of doing so.

I would like to also include an example of how to filter via point selections, but I can't figure out if this is possible in Panel currently as per https://discourse.holoviz.org/t/panel-0-13-0-release-candidate/3530/19


philippjfr:
This is great, thank you!

PR #3355, opened 2022-04-10
Auto-pagination allowed to be set to remote and to switch

maximlt:
Follow-up on https://github.com/holoviz/panel/pull/3306 which always set pagination to `local` if the number of rows in the df was greater than 200. Another change made in this PR is to avoid triggering the `_set_explicict_pagination` callback from `_apply_max_size`, allowing `pagination` to automatically switch from 'local' to 'remote' or inversely when the df size changes.


philippjfr:
Thanks for catching that!

PR #3359, opened 2022-04-10
Add docstrings to indicators

MarcSkovMadsen:
The docstrings make it so much easier to use the widgets in an IDE like VS Code.

image


MarcSkovMadsen:
Thanks for wrapping the docstrings @philippjfr . Could you enlighten me why you do it? (I would like to know such that you don't have to do it).

I normally work with a max -line-length of 100 and I can see Panels flake8 rule is 165

image

PR #3367, opened 2022-04-11
Adds the remaining widget docstrings

MarcSkovMadsen:
Adds the remaining widget docstrings and also docstrings to the `panel` and `widgets` modules.

The purpose is to make it much easier to navigate the code in a modern IDE like VS Code or PyCharm


MarcSkovMadsen:
Why do you add the `noqa` below @philippjfr? 

I have a feeling it is what hinders tooltips for `pn.widgets` to show up in VS Code??

image

Issue #3382, opened 2022-04-14
Tabulator: style not applied after a cell is edited


maximlt:
Bug observed on *master*. The following snippet creates a Tabulator widget whose empty cells should be styled with a red background. Adding content to an already empty cell should I guess reset the style and emptying the content of a cell should apply the style.

```python
import panel as pn
import pandas as pd

pn.extension('tabulator')

df = pd.DataFrame({'a': ['', 'y', '']})

w = pn.widgets.Tabulator(df)

def highlight_empty(s):
    is_empty = s == ''
    return ['background-color: red' if v else '' for v in is_empty]

w.style.apply(highlight_empty, subset='a')

w.servable()
```

![issue_tabulator_style](https://user-images.githubusercontent.com/35924738/163358407-9faa0a28-b895-4a30-b7f0-235996b3edfd.gif)

PR #3385, opened 2022-04-14
Update the style after an edit event

maximlt:
Fixes https://github.com/holoviz/panel/issues/3382

### DIFF
--- a/panel/widgets/tables.py
+++ b/panel/widgets/tables.py
@@ -1054,6 +1054,7 @@ class Tabulator(BaseTable):
                 event.old = self._old[event.column].iloc[event.row]
             for cb in self._on_edit_callbacks:
                 cb(event)
+            self._update_style()
         else:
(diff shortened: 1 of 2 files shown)


panel/widgets/tables.py
```diff
@@ -1054,6 +1054,7 @@ def _process_event(self, event):
                 event.old = self._old[event.column].iloc[event.row]
             for cb in self._on_edit_callbacks:
                 cb(event)
+            self._update_style()
```
philippjfr, line 1057: Surprised this is needed. An edit event triggers two separate events 1) the `table-edit` event and 2) a property change event on the data source. I would expect the styles to already be updated by the latter.

maximlt: Commenting this doesn't resolve the issue so it appears it's needed. The data source update doesn't trigger any restyling apparently.

PR #3389, opened 2022-04-15
Add downstream tests

maximlt:
This workflow runs when a new Panel release is made and launches and reports the results of the tests workflows of HoloViews and Lumen, running from their *master* branch.


maximlt:
Merging since this PR doesn't touch any code.

PR #3401, opened 2022-04-15
Show how to filter catgorical and temporal data from Altair/VegaLite

joelostblom:
This took me a little bit of time figure out, so I thought documenting it might help others trying to do the same thing. I also added headings since this section is quite long now.


philippjfr:
Thanks @joelostblom!

Issue #3400, opened 2022-04-15
Tab expansion working poorly in VS Code


MarcSkovMadsen:
I'm on Panel 0.13.0. If I write  from `panel.pane import ` and press `CTRL+Space` or start typing `Markdown` I would expect VS Code to help me identify what is possible to import. It does not.

https://user-images.githubusercontent.com/42288570/163605291-3ff10210-1955-48bc-a7dc-82cde2e5433b.mp4

This is just an example of a general problem that makes it harder than it has to be to use Panel.

MarcSkovMadsen:
Its the same with other combinations `pn.pane.` and `from panel import pane`. I simply cannot get any help that `Markdown` is available under `panel.pane`.

philippjfr:
What does VSCode use for completion here? Guess it's not dynamic.

MarcSkovMadsen:
VS Code uses the Python Language Server and static type analysis (similar to mypy).

I can see if I

1. Move the `from .layout import ...` to the top of the root `__init__.py` file
2. Restart VS Code
![image](https://user-images.githubusercontent.com/42288570/163608308-64ddd915-3c7d-444a-96ad-71966d735319.png)

the I start to be able to see some classes.

https://user-images.githubusercontent.com/42288570/163608340-51abecc3-c6da-4a21-ae8f-98554323d40f.mp4

MarcSkovMadsen:
I've tried to set the Log Level for pylance to `trace`

![image](https://user-images.githubusercontent.com/42288570/163608453-3de15fb3-b72e-4063-8a58-56a2fb59d2d2.png)

and inspect the logs. But don't see any errors. See the pylance logs below.

![image](https://user-images.githubusercontent.com/42288570/163608583-d5a49d1f-3991-46c8-a2b0-a7259d750c59.png)

[log.txt](https://github.com/holoviz/panel/files/8497503/log.txt)

MarcSkovMadsen:
It might have something todo with static analysis for panel not working at all as you can see below.

![image](https://user-images.githubusercontent.com/42288570/163608767-f67b8066-f73f-4045-96c2-c0dce0712cf8.png)

MarcSkovMadsen:
But we are not alone https://github.com/microsoft/pylance-release/issues/844.

But for example Streamlit just works.

MarcSkovMadsen:
Similarly if I run

[...]

PR #3403, opened 2022-04-16
#3400 Fix VS Code/ Pylance not working

MarcSkovMadsen:
Addresses #3400 

It removes the `from . import param` in the panel `__init__.py` root file.

I've tried to search the Panel code including example notebooks and tests. I cannot see any reason to import it.

I don't know why this solves the issues. But the `panel.param.py` file imports a lot of other modules and also the `param` package. So a lot of things could potentially be confusing Pylance and causing it.

### DIFF
--- a/panel/__init__.py
+++ b/panel/__init__.py
@@ -47,7 +47,6 @@ https://panel.holoviz.org/getting_started/index.html
 from . import layout # noqa
 from . import links # noqa
 from . import pane # noqa
-from . import param # noqa
 from . import pipeline # noqa
 from . import reactive # noqa
 from . import viewable # noqa


MarcSkovMadsen:
Hmmm. I tried to `pip install -e <path/to/panel` this change. It did not seem to help VS Code and Pylance ???

philippjfr:
Appreciate all your investigation of this issue, I'll dig into this as well. I don't think not exposing certain modules because pylance doesn't like them isn't the solution however. Let's find the actual issue in the `.param` module.

MarcSkovMadsen:
I close this one as its been replaced with https://github.com/holoviz/panel/pull/3411

PR #3411, opened 2022-04-16
add __all__ to prepare for static type checkers and IDEs

MarcSkovMadsen:
This adds a `__all__` to `__init__.py` files as required for static type checking and intellisense in VS code. C.f. https://github.com/microsoft/pylance-release/issues/2572#issuecomment-1100566151


panel/__init__.py
```diff
@@ -65,3 +65,29 @@
 from .param import Param # noqa
 from .template import Template # noqa
 from .widgets import indicators # noqa
+
+__all__ = [
```
MarcSkovMadsen, line 69: Using `__all__`.

panel/widgets/__init__.py
```diff
@@ -32,56 +32,86 @@
...
-    BooleanStatus,
-    Dial,
-    Gauge,
-    LinearGauge,
-    LoadingSpinner,
-    Number,
-    Progress,
-    Trend,
-    Tqdm,
+from .ace import Ace as Ace  # noqa
```
MarcSkovMadsen, line 35: *redundant form*. An alternative to `__all__`. 

I don't know which to prefer. Which would you prefer @philippjfr ?

MarcSkovMadsen:
Both the `__all__` and the *redundant* form seems to work.

https://user-images.githubusercontent.com/42288570/163689164-6ca54e92-6036-4371-99c4-4548b88324c2.mp4

panel/widgets/__init__.py
MarcSkovMadsen: Personally I would prefer __all__ because the new developers would ask why add `... as XYZ`, it seems redundant?

But `__all__` might be easier to forget to update?

philippjfr: Definitely `__all__`. Is this redundant form really something someone uses? It's crazy!

panel/__init__.py
```diff
@@ -65,3 +65,38 @@
...
+    "GridSpec",
+    "indicators",
+    "interact",
+    "ipywidget",
+    "layout",
+    "links",
+    "pane",
+    "Pane",
+    "panel",
+    # "param",
```
MarcSkovMadsen, line 89: I don't think `pn.param` is to be exposed?

panel/__init__.py
```diff
@@ -65,3 +65,38 @@
...
+    "config",
+    "depends",
+    "extension",
+    "FlexBox",
+    "GridBox",
+    "GridSpec",
+    "indicators",
+    "interact",
+    "ipywidget",
+    "layout",
```
MarcSkovMadsen, line 84: I think modules need to be exposed? I can see Pydatantic does it https://github.com/samuelcolvin/pydantic/blob/8997cc5961139dd2695761a33c06a66adbf1430a/pydantic/__init__.py#L27

image


But I could not find anything written about whether its needed.

panel/config.py
```diff
@@ -650,3 +650,5 @@ def _cleanup_server(server_id):
 panel_extension.add_delete_action(_cleanup_panel)
 if hasattr(panel_extension, 'add_server_delete_action'):
     panel_extension.add_server_delete_action(_cleanup_server)
+
+__all__ = []
```
MarcSkovMadsen, line 654: I don't think there is anything you want to publicly expose? So I set `__all__` equal to the empty list. I also do that in other places where I don't think there is anything you want to make a part of the public api to users of the panel package.

panel/links.py
```diff
@@ -655,3 +655,11 @@ def _get_code(self, link, source, src_spec, target, tgt_spec):
 Link.register_callback(callback=JSLinkCallbackGenerator)
 
 Viewable._preprocessing_hooks.append(Callback._process_callbacks)
+
+__all__ = [
```
MarcSkovMadsen, line 659: I'm in doubt here what you want to expose as part of public api to users of the package as I've very rarely used this functionality/ module.

Hoxbro:
I'm +1 on using `__all__`. 

I would properly add `__all__` directly below the imports, so it is clearly visible and not something you have to find at the end of the script.

MarcSkovMadsen:
Releasing this PR together with the various docs PRs will be a major productivity boost for users of VS Code and similar IDEs :-)

panel/__init__.py
jbednar: Seems like users will sometimes need to instantiate the Param pane, ParamMethod, and ParamFunction, so I don't see why not...

MarcSkovMadsen: If those funktions are considered public they should be documented.

I have never read about them and would not know what they do or when to use?

philippjfr: You use them **all the time**. They are what generates widgets from Parameters and what makes reactive functions (i.e. those wrapped with depends and bind) interactive.

panel/links.py
philippjfr: Yeah this really is mostly used internally and should probably be moved to a core module.

philippjfr:
Thanks @MarcSkovMadsen!

MarcSkovMadsen:
Thanks @philippjfr and others for supporting.

PR #3433, opened 2022-04-20
Add setuptools to the runtime dependencies

maximlt:
Panel depends on `pkg_resources` which is a module distributed by `setuptools`. I've seen this was missing in the runtime dependencies when playing around with toga/briefcase. Note that `pkg_resources` is deprecated and should be replaced at some point by `importlib.resources` and `importlib.metadat`, and their backports.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -104,7 +104,8 @@ install_requires = [
     'requests',
     'tqdm >=4.48.0',
     'pyct >=0.4.4',
-    'bleach'
+    'bleach',
+    'setuptools',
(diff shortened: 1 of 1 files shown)


philippjfr:
Damn, does that mean 0.13.0 installs are currently broken or does something else usually pull in setuptools?

maximlt:
I've never heard of anyone reporting an issue due to the missing `setuptools` requirement. I think in a pip/conda context it's super likely it gets installed.

Issue #3429, opened 2022-04-18
Markdown pane does not render latex without adding "mathjax" to pn.extension


MarcSkovMadsen:
I'm on Panel 0.13 trying out the new functionality. I cannot get it working.

```python
import panel as pn

pn.extension(sizing_mode="stretch_width")

pn.pane.Markdown("""
- Markdown panes now support $$LaTeX$$
- To render a snippet with LaTeX add two $ delimiters
- e.g. $$\sum_{j}{\sum_{i}{a*w_{j, i}}}$$
""", width=800).servable()
```

looks like

![image](https://user-images.githubusercontent.com/42288570/163867321-88e7004d-febc-441f-87ed-788f09470f25.png)

## Additional Context

If I add `mathjax` to `pn.extension`, I can get it working. But its not documented in the blog post https://blog.holoviz.org/panel_0.13.0.html or in the reference notebook https://panel.holoviz.org/reference/panes/Markdown.html

```python
import panel as pn

pn.extension("mathjax", sizing_mode="stretch_width")

pn.pane.Markdown("""
## Markdown $$LaTeX$$ support

The Markdown pane supports math rendering for string encapsulated with double $ delimiters: $$\sum_{j}{\sum_{i}{a*w_{j, i}}}$$
""", width=800).servable()
```

![image](https://user-images.githubusercontent.com/42288570/163867497-038185be-4f94-4421-8ecb-80a18794970e.png)

PR #3438, opened 2022-04-21
Ensure bokeh correctly detects whether HTML/Markdown contains latex

philippjfr:
Fixes https://github.com/holoviz/panel/issues/3429

### DIFF
--- a/panel/config.py
+++ b/panel/config.py
@@ -475,6 +475,7 @@ class panel_extension(_pyviz_extension):
                 except Exception:
                     pass
                 __import__(self._imports[arg])
+                self._loaded_extensions.append(arg)
             elif arg in reactive_exts:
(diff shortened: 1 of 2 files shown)


philippjfr:
@MarcSkovMadsen So the fast templates have this bit of CSS in them:

```python
  :not(:defined) {
    visibility: hidden;
  }
```

This seems to hide the `mjx-container` tags that MathJax adds to the page to render the LaTeX output. Do you recall why this rule had to be added?

MarcSkovMadsen:
Yes. The fast documentation specifiks they should be added to avoid having the page look write indtil web components have been defined.

philippjfr:
So it should be removed after the page is fully rendered?

philippjfr:
Will merge and fix the fast template issue separately.

PR #3443, opened 2022-04-22
DOC: vscode -> vscode notebook

raybellwaves:
Is this correct?

When I think vscode I think of a text editor but I believe this part is referring to rendering elements after cells in notebooks in vscode (https://code.visualstudio.com/blogs/2021/11/08/custom-notebooks)?


philippjfr:
Thanks @raybellwaves!

PR #3444, opened 2022-04-22
Improvements for pyodide handling

philippjfr:
- Add `state._is_pyodide` property to unify detection of pyodide context
- Suppress warnings

### DIFF
--- a/panel/io/__init__.py
+++ b/panel/io/__init__.py
@@ -17,7 +17,7 @@ from .notebook import ( # noqa
     load_notebook, push, push_notebook
 )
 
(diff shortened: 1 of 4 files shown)


MarcSkovMadsen:
Would it not be nice for users to have public access to the property?

Then they can import in one way if in pyodide and another if not etc.

philippjfr:
>Would it not be nice for users to have public access to the property?

On the one hand yes, but it's really not Panel's responsibility to handle this and in the long run pyodide may not be the only WASM runtime that is supported. So I'd prefer to keep this as private API for the time being.

Issue #3428, opened 2022-04-18
JSONEditor icons not showing in Fast template


MarcSkovMadsen:
I'm on Panel 0.13.0 trying out the new `JSONEditor`. It does not look well in the FastListTemplate.

![image](https://user-images.githubusercontent.com/42288570/163866321-f1490c3e-ae0a-4e6d-9b19-f05ba8adb170.png)

```python
import panel as pn

pn.extension('ace', 'jsoneditor', sizing_mode="stretch_width", template="fast")

editor = pn.widgets.JSONEditor(value={
    "icon_url" : "https://assets.chucknorris.host/img/avatar/chuck-norris.png",
    "id" : "aybvxdnmr82jc_jssvkegw",
    "url" : "",
    "value" : "When Chuck Norris throws exceptions, it's across the room."
})

pn.Row(
    editor,
    editor.controls(['menu', 'mode', 'selection', 'value']),
).servable()

pn.state.template.param.update(
    site="Awesome Panel", title="Panel 0.13 adds JSONEditor"
)
```

philippjfr:
Thanks for reporting, two issues:

1) The icons svg file isn't bundled into Panel
2) The fast-bokeh.css file styling rule for buttons is way too broad

PR #3447, opened 2022-04-23
Improve resources bundling for JSONEditor

philippjfr:
Fixes https://github.com/holoviz/panel/issues/3428


philippjfr:
@MarcSkovMadsen Could you check if the updated CSS rules for bokeh buttons are sufficiently broad? Styling all buttons is definitely not a solution that will work but I'm not sure styling `bk-btn` is sufficient either.

PR #3451, opened 2022-04-25
DOC: mention --autoreload

raybellwaves:
Some discussion of the PR at https://discourse.holoviz.org/t/dashboard-development-in-a-py-file/3616

I've put the `panel serve` command in a code block. I'm also adding the `--autoreload` flag which reflects the default of other dashboard tools e.g. streamlit. The goal is make it easier for new users.


philippjfr:
Will merge and perhaps add a bit more detail separately.

PR #3458, opened 2022-04-28
Fix Fast button color

### LINKED ISSUE #3457: Button in FastTemplate does not have correct color
Just tried out panel==0.13.1a2

I can see that for some reason the button color in a FastTemplate is no longer correct. It used to work.

image

It should have been white. Its a regression.

MarcSkovMadsen:
Fixes #3457


philippjfr:
Thanks!

Issue #3455, opened 2022-04-27
Fast template : 404 error fastbasetemplate/images/header-columns-over-bg.gif


xavArtley:
When using a pandas Dataframe widget in a fast template I receive this error?
![image](https://user-images.githubusercontent.com/18531147/165506058-ec31a0e5-a989-4992-98bb-38e1f6505150.png)
I suppose the gif is missing in resources

philippjfr:
Not quite sure why that would load a gif.

xavArtley:
I don't know but it seeams to be here :
https://github.com/holoviz/panel/blob/4429e319e1aa5145ee0baa1d109d58d0c981e626/panel/template/fast/css/fast_bokeh_slickgrid.css#L74

MarcSkovMadsen:
The original slickgrid also loads images.

philippjfr:
I see, problem here is that the url path is relative to where the CSS is served from and that original CSS referenced an image served from within slickgrid itself.

PR #3461, opened 2022-04-28
Replace slickgrid background image in custom FastTemplate CSS

philippjfr:
Replaces an image we can not easily serve with a simple brightness filter. Similar to what this CSS stylesheet does for slickgrid: https://github.com/DimitarChristoff/slickgrid-es6/blob/master/src/slick-default-theme.scss#L26

Fixes https://github.com/holoviz/panel/issues/3455

### DIFF
--- a/panel/template/fast/css/fast_bokeh_slickgrid.css
+++ b/panel/template/fast/css/fast_bokeh_slickgrid.css
@@ -69,9 +69,12 @@
   font-weight: 500px;
 }
 
-.bk-root .slick-header-column:hover,
+.bk-root .slick-header-column:hover {
+  filter: brightness(0.98);
+}
+
 .bk-root .slick-header-column-active {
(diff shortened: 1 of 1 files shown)


panel/template/fast/css/fast_bokeh_slickgrid.css
```diff
@@ -69,9 +69,12 @@
...
 }
 
-.bk-root .slick-header-column:hover,
+.bk-root .slick-header-column:hover {
+  filter: brightness(0.98);
+}
+
 .bk-root .slick-header-column-active {
-  background: transparent url('images/header-columns-over-bg.gif') repeat-x center bottom;
+  filter: brightness(0.9);
```
philippjfr, line 77:
```suggestion
  filter: brightness(0.95);
```

Issue #3440, opened 2022-04-22
Make debugging of Jupyter Panel Preview easier


MarcSkovMadsen:
I've experienced it many times and also seen others struggle with the Jupyter Panel Preview.

If the jupyter panel preview just displays a blank screen you have no way of finding the cause as you cannot find the server logs any where.

Here is an example of a user struggling https://discourse.holoviz.org/t/how-to-customize-the-display-url-from-panel-serve-for-use-behind-jupyterhub-with-jupyter-server-proxy/3571/5?u=marc. I'm trying to help but have no chance to find the root cause.

I don't know what the solution is but

- If the log exists somewhere then document how to find it.
- If no log exist then create one by writing to a file. Document where to find that file.
- If no log exist then make a general feature request to jupyter and maybe help them implement

tlvu:
I launched the docker image in https://discourse.holoviz.org/t/how-to-customize-the-display-url-from-panel-serve-for-use-behind-jupyterhub-with-jupyter-server-proxy/3571/ directly on my machine and I got this error when trying to launch the Panel preview:

```
ServerApp - ERROR - Uncaught exception GET /panel-preview/render/workdir/PAVICS-e2e-workflow-tests/binder/test-notebooks/TEST-panel-servable.ipynb (10.10.10.7)
HTTPServerRequest(protocol='http', host='lvu.ouranos.ca:8890', method='GET', uri='/panel-preview/render/workdir/PAVICS-e2e-workflow-tests/binder/test-notebooks/TEST-panel-servable.ipynb', version='HTTP/1.1', remote_ip='10.10.10.7')
Traceback (most recent call last):
  File "/opt/conda/envs/birdy/lib/python3.8/site-packages/tornado/web.py", line 1704, in _execute
    result = await result
  File "/opt/conda/envs/birdy/lib/python3.8/site-packages/panel/io/jupyter_server_extension.py", line 116, in get
    page = server_html_page_for_session(
  File "/opt/conda/envs/birdy/lib/python3.8/site-packages/bokeh/embed/server.py", line 255, in server_html_page_for_session
    bundle = bundle_for_objs_and_resources(None, resources)
  File "/opt/conda/envs/birdy/lib/python3.8/site-packages/bokeh/embed/bundle.py", line 211, in bundle_for_objs_and_resources
    js_files.extend(js_resources.js_files)
  File "/opt/conda/envs/birdy/lib/python3.8/site-packages/panel/io/resources.py", line 299, in js_files
[...]

PR #3469, opened 2022-04-28
fix root_dir support for "Render with Panel" green button 

tlvu:
Regression introduced when calling `fullpath(path)` when `path` is an URL fragment.  I think `fullpath()` was meant for path on disk only.

See discussion in https://github.com/holoviz/panel/issues/3440#issuecomment-1112616866

### DIFF
--- a/panel/io/jupyter_server_extension.py
+++ b/panel/io/jupyter_server_extension.py
@@ -94,9 +94,7 @@ class PanelHandler(DocHandler):
         pass
 
     async def get(self, path, *args, **kwargs):
-        origpath = path
-        path = os.path.join(self.application.root_dir, fullpath(path))
-        raise ValueError(f"{path} {origpath} {self.application.root_dir} {fullpath(origpath)}")
+        path = os.path.join(self.application.root_dir, path)
         if path in _APPS:
             app, context = _APPS[path]
         else:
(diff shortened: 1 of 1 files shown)


tlvu:
This is really odd, my fix seems to make the JupyterLab server restart each time I save the notebook to test the "Render on Save" checkbox.

This is my first PR.  I do not really understand how this plugin is interacting with JupyterLab.  Please help.

philippjfr:
Thanks @tlvu. I'll look at this later today, want this in 0.13.1 which I hope to release today or Monday.

philippjfr:
So I don't see the issue you're seeing at all. This works perfectly for me, and I can confirm that adding `fullpath` caused issues when setting the `--notebook-url`.

philippjfr:
So I don't see the mechanism by which this could force a restart either. The server extension is simply installed as an additional handler on the tornado server. If it's somehow killing or resetting jupyterlab then there's something else going on causing it to crash.

tlvu:
>  If it's somehow killing or resetting jupyterlab then there's something else going on causing it to crash.

@philippjfr You're right, completely my fault.  I had `--autoreload` given to `jupyter lab` and I forgot about it.  Sorry for the false alarm.  "Render on Save" checkbox works very well now.

While continuing to investigate the restart issue above, I noticed this line https://github.com/holoviz/panel/blob/e28703da931b29a06c75c47a689b7fa243a99b77/panel/io/jupyter_server_extension.py#L108

Given `path` is now really "full path on disk" and not a URL fragment anymore, that line seems wrong.  But when testing I do not observe anything wrong.  Is it worth to still fix that line?

philippjfr:
Thanks @tlvu. Good catch, I don't think the ApplicationContext.url value is used for anything important, which is why it seems to have no adverse consequences. Should still fix it though.

Issue #3475, opened 2022-04-30
mypy errors


MarcSkovMadsen:
I'm upgrading `panel-chemistry` to use Panel 0.13.0. When I run `mypy` on panel-chemistry I get the errors below. I believe the reason is that Panel is indicating its now typed via `py.typed` file, but its missing type hints.

```python
panel_chemistry\pane\pdbe_molstar.py:178: error: Incompatible types in assignment (expression has type "List[str]", base class "ReactiveHTML" defined the type as "None")
panel_chemistry\pane\ngl_viewer.py:141: error: Incompatible types in assignment (expression has type "None", base class "PaneBase" defined the type as "float")
panel_chemistry\widgets\jsme_editor.py:68: error: Incompatible types in assignment (expression has type "Type[JSMEEditor]", base class "Widget" defined the type 
as "None")
panel_chemistry\widgets\jsme_editor.py:73: error: Dict entry 0 has incompatible type "str": "None"; expected "str": "str"
Found 4 errors in 3 files (checked 21 source files)
```

philippjfr:
Should be resolved now I think.

Issue #2968, opened 2021-12-01
Support mypy by adding py.typed file


MarcSkovMadsen:
When running mypy on my code containing `panel` I get `error: Skipping analyzing "panel": found module but no type hints or library stubs`.

This is a problem for me because 

1) I cannot get help from my editor/ IDE because it does not understand the panel code
2) I have to spend time on *ignoring* this signal.

I believe its relatively easy to fix as its a matter of adding a py.typed file to the project. See https://stackoverflow.com/questions/60716482/error-skipping-analyzing-flask-mysqldb-found-module-but-no-type-hints-or-lib.

I know that adding type hints all over Panel could take time. This is not the request. It is just adding this file.

Thanks.

philippjfr:
I'm at the point where I would happily start adding typing to Panel but am also happy to accept a PR adding the typing file for now.

PR #3476, opened 2022-04-30
Add typehints to help developers and users

MarcSkovMadsen:
Fixes one of the issues in #3475.

- I added lots more typehints to `reactive.py`, to give you (@philippjfr ) an idea about what it takes to add typehints to a module as you say are are prepared for in https://github.com/holoviz/panel/issues/2968#issuecomment-983636021
- I also ran `isort` which reorganized some of the imports slightly.
- I focused on adding typehints to public objects.
- In a few places I would not know how to construct the typehint, so I did not add it.
- I don't add typehints to *parameters* on `Parameterized` classes as I assume they will be inferred automatically one day by automated type stub generation or by implementing the PEP on Param for DataClass like classes.
- I tried my best to use type hints support by Python 3.6 as setup.py says we support this. But the github tests only run for python >= 3.7. So I cannot be 100% sure.

### DIFF
--- a/panel/pane/base.py
+++ b/panel/pane/base.py
@@ -3,11 +3,10 @@ Defines the PaneBase class defining the API for panes which convert
 objects to a visual representation expressed as a bokeh model.
 """
 import warnings
-
 from functools import partial
+from typing import Callable, List, Optional, Union, TYPE_CHECKING, Any, TypeVar, Type
 
 import param
-
 from bokeh.models.layouts import GridBox as _BkGridBox
 
 from ..io import init_doc, push, state, unlocked
@@ -15,11 +14,14 @@ from ..layout import Panel, Row
 from ..links import Link
 from ..models import ReactiveHTML as _BkReactiveHTML
 from ..reactive import Reactive
-from ..viewable import Layoutable, Viewable, Viewer
 from ..util import param_reprs
+from ..viewable import Layoutable, Viewable, Viewer
 
+if TYPE_CHECKING:
+    import bokeh
+    import pyviz_comms
 
-def Pane(obj, **kwargs):
+def Pane(obj: Any, **kwargs) -> 'PaneBase':
     """
     Converts any object to a Pane if a matching Pane class exists.
     """
@@ -32,7 +34,7 @@ def Pane(obj, **kwargs):
     return PaneBase.get_pane_type(obj, **kwargs)(obj, **kwargs)
 
 
-def panel(obj, **kwargs):
+def panel(obj: Any, **kwargs) -> Viewable:
     """
     Creates a panel from any supplied object by wrapping it in a pane
     and returning a corresponding Panel.
@@ -76,13 +78,16 @@ class RerenderError(RuntimeError):
     Error raised when a pane requests re-rendering during initial render.
     """
 
+T = TypeVar('T', bound='PaneBase')
 
 class PaneBase(Reactive):
     """
     PaneBase is the abstract baseclass for all atomic displayable units
-    in the panel library. Pane defines an extensible interface for
+    in the panel library. We call any child class of `PaneBase` a `Pane`.
+    
+    Panes defines an extensible interface for
     wrapping arbitrary objects and transforming them into Bokeh models.
-
+    
     Panes are reactive in the sense that when the object they are
     wrapping is changed any dashboard containing the pane will update
     in response.
@@ -104,7 +109,7 @@ class PaneBase(Reactive):
     # numerical priority is selected. The default is an intermediate value.
     # If set to None, applies method will be called to get a priority
     # value for a specific object type.
-    priority = 0.5
+    priority: Optional[Union[float, bool]] = 0.5
 
     # Whether applies requires full set of keywords
     _applies_kw = False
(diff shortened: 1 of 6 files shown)


panel/reactive.py
```diff
@@ -4,38 +4,38 @@
...
 from .viewable import Layoutable, Renderable, Viewable
 
-log = logging.getLogger('panel.reactive')
+if TYPE_CHECKING:
+    import pandas as pd
 
-LinkWatcher = namedtuple("Watcher", Watcher._fields+('target', 'links', 'transformed', 'bidirectional_watcher'))
+log = logging.getLogger('panel.reactive')
 
+LinkWatcher: Tuple = namedtuple("Watcher", Watcher._fields+('target', 'links', 'transformed', 'bidirectional_watcher'))
```
MarcSkovMadsen, line 38: A `Tuple` is not specific enough. But `collections.namedtuple` is a factory function and not a class. So there is no meaningfull `namedtuple` class. Instead from Python 3.6 `typing.NamedTuple` is provided. See https://www.geeksforgeeks.org/typing-namedtuple-improved-namedtuples/. I'm not sure it would work here as you add attributes dynamically using `Watcher._fields`.

panel/reactive.py
```diff
@@ -399,22 +399,26 @@ class Reactive(Syncable, Viewable):
...
         parameter to a callback which is triggered when the parameter
         changes.
 
         Arguments
         ---------
-        target: object
+        target: param.Parameterized
           The target object of the link.
-        callbacks: dict
+        callbacks: dict | None
```
MarcSkovMadsen, line 419: From python 3.10 you can use `dict | None` instead of `Optional[None]`. I believe its more future proof to us the modern version in the docstrings. But keep the latter in the code for compatibility with python 3.6 for now.

panel/reactive.py
```diff
@@ -529,7 +533,7 @@ def controls(self, parameters=[], jslink=True, **kwargs):
             return controls.layout[0]
         return style.layout[0]
 
-    def jscallback(self, args={}, **callbacks):
+    def jscallback(self, args: Dict={}, **callbacks):
```
MarcSkovMadsen, line 536: The `args` type annotation should be `Dict[str, Something]` where I don't know what `Something` is. So I did not add it.

panel/reactive.py
```diff
@@ -555,10 +559,13 @@ def jscallback(self, args={}, **callbacks):
             callbacks[k] = self._rename.get(v, v)
         return Callback(self, code=callbacks, args=args)
 
-    def jslink(self, target, code=None, args=None, bidirectional=False, **links):
+    def jslink(self, target, code: Dict[str,str]=None,
```
MarcSkovMadsen, line 562: Here I don't now the relevant Union of target types.

panel/reactive.py
```diff
@@ -956,7 +966,7 @@ def __init__(self, **params):
         super().__init__(**params)
         self._old = None
     
-    def _update_selection(self, indices):
+    def _update_selection(self, indices: List[int]):
```
MarcSkovMadsen, line 969: I believe this is correct. But its hard to infer looking at the code. I.e. the typehint will actually help the next developer  :-)

Maybe its even allowed to be `Optional`, who knows?

panel/reactive.py
```diff
@@ -1291,9 +1301,9 @@ class ReactiveHTML(Reactive, metaclass=ReactiveHTMLMetaclass):
 
     __abstract = True
 
-    __css__ = None
-    __javascript__ = None
-    __javascript_modules__ = None
+    __css__: Optional[List[str]] = None
+    __javascript__: Optional[List[str]] = None
```
MarcSkovMadsen, line 1305: This is the one fix I needed.

panel/reactive.py
```diff
@@ -608,7 +617,7 @@ class SyncableData(Reactive):
     with the frontend via a ColumnDataSource.
     """
 
-    selection = param.List(default=[], doc="""
+    selection = param.List(default=[], class_=int, doc="""
```
MarcSkovMadsen, line 620: I believe the `class_` should be int. 

It will help when/ if we can automatically create *stubs* from `Parameterized` classes. Then the type stub can be inferred to `List[int]`.

panel/reactive.py
philippjfr: Happy to drop 3.6 asap but I guess that's not enough.

panel/reactive.py
philippjfr: Difficult one, I think it's `Union[bokeh.model.Model, holoviews.core.Dimensioned, panel.viewable.Renderable, panel.viewable.Viewer]`

panel/reactive.py
philippjfr: Yes, thanks!

panel/reactive.py
philippjfr: Great.

panel/reactive.py
philippjfr: I think that's right.

panel/reactive.py
philippjfr: Actually I think you can also link a simple `param.Parameterized`, will have to double check.

philippjfr:
@MarcSkovMadsen Let's consolidate type hinting to one PR instead of creating a separate PR for each module. I know we won't get through all of them immediately but I think we can make good progress.

panel/reactive.py
MarcSkovMadsen: I don't think there is a problem now. But the did python 3.10 is the minimum supported version the the type hints could be updated.

MarcSkovMadsen:
I'm afraid a larger PRs. Small ones get in. Big ones not always.

MarcSkovMadsen:
Would it be ok for you @philippjfr  to review these to small PRs?

It will minimize the risk I do a lot of work which cannot be approved later without lots of additional work.

philippjfr:
I'm committed to fully typing Panel so there's little risk of this not being mergeable. That said I'll do what I can this weekend and then merge and cut a dev release.

panel/reactive.py
Hoxbro: I would prefer to see the pipe notation from the start, and this will also avoid needing to update the type hints in a couple of years. 

Using the pipe notation is possible from python 3.7 by adding `from __future__ import annotations`. I see no harm in doing this as bokeh is pinned to at least 2.4, and that version does not support python 3.6 [[ref](https://github.com/bokeh/bokeh/wiki/BEP-9:-Downstream-Version-Support)]. Furthermore, if you look at the bokeh's code it also imports from the future.

panel/pane/base.py
```diff
@@ -153,24 +166,28 @@ def __getitem__(self, index):
     #----------------------------------------------------------------
 
     @property
-    def _linkable_params(self):
+    def _linkable_params(self) -> List[str]:
```
Hoxbro, line 169: I just want to note that with the future import, this could be replaced with `def _linkable_params(self) -> list[str]`.

PR #3477, opened 2022-04-30
Add typehints to base.py to help developers and users

MarcSkovMadsen:
Please review #3476 first for context.

Solves one of the issues in #3475.

### DIFF
--- a/panel/pane/base.py
+++ b/panel/pane/base.py
@@ -3,11 +3,10 @@ Defines the PaneBase class defining the API for panes which convert
 objects to a visual representation expressed as a bokeh model.
 """
 import warnings
-
 from functools import partial
+from typing import Callable, List, Optional, Union, TYPE_CHECKING, Any, TypeVar, Type
 
 import param
-
 from bokeh.models.layouts import GridBox as _BkGridBox
 
 from ..io import init_doc, push, state, unlocked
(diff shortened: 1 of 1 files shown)


panel/pane/base.py
```diff
@@ -104,7 +109,7 @@ class PaneBase(Reactive):
     # numerical priority is selected. The default is an intermediate value.
     # If set to None, applies method will be called to get a priority
     # value for a specific object type.
-    priority = 0.5
+    priority: Optional[Union[float, bool]] = 0.5
```
MarcSkovMadsen, line 112: This was the typehint I needed to fix 

```bash
panel_chemistry\pane\ngl_viewer.py:141: error: Incompatible types in assignment (expression has type "None", base class "PaneBase" defined the type as "float")
```

panel/pane/base.py
```diff
@@ -232,21 +237,24 @@ def _update(self, ref=None, model=None):
...
+        can render the object.
+        
+        If the priority of the pane is set to
+        `None`, this method may also be used to define a float priority
         depending on the object being rendered.
         """
         return None
 
-    def clone(self, object=None, **params):
+    def clone(self: T, object: Optional[Any]=None, **params) -> T:
```
MarcSkovMadsen, line 251: The should be the way to type annotate a factory method. See https://github.com/python/typing/issues/58#issuecomment-326240794

panel/pane/base.py
```diff
@@ -441,14 +450,14 @@ def _cleanup(self, root=None):
         self._inner_layout._cleanup(root)
         super()._cleanup(root)
 
-    def select(self, selector=None):
+    def select(self, selector: Union[type, Callable, None]=None) -> List[Viewable]:
```
MarcSkovMadsen, line 453: As far as I can infer it might also be `List[Panel, ReplacementPane]]` but since the docstrings stated `List[Viewable]`, I used that.

panel/pane/base.py
```diff
@@ -441,14 +450,14 @@ def _cleanup(self, root=None):
...
         """
         Iterates over the Viewable and any potential children in the
         applying the Selector.
 
         Arguments
         ---------
-        selector: type or callable or None
+        selector: (type | callable | None)
```
MarcSkovMadsen, line 460: For consistency with the way you can construct unions of types from Python 3.10 I changed the docstring.

philippjfr:
Closing and porting to #3476

PR #3492, opened 2022-05-06
Ensure overflow in MaterialTemplate is not clipped

philippjfr:
Previously content was simply cut-off in the `MaterialTemplate` making some dashboards unusable.


jbednar:
That sounds like a problem to report separately.

PR #3499, opened 2022-05-07
Add more type hints

MarcSkovMadsen:
Adds typehints to button.py and selected dependencies.

### DIFF
--- a/panel/links.py
+++ b/panel/links.py
@@ -414,7 +414,7 @@ class CallbackGenerator(object):
             list(src_model.js_event_callbacks.values()) # type: ignore
         )
         # Skip registering callback if already registered
-        if any(link_id in cb.tags for cb in callbacks):
+        if any(link_id in cb.tags for cbs in callbacks for cb in cbs):
             return
 
         references['source'] = src_model
--- a/panel/links.py
+++ b/panel/links.py
@@ -4,10 +4,12 @@ Defines Links which allow declaring links between bokeh properties.
 import difflib
 import sys
 import weakref
+from typing import Any, Dict
+
+from bokeh.models import CustomJS, LayoutDOM
+from bokeh.models import Model as BkModel
 
 import param
-
-from bokeh.models import CustomJS, Model as BkModel, LayoutDOM
 
 from .io.datamodel import create_linked_datamodel
 from .models import ReactiveHTML
@@ -72,7 +74,7 @@ def assert_target_syncable(source, target, properties):
                 "It requires a live Python kernel to have an effect."
             )
 
-
+UnknownType = Any
 class Callback(param.Parameterized):
     """
     A Callback defines some callback to be triggered when a property
@@ -100,12 +102,29 @@ class Callback(param.Parameterized):
     # Whether the link requires a target
     _requires_target = False
 
-    def __init__(self, source, target=None, **params):
+    def __init__(self, source: UnknownType, target: UnknownType=None, args: Dict[str, UnknownType]=None, code: Dict[str, str]=None):
+        """A Callback defines some callback to be triggered when a property
+        changes on the source object. A Callback can execute arbitrary
+        Javascript code and will make all objects referenced in the args
+        available in the JS namespace.
+
+        Args:
+            source (UnknownType): The source object
+            target (UnknownType, optional): _description_. Defaults to None.
+            args (Dict[str, UnknownType], optional): A `value` object can be referenced by the 
+                `key` name on the JS side. Defaults to None.
+            code (Dict[str, str], optional): _description_. Defaults to None.
+
+        Raises:
+            ValueError: _description_
+        """
         if source is None:
             raise ValueError('%s must define a source' % type(self).__name__)
         # Source is stored as a weakref to allow it to be garbage collected
         self._source = None if source is None else weakref.ref(source)
-        super().__init__(**params)
+        if not args:
+            args={}
+        super().__init__(args=args, code=code)
         self.init()
 
     def init(self):
@@ -163,6 +182,7 @@ class Callback(param.Parameterized):
         arg_overrides = {}
         if 'holoviews' in sys.modules:
             from holoviews.core.dimension import Dimensioned
+
             from .pane.holoviews import HoloViews, generate_panel_bokeh_map
             found = [
                 (link, src, tgt) for (link, src, tgt) in found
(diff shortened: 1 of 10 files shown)


panel/links.py
```diff
@@ -72,7 +74,7 @@ def assert_target_syncable(source, target, properties):
                 "It requires a live Python kernel to have an effect."
             )
 
-
+UnknownType = Any
```
MarcSkovMadsen, line 77: I use this to signal that I simply cannot figure out which types are allowed. I need your help @philippjfr .

panel/links.py
```diff
@@ -100,12 +102,29 @@ class Callback(param.Parameterized):
     # Whether the link requires a target
     _requires_target = False
 
-    def __init__(self, source, target=None, **params):
+    def __init__(self, source: UnknownType, target: UnknownType=None, args: Dict[str, UnknownType]=None, code: Dict[str, str]=None):
```
MarcSkovMadsen, line 105: When you define a custom `__init__` I would argue you have to make the function signature more specific to help the user: `**params` is not very helpful.

panel/reactive.py
```diff
@@ -11,44 +11,44 @@
...
     from bokeh.model import Model
     from bokeh.models.sources import DataDict, Patches
+    from holoviews.core.dimension import Dimensioned
     from pyviz_comms import Comm
 
     from .layout.base import Panel
     from .links import Callback, Link
 
+    # Type Alias
+    JSLinkTarget=Union[Viewable, Model, 'Dimensioned']
```
MarcSkovMadsen, line 50: I would like to make a Type Alias. Something that can be defined in one place and used where ever that specific type is needed.

I am not sure this is the right way. I believe from Python 3.10 you can/ should type annotate it `TypeAlias`.

```
JSLinkTarget: TypeAlias=Union[Viewable, Model, 'Dimensioned']
```

panel/widgets/button.py
```diff
@@ -125,28 +129,62 @@ class Button(_ClickButton):
...
+        callback: (Callable[[param.parameterized.Event], None])
             The function to run on click events. Must accept a positional `Event` argument
+
+        Returns
+        -------
+        watcher: param.Parameterized.Watcher
+          A `Watcher` that executes the callback when the button is clicked.
         """
-        self.param.watch(callback, 'clicks', onlychanged=False)
+        return self.param.watch(callback, 'clicks', onlychanged=False)
```
MarcSkovMadsen, line 187: I updated the function to return the `Watcher` I believe this will not break anything and its more useful as you then have a chance to later delete/ stop/ pause the `Watcher`.

panel/widgets/file_selector.py
```diff
@@ -3,23 +3,24 @@
...
 from ..util import fullpath
+from ..viewable import Layoutable
 from .base import CompositeWidget
 from .button import Button
 from .input import TextInput
 from .select import CrossSelector
 
 
-def scan_path(path, file_pattern='*'):
+def _scan_path(path: StrPath, file_pattern='*') -> Tuple[List[str], List[str]]:
```
MarcSkovMadsen, line 23: I made this function private. Its only used in this module.

panel/widgets/file_selector.py
```diff
@@ -100,7 +101,7 @@ class FileSelector(CompositeWidget):
 
     _composite_type = Column
 
-    def __init__(self, directory=None, **params):
+    def __init__(self, directory: Union[AnyStr, os.PathLike[AnyStr]]=None, **params):
```
MarcSkovMadsen, line 104: I put in the type I believe the will work from directory from looking at the code. But I am in doubt if this is the type hint your want @philippjfr

panel/widgets/button.py
philippjfr: Sounds good!

panel/widgets/file_selector.py
philippjfr: Sounds good!

panel/links.py
```diff
@@ -1,21 +1,40 @@
...
+        from holoviews.core.dimension import Dimensioned
+        JSLinkTarget = Union[Reactive, BkModel, 'Dimensioned']
+    except Exception:
+        JSLinkTarget = Union[Reactive, BkModel] # type: ignore
+    SourceModelSpec = Tuple[Optional[str], str]
+    TargetModelSpec = Tuple[Optional[str], Optional[str]]
 
-def assert_source_syncable(source, properties):
+
+def assert_source_syncable(source: 'Reactive', properties: Iterable[str]) -> None:
```
mattpap, line 37: When using `from __future__ import annotations`, i.e. lazy evaluation of annotations, parentheses aren't needed anymore (in case of own or forward references, etc.).

philippjfr: Yeah, gonna clean that up consistently at some point.

panel/links.py
```diff
@@ -123,32 +172,33 @@ def init(self):
...
+    def register_callback(cls, callback: Type['CallbackGenerator']) -> None:
         """
         Register a LinkCallback providing the implementation for
         the Link for a particular backend.
         """
         cls._callbacks[cls] = callback
 
     @property
-    def source(self):
+    def source(self) -> Reactive | None:
```
mattpap, line 191: This is my personal preference over `Optional[T]` and what we use exclusively in bokeh, but this is 3.10 syntax, and works here only due to lazy annotations and mypy's support. See https://github.com/bokeh/bokeh/issues/11637 discussion for possible issues.

panel/util.py
```diff
@@ -256,33 +264,36 @@ def datetime_as_utctimestamp(value):
     return value.replace(tzinfo=dt.timezone.utc).timestamp() * 1000
 
 
-def is_number(s):
+def is_number(s: Any) -> bool:
```
mattpap, line 267: It would be even better to make this a type guard `def is_number(s: Any) -> TypeGuard[float]` (needs `typing_extensions` for now). This way `s` will be known to be a `float` in `if is_number(s):` or after `assert is_number(s)`.

panel/links.py
philippjfr: Thanks, since bokeh was already using this I've simply adopted the decision but it'll be good to point people to that issue should it come up.

philippjfr:
I'm going to merge and continue in another PR. Otherwise we're inviting tons of merge conflicts.

PR #3514, opened 2022-05-12
Add pre-commit config

philippjfr:
As title says adds a pre-commit configuration file adding some basic linting rules and flake8 to pre-commit rather than running it in the linter test suite. Local setup is easy:

- Install pre-commit with `pip install pre-commit` or `conda install -c conda-forge pre-commit`
- Now run `pre-commit install` in the panel repo root directory
- Now the hooks will run before every commit

On CI the pre-commit hooks run before all the test jobs are started avoiding unnecessarily starting a bunch of jobs.


maximlt:
Excellent!

jbednar:
Can you also summarize the source code changes? What I can see on spot checks is removing trailing whitespace and adding newlines at the end of files. Anything else?

philippjfr:
That is it as far as I'm aware. pre-commit auto fixes the failures and that's the only ones that it reported.

panel/models/reactive_html.py
```diff
@@ -106,6 +106,7 @@ def handle_data(self, data):
         values_loop = re.findall(values_iter_re, data)
         items_loop = re.findall(items_iter_re, data)
         nloops = len(list_loop) + len(values_loop) + len(items_loop)
+        print(list_loop)
```
Hoxbro, line 109: Loose print statement. Wondering if pre-commit could/should catch a print statement like this?

maximlt: I've seen a pre-commit plugin that does that. There are a few other print statements in the code base, so first we'd have to check if they can be replaced by logging calls, and replace them if so.

philippjfr: Thanks @Hoxbro. Can also mark them explicitly to be ignored.

PR #3515, opened 2022-05-13
DateRangeSlider: value and value_throttled turned to DateRange Parameters

maximlt:
The `value` and `value_throttled` Parameters of the `DateRangeSlider` widget were of the Parameter type `Tuple`. This PR turns them into `DateRange` Parameters. Without that:

* their Param validation was not as strict as it could have been
* they could not be properly serialized/deserialized, i.e. Param currently doesn't serialize Parameters that contain datetime values. It will be able to serialize `DateRange` Parameters after  https://github.com/holoviz/param/pull/625 is merged, and maybe more Parameters in the future.

Before this PR the default value of `value` was `(None, None)`. This default value is not accepted by the `DateRange` Parameter that doesn't accept any `None` in the tuple it holds. The default is now `None`, which forced me to make some more changes, including no longer having the `DateRangeSlider` inheriting from `_RangeSliderBase` since it had logic that diverted from the new logic required to handle this change in default value.

Also, I noticed that before the PR it was possible to instantiate the widget without providing any parameter (e.g. without `start` or `end`). This behavior is common to a lot of widgets and is tested in the `tests/widgets/test_base.py` file. This behavior is still respected after the changed made in this PR. However, I've noticed that not providing the `start` or `end` parameters did come with some issues, even if the widget could be instantiated it could never be rendered, Bokeh needs the start/end properties to be set. This should probably be better handled (in another PR!), for now I've just added a warning when start or end aren't set on instantiation.

### DIFF
--- a/panel/widgets/slider.py
+++ b/panel/widgets/slider.py
@@ -574,10 +574,10 @@ class IntRangeSlider(RangeSlider):
         return msg
 
 
-class DateRangeSlider(_RangeSliderBase):
+class DateRangeSlider(_SliderBase):
     """
     The DateRangeSlider widget allows selecting a date range using a
-    slider with two handles. Supports datetime.datetime, datetime.data
+    slider with two handles. Supports datetime.datetime, datetime.date
     and np.datetime64 ranges.
 
     Reference: https://panel.holoviz.org/reference/widgets/DateRangeSlider.html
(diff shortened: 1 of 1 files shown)


panel/widgets/slider.py
```diff
@@ -594,8 +594,8 @@ class DateRangeSlider(_RangeSliderBase):
     ... )
     """
 
-    value = param.Tuple(default=(None, None), length=2, doc=
-        """The selected range as a tuple of values. Updated when one of the handles is
+    value = param.DateRange(default=None, doc="""
+        The selected range as a tuple of values. Updated when one of the handles is
```
philippjfr, line 598: Don't you also have to update `value_throttled`?

maximlt: You're right I missed it thanks for the catch! I'll fix that.

maximlt: I had not yet changed it, I'll submit another PR.

philippjfr: Sorry!

PR #3523, opened 2022-05-16
PANEL_COOKIE_SECRET not verified correctly

govinda18:
Some env variables such as PANEL_COOKIE_SECRET and PANEL_OAUTH_ENCRYPTION are not being verified correctly. This commit fixes that.

Issue#3524

### DIFF
--- a/panel/command/serve.py
+++ b/panel/command/serve.py
@@ -406,7 +406,7 @@ class Serve(_BkServe):
                         "base64-encoded bytes."
                     )
                 config.oauth_encryption_key = encryption_key
-            else:
+            elif not config.oauth_encryption_key:
                 print("WARNING: OAuth has not been configured with an "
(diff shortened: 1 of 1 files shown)


panel/command/serve.py
```diff
@@ -406,7 +406,7 @@ def customize_kwargs(self, args, server_kwargs):
                         "base64-encoded bytes."
                     )
                 config.oauth_encryption_key = encryption_key
-            else:
+            elif not config.oauth_encryption_key:
```
philippjfr, line 409: I agree this fix is indeed correct, but to be clear this fix is simply about the validation right? Because it has no effect on whether or not the env var is respected or not.

govinda18: Yes, I verified that they are being respected - just not verified correctly.

Updated the PR title too.

PR #3525, opened 2022-05-16
ci: cache conda envs betweeen successive builds in a day

MridulS:
Quicker turn around time for github actions. The cache is invalidated every day once (date is used in the key for the cache). It will also rerun if there is any change to setup.py.

There is a similar setup for hvplot.

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -9,7 +9,7 @@ on:
   workflow_dispatch:
   schedule:
     - cron: '0 19 * * SUN'
-    
+
 env:
   CACHE_VERSION: 1
 
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -9,6 +9,9 @@ on:
   workflow_dispatch:
   schedule:
     - cron: '0 19 * * SUN'
+    
+env:
+  CACHE_VERSION: 1
 
 jobs:
   pre_commit:
@@ -35,6 +38,13 @@ jobs:
       matrix:
         os: ['ubuntu-latest', 'macos-latest', 'windows-latest']
         python-version: [3.7, 3.8, 3.9]
+        include:
+        - os: ubuntu-latest
+          path: /usr/share/miniconda3/envs/
+        - os: macos-latest
+          path: /Users/runner/miniconda3/envs/
+        - os: windows-latest
+          path: C:\Miniconda3\envs\
         exclude:
           - os: windows-latest
             python-version: 3.8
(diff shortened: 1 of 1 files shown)


.github/workflows/test.yaml
```diff
@@ -35,6 +38,13 @@ jobs:
       matrix:
         os: ['ubuntu-latest', 'macos-latest', 'windows-latest']
         python-version: [3.7, 3.8, 3.9]
+        include:
+        - os: ubuntu-latest
+          path: /usr/share/miniconda3/envs/
+        - os: macos-latest
+          path: /Users/runner/miniconda3/envs/
+        - os: windows-latest
+          path: C:\Miniconda3\envs\
```
maximlt, line 47: I've seen that these absolute paths have been removed from the [documentation of the setup-miniconda action](https://github.com/conda-incubator/setup-miniconda). Maybe they're no longer required?

.github/workflows/test.yaml
```diff
@@ -65,7 +75,18 @@ jobs:
           channels: pyviz/label/dev,bokeh/label/dev,conda-forge,nodefaults
       - name: Fetch unshallow
         run: git fetch --prune --tags --unshallow
+      - name: Get Today's date for cache
+        run: echo "TODAY=$(date +'%Y-%m-%d')" >> $GITHUB_ENV
+      - name: conda cache
+        uses: actions/cache@v2
+        with:
+          path: ${{ matrix.path }}
+          key: ${{ runner.os }}-conda-${{ matrix.python-version }}-${{ hashFiles('setup.py') }}-${{ env.TODAY }}-${{ env.CACHE_VERSION }}
```
maximlt, line 84: It'd be nice to have the installed group extras added in the cache key, as done in hvplot:

https://github.com/holoviz/hvplot/blob/04f6dae12c449987abf69601be47c1e292e1ab3a/.github/workflows/test.yaml#L61

We may use caching for the docs workflow for instance (we sometimes trigger many docs build in a day, that could help), and need to make sure the tests key not the same as the docs key. Maybe instead of adding the extras group it'd make more sense to add a hash of the test.yaml file itself?

.github/workflows/test.yaml
```diff
@@ -65,7 +75,18 @@ jobs:
...
+          restore-keys: |
+           ${{ runner.os }}-conda-${{ matrix.python-version }}-${{ env.TODAY }}-${{ env.CACHE_VERSION }}
```
maximlt, line 86: In hvplot we had removed that, because of the different workflow types we may run in a single day.

.github/workflows/test.yaml
```diff
@@ -99,6 +121,7 @@ jobs:
         run: |
           eval "$(conda shell.bash hook)"
           conda activate test-environment
+          python -m pip install --no-deps --no-build-isolation -e .
```
maximlt, line 124: Could you move this to its own step before the `doit env_capture` step? I like to have the `doit env_capture` step super simple, that's a least one step I know is not causing any issue when I debug the CI 🙃

maximlt (changes requested):
Thanks for that! 🚀 Just a few suggestions to have caching closer to what has been set on hvplot.

philippjfr:
Thanks for this, it provided a good template. We've now implemented caching as part of a composite action: https://github.com/pyviz-dev/holoviz_tasks/tree/main/install

Issue #3431, opened 2022-04-19
Inverted data when editing a cell in a sorted column


maximlt:
When using header sorters and then editing the data in a tabulator widget the data gets inverted.

As an example, starting with this dataframe:
```
   a  b
0  1  x
1  2  y
2  3  z
```

Then applying a header/frontend sorter, editing a cell (here `3` to `4`) and applying the header sorter again (the data should be displayed in the same order):
```
   a  b
0  4  z
1  2  y
2  1  x
```

philippjfr:
Damn, was worried this could happen when syncing the `sorters`. We'll have to apply the sorters only for the `current_view` and not for the internal `_processed` attribute.

PR #3542, opened 2022-05-20
Apply sorting only on the current view

maximlt:
Fixes https://github.com/holoviz/panel/issues/3431

Also passes the sorters in the configuration properties on instantiation of the tabulator model, which fixes another bug.


maximlt:
Thanks @philippjfr !!!

PR #3543, opened 2022-05-21
Add two-level sidebar toctree

philippjfr:
Screen Shot 2022-05-21 at 18 12 28

### DIFF
--- a/doc/_templates/sidebar-nav-bs.html
+++ b/doc/_templates/sidebar-nav-bs.html
@@ -1,0 +1,11 @@
+<nav class="bd-links" id="bd-docs-nav" aria-label="{{ _('Main navigation') }}">
+  <div class="bd-toc-item active">
+    {{ generate_nav_html("sidebar",
+                         show_nav_level=theme_show_nav_level|int,
+                         maxdepth=theme_navigation_depth|int,
+                         collapse=theme_collapse_navigation|tobool,
+                         includehidden=True,
+                         startdepth=0,
+                         titles_only=True) }}
+  </div>
+</nav>
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -30,7 +30,7 @@ html_theme = "pydata_sphinx_theme"
 html_logo = "_static/logo_horizontal.png"
 html_favicon = "_static/favicon.ico"
 
-html_theme_options.update({
+html_theme_options = {
     "github_url": "https://github.com/holoviz/panel",
     "icon_links": [
         {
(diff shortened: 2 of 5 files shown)


doc/conf.py
```diff
@@ -30,7 +30,7 @@
 html_logo = "_static/logo_horizontal.png"
 html_favicon = "_static/favicon.ico"
 
-html_theme_options.update({
```
maximlt, line 33: The theme options dict was updated because it comes pre-configured from nbsite:

https://github.com/pyviz-dev/nbsite/blob/fe2cded1cfc3519b6c423cd55f52ebc044fabb91/nbsite/shared_conf.py#L96-L100

@philippjfr Did you overwrite it on purpose?

doc/_templates/sidebar-nav-bs.html
```diff
@@ -0,0 +1,11 @@
+<nav class="bd-links" id="bd-docs-nav" aria-label="{{ _('Main navigation') }}">
+  <div class="bd-toc-item active">
+    {{ generate_nav_html("sidebar",
+                         show_nav_level=theme_show_nav_level|int,
+                         maxdepth=theme_navigation_depth|int,
+                         collapse=theme_collapse_navigation|tobool,
+                         includehidden=True,
+                         startdepth=0,
```
maximlt, line 8: Ok I see, so adding `startdepth=0` seemed to do the trick! It's the only difference I can spot with the template from the pydata sphinx theme, do you confirm?

https://github.com/pydata/pydata-sphinx-theme/blob/b806673afbcc5e13a68d2a28b1dde92c8729cbd7/src/pydata_sphinx_theme/theme/pydata_sphinx_theme/components/sidebar-nav-bs.html#L7

Maybe we should submit that upstream, as an optional configuration?

philippjfr: Chris already submitted a PR to this effect but closed it: https://github.com/pydata/pydata-sphinx-theme/pull/536/files

doc/conf.py
philippjfr: Yes, the only option in the shared conf were `footer_items` and these look cleaner than the override in nbsite.

doc/_templates/sidebar-nav-bs.html
maximlt: It's not exactly the same as your change affects all the pages while he was only addressing that on the landing page. Sounds useful to me!

doc/conf.py
```diff
@@ -122,8 +126,8 @@
     'deployment_url': 'https://panel-gallery.pyviz.demo.anaconda.com/'
 }
 
-templates_path += [
```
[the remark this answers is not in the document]
philippjfr: And holoviews.org doesn't use it.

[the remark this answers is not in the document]
maximlt: I should update all the sites in a batch but it's a pain :D
No strong opinion either, and this is a pretty minor difference.

PR #3550, opened 2022-05-23
Add `execute` method to run callbacks and add support to asynchronous `on_edit`/`on_click` tabulator callbacks

maximlt:
This is achieved through the addition of the `run_callback` method on the states that delegates to the callback call to the right context.

Should there me more usage of `run_callback` throughout the code base? e.g. should the`on_click` button events make use of it?

### DIFF
--- a/panel/io/state.py
+++ b/panel/io/state.py
@@ -435,10 +435,11 @@ class _state(param.Parameterized):
 
     def execute(self, callback: Callable([], None)) -> None:
         """
-        Executes both synchronous and asynchronous callbacks appropriately depending on
-        the context the application is running in. When running on the server callbacks are
-        scheduled on the event loop ensuring the Bokeh Document lock is acquired and
-        models can be modified directly.
+        Executes both synchronous and asynchronous callbacks
+        appropriately depending on the context the application is
+        running in. When running on the server callbacks are scheduled
+        on the event loop ensuring the Bokeh Document lock is acquired
+        and models can be modified directly.
 
         Arguments
         ---------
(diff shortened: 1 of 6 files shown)


philippjfr:
>Should there me more usage of run_callback throughout the code base? e.g. should theon_click button events make use of it?

Yes.

panel/io/state.py
```diff
@@ -553,6 +553,24 @@ def publish(
...
+        Run a synchronous callback immediately or schedule an asynchronous
+        callback on the event loop.
+
+        Arguments
+        ---------
+        callback: callable
+          Callback to execute
+        """
+        if param.parameterized.iscoroutinefunction(callback):
+            print(param.parameterized.async_executor.__module__)
```
philippjfr, line 567: Stray print:

```suggestion
```

panel/io/state.py
```diff
@@ -553,6 +553,24 @@ def publish(
             self._rest_endpoints[endpoint] = ([parameterized], parameters, cb)
         parameterized.param.watch(cb, parameters)
 
+    def run_callback(self, callback: Callable([], None)) -> None:
+        """
+        Run a synchronous callback immediately or schedule an asynchronous
+        callback on the event loop.
```
philippjfr, line 559: Something like this:

```suggestion
        Executes both synchronous and asynchronous callbacks appropriately depending on 
        the context the application is running in. When running on the server callbacks are 
        scheduled on the event loop ensuring the Bokeh Document lock is acquired and
        models can be modified directly.
```

PR #3560, opened 2022-05-26
Add type annotation to priority attribute

MarcSkovMadsen:
Fixes type errors when trying to inherit from a Panel class and setting the `priority`.

image

### DIFF
--- a/panel/pane/alert.py
+++ b/panel/pane/alert.py
@@ -3,8 +3,9 @@ Bootstrap inspired Alerts
 
 See https://getbootstrap.com/docs/4.0/components/alerts/
 """
-import param
+from typing import Union
 
+import param
 from panel.pane.markup import Markdown
 
 ALERT_TYPES = ["primary", "secondary", "success", "danger", "warning", "info", "light", "dark"]
@@ -24,7 +25,7 @@ class Alert(Markdown):
 
     alert_type = param.ObjectSelector("primary", objects=ALERT_TYPES)
 
-    priority: float | bool | None = 0
+    priority: Union[float, bool, None] = 0
 
     _rename = dict(Markdown._rename, alert_type=None)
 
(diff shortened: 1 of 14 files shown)


panel/pane/alert.py
```diff
@@ -24,7 +24,7 @@ class Alert(Markdown):
 
     alert_type = param.ObjectSelector("primary", objects=ALERT_TYPES)
 
-    priority = 0
+    priority: float | bool | None = 0
```
MarcSkovMadsen, line 27: I used the type `float | bool | None` as it was already used in the `PaneBase` class.

panel/pane/base.py
```diff
@@ -5,17 +5,14 @@
...
+                    TypeVar, Union)
 
 import param
-
-from bokeh.models.layouts import (
-    GridBox as _BkGridBox, Tabs as _BkTabs, Panel as _BkPanel
-)
+from bokeh.models.layouts import GridBox as _BkGridBox
+from bokeh.models.layouts import Panel as _BkPanel
+from bokeh.models.layouts import Tabs as _BkTabs
```
philippjfr, line 15: I know isort does this but I really don't like it. Is there an isort configuration that avoids this because at that point I'd be happy to enable it?

panel/pane/base.py
```diff
@@ -119,7 +116,7 @@ class PaneBase(Reactive):
     # numerical priority is selected. The default is an intermediate value.
     # If set to None, applies method will be called to get a priority
     # value for a specific object type.
-    priority: float | bool | None = 0.5
+    priority: Union[float, bool, None] = 0.5
```
philippjfr, line 119: Please use this throughout:

```python
...
    priority: float | bool | None = 0.5
```

panel/pane/base.py
mattpap: See https://github.com/bokeh/bokeh/blob/530c735f75f065b33acced0096f58e265aae4a65/bokeh/.isort.cfg#L5.

panel/pane/base.py
MarcSkovMadsen: Thanks I also prefer that, but did not know it was possible.

philippjfr:
Going to close this and redo this. To painful to rebase at this point.

PR #3561, opened 2022-05-26
Add type annotation to _rename attribute

MarcSkovMadsen:
Fixes typing errors when trying to inherit from Panel class with `_rename` attribute.

image

### DIFF
--- a/panel/io/location.py
+++ b/panel/io/location.py
@@ -7,7 +7,7 @@ import json
 import urllib.parse as urlparse
 
 from typing import (
-    TYPE_CHECKING, Any, Callable, Dict, List, Mapping, Optional,
+    TYPE_CHECKING, Any, Callable, ClassVar, Dict, List, Mapping, Optional,
 )
 
 import param
@@ -58,7 +58,7 @@ class Location(Syncable):
         should be set to False""")
 
     # Mapping from parameter name to bokeh model property name
-    _rename: Mapping[str, Union[str, None]] = {"name": None}
+    _rename: ClassVar[Mapping[str, str | None]] = {"name": None}
 
     def __init__(self, **params):
         super().__init__(**params)
--- a/panel/io/location.py
+++ b/panel/io/location.py
@@ -5,10 +5,7 @@ from __future__ import annotations
 
 import json
 import urllib.parse as urlparse
-
-from typing import (
-    TYPE_CHECKING, Any, Callable, Dict, List, Mapping, Optional
-)
+from typing import TYPE_CHECKING, Any, Callable, Dict, List, Mapping, Optional
 
 import param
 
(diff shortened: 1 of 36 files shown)


panel/widgets/tables.py
```diff
@@ -1,29 +1,30 @@
...
-
 from bokeh.models import ColumnDataSource
-from bokeh.models.widgets.tables import (
-    AvgAggregator, CellEditor, CellFormatter, CheckboxEditor,
-    DataCube, DataTable, DateEditor, DateFormatter, GroupingInfo,
-    IntEditor, MaxAggregator, MinAggregator, NumberEditor,
-    NumberFormatter, RowAggregator, StringEditor, StringFormatter,
-    SumAggregator, TableColumn
-)
+from bokeh.models.widgets.tables import (AvgAggregator, CellEditor,
```
MarcSkovMadsen, line 10: I use isort to autosort. This is how it changes the code.

panel/io/location.py
```diff
@@ -58,7 +56,7 @@ class Location(Syncable):
         should be set to False""")
 
     # Mapping from parameter name to bokeh model property name
-    _rename = {"name": None}
+    _rename: Mapping[str, Union[str, None]] = {"name": None}
```
MarcSkovMadsen, line 59: I used `Mapping[str, Union[str, None]]` type annotation instead of  `KeysAndGetItem[str, Union[str, None]]` as it was already used it the Panel code. I am not sure if its the right choice though.

mattpap: I presume `_rename` is a static member. Then the type should be wrapped with `ClassVar` to indicate this (everywhere).

mattpap: `KeysAndGetItem` seems to be an oddly specific type (where does it come from btw.?). `Mapping` is fine in this context, but it really depends on the interface you want to expose. The smaller the surface of the API, the better. One can always expand it, by changing the type to the more specific `Dict[..., ...]`, but not the other way, at least without breaking compatibility.

MarcSkovMadsen: `KeysAndGetItem` was the type reported by mypy as the expected type. It was inferred from the attribute value being a dictionary.

Thanks for mentioning ClassVar. I learned something new there.

philippjfr: Yep, also just learned about ClassVar.

panel/widgets/indicators.py
```diff
@@ -32,7 +35,7 @@
 from ..models import (
     HTML, Progress as _BkProgress, TrendIndicator as _BkTrendIndicator,
 )
-from ..pane.markup import Str
+from ..pane import Str
```
philippjfr, line 38: Looks like `Str` actually needs to be exported in `panel/pane/__init__.py`. Not sure why it's not.

PR #3562, opened 2022-05-26
Improve BasicTemplate type hints

MarcSkovMadsen:
Fixes issues like below

image

### DIFF
--- a/panel/io/callbacks.py
+++ b/panel/io/callbacks.py
@@ -9,11 +9,9 @@ import time
 
 import param
 
-from bokeh.io import curdoc as _curdoc
-
 from ..util import edit_readonly, function_name
 from .logging import LOG_PERIODIC_END, LOG_PERIODIC_START
-from .state import state
+from .state import curdoc_locked, state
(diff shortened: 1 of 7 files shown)


MarcSkovMadsen:
I don't know how you got type hints in using `| ` @philippjfr ? These will fail when running on Python < 3.10 I believe.

image

philippjfr:
The union operator is compatible with `from __future__ import annotations` in older versions of python. You just can't express type aliases that way.

MarcSkovMadsen:
> Thanks. Does it mean I should refactor this one to annotate using `Path | str | List[Path| str]` instead of `Union[Path, str, List[Union[Path, str]]` and add `from __future__ import annotations` @philippjfr ?

I can see from https://github.com/holoviz/panel/pull/3560#discussion_r882431205 that it does. I will update.

philippjfr:
Perfect, thanks!

philippjfr:
Finished typing `template/base.py` in this PR.

Issue #3584, opened 2022-06-03
Player.interval required to be an integer parameter when float should be acceptable


dmarx:
The "interval" attribute can be used to set the starting play speed of the Player widget.

Under the hood, usage of this value clearly does not presume that it is an integer, so it seems unnecessary (and constraining) to restrict users to providing integer values in the widget's init method:

https://github.com/holoviz/panel/blob/646f4d297994ac3798186276152ddba48542871e/panel/models/player.ts#L264-L270

Forcing users to provide integer values only permits arbitrarily accelerating the animation

demo code:

```
# permitted:
kargs['i'] = pn.widgets.Player(interval=1, name='step', start=1, end=n_imgs_per_group, step=1, value=1, loop_policy='reflect')

# raises an error:
kargs['i'] = pn.widgets.Player(interval=0.5, name='step', start=1, end=n_imgs_per_group, step=1, value=1, loop_policy='reflect')
```

dmarx:
oh you know what... I think my confusion here was just that the default interval is 500. this probably doesn't need to be a float, I misinterpreted this parameter as a multiplier on the default play speed rather than a number of microseconds. clarifying that this interval is "microseconds between updates" in the docstring and help docs would probably resolve this confusion going forward.

maximlt:
You're right the docs could be improved there, do you feel like opening a PR to fix that?

dmarx:
Sure, no problem

PR #3585, opened 2022-06-03
docstrings - interval is specified in ms

dmarx:
followup from #3584

### DIFF
--- a/panel/widgets/player.py
+++ b/panel/widgets/player.py
@@ -11,7 +11,9 @@ from .select import SelectBase
 
 class PlayerBase(Widget):
 
-    interval = param.Integer(default=500, doc="Interval between updates, in milliseconds. Default is 500, i.e. two updates per second")
+    interval = param.Integer(default=500, doc="""
+        Interval between updates, in milliseconds. Default is 500, i.e. 
+        two updates per second.""")
 
     loop_policy = param.ObjectSelector(default='once',
                                        objects=['once', 'loop', 'reflect'], doc="""
--- a/panel/widgets/player.py
+++ b/panel/widgets/player.py
@@ -11,7 +11,7 @@ from .select import SelectBase
 
 class PlayerBase(Widget):
 
-    interval = param.Integer(default=500, doc="Interval between updates")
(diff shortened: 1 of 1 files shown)


jbednar (approved):

panel/widgets/player.py
```diff
@@ -11,7 +11,7 @@
 
 class PlayerBase(Widget):
 
-    interval = param.Integer(default=500, doc="Interval between updates")
+    interval = param.Integer(default=500, doc="Interval between updates, in milliseconds. Default is 500, i.e. two updates per second")
```
maximlt, line 14:
```suggestion
    interval = param.Integer(default=500, doc="""
        Interval between updates, in milliseconds. Default is 500, i.e. 
        two updates per second.""")
```

maximlt (changes requested):
Thanks for this PR, that was fast 👍 ! I'm just suggesting a small formatting change (we don't use `black` as it doesn't format well the declaration of Parameters like `interval`, unfortunately).

panel/widgets/player.py
mattpap: Thinking about this from API design perspective, this information could be encoded in the type of the property, i.e. have `param.Milliseconds(...)`, which would be, at least at first, a simple alias for `param.Integer` (well, I suppose non-negative integer, as most likely we don't want negative intervals to be allowed). This then would allow to push the API even further, e.g. allowing string values like `1s` that would be parsed and normalized to the base type and units.

dmarx: interesting idea. you could probably generalize that a bit further by making it a generic [timedelta](https://docs.python.org/3/library/datetime.html#timedelta-objects) type with a "default units" option or something like that.

jbednar: Unit support in Param is a longstanding issue, and is something that is orthogonal to the Parameter type in general. A Parameter _instance_ (not necessarily class) should be able to declare its type, e.g. using [pint](https://pint.readthedocs.io), and then if someone passes a value with a compatible type, it will be converted into the desired type. We had a partial implementation at some point, but I don't see it in the PRs, so that might have predated this project being on Github.  Still something worth doing!

Issue #3583, opened 2022-06-03
Components do not maximize height when used in (Fast-) template


MarcSkovMadsen:
My users and co-developers are asking me to create applications that adjust nicely to big traders screens, smaller laptop screens and custom window sizes.

Its normally not a problem to get `width` working. But `height` I can very often not get working when using either prebuilt or custom templates.

There is an example below. Any fix or workaround suggestion would be appreciated.

## No Template - works well

![image](https://user-images.githubusercontent.com/42288570/171833611-79550531-a028-479b-a6e1-d52a7c254a4a.png)

```python
import panel as pn
import pandas as pd
import hvplot.pandas

pn.extension(
  sizing_mode="stretch_width",
  # template="fast",
)

data = pd.DataFrame({
    "x": list(range(0,100)), "y": list(range(0,100))
})
table_pane = pn.widgets.Tabulator(
  data,
  show_index=False,
  theme="fast",
  height=500,
  sizing_mode="stretch_both",
  disabled=True,
)
component = pn.Column(
    table_pane, 
    pn.widgets.Button(name="download"), 
    sizing_mode="scale_both", 
    background="lightgray"
)

component.servable()
```

## Template - does not work well

![image](https://user-images.githubusercontent.com/42288570/171833454-5e5daeb8-f983-476a-a66c-f062cc96401f.png)

Please notice if I don't add the `heigth=500` the height will only be around 50!

```python
import panel as pn
import pandas as pd
import hvplot.pandas

pn.extension(
  sizing_mode="stretch_width",
  template="fast",
)

data = pd.DataFrame({
    "x": list(range(0,100)), "y": list(range(0,100))
})
table_pane = pn.widgets.Tabulator(
  data,
  show_index=False,
  theme="fast",
  height=500,
  sizing_mode="stretch_both",
  disabled=True,
)
component = pn.Column(
    table_pane, 
    pn.widgets.Button(name="download"), 
    sizing_mode="scale_both", 
    background="lightgray"
)

component.servable()
```

MarcSkovMadsen:
## Template -  Workaround

If I set the `.bk-root` my self to something responsive it seems to work

![image](https://user-images.githubusercontent.com/42288570/171839320-4a5f66c4-7dd5-4e29-b61f-e1e51016ec18.png)

```python
import panel as pn
import pandas as pd
import hvplot.pandas

pn.config.raw_css.append(
  """
  .bk-root {
    height: calc( 100vh - 200px ) !important;
  }
[...]

PR #3607, opened 2022-06-12
Feature/maximize fast panels

MarcSkovMadsen:
Addresses #3583

Adds support for maximizing roots/ panels responsively. This makes it much easier to make the layouts I want to create 80% of the time.

### Design Considerations

I chose the parameter name `maximize_panels`. I also considered `maximize_roots` (Do users understand what a root is?) and `maximize_main_panels`.

### Reference Test Example

https://user-images.githubusercontent.com/42288570/173219553-dd3062f1-1943-4053-b8c3-9e1fe67dacfa.mp4

```python
import panel as pn
import pandas as pd
import hvplot.pandas

pn.extension(
  sizing_mode="stretch_width",
  template="fast",
)
pn.state.template.param.update(maximize_panels=True)

data = pd.DataFrame({
    "x": list(range(0,100)), "y": list(range(0,100))
})
table_pane = pn.widgets.Tabulator(
  data,
  show_index=False,
  theme="fast",
  sizing_mode="stretch_both",
  disabled=True,
)
component = pn.Column(
    table_pane, 
    pn.widgets.Button(name="download"), 
    sizing_mode="scale_both", 
)

component.servable()
```

### DIFF
--- a/panel/template/fast/list/fast_list_template.html
+++ b/panel/template/fast/list/fast_list_template.html
@@ -54,6 +54,11 @@
       box-shadow: 2px 2px 10px silver;
   }
   {% endif %}
+  {% if style.maximize_panels %}
+  .bk-root {
+    height: calc( 100vh - 100px ) !important;
+  }
+  {% endif %}
 </style>
 
 <!-- Template JS -->
--- a/panel/template/fast/list/fast_list_template.html
+++ b/panel/template/fast/list/fast_list_template.html
@@ -55,7 +55,7 @@
   }
   {% endif %}
   {% if style.maximize_panels %}
-  .bk-root {
+  .main .bk-root {
     height: calc( 100vh - 100px ) !important;
   }
   {% endif %}
--- a/examples/reference/templates/FastListTemplate.ipynb
+++ b/examples/reference/templates/FastListTemplate.ipynb
@@ -57,6 +57,7 @@
     "* **`corner_radius`** (str): The corner radius applied to controls.\n",
     "* **`font`** (str): A font url to import.\n",
     "* **`font_url`** (str): A font url to import.\n",
+    "* **`maximize_panels`** (bool): If `True`, the root container for each panel added to the main area will be maximized. Additionally you will have to add sizing_mode='stretch_both' to all panels to force them to take up the available window space. Default is `False`.\n",
     "* **`shadow`** (str): Optional shadow override. Whether or not to apply shadow.\n",
     "* **`main_layout`** (str): What to wrap the main components into. Options are '' (i.e. none) and 'card' (Default). Could be extended to Accordion, Tab etc. in the future.\n",
     "\n",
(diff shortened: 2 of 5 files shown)


panel/template/fast/list/fast_list_template.html
```diff
@@ -54,6 +54,11 @@
       box-shadow: 2px 2px 10px silver;
   }
   {% endif %}
+  {% if style.maximize_panels %}
+  .bk-root {
+    height: calc( 100vh - 100px ) !important;
```
philippjfr, line 59: This scares me, you certainly don't want all roots to be full screen height right?

philippjfr:
Personally -1 on this change, I don't really see the utility of all roots using full screen height. There's definitely something off about height scaling in the Fast templates but I don't think this is the right fix.

panel/template/fast/list/fast_list_template.html
MarcSkovMadsen: My intention is to maximize all roots in the main area. I can see I do more and will adjust.

The alternative is to add functionality to maximize the first root only.

If I look at my apps awesome-panel.org and at work most of these are are should just have one maximized main root. Its such a bad user experience currently where a main root is around 500px by default or alternatively some fixed size specified by the user. 

Users come complaining it does not fit their big trader screen, their small laptop screen or a custom window size they need to fit it in their complex set of windows. Developers come complaining that laying out apps is too difficult in Panel.

MarcSkovMadsen: I don't see and option that is `False` by default as very scary. But it would make it so much easier to create apps that layout nicely.

MarcSkovMadsen: I've changed the selector to `.main .bk-root`.

MarcSkovMadsen:
How would you solve this issue @philippjfr ? I don't have the imagination to see another solution technically.

philippjfr:
Not sure yet, maybe put together some apps that you want to behave nicely and then let's iterate on it.

MarcSkovMadsen:
At awesome-panel.org I would actually like the second main root maximized as the first normally contains a description.

## Leaflet component using ReactiveHTML

For example this one I would like to maximize the root with the plot https://awesome-panel.org/reactive_leaflet. The source code is here https://github.com/MarcSkovMadsen/awesome-panel/blob/master/awesome_panel/apps/reactive_leaflet.py.

image

## Google Map Viewer

https://awesome-panel.org/google_map_viewer. The code is here https://github.com/MarcSkovMadsen/awesome-panel/blob/master/awesome_panel/apps/google_map_viewer.py

image

MarcSkovMadsen:
Ok. I can see we need something better than maximizing either the first or all main roots. We probably need the ability to maximize one or more specific main roots. Maybe the solution is to document how it can be done using custom `.css`?

philippjfr:
I think we just need to ensure that `sizing_mode` behaves correctly and that a root declared with `stretch_both` or `stretch_height` actually fills the available space.

philippjfr:
Appreciate you getting this started. I think it's definitely an issue and I've noticed it too, I just don't believe in this approach and would rather iterate on #3624.

Issue #3249, opened 2022-03-18
Patching values in Tabulator causes the vertical scrollbar to reset


Kadek:
#### ALL software version info
Python 3.8.8
panel 0.13.0a43

#### Description of expected behavior and the observed behavior

Expected behaviour: When patching data in Tabulator the vertical scrollbar should stay in the same position.
Observed behaviour: When using patch function on Tabulator the vertical scrollbar jumps to the top

#### Complete, minimal, self-contained example code that reproduces the issue

```
import numpy as np
import pandas as pd
import panel as pn

df = pd.DataFrame(np.random.randn(10, 5), columns=list('ABCDE'))
styled = pn.widgets.Tabulator(df, height=200)

def highlight_max(s):
    '''
    highlight the maximum in a Series yellow.
    '''
    is_max = s == s.max()
    return ['background-color: yellow' if v else '' for v in is_max]

styled.style.apply(highlight_max)


def start_updating(*args):
    pn.state.add_periodic_callback(update, period=1000)

def update():
    df = pd.DataFrame(np.random.randn(10, 5), columns=list('ABCDE'))
    styled.patch(df)

button = pn.widgets.Button(name="Start")
button.on_click(start_updating)

pn.Column(button, styled).servable()

```

#### Screenshots or screencasts of the bug in action

https://user-images.githubusercontent.com/2953645/159003446-e99ebde0-53c4-4bcf-a417-8a3c109c865c.mp4

philippjfr:
Apparently this is still an issue. Will attempt reproduce tomorrow.

Kadek:
I believe there are two problems with the current fix.

1. When calling scrollToRow we set ifVisible = false. In the example visible in the video if we scroll so that 5th row is the top visible one. Then we won't scroll to it because it's already visible when we reset the scroll bar on patch.
2. The scrolling works correctly if we add more rows and scroll to something not visible on the initial load. The interesting problem that occurs then is that if we scroll to the 8th row, then a little bit of the 7th row is visible. On the next patch we will therefore scroll to the 7th row. Then a little bit of 6th row is visible and so we repeat until we are back in the case 1. I think this requires changing the second parameter of scrollToRow but haven't had a chance to experiment yet.

Kadek:
[...]

PR #3645, opened 2022-06-23
Try to restore tabulator's scrollbar position on every renderComplete not only on the initial one

Kadek:


maximlt:
Continuing the discussion that was going on in the issue https://github.com/holoviz/panel/issues/3249 here. Thanks for rebasing, that's helpful. Indeed the failing tests seem unrelated, they'll have to be fixed separately.

I recently added a UI Tabulator test that was marked as `xfail`, i.e. that is expected to fail, thinking that it would cover the issue your opened:
https://github.com/holoviz/panel/blob/ff39f02493f349569b2b18bfe4f863e67fc3a034/panel/tests/ui/widgets/test_tabulator.py#L1072-L1073

Fixing [your issue](https://github.com/holoviz/panel/issues/3249) in this PR should have made the test pass, and as it's marked as `xfail` it should actually show up as failed this time. It's not the case so something went wrong. It might be that the test I wrote is still failing for another reason, or that it was not covering the issue you reported but another one. I think it's the latter. I'll investigate and write a new test if need be.

philippjfr:
Thanks @Kadek and sorry this has sat here for so long. Things have changed quite a bit so I'll likely have to find a new solution.

PR #3679, opened 2022-07-09
Implement ability to reuse session for initial render

philippjfr:
This is an experiment to see if we can improve the speed of handling the initial request. The basic idea behind this is that in most cases the initial page served to the user is identical, i.e. the template does not differ between sessions. As long as that holds true we can simply cache the first session for each endpoint and reuse that to render the template served to the user. In the background we can then still create a session for each user which the rendered template connects to.

The way this works is that we simply modify the `token` which tells the frontend which session it should connect to. So the steps are:

1. User visits application
2. The `DocHandler` uses a cached session to render the template
3. We swap out the `session_id` in the token so that the frontend does not re-connect to the existing session
4. When the frontend establishes the Websocket connection a new session is created

This will significantly reduce time to first render but also comes with some caveats and drawbacks:

- Even though the initial render is faster establishing the websocket connection will be slower because we have simply postponed when the session is created
- This will not work if the initial page render is dependent on some state, e.g. a query parameter determines the theme or a header/cookie is used to render the user into the header.

### DIFF
--- a/panel/command/serve.py
+++ b/panel/command/serve.py
@@ -209,6 +209,10 @@ class Serve(_BkServe):
             help    = "The endpoint for the liveness API.",
             default = "liveness"
         )),
+        ('--reuse-sessions', dict(
+            action  = 'store_true',
+            help    = "Whether to reuse sessions when serving the initial request."
+        )),
     )
 
     # Supported file extensions
@@ -277,6 +281,7 @@ class Serve(_BkServe):
             raise ValueError("rest-provider %r not recognized." % args.rest_provider)
 
         config.autoreload = args.autoreload
+        config.reuse_sessions = args.reuse_sessions
 
         if config.autoreload:
             for f in files:
--- a/panel/command/serve.py
+++ b/panel/command/serve.py
@@ -211,7 +211,7 @@ class Serve(_BkServe):
         )),
         ('--reuse-sessions', dict(
             action  = 'store_true',
-            help    = "Whether to reuse sessions when serving the initial request."
+            help    = "Whether to reuse sessions when serving the initial request.",
         )),
     )
 
--- a/panel/config.py
+++ b/panel/config.py
@@ -156,7 +156,13 @@ class _config(_base_config):
         the initial page render. Note that if the initial page differs
         between sessions, e.g. because it uses query parameters to modify
         the rendered content, then this option will result in the wrong
-        content being rendered.""")
+        content being rendered. Define a session_key_func to ensure that
+        reused sessions are only reused when appropriate.""")
+
+    session_key_func = param.Callable(default=None, doc="""
+        Used in conjunction with the reuse_sessions option, the
+        session_key_func is given a tornado.httputil.HTTPServerRequest
+        and should return a key that uniquely captures a session.""")
 
     safe_embed = param.Boolean(default=False, doc="""
         Ensure all bokeh property changes trigger events which are
--- a/panel/config.py
+++ b/panel/config.py
@@ -151,6 +151,13 @@ class _config(_base_config):
         'pyinstrument', 'snakeviz', 'memray'], doc="""
         The profiler engine to enable.""")
 
+    reuse_sessions = param.Boolean(default=False, doc="""
+        Whether to reuse a session for the initial request to speed up
+        the initial page render. Note that if the initial page differs
+        between sessions, e.g. because it uses query parameters to modify
+        the rendered content, then this option will result in the wrong
+        content being rendered.""")
+
     safe_embed = param.Boolean(default=False, doc="""
         Ensure all bokeh property changes trigger events which are
         embedded. Useful when only partial updates are made in an
--- a/panel/io/server.py
+++ b/panel/io/server.py
@@ -359,7 +359,7 @@ class DocHandler(BkDocHandler, SessionPrefixHandler):
         from ..config import config
         path = self.request.path
         session = None
-        if path in self._session_key_funcs:
+        if config.reuse_sessions and path in self._session_key_funcs:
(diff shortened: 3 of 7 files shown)


examples/user_guide/Performance_and_Debugging.ipynb
```diff
@@ -73,7 +73,43 @@
...
+    "\n",
+    "Luckily, as long as certain conditions are met you can easily cache sessions ensuring that the initial page is rendered very quickly. You can enable this behavior by setting `pn.config.reuse_sessions = True` (or equivalently `pn.extension(..., reuse_sessions=True)`).\n",
+    "\n",
+    "So let us explore what conditions need to be met for it to be safe to enable this behavior. Luckily there really is only one condition:\n",
+    "\n",
+    "> The the placement and number of Panel components as well as any template configuration should not be dependent on any external state that can change.\n",
```
maximlt, line 84:
```suggestion
    "> The placement and number of Panel components as well as any template configuration should not be dependent on any external state that can change.\n",
```

examples/user_guide/Overview.ipynb
```diff
@@ -172,6 +172,7 @@
     "> - `loading_color`: The color of the global loading indicator as a hex color, e.g. #6a6a6a\n",
     "> - `nthreads`: If set will start a `ThreadPoolExecutor` to dispatch events to for concurrent execution on separate cores. By default no thread pool is launched, while setting nthreads=0 launches `min(32, os.cpu_count() + 4)` threads.\n",
     "> - `raw_css`: List of raw CSS strings to add to load.\n",
+    "> - `reuse_sessions`: Whether to reuse a session for the initial request to speed up the initial page render. Note that if the initial page differs between sessions, e.g. because it uses query parameters to modify the rendered content, then this option will result in the wrong content being rendered.\n",
```
maximlt, line 175: > Note that if the initial page differs between sessions, e.g. because it uses query parameters to modify the rendered content, then this option will result in the wrong content being rendered.

According to the docs it seems like it's more subtle than that. Maybe linking to the docs section introducing reusable sessions would be better than trying to explain the limitations here?

examples/user_guide/Performance_and_Debugging.ipynb
```diff
@@ -73,7 +73,43 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "## Layouts\n",
+    "## Reuse sessions\n",
+    "\n",
+    "In order to render a page for a user visiting your application the Panel server creates a so called session. This session has to execute the code associated with your application and then render the page template to HTML. Once the frontend has rendered the initial template it will establish a Websocket connection and render the actual Panel components into the template. This means that in order for the user to see any actual content the entire application has to be executed. While this can be sped up through the use of caching it will always take a little while for the initial page to be rendered.\n",
```
maximlt, line 78: If I'm correct, when a Panel app is served without using a Panel template (e.g. Bootstrap) there is a default template being served (from Panel or Bokeh?). That may need to be introduced.

philippjfr: Agreed, was kind of skirting around this but best to explain this properly.

examples/user_guide/Performance_and_Debugging.ipynb
```diff
@@ -73,7 +73,43 @@
...
+    "\n",
+    "- The number of items you have marked as `.servable` or added to a template's `main`, `sidebar` and `header` areas.\n",
+    "- Secondly, if you are using a `Template` its parameters should not change.\n",
+    "\n",
+    "Some examples include:\n",
+    "\n",
+    "- You may not load a dataset that might change and then add a variable number of plots.\n",
+    "- Set a template parameter based on some changing value `BootstrapTemplate(title=f'Dashboard at {dt.datetime.now()}')`\n",
+    "\n",
+    "One exception is when the template changes based on something in the HTTP request, e.g. the URL query parameters, headers or cookies. In these cases you have to tell Panel that there is some dependency on a request argument. Specifically we have to define a `pn.config.session_key_func`. Let's say for example that we switch the theme depending on a 'theme' query parameter. To ensure Panel caches the session for all possible theme values we define a function that returns a key that includes the entire URI `request.uri` (e.g. `/myapp?theme=dark`):\n",
```
maximlt, line 96: I find this part not very easy to understand. In the tests you return a tuple, how does that work?

philippjfr: The session is cached depending on the key. Any hashable value will work as a key, e.g. request.path, request.uri or a tuple of values will all work. By default sessions are cached entirely based on the `request.path` (e.g. the `/app` or `/app2`), however if the contents of your template changes based on some query argument (e.g. `?theme=dark`) you can provide this custom function that returns a key ensuring that the cache does not conflate a rendered template for `theme='dark'` and `theme='default. Is that clearer? Would love suggestions on how to clarify this in docs.

maximlt:
```suggestion
    "One exception is when the template changes based on something in the HTTP request, e.g. the URL query parameters, headers or cookies. In these cases you have to tell Panel that there is some dependency on a request argument. Specifically we have to define `pn.config.session_key_func`, giving it a callable that should return a key that will be used to cache the session. Let's say for example that we switch the theme depending on a 'theme' query parameter. To ensure Panel caches the session for all possible theme values we define a function that returns a key that includes the entire URI `request.uri` (e.g. `/myapp?theme=dark`):\n",
```

maximlt: Small suggestion to improve the docs based on your previous comment. I just missed that the returned value would be the cache key (even if it's written in the paragraph), so emphasizing that early.

PR #3686, opened 2022-07-12
Wrong example signature

keul:
Callback should follow the `_{node-id}_{event}`, but example uses a double underscore (which is not working)


philippjfr:
Thanks again!

PR #3690, opened 2022-07-13
Properly describe where the .type attribute of DOMEvent is found

keul:
Documentation is currently wrong, as there's no `.type` attribute on DOMEvent, but it's `.data.type`


philippjfr:
Thanks @keul!

PR #3691, opened 2022-07-13
Add Accordion UI tests

thuydotm:
Adding UI tests to `panel.Accordion`. The newly added tests verify if Accordion works properly with different parameter setting at initialization. Some tests are to check the behavior of Accordion as a list-like object when performing following methods: `append, extend, clear, insert, pop, remove`.


philippjfr:
Unrelated and spurious test failure. Thanks for adding these @thuydotm!

PR #3694, opened 2022-07-14
Update Admin Logs page to use Tabulator

### LINKED ISSUE #3435: Streaming new data to tabulator widget with show_index=False raises a ValueError exception
When streaming new data to a tabulator widget, if the show_index value is set to False then a ValueError exception is raised.
It happens with both DataFrames and Dictionaries.

This code below demonstrates it : 

```
import pandas as pd
import panel as pn

df = pd.DataFrame(np.random.randn(10, 5), columns=list('ABCDE'))
tabulator=pn.widgets.Tabulator(df, show_index=False)
def stream_data(event):
    stream_dic = {
        "A":0,
        "B":0,
        "C":0,
        "D":0,
        "E":0
    }
    stream_df = pd.DataFrame(np.random.randn(10, 5), columns=list('ABCDE'))
    tabulator.stream(stream_df)

button = pn.widgets.Button(name="stream")
button.on_click(stream_data)
pn.Row(tabulator, button)
```

If you set show_index to True, the new data is streamed as expected. 
This did not happen in version 0.12.6.

Current environment :
Panel 0.13.0
Pandas 1.3.2
Jupyter Lab

### LINKED ISSUE #3623: Admin Panel: allow to filter the previous logs
The Logs tab currently rely on the Terminal widget to show the logs emitted by Panel directly. It also offers widgets to filter the logs by sessions and components. However these widgets only affect the future logs, not the logs already recorded and streamed to the Terminal. In most cases I believe users will be interested in filtering previous logs, for instance to see what went wrong with a given session.

One way to improve the Logs tab would be to stream the logs to a Tabulator widget instead of to the Terminal widget, which could have columns such as timestamp, log level, session, component and message and activate header filters and sorters.

Hoxbro:
Fixes #3623 
Fixes #3435 

The primary focus of this PR has been to replace a Terminal with Tabulator to be able to filter existing logs. This is done by subclassing `Tabulator` and adding a write function that takes the log message and streams it to the Tabulator data. 

More filters have been added to work with the respective columns. Suggestions for alternative column names are welcome! 

At last, an option for downloading the filtered log file has been added. 

### Example of the new log system:
(I have moved the logs tab to the front locally for easier access during development)
 
<details>
  <summary>Code used </summary>

``` python
# Run with `panel serve tabulator_log.py --admin`

import panel as pn
import param

class Example(param.Parameterized):
    i = param.Number()

    @param.depends("i", watch=True)
    def test(self):
        return 42


e = Example()
pn.Column(e).servable()
```

</details>

https://user-images.githubusercontent.com/19758978/178973411-fc9921bb-4c65-4ada-b361-850a6ef91bf2.mp4

### Problems
Three problems were discovered during the implementation, all somewhat related to streaming data:

1. A rediscover of [#3435](https://github.com/holoviz/panel/issues/3435) of `show_index=False` with streaming data.
2. Calculation of `max_page` or `self.page` with empty data, which was fixed by always having them to be at least 1. 
3. When streaming data and reloading the admin page, it does not show previous logs. This is solved by triggering the `log_terminal.value` in `log_component()`. I don't know if this is by design or a bug.

https://user-images.githubusercontent.com/19758978/178973852-81d932fd-b1f6-4a05-bb69-a04cff281e91.mp4

### Notes:
- As it is (and was before this PR), the filters are synced between multiple users.
- I _think_ that when having `rollover` enabled in streaming, all data will be kept in memory as the rollover here is a view and not a copy. 

https://github.com/holoviz/panel/blob/9ca3a472ca0a590843e6e3d03959b56bcce5098c/panel/widgets/tables.py#L589-L590

- I haven't converted the datetime column to a `datetime` type as this gave some problems when showing the correct value in the Tabulator. 
- I'm not sure if wrapping the write functionality in a try/except block is correct.


panel/io/admin.py
```diff
@@ -75,6 +73,44 @@ def emit(self, record):
...
+
+    def __init__(self, **params):
+        params["value"] = self._create_frame()
+        params["theme"] = "midnight"
+        params["layout"] = "fit_data_stretch"
+        params["show_index"] = False
+        params["sorters"] = [{'field': 'datetime', 'dir': 'dsc'}]
+        params["disabled"] = True
+        params["pagination"] = "local"
+        params["page_size"] = 18
```
philippjfr, line 86: Can we maybe move this onto a class parameter as a single dictionary so someone could override these defaults? Alternatively I suppose the correct thing to do would be to override the actual parameters with new defaults.

panel/widgets/tables.py
```diff
@@ -476,8 +476,11 @@ def _process_column(self, values):
         return values
 
     def _get_data(self) -> Tuple[DataFrameType, DataDict]:
+        return self._get_data_any_df(self.value)
+
+    def _get_data_any_df(self, df: DataFrameType) -> Tuple[DataFrameType, DataDict]:
```
philippjfr, line 481: Not sure I love the name of this method, something like `_convert_df_to_cds_data` seems better but would be happy with other suggestions.

philippjfr:
Had some minor comments, otherwise looks good to me! This will be much more useful than the terminal.

panel/io/admin.py
Hoxbro: Is there a best practice for overriding the defaults?

panel/widgets/tables.py
Hoxbro: Yeah, I agree the name could be (much) better. But maybe make the name even more declarative because the function also filters and aligns multi-index the dataframe, maybe `_process_df_and_convert_to_cds`?

philippjfr: Sounds good!

panel/widgets/tables.py
```diff
@@ -1278,7 +1281,7 @@ def _stream(self, stream, rollover=None, follow=True):
         if self.pagination == 'remote':
             length = self._length
             nrows = self.page_size
-            max_page = length//nrows + bool(length%nrows)
+            max_page = length//nrows + bool(length%nrows) or 1
```
philippjfr, line 1284: Sorry for coming back to this so late? Is the logical grouping here correct? Doesn't quite look like it, because `True or 1` and `False or 1` both evaluate to 1. Should this be:

```suggestion
            max_page = (length//nrows + bool(length%nrows)) or 1
```

panel/widgets/tables.py
```diff
@@ -1291,8 +1294,10 @@ def stream(self, stream_value, rollover=None, reset_index=True, follow=True):
         if follow and self.pagination:
             length = self._length
             nrows = self.page_size
-            self.page = length//nrows + bool(length%nrows)
+            self.page = length//nrows + bool(length%nrows) or 1
```
philippjfr, line 1297: Same here:

```suggestion
            self.page = (length//nrows + bool(length%nrows)) or 1
```

panel/widgets/tables.py
```diff
@@ -1342,7 +1347,7 @@ def _update_selectable(self):
     def _update_max_page(self):
         length = self._length
         nrows = self.page_size
-        max_page = length//nrows + bool(length%nrows)
+        max_page = length//nrows + bool(length%nrows) or 1
```
philippjfr, line 1350: and here:

```suggestion
        max_page = (length//nrows + bool(length%nrows)) or 1
```

philippjfr:
> As it is (and was before this PR), the filters are synced between multiple users.

We should probably fix this separately later.

> I think that when having rollover enabled in streaming, all data will be kept in memory as the rollover here is a view and not a copy.

Yes, that is intended.

> I haven't converted the datetime column to a datetime type as this gave some problems when showing the correct value in the Tabulator.

Can you elaborate? Would definitely be nice.

panel/widgets/tables.py
Hoxbro: I think this is a case of being too fancy. I will replace this with  `max(length//nrows + bool(length%nrows), 1)`, which makes it easier to read and understand. 

For the current implementation, I don't think parenthesis is necessary to get the desired effect, e.g., `1 + 1 or 1` equals 2, which is what we want.

philippjfr: Hmm, not sure I follow, `True/False or 1` will always be equal to 1 so what is the point of even including the `bool(length%nrows)` condition.

Hoxbro: Bad example...

The `+` operator has higher precedence than `or` [ref](https://www.programiz.com/python-programming/precedence-associativity).

image

philippjfr: Thanks for your patience 🤦

Hoxbro:
> Can you elaborate? Would definitely be nice.

The datetime will appear in the logging but only as a string. 

An example of the behavior mentioned can be seen here:

https://user-images.githubusercontent.com/19758978/183616164-9eba7585-0eaa-48ca-8209-946fbe9bbfed.mp4


Also the `.csv` will not show the number:
``` csv
"datetime","level","app","session","message"
"1660044574463","DEBUG","panel.viewable.Location","139986823043728","finished processing events {'pathname': '/3694_admin_tabulator_log', 'href': 'http://localhost:5006/3694_admin_tabulator_log', 'hostname': 'localhost', 'protocol': 'http:', 'port': '5006'}"
"1660044574455","DEBUG","panel.viewable.Location","139986823043728","received events {'pathname': '/3694_admin_tabulator_log', 'href': 'http://localhost:5006/3694_admin_tabulator_log', 'hostname': 'localhost', 'protocol': 'http:', 'port': '5006'}"
"1660044574388","INFO","panel.io.server","139986823043728","rendered"
"1660044574102","INFO","panel.io.server","139986823043728","created"
"1660044574008","INFO","panel.io.server","139986823043728","launching"
```

philippjfr:
Thanks, my question was more about what issues you faced formatting the timestamp as a datetime properly?

Hoxbro:
I did not have any issue converting the string to datetime. The issues were with how Tabulator handled the datetime objects.

Hoxbro:
The picture [here](https://panel.holoviz.org/user_guide/Performance_and_Debugging.html#logs) should be updated with the new interface. 

The only way I can see how this can be done is by replacing it with a new URL, so the current site does not show the new interface.

philippjfr:
>The only way I can see how this can be done is by replacing it with a new URL, so the current site does not show the new interface.

The current site should continue to show it until we release the new version, all you have to do is update examples/assets/admin_logs.png.

PR #3698, opened 2022-07-14
Allow setting the Oauth provider via the environment.

### LINKED ISSUE #3675: PANEL_OAUTH_PROVIDER not used
Panel 0.13.1

Setting PANEL_OAUTH_PROVIDER is ignored, authentication is only used when passing --oauth-provider to panel serve. The authentication user guide suggests otherwise
"the `oauth_provider` can be set on the commandline using the `--oauth-provider` CLI argument to `panel serve` or the `PANEL_OAUTH_PROVIDER` environment variable"

janimo:
Fixes #3675

### DIFF
--- a/panel/command/serve.py
+++ b/panel/command/serve.py
@@ -351,8 +351,14 @@ class Serve(_BkServe):
                 )
             config.nthreads = args.num_threads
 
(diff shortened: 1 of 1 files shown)


philippjfr:
Thanks @janimo!

Issue #3708, opened 2022-07-21
Open server side event management to CustomEvent


keul:
#### Is your feature request related to a problem? Please describe.

The feature request is originated by this question
https://discourse.holoviz.org/t/reactivehtml-data-provided-to-server-side-event-handlers-by-dom-events/3944

Following the documentation for [DOM events](https://panel.holoviz.org/user_guide/Custom_Components.html?highlight=events#dom-events) I find a very powerful tool in panel: a client side event can be captured from a server side bokeh Python method (👍)

But this seems limited to just a pre-defined set of DOM events, any other event type is ignored (👎 ) 

Spelunking in the panel code I found that all of the logic is stored in [event-to-object](https://github.com/holoviz/panel/blob/2e987887d611eb5060226d84bea1af7f9d5d4d82/panel/models/event-to-object.ts).

Every event is checked to be part of the full set of know browsers events. For example: only if I trigger a `mousemove` event I will see my `clientX`attribute passed through. Any other custom data added to the event is discarded.

Is this OK? Maybe…

#### Describe the solution you'd like

The current code provides a full list of well-knows DOM events, but the JavaScript implementation seems to forget the existence of event triggered manually by the client (using [`.dispatchEvent`](https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/dispatchEvent))

I think that, if not _all_ kind of events, at least [CustomEvent](https://developer.mozilla.org/en-US/docs/Web/API/CustomEvent) should be supported. `CustomEvent` works exactly as default `Event`, but handle better data collision by defining a `detail` attribute, to be reserved to custom data.

Brefly speaking: I think that this JavaScript command should be totally legit to be captured by the proper server side Python code by passing through all of the event's data:

```javascript
// this should be captured by a `_mydomnode_myevent` method
mydomnode.dispatchEvent(new CustomEvent('myevent', { detail: { foo: 'bar', baz: [1, 2, 3] } }))
```

This change will greatly improve the integration of external libraries and let the user to build complex JavaScript applications that can freely talk to the server by sending events (e.g integrated complex libraries, like Leaflet, easily).

#### Describe alternatives you've considered

I didn't found any alternative, at least not in Panel (maybe going down at Bokeh level?).
Reading the whole documentation I didn't find an easy way to directly talk to Python

PR #3709, opened 2022-07-21
DOMEvent support for CustomEvent

keul:
See #3708

### DIFF
--- a/panel/models/event-to-object.ts
+++ b/panel/models/event-to-object.ts
@@ -26,7 +26,7 @@ export function serializeEvent(event: any): any {
   const data: any = {};
 
   // support for CustomEvents: the whole `detail` object is serialized
-  if (Object.getPrototypeOf(event).initCustomEvent !== undefined) {
+  if (event.detail !== undefined) {
     Object.assign(data, { detail: event.detail } );
   }
 
--- a/panel/models/event-to-object.ts
+++ b/panel/models/event-to-object.ts
@@ -25,6 +25,11 @@ SOFTWARE.
 export function serializeEvent(event: any): any {
   const data: any = {};
 
+  // support for CustomEvents: the whole `detail` object is serialized
+  if (Object.getPrototypeOf(event).initCustomEvent !== undefined) {
+    Object.assign(data, { detail: event.detail } );
+  }
+
   if (event.type in eventTransforms) {
     Object.assign(data, eventTransforms[event.type](event));
   }
--- a/panel/models/event-to-object.ts
+++ b/panel/models/event-to-object.ts
@@ -27,7 +27,7 @@ export function serializeEvent(event: any): any {
 
   // support for CustomEvents: the whole `detail` object is serialized
   if (event.detail !== undefined) {
-    Object.assign(data, { detail: event.detail } );
+    Object.assign(data, { detail: JSON.parse(JSON.stringify(event.detail)) } );
   }
 
   if (event.type in eventTransforms) {


philippjfr:
This seems completely reasonable to me. My only concern here is that `detail` may contain certain objects which can't be naively serialized to JSON and will therefore error, e.g. `Undefined` objects are one example that I've frequently seen issues with.

panel/models/event-to-object.ts
```diff
@@ -25,6 +25,11 @@ SOFTWARE.
 export function serializeEvent(event: any): any {
   const data: any = {};
 
+  // support for CustomEvents: the whole `detail` object is serialized
+  if (Object.getPrototypeOf(event).initCustomEvent !== undefined) {
```
mattpap, line 29: Any reason to use this specific detection code, instead of `event instanceof CustomEvent`?

keul: @mattpap yep, more simple, but it will be also true for `Event` instance.

keul:
> This seems completely reasonable to me. My only concern here is that `detail` may contain certain objects which can't be naively serialized to JSON and will therefore error, e.g. `Undefined` objects are one example that I've frequently seen issues with.

@philippjfr uops, you are right about this.
Any suggestion on how to improve? JSON.stringify/JSON.parse would fix this but maybe is not that efficient.

panel/models/event-to-object.ts
keul: … I mean: there's nothing wrong on supporting also `Event`, but it's the general design of this module seems to want to limit properties read from an event and I didn't wanted to go against this

panel/models/event-to-object.ts
```diff
@@ -25,6 +25,11 @@ SOFTWARE.
...
+    Object.assign(data, { detail: event.detail } );
```
philippjfr, line 30: Maybe simply add a check if `detail` is undefined to the conditional.

keul: @philippjfr to resume: do you want to remove the whole poor man typechecking and just test for a `detail` attribute?

keul:
> > This seems completely reasonable to me. My only concern here is that `detail` may contain certain objects which can't be naively serialized to JSON and will therefore error, e.g. `Undefined` objects are one example that I've frequently seen issues with.
> 
> @philippjfr uops, you are right about this. Any suggestion on how to improve? JSON.stringify/JSON.parse would fix this but maybe is not that efficient.

@philippjfr I would go for trying to serialize to JSON and abort with a console error in case it fails

panel/models/event-to-object.ts
philippjfr: Probably simpler in the end tbh.

keul: Done!

philippjfr: Okay, and this does what you need?

keul: Yes, with https://github.com/holoviz/panel/pull/3709/commits/2ec39f04d857c98ca4a3dc26d4837dbf6c787325 I'm still able pass through the server serialized data using `.detail` as defined by MDN.

There's still an open issue about what to do if `detail` contains stuff that can't be converted to JSON.

philippjfr:
> @philippjfr I would go for trying to serialize to JSON and abort with a console error in case it fails

This seems reasonable and shouldn't ever be too expensive.

keul:
> > @philippjfr I would go for trying to serialize to JSON and abort with a console error in case it fails
> 
> This seems reasonable and shouldn't ever be too expensive.

Hi @philippjfr

I investigated the issue a little more and I misunderstood something. There's no issue on serializing an object using `JSON.stringify`. For example:

```javascript
JSON.stringify({aaa: 555, ggg: () => {}, bbb: undefined});
'{"aaa":555}'
```

But if I use this same value for `.detail`, a JavaScript error is raised: `[object Function] is not serializable`.
This came from bokeh itself, specifically calling a [to_serializable method](https://github.com/bokeh/bokeh/blob/9c677362d832dafb5c10036fd6fe53ffbb4c258c/bokehjs/src/lib/core/serializer.ts#L89).

Also: not that this serializer module seems a lot different in the [bokeh 3 branch](https://github.com/bokeh/bokeh/blob/fca16442ae90afcd2ac61f4e554e538776730831/bokehjs/src/lib/core/serialization/serializer.ts)

So: the only way to test `detail` for being valid is to import this serializer module from bokeh, but to be honest it seems too much to me.

Currently we already have a JavaScript exception with an error message. I would leave this check to bokeh internal.

philippjfr:
For now let's just roundtrip it through `JSON.stringify` and `JSON.parse` so it drops the unserializable parts.

philippjfr:
Thanks for your patience with this one @keul!

PR #3710, opened 2022-07-22
Added Developing Custom Models to the table of contents

Hoxbro:
You can access the webpage if you know it exists: https://panel.holoviz.org/developer_guide/DevelopingCustomModels.html

Will look like this:
image


jbednar:
Oops! Good catch.

PR #3714, opened 2022-07-23
Added add_periodic_callback coroutine typing

WoLpH:
The `add_periodic_callback` method only had the `Callable` type hint while it also supports asyncio functions. This adds that type hint so `pyright` won't complain anymore :)

### DIFF
--- a/panel/io/state.py
+++ b/panel/io/state.py
@@ -18,8 +18,8 @@ from collections.abc import Iterator
 from contextlib import contextmanager
 from functools import partial
 from typing import (
-    TYPE_CHECKING, Any, Callable, ClassVar, Dict, Iterator as TIterator, List,
-    Optional, Tuple, Union, Coroutine,
+    TYPE_CHECKING, Any, Callable, ClassVar, Coroutine, Dict,
+    Iterator as TIterator, List, Optional, Tuple, Union,
 )
 from urllib.parse import urljoin
 from weakref import WeakKeyDictionary
--- a/panel/io/state.py
+++ b/panel/io/state.py
@@ -19,7 +19,7 @@ from contextlib import contextmanager
 from functools import partial
 from typing import (
     TYPE_CHECKING, Any, Callable, ClassVar, Dict, Iterator as TIterator, List,
-    Optional, Tuple, Union,
+    Optional, Tuple, Union, Coroutine,
 )
 from urllib.parse import urljoin
 from weakref import WeakKeyDictionary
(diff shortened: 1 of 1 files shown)


panel/io/state.py
```diff
@@ -379,8 +379,9 @@ def as_cached(self, key: str, fn: Callable[[], None], ttl: int = None, **kwargs)
         return ret
 
     def add_periodic_callback(
-        self, callback: Callable[[], None], period: int=500,
-        count: Optional[int] = None, timeout: int = None, start: bool=True
+        self, callback: Union[Callable[[], None], Coroutine[Any, Any, None]],
```
maximlt, line 382: I believe the preferred syntax for listing types in the code base is `|` instead of `Union`. Is that right @philippjfr ?

```suggestion
        self, callback: Callable[[], None] | Coroutine[Any, Any, None],
```

maximlt (commented):
Thanks @WoLpH! Just a minor suggestion.

panel/io/state.py
philippjfr: No `|` is preferred.

WoLpH: I didn't go for the pipe because I noticed that Python 3.6 is still supported in the code base. As far as I know that feature has been backported to 3.7 and above.

But I prefer it as well, 3.6 is EOL anyhow.

maximlt: In practice Panel has dropped support of Python 3.6 as it relies on Bokeh >= 2.4.0 which itself no longer supports it. The setup.py file still has to be updated to reflect that.

maximlt: @WoLpH the test suite is failing as `Coroutine` is not imported. Would you like to fix that?

WoLpH: Oops... not sure how that was not committed but I only have that change locally. I've added it and updated using the `|` instead of `Union`

panel/io/state.py
```diff
@@ -19,7 +19,7 @@
 from functools import partial
 from typing import (
     TYPE_CHECKING, Any, Callable, ClassVar, Dict, Iterator as TIterator, List,
```
philippjfr, line 21:
```suggestion
    TYPE_CHECKING, Any, Callable, ClassVar, Coroutine, Dict,
```

panel/io/state.py
```diff
@@ -19,7 +19,7 @@
...
-    Optional, Tuple, Union,
+    Optional, Tuple, Union, Coroutine,
```
philippjfr, line 22:
```suggestion
    Iterator as TIterator, List, Optional, Tuple, Union,
```

philippjfr:
Test failures are unrelated, merging.

philippjfr:
Thanks @WoLpH!

Issue #3628, opened 2022-06-16
New warning in panel 0.13.1 –param function


mayonnaisecolouredbenz7:
Hi, I have been using panel to build a large scale multi page app. To achieve the multi page aspect I have been using tabs combined with param functions so as to defer the load of the page until the tab is selected, as is showing in an example in the documentation page. [Tabs — Panel v0.13.1 1](https://panel.holoviz.org/reference/layouts/Tabs.html)

After updating to 0.13.1 I get the following warning

> WARNING:param.Function 1: The function ‘func1’ does not have any dependencies and will never update. Are you sure you did not intend to depend on or bind a parameter or widget to this function? If not simply call the function before passing it to Panel. Otherwise, when passing a parameter as an argument, ensure you pass at least one parameter and reference the actual parameter object not the current value, i.e. use object.param.parameter not object.parameter.

Here is a MRE:
![image](https://user-images.githubusercontent.com/38597661/173980464-5e6b8595-1f80-4f97-9c59-65857eea1040.png)

Interestingly this warning is now reflected in the Tabs documentation, but not explained.

This issue was first presented by myself on the holoviz discourse.  Was responded to and now opened up as an issue here.

philippjfr:
Thanks for reporting this, this is a usage of `ParamFunction` I had not considered when adding that warning but it does make sense of course. I'll think about it more and come up with a solution.

PR #3715, opened 2022-07-24
No warning when ParamFunction is lazy

maximlt:
Fixes https://github.com/holoviz/panel/issues/3628

In this PR I've simply skipped emitting a warning when the ParamFunction is lazy. Note that https://github.com/holoviz/panel/commit/3c59d4ddae9830f697455b8d5349fa1a01f970dc is unrelated, it is touching code that is close to what I changed in the PR and that seemed useless given the signature of the method `def _replace_pane(self, *args, force=False)`, whereby `*args` isn't used in the method and would swallow the passed argument I removed.

This change doesn't fix the spurious warning observed in https://github.com/holoviz/hvplot/issues/770. It happens there because of the usage of `pn.panel(self._callback)`

https://github.com/holoviz/hvplot/blob/3f0f4596228f92a8951dc0459ade91e830e62c02/hvplot/interactive.py#L495

Where `self._callback` is a property returning a function decorated with `pn.depends`.

https://github.com/holoviz/hvplot/blob/3f0f4596228f92a8951dc0459ade91e830e62c02/hvplot/interactive.py#L140-L143

`pn.panel()` in this case falls back to the `ParamFunction` pane and as in some cases the `pn.depends` decorator is given an empty list then in those cases it is similar to running `pn.panel(pn.bind(foo))`, for which a warning is expected. To solve that at hvPlot's level I *think* it would be fine to always call `pn.panel` with `lazy=True` in:

https://github.com/holoviz/hvplot/blob/3f0f4596228f92a8951dc0459ade91e830e62c02/hvplot/interactive.py#L495

But I'd need this to be confirmed and tested.

If this warning causes too much trouble, then we may just have to revert it (even if I think it was a nice addition!).

### DIFF
--- a/panel/param.py
+++ b/panel/param.py
@@ -758,7 +758,7 @@ class ParamMethod(ReplacementPane):
         self._link_object_params()
         if object is not None:
             self._validate_object()
-            self._replace_pane(not self.lazy)
+            self._replace_pane()
 
     @param.depends('object', watch=True)
     def _validate_object(self):
--- a/panel/param.py
+++ b/panel/param.py
@@ -901,7 +901,7 @@ class ParamFunction(ParamMethod):
     def _link_object_params(self):
         deps = getattr(self.object, '_dinfo', {})
         dep_params = list(deps.get('dependencies', [])) + list(deps.get('kw', {}).values())
-        if not dep_params:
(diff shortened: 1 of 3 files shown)


philippjfr:
Thanks @maximlt!

PR #3716, opened 2022-07-25
Large pdf improvements

### LINKED ISSUE #3696: PDF Pane not always rendering the document
Local large PDF files cannot all be rendered, and even files of moderate size. The implementation converts the PDF to a base64-string, and uses the HTML `embed` tag whose `src` is assigned a data URL built from the base64-string. When that string is over a limit, that depends on the browser in use, the PDF pane isn't rendered at all and there's no error logged.

I'm sharing this 2Mb PDF file that cannot be rendered on Chrome when served locally but can be rendered on Firefox.

[pdfsample50.pdf](https://github.com/holoviz/panel/files/9115028/pdfsample50.pdf)

There are two posts on Discourse that seem related to this issue:
* https://discourse.holoviz.org/t/pdf-pane-pn-pane-pdf/3640/14
* https://discourse.holoviz.org/t/issue-on-displaying-pdfs/3884

Hoxbro:
Fixes #3696 

The method used to be able to render large PDFs is the one described [here](https://discourse.holoviz.org/t/pdf-pane-pn-pane-pdf/3640/14). Which converts the base64 to a blob and inserted as an URL with `URL.createObjectURL(blob);` in the Typescript Bokeh model.  

I have also added `start_page` to be able to set the start page when loading the PDF.

The pane is located in `image.py` and not in `markup.py` because it is a subclass of `FileBase`, which is located in `Image.py`. 

# Todo
- [ ] Confirm that this is the method to go forward with.
- [x] Write unit tests for this method.
- [x] Update existing unit tests to the method.
- [ ] Maybe add other URL parameters `navpanes=0&scrollbar=0&toolbar=0&zoom=100`?
- [ ] If there is some way to cache the data of `convert_base64_to_blob`.


panel/models/pdf.ts
```diff
@@ -0,0 +1,79 @@
...
+import { htmlDecode } from "./html";
+
+export class PDFView extends PanelMarkupView {
+  model: PDF;
+
+  connect_signals(): void {
+    super.connect_signals();
+    const p = this.model.properties;
+    const { text, width, height, embed, start_page } = p;
+    this.on_change([text, width, height, embed, start_page], () => {
```
philippjfr, line 13: Do you think we can somehow control the page independently of the embedded content to avoid re-rendering the PDF from scratch each time?

Hoxbro: I don't think it is possible with the current implementation. This is why I called it `start_page` and not `page`. If we want full control over the PDF, we need to change the implementation to use `PDF.js`.

examples/reference/panes/PDF.ipynb
```diff
@@ -23,7 +23,8 @@
     "\n",
     "* **``embed``** (boolean, default=False): If given a URL to a file this determines whether the image will be embedded as base64 or merely linked to.\n",
     "* **``object``** (str or object): The PDF file to display. Can be a string pointing to a local or remote file, or an object with a ``_repr_pdf_`` method.\n",
-    "* **``style``** (dict): Dictionary specifying CSS styles\n",
+    "* **``style``** (dict): Dictionary specifying CSS styles.\n",
+    "* **``start_page``** (int): Start page of the `.pdf` file when loading the page. \n",
```
philippjfr, line 27:
```suggestion
    "* **``start_page``** (int): Start page of the `.pdf` file when loading the page. \n",
    "* **``style``** (dict): Dictionary specifying CSS styles.\n",
```

philippjfr (approved):
Sorry I didn't merge this one earlier. One balance I think this is an improvement, so I'm going to merge.

PR #3721, opened 2022-07-27
Tabulator errors with selectable_rows and nonallowed selection

### LINKED ISSUE #3644: Tabulator: Setting `selection` to one of the non allowed rows by `selectable_rows` should raise an error
As per the title and the example below:

```python
from pandas._testing import makeMixedDataFrame

df = makeMixedDataFrame()

table = pn.widgets.Tabulator(df, selectable_rows=lambda df: [0])
# This should raise an error
table.selection = [1]
```

hoxbro:
Fixes #3644

I could only stop values that were not allowed by `selectable_rows` by using a subclass of `param.List`. 

If I used something like `_process_param_change` raised the error but the parameter `Tabulator.selection` will have the wrong value.


panel/widgets/tables.py
```diff
@@ -896,6 +896,29 @@ def _update_aggregators(self, model):
             g.aggregators = self._get_aggregators(index)
 
 
+class _ListValidateWithCallable(param.List):
```
philippjfr, line 899: If we are defining custom parameters (which we should do much more of) I think we should put them in a shared place.

Hoxbro: The logical place will be `panel/param.py`, but the file could get pretty large if a lot of parameters are added.  

A small thing is the current error message is tailored to make sense with `Tabulator` only.

maximlt: I'd love for `param` to provide a way to add validators that'd be called before the new value is set.

philippjfr: Agreed, I made an issue for that years ago https://github.com/holoviz/param/issues/324

Coderambling:
Is this one fixed or still an issue?

philippjfr (approved):
In this case, `tables.py` is probably the correct place for the parameter.

PR #3722, opened 2022-07-27
Fix flaky tests

maximlt:
There are a few flaky tests in the test suite, attempting to fix those in this PR.


philippjfr:
My vote is to use flaky (https://pypi.org/project/flaky/) to re-run tests a maximum of 3 times and it should only ever be used on UI tests.

philippjfr:
Yay, thanks!

PR #3729, opened 2022-07-30
Improve docstring and type annotation of as_cached

MarcSkovMadsen:
I'm trying to help a user on discourse. I wanted to use `as_cached` for the first time. 

Its just much faster (for me) to look at an example before I study the more systematic documentation. So I added an example.

I added a better type annotation than return type `None` as described here https://stackoverflow.com/questions/62153257/specify-type-annotation-for-python-function-returning-type-of-argument

## Comments

- Why did you pick the name `as_cached`. Its not a term I know from elsewhere. It does not give me associations to anything I already know and love which would have been helpful 😄 .

### DIFF
--- a/panel/io/state.py
+++ b/panel/io/state.py
@@ -19,7 +19,7 @@ from contextlib import contextmanager
 from functools import partial
 from typing import (
     TYPE_CHECKING, Any, Callable, ClassVar, Coroutine, Dict,
-    Iterator as TIterator, List, Optional, Tuple, Union,
+    Iterator as TIterator, List, Optional, Tuple, TypeVar, Union, 
 )
 from urllib.parse import urljoin
 from weakref import WeakKeyDictionary
@@ -51,6 +51,8 @@ if TYPE_CHECKING:
     from .location import Location
     from .notifications import NotificationArea
     from .server import StoppableThread
+
+    T = TypeVar("T")
 
 
 @contextmanager
@@ -333,7 +335,7 @@ class _state(param.Parameterized):
     # Public Methods
     #----------------------------------------------------------------
 
-    def as_cached(self, key: str, fn: Callable[[], None], ttl: int = None, **kwargs) -> None:
+    def as_cached(self, key: str, fn: Callable[[], T], ttl: int = None, **kwargs) -> T:
         """
         Caches the return value of a function, memoizing on the given
         key and supplied keyword arguments.
--- a/panel/io/state.py
+++ b/panel/io/state.py
@@ -340,6 +340,11 @@ class _state(param.Parameterized):
 
         Note: Keyword arguments must be hashable.
 
+        Example:
+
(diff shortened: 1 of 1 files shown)


MarcSkovMadsen:
I looked at the failing tests. They don't seem related to this PR as far as I can see.

panel/io/state.py
```diff
@@ -333,13 +335,18 @@ async def _scheduled_cb(self, name: str) -> None:
...
         """
         Caches the return value of a function, memoizing on the given
         key and supplied keyword arguments.
 
         Note: Keyword arguments must be hashable.
 
+        Example:
+
+        >>> import seaborn as sns
+        >>> planets = pn.state.as_cached("seaborn-planets", sns.load_dataset, name='planets')
```
maximlt, line 348: I didn't find the example very intuitive, in particular without knowing what `sns.load_dataset` does and what signature it has. I've come up with this suggestion, feel free to adapt it more if you think it can be improve, or to reject it entirely if you don't like it!

```suggestion
        >>> def load_dataset(name):
        >>>     # some slow operation that uses name to load a dataset....
        >>>     return dataset
        >>> penguins = pn.state.as_cached('dataset-penguins', load_dataset, name='penguins')
```

maximlt (commented):
Thanks Marc!

In particular on the typing improvement, TIL!

I made a suggestion to improve the example.

> Why did you pick the name as_cached. Its not a term I know from elsewhere. It does not give me associations to anything I already know and love which would have been helpful 😄 .

Naming is hard! I can't reply for Philipp, but if you have other names to suggest I guess it's never too late, in particular if they're a net improvement over the current state.

> I looked at the failing tests. They don't seem related to this PR as far as I can see.

They seem related actually, pre-commit failed because of a trailing whitespace: https://github.com/holoviz/panel/runs/7589741660?check_suite_focus=true#step:5:79

panel/io/state.py
```diff
@@ -333,13 +335,18 @@ async def _scheduled_cb(self, name: str) -> None:
     # Public Methods
     #----------------------------------------------------------------
 
-    def as_cached(self, key: str, fn: Callable[[], None], ttl: int = None, **kwargs) -> None:
+    def as_cached(self, key: str, fn: Callable[[], T], ttl: int = None, **kwargs) -> T:
```
philippjfr, line 338:
```suggestion
    def as_cached(self, key: str, fn: Callable[[], T], ttl: int = None, **kwargs) -> T:
```

Can you explain the use of `T`? Shouldn't it simply be `Any`?

MarcSkovMadsen: Nope :-) I want the editor to understand that `as_cached` return the same value (`T`) as the callable `fn` - not Any as I then will not be able to get any help from my editor from that point.

philippjfr: Ah got it.

PR #3737, opened 2022-08-02
Add extension entry point

jmosbacher:
This is just a draft to test the waters on whether maintainers would be open to this, if so can polish/alter it as requested.

The PR adds an entry point group named `panel.extension` and loads them when `panel.extension()` is called.
A flag named `load_entry_points` is added to the global config object which can disable the loading in cases where a user wishes to disable external plugin functionality. This can be generalized to perhaps a list of plugins to ignore when loading entry points, depending on what the maintainers think.

This can be useful in various scenarios:

- packages wishing to extend panel functionality but want to register this functionality in scenarios when `panel.extension()` is called. e.g. in a notebook
- packages which offer a panel based GUI as optional, if panel is installed. In these cases we can avoid many `try-except ImportError` blocks by simply putting all the GUI logic in a module loaded only by panel so its never imported unless panel is installed. For example registering a `__repr_mimebundle__` method on classes only when `pn.extension()` was called.
- packages that monkey patch panel, avoids needing to import the package just from the patching logic to run.

Addresses issue #3738


panel/entry_points.py
```diff
@@ -0,0 +1,51 @@
+""" This module defines a function for iterating over entry points
```
philippjfr, line 1: The root directory of the panel package is getting a little crowded. I think it's time to bite the bullet and add a subpackage for some stuff, not quite sure what to sensibly group in there yet though. I'll consider and get back to you.

panel/config.py
```diff
@@ -144,6 +144,9 @@ class _config(_base_config):
     throttled = param.Boolean(default=False, doc="""
         If sliders and inputs should be throttled until release of mouse.""")
 
+    load_entry_points = param.Boolean(default=True, doc="""
```
philippjfr, line 147: Could we define the parameter in alphabetical order?

philippjfr (changes requested):
Overall this looks good to me, had some minor comments and have to consider how to reorganize the package a bit but that can happen in another PR. One other thing is that it would be nice to document this in the developer docs. Could you create an `doc/developer_guide/extensions.md` and document this? I'll probably follow up in another PR and add some more detail around other ways of extending Panel.

panel/config.py
jmosbacher: np, done.

jmosbacher:
Thanks for being open to this, I think it would be a really nice way for me to register the Pydantic pane from `pydantic-panel` package. 
I added some documentation for entrypoints but would appreciate if you checked its correctness. I never used the `panel.auth` or `panel.io.rest` ones so I just wrote what i understood from the code, may have misunderstood their exact usage.

When you decide on where to put this module just let me know and ill move it there and mark this ready fro review.

doc/developer_guide/extensions.md
```diff
@@ -0,0 +1,29 @@
...
+
+ - panel.auth
+ - panel.io.rest
+ - panel.extension
+
+### Authentication plugins
+The `panel.auth` group should be used to register new authentication providers. The entrypoint should resolve to a subclass of `tornado.web.RequestHandler` that can handle auhtentication requests. The handler can be enabled for a given app by setting the `auth` config to to the name of the entrypoint.
+
+### Rest plugins
+The `panel.io.rest` entrypoit group can be used to register rest providers that will be served along the panel app. The entrypoint should resolve to a callable with the following signature:
```
philippjfr, line 14:
```suggestion
### Rest plugins

The `panel.io.rest` entrypoint group can be used to register rest providers that will be served along the panel app. The entrypoint should resolve to a callable with the following signature:
```

doc/developer_guide/extensions.md
```diff
@@ -0,0 +1,29 @@
...
+
+## Plugin support
+Panel supports a number of entrypoint groups which can be used by external packages to extend its functionality.
+
+ - panel.auth
+ - panel.io.rest
+ - panel.extension
+
+### Authentication plugins
+The `panel.auth` group should be used to register new authentication providers. The entrypoint should resolve to a subclass of `tornado.web.RequestHandler` that can handle auhtentication requests. The handler can be enabled for a given app by setting the `auth` config to to the name of the entrypoint.
```
philippjfr, line 11:
```suggestion
### Authentication plugins

The `panel.auth` group should be used to register new authentication providers. The entrypoint should resolve to a subclass of `tornado.web.RequestHandler` that can handle authentication requests. The handler can be enabled for a given app by setting the `auth` config to to the name of the entrypoint.
```

philippjfr:
Test failures are unrelated, the docs look like a good starting point and I'll take care of reorganizing the root directory. Really appreciate your work here @jmosbacher. Merging!

PR #3739, opened 2022-08-03
Add hard bounds to editable sliders

### LINKED ISSUE #2735: Location sync do not respect start and end of widget
panel version 1.12.1

I have a widget which I want to sync with `pn.state.location.sync`. The problem is I can exceed the limits of the slider with the url, I expected that it will not exceed the start and end of the widget. 

``` python
import panel as pn

widget = pn.widgets.IntSlider(name="Slider", start=0, end=10)
pn.state.location.sync(widget, {"value": "value"})
widget.servable()

```
image

### LINKED ISSUE #3637: Editable slider widgets should have hard bounds
Currently the editable slider widget have one regular set of bounds defined as `start` and `end` which define the limits of the slider. The text input does not have any bounds applied to it however which is problematic because you may want to define hard bounds.

Hoxbro:
Fixes #3637 

The way to set the hard bounds is by using `fixed_start` and `fixed_end`, which are then set on the `value` parameter as bounds and `start` and `end` of the text input widgets. The reason for using two parameters and not one is to mimic how the widget `start` and `end` work.

I tried to set the `bounds` of the `value` parameter of the non-editable sliders to fix the [#2735](https://github.com/holoviz/panel/issues/2735), but it broke a lot of the tests, so I removed it again (for now).

The removal of `self._slider.jscallback` is because it didn't seem to do anything, but maybe it is to avoid multiple synchronizations between the bokeh model and the param class. I will add it back again with a comment if needed.

The last thing is on the `EditableRangeSlider`, I set the start value of the text input always to be smaller than the end value and vice versa.

### DIFF
--- a/panel/param.py
+++ b/panel/param.py
@@ -456,6 +456,10 @@ class Param(PaneBase):
                         widget_class = LiteralInput
             if hasattr(widget_class, 'step') and getattr(p_obj, 'step', None):
                 kw['step'] = p_obj.step
+            if hasattr(widget_class, 'fixed_start') and getattr(p_obj, 'bounds', None):
+                kw['fixed_start'] = p_obj.bounds[0]
+            if hasattr(widget_class, 'fixed_end') and getattr(p_obj, 'bounds', None):
+                kw['fixed_end'] = p_obj.bounds[1]
 
         kwargs = {k: v for k, v in kw.items() if k in widget_class.param}
 
--- a/panel/tests/ui/widgets/test_sliders.py
+++ b/panel/tests/ui/widgets/test_sliders.py
@@ -316,7 +316,7 @@ def test_editablerangeslider_button_start(page, port, widget):
 
 
 def test_editablerangeslider_no_overlap(page, port):
-    widget = EditableRangeSlider(value=(0, 1), step=1)
+    widget = EditableRangeSlider(value=(0, 2), step=1)
 
     serve(widget, port=port, threaded=True, show=False)
     time.sleep(0.2)
@@ -324,11 +324,14 @@ def test_editablerangeslider_no_overlap(page, port):
     page.goto(f"http://localhost:{port}")
 
     up_start = page.locator("button").nth(0)
+    down_start = page.locator("button").nth(1)
     down_end = page.locator("button").nth(3)
 
     up_start.click(click_count=3)
-    wait_until(page, lambda: widget.value == (1, 1))
+    wait_until(page, lambda: widget.value == (2, 2))
 
-    widget.value = (0, 1)
+    down_start.click()
+    wait_until(page, lambda: widget.value == (1, 2))
+
     down_end.click(click_count=3)
-    wait_until(page, lambda: widget.value == (0, 0))
+    wait_until(page, lambda: widget.value == (1, 1))
--- a/panel/tests/ui/widgets/test_sliders.py
+++ b/panel/tests/ui/widgets/test_sliders.py
@@ -10,6 +10,7 @@ except ImportError:
 pytestmark = pytest.mark.ui
 
 from panel.io.server import serve
+from panel.tests.util import wait_until
 from panel.widgets import (
     EditableFloatSlider, EditableIntSlider, EditableRangeSlider,
 )
@@ -57,21 +58,18 @@ def test_editableslider_textinput_end(page, port, widget, val1, val2, val3, func
     text_input = _editable_text_input(page)
 
     text_input.value = val1
-    page.wait_for_timeout(200)
-    assert widget.value == val1
+    wait_until(page, lambda: widget.value == val1)
     assert widget._slider.end == val1
     assert func(text_input.value) == val1
 
     text_input.value = val2
-    page.wait_for_timeout(200)
-    assert widget.value == val2
+    wait_until(page, lambda: widget.value == val2)
     assert widget._slider.end == val1
     assert func(text_input.value) == val2
 
     # Setting fixed end
     widget.fixed_end = val3
-    page.wait_for_timeout(200)
-    assert widget.value == val3
+    wait_until(page, lambda: widget.value == val3)
     assert widget._slider.end == val3
     assert func(text_input.value) == val3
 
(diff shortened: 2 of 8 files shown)


philippjfr:
Looking good to me! Not entirely sure `fixed_start` and `fixed_end` convey the concept of a hard bound but don't love `hard_start` and `hard_end` either. Any other suggestions?

Separately I think this PR should also update the `Param` pane to map the `softbounds` and `bounds` to these new parameters on the editable sliders.

Hoxbro:
I don't favor `fixed_*` either but also liked it better than `hard_*`. 

Some suggestions for another name:  pinned, immovable, locked, strict, firm.

philippjfr:
Maybe let's move away from the "hard" connotation and instead describe what it actually bounds, e.g. `editor_start` and `editor_end`? Ping @maximlt and @jbednar for votes/suggestions.

jbednar:
I think "hard bounds" is self-explanatory; that's something people say in the real world. But "hard_end" doesn't really work. It can't be a tuple to avoid that problem?

I don't think "editor" is the right concept, certainly not at the Param level but it also seems odd to me at the Panel level. "Soft bounds" are really "suggested bounds", i.e., they are a weak suggestion of a useful range, a hint that doesn't specifically have to do with editors or widgets, but a declaration that the author of this app thought that this range was useful or the best starting point.  Such a suggestion maps nicely to the initial bounds of a widget, but that's not its only interpretation.

philippjfr:
For compatibility with other sliders the `start` and `end` values will continue to determine the slider bounds. Adding additional parameters to control the minimum and maximum values that can be entered in the editor/spinner box is what we are discussing here.

How that maps onto the `bounds` and `softbounds` is entirely up to us but also a separate issue from naming the parameters that determine the hard bounds on the widget.

panel/param.py
```diff
@@ -456,6 +456,10 @@ def widget(self, p_name):
                         widget_class = LiteralInput
             if hasattr(widget_class, 'step') and getattr(p_obj, 'step', None):
                 kw['step'] = p_obj.step
+            if hasattr(widget_class, 'fixed_start') and getattr(p_obj, 'bounds', None):
```
philippjfr, line 459: I'll have to consider this logic more deeply, I don't think this is actually sufficient since it's the `softbounds` that we want to apply.

maximlt:
I like better `hard_*` but that's maybe just because I got used to that working with Param.

One note, the docs will have to be updated, this is e.g. the current definition of the `EditableFloatSlider` that doesn't say that the editor can be used to set a value outside of the soft-bounds offered by the slider.

> The EditableFloatSlider widget allows selecting selecting a numeric floating-point value within a set bounds using a slider and for more precise control offers an editable number input box.

By the way are we okay with the fact that if `hard/fixed/..._*` isn't set then that does mean that by default there's **no hard bound**? Not entirely sure but I would guess that in most cases people actually want to have hard bounds when they use sliders in their app. They would have to write `pn.widgets.Editable...(start=1, hard_start=1, end=2, hard_end=2)` to get that. I see the editable slider as a way to provide users an easy way to enter a specific value, more than a way to enter a value outside of the slider bounds.

(That's another issue but the Input number widgets don't seem to have a way to indicate their bounds)

panel/param.py
Hoxbro: I'm not sure I understand what you are meaning. The `softbounds` functionality for the slider is already implemented, and the `hard/fixed/..._*` should be associated with the hard `bounds`, right?

https://user-images.githubusercontent.com/19758978/183444918-4cbb8465-6954-49ca-a787-22d4124673bc.mp4

``` python
import panel as pn
import param
import pytest

pn.extension()

    i = param.Number(default=1, softbounds=(1, 5), bounds=(0, 10), step=1)


t = Test()
w = pn.widgets.EditableFloatSlider.from_param(t.param.i)
w
```

--- 
Unrelated to this PR, I noted that `bounds` / `softbounds` have priority over `start`/`end`. Is this what is expected? 

image

jbednar: Yes, putting in a value outside the soft bounds but inside the hard bounds is precisely how you can make use of the difference between soft and hard bounds.

maximlt: That does mean that it's not possible to reset the soft bounds to their defaults.

jbednar: Yes. But still better than being bound inappropriately by soft bounds, as one can usually reload the app, but can never hack around falsely hard bounds.

jbednar: See https://github.com/bokeh/bokeh/issues/9943 for my rant about this.

jbednar: Oh, I see you chimed in there before even joining our group. :-)

jbednar:
> By the way are we okay with the fact that if hard/fixed/..._* isn't set then that does mean that by default there's no hard bound? Not entirely sure but I would guess that in most cases people actually want to have hard bounds when they use sliders in their app. 

I would have assumed that any previously supported single bounds would have to be hard bounds, and that any new set of bounds are optional soft bounds that override the previous bounds if present. Isn't that the only way the system can behave reasonably?

philippjfr:
I'm actually okay with `fixed_start` and `fixed_end` now since the semantics are that start/end can be overridden by manually entering a value but the `fixed_` bounds cannot. Just want to see updates to the docs for the various `Editable` widgets and then I'm very happy with this PR. Thanks @Hoxbro!

Hoxbro:
When updating the documentation, I updated a small bug, which meant that the `start`/`end` of the `_slider` could exceed the `fixed_start` and `fixed_end`, as the following video show. This should now be fixed.

https://user-images.githubusercontent.com/19758978/183668197-70299768-17b2-44d3-b610-c17da86fc95f.mp4

---

The `fixed_start` and `fixed_end` of EditableRangeSlider were set to `param.Integer` which was a copy/paste error...

panel/tests/ui/widgets/test_sliders.py
```diff
@@ -0,0 +1,358 @@
...
+    serve(widget, port=port, threaded=True, show=False)
+    time.sleep(0.2)
+
+    page.goto(f"http://localhost:{port}")
+
+    text_input = _editable_text_input(page)
+
+    text_input.value = val1
+    page.wait_for_timeout(200)
+    assert widget.value == val1
```
maximlt, line 61: The `wait_until` utility wasn't working?

Hoxbro: A classic case of `I forgot it`...

philippjfr:
Looks like one of the tests is flakey. Can you see if you can fix it? Otherwise use the flaky marker with max_runs=3.

There also seems to be an unrelated VTK issue on Windows.

philippjfr:
Thanks, merging (py 3.9 failure still unrelated).

Issue #3688, opened 2022-07-12
Regression: Names of Parameterized options not used as labels of Selectors


MarcSkovMadsen:
Panel 0.13.1

```python
import panel as pn

component1=pn.panel("Value1")
component2=pn.panel("Value2")

selection = pn.widgets.RadioBoxGroup(value=component1, options=[component1, component2])

pn.Column(
    selection
).servable()
```

shows

![image](https://user-images.githubusercontent.com/42288570/178573885-12e5afef-ecda-4332-a970-fe6c17ef26b8.png)

I would expect "Value1" and "Value2" to be shown as labels. It used to be like that and it was simple and powerful.

maximlt:
Even if it worked before, I'm not sure it should actually have, as I find this usage as shown in the example pretty confusing. Do you have a more specific use case in mind that would motivate supporting that?

philippjfr:
I can't think of any mechanism that would have allowed this to work in the past tbh. I'm -0 on supporting this, I think it's not unreasonable to support but it seems of very limited utility. The `.object` of a pane is very rarely something that would make sense as an option label (e.g. what happens for non-string types and what happens if the string is very long). It does seem better to be explicit about the actual label.

philippjfr:
Also just tested all Panel versions in the 0.11.x, 0.12.x as well as 0.13.0 and this has never worked as far as I can tell. Therefore I'm going to close as invalid.

philippjfr:
The reason why I also think this is wrong is that the `.object` of a pane is mutable. If you update it and the label does not also update it is very confusing.

jbednar:
Was this maybe the intent of the original post?

```python
import panel as pn
pn.extension()

component1=pn.panel("Value1", name="V1")
component2=pn.panel("Value2", name="V2")

selection = pn.widgets.RadioBoxGroup(value=component1, options=[component1, component2])

pn.Column(
    selection
).servable()
```

In that case I would have expected `name` to be used as the option label, but it isn't:

<img width="247" alt="image" src="https://user-images.githubusercontent.com/1695496/180331530-d0419461-e167-4ecc-a2d6-ce6aeaa570de.png">

jbednar:
Sure, but it's worth checking if it used to work this way...

MarcSkovMadsen:
I've checked panel==0.10.0/ param==1.9.3 with this example

```python
import param
import panel as pn

class MyClass(param.Parameterized):
    value = param.Number()

options = [
[...]

PR #3755, opened 2022-08-13
Use non-default Parameterized name as Select label

philippjfr:
Fixes https://github.com/holoviz/panel/issues/3688


panel/util.py
```diff
@@ -37,6 +37,9 @@
 bokeh_version = Version(bokeh.__version__)
 
 
+PARAM_NAME_PATTERN = re.compile('^.*\d{5}$')
```
jbednar, line 40: This seems like something that should be internal to Param (a method or function to determine if a name is non-default), not hardcoded into Panel.

Issue #3436, opened 2022-04-20
Render With Panel work with Different Kernels/Conda Environments in JupyterLab


iwalmsley:
Hi there, thanks very much for the software, very much enjoying using it.

#### Is your feature request related to a problem? Please describe.
In my team we make heavy use of Conda environments, hence although we have a shared JupyterLab image which includes panel, I might be working in a Conda environment with contains a number of custom packages, however when I select "Render with Panel", this uses the base environment of Jupyter, hence fails to load, e.g;

```
import intake
import panel as pn

pn.extension()
col = pn.Column(
  '#Title',
)
col.servable()
```

Which works fine when I run the cell gives the following when I select render;

```
ModuleNotFoundError: No module named 'intake'
Traceback (most recent call last):
  File "/opt/conda/lib/python3.9/site-packages/bokeh/application/handlers/code_runner.py", line 231, in run
    exec(self._code, module.__dict__)
  File "/data/notebooks/jupyterlab-juliatest/wrf-panel-app/Untitled.ipynb", line 7, in <module>
    "metadata": {},
ModuleNotFoundError: No module named 'intake'
```

This is frustrating as although I can quite easily run the cells and simulate using it, it's much nicer to render the entire app in a separate panel.

#### Describe the solution you'd like
It'd be great if the Conda environment active in the current Kernel was used to render the panel app. I do understand that panel itself would need to be installed in the environment as well which does make things a bit confusing.

philippjfr:
Thanks for the issue, I do not believe this to be possible to be honest. The "Preview" component works using a jupyter serverextension which by design has to run in the same environment as the Jupyter server.

iwalmsley:
Thanks a lot for the response, closing the request as not possible!

tlvu:
I have a reproducible runtime environment on Binder here https://mybinder.org/v2/gh/Ouranosinc/PAVICS-e2e-workflow-tests/47ef7b0c668acc84cb9c1c7cedc2f135aa54dd3a.

Very similar problem, the green icon "Render with Panel" do not work with the notebook `TEST-panel-servable.ipynb` that is absolutely trivial:
```python
import panel as pn
pn.panel('# Ceci est un test').servable()
```

That same trivial notebook work fine under the Panel launcher (https://github.com/holoviz/jupyter-panel-proxy), try it on my Binder link above.

tlvu:
"Render with Panel" green button works fine for me know.  Updating here in case this helps someone.

[...]

PR #3763, opened 2022-08-17
Rewrite server extension to run Panel applications in kernels

philippjfr:
This PR reimplements the panel preview server extension in its entirety to execute Panel applications in custom kernels. It does this by querying the notebook for its kernel spec and then requesting a kernel from Jupyter. In this kernel it executes the following:

```python
from panel.io.jupyter_server_extension import PanelExecutor
executor = PanelExecutor('{{ path }}', '{{ token }}', '{{ root_url }}')
executor.render()
```

Internally the `PanelExecutor` creates a bokeh `ServerSession` and connects it to a Jupyter `Comm`. The `PanelHandler` will then serve the result of `executor.render()` and a request to open a WebSocket will be sent to the `PanelWSHandler`. This in turns forwards any messages it receives via the `kernel.shell_channel` to the `Comm`. This way we proxy any WS messages being sent to and from the server to the kernel.

The major benefit of this is:

- Panel preview now actually uses the appropriate kernel (and therefore environment) that it was intended for.

Drawbacks:

- Spinning up a kernel per session is expensive (we may be able to fix this by keeping only one kernel per kernelspec).

## ToDo

- [x] Fixes https://github.com/holoviz/panel/issues/3436#issuecomment-1217832128
- [x] Add kernel selector via query parameter?
- [x] Ensure kernels are cleaned up eventually
- [x] Support autoreload
- [x] Add tests


MarcSkovMadsen:
I can't see there is a high probability of someone spinning up lots of kernels unless they have Jupyter lab for ever running.

But maybe in a Jupyter Hub environment this could become an issue?

Our JupyterHub closes the session/ server after 1 hour of inactivity though.

Issue #3647, opened 2022-06-24
Tabulator: `page` not updated when pagination is `'local'`


maximlt:
As per the title, when `pagination='local'` the widget's `page` attribute isn't updated and is always 1.

Issue #3663, opened 2022-06-30
TableEditEvent row index wrong when pagination is remote and a sorter is applied


maximlt:
When pagination is `remote` and a client sorter is applied to the table, editing a cell emits a TableEditEvent whose row index is wrong, i.e. it's not the one of the original DataFrame given to the Tabulator Widget.

maximlt:
Same issue with the click event.

Issue #3662, opened 2022-06-30
TableEditEvent row index wrong when a Python filter is applied


maximlt:
When a Python filter  is applied to a Tabulator widget, i.e. with `add_filter(value, column)`, and a cell is changed in the filtered table, the row index returned by the `TableEditEvent` is not the one of the original DataFrame but the one of the `current_view`.

maximlt:
Same issue with the click event.

Issue #3660, opened 2022-06-29
Error raised on edit events when a header filter is active


maximlt:
Execute the example below in a notebook, filter *col1* by `a` and edit one of the rows in *col2*.

```python
import pandas as pd
import panel as pn
pn.extension('tabulator')

df = pd.DataFrame({
    'col1': list('aaabcd'),
    'col2': list('ABCDEF')
})
header_filters = {
    'col1': {'type': 'input', 'func': 'like'}
}
widget = pn.widgets.Tabulator(
    df,
    header_filters=header_filters,
)
def cb(e):
    print(e)

widget.on_edit(cb)
```

Traceback:
```
Traceback (most recent call last):
  File "/Users/mliquet/miniconda3/envs/panel-dev38/lib/python3.8/site-packages/pyviz_comms/__init__.py", line 325, in _handle_msg
    self._on_msg(msg)
  File "/Users/mliquet/work/dev/panel/panel/viewable.py", line 292, in _on_msg
    doc.unhold()
  File "/Users/mliquet/miniconda3/envs/panel-dev38/lib/python3.8/site-packages/bokeh/document/document.py", line 799, in unhold
    self.callbacks.unhold()
  File "/Users/mliquet/miniconda3/envs/panel-dev38/lib/python3.8/site-packages/bokeh/document/callbacks.py", line 396, in unhold
    self.trigger_on_change(event)
  File "/Users/mliquet/miniconda3/envs/panel-dev38/lib/python3.8/site-packages/bokeh/document/callbacks.py", line 373, in trigger_on_change
    invoke_with_curdoc(doc, event.callback_invoker)
  File "/Users/mliquet/miniconda3/envs/panel-dev38/lib/python3.8/site-packages/bokeh/document/callbacks.py", line 408, in invoke_with_curdoc
    return f()
  File "/Users/mliquet/miniconda3/envs/panel-dev38/lib/python3.8/site-packages/bokeh/util/callback_manager.py", line 191, in invoke
    callback(attr, old, new)
  File "/Users/mliquet/work/dev/panel/panel/reactive.py", line 392, in _comm_change
    self._schedule_change(doc, comm)
  File "/Users/mliquet/work/dev/panel/panel/reactive.py", line 376, in _schedule_change
    self._change_event(doc)
  File "/Users/mliquet/work/dev/panel/panel/reactive.py", line 370, in _change_event
    self._process_events(events)
[...]

PR #3771, opened 2022-08-22
Tabulator: consistent event row index and selection

maximlt:
Fixes https://github.com/holoviz/panel/issues/3647, https://github.com/holoviz/panel/issues/3660, https://github.com/holoviz/panel/issues/3662, https://github.com/holoviz/panel/issues/3663

The goals of this PR were:
* to have the click and edit events consistently returning the index of the value DataFrame
* ~~to have the selection being based on the value DataFrame, not on the current_view~~

The code relied on the fact that the index of the value DataFrame should only have unique values. I'm using that even more.


philippjfr:
Thanks for this. Seems like there's one new unit test failure: https://github.com/holoviz/panel/runs/7955050138?check_suite_focus=true#step:5:1719

Old behavior that is actually incorrect?

maximlt:
When #3784 is merged there will be a few changes to make to this PR, as in Tabulator 5.3 `getPosition` has changed. This is the diff to apply. I've had first to solve a merge conflict, to make sure the `dataFiltering` function correctly contains the page update code.

```diff
diff --git a/panel/models/tabulator.ts b/panel/models/tabulator.ts
index 6e48d01d..f9bff329 100644
--- a/panel/models/tabulator.ts
+++ b/panel/models/tabulator.ts
@@ -768,7 +768,7 @@ export class DataTabulatorView extends PanelHTMLBoxView {
           tab_column.headerSortStartingDir = sort.dir
       }
       tab_column.cellClick = (_: any, cell: any) => {
-        const index_processed = cell.getRow().getPosition(true)
+        const index_processed = cell.getRow().getPosition() - 1
         this.model.trigger_event(new CellClickEvent(column.field, index_processed))
       }
       if (config_columns == null)
@@ -782,7 +782,7 @@ export class DataTabulatorView extends PanelHTMLBoxView {
         formatter: button_formatter,
         hozAlign: "center",
         cellClick: (_: any, cell: any) => {
-          const index_processed = cell.getRow().getPosition(true)
+          const index_processed = cell.getRow().getPosition() - 1
           this.model.trigger_event(new CellClickEvent(col, index_processed))
         }
       }
@@ -1086,8 +1086,7 @@ export class DataTabulatorView extends PanelHTMLBoxView {
 
   cellEdited(cell: any): void {
     const field = cell._cell.column.field;
-    // true to get the position of the filtered/sorted row index
-    const index_processed = cell.getRow().getPosition(true);
+    const index_processed = cell.getRow().getPosition() - 1;
     const index = cell._cell.row.data._index;
     const value = cell._cell.value;
     this._tabulator_cell_updating = true
diff --git a/panel/widgets/tables.py b/panel/widgets/tables.py
index 4f9c9dd4..49cb23c0 100644
--- a/panel/widgets/tables.py
+++ b/panel/widgets/tables.py
@@ -1106,7 +1106,7 @@ class Tabulator(BaseTable):
     def _process_event(self, event):
         event_col = self._renamed_cols.get(event.column, event.column)
         processed = self._sort_df(self._processed)
-        if self.pagination == 'remote':
+        if self.pagination in ['local', 'remote']:
             nrows = self.page_size
             event.row = event.row+(self.page-1)*nrows
         event_row = event.row
```

maximlt:
In https://github.com/holoviz/panel/pull/3771/commits/cc9448696d25945c8633df905780774c34e09850 I switched the sorting algorithm used by Pandas to `mergesort`. While testing this branch on a real app I found out that in some cases the index returned was slightly wrong. In this PR the Python code now in `_process_event` needs to sort the data as the index sent by the front-end is the one of the filtered/sorted table. However, Pandas sorting algorithm (quicksort in `.sort_values`) doesn't necessarily lend to the same result compared to the result computed by Tabulator JS.

Switching to `mergesort` instead of `quicksort` fixed the issue I encountered. However, as a matter of fact Tabulator JS seems to implement its own sorters for the different data types it handles, see https://github.com/olifolkerd/tabulator/tree/master/src/js/modules/Sort/defaults/sorters. This means that I'm not sure that the sorting we do is always the same as the one Tabulator JS does, and this is actually pretty hard to test. In the future we may want to find a more robust solution to this, maybe adding our own hidden index to keep track of such things.

maximlt:
Tested this PR on multiple real apps and saw no issue. It's unfortunate I had to revert the start of the work on making the selection indexes and selected_dataframe be more consistent, that will have to wait for another PR.

Issue #3695, opened 2022-07-14
Only part of the Tabulator widget displayed after an update


maximlt:
Triggering an update of a Tabulator table shrinks its display to 20 lines.

```python
import numpy as np
import pandas as pd
import panel as pn

pn.extension('tabulator')

df = pd.DataFrame(np.random.rand(50, 3), columns=['a', 'b', 'c'])

table = pn.widgets.Tabulator(df)

bt = pn.widgets.Button(name='click')

def cb(e):
    table.param.trigger('value')

bt.on_click(cb)

pn.Column(bt, table).servable()
```

![issue_tabulator_update](https://user-images.githubusercontent.com/35924738/179005150-59571d24-410a-47be-9133-eee36155eca4.gif)

Hoxbro:
This seems to work for me:
``` python
def cb(e):
    try:
        table._updating = True
        table.param.trigger('value')
    finally:
        table._updating = False
```

I noticed that the disappearing only happened when I was zoomed in 100% if 90% or 110%, the rows would not disappear.

maximlt:
Ok strange as it seems then it's a UI problem more than a Python problem, if it only happens for different zoom levels.

maximlt:
I believe this is a Tabulator issue that was fixed in a recent release: https://github.com/olifolkerd/tabulator/issues/3380

The rows disappear when this line is executed, which happens all in Tabulator JS:
https://github.com/holoviz/panel/blob/9cc50af8a3f3739e32b8952011eb5ab0df31e358/panel/models/tabulator.ts#L848

One possible workaround is to disable the virtual DOM rendering since this is what is causing the issue, with `configuration={'virtualDom': False}`, however I don't know to what other things this might break.

The real fix will be to update the Tabulator dependency to its latest version.

holovizbot:
This issue has been mentioned on **HoloViz Discourse**. There might be relevant details there:

https://discourse.holoviz.org/t/getting-tabulator-to-display-all-columns/8394/1

PR #3784, opened 2022-08-26
Upgrade to Tabulator 5.3.2

maximlt:
Fixes https://github.com/holoviz/panel/issues/3695


maximlt:
@philippjfr I expected this test - `test_tabulator_header_filter_no_horizontal_rescroll[None]` - to fail. It tests that when you have a wide table with an horizontal scrollbar, then after scrolling right, setting a header filter shouldn't update the scroll position. What happens after the upgrade is that the header filter entirely disappears! I've added a test that checks that this doesn't occur all the time, i.e. it seems to only happen when there's a horizontal scrollbar. Tabulator JS now emits more warnings when events are triggered when they should not, there are a few warnings triggered when the widget is rendered, so maybe that's an area to explore to try to fix that. I've attempted to change some of that logic, without luck unfortunately.

maximlt:
Testing this on real apps, I've seen many warnings related to events being triggered when they shouldn't (redraw in particular). I've also seen this error multiple times: `Uncaught (in promise) TypeError: Cannot read properties of null (reading offsetWidth)`.

philippjfr:
> What happens after the upgrade is that the header filter entirely disappears! I've added a test that checks that this doesn't occur all the time, i.e. it seems to only happen when there's a horizontal scrollbar.

So it doesn't actually disappear, the problem is that the scrollbar disappears when there are no rows, which in turn means you cannot see all the rows in the header. Not sure I'll be able to resolve this immediately.

>Tabulator JS now emits more warnings when events are triggered when they should not, there are a few warnings triggered when the widget is rendered, so maybe that's an area to explore to try to fix that. I've attempted to change some of that logic, without luck unfortunately.

Got a fix for this I'll push up shortly.

philippjfr:
>Not sure I'll be able to resolve this immediately.

I did resolve it. Had to add this:

```javascript
  this.tabulator.on("dataFiltered", (_: any, rows: any[]) => {
      if (this._initializing)
	return
      // Ensure that after filtering empty scroll renders
      if (rows.length === 0)
	this.tabulator.rowManager.renderEmptyScroll()
      // Ensure that after filtering the page is updated
      this.updatePage(this.tabulator.getPage())
    })
```

PR #3788, opened 2022-08-29
Supporting datetime bounds for DatetimePicker and DatetimeRangePicker

Hoxbro:
I noticed the options `minDateHasTime` and `maxDateHasTime` 

https://user-images.githubusercontent.com/19758978/187180672-d8ab009a-58ec-4c18-b38c-3d1c475d5ce4.mp4

``` python
import datetime

import panel as pn

datetime_picker = pn.widgets.DatetimePicker(
    name="Datetime Picker",
    value=datetime.datetime(2021, 3, 2, 12, 10),
    end=datetime.datetime(2021, 3, 4, 12, 10),
    start=datetime.datetime(2021, 3, 1, 12, 10),
)

pn.Row(
    datetime_picker,
    f"start: {datetime_picker.start}",
    f"end: {datetime_picker.end}",
).servable()
```

PR #3797, opened 2022-08-30
Support datetime bounds 2

Hoxbro:
I was not completely finished with the original PR #3788 before it got merged.

My original idea was to _only_ support datetime objects for `start`/`end`, but after a night's sleep, I think it is too aggressive. This adds support back for date objects too.

---

The reason why I convert the `min_date`/`max_date` in `_process_param_change` is that if a datestr (and not datetimestr) is passed to `flatpickr` in the bokeh model with the `maxDateHasTime`/`minDateHasTime` set to True, it has some weird behavior. This is shown here where the `end`/`max_date` is a date-object. [StackOverflow issue](https://stackoverflow.com/questions/65369608/unable-to-set-minhour-to-00-in-flatpicker-for-maxdate) about the problem. 

https://user-images.githubusercontent.com/19758978/187371557-ae83f8c3-3ad6-4e18-af51-cfae326d697a.mp4

### DIFF
--- a/panel/tests/widgets/test_input.py
+++ b/panel/tests/widgets/test_input.py
@@ -61,7 +61,7 @@ def test_date_picker(document, comm):
 def test_datetime_picker(document, comm):
     datetime_picker = DatetimePicker(
         name='DatetimePicker', value=datetime(2018, 9, 2, 1, 5),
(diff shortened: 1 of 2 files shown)


philippjfr:
Sorry for merging to early? Is this one ready?

Hoxbro:
Yes!

PR #3803, opened 2022-09-03
Add description of literal options to the docs for ReactiveHTML

### LINKED ISSUE #3802: Enable or document how to render options in a List Parameter as Literal elements
## Request

Enable or document how to render options in a List Parameter as Literal elements

## Motivation

I'm trying to show a user how to create a custom dropdown component using ReactiveHTML in https://discourse.holoviz.org/t/customize-panel-components-widgets-design/4187.

There is already an example of a *child* template/ dropdown here https://panel.holoviz.org/user_guide/Custom_Components.html#child-templates

Unfortunately the drop down options are not rendered as Literal values but instead as panels, i.e. wrapping the literal value in `bk` `div`s.

In my example it makes the dropdown not look nice because a lot of margin is added.

image

```python
import param

class ShoelaceSelect(ReactiveHTML):
    value = param.Parameter()
    options = param.List()
    _template = """
<sl-dropdown>
  <sl-button slot="trigger" caret>Dropdown</sl-button>
  <sl-menu>
   {% for obj in options %}
    <sl-menu-item id="option">${obj}</sl-menu-item>
    {% endfor %}
  </sl-menu>
</sl-dropdown>
"""

    __javascript_modules__=["https://cdn.jsdelivr.net/npm/@shoelace-style/shoelace@2.0.0-beta.82/dist/shoelace.js"]
    __css__ = ["https://cdn.jsdelivr.net/npm/@shoelace-style/shoelace@2.0.0-beta.82/dist/themes/light.css"]
    # _child_config = {'obj': 'literal'}

import panel as pn

pn.extension()

select = ShoelaceSelect(options=["apple", "pear", "banana"])

pn.Column(select).servable()
```

I simple cannot find any documentation or example telling me how to provide these a literal values.

MarcSkovMadsen:
closes #3802

### DIFF
--- a/examples/user_guide/Custom_Components.ipynb
+++ b/examples/user_guide/Custom_Components.ipynb
@@ -157,6 +157,16 @@
     "  <option id=\"option\">${obj}</option>\n",
     "  {% endfor %}\n",
     "</select>\n",
+    "```\n",
+    "\n",
+    "This will insert the children as Panel components. If you want to insert the children as *literal* values, you can use the syntax\n",
+    "\n",
+    "```html\n",
+    "<select>\n",
+    "  {% for obj in parameter %}\n",
+    "  <option id=\"option-{{ loop.index0 }}\">{{ obj }}</option>\n",
+    "  {% endfor %}\n",
+    "</select>\n",
     "```"
    ]
   },
@@ -331,7 +341,7 @@
    "source": [
     "#### Child templates\n",
     "\n",
-    "If we want to provide a template for the children of an HTML node we have to use Jinja2 syntax to loop over the parameter. The component will automatically assign each `<option>` tag a unique id and insert the loop variable `option` into each of the tags:"
+    "If we want to provide a template for the children of an HTML node we have to use Jinja2 syntax to loop over the parameter. The component will insert the loop variable `option` into each of the tags:"
    ]
   },
   {
@@ -349,7 +359,7 @@
     "    _template = \"\"\"\n",
     "    <select id=\"select\" value=\"${value}\" style=\"width: ${model.width}px\">\n",
     "      {% for option in options %}\n",
-    "      <option id=\"option\">${option}</option>\n",
+    "      <option id=\"option-{{ loop.index0 }}\">{{ option }}</option>\n",
     "      {% endfor %}\n",
     "    </select>\n",
     "    \"\"\"\n",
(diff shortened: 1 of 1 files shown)


examples/user_guide/Custom_Components.ipynb
```diff
@@ -349,7 +359,7 @@
     "    _template = \"\"\"\n",
     "    <select id=\"select\" value=\"${value}\" style=\"width: ${model.width}px\">\n",
     "      {% for option in options %}\n",
-    "      <option id=\"option\">${option}</option>\n",
+    "      <option id=\"option-{{ loop.index0 }}\">{{ option }}</option>\n",
```
MarcSkovMadsen, line 362: You can keep the existing example. But it is misleading as you will not use that syntax in practice to create a Select element for literal values as the example shows. That the example is misleading takes a long time to figure out as you *trust* the documentation.

examples/user_guide/Custom_Components.ipynb
```diff
@@ -540,9 +550,22 @@
   }
  ],
  "metadata": {
+  "kernelspec": {
```
MarcSkovMadsen, line 553: Please tell me if I can do something to keep these meta data changes out. Thanks.

examples/user_guide/Custom_Components.ipynb
MarcSkovMadsen, line 362: I don't understand why you cannot use `id='option'`. But it raises an exception.

MarcSkovMadsen:
It fails test ` test_plotly_3d_plot` on MacOS, I can't see its related to this PR.

examples/user_guide/Custom_Components.ipynb
philippjfr: You absolutely should be able to as it should automatically add the index.

examples/user_guide/Custom_Components.ipynb
philippjfr: On Unix based systems I use:

```
#!/bin/bash
jq --sort-keys --indent 1 \
    '
    (.cells[] | select(has("outputs")) | .outputs) = []
    | (.cells[] | select(has("execution_count")) | .execution_count) = null
    | .metadata = {"language_info": {"name":"python", "pygments_lexer": "ipython3"}}
    | .cells[].metadata = {}
    | del(.cells[].id)
    ' "$1" > /tmp/$(basename $1) && cat /tmp/$(basename $1) > "$1"
```

to clear out the metadata.

philippjfr:
Closing in favor of https://github.com/holoviz/panel/pull/3813 which makes these instructions unncessary.

philippjfr:
Actually some of the updates here should still be made. Maybe update it so it doesn't instruct users to use `{{ loop.index0 }}` as that should now be unncessary.

examples/user_guide/Custom_Components.ipynb
philippjfr: It's not necessarily misleading I guess just a little silly to render some text by creating an entire child model.

philippjfr:
pre-commit.ci autofix

PR #3817, opened 2022-09-08
Add documentation for working with WebAssembly/Pyodide/Pyscript

philippjfr:
@jbednar review would be appreciated before I make the page look pretty


jbednar:
Nice! I made some edits; please accept or revert them as you wish. A few remaining issues:

- The title probably won't mean much to people who don't know about this functionality already. Can we phrase it more generically, e.g. something like "Exporting as Python in the browser"?  
- The option to export a full app seems like the most directly relevant to most Panel users, since it doesn't require learning about DOMs or HTML. Shouldn't that be described first, then the other approaches as a deeper topic?
- Doesn't it need a section talking about how to access data, things that won't work in PyScript, async vs. non-async calls, and other gotchas?

doc/user_guide/Running_in_Webassembly.md
```diff
@@ -0,0 +1,176 @@
+# Running in Webassembly
+
+Panel lets you write dashboards and other applications in Python that are accessed using a web browser. Typically, the Python interpreter runs as a separate Jupyter or Bokeh server process, communicating with JavaScript code running in the client browser. However, it is now (as of 2022) possible to run Python directly in the browser, with no separate server needed!
```
MarcSkovMadsen, line 3: remove (as of 2022). Its not necessary and in 1 or 2 years someone will see the documentation as outdated due to this.

doc/user_guide/Running_in_Webassembly.md
```diff
@@ -0,0 +1,176 @@
...
+
+    <script type="text/javascript" src="https://cdn.bokeh.org/bokeh/release/bokeh-2.4.3.js"></script>
+    <script type="text/javascript" src="https://cdn.bokeh.org/bokeh/release/bokeh-widgets-2.4.3.min.js"></script>
+    <script type="text/javascript" src="https://cdn.bokeh.org/bokeh/release/bokeh-tables-2.4.3.min.js"></script>
+    <script type="text/javascript" src="https://cdn.jsdelivr.net/npm/@holoviz/panel@0.14.0/dist/panel.min.js"></script>
+
+  </head>
+  <body>
+    <script type="text/javascript">
+      async function main(){
```
MarcSkovMadsen, line 38: Make the example able to run something like outputting `hello world` in a `div`. I know from experience that users might not always know how to turn an indicative example into a working example.

doc/user_guide/Running_in_Webassembly.md
```diff
@@ -0,0 +1,176 @@
...
+    <script defer src="https://pyscript.net/stable/pyscript.js"></script>
+
+  </head>
+  <body>
+    <py-env>
+      - panel
+      ...
+    </py-env>
+    <py-script>
+      import panel as pn
```
MarcSkovMadsen, line 86: As above: make the example work.

doc/user_guide/Running_in_Webassembly.md
```diff
@@ -0,0 +1,176 @@
...
+      --index               Whether to create an index if multiple files are served.
+      --pwa                 Whether to add files to allow serving the application as a Progressive Web App.
+      --requirements        Explicit list of Python requirements to add to the converted file.
+
+### Formats
+
+Using the `--to` argument on the CLI you can control the format of the file that is generated by `panel convert`. You have three options, each with distinct advantages and disadvantages:
+
+- **`pyodide`** (default): Run application using Pyodide running in the main thread. This option is less performant than pyodide-worker but produces completely standalone HTML files that do not have to be hosted on a static file server.
+- **`pyodide-worker`**: Generates an HTML file and a JS file containing a Web Worker that runs in a separate thread. This is the most performant option, but files have to be hosted on a static file server.
```
MarcSkovMadsen, line 161: Extend line to "static file server like github pages". Users might not know what a static file server means.

droumis:
> * The option to export a full app seems like the most directly relevant to most Panel users, since it doesn't require learning about DOMs or HTML. Shouldn't that be described first, then the other approaches as a deeper topic?

Another reason this is important is that this section contains the advantages and disadvantages of pyodide, pyodide-worker, and pyscript. Having this closer to the beginning helps to justify why there are multiple options being presented and orients their attention to what is more relevant to them (performance, readability, static file server).

doc/user_guide/Running_in_Webassembly.md
```diff
@@ -0,0 +1,176 @@
...
+  def callback(new):
+      return f'Amplitude is: {new}'
+
+  pn.Row(slider, pn.bind(callback, slider))
+</py-script>
+```
+
+## Converting Panel applications
+
+Writing an HTML file from scratch with all the Javascript and Python dependencies and other boilerplate can be quite annoying, and requires learning a good bit of HTML. To avoid writing all the boilerplate, Panel provides support for converting an entire application (including Panel templates) to an HTML file, using the `panel convert` API. As a starting point create one or more Python scripts or notebook files containing your application. The only requirement is that they import only global modules (relative imports of other scripts or modules is not supported).
```
droumis, line 139: the 'global' modules still have to be on the [pyodide supported list](https://github.com/pyodide/pyodide/tree/main/packages) no? If so, maybe mention and add link.

droumis (commented):
Looks great! I just had couple of minor suggestions:

doc/user_guide/Running_in_Webassembly.md
```diff
@@ -1,3 +1,5 @@
+# Running in Webassembly
```
mattpap, line 1:
```suggestion
# Running in WebAssembly
```

doc/user_guide/Running_in_Webassembly.md
philippjfr: Good point, that's not quite accurate but I'll clarify.

doc/user_guide/Running_in_Webassembly.md
```diff
@@ -185,3 +185,52 @@ Current versions of PyScript will automatically render the output of the last ce
   pn.Row(slider, pn.bind(callback, slider))
 </py-script>
 ```
+
+## Embedding in documentation
+
+One more option is to include live Panel examples in your Sphinx documentation using the `nbsite.pyodide` directive. Simply install latest nbsite with `pip` or `conda`:
```
maximlt, line 191: Do we really want people to use `nbsite`?

MarcSkovMadsen:
Other interesting options is embedding in simple Markdown documents. Is it possible? How?

And will it work for example in Github Readme.md file?

philippjfr:
Good question but I don't think you load and execute JS in a Markdown file.

PR #3825, opened 2022-09-12
Tabulator: ignore case when sorting strings

maximlt:
Tabulator JS implements its own sorting algorithms which for the string type ignores the case, while `df.sort_values` does and is used internally to mirror what happens on the client side. This PR adds a key function to `sort_values` so that strings are temporally lower-cased to sort them.

### DIFF
--- a/panel/widgets/tables.py
+++ b/panel/widgets/tables.py
@@ -311,7 +311,7 @@ class BaseTable(ReactiveData, Widget):
         def tabulator_sorter(col):
             # Tabulator JS defines its own sorting algorithm:
             # - strings's case isn't taken into account
-            if col.dtype.kind not in 'SUO': 
+            if col.dtype.kind not in 'SUO':
                 return col
             try:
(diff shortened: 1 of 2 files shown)


panel/widgets/tables.py
```diff
@@ -308,7 +308,18 @@ def _sort_df(self, df: DataFrameType) -> DataFrameType:
         else:
             rename = False
 
-        df_sorted = df.sort_values(fields, ascending=ascending, kind='mergesort')
+        def tabulator_sorter(col):
+            # Tabulator JS defines its own sorting algorithm:
+            # - strings's case isn't taken into account
+            if col.dtype != 'O':
```
philippjfr, line 314:
```suggestion
            if col.dtype.kind not in 'SUO': 
```

philippjfr:
One suggestion about taking fixed length string and unicode columns into account, otherwise this looks good.

PR #3826, opened 2022-09-12
Tabulator: fix default values of a list header filter

maximlt:
Before the Tabulator JS upgrade `values: True` was alright but now we need to use `valuesLookup: True`.


examples/reference/widgets/Tabulator.ipynb
```diff
@@ -216,7 +216,7 @@
     "    'int': None,\n",
     "    'float': {'type': 'number', 'max': 10, 'step': 0.1},\n",
     "    'bool': {'type': 'tickCross', 'tristate': True, 'indeterminateValue': None},\n",
-    "    'str': {'type': 'autocomplete', 'values': True},\n",
+    "    'str': {'type': 'list', 'valuesLookup': True},\n",
```
philippjfr, line 219: Is 'values' entirely ignored now or does it have another meaning? Trying to work out whether we can provide backward compatibility.

maximlt: It seems to be that when `values` is set to True it is just ignored by Tabulator JS. Otherwise `values` accept a list of values or dict of value/label.

We could look for `values` in the editor or header filter params, when its type is `select` or `autocomplete`, and replace it by `valuesLookup`?

philippjfr: Sounds good!

maximlt: I'm adding backwards compatibility to `select` and `autocomplete` in [6ec3c1e](https://github.com/holoviz/panel/pull/3826/commits/6ec3c1ec648144872fa81a3ccdac330de8dbd324), for the editor and header filter. The editor didn't set `values` to True by default, so I'm sticking to that.

I've also added a warning to encourage users to switch to `list` and `lookupValues`. Used `self.param.warning`, is that the preferred way over `warnings.warn`? I'm never sure.

philippjfr: Honestly I have lost track about warnings too, I think at some point we just need to go through and make everything consistent.

philippjfr:
Looks good, thanks!

Issue #3129, opened 2022-01-21
Events are not being sent in Panel/Django apps


sdc50:
Panel=0.13.0a31
Bokeh=2.4.2
Django=2.2.14

Ever since #3115 I can't get a Panel app embedded in Django to work. It initially renders fine, but it stalls on any interaction. It appears that events messages are not being sent over the websocket. 

WebSocket messages with Panel=0.13.0a31
![image](https://user-images.githubusercontent.com/6454472/150607171-1a7f00ec-3b4f-4202-873d-84101c75f31e.png)

WebSocket messages with Panel=0.13.0a28
![image](https://user-images.githubusercontent.com/6454472/150607392-bcdbeac3-3db2-448b-8799-99ba26f56603.png)

sdc50:
@philippjfr I think this is the issue that I've been dealing with in updating the bokeh-django repo. When I installed Panel=0.13.0a28 then it works. I'll start looking at what has changed since that tag to see if I can figure out how to fix it.

philippjfr:
Hmm, yeah if you can track that down that would be tremendously helpful.

sdc50:
fixed by #3828

PR #3828, opened 2022-09-13
Fix event handling for Django integration

sdc50:
This is likely not the "right" solution, but after comparing the latest code with the last functioning version this was the easiest way to restore the same behavior. Hopefully it illustrates what the issue is helps lead toward a better solution.

Fixes issue #3129

### DIFF
--- a/panel/io/django.py
+++ b/panel/io/django.py
@@ -1,7 +1,7 @@
 from contextlib import contextmanager
 from urllib.parse import urljoin, urlparse
 
-from bokeh.server.django.consumers import AutoloadJsConsumer, DocConsumer
+from bokeh_django.consumers import DocConsumer, AutoloadJsConsumer
 
 from ..util import edit_readonly
 from .resources import Resources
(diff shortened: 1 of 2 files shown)


panel/io/django.py
```diff
@@ -1,7 +1,7 @@
 from contextlib import contextmanager
 from urllib.parse import urljoin, urlparse
 
-from bokeh.server.django.consumers import AutoloadJsConsumer, DocConsumer
+from bokeh_django.consumers import DocConsumer, AutoloadJsConsumer
```
sdc50, line 4: @philippjfr I'm wondering if it makes sense to move this code to the `bokeh_django` repo rather than maintaining it here. We could conditionally import `panel` and if present then modify the `DocConsumer` and `AutoloadJsConsumer` accordingly.

panel/io/document.py
```diff
@@ -16,6 +16,7 @@
 from bokeh.application.application import SessionContext
 from bokeh.document.document import Document
 from bokeh.document.events import DocumentChangedEvent, ModelChangedEvent
+from bokeh_django.consumers import WSConsumer
```
sdc50, line 19: I'm hoping there is a better condition to test in the change made below, but if using `WSConsumer` ends up being the best case then maybe we want to conditionally import it here so `panel` doesn't have to depend on the `bokeh_django` repo.

philippjfr:
Superseded by https://github.com/holoviz/panel/pull/3835

PR #3835, opened 2022-09-14
Handle Django compatibility

philippjfr:
Improve dispatching to DjangoConsumer in panel.io.document.unlocked contextmanager and add forward compatibility for bokeh_django.

@sdc50 Please test.

### DIFF
--- a/panel/io/django.py
+++ b/panel/io/django.py
@@ -1,7 +1,10 @@
 from contextlib import contextmanager
 from urllib.parse import urljoin, urlparse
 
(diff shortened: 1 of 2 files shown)


sdc50:
@philippjfr Sorry I didn't see this yesterday. I just tested and it works great!

I'm still working through what needs to be fixed when setting `BOKEH_RESOURCES` to `server`. From what I can tell so far it seems to be a limitation with `bokeh.embed.bundle.bundle_for_objs_and_resources` not allowing you to specify a `root_url`.

PR #3837, opened 2022-09-14
Making the test suite more robust

maximlt:
Generalize the `wait_until` utility to be usable outside of UI tests, plus some other minor changes. Let's see how that goes.


philippjfr:
It's an improvement in any case. Clearly more work to do though.

PR #3839, opened 2022-09-15
Tabulator: custom datetime sorter

maximlt:
As already pointed out before Tabulator JS defines its own sorters, which can behave differently from `df.sort_values`. The result of these two sorters was different when applied to datetime columns that contain NaN values (this causes issues with the index of edit/click events returned when the sorters aren't aligned on their result...). Datetime cols are actually serialized as UNIX timestamps by Bokeh, and NaN values to `-9223372036854776` (see https://github.com/bokeh/bokeh/issues/10448). Tabulator JS was using its *number* built-in sorter to sorter this kind of column in the end, the NaN values being first when sorting in an ascending way. The default behavior of `df.sort_values` is put NaNs at the end (see https://pandas.pydata.org/docs/reference/api/pandas.Series.sort_values.html#pandas.Series.sort_values). This is where the two sorting algorithms were diverging. Pandas' `df.sort_values` has the `na_position` parameter that can be set to `'first'`, except that it applies to all the columns which I didn't want to do that now, as it affects all the dtypes. Instead this PR adds a custom sorter for datetime objects which always puts the NaNs at the end as `df.sort_values`.

### DIFF
--- a/panel/models/tabulator.ts
+++ b/panel/models/tabulator.ts
@@ -120,6 +120,53 @@ function group_data(records: any[], columns: any[], indexes: string[], aggregato
   return grouped
 }
 
+
+const timestampSorter = function(a: any, b: any, _aRow: any, _bRow: any, _column: any, _dir: any, _params: any){
+  // Bokeh serializes datetime objects as UNIX timestamps.
+
+  //a, b - the two values being compared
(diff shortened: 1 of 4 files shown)


philippjfr:
Ready to merge?

maximlt:
Yes

PR #3841, opened 2022-09-15
Tabulator: simplify event row calculation

maximlt:
Instead of relying on the actual (i.e. visual) position of the row clicked/edited this PR computes an index mapping - that relies on the existence of the hidden `_index` column used on the front-side for other reasons - and use that index when processing a click/edit event to obtain the actual row iloc in the value DataFrame. This has the effect to not have to rely on sorting data to get the row iloc, which as seen recently caused issue due to divergences between Tabulator JS' sorting algorithms and Pandas `df.sort_values()`.


philippjfr:
Agree this is significantly cleaner!

PR #3844, opened 2022-09-15
Update Django Example Apps

sdc50:
Update django example apps to work with multiple versions of Django/Channels.

Changes will allow Django example apps to be run with Django/Channels 2 and with Django 3/4 (with Channels 3). 

Conditionally imports `bokeh_django` to enable using Channels 3.


sdc50:
@philippjfr once we release `bokeh_django` then we can update the docs to explain how to install it.

philippjfr:
Worth making these updates even if `bokeh_django` isn't available yet. Will merge shortly.

Issue #3848, opened 2022-09-16
Tabulator does not support editting Int64Dtype


Hoxbro:
Pandas dtype for nullable integer is not currently supported in Tabulator when editing columns.

``` python
import numpy as np
import pandas as pd
import panel as pn
pn.extension("tabulator")

df = pd._testing.makeDataFrame()
df["B"][::5] = np.nan
df["B"] = df["B"].round().astype(pd.Int64Dtype())
pn.widgets.Tabulator(df)
```

``` python-traceback
Traceback (most recent call last):
  File "/home/shh/miniconda3/envs/holoviz/lib/python3.9/site-packages/pyviz_comms/__init__.py", line 346, in _handle_msg
    self._on_msg(msg)
  File "/home/shh/Development/holoviz/repos/panel/panel/viewable.py", line 292, in _on_msg
    doc.unhold()
  File "/home/shh/miniconda3/envs/holoviz/lib/python3.9/site-packages/bokeh/document/document.py", line 799, in unhold
    self.callbacks.unhold()
  File "/home/shh/miniconda3/envs/holoviz/lib/python3.9/site-packages/bokeh/document/callbacks.py", line 396, in unhold
    self.trigger_on_change(event)
  File "/home/shh/miniconda3/envs/holoviz/lib/python3.9/site-packages/bokeh/document/callbacks.py", line 373, in trigger_on_change
    invoke_with_curdoc(doc, event.callback_invoker)
  File "/home/shh/miniconda3/envs/holoviz/lib/python3.9/site-packages/bokeh/document/callbacks.py", line 408, in invoke_with_curdoc
    return f()
  File "/home/shh/miniconda3/envs/holoviz/lib/python3.9/site-packages/bokeh/util/callback_manager.py", line 191, in invoke
    callback(attr, old, new)
  File "/home/shh/Development/holoviz/repos/panel/panel/reactive.py", line 389, in _comm_change
    self._schedule_change(doc, comm)
  File "/home/shh/Development/holoviz/repos/panel/panel/reactive.py", line 373, in _schedule_change
    self._change_event(doc)
  File "/home/shh/Development/holoviz/repos/panel/panel/reactive.py", line 369, in _change_event
    self._process_events(events)
  File "/home/shh/Development/holoviz/repos/panel/panel/reactive.py", line 1108, in _process_events
    self._process_data(events.pop('data'))
  File "/home/shh/Development/holoviz/repos/panel/panel/widgets/tables.py", line 1207, in _process_data
[...]

PR #3850, opened 2022-09-16
Support editing of pandas masked array dtypes in Tabulator

philippjfr:
Fixes https://github.com/holoviz/panel/issues/3848

### DIFF
--- a/panel/reactive.py
+++ b/panel/reactive.py
@@ -26,6 +26,7 @@ import param
 
 from bokeh.core.property.descriptors import UnsetValueError
 from bokeh.model import DataModel
+from packaging.version import Version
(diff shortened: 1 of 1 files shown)


panel/reactive.py
```diff
@@ -1049,6 +1050,13 @@ def _convert_column(
                         iv = dt.date.fromtimestamp(iv/1000)
                     new_values.append(iv)
                 converted = new_values
+        elif 'pandas' in sys.modules:
+            import pandas as pd
+            if Version(pd.__version__) >= '1.1.0':
```
Hoxbro, line 1055:
```suggestion
            if Version(pd.__version__) >= Version('1.1.0'):
```

PR #3866, opened 2022-09-20
Upgrade GitHub Actions

cclauss:


philippjfr:
Looking good, thanks for the update! One issue we have with adding 3.10 to the test matrix is that it puts us (significantly) over the size of the GH Actions cache, which means that some environments will be evicted from the cache and will end up being rebuilt each time. I'm wondering whether we have a policy where we test the oldest Python and the two most recent Python versions (and therefore drop 3.8 from the matrix). @maximlt Do you have any opinions?

cclauss:
Evens or odds also work: 3.6, 3.8, 3.10 or 3.7, 3.9, 3.11 (only two weeks away).

maximlt:
I would rather drop 3.9 in that case, based on the numbers of downloads I see on [PyPiStats](https://pypistats.org/packages/panel) there are way more downloads for 3.8 than 3.9. In practice I'm not sure I've ever seen on Panel an error introduced in the code that would break one of the intermediate supported versions. I find the tests useful though to spot errors in the ecosystem, which can happen on any version. That's why I suggest making that decision based on download counts.

And I would suggest running the complete matrix on scheduled runs, manual runs (triggered from GH UI) and dev releases.

philippjfr:
Okay, seems like all the changes here have now been applied. Thanks again @cclauss and sorry we didn't end up merging this specific PR!

PR #3880, opened 2022-09-23
Added testing of python in markdown documentation

### LINKED ISSUE #3877: Test documentation markdown documentation python
With the switch of some of the documentation from notebook files to markdown files, it makes sense to add a way to test all the code blocks to verify it is able to run from top to bottom.

Hoxbro:
Fixes #3877 

Found some missing imports, etc., which made it impossible to run the document from top to bottom.


maximlt:
Nice! Just some suggestions:

- I guess this test should run as part of the test_examples step.
- As it's running a single test that executes many files, it may be interesting to use https://github.com/pytest-dev/pytest-subtests to separate more clearly the tests in the output (never used subtests before but heard of it recently)

Hoxbro:
> I guess this test should run as part of the test_examples step.

Definitely, though I don't know what the best way to do it is. 

> As it's running a single test that executes many files

I have refactored the code to use `pytest.mark.parametrize`, which separates the test for each file and removes the need for an extra dependency.

philippjfr:
Thanks @Hoxbro! Quite happy to have something like this now. Eventually I want to implement a Markdown converter and that will either reuse this code or implement it separately. Until then this is nice to have.

PR #3882, opened 2022-09-25
Add ability to defer load until after page is rendered

### LINKED ISSUE #3881: Make it easier to defer the loading of panels
## Request

Make it easier to defer computations and displaying complex/ big objects to after the page has loaded.

## Motivation

I often see users of Panel develop applications that takes a long time to load. In the mean time the application appears unresponsive. This then leads to a bad user experience because users are not used to this and starts wondering if the application has failed or not. They  might refresh the application leading to more stress on the server.

Working with deployment on Lighting AI I can also see they have built in to their framework an expectation that the application responds fast (less than a second). This works for all the existing frameworks they support including data app frameworks like Streamlit, Dash and Gradio. 

- Streamlit works because it only responds with a minimum response to connect the web socket. The rest is loaded over the web socket.
- Gradio normally does not do anything heavy before the user has clicked a "submit" button.

For Panel it can be quite hard to respond within 1 second if you start loading data, using a template and/ or respond with a complicated dashboards.

Streamlit provides a `st.spinner` context manager to make this really easy. So they have never had the problem that their apps appears unresponsive.

## Existing solution

To defer load Panel provides the `pn.state.onload` function that you can use defer the heavy lifting to after the initial response.

This works. But its a callback api and means you have to write a lot of boiler plate code to use it. 

## Potential new solutions

- Change the initial request of Panel to do the same as Streamlit, i.e. respond with the minimum required to enable the app to transfer the rest via the web socket
    - This will solve the lightning ai 1 second issue
    - This will **not** give the user a better UI.
- Provide a `pn.state.loaded` parameter that users can `pn.bind` and `pn.depends` to.
    - This will remove the callback api
    - Users will have to provide an extra loaded argument to their functions which is not ideal.
- Add a `defer_load` parameter to  `pn.panel` to signal that the object or bound function should not load before the page has loaded. In the meantime some spinner should be shown.
- Provide a `defer_load` parameter to `pn.extension` that will defer the load of all `pn.panel`s to after the initial load. And before the initial load they will individually show a loading spinner.
- `defer_load` could potentially be added to `pn.bind` and `pn.depends` instead of `pn.panel`.
- `pn.defer_load` could potentially be an annotation and function instead.

## Basic App

This app takes a long time to load. In the mean time the app is unresponsive. This is what we normally teach our users to do. This is a bad user experience.

https://user-images.githubusercontent.com/42288570/192103784-97efd2c5-b124-4253-aeee-7e16c6bdee05.mp4

```python
import time
start = time.time()

import pandas as pd
import panel as pn
import hvplot.pandas

pn.extension(sizing_mode="stretch_width", template="fast")

def duration():
    return time.time()-start

def big_load():
    time.sleep(2)
    return pd.DataFrame({"x": [1,2,3], "y": [1,2,3]})

def big_plot():
    time.sleep(2)
    return pd.DataFrame({"x": [1,2,3], "y": [1,2,3]}).hvplot()

pn.Column(
    duration, big_load, big_plot
).servable()
```

## Basic App with `pn.state.onload`

This is provides a much better user experience. But the developer experience is not great. Its a callback and the callback might become complicated if there are many `.servables` you want to update. This is so hard to explain to the *new* or *inexperienced* users I know.

It seems like a lot of boilerplate code that you always must add your self if you want to provide a nice user experience.

https://user-images.githubusercontent.com/42288570/192106000-8f1d13b6-0310-4d80-8d27-cfdc35a63353.mp4

```python
import time
start = time.time()

import pandas as pd
import panel as pn
import hvplot.pandas

pn.extension(sizing_mode="stretch_width", template="fast")

def duration():
    return time.time()-start

def big_load():
    time.sleep(2)
    return pd.DataFrame({"x": [1,2,3], "y": [1,2,3]})

def big_plot():
    time.sleep(2)
    return pd.DataFrame({"x": [1,2,3], "y": [1,2,3]}).hvplot()

layout = pn.Column(
    pn.indicators.LoadingSpinner(value=True, width=200, height=200)
)

def on_load():
    layout[:]=[big_load, big_plot, duration]
pn.state.onload(on_load)

layout.servable()
```

philippjfr:
Makes it possible to defer the loading of components until after the page is rendered. Can be used explicitly with `ParamMethod` and `ParamFunction` (or via `panel()`) OR by setting `config.defer_load`.  

```python
import time
start = time.time()

import pandas as pd
import panel as pn
import hvplot.pandas

pn.extension(
    sizing_mode='stretch_width', template='fast', defer_load=True,
    loading_spinner='arc', loading_color='blue'
)

def duration():
    return time.time()-start

def big_load():
    time.sleep(2)
    return pd.DataFrame({"x": [1,2,3], "y": [1,2,3]})

def big_plot():
    time.sleep(2)
    return pd.DataFrame({"x": [1,2,3], "y": [1,2,3]}).hvplot()

pn.Column(big_load, big_plot, duration).servable()
```

defer_load

- [x] Fixes #3881 
- [x] Add docs

### DIFF
--- a/doc/user_guide/Overview.md
+++ b/doc/user_guide/Overview.md
@@ -148,6 +148,7 @@ The `pn.config` object allows setting various configuration variables, the confi
 > - `js_files`: External JS files to load. Dictionary should map from exported name to the URL of the JS file.
 > - `loading_spinner`: The style of the global loading indicator, e.g. 'arcs', 'bars', 'dots', 'petals'.
 > - `loading_color`: The color of the global loading indicator as a hex color, e.g. #6a6a6a
+> - `defer_load`: Whether reactive function evaluation is deferred until the page is rendered.
 > - `nthreads`: If set will start a `ThreadPoolExecutor` to dispatch events to for concurrent execution on separate cores. By default no thread pool is launched, while setting nthreads=0 launches `min(32, os.cpu_count() + 4)` threads.
 > - `raw_css`: List of raw CSS strings to add to load.
 > - `reuse_sessions`: Whether to reuse a session for the initial request to speed up the initial page render. Note that if the initial page differs between sessions, e.g. because it uses query parameters to modify the rendered content, then this option will result in the wrong content being rendered. See the [Performance and Debugging guide](Performance_and_Debugging.rst#Reuse-sessions) for more information.
--- a/examples/user_guide/Session_State_and_Callbacks.ipynb
+++ b/examples/user_guide/Session_State_and_Callbacks.ipynb
@@ -197,6 +197,25 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
+    "Alternatively we may also use the `defer_load` option to wait to evaluate a function until the page is loaded. This will render a placeholder and display the global `config.loading_spinner`:"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "def render_on_load():\n",
+    "    return pn.widgets.Select(options=['A', 'B', 'C'])\n",
+    "\n",
+    "pn.Row(pn.panel(render_on_load, defer_load=True))"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
     "#### pn.state.on_session_destroyed\n",
     "\n",
     "In many cases it is useful to define on_session_destroyed callbacks to perform any custom cleanup that is required, e.g,  dispose  a database engine, or when a user is logged out. These callbacks can be registered with `pn.state.on_session_destroyed(callback)`"
--- a/panel/config.py
+++ b/panel/config.py
@@ -112,6 +112,9 @@ class _config(_base_config):
     autoreload = param.Boolean(default=False, doc="""
         Whether to autoreload server when script changes.""")
 
+    defer_load = param.Boolean(default=False, doc="""
+        Whether to defer load of rendered functions.""")
+
     load_entry_points = param.Boolean(default=True, doc="""
         Load entry points from external packages.""")
 
--- a/panel/io/server.py
+++ b/panel/io/server.py
@@ -275,7 +275,8 @@ class Application(BkApplication):
         super().initialize_document(doc)
         if doc in state._templates and doc not in state._templates[doc]._documents:
             template = state._templates[doc]
-            template.server_doc(title=template.title, location=True, doc=doc)
+            with set_curdoc(doc):
+                template.server_doc(title=template.title, location=True, doc=doc)
 
 bokeh.command.util.Application = Application # type: ignore
 
--- a/panel/io/state.py
+++ b/panel/io/state.py
@@ -99,6 +99,9 @@ class _state(param.Parameterized):
        Object with encrypt and decrypt methods to support encryption
        of secret variables including OAuth information.""")
 
+    loaded = param.Boolean(default=False, doc="""
+       Whether the page is fully loaded.""")
+
     rel_path = param.String(default='', readonly=True, doc="""
        Relative path from the current app being served to the root URL.
        If application is embedded in a different server via autoload.js
@@ -162,6 +165,7 @@ class _state(param.Parameterized):
     # Dictionary of callbacks to be triggered on app load
     _onload: ClassVar[Dict[Document, Callable[[], None]]] = WeakKeyDictionary()
     _on_session_created: ClassVar[List[Callable[[BokehSessionContext], []]]] = []
+    _loaded: ClassVar[WeakKeyDictionary[Document, bool]] = WeakKeyDictionary()
 
     # Module that was run during setup
     _setup_module = None
@@ -323,6 +327,7 @@ class _state(param.Parameterized):
 
     def _on_load(self, doc: Optional[Document] = None) -> None:
         doc = doc or self.curdoc
+        self._loaded[doc] = True
         callbacks = self._onload.pop(doc, [])
         if not callbacks:
             return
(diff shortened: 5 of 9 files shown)


philippjfr:
@MarcSkovMadsen Give this one a go. For now it simply uses the `loading` parameter and therefore displays the globally configured `loading_spinner`.

MarcSkovMadsen:
Thanks for taking this seriously @philippjfr . I actually believe its the most important missing piece of the Panel api to make it easy to create data apps with a great look and feel. There are still a few things around `pn.bind`. But this is actually a big step and its so easy to explain to users.

I will review now.

MarcSkovMadsen:
Using the code example below I noticed a few thing. 

- The last `duration` function for some reason does not show a loading indicator ??
- The theme toggle is not white https://github.com/holoviz/panel/issues/3883
- The plots update really slowly compared to usual (I believe) when I drag the slider ??

https://user-images.githubusercontent.com/42288570/192145216-cd0ba4cd-a093-46a3-adb4-b74a0ec588ee.mp4

```python
import time
start = time.time()

import pandas as pd
import panel as pn
import hvplot.pandas

pn.extension(
    sizing_mode='stretch_width', template='fast', defer_load=True,
    loading_spinner='arc', loading_color='#0072B5'
)

count = pn.widgets.IntSlider(value=3, start=1, end=10, name="Count").servable(target="sidebar")

def big_load():
    time.sleep(1)
    return pd.DataFrame({"x": [1,2,3], "y": [1,2,3]})

def big_plot(count=3):
    # time.sleep(2)
    return pd.DataFrame({"x": range(0,count), "y": range(0,count)}).hvplot(height=100, xlim=(0,10))

@pn.depends(count)
def big_plot1(count=3):
    return big_plot(count)

def duration():
    time.sleep(3)
    return time.time()-start

ibig_plot = pn.bind(big_plot, count=count)

layout= pn.Column(big_load, big_plot, big_plot1, ibig_plot, duration)

add = pn.widgets.Button(name="Add").servable(target="sidebar")

def add_panel(_):
    print("update")
    layout.append(pn.panel(ibig_plot))

add.on_click(add_panel)

layout.servable()
```

MarcSkovMadsen:
I this example I would expect the value of the `duration_no_sleep` to be shown immediately, but it is not.

https://user-images.githubusercontent.com/42288570/192145567-26fe0a7f-36bd-46fa-bf0b-33a5d4fc1c17.mp4

```python
import time
start = time.time()

import pandas as pd
import panel as pn
import hvplot.pandas

pn.extension(
    sizing_mode='stretch_width', template='fast', defer_load=True,
    loading_spinner='arc', loading_color='#0072B5'
)

count = pn.widgets.IntSlider(value=3, start=1, end=10, name="Count").servable(target="sidebar")

def big_load():
    time.sleep(3)
    return pd.DataFrame({"x": [1,2,3], "y": [1,2,3]})

def duration_no_sleep():
    return f"This text is important to show immediately {round(time.time()-start,2)}"

layout= pn.Column(big_load, pn.panel(duration_no_sleep, defer_load=False))

layout.servable()
```

MarcSkovMadsen:
In this example I get the warnings below. I would not expect to get those.

```python
import time
start = time.time()

import pandas as pd
import panel as pn
import hvplot.pandas

pn.extension(
    sizing_mode='stretch_width', template='fast',
    loading_spinner='arc', loading_color='#0072B5'
)

count = pn.widgets.IntSlider(value=3, start=1, end=10, name="Count").servable(target="sidebar")

def big_load():
    print("big_load")
    time.sleep(2)
    return pd.DataFrame({"x": [1,2,3], "y": [1,2,3]})

def duration():
    print("duration")
    time.sleep(2)
    return f"This text is important to show immediately {round(time.time()-start,2)}"

layout= pn.Column(pn.panel(big_load, defer_load=False), pn.panel(duration, defer_load=True))

layout.servable()
```

Warnings

```
WARNING:param.ParamFunction00953: The function 'big_load' does not have any dependencies and will never update. Are you sure you did not intend to depend on or bind a parameter or widget to this function? If not simply call the function before passing it to Panel. Otherwise, when passing a parameter as an argument, ensure you pass at least one parameter and reference the actual parameter object not the current value, i.e. use object.param.parameter not object.parameter.
2022-09-25 13:22:30,231 The function 'big_load' does not have any dependencies and will never update. Are you sure you did not intend to depend on or bind a parameter or widget to this function? If not simply call the function before passing it to Panel. Otherwise, when passing a parameter as an argument, ensure you pass at least one parameter and reference the actual parameter object not the current value, i.e. use object.param.parameter not object.parameter.
```

MarcSkovMadsen:
In this example I would expect the spinner to be shown for 2 seconds. But its only there for a split second.

https://user-images.githubusercontent.com/42288570/192145998-b3bc91f4-ebec-42c7-9773-352d749850e7.mp4

```python
import time
start = time.time()

import pandas as pd
import panel as pn
import hvplot.pandas

pn.extension(
    sizing_mode='stretch_width', template='fast',
    loading_spinner='arc', loading_color='#0072B5'
)

count = pn.widgets.IntSlider(value=3, start=1, end=10, name="Count").servable(target="sidebar")

def big_load():
    print("big_load")
    time.sleep(2)
    return pd.DataFrame({"x": [1,2,3], "y": [1,2,3]})

def duration():
    print("duration")
    time.sleep(2)
    return f"This text is important to show immediately {round(time.time()-start,2)}"

layout= pn.Column(pn.panel(big_load, defer_load=True), duration)

layout.servable()
```

philippjfr:
Thanks for testing, will look into the issues tonight.

philippjfr:
>The plots update really slowly compared to usual (I believe) when I drag the slider ??

This is because the duration callback blocks updates while time.sleep is running. Running the app with multiple threads (using `--num-threads`) will fix this.

>In this example I would expect the value of the duration_no_sleep output to be shown immediately, but it is not.

Again, ordering matters, the `big_load` callback is blocking so no other callbacks can execute. If you enable threading this works fine once again.

>In this example I get the warnings below. I would not expect to get those.

The warning comes from the callback without `defer_load`. I'm also wavering on how useful these warnings are in practice.

>In this example I would expect the spinner to be shown for 4 seconds. But its only there for a split second.

Looking into this one, it does seem like a bug.

MarcSkovMadsen:
**Comment 1 - referring to Feedback 1 above**

image

This does not make sense to me: `duration` should not be executed when I drag the slider.

** Comment 2 - referring to Feedback 2 above**

image

This does not make sense to me: I would expect to be able to specify some functions to be deferred (`big_load`) and some to be executed before the page loads (`duration_no_sleep`). The implementation I had in the issue would have worked in that way.

philippjfr:
Yeah, clearly didn't study those examples enough. Thanks for double checking!

Issue #3893, opened 2022-09-27
Make it easy to handle and report exceptions


MarcSkovMadsen:
## Request

Provide functionality to make exception handling and reporting easy

## Motivation

One of the things that make it hard to make Panel applications work efficiently is exception handling. Panel provides no functionality or guidance to make this easy.

As a consequence I have seen so many apps that just fails without the end user knowing anything about this. The end user does not know if the code actually executed or not because they don't see anything changing. As a consequence the users start distrusting the app and the developers do not get any feedback from users.

The developers can look into the logs and even setup automated monitoring (if they have access). But it requires infrastructure work to get there.

For me this seems just as important as the nice `deferred_load` functionality you just implemented @philippjfr 👍 

## Solutions

1. Document how to handle exceptions inside individual callbacks and report to a `TextArea` or via a `notification`.
2. Document how to build some reusable functionality to handle and report exceptions.
3. Provide built in functionality to make this easy [PREFERRED]

## My proposal

- enable developers to configure Panel via `pn.extension` to `catch_and_report_exceptions`.
  - The catched exceptions should be logged to server and notified to client by default. But this should be configurable.
  - enable developers to `pn.bind` and/ or `pn.depend` to a `config.exception` parameter for custom error handling.

## Concept Code

I don't know exactly how to implement this, for now I had to make this a context handler `with config.exception_handler():`. I believe this can be avoid in the final solution.

But this shows the principle. 

https://user-images.githubusercontent.com/42288570/192556205-af34a59d-2507-4436-bd37-3ba6a359d7c3.mp4

```python
import param
import logging
import panel as pn
from contextlib import contextmanager

class Config(param.Parameterized):
    exception = param.ClassSelector(class_=Exception)

    log_exception_to_server = param.Boolean(default=True)
    log_exception_to_client = param.Boolean(default=True)

    @contextmanager
    def exception_handler(self):
        try:
            yield None
        except Exception as ex:
            self.exception=ex
            if self.log_exception_to_server:
                logging.error("Error", exc_info=ex)
            if self.log_exception_to_client:
                pn.state.notifications.error(f'{ex}')


config = Config()

# ***** General config object above ****
# ***** General app below ****

import panel as pn

pn.extension(notifications=True, template="fast") # catch_and_handle_errors=True

button = pn.widgets.Button(name="Raise Exception").servable()

def _raise_exception(_):
    with config.exception_handler():
        raise ValueError("Some button error")

button.on_click(_raise_exception)

[...]

Issue #3574, opened 2022-05-29
Errors not logged when nthreads is enabled


maximlt:
Errors raised in callbacks dispatched with `--num-threads` enabled are not logged in the console.

```python
# app.py
import panel as pn

def cb(e):
    raise ValueError('Not logged')

b = pn.widgets.Button(name='Action')
b.on_click(cb)

b.servable()
```

served with `panel serve app.py --num-threads 2`

TheoMathurin:
Hope this will get attention, debugging without error logs is a real pain.

jbednar:
Any debugging help would be greatly appreciated!

TheoMathurin:
Ok, right now I don't know how to do so, but I will give it a try when I get a chance.

Hoxbro:
I'm still not seeing anything on master.

PR #3896, opened 2022-09-28
Add ability to configure exception handler

philippjfr:
Fixes https://github.com/holoviz/panel/issues/3574
Fixes https://github.com/holoviz/panel/issues/3893

### DIFF
--- a/doc/user_guide/Overview.md
+++ b/doc/user_guide/Overview.md
@@ -149,6 +149,7 @@ The `pn.config` object allows setting various configuration variables, the confi
 > - `loading_spinner`: The style of the global loading indicator, e.g. 'arcs', 'bars', 'dots', 'petals'.
 > - `loading_color`: The color of the global loading indicator as a hex color, e.g. #6a6a6a
 > - `defer_load`: Whether reactive function evaluation is deferred until the page is rendered.
+> - `exception_handler`: A custom exception handler can be defined which should accept any exceptions raised while processing events originating from the frontend and onload callbacks.
 > - `nthreads`: If set will start a `ThreadPoolExecutor` to dispatch events to for concurrent execution on separate cores. By default no thread pool is launched, while setting nthreads=0 launches `min(32, os.cpu_count() + 4)` threads.
 > - `raw_css`: List of raw CSS strings to add to load.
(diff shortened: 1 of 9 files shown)


philippjfr:
Not handling that in this PR.

MarcSkovMadsen:
Makes sense, will be a big improvement already.

philippjfr:
Question for you @MarcSkovMadsen. Should the handler accept just the exception or should it also get the event that caused the exception?

MarcSkovMadsen:
If it does not slow down Panel I think it should provide both the exception and the event because it would be hard to change the api later.

For my current use case I only need the exception.

philippjfr:
Looks like I have to punt on that anyway because it's a ton of work to make that information available in all cases.

philippjfr:
In future I can search for an explicit `event` argument in the signature and add it without breaking backward compatibility.

MarcSkovMadsen:
I just tested the main branch and it works as expected for me. You can find the test example code and a demo video in https://github.com/holoviz/panel/issues/3897#issue-1389402544.

MarcSkovMadsen:
There is a `.gif` for the release blog post here

panel-exception-handling

```python
import logging
import panel as pn

def exception_handler(ex):
    logging.error("Error", exc_info=ex)
    pn.state.notifications.error(f'{ex}')

pn.extension(exception_handler=exception_handler, notifications=True, template="fast")

pn.state.template.param.update(site="Awesome Panel", title="Easy and powerful exception handling added for production apps")

button = pn.widgets.Button(name="Raise Exception").servable()

def raise_exception(_=None):
    raise ValueError(f"Something went wrong when clicking the button")

button.on_click(raise_exception)

slider = pn.widgets.IntSlider(value=0, start=0, end=5, name="Raise exception on 3").servable()

@pn.depends(value=slider, watch=True)
def _raise_exception_2(value):
    if value==3:
        raise ValueError("Some slider error")

pn.pane.Markdown("""
```python
import logging
import panel as pn

def exception_handler(ex):
    logging.error("Error", exc_info=ex)
    pn.state.notifications.error(f'{ex}')

pn.extension(exception_handler=exception_handler, notifications=True, template="fast")
```
""").servable()
```

PR #3909, opened 2022-09-29
Add example of panel convert

### LINKED ISSUE #3905: Document example of running panel convert with requirements
Looking at https://pyviz-dev.github.io/panel/user_guide/Running_in_Webassembly.html I can see the reference api

image

but for example as a user I would have to guess if the requirements are space separated, comma separated or ...

Another question is if I should include `panel` or not. That is not clear either.

Another question is if I can version requirements?

Please add an example like

```bash
panel convert name-of-script.py --to pyodide-worker --prerender --requirements panel sklearn xgboost
```

MarcSkovMadsen:
Closes #3905


MarcSkovMadsen:
Image for this documentation. If you want this PR @philippjfr please copy this image to the CDN and update the docs with a link.

image

Issue #3920, opened 2022-10-01
Panel Sharing


MarcSkovMadsen:
## Request

Make it easy to share a link to my Panel app running in the browser:

```bash
panel.holoviz.org/sharing?<base64encoded string>
```

## Motivation

I'm inspired by

- [stlite](https://github.com/whitphx/stlite) Streamlit in the browser and [stlite sharing](https://edit.share.stlite.net/)
- [shiny live](https://shiny.rstudio.com/py/docs/shinylive.html)

I would also like an easy way to share Panel apps running entirely in the browser by simply sharing a link.

On online editor would also be really, really great and remove any friction to develop and share the kind of Panel apps that can run in the browser.

## Ideas

- An option to output a shareable link via `panel convert .... --shareable-link`.
- A way to get a short and maybe even meaningful link instead of the long base64 encoded string
- An online editor like [stlike sharing](https://edit.share.stlite.net/) that makes it easy to edit and share Panel apps. No need to install anything locally.
  - Ideally would support both a .py (VS Code) and an .ipynb (Jupyterlite) editor.
- A way for the community to create and share the apps (A Gallery)
- The base64encoded string might contain the information needed to run `panel convert` on it and return the app as a response. Alternatively the base64encoded string could contain the already converted app.
- It might require a live server to convert. It might also be possible to create a static page with some js that can actually handle and convert the base64encoded string to a live app???

MarcSkovMadsen:
I'm working on a POC in #3923

PR #3923, opened 2022-10-01
Panel Sharing

MarcSkovMadsen:
For now this is a *Proof of concept* and play ground for discussion around #3920 and the elements

- Panel converter and sharing endpoint
- `panel convert .... --share`.

For now its a Panel app that will render a `Template`.

It might be more efficient later to make it a Tornado event handler or a standalone FastAPI app.


MarcSkovMadsen:
The POC is there for an Editor + Sharing.

I gave up on the base64 encoded url idea. It would not work in practice for me. I would like to be able to share links, also on social media and they would be too long.

The basic things to figure out is

- how users can log in and verify their github account. I would like ot use their username as a part of the url.
- How I could/ should persist the source and target files when saved. Probably a blob storage or similar.
- Where to host the app it self.

Uploading panel-sharing.mp4…

MarcSkovMadsen:
As discussed I've moved this to awesome panel.

A prototype is running at awesome-panel.org.

image

image

The code is available at https://github.com/awesome-panel/panel-sharing

Issue #3939, opened 2022-10-03
App generated by convert_apps from a larger Panel app is not working.


MarcSkovMadsen:
I am working on creating a concept for easily sharing Panel apps live via `convert_apps`. In the beginning this worked out really, really well as `convert_apps` 1) worked 2) was super fast.

But at some point in time the converted app could no longer render - there was an error on the js side. The prerendering still looked fine. The app could still render if converted via `panel convert`.

To figure out the cause I compared the `.html` file generated by `convert_apps` and `panel convert`. I could see that the one generated by `convert_apps` contained traces of the `Ace` and a custom `ReactiveHTML` component. None of these where in the script I converted. But they where in the Panel app where I executed the `convert_apps` function by clicking a button.

My hypothesis is that this is the cause.

I tried to create a minimum reproducible example from scratch but could not. I hope to get back to to creating a minimum reproducible example later. But right now I want to focus on my app and I can execute `panel convert` via `subprocess.run`. It works but is much slower. So eventually if the Panel sharing app takes off I would need this fixed.

For now this is mostly for info.

-------

The app I was converting.

```python
import panel as pn

pn.panel("Hello World").servable()
```

philippjfr:
Looking into it. Initially I had a ton of these issues because all the apps were converted in the same process but then I switched to using multiprocessing to run each conversion in a different process. But I think because these processes are forked from the main process they are still not fully isolated.

MarcSkovMadsen:
Agree.

MarcSkovMadsen:
Ahh. Now I changed to `convert_app` instead of `convert_apps`.

Is `convert_app` running in the same process?

MarcSkovMadsen:
Problem solved for `convert_apps`. 

![image](https://user-images.githubusercontent.com/42288570/193608260-595befc6-e0d2-48db-9f69-c2189f92ed4b.png)

MarcSkovMadsen:
Should I keep this issue open because `convert_app` has the issue @philippjfr ?

MarcSkovMadsen:
And thanks for the effort. You have made so made big and small things here. I think the whole loading experience of Panel on Pyodide is amazing now.

This below video is speeded up twice. Its beautiful. I'm really enjoying the Panel experience. The `exception_handler` and the `notifications` work so great.

![panel-sharing-speedup](https://user-images.githubusercontent.com/42288570/193609852-b318c1a0-c769-4b99-9314-2c2c73aa95da.gif)

[...]

PR #3942, opened 2022-10-03
Explicitly request spawn start method in convert processes

philippjfr:
Attempts to fix https://github.com/holoviz/panel/issues/3939

At least for me 'spawn' is already the default start method so this PR makes no difference but it may be different on other systems.

@MarcSkovMadsen could you try this?

### DIFF
--- a/panel/io/convert.py
+++ b/panel/io/convert.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 
 import concurrent.futures
 import dataclasses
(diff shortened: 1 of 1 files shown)


philippjfr:
I'm going to merge just because being explicit here is a good thing regardless but if you could still test @MarcSkovMadsen that would be helpful.

MarcSkovMadsen:
I'm on linux so I believe this should change things according to

image

PR #3973, opened 2022-10-09
Improve support for requests in pyodide

philippjfr:
- Adds [`pyodide-http`](https://github.com/koenvo/pyodide-http) as a dependency for converted pyodide apps ensuring that `requests` and `urllib3` work (even for binary files if using the `pyodide-worker` target)
- Additionally we also patch `pd.read_parquet` and enable synchronous queries on the `Image` pane.

### DIFF
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -153,7 +153,7 @@ else:
     bokeh_req = f'{CDN_DIST}wheels/bokeh-{BOKEH_VERSION}-py3-none-any.whl'
 
 nbsite_pyodide_conf = {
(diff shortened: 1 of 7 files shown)


koenvo:
thank you for using pyodide-http!

Pyodide-http patches urllib. Under the hood urllib is used by pandas to read remote data. All read_* function should work out of the box after applying the patch.

philippjfr:
Nice, thanks! Was going to do some experimenting, particularly in regard to fsspec. I guess that also uses urllib under the hood in which case this is even more exciting than I'd dared to hope.

MarcSkovMadsen:
Wow. I've been dreaming about this. Thought it was not possible.

Issue #3970, opened 2022-10-08
panel serve ... --glob: NameError: name 'base_url' is not defined


MarcSkovMadsen:
panel==0.14.0

I'm trying to panel serve some apps with the `--glob` flag. When I go to the index page I get `NameError: name 'base_url' is not defined`.

## Reproducible example

Create two files `apps/app1.py`and `apps/app2.py` with the code

```python
import panel as pn

pn.panel("hello").servable()
```

```bash
panel serve apps/*.py --glob
```

When I navigate to `https://localhost:5006/ I see

```bash
2022-10-08 13:38:28,473 Starting Bokeh server version 2.4.3 (running on Tornado 6.2)
2022-10-08 13:38:28,474 User authentication hooks NOT provided (default user enabled)
2022-10-08 13:38:28,478 Bokeh app running at: http://localhost:5006/app1
2022-10-08 13:38:28,478 Bokeh app running at: http://localhost:5006/app2
2022-10-08 13:38:28,479 Starting Bokeh server with process id: 2153
2022-10-08 13:38:35,174 Uncaught exception GET / (::1)
HTTPServerRequest(protocol='http', host='..', method='GET', uri='/', version='HTTP/1.1', remote_ip='::1')
Traceback (most recent call last):
  File "/home/jovyan/repos/private/awesome-panel/panel-cli/.venv/lib/python3.9/site-packages/tornado/web.py", line 1713, in _execute
    result = await result
  File "/home/jovyan/repos/private/awesome-panel/panel-cli/.venv/lib/python3.9/site-packages/bokeh/server/views/root_handler.py", line 66, in get
    self.render(index, prefix=prefix, items=sorted(self.applications.keys()))
  File "/home/jovyan/repos/private/awesome-panel/panel-cli/.venv/lib/python3.9/site-packages/tornado/web.py", line 865, in render
    html = self.render_string(template_name, **kwargs)
  File "/home/jovyan/repos/private/awesome-panel/panel-cli/.venv/lib/python3.9/site-packages/tornado/web.py", line 1014, in render_string
    return t.generate(**namespace)
  File "/home/jovyan/repos/private/awesome-panel/panel-cli/.venv/lib/python3.9/site-packages/tornado/template.py", line 362, in generate
    return execute()
  File "/home/jovyan/repos/private/awesome-panel/panel-cli/_venv/lib/python3_9/site-packages/panel/io/__/_templates/index_html.generated.py", line 5, in _tt_execute
[...]

PR #3980, opened 2022-10-10
Fix panel serve index template

philippjfr:
Fixes https://github.com/holoviz/panel/issues/3970


MarcSkovMadsen:
I think I see the speedups from the new CDN in Panel 0.14. They seem more snappy.

I noticed pyodide is coming from jsdelivr. If you ever talk to those guys they might consider running their own cdn one day.

PR #3990, opened 2022-10-12
Add option to clear value of DatetimePicker

### LINKED ISSUE #3975: Clearing value of a DatetimePicker
Not sure if this is a bug or a new feature to Panel. Let's say I have a layout consisting of a button named "Edit", a DatetimePicker disabled with no default value, and a button named "Submit". At the time of initialization, the value of DatetimePicker is Null. The way these objects interact is as follows:
- Click "Edit" button, DatetimePicker is enabled so user can select a specific time value.
- Click "Submit" button, the selected time value will be pushed to the DB, and the DatetimePicker will be disabled and reset back to Null.

I have tried several ways with no success in clearing the value of the DatetimePicker.

```
time_widget = pn.widgets.DatetimePicker(disabled=True)
time_widget.value = now()

# how to set value back to None?
time_widget.value = None/pandas.NaT/np.nan => all causes error
```

Hoxbro:
Fixes #3975

This PR will make it possible to clear the value of DatetimePicker. 

Example:

``` python
import panel as pn
from datetime import datetime

pn.extension()

time_widget = pn.widgets.DatetimePicker()
btn = pn.widgets.Button(name="click")

@pn.depends(btn, watch=True)
def tmp(*event):
    if time_widget.value:
        time_widget.value = None
    else:
        time_widget.value = datetime.today()


pn.Column(btn, time_widget).servable()
```

https://user-images.githubusercontent.com/19758978/195400548-826677c5-6e73-4763-a133-b1059ef2c875.mp4

TODO:
- [x] Write unit test for it
- <s>Add support for other np.nan / pd.NaT </s>

### DIFF
--- a/panel/models/datetime_picker.py
+++ b/panel/models/datetime_picker.py
@@ -10,7 +10,7 @@ class DatetimePicker(InputWidget):
 
     '''
 
-    value = String(help="""
+    value = Nullable(String, help="""
     The initial or picked date.
     """)
 
--- a/panel/models/datetime_picker.ts
+++ b/panel/models/datetime_picker.ts
@@ -36,7 +36,7 @@ export class DatetimePickerView extends InputWidgetView {
 
     const {value, min_date, max_date, disabled_dates, enabled_dates, position, inline,
       enable_time, enable_seconds, military_time, date_format, mode} = this.model.properties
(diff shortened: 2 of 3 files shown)


philippjfr (approved):
Looks good!

PR #3992, opened 2022-10-13
Add compute_hash in panel.io.__init__ imports

TheoMathurin:
The `compute_hash` function is documented as part of the public API but was not imported in panel.io.\_\_init__

### DIFF
--- a/panel/io/__init__.py
+++ b/panel/io/__init__.py
@@ -4,7 +4,7 @@ model state, and rendering panel objects.
 """
 import sys
 
-from .cache import cache  # noqa
+from .cache import cache, compute_hash  # noqa
 from .callbacks import PeriodicCallback  # noqa
 from .document import init_doc, unlocked, with_lock  # noqa
 from .embed import embed_state  # noqa


philippjfr:
Where's it documented? I would much prefer not to expose it here.

TheoMathurin:
It appears [in the docs](https://panel.holoviz.org/api/panel.io.html#panel.io.cache.compute_hash) and is behind a "Public API" [banner](https://github.com/holoviz/panel/blob/master/panel/io/cache.py#L245) in the source code so I assumed it was a mere omission.

For my custom implementation where I use the cache conditionally, I need to be able to pop items from the cache and therefore to compute hash values the same way it is done in the cache module.

philippjfr:
>It appears in the docs and is behind a "Public API" banner in the source code so I assumed it was a mere omission. Sorry if it's not, I should have file an issue for a discussion first.

Fair enough, I should move it out of that section but at the same time even if we consider it public API at the `panel.io.cache` level doesn't mean it should necessarily be exposed that the `panel.io` level. 

>For my custom implementation where I use the cache conditionally, I need to be able to pop items from the cache and therefore to compute hash values the same way it is done in the cache module.

I don't have any issue continuing to treat it as public API in that case but still do not want to expose it in `panel.io`

TheoMathurin:
Alright.

Tell me if I'm wrong, but for the time being you can't access this function, since what is imported with `import panel.io.cache` is the `cache` function and not the `panel.io.cache` module.

philippjfr:
Not sure I understand what you mean, why can't you do `from panel.io.cache import compute_hash`

jbednar:
That fails because panel.io clobbered the `cache` module with the `cache` function: `from .cache import cache`

image

philippjfr:
>That fails because panel.io clobbered the cache module with the cache function: from .cache import cache

That's not how module imports in Python work. You must just have an older version of Panel.

jbednar:
D'oh! I was surprised I hadn't encountered such clobbering before, and indeed it's because my installed and source versions of Panel were different. In any case I'd argue it's a bad idea to overwrite `cache` in the namespace, regardless.

philippjfr:
Don't necessarily agree this is bad practice when the module is not intended to be user-level API. That said even if were to agree I cannot deprecate `pn.io.state` or `pn.io.cache` without a major version bump.

Issue #3976, opened 2022-10-10
Tabulator sorts columns as string instead of number if initial dataframe is empty


xavArtley:
version panel 0.13.1

I need to initialize and empty Tabulator, then a dataframe is inserted with some numerical columns

I'd like to sort these numerical columns hower they are processed as string instead of number on tabulator side

Is there a way to define the type of columns at instantiation?

MRE:
```python
import json
import panel as pn
import pandas as pd
from bokeh.models.widgets.tables import NumberEditor, SelectEditor
pn.extension("tabulator")

data = '{"start":{"0":0,"1":96,"2":116},"end":{"0":79,"1":102,"2":123},"type":{"0":"Standard","1":"Outlier","2":"Outlier"},"use":{"0":"Train","1":"Test","2":"Test"}}'
df = pd.DataFrame.from_dict(json.loads(data))
df_init = pd.DataFrame([], columns=["start", "end", "type", "use"]).astype({"start":int, "end": int, "type": str, "use": str})
options = {"editors": {
        "start": NumberEditor(),
        "end": NumberEditor(),
        "type": SelectEditor(options=["Standard", "Outlier"]),
        "use": SelectEditor(options=["Train", "Test", "Train/Test", ""]),
    },
    "widths": {"use": 80},
    "selectable": "checkbox",
    "show_index": False,
    "height": 150,
    "sorters": [
        {"field": "start", "dir": "desc"},
    ]
}
tab = pn.widgets.Tabulator(df_init, **options)
pn.Row(tab, pn.widgets.Tabulator(df, **options))
```
in next cell:
```python
tab.value=df
tab.value.dtypes
```
result:
![tabulator_sort](https://user-images.githubusercontent.com/18531147/194853667-c880f548-60b0-4113-9142-32976224fc55.gif)

PR #3999, opened 2022-10-14
Sorting number rather than string in tabulators

liu-kan:
Tabulator JS has its [Built In Sorters](http://tabulator.info/docs/5.4/sort#func-builtin) for data other than string.
Default behaviours for sorting data, include numbers, in panel tabulator, is treating them as string. 
This PR fix this issue.


maximlt (approved):
This is the change I also found that was required to fix https://github.com/holoviz/panel/issues/3976, thanks a lot @liu-kan !!!

PR #4003, opened 2022-10-14
Tabulator: re-introduce headerSort in the docs

maximlt:
Documenting `columnDefaults` seems relevant.

```python
pn.widgets.Tabulator(df, configuration={
    'clipboard': True,
    'rowHeight': 50,
    'columnDefaults': {
        'headerSort': False,
    },
})
```


philippjfr:
Agree that documenting `columnDefaults` is a better approach but please also update the text describing the options above the code cell.

philippjfr (approved):
Perfect, thanks!

PR #4013, opened 2022-10-16
Bump setup-python and install actions

maximlt:
Committed via https://github.com/asottile/all-repos

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -54,7 +54,7 @@ jobs:
       # it as one of the sources.
       PYCTDEV_SELF_CHANNEL: "pyviz/label/dev"
     steps:
-      - uses: pyviz-dev/holoviz_tasks/install@v0.1a2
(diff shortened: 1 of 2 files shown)


.github/workflows/test.yaml
```diff
@@ -109,7 +109,7 @@ jobs:
       # it as one of the sources.
       PYCTDEV_SELF_CHANNEL: "pyviz/label/dev"
     steps:
-      - uses: pyviz-dev/holoviz_tasks/install@v0.1a2
+      - uses: pyviz-dev/holoviz_tasks/install@v0.1a4
```
philippjfr, line 112: Different versions?

maximlt: Good catch thanks that was an oversight. I wanted anyway to wait until you release 0.15.1 to make these changes since they're just meant to remove some deprecation warnings emitted by GitHub.

Issue #4026, opened 2022-10-19
Incorrect type annotation for `ServableMixin`.


MarcSkovMadsen:
panel==0.14.0

## Reproduce

```python
import panel as pn

component = pn.pane.Markdown("Hello").servable()

pn.Column(component.param.object).servable()
```

```bash
$ mypy script.py
script.py:5: error: "ServableMixin" has no attribute "param"
Found 1 error in 1 file (checked 1 source file)
```

## Cause

The return type should not be `ServableMixin` but the generic type of `self`.

![image](https://user-images.githubusercontent.com/42288570/196621963-f3bb01c4-0a21-4c21-9294-087dd4c87f4d.png)

![image](https://user-images.githubusercontent.com/42288570/196622036-c18028ee-8490-4e0f-b95e-1315ab087512.png)

PR #4027, opened 2022-10-19
Fix typing of .servable method

philippjfr:
Fixes https://github.com/holoviz/panel/issues/4026

### DIFF
--- a/panel/viewable.py
+++ b/panel/viewable.py
@@ -29,6 +29,7 @@ from bokeh.document import Document
 from bokeh.resources import Resources
 from jinja2 import Template
 from pyviz_comms import Comm, JupyterCommManager  # type: ignore
+from typing_extensions import Self
 
 from .config import config, panel_extension
 from .io import serve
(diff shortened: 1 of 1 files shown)


panel/viewable.py
```diff
@@ -326,7 +327,7 @@ def _on_stdout(self, ref: str, stdout: Any) -> None:
     def servable(
         self, title: Optional[str] = None, location: bool | 'Location' = True,
         area: str = 'main', target: Optional[str] = None
-    ) -> 'ServableMixin':
+    ) -> Self:
```
mattpap, line 330: As far as I can tell, mypy doesn't support type checking this yet (pyright does though).

philippjfr: Okay, so I should use:

```python
_Self = TypeVar('_Self', bound='ServableMixin')
```

for now?

mattpap: Yes, sadly that's the only way for now. mypy seems quite behind the curve when it comes to implementing new PEPs.

PR #4042, opened 2022-10-21
Allow passing parameters by reference

philippjfr:
This is something I've been wanting to experiment with for the longest time (specifically I suggested it here: https://github.com/holoviz/param/issues/363). While I still think this might best be implemented at the Param level I think I would like to experiment with supporting it in Panel at first. The basic idea is that you can pass a parameter instance (or widget) to a constructor and it will treated as a reference, i.e. it will automatically link the parameters:

As a simple example here we create a slider that will control the page_size of a Tabulator widget:

```python
page_size = pn.widgets.IntSlider(start=1, end=10)
pn.widgets.Tabulator(df, page_size=page_size, pagination='remote')
```

Screen Shot 2022-10-21 at 12 19 48

This means that in many cases you won't have to use the equivalent `pn.bind` expression anymore:

```python
pn.bind(pn.widgets.Tabulator, df, page_size=page_size, pagination='remote')
```

It also means that we really have another API for building reactive apps but to me it's by far the most natural way of working. The other major benefit of supporting this is that it's much easier to inspect and serialize such an object than a `pn.bind` expression would be which is important for some usecases (e.g. Lumen).

### DIFF
--- a/panel/reactive.py
+++ b/panel/reactive.py
@@ -465,6 +465,34 @@ class Reactive(Syncable, Viewable):
     the parameters to other objects.
     """
 
+    def __init__(self, **params):
+        params, refs = self._extract_refs(params)
+        super().__init__(**params)
+        self._setup_refs(refs)
+
+    def _extract_refs(self, params):
+        from .depends import param_value_if_widget
+        processed, refs = {}, {}
+        for pname, value in params.items():
+            if pname not in self.param:
+                processed[pname] = value
+                continue
+            value = param_value_if_widget(value)
+            if isinstance(value, param.Parameter):
+                refs[pname] = value
+                value = getattr(value.owner, value.name)
+                if self.param[pname].allow_None and value is None:
+                    continue
+            processed[pname] = value
+        return processed, refs
+
+    def _setup_refs(self, refs):
+        groups = defaultdict(list)
+        for pname, p in refs.items():
+            groups[p.owner].append((pname, p.name))
+        for owner, pnames in groups.items():
+            self.link(owner, bidirectional=True, **dict(pnames))
+
     #----------------------------------------------------------------
     # Public API
     #----------------------------------------------------------------


MarcSkovMadsen:
Can you also provide a pn.depends or pn.bind function?

Can you just provide the widget and not widget.param.value? Please consider treating any parameterized with a value parameter as a widget. That would make using custom widgets based on Viewer or ReactiveHTML easier to use.

philippjfr:
Yes to all of those. At least that's the intention.

jbednar:
Sounds great to me. I'd vote for it to be supported at the Param level rather than the Panel level, though.

maximlt:
To be honest, my first reaction was "Oh no another API!" 🙃 I see users struggling on Discourse with the various Panel APIs, and it gets worse when they want to combine it with hvPlot/HoloViews API to create dashboards with dynamic visualizations.

Then I thought that this ship has maybe already sailed, at least in hvPlot, as it's already to provide [widgets and bound functions](https://hvplot.holoviz.org/user_guide/Widgets.html#using-widgets-as-arguments) to quickly create dashboard. However, I don't know to which extent this feature is used (anecdotally, I discovered it pretty recently).

Thinking about typing (statically or dynamically with Param), I realize that the types listed in hvPlot's docstrings are incomplete, even after @MarcSkovMadsen's massive update :( Since they should indicate that for instance `x` accepts widgets (well, widgets that hold a string value...) and bound functions. Having docstrings that correctly represent the types accepted is without any doubt a good way to let users know about that kind of API. This is true for hvPlot, and true for Panel too. This might also make static type checkers even unhappier than they currently are with Param parameters.

> This means that in many cases you won't have to use the equivalent `pn.bind` expression anymore:

I think I'd like to see more compelling examples than the one provided in the original post (I'll also try to think about some more).

Overall, I'm not against this new API. I think though that there needs a be a clear path on how to integrate it within the documentation, i.e. in the guides, the galleries and API reference. This should be a hard constraint for adding a new API, as the site currently shows a mix of API usages. Of course I'm happy to help updating the website to showing usage of the recommended API, when this is decided upon (could be a good idea for the Pydata Global spring @droumis).

One thing I like for sure with this new API is that until now `pn.bind` didn't offer me a way to properly control some widgets properties, and I had to use `@pn.depends(widget, watch=True)` to update them, as in the example below:

```python
w1 = pn.widgets.TextInput()
w2 = pn.widgets.TextInput()
b = pn.widgets.Button(name='submit')

def submit(e):
    print('submit', w1.value, w2.value)

b.on_click(submit)

@pn.depends(w1, w2, watch=True)
def _enable_button(val1, val2):
    b.disabled = not val1 or not val2

pn.Column(w1, w2, b)
```

I think that new API makes that nicer to write (I'm not a big fan of `@pn.depends` that adds a layer of indirection):

```python
w1 = pn.widgets.TextInput()
w2 = pn.widgets.TextInput()

def _disabled_control(val1, val2):
    return not val1 or not val2

b = pn.widgets.Button(name='submit', disabled=pn.bind(_disabled_control, w1, w2))

def submit(e):
    print('submit', w1.value, w2.value)

b.on_click(submit)
pn.Column(w1, w2, b)
```

> Allow passing parameters by reference

We'll have to be careful with how this is named and documented. *parameter* is a a generic term for most but it has a peculiar meaning in the HoloViz ecosystem. Of course I know what you mean but that's just because I've been HoloViz-brainwashed 🙃 Again, this might just be a matter of better documenting things, and that might just be the most important thing to consider. Cross-referencing https://github.com/holoviz/holoviz/issues/315

jbednar:
> I have some concerns about that because a Parameter/Widget/Bound function could well passed as a parameter by value rather than by reference. 

In many cases the declared type of the parameter would ensure that such objects wouldn't be valid by value anyway, so we'd just have to worry about the case where it is ambiguous. In particular, Dynamic Parameters, including Numbers, accept a callable and expect it to return a value of the declared type, so we'd have to be able to distinguish between one of those three items and a generic callable. That should be possible though, right?

The other worry would be if someone is using an untyped param.Parameter, but in that case the advice would just be "don't do that -- use a more specific type", right?

PR #4047, opened 2022-10-22
Add Gallery VideoStream example

### LINKED ISSUE #4046: Make it easy and useful to use the VideoStream
I've never seen any showcases of the `VideoStream` in Panel. As Panel can update relatively quickly via Tornado and websockets it should be the best data app framework for Video. And with `panel convert` even better.

But there are lot of issues

- There are no inspirational examples out there
- When you try out you run into several issues
  - The VideoStream transfers a high resolution image
  - The VideoStream transfers a .png base64 data url. It is slow to work with in Pillow and other frameworks. JPEG is much faster to load and transform
  - You need to be able to convert from base64 to PIL.Image for using Pillow and np.ndarray for using scikit-learn etc. This might be easy for experienced data vision people. But for beginners this is so hard.

To make it easier for Panel users to get started and create something useful. I propose we add an example to the Gallery that people can use for inspiration. We can use the example to identify and solve any performance issues.

The example should end up being easy to understand and use + super performant.

## Additional Context

[Yuichiro](https://twitter.com/whitphx) has done some really amazing work to make this possible in Streamlit. We should be able to find inspiration and make something at least as awesome. See https://edit.share.stlite.net/?sampleAppId=realtime_image_processing

[image](https://twitter.com/whitphx/status/1583708572063772675?s=20&t=lZLbNEIl7LQpVWu1RyNEiQ)

MarcSkovMadsen:
Closes #4046

## Server

The video shows the performance when running on the server. The key to getting this performance is setting the `VideoStream` format to `jpeg`. `png` is slow.

https://user-images.githubusercontent.com/42288570/197378169-b2de8538-efac-4924-8628-a5da87815d15.mp4

You can get more than 10fps. This includes a round trip from DK to Amsterdam and back. That is pretty amazing. And it means you can use deep learning frameworks like PyTorch that are not available in Webassembly.

## Webassembly

The video shows the performance when converted to webassembly. Its fucking fast! (sorry)

https://user-images.githubusercontent.com/42288570/197385376-824c8afc-7dd4-4c70-bd1f-d9e2704a1aa5.mp4

### Online

https://awesome-panel.github.io/examples/ (Works in Chrome, not firefox)

### Panel Serve

```bash
pip install panel  numpy scikit-image
```

```bash
panel serve examples/gallery/streaming/video/app.py --autoreload
```

### Panel Convert to webassembly

```bash
panel convert examples/gallery/streaming/video/app.py --to pyodide-worker --out pyodide --requirements examples/gallery/streaming/video/requirements.txt --watch
```

```bash
python -m http.server
```

open [http://localhost:8000/pyodide/app.html](http://localhost:8000/pyodide/app.html).

### Issues identified and solved

- [x] The default `.png` format of the stream is slow to work with. `.jpeg` is much better.
- [x] The server can get "behind" the videostream. This should be avoided.
- [x] When using the face recognition algorithm the colors get strange. They are not like that in the [Yuichiro example](https://edit.share.stlite.net/?sampleAppId=realtime_image_processing).

### Potential future improvements

Someday someone might be willing to invest in

- [] Transfering reduced size video stream from client to server. Would require adjusting the videostream.ts file.
- [] There might be a more efficient form of communication than base64 data urls from client to server?
- [] There might be a more efficient form of communication than a jpeg file buffer from server to client embedded in a JPEG pane. Maybe a HoloViews Image is faster?
- [] Enabling interaction with the video output. Like clicking or drawing.
- Someone could build a `panel-videostream` package that provides the reusable code from this example such that for users they only have to implement a `CustomTransform` in 2-10 lines of code.
    - Gradio plans to provide this in https://github.com/gradio-app/gradio/issues/1637
    - It will probable work similarly to https://gradio.app/real_time_speech_recognition/
- Getting it to work in Firefox. When loading scikit-image it throws an out of memory error.

## Credits

[Yuichiro](https://twitter.com/whitphx) for inspiration


philippjfr:
>The video shows the performance when converted to webassembly.

Looking good 😉 Definitely want to merge this one, can you remove the built app.html and app.js? They will get built automatically.

We do have to work out how to link the pyodide versions of all the gallery examples from the docs though.

MarcSkovMadsen:
Yes. I will remove them. But also plan to convert to notebook. But need to solve some of the issues first

MarcSkovMadsen:
The app is now working really great. Next step is to turn it into a Notebook.

MarcSkovMadsen:
First version of notebook created.

MarcSkovMadsen:
The notebook is now ready for review.

---

The assets below can be used for promotion if relevant.

streaming-videostream-speedup

https://user-images.githubusercontent.com/42288570/208600967-e01c9884-132e-44ed-8b9d-5ff6600396ec.mp4

setup.py
```diff
@@ -190,6 +190,7 @@ def run(self):
         'xgboost',
         'ipyvuetify',
         'reacton'
```
Hoxbro, line 192:
```suggestion
        'reacton',
```

MarcSkovMadsen:
I can see that the tests fail because it tries to run the VideoStream notebook with the kernel `conda-root-py`. It seems not to exist. Where does this kernel name come from and what should I do to fix @philippjfr ?

philippjfr:
> I can see that the tests fail because it tries to run the VideoStream notebook with the kernel conda-root-py. It seems not to exist. Where does this kernel name come from and what should I do to fix @philippjfr ?

That's because you didn't clear out the metadata in the notebook. Will do that shortly.

philippjfr:
pre-commit.ci autofix

philippjfr (approved):
Amazing example and great work as always @MarcSkovMadsen, thanks!

PR #4053, opened 2022-10-24
Fix FloatInput notebook documentation for `value_throttled`

sjdemartini:
The doc (https://panel.holoviz.org/reference/widgets/FloatInput.html) mistakenly contained the description for `value` again instead of `value_throttled`.

The description here was copied from `widgets/input.py`.


philippjfr:
Thank you @sjdemartini!

PR #4056, opened 2022-10-25
Silence EMPTY_LAYOUT warnings

maximlt:
The `EMPTY_LAYOUT` warning is emitted when an empty layout (e.g. `pn.Row()`) is rendered. Panel use cases are different than Bokeh's ones, Panel is more used to serve interactive apps where it's perfectly valid to render an empty layout that is later populated with some content. In this case this warning is just noise.

The ability to silence specific Bokeh warnings was added to Bokeh a few years ago https://github.com/bokeh/bokeh/pull/8739. This PR is using the `silence` function to silence this warning, for both `panel serve` and `pn.serve()` usages.

### DIFF
--- a/panel/command/serve.py
+++ b/panel/command/serve.py
@@ -19,6 +19,8 @@ from bokeh.application.handlers.document_lifecycle import (
 from bokeh.application.handlers.function import FunctionHandler
 from bokeh.command.subcommands.serve import Serve as _BkServe
 from bokeh.command.util import build_single_handler_applications
+from bokeh.core.validation import silence
+from bokeh.core.validation.warnings import EMPTY_LAYOUT
 from bokeh.server.contexts import ApplicationContext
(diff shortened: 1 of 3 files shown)


mattpap:
I would prefer to remove that warning (even as soon as now, i.e. 3.0). I always found it annoying that we warn about a perfectly valid behavior, which is counter productive at best.

maximlt:
I've seen a rather long discussion on Bokeh about this warning, with some being in favor of removing it, and some being in favor of keeping it, but that was some years ago so opinions may have changed!

mattpap:
Here's https://github.com/bokeh/bokeh/issues/8725 the discussion. I will submit a PR.

philippjfr:
Still going to merge this but thankfully we can then remove it again on the 3.0 branch.

PR #4062, opened 2022-10-27
Add Python 3.10 and 3.11 support

maximlt:
Python 3.11 just got out, let's try to get the CI running on 3.10 again.

EDIT:
This PR says that it's adding Python 3.10 and 3.11 support but all that it's really doing is:

- setup the CI to run on these versions, which at the time of writing involves not running the examples! As even on Python 3.10 all the dependencies aren't yet available on conda-forge.
- make the test suite a little more robust by doing some minor changes here and there in the test suite
- make the CI only run on 3.7, 3.9 and 3.11, except on scheduled, tag, and manually dispatched events where it runs the full set from 3.7 to 3.11.
- run the docs Markdown part of the examples tests, as they need the examples dependencies.


philippjfr:
Let's use this PR to also fix the rest of the CI. It's in pretty poor shape right now.

maximlt:
> Let's use this PR to also fix the rest of the CI. It's in pretty poor shape right now.

I agree.

philippjfr:
Looking good, and thanks for doing this!

maximlt:
It's not yet ready!

philippjfr:
Just showing appreciation for you 🤗

philippjfr (approved):
Looks good, thank you!

maximlt:
Oh it just failed when I wrote the last message :( Seems there are intermittent install failures on Windows / Python 3.7. I might just remove it from the matrix. Anyway, interested in your feedback on this.

Issue #4088, opened 2022-11-07
Panel ipywidgets interaction stops working after a new ipywidget is mounted on the page


govinda18:
#### ALL software version info
panel master

#### Description of expected behavior and the observed behavior
In the code below, after clicking Append Text, Update Text button stops working

#### Complete, minimal, self-contained example code that reproduces the issue

```
import panel as pn

import ipywidgets as w

pn.extension()

a = w.HTML("Asd")
b = w.HTML("Qwe")

x = pn.Column()

btn1 = pn.widgets.Button(name="Update Text")
btn2 = pn.widgets.Button(name="Append Text")

def f1(_):
    a.value = "asdasd" if a.value == "qweqew" else "qweqew"

def f2(_):
    x.clear()
    x.append(b)
    
btn1.on_click(f1)
btn2.on_click(f2)

pn.Row(a, x, btn1, btn2).servable()

```

PR #4091, opened 2022-11-08
Ipywidget comm breaks when a new ipywidget is rendered

govinda18:
Fixes https://github.com/holoviz/panel/issues/4088

This commit drops explicitly calling ipywidget.comm.open() as it causes a protocol version mismatch issue - reproducer in the referenced issue. When opening a new comm, as per the [docs](https://github.com/jupyter-widgets/ipywidgets/blob/7.6.x/packages/schema/messages.md#instantiating-a-widget-object-1), we need to pass data and meta data. So the w.comm.open() call should be something like:

```python
metadata = {"version": "2.0.0"}
data = {"state": w.get_state()}
w.comm.open(data=data, metadata=metadata)
```

This seems to fix the error but the widgets still do not update. This is because when we create a new comm, ipywidgets seems to create a new model [reference](https://github.com/jupyter-widgets/ipywidgets/blob/74774dae5becb9f4781c3438a5fece1f8ca60415/packages/base-manager/src/manager-base.ts#L252). In my understanding, we need not explicitly open a comm to the frontend as the kernel will automatically reconstruct an existing comm. If for some reason we do want to open a comm, then ipywidgets bokeh needs to be updated so that the view uses the correct model.

### DIFF
--- a/panel/pane/ipywidget.py
+++ b/panel/pane/ipywidget.py
@@ -63,7 +63,6 @@ class IPyWidget(PaneBase):
                 widgets = (obj._active_widgets if hasattr(obj, '_active_widgets') else obj.widgets).values()
                 for w in widgets:
                     w.comm.kernel = kernel
-                    w.comm.open()
 
         model = IPyWidget(widget=obj, **kwargs)
         return model
--- a/panel/pane/ipywidget.py
+++ b/panel/pane/ipywidget.py
@@ -63,6 +63,8 @@ class IPyWidget(PaneBase):
                 widgets = (obj._active_widgets if hasattr(obj, '_active_widgets') else obj.widgets).values()
                 for w in widgets:
                     w.comm.kernel = kernel
+                    if w.comm._closed:
+                        w.comm.open()
 
         model = IPyWidget(widget=obj, **kwargs)
         return model


philippjfr:
Thanks @govinda18. Would love to sync with you some time to survey the current issues with ipywidgets + bokeh/panel and then make a plan for resolving them.

govinda18:
> Thanks @govinda18. Would love to sync with you sometime to survey the current issues with ipywidgets + bokeh/panel and then make a plan for resolving them.

Surely, feel free to send an invite to govindamaheshwari21@gmail.com.

govinda18:
@philippjfr let me know whenever you are free to catch up. My colleague @mlucool would also love to join at udoff@deshaw.com.

Also, let me know if you have any feedback on the PR.

philippjfr:
Together with https://github.com/bokeh/ipywidgets_bokeh/pull/69/ this fixes the issue.

panel/pane/ipywidget.py
```diff
@@ -63,7 +63,6 @@ def _get_ipywidget(self, obj, doc, root, comm, **kwargs):
                 widgets = (obj._active_widgets if hasattr(obj, '_active_widgets') else obj.widgets).values()
                 for w in widgets:
                     w.comm.kernel = kernel
```
philippjfr, line 65:
```suggestion
                    w.comm.kernel = kernel
                    if w.comm._closed:
                        w.comm.open()
```

PR #4092, opened 2022-11-08
Make panel convert help text more useful

MarcSkovMadsen:
Small improvement so that we don't have to lookup things in the docs


philippjfr:
Thanks @MarcSkovMadsen!

PR #4093, opened 2022-11-08
Add support for tqdm's process_map

peterfpeterson:
There are a few missing functions on the Tqdm object that are needed then it works

### DIFF
--- a/panel/tests/widgets/test_tqdm.py
+++ b/panel/tests/widgets/test_tqdm.py
@@ -19,6 +19,18 @@ def test_tqdm():
     assert isinstance(tqdm.layout, pn.Row)
 
 
+def test_process_map():
+    tqdm_obj = Tqdm()
+    # make sure the bar starts at zero
+    assert tqdm_obj.value == 0
+
+    NUM_ITEMS = 10
+    # run process map to sleep .3 seconds for each of ten items
+    _ = process_map(time.sleep, [0.3] * NUM_ITEMS, max_workers=2, tqdm_class=tqdm_obj)
+    # make sure the bar finishes where it should
+    assert tqdm_obj.value == NUM_ITEMS
+
+
 def test_tqdm_leave_false():
     tqdm = Tqdm(layout="row", sizing_mode="stretch_width")
 
--- a/panel/widgets/indicators.py
+++ b/panel/widgets/indicators.py
@@ -18,6 +18,7 @@ How to use indicators
 from __future__ import annotations
 
 import math
+from multiprocessing import Lock
 import os
 import sys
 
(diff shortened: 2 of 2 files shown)


KedoKudo:
@MarcSkovMadsen Could you take a look at this PR?  It is a feature expansion for the existing tqdm widget. :pray:

panel/widgets/indicators.py
```diff
@@ -1137,6 +1138,9 @@ class Tqdm(Indicator):
     layout = param.ClassSelector(class_=(Column, Row), precedence=-1, constant=True, doc="""
         The layout for the text and progress indicator.""",)
 
+    lock = param.ClassSelector(class_=object, default=None,
+                               doc="""The layout for the text and progress indicator.""",)
```
MarcSkovMadsen, line 1142: The doc string looks incorrect.

panel/tests/widgets/test_tqdm.py
```diff
@@ -19,6 +19,18 @@ def test_tqdm():
     assert isinstance(tqdm.layout, pn.Row)
```
MarcSkovMadsen, line 21: It would be very nice if an example could be included in the TQDM reference docs?

At least please include a simple minimum, reproducible Panel app example in this PR, that shows how this can be used in a Panel app. Then its easy to later include in the docs if needed.

Thanks.

panel/widgets/indicators.py
```diff
@@ -1137,6 +1138,9 @@ class Tqdm(Indicator):
     layout = param.ClassSelector(class_=(Column, Row), precedence=-1, constant=True, doc="""
         The layout for the text and progress indicator.""",)
 
+    lock = param.ClassSelector(class_=object, default=None,
```
MarcSkovMadsen, line 1141: @philippjfr would know much better than me if there could be any unanticipated sideeffects of using a Lock like this with Panel?

panel/tests/widgets/test_tqdm.py
peterfpeterson: Is there a place in the docs I should edit as part of this PR? I'm looking for a path in the repo.

The other option is to add an example to the PR description. I'm unclear on what form the example should be.

peterfpeterson:
Here is a demo ipynb (rename the file)

[progress_bar_demo.ipynb.txt](https://github.com/holoviz/panel/files/9981314/progress_bar_demo.ipynb.txt)

panel/widgets/indicators.py
```diff
@@ -18,6 +18,7 @@
 from __future__ import annotations
 
 import math
+from multiprocessing import Lock
```
philippjfr, line 21: `multiprocessing` can't be imported in Pyodide environments so we have to find a different solution here or guard the import.

panel/widgets/indicators.py
philippjfr: No real problem we just have to handle the pyodide case gracefully.

peterfpeterson:
The last commit should resolve the reviewer comments.

philippjfr (approved):
Thanks for applying the requested changes. Looks good to merge now.

PR #4105, opened 2022-11-11
perspective: Add documentation about deleting rows

geronimogoemon:
Following up on the discussion present here https://discourse.holoviz.org/t/delete-row-from-perspective/4537 I am adding small explanation on a proper way to remove rows from Perspective.


philippjfr:
Thanks @geronimogoemon.

philippjfr (approved):
Looks good.

PR #4107, opened 2022-11-11
Remove unused imports

joelostblom:
These are not used anywhere on the page as far as I can see.


philippjfr:
Thanks @joelostblom

PR #4115, opened 2022-11-14
Improve detection of comms in vscode and colab

### LINKED ISSUE #3767: Automatic `comms` detection is sensitive to import order
First noticed here https://github.com/holoviz/holoviews/pull/5398

When `hv.extension()` (or a import of `hvplot`) is done before `pn.extension()` it does not automatic set the current `comms` configuration.

| | Colab | vscode |
|-|---------|-----------|
|`holoviews` / `hvplot` import first | Screenshot 2022-08-22 10 49 59 | Screenshot 2022-08-22 10 55 41 |
|`panel` import first |  Screenshot 2022-08-22 10 50 49 | Screenshot 2022-08-22 10 55 13 |  

The offending code seems to be here:

https://github.com/holoviz/panel/blob/4f8ab43dac9ca5971f5bd84c35e7882ca676e541/panel/config.py#L597-L599

This could be easy to fix, but I'm not sure if this will break some other underlying logic.

Hoxbro:
Fixes #3767

Before this PR if `hv.extension` was called before `pn.extension` it will be set  `pn.config.comms` to `"default"` and not `vscode` in vs code or `colab` in Google Colab. 

This PR refactors the detection into its own method and always calls in a notebook. On the first run, it will auto-detect the comms and replace `None` with one of `default`, `ipywidgets`, `vscode`, or  `colab`. Then in the following calls, it will only change the `comms` if it is explicitly set.

image

I'm unsure if the order of `pyodide` and `vscode` should be swapped.

### DIFF
--- a/panel/config.py
+++ b/panel/config.py
@@ -166,7 +166,7 @@ class _config(_base_config):
     _admin = param.Boolean(default=False, doc="Whether the admin panel was enabled.")
 
     _comms = param.ObjectSelector(
-        default='default', objects=['default', 'ipywidgets', 'vscode', 'colab'], doc="""
+        default=None, objects=[None, 'default', 'ipywidgets', 'vscode', 'colab'], doc="""
         Whether to render output in Jupyter with the default Jupyter
         extension or use the jupyter_bokeh ipywidget model.""")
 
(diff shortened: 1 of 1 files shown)


philippjfr:
Looks good to me. Ready to merge?

Hoxbro:
Yes

PR #4116, opened 2022-11-14
Fixes a memory leak in holoviews Pane.

### LINKED ISSUE #4096: Panel + Datashader app indicate serious memory leak
At awesome-panel.org and at work I can see the memory of my apps steadily increase to a point where I need to restart them. Its been like that for a long period of time. A liveness probe has to some extend solved the problem so I've been caring about other problems.

But I see awesome-panel.org getting slower and slower over time until I restart it. Triggered by @sophiamyang that reached out yesterday about the slowness I now want to identify the cause.

One issue I see once in a while in the logs is

```bash
bokeh.document.modules - ERROR - Module <module 'bokeh_app_35f6d52a842147efb3c893bb26bdfdb1' from '/app/examples/lib_datashader.py'> has extra unexpected referrers! This could indicate a serious memory leak. Extra referrers: [<cell at 0x7efd70a65550: module object at 0x7efd47b41900>]
```

It is caused by this app

```python
"""
The purpose of this app is to demonstrate that Panel works with the tools you know and love
&#10084;&#65039;, including Datashader. It supports both light and dark theme.
"""
import hvplot.xarray  # pylint: disable=unused-import
import panel as pn
import xarray as xr

pn.extension(template="fast")

ACCENT = "#1f77b4"

if not "air" in pn.state.cache:
    air = pn.state.cache["air"] = xr.tutorial.open_dataset("air_temperature").load().air
else:
    air = pn.state.cache["air"]


def get_plot(accent_base_color=ACCENT):
    """Returns a datashaded hvplot"""
    plot = air.hvplot.scatter(
        "time",
        groupby=[],
        rasterize=True,
        dynspread=True,
        responsive=True,
        cmap="YlOrBr",
        colorbar=True,
    ) * air.mean(["lat", "lon"]).hvplot.line("time", color=accent_base_color, responsive=True)
    plot.opts(responsive=True, active_tools=["box_zoom"])
    return plot


PLOT = get_plot()

pn.pane.HoloViews(PLOT, min_height=500, sizing_mode="stretch_both").servable()
```

Unfortunately, I cannot trigger the problem.

Hoxbro:
Fixes #4096

Fixes a memory leak by setting `self.object` to None before destroying it in the Holoviews pane. I'm not completely sure this fixes the problem at the root, but at least it stops it. 

This would give the following memory usage with `panel serve 4096.py --admin --unused-session-lifetime 100 --check-unused-sessions 500` and reloading it 10 times and closing the tab:
``` python
import hvplot.xarray  # noqa
import panel as pn
import xarray as xr


def get_plot():
    with xr.tutorial.open_dataset("air_temperature") as ds:
        air = ds.air
    plot = air.hvplot.line("time")
    return plot


pn.pane.HoloViews(get_plot()).servable()

```

| Before | After |
| --------- | ------- |
|image |image |

### DIFF
--- a/panel/pane/holoviews.py
+++ b/panel/pane/holoviews.py
@@ -31,6 +31,7 @@ from .plotly import Plotly
 if TYPE_CHECKING:
     from bokeh.document import Document
     from bokeh.model import Model
+    from bokeh.server.contexts import BokehSessionContext
     from pyviz_comms import Comm
 
 
@@ -353,6 +354,11 @@ class HoloViews(PaneBase):
             if old_pane:
                 old_pane._cleanup(root)
         super()._cleanup(root)
+
+    def _server_destroy(self, session_context: 'BokehSessionContext') -> None:
+        with param.discard_events(self):
+            self.object = None
+        return super()._server_destroy(session_context)
 
     #----------------------------------------------------------------
     # Public API


panel/pane/holoviews.py
```diff
@@ -354,6 +355,11 @@ def _cleanup(self, root: Model | None = None) -> None:
                 old_pane._cleanup(root)
         super()._cleanup(root)
 
+    def _server_destroy(self, session_context: BokehSessionContext) -> None:
+        with param.discard_events(self):
+            self.object = None
```
philippjfr, line 360: This is not safe to do. While in general Panel components are created per session it is possible to reuse them across sessions so you cannot clean up the parameters in this way.

I'm also quite dubious about this actually causing a true memory leak. Certainly it can result in a circular reference being created that means that garbage collection has to do a full collection to free up the memory. If there is a true memory leak then this fix is simply reducing its severity.

philippjfr:
Will close since this change is not safe to apply.

Issue #4080, opened 2022-11-02
Panel 0.14.0 breaks Fast Templates


mayonnaisecolouredbenz7:
Running the code exactly from documentation https://panel.holoviz.org/reference/templates/FastListTemplate.html 

Using panel 0.14.1 gives:

![image](https://user-images.githubusercontent.com/38597661/199426176-1e46ddff-f7fe-4c0a-84df-19b565f1299b.png)

MarcSkovMadsen:
Hi @mayonnaisecolouredbenz7 

Could you share a bit more information

1. How and where are you running this? (I.e. running it behind an enterprise firewall etc could cause issues)
2. What are the logs from the server?
3. Could you share the errors from the browsers console? (right click, inspect, select the *console*).

mayonnaisecolouredbenz7:
1. Running it locally, windows machine in a python virtual environment. And running installs for only panel and holoviews
![image](https://user-images.githubusercontent.com/38597661/199455742-6076bad8-d893-49e7-8788-a0a02b72e836.png)

2. Logs look normal
```
2022-11-02 20:40:12,077 Bokeh app running at: http://localhost:5006/test
2022-11-02 20:40:12,077 Starting Bokeh server with process id: 24196
2022-11-02 20:40:20,200 Subprotocol header received
2022-11-02 20:40:20,200 WebSocket connection opened
2022-11-02 20:40:20,201 Receiver created for Protocol()
2022-11-02 20:40:20,202 ProtocolHandler created for Protocol()
2022-11-02 20:40:20,202 ServerConnection created
2022-11-02 20:40:20,293 Sending pull-doc-reply from session 's1RzWcutYRiBzQenVwFj5UhsoIjjFzntCG1eSTRmphvB'
2022-11-02 20:40:27,067 [pid 24196] 1 clients connected
2022-11-02 20:40:27,067 [pid 24196]   /test has 1 sessions with 0 unused
```

3. Looks like some browser console errors/warnings of note
![image](https://user-images.githubusercontent.com/38597661/199457182-a24ee51c-1c5d-4bac-ba8c-6a13f7cad560.png)

xavArtley:
Ok it's due to a windows register key
I don't know at wich level it should be considered as an issue but the
solution I found is the following one :
```python
import mimetypes
mimetypes.add_type("application/javascript", ".js")
```

mayonnaisecolouredbenz7:
[...]

PR #4118, opened 2022-11-14
Ensure .js mimetype is set correctly

philippjfr:
Fixes https://github.com/holoviz/panel/issues/4080

### DIFF
--- a/panel/io/resources.py
+++ b/panel/io/resources.py
@@ -101,7 +101,7 @@ JS_URLS = {
 
 extension_dirs['panel'] = str(DIST_DIR)
 
-if mimetypes.types_map.get('.js') == 'application/javascript':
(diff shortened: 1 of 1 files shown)


panel/io/resources.py
```diff
@@ -97,6 +101,15 @@ def conffilter(value):
 
 extension_dirs['panel'] = str(DIST_DIR)
 
+if mimetypes.types_map.get('.js') == 'application/javascript':
```
Hoxbro, line 104: Shouldn't this be `'text/plain'` or `!=`? And maybe only be done on Windows?

philippjfr: Yep, it's wrong. Committed code I was testing with.

philippjfr, line 104:
```suggestion
if mimetypes.types_map.get('.js') == 'text/plain':
```

Issue #4097, opened 2022-11-10
Bokeh 3.0 compatibility


philippjfr:
The migration to support Bokeh 3.0 is a huge project that will take several months to complete. Initial work for compatibility has started in https://github.com/holoviz/panel/pull/3752 but it is not sustainable to keep pushing commit after commit into that PR. Therefore we have decided to create `branch-1.0` and merge the current progress into that PR and then follow up with further fixes (and features) in followup PRs.

For reference, a migration guide to bokeh 3.0 can be found [here](https://github.com/bokeh/bokeh/wiki/Migration-Guides).

# Components

This list of components highlights which items render correctly. Note that unchecked items may work and checked items may still have weird interactions when laid out with other components).

## Panes

- [x] Alert (needs some color tweaks)
- [x] Audio
- [x] Bokeh
- [x] DataFrame
- [x] DeckGL
- [x] ECharts
- [x] Folium
- [x] GIF
- [x] HTML
- [x] HoloViews
- [x] ICO
- [x] IDOM (deprecated)
- [x] IPyLeaflet (@maximlt same as IpyWidget)
- [x] IPyWidget (@maximlt https://github.com/holoviz/panel/issues/4410)
- [x] Interactive
- [x] JPG
- [x] JSON
- [x] LaTeX (`KaTeX` model does not yet work correctly - @Hoxbro)
- [x] Markdown
- [x] Matplotlib
- [x] PDF
- [x] PNG
- [x] Param
- [x] ParamFunction
- [x] ParamMethod
- [x] Perspective (scrollbars not right) (@philippjfr)
- [x] Plotly
- [x] RGGPlot (@maximlt)
- [x] SVG
- [x] Str
- [x] Streamz (@Hoxbro)
- [x] VTK
- [x] VTKJS (@philippjfr)
- [x] VTKVolume (@philippjfr)
- [x] Vega (@maximlt https://github.com/holoviz/panel/issues/4379)
- [x] Video
- [x] YT (deprecated)

## Widgets

- [x] Ace (weird height sizing - @Hoxbro)
- [x] ArrayInput
- [x] AutocompleteInput
- [x] BooleanStatus (#4129)
- [x] Button
- [x] CheckBoxGroup
- [x] CheckButtonGroup
- [x] Checkbox
- [ ] ColorPicker (https://github.com/holoviz/panel/issues/4385)
- [x] CrossSelector
- [x] DataFrame
- [x] DatePicker
[...]

PR #4129, opened 2022-11-17
Bokeh 3 - Indicators

Hoxbro:
Adding support for indicators with Bokeh 3 https://github.com/holoviz/panel/issues/4097 

- [x] BooleanStatus
- [x] LoadingSpinner
- [x] Progress
- [x] Number
- [x] String

``` python
import panel as pn
import bokeh

pn.panel(f"# Bokeh: {bokeh.__version__}").servable()

pn.indicators.BooleanStatus(value=True, color="primary").servable()
pn.indicators.LoadingSpinner(value=True, color="primary").servable()
pn.indicators.Progress(value=50, bar_color="primary").servable()
pn.indicators.Progress(active=True, bar_color="primary").servable()

pn.indicators.Number(name="Rate", value=72, format="{value}%").servable()
pn.indicators.String(value="hello").servable()

```

https://user-images.githubusercontent.com/19758978/202415989-30aa7bc3-b878-4f7d-86aa-1b4708600bb3.mp4

image


panel/dist/css/progress.css
```diff
@@ -0,0 +1,209 @@
...
+  background-position: 10%;
+  animation-name: stripes;
+  animation-duration: 3s;
+  animation-timing-function: linear;
+  animation-iteration-count: infinite;
+}
+
+progress.primary[value]::-webkit-progress-value {
+  background-color: #007bff;
+}
```
mattpap, line 70: Given the extent of changes, I would really suggest start using CSS variables to keep these constants organized, e.g:
```css
:host {
  --primary-bg-color: #007bff;
}

progress.primary[value]::-webkit-progress-value {
  background-color: var(--primary-bg-color);
}
```
This way you get a component private CSS variable. Then, when a value is shared between components, this can be generalized using layered stylesheets, like bokehjs does, though our CSS variable usage is still quite limited.

philippjfr:
This is great! Could you also update alerts.css with these variables?

Hoxbro:
> Could you also update alerts.css with these variables?

The colors are not the same as for the indicators. Do you still want me to do it?

philippjfr:
Yes let's align them.

Hoxbro:
Alerts have now been aligned. I have used https://app.contrast-finder.org to find a good contrast text color.

Before:

before

After:
after

philippjfr:
>Alerts have now been aligned. I have used https://app.contrast-finder.org to find a good contrast text color.

Sweet thanks, would be nice if we could somehow reduce the opacity a bit. I think that's only possible with `rgba` color definitions.

PR #4130, opened 2022-11-17
Add Switch widget and description parameter to widgets

philippjfr:
- Adds `Switch` widget
- Adds `description` parameter to all widgets (that support it)
- Maps parameter doc attribute to the description


panel/widgets/input.py
```diff
@@ -112,7 +115,7 @@ class TextAreaInput(TextInput):
     Lines are joined with the newline character `\n`.
 
     Reference: https://panel.holoviz.org/reference/widgets/TextAreaInput.html
-
+L
```
mattpap, line 118:
```suggestion
```

PR #4140, opened 2022-11-21
Bokeh 3 - Fix tests

Hoxbro:
Fixes the tests
- Update background to `styles["background"]`
- Tabulator tests.
- Dial not having a _widget_type.

Also:
- Remove a version guard for bokeh. 
---
### Background 
A lot of the failing test on branch-1.0 is because Bokeh 3 no longer uses `model.background` but instead uses `model.styles["background"]`.

This PR fixes these failing tests by converting  `model.background`  -> `model.styles["background"]` and then gives a warning. 

The changes made in this PR do not work in Firefox and will be fixed in https://github.com/bokeh/bokeh/issues/12633.

### DIFF
--- a/panel/viewable.py
+++ b/panel/viewable.py
@@ -251,23 +251,6 @@ class Layoutable(param.Parameterized):
             elif config.sizing_mode == 'stretch_height' and 'height' not in params:
                 params['sizing_mode'] = config.sizing_mode
         super().__init__(**params)
-        self._set_background()
-        self.param.watch(self._set_background, 'background')
-
-    def _set_background(self, *event):
-        if self.background == self.styles.get("background", None):
-            return
-
-        # Warning
-        prev = f'{type(self).name}(background="{self.background}")'
-        new = f'{type(self).name}(styles={{"background": "{self.background}"}}'
-        deprecation_warning(
-            f"{prev!r} is deprecated and will stop working, "
-            f"use {new!r} instead of."
-        )
-
-        self.styles["background"] = self.background
-        self.param.trigger("styles")
 
 
 _Self = TypeVar('_Self', bound='ServableMixin')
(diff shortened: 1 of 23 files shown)


panel/viewable.py
```diff
@@ -251,6 +251,23 @@ def __init__(self, **params):
             elif config.sizing_mode == 'stretch_height' and 'height' not in params:
                 params['sizing_mode'] = config.sizing_mode
         super().__init__(**params)
+        self._set_background()
+        self.param.watch(self._set_background, 'background')
```
Hoxbro, line 255: @philippjfr this line makes `test_param_function_pane_update` fails. Even if I replace it with something like `self.param.watch(lambda x: print(x), 'background')`. I can recreate it on the current master.  Do you know why it this is the case?

``` python
document = <bokeh.document.document.Document object at 0x7f6569a7a4d0>, comm = Comm(id='c5831af2ef354ca39bcc1b228eabc6d5', name='Comm01251')

    def test_param_function_pane_update(document, comm):
        test = View()
    
        objs = {
            0: HTML("012"),
            1: HTML("123")
        }
    
        @param.depends(test.param.a)
        def view(a):
            return objs[a]
    
        pane = panel(view)
        inner_pane = pane._pane
>       assert inner_pane is not objs[0]
E       assert HTML(str) is not HTML(str)

panel/tests/test_param.py:1109: AssertionError

```

Hoxbro: Same thing with these tests:
```
FAILED panel/tests/io/test_embed.py::test_embed_param_jslink - KeyError: 'target'
FAILED panel/tests/io/test_embed.py::test_embed_select_str_jslink - AssertionError: assert 2 == 1
FAILED panel/tests/io/test_embed.py::test_embed_checkbox_str_jslink - AssertionError: assert 2 == 1
FAILED panel/tests/io/test_embed.py::test_embed_slider_str_jslink - AssertionError: assert 2 == 1
```

Hoxbro: Found a workaround in dec2eb7a29334e11416f70d8db09f6818d616f23

PR #4148, opened 2022-11-23
Sync branch-1.0 with master

Hoxbro:
Not sure if this is the best way to keep the branch-1.0 in sync with master. I have used a `git merge`, but it could also be done with `git rebase` directly on the branch.


philippjfr:
We do need to work out what to do about this. We can't take the usual approach of squash merging once we decide that branch-1.0 is ready to be merged into master/main because that will lose too much information. At the same time branch-1.0 is also too messy to merge into master/main without some cleanup. The problem I see with doing a `git merge` from master into branch-1.0 is that we'll end up with an even bigger tangle of commits. Open to suggestions @maximlt and @Hoxbro .

maximlt:
> We can't take the usual approach of squash merging once we decide that branch-1.0 is ready to be merged into master/main because that will lose too much information

Yes that makes sense.

> At the same time branch-1.0 is also too messy to merge into master/main without some cleanup.

Really? It doesn't look too bad.

image

philippjfr:
Good point, I had incorrectly assumed that the initial Bokeh compatibility branch was merged into branch-1.0 without squashing.

philippjfr:
Will close and rebase branch-1.0 manually.

Issue #2731, opened 2021-09-12
Support step and format on DateSlider


MarcSkovMadsen:
The current version of Panel is 0.12.1.

If I look at the Bokeh [DateSlider](https://docs.bokeh.org/en/latest/docs/reference/models/widgets.sliders.html#bokeh.models.widgets.sliders.DateSlider) I can see it supports a `step` and `format` which the Panel `DateSlider` does not.

I think supporting that would be quite useful for Panel as that would allow sliding through other timeperiods than days down to milliseconds as far as I can see.

## Todo

It's probably as simple as

- Add `step` and `format` parameters to Panels `DataSlider`
- Add tests or test manually that it works
- Update reference guide.

## Additional Context

It would for example increase the possibilities for *Creating a dynamic map with datetime range*. See https://github.com/holoviz/panel/issues/673

Axeldnahcram:
Hi, long time lurker but I guess I kinda want to get involved after using panel for more than a year now. How can I push my changes to a branch to make my first PR @MarcSkovMadsen ?

Axeldnahcram:
I guess I need to fork my own version of panel. I will do it tonight

maximlt:
Hi @Axeldnahcram,

Yes indeed as an external contributor you must fork this repository, then create a new branch in your local fork and push it to your remote repo. At this point Github will offer you to merge your branch into this repository, that's a Pull Request.

You can find the developer instructions here for more details: https://panel.holoviz.org/developer_guide/index.html

PR #4152, opened 2022-11-24
Now support step format in date sliders

Axeldnahcram:
Original issues are [2731](https://github.com/holoviz/panel/issues/2731) and [2894](https://github.com/holoviz/panel/issues/2894).

I have not really added tests since the formatting is done in the JS of the bokeh object and I feel like testing must be done on their side rather than ours then. And I kinda do not really like simply testing that if I call `slider=Slider(format=5)` then `slider.format=5` but if needed I can add it.

If needed I can also add some BFSoup checker somewhere to test both steps and format, of course if you have an original example of pr that had to deal with that I will be happy to comply to any specific test format.

On a side note, it looks like the `.controls()` function is oftentimes not displaying the full controls. I would be happy to investigate.

Also if you have a tool that strips the notebooks of the cell metadata updates so that my commits on notebooks are clean I would be very happy.

It is my first ever PR on this project, let me know if there are any specifics to consider when contributing. 

Have a good day

### DIFF
--- a/examples/reference/widgets/DateRangeSlider.ipynb
+++ b/examples/reference/widgets/DateRangeSlider.ipynb
@@ -43,6 +43,7 @@
     "* **``name``** (str): The title of the widget\n",
     "* **``orientation``** (str): Whether the slider should be displayed in a 'horizontal' or 'vertical' orientation.\n",
     "* **``tooltips``** (boolean): Whether to display tooltips on the slider handle\n",
+    "* **``format``** (str): The datetime format\n",
     "\n",
     "___\n",
     "\n",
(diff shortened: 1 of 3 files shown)


philippjfr (approved):
This looks good, thanks! In general unless the model is implemented in Panel itself (i.e. not just wrapping something in Bokeh) we add UI tests but since that's not the case here this is fine. You could add a test that checks that the bokeh model returned by `obj.get_root()` correctly sets the `format` and is updated when the `format` parameter is changed but it's not critical. I'll merge and look at why the tests are being flakey.

Issue #3996, opened 2022-10-13
npm missnamed in server.py


t-houssian:
Python 3.10.6
Panel 0.14

It appears that npm was misnamed in the DocHandler causing authentication to not work with pn.serve().

See line 330 in panel/io/server.py

![Screenshot from 2022-10-13 14-13-05](https://user-images.githubusercontent.com/25035266/195711773-3a11b1bb-7968-4b73-9e12-a29ab15f5936.png)
![Screenshot from 2022-10-13 16-09-13](https://user-images.githubusercontent.com/25035266/195711786-9783809b-6e24-42b3-83c8-bfc120fcd944.png)

```
npm_cdn=config.npn_cdn,
```

Needs to be

```
npm_cdn=config.npm_cdn,
```

A few steps are needed if you want to reproduce the error. But you should easily be able to see the typo.

Here is a reproducible example. You will have to use your own google web api credentials and add the correct redirect uri and your own encryption and cookie keys.

```python
import panel as pn

def authorize(user_info):
        with open('users.txt') as f:
            valid_users = f.readlines()
        return user_info['email'] in valid_users

    pn.config.authorize_callback = authorize

def abc():
    return pn.pane.JSON({'abc': 123})

def bcd():
    return pn.pane.JSON({'bcd': 123})

template = pn.template.FastListTemplate(
    main=abc()
).servable()

template2 = pn.template.FastListTemplate(
    main=bcd()
).servable()

APP_ROUTES = {
    "/": template,
    "/bcd": template2
}

pn.serve(
    APP_ROUTES,
    port=5010,
    allow_websocket_origin=["127.0.0.1:5010"],
    address="127.0.0.1",
    show=True,
    autoreload=True,
    oauth_provider="google",
    oauth_key="KEY",
    oauth_secret="SECRET",
    oauth_encryption_key="k",
    cookie_secret="c",
    oauth_redirect_uri="http://127.0.0.1:5010/bcd"
)
```

I can open a merge request for this as well if needed.

jbednar:
Yes please! 🙂

PR #4162, opened 2022-11-29
auth.py npm reference fix

minasouliman:
small fix on npm reference in auth. same as https://github.com/holoviz/panel/issues/3996


philippjfr:
Thanks @minasouliman!

PR #4169, opened 2022-11-30
Better estimation of value for editable sliders with fixed_start / fixed_end

Hoxbro:
Before this PR, the value will always be set to `0` or `(0,1)` even if the value was outside the hard bounds of `fixed_start` / `fixed_end`.


philippjfr (approved):
Looks like a pain, but also looks correct.

PR #4174, opened 2022-12-05
Don't raise TypeError for a class containing  `__panel__` and not Viewer

Hoxbro:
Don't raise TypeError for a class containing  `__panel__` and not `pn.viewable.Viewer`  by checking if the class is an instance of type before checking if it is a subclass. 

Before:
image


philippjfr (approved):
Looks good!

PR #4179, opened 2022-12-07
Fix branch-1.0 tests after rebase

Hoxbro:
Mostly small fixes. 

Most of the fixes are related to changes made in #4130, where `panel.widgets.Checkbox` was switched from using `bokeh.models.CheckboxGroup` to `bokeh.models.Checkbox` as the `_widget_type`.


philippjfr:
Looks great, thanks!

PR #4193, opened 2022-12-14
Add Reacton pane

philippjfr:
Add `Reacton` pane to make it easy to render Reacton components.

Screen Shot 2022-12-14 at 20 15 18


philippjfr:
Ping @maartenbreddels just a heads up that I'm adding this.

maartenbreddels:
Cool, please give me some time to review this, I think I have time tomorrow

MarcSkovMadsen:
Some of the question answers that would be nice to have in the docs are

- can you use panel components in reacton?
- are there any performance hits when using reacton with python?
- is it possible to use reacton to  get some of the ipywidgets that do not work with Panel today to work?

philippjfr:
Found a good number of issues with IPywidgets support in Panel but now that those are resolved this works both in the notebook and in an app:

Screen Shot 2022-12-15 at 11 43 34

philippjfr:
Will try to expand the documentation. For now here are my answers:

>can you use panel components in reacton?

In theory yes if you wrap your Panel component in an ipywidget with `pn.ipywidget`, however in practice I would probably not recommend it and there's bound to be some weirdness here.

>are there any performance hits when using reacton with python? I.e. is this useful or just a show off?

Are you asking whether Reacton in general is useful? That's a question for @maartenbreddels but I presume so because I don't think he'd have built it otherwise.

>is it possible to use reacton to get some of the ipywidgets that do not work with Panel today to work?

No, Panel supports all widgets that have been correctly implemented, i.e. publish the appropriate bundles to CDN. This doesn't change anything about that.

maartenbreddels:
There are performance penalties for using Reacton, similar in how React is slower than HTML DOM manipulation. However, we do our best to keep that minimal. So I think Reacton should be seen as making it easier to build complex UIs. It shouldn't have any impact on visualizations for example, those are standalone, and don't have much todo with Reacton.
If there are performance issues, i'd like to hear that so we can improve it.

Reacton is pure Python, so it only uses the widgets as they are.

Long term, Panel might be able to benefit from Solara, since that basically replaces https://github.com/maartenbreddels/flask-ipywidgets which I think panel ipywidgets is based on (I'll contact you on this in the future).

philippjfr:
Note our developer instructions currently suck so if you want to try I'd suggest doing this:

1. Clone Panel
2. Create an env

```
conda create -c conda-forge -n panel-dev python=3.10 jupyterlab nodejs bokeh=2.4.3 ipywidgets
cond activate panel-dev
pip install reacton ipywidgets_bokeh
```

3. Develop install Panel

```
pip install -e .
```

4. Run the notebook in `examples/reference/pane/Reacton.ipynb`
5. Add `.servable()` to one of the Reacton examples and then `panel serve examples/reference/pane/Reacton.ipynb` OR just use the Panel preview button in JLab.

maartenbreddels:
I have it running (need to manually do `panel build panel`). How can I trigger the call to `_cleanup`?

philippjfr:
It should run when a cell is deleted and/or when the component is removed from the app, e.g. when you do:

```python
layout = pn.Row(reacton_component)
...
layout[:] = []
```

maartenbreddels:
```python
@reacton.component
def ButtonClick():
    # first render, this return 0, after that, the last argument
    # of set_clicks
    clicks, set_clicks = reacton.use_state(0)

    def test_effect():
        print("run effect")
        def cleanup():
            print("cleanup effect")
        return cleanup
    reacton.use_effect(test_effect, [])
    
    def my_click_handler():
        # trigger a new render with a new value for clicks
        set_clicks(clicks+1)

    button = w.Button(description=f"Clicked {clicks} times",
                      on_click=my_click_handler)
    return button

r = pn.panel(ButtonClick())
layout = pn.Row(r)
layout
```

If correct, it should print 'run effect', and when cleanup happens (`layout[:] = []`), it should print 'cleanup effect'. The current branch doesn't do that, but with this diff it does:
```diff
diff --git a/panel/pane/ipywidget.py b/panel/pane/ipywidget.py
index 5c94d7c9..04c634b1 100644
--- a/panel/pane/ipywidget.py
+++ b/panel/pane/ipywidget.py
@@ -94,14 +94,13 @@ class Reacton(IPyWidget):
 
     def _cleanup(self, root: Model | None = None) -> None:
         if root and root.ref['id'] in self._widget_wrappers:
-            box = self._widget_wrappers.pop(root.ref['id'])
-            box.layout.close()
-            box.close()
+            self.rc.close()
         super()._cleanup(root)
 
     def _get_ipywidget(
         self, obj, doc: Document, root: Model, comm: Optional[Comm], **kwargs
     ):
-        from reacton.core import make
-        self._widget_wrappers[root.ref['id']] = widget = make(obj)
+        import reacton
+        widget, self.rc = reacton.render(obj)
+        self._widget_wrappers[root.ref['id']] = widget
         return super()._get_ipywidget(widget, doc, root, comm, **kwargs)
```

philippjfr:
Will apply and write a test to that effect.

philippjfr:
Have to change it slightly, presumably each widget should get its own context?

philippjfr:
Okay, added a UI test and fixed the cleanup logic.

philippjfr:
Note using Panel components in Reacton is possible but requires some fiddling:

Screen Shot 2022-12-15 at 15 46 02

MarcSkovMadsen:
Would it be possible to add some code that handled the wrapping as ipywidget and into layout for the user? Something that enables the user to use Panel components directly? Then Reaction would be an api for both Ipywidgets and Panel widgets, which would be quite powerful.

philippjfr:
You mean for Reacton to handle Panel components? That would be quite nice from my perspective. On the Panel side though I'm not entirely convinced that a helper function that does the equivalent of `w.VBox(children=[pn.ipywidget(panel_component)])` is something we should add.

maartenbreddels:
This works, because Reacton doesn't recognise the argument, and just passes it along. But this will create a new widget on each render (so each button click). The idea of Reacton, is that it doesn't do that, but returns elements (which is a description of what you want).
So `pn.ipywidget` would need to be replaced by a something that returns an element. 
I'm also wondering what happens if you create a pn.widget.Button? is that expensive? If so, it would require generating wrappers for panel as well.
There are other ways this can be done, but they will be less elegant.

philippjfr:
>I'm also wondering what happens if you create a pn.widget.Button? is that expensive? If so, it would require generating wrappers for panel as well.

It's not super expensive no, wrappers would be nice yes. Basically they'd probably do the same thing our reactive functions (like `pn.bind`) do, which is to update the existing "model" with the state from the new one.

philippjfr:
Haven't read any of the internals of Reacton but I'm guessing internally the Elements have a rendering context and then inspect the stack and return the same widget each time? Sounds like it wouldn't be too hard to write a wrapper for that.

maartenbreddels:
> but I'm guessing internally the Elements have a rendering context

Not exactly, because every time to function body gets executed, new elements get created. New elements get matches to the previous run based on the 'location' they are within the widget tree. If the type matches, but the arguments change, we update the widget values. If there is no match, we either create or destroy the widget.

maartenbreddels:
🎉

Issue #4183, opened 2022-12-08
Dynamically scaling Panel


govinda18:
#### Is your feature request related to a problem? Please describe.
Currently two users simultaneously using a panel app would block each other. One workaround to this is to use num-threads but this however is not the most elegant solution as it requires the developer to be cautious with handling GIL. Logging this request to explore solutions on dynamically scaling a panel server as more users join.

#### Describe the solution you'd like
One idea is to have several workers at the server's disposal and events can be redirected to the workers. This requires a mechanism to appropriately store state such that the tasks assigned to workers can be uniquely identified with the clients that they are associated to.

#### Describe alternatives you've considered
One of the solutions using static scaling is to make multiple panel servers sit behind an nginx load balancer. This is what we are currently using.

MarcSkovMadsen:
Have you tried `--num-procs` @govinda18? This will start the number of workers you provide (or an optimal number of works if you provide `0`).

govinda18:
> Have you tried --num-procs @govinda18? This will start the number of workers you provide

It does not work as one would want it to actually due to https://stackoverflow.com/a/63849068/7398517. Therefore a better load-balancing solution any way is to use https://docs.bokeh.org/en/latest/docs/user_guide/server/deploy.html#load-balancing instead.

However, these options limit you in terms of optimal resource allocation. If you use too many, you may be consuming a lot of resources, mostly idle and too few, which would not help balance the load. Ideally, I would want to scale things as and when the load increases dynamically. Maybe use some standard request-response designs or something of that sort.

TheoMathurin:
I would also very much like a way to scale panel apps dynamically for the same reasons mentioned by @govinda18. It would dramatically improve both the developper and user experience for production apps which are likely to be simultaneously accessed by several clients.

Threading using `--num-threads` helps, but not that much for callbacks with CPU-bound tasks which make your process busy for some time. Also, the fact that bokeh and panel apps are stateful means a client is connected to the same python process throughout. This has advantages but it makes scaling less straightforward compared to a stateless server. It may be a lot of work or even not really feasible, but if the said process could indeed hand tasks to a pool of independent workers so that it's still able to handle further requests, that would be a game-changer IMHO.

By the way the panel docs [section](https://panel.holoviz.org/user_guide/Performance_and_Debugging.html#multiple-processes) mentioning `--num-procs`  is quite misleading as it makes it look like you get something equivalent to a standard load balancer, while it's clear that requests are not balanced.

[...]

PR #4196, opened 2022-12-15
Update "Concurrent processing" section

TheoMathurin:
Mention in the docs that the Tornado multi-process mode does not do any load balancing.

Discussed here: https://github.com/holoviz/panel/issues/4183


philippjfr (approved):
Thanks, clarifications look good!

PR #4198, opened 2022-12-15
Migrate panel.js to CSS modules

mattpap:
This is an early preview that sets up things and makes `variables.css` to be included via a CSS module.


panel/styles.d.ts
```diff
@@ -0,0 +1,36 @@
...
+}
+declare module "styles/progress.css" {
+  export default ""
+}
+declare module "styles/variables.css" {
+  export default ""
+}
+declare module "styles/widgets.css" {
+  export default ""
+}
```
mattpap, line 36: bokehjs can auto-generate this, but the extensions' compiler was not updated to support this, so I will have to do that in 3.1. However, for now this will work and when the compiler is updated, then this file can be removed

mattpap: Not needed anymore with bokeh 3.1.dev1.

philippjfr:
Thanks for this! At this point this would be a really painful merge/rebase. I'll probably re-apply this in some form in a new PR.

PR #4211, opened 2022-12-17
Refactor util and add util.warnings

Hoxbro:
At first, this PR was meant only to give a visible warning to users when running `pn.Pane`.

I wanted to add a `find_stack_level` and the logical place to have it was in `util.py`, but I felt like it would pollute the file. So I moved it to `util/__init__.py`  and split the checks into `util/checks.py`, and then imported them back into the `__ini__.py`. This should make it backward-compatible. And because I already was in these files, I made some small refactoring mostly related to type hints. 

Finally, I created a `util/warnings.py` and created a `deprecated` function, which needs a version when the function is removed from Panel. I have added an error for when this version is exceeded by Panel itself.  

Some questions:
- I would like to change all the check functions to start with `is_XXX`. But for now, I haven't done it. 
- Should I add a `deprecated` to other things, here I'm mainly thinking about `pn.pane.VTK`.

### DIFF
--- a/panel/tests/test_param.py
+++ b/panel/tests/test_param.py
@@ -13,7 +13,7 @@ from bokeh.models import (
 from panel.io.state import set_curdoc, state
 from panel.layout import Row, Tabs
 from panel.pane import (
-    HTML, Bokeh, Matplotlib, Pane, PaneBase, Str, panel,
+    HTML, Bokeh, Matplotlib, PaneBase, Str, panel,
 )
 from panel.param import (
     JSONInit, Param, ParamFunction, ParamMethod,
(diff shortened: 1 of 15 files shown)


panel/tests/test_param.py
```diff
@@ -25,6 +25,10 @@
 )
 
 
+def Pane(obj, **kwargs):
+    return PaneBase.get_pane_type(obj, **kwargs)(obj, **kwargs)
```
philippjfr, line 29: What's up with this? Not easy to replace with `pn.panel`?

Hoxbro: It is for functions that expect a `pn.Param` output.  `pn.panel` give a `Column` output. 

I could also change the type to assert with to `Column` if this is more desirable for you. 

``` python
def test_instantiate_from_class():

        a = param.Number()

    assert isinstance(Pane(Test), Param)
```

philippjfr: Gotcha

philippjfr (approved):
Thanks @Hoxbro! Looks good to me.

PR #4212, opened 2022-12-17
Update unittest

Hoxbro:
Two things 
1) Rename `test_` which have a return in it to, `manualtest_` to avoid the `PytestReturnNotNoneWarning`.
2) Trying to make the unit test run in parallel, and small updates to make the test_suite run.

### DIFF
--- a/panel/tests/pane/test_alert.py
+++ b/panel/tests/pane/test_alert.py
@@ -36,7 +36,7 @@ def test_existing_css_classes():
     assert set(alert.css_classes) == {"alert", "alert-info", "important"}
 
 
-def test_all_view():
+def manualtest_all_view():
     """Test that we can construct and view all Alerts"""
     alerts = []
     for alert_type in ALERT_TYPES:
@@ -63,4 +63,4 @@ def test_all_view():
 
 if __name__.startswith("bokeh"):
     pn.extension(sizing_mode="stretch_width")
-    test_all_view().servable()
+    manualtest_all_view().servable()
--- a/panel/tests/pane/test_vega.py
+++ b/panel/tests/pane/test_vega.py
@@ -230,7 +230,7 @@ def test_altair_pane(document, comm):
     chart.mark = 'point'
     chart.data.values[0]['x'] = 'C'
     pane.object = chart
-    point_example = dict(vega_example, mark='point')
+    point_example = dict(vega_example, data={},  mark='point')
     if altair_version >= Version('4.0.0'):
         point_example['config'] = vega4_config
     assert dict(model.data, **blank_schema) == dict(point_example, **blank_schema)
--- a/panel/tests/pane/test_vega.py
+++ b/panel/tests/pane/test_vega.py
@@ -155,7 +155,7 @@ def test_vega_pane(document, comm):
     point_example = dict(vega_example, mark='point')
     point_example['data']['values'][0]['x'] = 'C'
     pane.object = point_example
-    point_example['data'].pop('values')
+    point_example = dict(point_example, data={})
     assert model.data == point_example
     cds_data = model.data_sources['data'].data
     assert np.array_equal(cds_data['x'], np.array(['C', 'B', 'C', 'D', 'E']))
--- a/panel/tests/widgets/test_speech_to_text.py
+++ b/panel/tests/widgets/test_speech_to_text.py
@@ -277,5 +277,4 @@ def manualtest_get_color_app():
 
 if __name__.startswith("bokeh"):
     pn.extension(sizing_mode="stretch_width")
-    # manualtest_get_advanced_app().servable()
     manualtest_get_color_app().servable()
--- a/panel/tests/widgets/test_speech_to_text.py
+++ b/panel/tests/widgets/test_speech_to_text.py
@@ -120,7 +120,7 @@ def test_can_create_result_from_list():
     assert actual[0].alternatives[0].transcript == "and why"
 
 
-def test_get_advanced_app():
+def manualtest_get_advanced_app():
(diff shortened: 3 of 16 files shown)


Hoxbro:
Even if the speed-up does not look like much on the CI, I can run almost all the unit tests in under 15 seconds, whereas without, it takes 90 seconds.

panel/tests/pane/test_vega.py
```diff
@@ -230,7 +230,7 @@ def test_altair_pane(document, comm):
     chart.mark = 'point'
     chart.data.values[0]['x'] = 'C'
     pane.object = chart
-    point_example = dict(vega_example, mark='point')
+    point_example = dict(vega_example, data={},  mark='point')
```
Hoxbro, line 233: This change is needed to be able to run this test in parallel, as it seems to be depends on the previous tests in this file have been run before it. I don't know why this is the case. 

E.g. this fails `pytest panel/tests -k "altair"` but this does not `pytest panel/tests -k "altair or vega"` before this change.

``` python-traceback
============================================== FAILURES =====================================================
_________________________________________________ test_altair_pane _________________________________________________

document = <bokeh.document.document.Document object at 0x7f1318570b80>
comm = Comm(id='bb34c97071754b02b7db8bc3f4a895d3', name='Comm01263')

    @altair_available
    def test_altair_pane(document, comm):
        pane = pn.panel(altair_example())
    
        # Create pane
        model = pane.get_root(document, comm=comm)
        assert isinstance(model, VegaPlot)
    
        expected = dict(vega_example, data={})
        if altair_version >= Version('4.0.0'):
            expected['config'] = vega4_config
        assert dict(model.data, **blank_schema) == dict(expected, **blank_schema)
    
        cds_data = model.data_sources['data'].data
        assert np.array_equal(cds_data['x'], np.array(['A', 'B', 'C', 'D', 'E']))
        assert np.array_equal(cds_data['y'], np.array([5, 3, 6, 7, 2]))
    
        chart = altair_example()
        chart.mark = 'point'
        chart.data.values[0]['x'] = 'C'
        pane.object = chart
        point_example = dict(vega_example, mark='point')
        if altair_version >= Version('4.0.0'):
            point_example['config'] = vega4_config
>       assert dict(model.data, **blank_schema) == dict(point_example, **blank_schema)
E       AssertionError: assert {'$schema': '...ative'}}, ...} == {'$schema': '...ative'}}, ...}
E         Omitting 4 identical items, use -vv to show
E         Differing items:
E         {'data': {}} != {'data': {'values': [{'x': 'A', 'y': 5}, {'x': 'B', 'y': 3}, {'x': 'C', 'y': 6}, {'x': 'D', 'y': 7}, {'x': 'E', 'y': 2}]}}
E         Full diff:
E           {
E            '$schema': '',
E            'config': {'view': {'continuousHeight': 300,...
E         
E         ...Full output truncated (19 lines hidden), use '-vv' to show

panel/tests/pane/test_vega.py:236: AssertionError
============================================= short test summary info ==============================================
FAILED panel/tests/pane/test_vega.py::test_altair_pane - AssertionError: assert {'$schema': '...ative'}}, ...} == {'$schema': '...ative'}}, ...}

Comment (philippjfr) [reply panel/tests/pane/test_vega.py:233]: Hmm, is the Vega pane modifying the data inplace somehow. Based on a quick scan I couldn't find anything that could cause this.

Comment (Hoxbro) [reply panel/tests/pane/test_vega.py:233]: I think the `data` from `point_example` should be empty. It does not make any sense that it has `data.values[0]['x'] = 'A'` when we a couple of lines above changed it to be `C`.  

The problem is because of a `pop` in `test_vega_pane` will push a fix.

File: panel/tests/widgets/test_speech_to_text.py [hunk 2]
```diff
@@ -277,5 +277,5 @@ def update_result_panel(results_last):
 
 if __name__.startswith("bokeh"):
     pn.extension(sizing_mode="stretch_width")
-    # test_get_advanced_app().servable()
-    test_get_color_app().servable()
+    # manualtest_get_advanced_app().servable()
```
philippjfr, line 280:
```suggestion
```

philippjfr:
Thanks, happy to merge this now. That said I'm not sure if I'm even happy to keep these manual tests going forward. It's something @MarcSkovMadsen added but since no one actually runs these they will just rot and eventually break. Where possible manual tests should probably be converted to UI tests.

panel/tests/pane/test_alert.py
```diff
@@ -36,7 +36,7 @@ def test_existing_css_classes():
     assert set(alert.css_classes) == {"alert", "alert-info", "important"}
 
 
-def test_all_view():
+def manualtest_all_view():
```
maximlt, line 39: Does it mean these tests will actually no longer run on the CI?

Hoxbro: Correct. The `manualtest` have no assert in them. It could, of course, find errors in the code, but mostly they are creating a servable dashboard. 

If any test with `return` in it also has an `assert`, I have made two tests one with the assert named `test_name` and one with return named `manualtest_name`.

maximlt: Oh ok I was unlucky then as this one has a couple of `assert`s. I didn't know there were kind of manuel tests in the test suite. Did you know? Just thinking that code that is not executed is code than can rot fast, if it's of no use maybe we could just get rid of it.

Hoxbro: I think we should either remove them or move them to a separate folder for now. So it is clear that these tests should be checked manually.

philippjfr (approved):
Happy to defer decisions about `manualtest_` until later. Thanks for this!

PR #4216, opened 2022-12-17
Modernize home page

### LINKED ISSUE #4214: The home page is outdated
If I am looking at the home page https://panel.holoviz.org/ I see something that does not represent what data apps and Panel are in 2022.

## Title

"A high-level app and dashboarding solution for Python". This is not *catchy* enough. There are many of these frameworks today. What sets Panel apart? 

I would change it to "The flexible data app and dashboarding solution for the data science pros. Works with the tools you know and love". I believe this is our edge. I would also like our framework to be great for the beginner - but I don't believe that is our edge.

## Examples

image

In general we should 

- make sure the examples are styled well. the Fast templates are the best, so why not use them?
- make sure the examples are responsive (stretching to the available width)
- make sure the examples are responsive (i.e. loads fast and responds fast when the user interacts with them. If they take time to load or run functions this should be clearly indicated with loading spinners etc.)
- Make sure our users can find the code of the application
- Maybe add a bit of text to explain why this specific example is important for showcasing the power of Panel.

We should 

- replace the Attractors example. It styling is very ugly. Its a slow application. And its not representative of the applications our users are looking to create. If we can style it better and make it indicate when its working we can keep it. But it needs some text to describe why this is an important example for Panel to show as the first one to our potential users.
- Change the rest of the examples to use Fast templates and stretch appropriately.
- Make sure the apps work. https://penguin-crossfilter.pyviz.demo.anaconda.com/penguin_crossfilter does not.

image

- Make sure the apps reference HoloViz and not PyViz

image

- Make sure the gapminder apps shows 5 plots, has a play button and is styled well.

image

## Video

We should add a video showcasing how Panel works and what can be created. For inspiration see https://s3-us-west-2.amazonaws.com/assets.streamlit.io/videos/hero-video.mp4.

## Embedded Examples

We should

- Remove the Penguin K-Means clustering example. Its simply ugly and not representative of an "end product". Its for exploration.
- Update the Slideshow. We should provide the full code to the user. A new user needs working code. They cannot "guess" the rest. The pictures should be nice looking pictures of Panel guides, applications, blog posts, tweets etc. When the user clicks the image they should be sent to the resource. We should also add custom styling to the Tab headers to make them look modern and nice.

## Comparison

Move it below the description of Panel and update it to 2022.

## Panel highlights

Move it above the comparison and update it to 2022.

## Usage

Make it very, very clear to new users what they should click to *get started*.

We should also add links to Panelite.holoviz.org and Binder if we believe they are in a working state.

## Alternatively

As an alternative to the above we could also change the concept of the web page with inspiration from **hvplot**. I think the hvplot web page very clearly and visually communicates what hvplot is and can do. And the user can even find working code to get started.

We could also get some inspiration from *Param*

image

For example we could write something like *Do you want to be a data science pro? Then you need Panel to make your data and models interactive. Check out our 5 minute intro video to see why!*.

MarcSkovMadsen:
An attempt to modernize the home page

Closes #4214

Dependencies

- [] https://github.com/pyviz-topics/examples/pull/212

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -111,6 +111,9 @@ ENV/
 doc/**/*.rst
 doc/**/*.ipynb
 doc/**/*.json
+doc/_build/*
+doc/gallery/*
+doc/reference/*
 # this dir contains the whole website and should not be checked in on main
 builtdocs/
 .doit.db
--- a/doc/_static/css/custom.css
+++ b/doc/_static/css/custom.css
@@ -1,6 +1,6 @@
 :root[data-theme="light"] {
     --pst-color-primary: rgb(47, 47, 47);
-    --pst-color-link: rgb(0, 170, 65);
+    --pst-color-link: #008000;
 }
 
 .nav-link {
--- a/doc/index.md
+++ b/doc/index.md
@@ -18,7 +18,7 @@ sd_hide_title: true
 :link: https://gapminders.pyviz.demo.anaconda.com
 :link-type: url
 
-```{image} https://assets.holoviews.org/panel/thumbnails/index/gapminders.png
+```{image} https://user-images.githubusercontent.com/42288570/208299928-ef3253eb-000d-4c0c-94df-55c8ff5ac224.jpg
 ```
 
 :::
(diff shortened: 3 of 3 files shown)


MarcSkovMadsen:
I abandon this one. The Panel Home page has been updated a lot since I made this one to the better. And I also have a more up to date PR here https://github.com/holoviz/panel/pull/4654.

PR #4226, opened 2022-12-20
Add Vizzu pane

philippjfr:
Still experimenting with API and deciding whether to publish as panel-vizzu extension package but wanted to share my experiments.

vizzu_panel


MarcSkovMadsen:
I had just the same idea to start working on this 😉

PR #4228, opened 2022-12-21
fix: require wheel for setup

Midnighter:
Small fix that gets rid of some of the warnings when installing from this branch.


philippjfr:
Failures are unrelated. Will merge.

PR #4252, opened 2023-01-05
Simplify determining whether script is executed as application

philippjfr:
It is sometimes useful to write a Panel application in a script and also do some things that have side-effects such as setting `pn.config` options. If you also want this module to be importable without any side-effects you currently have to write something like this:

```python
if __name__.startswith('bokeh_app_'):
    ...
```

This is ugly and weird since it looks like magic and it's a Panel app not a bokeh app. Therefore I think we should provide a helper that makes it easy to determine if we are in a `panel serve` execution context. Therefore I propose introducing `pn.state.served`:

```python
if pn.state.served:
    ...
```

Ping @MarcSkovMadsen since you are the main person I've seen use this pattern. I don't love the `in_app` naming so am very eager to hear suggestions.

### DIFF
--- a/doc/api/state.md
+++ b/doc/api/state.md
@@ -36,6 +36,9 @@ The `pn.state` object makes various global state available and provides methods 
 `refresh_token`
 : The refresh token issued by the OAuth provider to authorize requests to its APIs (if available these are usually longer lived than the `access_token`).
 
+`served`
+:Whether we are currently inside a script or notebook that is being served using `panel serve`.
+
 `session_args`
 : When running a server session this return the request arguments.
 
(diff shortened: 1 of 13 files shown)


Hoxbro:
I also use it a lot and would prefer this method over it. :+1: 

For name suggestions:  `app_running`, `running_app`, `serving_app`

Hoxbro (approved):

MarcSkovMadsen:
What about .served or .is_served?

philippjfr:
Thanks @MarcSkovMadsen, decided `pn.state.served` was short and clear.

maximlt (approved):
> What about .served, .is_served or .in_served_context?

I had in mind exactly these alternatives :) Happy with `.served` !

PR #4254, opened 2023-01-06
Deprecate IDOM

philippjfr:
I've decided to deprecate the [IDOM](https://github.com/idom-team/idom) pane. IDOM is a cool project but after I initially implemented it on top of preact the architecture changed significantly which effectively meant that we could only support older versions. I think supporting it would still be nice but I would probably want to move it into a separate `panel-idom` package.

For now I've put a `IDOM` stub object into `panel.pane` that simply errors when you instantiate it. This is because we did not really telegraph the deprecation.


panel/pane/__init__.py
```diff
@@ -55,6 +54,16 @@
 from .vega import Vega  # noqa
 from .vtk import VTK, VTKVolume  # noqa
 
+
+class IDOM:
+    def __init__(self, *args, **kwargs):
+        raise RuntimeError(
+            "The IDOM pane was deprecated from Panel in version 1.0 "
```
Hoxbro, line 61:
```suggestion
            "The IDOM pane was removed from Panel in version 1.0 "
```

setup.py
```diff
@@ -146,7 +146,7 @@ def run(self):
     'markdown-it-py',
     'ipyvuetify',
     'reacton',
-    # Added lxml temporarily as installing pyechars or idom on Python 3.11
+    # Added lxml temporarily as installing pyecharts or idom on Python 3.11
```
Hoxbro, line 149:
```suggestion
    # Added lxml temporarily as installing pyecharts on Python 3.11
```

maximlt:
Shouldn't there first be a release deprecating it and then removing it? That might send the signal that Panel can remove things any time.

Issue #4260, opened 2023-01-08
Improve Snyk assessment


MarcSkovMadsen:
Snyk Advisor is a quick way to see if a package is *safe* to use. This might be used to pre-screen data app frameworks.

Both Streamlit and Dash score 97/100 points. Panel scores 90/100 points.

See https://snyk.io/advisor/python/panel

As far as I can see we are missing

- CONTRIBUTING.md
- Funding (?)

Snyk also evaluates the usage of Panel as *uncommon*. There are not so many public projects using the framework. I don't know if this contributes to the score.

[![image](https://user-images.githubusercontent.com/42288570/211185998-51aa5341-71b6-4bfa-becf-0457573be1c2.png)](https://snyk.io/advisor/python/panel).

philippjfr:
No objection to adding a CONTRIBUTING.MD. @maximlt @droumis I think we should add that as part of the overhaul of the developer docs.

On the funding front I don't really get how they assess that.

philippjfr:
Oh we should also have a documented security policy. I don't know offhand what the standard way of doing that is but we should set up a vulnerability disclosure program with a dedicated email address.

MarcSkovMadsen:
For inspiration here are links to a few CONTRIBUTING.md files

- [Dask](https://github.com/dask/dask/blob/main/CONTRIBUTING.md)
- [Streamlit](https://github.com/streamlit/streamlit/blob/develop/CONTRIBUTING.md)
- [Dash](https://github.com/plotly/dash/blob/dev/CONTRIBUTING.md)

MarcSkovMadsen:
We now have same score as comparable frameworks. The thing we can improve is addd a security policy.

PR #4262, opened 2023-01-08
Add contributing guide

MarcSkovMadsen:
Addresses part of https://github.com/holoviz/panel/issues/4260 by adding a CONTRIBUTING.md guide.

@philippjfr . You could consider giving this a quick review and merge for now. Then we can do a second iteration later where we dig deeper into improving the CONTRIBUTING docs in general?


philippjfr:
Happy with this as a first cut. We'll probably make adjustments as we overhaul the developer documentation.

PR #4264, opened 2023-01-08
Improve README

MarcSkovMadsen:
Closes #4263

I will be inspired by the READMEs listed in #4263. Especially the hvPlot README.

You can preview the readme [here](https://github.com/holoviz/panel/tree/feature/improve-readme).

PR #4278, opened 2023-01-13
Move enrironments docs down

MarcSkovMadsen:
This is a follow up to #4264. To temporarily address the comment by JBednar.

I believe the README will be more readable/ useful if we move the environments to the bottom of the README.


MarcSkovMadsen:
Arg. Go to many files merged when trying to fix windows encoding issue.

PR #4290, opened 2023-01-16
Migrate Customization User Guide content

droumis:
I moved most of the content into new How-to pages under `Create Components > Style Components` and `Create Components > Arrange and Size Components`. There wasn't much in terms of Background or Core Concept material. There may be some material that could also go into a reference, but we first need to align on the next step for reference material.

In both `apply_css.md` and `load_icon.md`, I added an admonition that says that the embedded code may not render on the page due to incompatibility with pyodide. I guess it could be some interaction with setting the panel configuration. This is just what I observed when built locally. Is this a bug or am I doing something incorrectly?

image

other previews:
image

image

image


doc/how_to/components/index.md
```diff
@@ -33,3 +35,66 @@ How to access and set widget values.
 How to add or remove components from a panel.
 :::
 ::::
+
+## Style Components
+
+::::{grid} 1 2 2 3
+:gutter: 1 1 1 2
+
+:::{grid-item-card} Apply CSS
```
philippjfr, line 44: If we can find icons octicon icons for all of these that would be fabulous: https://sphinx-design.readthedocs.io/en/latest/badges_buttons.html#octicon-icons

philippjfr (changes requested):
Looks great, thank you! I'll look at the issues you raised separately. One last thing, could you please remove the Customization.ipynb now that you've migrated the contents.

philippjfr (approved):
Okay, I removed it. Thanks again!

PR #4293, opened 2023-01-16
Deprecate ReactTemplate, IDOM pane and Viewable.app

philippjfr:
After some discussion I've decided to execute a few deprecations ahead of the 1.0 release. 

For templates I've simply decided that we do not need 5 different ListTemplate implementations and multiple grid template implementations. Deprecating some of these implementations will allow us to better focus on styling existing templates and reduces the maintenance burden.

Here are the deprecations I'm suggesting:

- `IDOM` pane: We only support version 0.24 currently which is very old. Since we do not have a maintainer who could update it and Reacton fills the same niche I've decided to remove it. It may be it could be re-added later in a `panel-idom` package but without anyone requesting support I think the right decision is to remove it.
- `ReactTemplate`: The `ReactTemplate` is overall pretty nice BUT the naming is terrible (it's built on top of a React library but it doesn't actually allow you to use React or anything so really that's an implementation detail. Secondly, the `FastGridTemplate` lets you do the exact same thing but looks prettier.
- `Viewable.app`: The `.app` method allows starting a server in Jupyter and then embedding it as an iframe. This relies on the port being accessible and to specify the `notebook_url`. This is brittle without much benefit at all because you already get interactivity in a notebook without running a server. I've left the function that will let you do the same thing but we don't need a method on all objects to do this.

Please chime in with thoughts and/or objections.  

Ping @MarcSkovMadsen, @jbednar, @maximlt, @Hoxbro

### DIFF
--- a/panel/template/react/__init__.py
+++ b/panel/template/react/__init__.py
@@ -9,6 +9,7 @@ from ...config import config
 from ...depends import depends
 from ...io.resources import CSS_URLS
 from ...layout import Card, GridSpec
+from ...util.warnings import deprecated
 from ..base import BasicTemplate
 from ..theme import DarkTheme, DefaultTheme
 
@@ -64,6 +65,8 @@ class ReactTemplate(BasicTemplate):
     }
 
     def __init__(self, **params):
+        deprecated('1.0', 'panel.template.ReactTemplate', 'panel.template.FastGridTemplate')
+        super().__init__(**params)
         if 'main' not in params:
             params['main'] = GridSpec(ncols=12, mode='override')
         super().__init__(**params)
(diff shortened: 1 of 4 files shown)


MarcSkovMadsen:
Sounds sensible. You could collect the bereaking changes you want to do. Announce them on discourse, Twitter and LinkedIn.

Then its hard for users to get upset if they did not react. And if you overlooked something you could get the feedback.

panel/template/react/__init__.py
```diff
@@ -64,6 +65,8 @@ class ReactTemplate(BasicTemplate):
     }
 
     def __init__(self, **params):
+        deprecated('1.0', 'panel.template.ReactTemplate', 'panel.template.FastGridTemplate')
+        super().__init__(**params)
```
Hoxbro, line 69:
```suggestion
```

Hoxbro:
I have no objection to the removal of these. 

Does this mean that VTK is being kept in Panel 1.0?

philippjfr:
> Does this mean that VTK is being kept in Panel 1.0?

I had planned to write a `panel-extension-cookiecutter` and then use it to distribute `panel-vtk` but I doubt I'll have time.

jbednar:
Sounds reasonable to me.

PR #4294, opened 2023-01-17
fix ref link to multi app serving

droumis:
As identified by 'phidias' user on gitter


philippjfr:
Thanks, this will have to be updated again when we make the how-to guides go live.

Issue #4305, opened 2023-01-20
Proposal for Reference Gallery


philippjfr:
Our reference gallery serves as a valuable visual index of all the different components that Panel has to offer. Its strength lies is the fact that it is both browsable and searchable, unlike a more classical API reference. However once we reach the actual individual reference guides it becomes a little messier. They all follow a general structure that roughly follows:

- Description of what the component is and what it does.
- Some high-level comments about usage
- A list of parameters that are specific to that component and what they do
- Now come a number of usage example that highlight basic usage and then a non-exhaustive list of examples
- Lastly most Pane and Widget components have a `Controls` section that uses a JS linked form to allow you to play with at least some of the parameters.

There are two main problem with this approach:

1. The high-level description and list of parameters and their descriptions are completely disconnected from the docstrings and the code so it's easy for them to get out of sync and they are annoying to update.
2. Usage examples aren't standalone and if you want to contribute one you have to make it fit into the flow of the overall page.

## Proposal

What I propose is that we automatically generate these pages by pulling content from a number of places:

1. The description of the component should be pulled directly from the docstring
2. The list of parameters and their descriptions should be automatically generated from the code
3. We automatically embed a collapsible section of any inherited parameters
4. The examples section will be generated automatically by pulling in small, self-contained examples.

For 4. I envision the individual examples as very small how-tos that demonstrate one thing about that particular component. We would then need some mechanism to tell it which of these examples would get rendered directly on the page and which would live in a small sub-gallery.

## Sketch

<img width="409" alt="Screen Shot 2023-01-20 at 11 38 58" src="https://user-images.githubusercontent.com/1550771/213675702-36c8ce1b-e472-4fb3-b348-3bda56efc6eb.png">

maximlt:
I think 1., 2. and 3. would be nice improvements compared to the current state.

If you add to that the list of methods supported by the component (I guess in a collapsed container), doesn't this mean that there's actually no need of the component to be in the API reference? (Making it easier to merge the API reference and visual reference)

I am much more doubtful about the usefulness of 4:

- I don't think the Reference Gallery in its current state is so bad (e.g. compared to how the user guides, the API reference and to some extent the Gallery which is a mix of How-Tos and Demos) that it needs such a big change. The Reference Gallery is currently and imo the best section of the whole documentation.
- I really like to have every thing in a single page and do CTRL+F to look for what I'm after. I do it very often. Being able to quickly scroll over a component's page to see what it offers is also nice.
[...]

PR #4311, opened 2023-01-21
Migrate `Interact` and remove `Widgets` User Guides

droumis:
Interact User Guide
- Most of the content has been reworked into How-tos:
  - Generate Widgets for Function Args
  - Customize Layout with `Interact`
  - Fix Values with `Interact`
  - Disable Continuous Updates with `Interact`
- The interact widget abbreviations table has been used to initiate a new background page

Widgets User Guide:
- Sections about creating a widget, setting/accessing value, layout, and `param.watch` are already covered in other how-tos.
- Section listing types of widgets is incomplete and better represented by the thumbnail index of the reference gallery (which will be the better way to access their reference material as we improve the reference sections). 

Also:
- Changing the linked resources at the bottom of the pages from being an admonition to a subsection so we can use consistent linking and styling rather than having to use raw HTML in admonitions.
- Starting each newly created how-to with a quick statement about what this page solves or addresses. It would nice to style this consistently in some way to distinguish from the rest of the text.
- I still need to add a link to the reference material for `panel.interact`, but this can wait until after to revamp the reference section (see #4305).

`Interact` How-to's previews:
image
image
image


philippjfr (approved):
Happy to merge this for now but I want to consider whether a whole section for Interact isn't overkill (I don't even like `interact`).

PR #4314, opened 2023-01-22
Rework `Template` User Guide into How-to and Background

droumis:
- Reworked most of the Template User Guide into new How-To guides
- Some of the explanation has been used to initiate a templates background page

image

image

image


philippjfr (approved):
This is great, thank you @droumis! I'll probably do a fair bit of work on these anyway so I won't review in detail right now.

MarcSkovMadsen:
One thing i believe is really lacking are nice notebook templates. My initial FastListTemplate worked there. But for some unknown reason, the current implementation does not work there. And there are not many ways to make things look great in notebooks. IVuewidgets is the only way I know of.

philippjfr:
Watch this space 😉

Issue #4317, opened 2023-01-23
Can't serve local __javascript_modules__ files in ReactiveHtml


c-meier:
#### ALL software version info
panel = "0.14.2"

#### Description of expected behavior and the observed behavior

#3284 added support for local ressources for ReactiveHtml components (as seen in #3241), it works for `__css__` and `__javascript__`, however it doesn't work for `__javascript_modules__`.

#### Complete, minimal, self-contained example code that reproduces the issue

Un fichier javascript de test: `test.mjs`

```js
console.log("Test local module script")
```

```python
from panel.reactive import ReactiveHTML

class SimpleComponent(ReactiveHTML):
    _template = """
    <p id="test">Hello World!</p>
    """
    
    __javascript_modules__ = [
        'test.mjs',
    ]
 

if __name__ == '__main__':
    test_component = SimpleComponent()
    test_component.servable()
```

#### Possible solution

I resolved the issue by modifying two lines of code:

At `panel/io/resources.py` in `Bundle.__init__`:
```diff
- js_modules.append(js_module)
+ js_modules.append(component_resource_path(model, '__javascript_modules__', js_module))
```

At `panel/io/server.py` in `ComponentResourceHandler._resource_attrs`:
```diff
  _resource_attrs = [
-     '__css__', '__javascript__', '__js_module__',  '_resources',
+     '__css__', '__javascript__', '__js_module__', '__javascript_modules__',  '_resources',
      '_css', '_js', 'base_css', 'css'
  ]
```

However I don't know the consequences of doing this change.

MarcSkovMadsen:
Thx. It would help if you made a PR with the proposed fixes. If a test could be added it would also be very helpful 👍

PR #4319, opened 2023-01-24
Fix bundling of javascript modules in ReactiveHTML

MarcSkovMadsen:
Fix suggested by @c-meier in #4317.

I've just started the PR to help Phillip or others finish it.

I don't intend to do more now.

## Todo

- Architect and implement solution
- Add tests


philippjfr:
Will follow up with tests separately.

Issue #4341, opened 2023-01-27
Ideas for Tabulator improvements when comparing to AG Grid + Dash


MarcSkovMadsen:
I've taken the time to compare `Tabulator + Panel` to `AG Grid + Dash` and found some ideas for improvement that I would like to share here.

## Summary Results

The Dash [AG Grid announcment](https://medium.com/plotly/announcing-dash-ag-grid-fbb4a1c83e62) contains a high level summary table that I've updated for Tabulator + Panel below.

![image](https://user-images.githubusercontent.com/42288570/215016212-ecce7564-e41e-4bca-8bbb-464a5cd25619.png)

## Survey

[PanelTabulatorvsDashAGrid.csv](https://github.com/holoviz/panel/files/10515845/PanelTabulatorvsDashAGrid.csv)

(rename to `PanelTabulatorvsDashAGrid.ipynb`)

### Column Settings for specific columns

For column settings not support by Panel Tabulator we can try to define them as `columnDefaults`

```python
configuration = {
    "columnDefaults":{
        "tooltip": True,
    }
}
```

But I cannot find any way to do this for specific columns. For example it powerful to be able to define *resizable columns* or *tooltips* for specific columns.

As there will always be some settings that we don't support yet, it would be really nice to have some fallback solution. like

```python
Tabulator(df, column_configuration={
    "column_x", {resizable: False, tooltip: True}
}
```

The `column_configuration` would be used as the starting point for the Panel Tabulator column configuration.

## moveable_columns parameter

`movableColumns` seem to be set to `true` by default. But would be nice to expose as a `moveable_columns` parameter.

## Custom HTML Tooltips

It seems we can set `toolitip` to `true` in general. But these tooltips are not very nice as they are not formatted. So for example dates are not shown nicely.

It would be nice to have custom html tooltips similar to what Bokeh plots can offer. An alternative would be a pop up/ modal.

MarcSkovMadsen:
We now have an example doing a direct comparison in #4350 

![image](https://user-images.githubusercontent.com/42288570/215251637-8308ae2f-97de-4538-946c-9d674c515802.png)

Khalilsqu:
Aggregation function are availabe with perspective!

robml:
@Khalilsqu what do you mean they are available with perspective? I don't see how we can add something like bottomCalc yet.

maxrausch:
Is it possible for a developer to use ag-grid with panel as a custom component?

jmgablon:
[...]

Issue #4327, opened 2023-01-26
A example showing how to build a dynamically updating, excel like table with Tabulator


MarcSkovMadsen:
In Excel user develop dynamic tables all the time. I.e. they can change the value in a cell or column and other cells or columns update.

The Tabulator documentation does not contain any example showing how to best do this. Some of my colleagues experienced it was difficult to develop a working concept. It would really help with an example to start from.

Please add. Thanks.

PR #4350, opened 2023-01-28
add excel like editing example

MarcSkovMadsen:
Contributes a Portfolio Analyser app.

This contribution is heavily inspired by a [Dash + AG Grid Example](https://github.com/plotly/dash-ag-grid/blob/dev/docs/demo_stock_portfolio.py).

It both provides an example of an excel like table for #4327 and makes it possible compare Panel+Tabulator to Dash + AG Grid as in #4341. It also adds a financial example to the gallery which I also believe is important.

portfolio-analysis

Serve it with

```bash
panel serve examples/gallery/demos/portfolio_analyser.py --autoreload
```

## Todo

- Convert to notebook
- Add some additional quant columns
- Add support for clicking donut also
- Showcase as much of Tabulators Features as possible.
- Add *download data to excel* and *download as html report* buttons
- Use the Modal for something.

## Issues (To be updated)

- #4351
- #4353 
- #4354
- #4356
- patching with a slice with no start  `slice(8)` raises errors. `slice(0,8)` works.
- #4357
- #4361
- Cannot use yfinance with panel convert due to lxml and CORS issue.

## Feature Requests (To be updated)

- Support Keyboard navigation in Tabulator
- Support copy paste into the Tabulator table

## Ideas

- It could be powerful to also provide versions using HoloViews plots, Altair and Echarts plots.

### DIFF
--- a/examples/gallery/demos/portfolio_analyser.py
+++ b/examples/gallery/demos/portfolio_analyser.py
@@ -7,10 +7,11 @@ Having both its possible to compare them. See also
 https://github.com/holoviz/panel/issues/4341
 """
 import pandas as pd
-import panel as pn
 import plotly.express as px
 import plotly.graph_objects as go
 import yfinance as yf
+
+import panel as pn
 
 ACCENT = "#BB2649"
 LINK_SVG = """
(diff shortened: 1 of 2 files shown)


philippjfr:
Going to merge and rework it slightly when I add the new "featured" gallery.

philippjfr (approved):

Issue #4325, opened 2023-01-26
Responsive Holoviews plot has zero height inside column with sizing_mode="stretch_both"


MarcSkovMadsen:
panel==0.14.2, bokeh==2.4.4

DrAstro cannot get his heatmap to `stretch_both` in https://discourse.holoviz.org/t/responsive-sizing-mode-and-pn-depends-heatmap-rendering-with-0-height/4863/5.

![image](https://user-images.githubusercontent.com/42288570/214768019-e80c6b65-82fd-45e2-b89c-1a0202a0b03f.png)

```python
import pandas as pd
import hvplot.pandas
import numpy as np
import panel as pn
from bokeh.sampledata import sea_surface_temperature as sst

df = sst.sea_surface_temperature

df['sea'] = 'sea1'

df2 = df.copy()
df2['sea'] = 'sea2'
df2_temps = df2['temperature'].to_numpy()
rng = np.random.default_rng()
rng.shuffle(df2_temps)
df2['temperature'] = df2_temps

df_plot = pd.concat([df, df2])

group_select = pn.widgets.Select(name='Select Sea', options=df_plot['sea'].unique().tolist(), value='sea1')

@pn.depends(group_select)
def get_heatmap(group_select):
    to_plot = df_plot.groupby('sea').get_group(group_select)
    return to_plot.hvplot.heatmap(
        x='time.month', 
        y='time.day',
        C='temperature',
        title=f"Sea: {group_select}",
        responsive=True
    )

pn.Column(pn.WidgetBox(group_select), get_heatmap, sizing_mode='stretch_both').servable()
```

## Request

- Explain why this happens. We need a mental model of this in the community
- Fix the issue, make it easier to fix for users or document what to do

## Question for Philipp: How to solve?

Could you guide us here @philippjfr ? Which solution do you see/ prefer?

## Workaround

I have seen issues like these from time to time and have a list of things that I would try. Here you can wrap `get_heapmap` in a `pn.panel` with `sizing_mode="stretch_both"`. But new users would not have that list anywhere. They would be stuck.

![image](https://user-images.githubusercontent.com/42288570/214767869-db538cd6-5d20-4566-a364-b0edcd68b2c4.png)

```python
import pandas as pd
import hvplot.pandas
import numpy as np
import panel as pn
from bokeh.sampledata import sea_surface_temperature as sst

df = sst.sea_surface_temperature

df['sea'] = 'sea1'

df2 = df.copy()
df2['sea'] = 'sea2'
[...]

Issue #2774, opened 2021-09-23
Make it easy to have stretch_both for holoviews plot


Hoxbro:
#### ALL software version info

<details>
  <summary> doit env_capture </summary>

``` yaml
.  env_capture


     active environment : base

    active env location : C:\Users\shh\miniconda3\envs\holoviz

            shell level : 2

       user config file : C:\Users\shh\.condarc

 populated config files : C:\Users\shh\.condarc

          conda version : 4.10.3

    conda-build version : 3.21.4

         python version : 3.7.10.final.0

       virtual packages : __cuda=11.3=0

                          __win=0=0

                          __archspec=1=x86_64

       base environment : C:\Users\shh\miniconda3\envs\holoviz  (writable)

      conda av data dir : C:\Users\shh\miniconda3\envs\holoviz\etc\conda

  conda av metadata url : None

           channel URLs : https://conda.anaconda.org/pyviz/win-64

                          https://conda.anaconda.org/pyviz/noarch

                          https://conda.anaconda.org/conda-forge/win-64

                          https://conda.anaconda.org/conda-forge/noarch

                          https://repo.anaconda.com/pkgs/main/win-64

                          https://repo.anaconda.com/pkgs/main/noarch

                          https://repo.anaconda.com/pkgs/r/win-64

                          https://repo.anaconda.com/pkgs/r/noarch

                          https://repo.anaconda.com/pkgs/msys2/win-64

                          https://repo.anaconda.com/pkgs/msys2/noarch

          package cache : C:\Users\shh\miniconda3\envs\holoviz\pkgs

                          C:\Users\shh\.conda\pkgs

                          C:\Users\shh\AppData\Local\conda\conda\pkgs

       envs directories : C:\Users\shh\miniconda3\envs\holoviz\envs

                          C:\Users\shh\.conda\envs

                          C:\Users\shh\AppData\Local\conda\conda\envs

               platform : win-64

             user-agent : conda/4.10.3 requests/2.26.0 CPython/3.7.10 Windows/10 Windows/10.0.19041

          administrator : False

             netrc file : None

           offline mode : False



# packages in environment at C:\Users\shh\miniconda3\envs\holoviz:

#

# Name                    Version                   Build  Channel

anyio                     3.3.0            py37h03978a9_0    conda-forge

argcomplete               1.12.3             pyhd8ed1ab_2    conda-forge

argon2-cffi               20.1.0           py37hcc03f2d_2    conda-forge

[...]

PR #4360, opened 2023-01-29
Propagate sizing mode

philippjfr:
This PR ensures that any dynamic components like `ParamFunction`, `ParamMethod` and `Interactive` reflect the parameters (especially the `sizing_mode`) of the components they are wrapping. It also propagates responsive sizing modes on a HoloViews/Bokeh object to the pane wrapping it ensuring that sizing modes and other options can propagate upwards from a plot all the way to a dynamic pane.

- Fixes https://github.com/holoviz/panel/issues/2774
- Fixes https://github.com/holoviz/panel/issues/4325


philippjfr:
pre-commit.ci autofix

PR #4377, opened 2023-02-01
Add note about raw strings for LaTeX

### LINKED ISSUE #4376: LaTeX in Markdown not working unless you provide raw strings
I'm on Panel 0.14.3.

A colleague of mine wanted to start including LaTeX in his app. He was using the Markdown pane as is should support LaTeX. He could not get it working. After some initial confusion we figured out its because the markdown string provided needs to be a raw string.

This was quite confusing as the documentation did not make that clear. Please make this clear.

https://panel.holoviz.org/reference/panes/Markdown.html

image

## Reproducible Example

```python
import panel as pn

pn.extension("mathjax")

pn.pane.Markdown("$$\frac{1}{n}$$")
pn.pane.Markdown(r"$$\frac{1}{n}$$")
```

image

MarcSkovMadsen:
Closes #4376

### DIFF
--- a/examples/reference/panes/LaTeX.ipynb
+++ b/examples/reference/panes/LaTeX.ipynb
@@ -41,9 +41,28 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "latex = pn.pane.LaTeX('The LaTeX pane supports two delimiters: $LaTeX$ and \\(LaTeX\\)', \n",
+    "latex = pn.pane.LaTeX(r'The LaTeX pane supports two delimiters: $LaTeX$ and \\(LaTeX\\)', \n",
     "                      style={'font-size': '18pt'}, width=800)\n",
     "latex"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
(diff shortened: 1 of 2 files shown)


examples/reference/panes/LaTeX.ipynb
```diff
@@ -41,11 +41,30 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "latex = pn.pane.LaTeX('The LaTeX pane supports two delimiters: $LaTeX$ and \\(LaTeX\\)', \n",
+    "latex = pn.pane.LaTeX(r'The LaTeX pane supports two delimiters: $LaTeX$ and \\(LaTeX\\)', \n",
```
MarcSkovMadsen, line 44: I added the `r` every where. I believe this will help users. If you see a working example without it, and your example does not work, then the cause is not very clear.

MarcSkovMadsen:
I cannot see how the failing test on Mac OS Python 3.11 is in any way related to this PR. I don't plan to do more before feedback is provided :-)

philippjfr (approved):
Looks good, thanks!

PR #4378, opened 2023-02-02
Fix serialization of buffer headers in notebook

philippjfr:
Seemingly buffer headers were json serialized twice causing issues when deserializing them. This did not seem to cause issues for bokeh <3 but does now.


philippjfr:
pre-commit.ci autofix

PR #4380, opened 2023-02-02
RGGPlot: rename the method that gets the image data

maximlt:
This PR renames the RGGPlot `_img` method to `_data`. This pane being untested this slipped through some (old?) refactoring. The pane being undocumented too, nobody complained about it because presumably nobody's really using it :)


philippjfr:
Yeah considering deprecating it but for now I'll merge.

PR #4383, opened 2023-02-02
More specific front-end parameter typing for SpeechToText

maximlt:
Saw the following error working on the branch 1.0 as long as with an error about `results` (didn't catch the error in the log):

> Error rendering Bokeh items: Error: panel.models.speech_to_text.SpeechToText(p1030).service_uri is unset


philippjfr (approved):
Looks good, thanks!

PR #4386, opened 2023-02-03
Standardize parameter mapping APIs

philippjfr:
Over time we accumulated a bunch of workaround for mapping parameters to properties. Specifically we had the `_rename` dictionary used for mapping from parameter to property names, then we had various `_init_params` and `_get_properties` methods used to provide parameters to initialize and update properties on the underlying bokeh model. This PR attempts to rationalize this a bit:

- `_rename` consistently used to declare parameter mapping (i.e. avoid manual workarounds for this)
- Bokeh models are now always initialized by calling `_get_properties` which itself calls `_init_params` and then transforms those parameters with `_process_param_change`, ensuring that as much logic for mapping from Panel parameters to Bokeh properties is encapsulated in this method.
- Introduce a `ModelPane` baseclass for panes that very simply map to an underlying bokeh model and apply some transformations to parameters.

### DIFF
--- a/panel/layout/base.py
+++ b/panel/layout/base.py
@@ -72,11 +72,13 @@ class Panel(Reactive):
     # Callback API
     #----------------------------------------------------------------
 
-    def _init_params(self):
-        return {p: v for p, v in self.param.values().items()
-                if v is not None and p != 'objects'}
+    def _init_params(self) -> Dict[str, Any]:
+        return {
+            p: v for p, v in self.param.values().items()
+            if v is not None and p != 'objects'
+        }
 
-    def _get_properties(self):
+    def _get_properties(self) -> Dict[str, Any]:
         return self._process_param_change(self._init_params())
 
     def _update_model(
--- a/panel/layout/base.py
+++ b/panel/layout/base.py
@@ -790,7 +790,7 @@ class WidgetBox(ListPanel):
 
     @param.depends('disabled', 'objects', watch=True)
     def _disable_widgets(self) -> None:
-        from ..widget import Widget
+        from ..widgets import Widget
         for obj in self.select(Widget):
             obj.disabled = self.disabled
 
--- a/panel/layout/base.py
+++ b/panel/layout/base.py
@@ -41,9 +41,6 @@ class Panel(Reactive):
     # Bokeh model used to render this Panel
     _bokeh_model: ClassVar[Type[Model]]
 
-    # Properties that should sync JS -> Python
-    _linked_props: ClassVar[List[str]] = []
-
     # Parameters which require the preprocessors to be re-run
     _preprocess_params: ClassVar[List[str]] = []
 
@@ -155,7 +152,7 @@ class Panel(Reactive):
         objects = self._get_objects(model, [], doc, root, comm)
         model.update(**{self._property_mapping['objects']: objects})
         self._models[root.ref['id']] = (model, parent)
-        self._link_props(model, self._linked_props, doc, root, comm)
+        self._link_props(model, self._linked_properties, doc, root, comm)
         return model
 
     #----------------------------------------------------------------
@@ -678,11 +675,11 @@ class NamedListPanel(NamedListLike, Panel):
     __abstract = True
 
     def _process_param_change(self, params: Dict[str, Any]) -> Dict[str, Any]:
-        scroll = params.pop('scroll', None)
-        css_classes = self.css_classes or []
-        if scroll:
-            params['css_classes'] = css_classes + ['scrollable']
-        elif scroll == False:
+        if 'scroll' in params:
+            scroll = params.pop('scroll')
+            css_classes = list(self.css_classes or [])
+            if scroll:
+                css_classes += ['scrollable']
             params['css_classes'] = css_classes
         return super()._process_param_change(params)
 
--- a/panel/layout/base.py
+++ b/panel/layout/base.py
@@ -777,7 +777,7 @@ class WidgetBox(ListPanel):
     }
 
     _rename: ClassVar[Mapping[str, str | None]] = {
-        'objects': 'children', 'horizontal': None
+        'disabled': None, 'objects': 'children', 'horizontal': None
     }
 
     @property
@@ -786,9 +786,9 @@ class WidgetBox(ListPanel):
 
     @param.depends('disabled', 'objects', watch=True)
     def _disable_widgets(self) -> None:
-        for obj in self:
-            if hasattr(obj, 'disabled'):
-                obj.disabled = self.disabled
+        from ..widget import Widget
+        for obj in self.select(Widget):
+            obj.disabled = self.disabled
 
     def __init__(self, *objects: Any, **params: Any):
         super().__init__(*objects, **params)
--- a/panel/layout/base.py
+++ b/panel/layout/base.py
@@ -75,6 +75,13 @@ class Panel(Reactive):
     # Callback API
     #----------------------------------------------------------------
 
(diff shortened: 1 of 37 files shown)


panel/pane/image.py
```diff
@@ -402,26 +422,20 @@ class PDF(FileBase):
...
-        p["embed"] = self.embed
-        p["start_page"] = self.start_page
-        p["width"] = self.width
-        p["height"] = self.height
+    _rerender_params: ClassVar[List[str]] = FileBase._rerender_params + ['start_page']
 
-        if self.object is None:
-            return dict(p, text='<embed></embed>')
-        if self.embed:
+    def _transform_object(self, obj):
```
philippjfr, line 429:
```suggestion
    def _transform_object(self, obj: Any) -> Dict[str, Any]:
```

panel/pane/markup.py
```diff
@@ -243,32 +226,33 @@ def _cleanup(self, root: Model | None = None) -> None:
...
-                          if p not in DivPaneBase.param and p != '_object'}
-                html = df.to_html(**kwargs)
+                          if p not in HTMLBasePane.param and p != '_object'}
+                html = obj.to_html(**kwargs)
         else:
             html = ''
-        return dict(properties, text=escape(html))
+        return dict(object=escape(html))
+
+    def _init_params(self):
```
philippjfr, line 248:
```suggestion
    def _init_params(self) -> Dict[str, Any]:
```

panel/pane/markup.py
```diff
@@ -430,13 +422,22 @@ def applies(cls, obj: Any, **params) -> float | bool | None:
...
         except Exception:
-            data = self.object
+            data = obj
         text = json.dumps(data or {}, cls=self.encoder)
-        depth = None if self.depth < 0 else self.depth
-        return dict(text=text, theme=self.theme, depth=depth,
-                    hover_preview=self.hover_preview, **properties)
+        return dict(object=text)
+
+    def _process_property_change(self, properties):
```
philippjfr, line 433:
```suggestion
    def _process_property_change(self, properties: Dict[str, Any]) -> Dict[str, Any]:
```

panel/pane/markup.py
```diff
@@ -430,13 +422,22 @@ def applies(cls, obj: Any, **params) -> float | bool | None:
...
+        properties = super()._process_property_change(properties)
+        if 'depth' in properties:
+            properties['depth'] = -1 if properties['depth'] is None else properties['depth']
+        return properties
+
+    def _process_param_change(self, params):
```
philippjfr, line 439:
```suggestion
    def _process_param_change(self, params: Dict[str, Any]) -> Dict[str, Any] :
```

panel/reactive.py
```diff
@@ -128,6 +130,25 @@ def __init__(self, **params):
...
+        return rename
+
+    @property
+    def _linked_properties(self):
+        return tuple(
+            self._property_mapping.get(p, p) for p in self.param
+            if p not in Viewable.param and self._property_mapping.get(p, p) is not None
+        )
+
+    def _get_properties(self):
```
philippjfr, line 149:
```suggestion
    def _get_properties(self) -> Dict[str, Any]:
```

panel/reactive.py
```diff
@@ -128,6 +130,25 @@ def __init__(self, **params):
     # Model API
     #----------------------------------------------------------------
 
+    @classproperty
+    @lru_cache(maxsize=None)
+    def _property_mapping(cls):
```
philippjfr, line 135:
```suggestion
    def _property_mapping(cls) -> Dict[str, str | None]:
```

panel/widgets/select.py
```diff
@@ -244,11 +247,10 @@ def _validate_options_groups(self, *events):
             )
 
     def _process_param_change(self, msg):
```
philippjfr, line 249:
```suggestion
    def _process_param_change(self, msg: Dict[str, Any]) -> Dict[str, Any]:
```

panel/widgets/select.py
```diff
@@ -312,6 +314,8 @@ class _MultiSelectBase(SingleSelectBase):
 
     _supports_embed: ClassVar[bool] = False
 
+    __abstract = True
+
     def _process_param_change(self, msg):
```
philippjfr, line 319:
```suggestion
    def _process_param_change(self, msg: Dict[str, Any]) -> Dict[str, Any]:
```

Hoxbro:
I would like to see all the type changes be made in another PR.

philippjfr:
Fair, just to clarify that is for review purposes, to avoid merge conflicts or something else?

Hoxbro:
Review purposes and to be able to look at the PR in the future if needed.

philippjfr:
Done, please review in that case 😄

philippjfr:
Want to make progress elsewhere (and merge this into the UI test PR) so I'll probably go ahead and merge. This is extremely difficult to review but if you do find the time/desire to do so I'll apply any fixes in a new PR.

panel/layout/base.py
```diff
@@ -616,6 +625,8 @@ class ListPanel(ListLike, Panel):
         Whether to add scrollbars if the content overflows the size
         of the container.""")
 
+    _rename: ClassVar[Mapping[str, str | None]] = {'scroll': None}
```
Hoxbro, line 628: `Mapping[str, str | None]` is used a lot maybe we should make it a type alias?

panel/pane/alert.py
```diff
@@ -30,7 +33,7 @@ class Alert(Markdown):
 
     priority: ClassVar[float | bool | None] = 0
 
-    _rename: ClassVar[Mapping[str, str | None]] = dict(Markdown._rename, alert_type=None)
+    _rename: ClassVar[Mapping[str, str | None]] = {'alert_type': None}
```
Hoxbro, line 36: Why was `Markdown._rename` removed?

panel/pane/base.py
```diff
@@ -1,13 +1,13 @@
 """
-Defines the PaneBase class defining the API for panes which convert
-objects to a visual representation expressed as a bokeh model.
+Defines base classes for Pane components which allow wrapping a Python
+object transforming it into a Bokeh model that can be rendered.
```
Hoxbro, line 3: Would probably update with:
`...object and transforming...`
`...that can be rendered in the browser.`

panel/reactive.py
```diff
@@ -1651,6 +1683,7 @@ def _get_template(self) -> Tuple[str, List[str], Mapping[str, List[Tuple[str, Li
                 html = html.replace('${%s}' % child_name, '')
         return html, parser.nodes, p_attrs
 
+    @property
     def _linked_properties(self) -> List[str]:
```
Hoxbro, line 1687: Should be returning tuple type.

panel/widgets/indicators.py
```diff
@@ -274,7 +268,7 @@ class Number(ValueIndicator):
     title_size = param.String(default='18pt', doc="""
         The size of the title given by the name.""")
 
-    _rename: ClassVar[Mapping[str, str | None]] = {}
+    _rename: ClassVar[Mapping[str, str | None]] = {'name': 'name'}
```
Hoxbro, line 271: I'm not sure I follow this rename... Is it because the parent class rename `name`?

Hoxbro (commented):
Have added some comment.

Add some point I think we should use `list`/`tuple`/`dict` instead of  `List`/`Tuple`/`Dict`. This can be done automatic with `pyupgrade` or `ruff`.

PR #4411, opened 2023-02-06
Add Pandas to the runtime dependencies

maximlt:
Bokeh 3 directly depends on Pandas (`>=1.2`) and it makes sense for Panel to also depend on it, which is what this PR does. However, as Bokeh does, Pandas is never a top-level import (except in the tests) to avoid increasing the import time for non-Pandas users.


philippjfr:
Thanks for this, test failures unrelated.

PR #4412, opened 2023-02-06
Custom components User Guide reorg

droumis:
Migrates the Custom components User Guide into:

1. How To about Viewer components (combining existing components)
2. How To about ReactiveHTML (building from scratch recipes)
3. Background about ReactiveHTML (building from scratch, with all the discussion and details)

image


doc/how_to/components/custom_reactiveHTML.md
```diff
@@ -0,0 +1,180 @@
+# Build Components from Scratch
```
philippjfr, line 1: Would suggest to move these to a "Building Custom Components" section in the advanced section of the How To.

philippjfr (approved):
Looks great but I'd probably suggest we move the two new How To guides into a section about extending Panel.

doc/how_to/components/custom_reactiveHTML.md
droumis: hmm it makes sense to me to have subsections by topic OR difficulty, but not both; i.e. to not have an 'Advanced' section of how-to's. I could imagine that each topic section could have advanced how-to's and we probably wouldn't want to separate them from related guides (of any difficulty) in the main view. I think I prefer some sort of minimal tag like I've done here to label the advanced ones as a sort of warning. I don't think anyone will start their search looking specifically for 'Advanced' guides of any topic... 

I'm ok with moving them both to a new section about extending Panel

philippjfr: We already have an advanced section. It's simply a heading in the index.md.

philippjfr: I appreciate what you're saying though, thematic subsections should be grouped together whether they are simple or advanced. I think it's reasonable to break down the actual subsections into core and advanced groupings.

philippjfr:
Thanks @droumis!

PR #4423, opened 2023-02-08
Tabulator: keywords and like filters aren't regex

maximlt:
`Tabulator JS` defines the *like* and *keywords* filters as follows:

> The like filter displays any rows with data that contains the specified string anywhere in the specified field. (case insensitive)

> The keywords filter displays any rows with data containing any space separated words in the specified string (case insensitive)

These filters were implemented with Pandas' `.str.contains` method that by default expects the pattern to be a regular expression. This PR sets the Pandas `regex` parameter to `False` to avoid that. Additionally it sets the `case` parameter to `False` too as as described in `Tabulator JS`' docs these two filters are case insensitive. There was no issue with that as all the strings compared were lower-cased (I believe `.casefold()` is a better fit to compare strings ignoring their case), it's just a little cleaner.

### DIFF
--- a/panel/widgets/tables.py
+++ b/panel/widgets/tables.py
@@ -480,7 +480,7 @@ class BaseTable(ReactiveData, Widget):
             elif op == 'in':
                 filters.append(col.isin(val))
             elif op == 'like':
-                filters.append(col.str.lower().str.contains(val.lower()))
+                filters.append(col.str.contains(val, case=False, regex=False))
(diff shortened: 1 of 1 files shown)


philippjfr (approved):
Looks good and makes sense. A test would be great!

philippjfr:
Oops, meant to merge another PR but was on my phone. Just add the test sometime.

Issue #4420, opened 2023-02-08
pn.pane.JPG throws ValueError is pointed to an `.jpeg` file


krlng:
The [JPG doc](https://panel.holoviz.org/reference/panes/JPG.html) states the component would understand `.jpeg` files. However, if you point it to one, it throws: `ValueError: JPG pane cannot parse a string that is not a filename or URL.`.

When I rename the `.jpeg` file to `.jpg` everything works nicely.

MarcSkovMadsen:
Hi @krlng . Would you be interested in contributing a PR to fix this? If YES please let us know if you need help getting started.

krlng:
I had a short look and think it is not as easy as one might think at first hand. 

To do this properly, [filetype](https://github.com/holoviz/panel/blob/main/panel/pane/image.py#L30) should be changed to a list. However [applies](https://github.com/holoviz/panel/blob/main/panel/pane/image.py#L54) and `_b64` and `_data` methods all use it and are not documented. So I am unsure, what they do exactly and how do deal with this. 

I guess this is something philipp can do really fast, but which would take me a couple of hours (I sadly don't have right now)

krlng:
Btw: I just saw in the source code you added a general Image component. That is just great. I actually had something small to do the same on my own and I guess many who worked with the image components had the same....

krlng:
Okay, sorry for texting you so much. If there is a new Image component, users most likely will not use the dedicated file-type extensions much more. So it would probably also be okay to just add a dedicated JPEG component (which basically means just copy-paste the JPG). This is a bit redundant, how ever, it keeps code less complex which is always a good thing for open source. 

So if you give me a thumbs up, I can just add the dedicated JPEG component.

philippjfr:
Fair points, had missed this background when you submitted the PR. I'll mull it over a bit.

PR #4427, opened 2023-02-09
add JPEG image class

krlng:
solves https://github.com/holoviz/panel/issues/4420

### DIFF
--- a/panel/pane/image.py
+++ b/panel/pane/image.py
@@ -308,7 +308,7 @@ class ICO(ImageBase):
 
 class JPG(ImageBase):
     """
-    The `JPG` pane embeds a .jpg or .jpeg image file in a panel if provided a
+    The `JPG` pane embeds a .jpg image file in a panel if provided a
     local path, or will link to a remote image if provided a URL.
 
     Reference: https://panel.holoviz.org/reference/panes/JPG.html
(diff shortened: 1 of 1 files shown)


panel/pane/image.py
```diff
@@ -344,6 +344,26 @@ def _imgshape(cls, data):
         return int(w), int(h)
 
 
+class JPEG(JPG):
+    """
+    The `JPEG` pane embeds a .jpeg image file in a panel if provided a
+    local path, or will link to a remote image if provided a URL.
+
+    Reference: https://panel.holoviz.org/reference/panes/JPEG.html
```
jbednar, line 352: I don't know if having a separate class for this is the _best_ approach, but it's what I thought of as well when I saw the original issue. That said, this link points to nowhere, so if other people think this is the way to go, I think the docstring should just say something like "Same as pn.pane.JPG, but accepting an extension .jpeg" and not trying to have a separate reference page.

philippjfr:
Thanks @krlng! Really appreciate the effort but I think we should just have the `JPG` component support `.jpeg` files. The filetype abstraction isn't ideal but I think the only place we have to come up with a different approach is here:

```python
    @classmethod
    def applies(cls, obj: Any) -> float | bool | None:
         filetype = cls.filetype.split('+')[0]
         ...
             if isfile(obj) and obj.lower().endswith(f'.{filetype}'):
         ...
```

philippjfr:
Thanks again @krlng, I've added the ability to override the supported extensions with a `_extensions` attribute and will merge that in as part of https://github.com/holoviz/panel/pull/4532.

PR #4431, opened 2023-02-09
fix dataframe pane layout issues

### LINKED ISSUE #4426: pn.panel and pn.pane.DataFrame highly problematic
panel==0.14.3

I was trying to help a colleague get into Panel and wanted to write a getting started tutorial for my team. 

In my team the Pandas DataFrame is the starting point for much of our work. So great support for displaying DataFrames is key

I wanted to describe the basic concepts and I believe(d) that `pn.panel` is key. Its the `print` statement of Panel and similar to `st.write` in Streamlit.

But when you use `pn.panel` on DataFrames they don't display nicely. Neither in a Notebook or in a server App. And its almost impossible to get them to do this if you have lots of data as you cannot control the scrollbars.

Please make `pn.panel` and `pn.pane.DataFrame` an easy to use starting point.

### Example 1

I've seen colleagues see this many times. They don't understand what goes on and how to solve this because they are not used to seeing tables overlap in other web pages they visit. They are sort of stuck. And as a beginner you don't know which part of the documentation to find the solution in if any.

Here I would recommend the `pn.pane.DataFrame` to automatically figure out if horizontal or vertical scroll is needed. Even better I would like Panel to have `sizing_mode='stretch_width` by default. I think its a general design mistake to have fixed width by default. I would really, really recommend changing this default for Panel 1.0.

image

You can solve the horizontal issue by adding `width=800` or `sizing_mode='stretch_width'`. But it is a complication that other Frameworks don't have as they just make things responsive by default.

### Example 2

If you set the `width` and `height` in the notebook a vertical scrollbar appears. That is nice and the behaviour I would have hoped to experience by default.

image

But the problem is that the scrollbar is not added by Panel. Its added by Jupyter. Thus if you start using the DataFrame pane in a Column you start having the overflow problem again.

image

### Example 3

Ok so you can try to add scroll to the column. But only both which leads to the below problem. Where you can see the whole table width.

image

### Example 4

Ok so you can make the Column Wider

image

But then you have the ugly not needed indication of a horizontal scrollbar. And you had to add a 2x`width`, 1x`heigth` and 1x`scroll` parameters to get to something that at least is functional.

### Example 5

There is no help in the `pn.pane.DataFrame` guide as it only contains example with a few rows and columns. And even that is not given a proper width.

image

### Example 6

Just for fun I tried setting the height of the `Column` and something is clearly wrong.

image

### Recommended Actions

- Make `pn.panel` and `pn.pane.DataFrame` able to display larger dataframes nicely by default. Something has lots of columns and rows.
- Add parameter to `pn.pane.DataFrame` to control the scroll: None, Automatic, Horizontal, Vertical, Both.
- Add an example to the `pn.pane.DataFrame` demonstrating that it now works.

As an alternative you could consider dropping the `pn.pane.DataFrame` and use `pn.widgets.Tabulator`. Its much, much more useful.

It would be an added bonus if the number formatting was changed to something that works in a business. I almost never see scientific notation used.

MarcSkovMadsen:
Closes #4426

### DIFF
--- a/doc/_static/css/dataframe.css
+++ b/doc/_static/css/dataframe.css
@@ -6,7 +6,7 @@ table.panel-df {
     border-spacing: 0;
     color: black;
     font-size: 12px;
-    table-layout: fixed;
+    table-layout: auto;
     width: 100%;
 }
 
--- a/examples/reference/panes/DataFrame.ipynb
+++ b/examples/reference/panes/DataFrame.ipynb
@@ -16,7 +16,9 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "The ``DataFrame`` pane renders pandas, dask and streamz ``DataFrame`` and ``Series`` types as an HTML table. If you need to edit the values of a `DataFrame` use the `DataFrame` widget instead. The Pane supports all the arguments to the `DataFrame.to_html` function. \n",
+    "The ``DataFrame`` pane renders pandas, dask and streamz ``DataFrame`` and ``Series`` types as an HTML table. The Pane supports all the arguments to the `DataFrame.to_html` function. \n",
+    "\n",
+    "If you need to edit the values of a `DataFrame` use the `Tabulator` or `DataFrame` widgets instead.\n",
     "\n",
     "#### Parameters:\n",
     "\n",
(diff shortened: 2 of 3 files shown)


MarcSkovMadsen:
Damn. Did not start from latest `main`. Everything has changed :-)

MarcSkovMadsen:
Close this one to start from a fresh `main` branch.

PR #4326, opened 2023-01-26
Provide ability to set design system independently from template

philippjfr:
This PR introduces the concept of a `Design` which can be configured directly on individual components, globally on the `pn.config` and via a template (i.e. the MaterialTemplate uses the `Material` `Design` object).

This allows applying design system to our components independently of templates, e.g. also in notebooks. Here is an overview of the design of different component types in a notebook. This is very much a WIP and I'm not aiming to get everything looking perfectly in this PR. I'm mostly trying to ensure that the different design systems we offer render at all and further tweak specific CSS in subsequent PRs.

Below are the three design systems displayed in JupyterLab (using the default theme and adapting to the JupyterLab theme). Note that any properties set in Python, e.g. the Bokeh theme, do not automatically adapt to the JupyterLab dark theme.

## Bootstrap

Screen Shot 2023-02-14 at 19 25 56

Screen Shot 2023-02-14 at 23 24 00

Screen Shot 2023-02-14 at 19 26 05

Screen Shot 2023-02-15 at 13 02 41

Screen Shot 2023-02-14 at 19 26 13

Screen Shot 2023-02-15 at 13 02 50

## Material

Screen Shot 2023-02-14 at 19 28 03

Screen Shot 2023-02-15 at 13 01 14

Screen Shot 2023-02-14 at 19 29 31

Screen Shot 2023-02-15 at 13 01 36

Screen Shot 2023-02-14 at 19 28 35

Screen Shot 2023-02-15 at 13 02 10

## Fast

Screen Shot 2023-02-15 at 12 54 22

Screen Shot 2023-02-15 at 12 28 45

Screen Shot 2023-02-15 at 12 54 46

Screen Shot 2023-02-15 at 12 55 22

Screen Shot 2023-02-15 at 12 55 00

Screen Shot 2023-02-15 at 12 55 42

PR #4443, opened 2023-02-11
Fix fast glitch

### LINKED ISSUE #4442: Fast template header color is green and background white on startup
I'm on the current `main` branch of Panel. One thing that has been annoying me about the FastTemplates is that if you network is slow, your template is big or some javascript goes wrong you end up with an app where the header background is white and the color is green. At least for a split of a second.

This looks very unprofessional. I will try to fix.

You can see it in the upper right corner in the video below

https://user-images.githubusercontent.com/42288570/218275833-84184f42-c52e-469a-96ea-2813dea5842e.mp4

```python
import panel as pn
import pandas as pd

pn.extension(template="fast")

data = pd.DataFrame({"x": [1]*1000, "y": [2]*1000})

pn.panel(data).servable()
```

MarcSkovMadsen:
Closes #4442

### DIFF
--- a/panel/template/fast/grid/fast_grid_template.html
+++ b/panel/template/fast/grid/fast_grid_template.html
@@ -54,6 +54,8 @@
   }
   #header-design-provider {
     --neutral-foreground-rest: {{ style.header_color }};
+    background: {{ style.header_background }};
+    color: {{ style.header_color }};
   }
   #sidebar {
     width: {{ sidebar_width }}px;
--- a/panel/template/fast/theme.py
+++ b/panel/template/fast/theme.py
@@ -40,7 +40,7 @@ class FastStyle(param.Parameterized):
     accent_base_color = param.String(default="#0072B5")
     collapsed_icon = param.String(default=COLLAPSED_SVG_ICON)
     expanded_icon = param.String(default=EXPANDED_SVG_ICON)
-    color = param.String(default="#00aa41")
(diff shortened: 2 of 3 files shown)


panel/template/fast/theme.py
```diff
@@ -40,7 +40,7 @@ class FastStyle(param.Parameterized):
     accent_base_color = param.String(default="#0072B5")
     collapsed_icon = param.String(default=COLLAPSED_SVG_ICON)
     expanded_icon = param.String(default=EXPANDED_SVG_ICON)
-    color = param.String(default="#00aa41")
+    color = param.String(default="#2B2B2B")
```
MarcSkovMadsen, line 43: This is where the grenn header color was coming from.

panel/template/fast/grid/fast_grid_template.html
```diff
@@ -54,6 +54,8 @@
   }
   #header-design-provider {
     --neutral-foreground-rest: {{ style.header_color }};
+    background: {{ style.header_background }};
```
MarcSkovMadsen, line 57: If we set these, we will see the header_background and header_color immediately. And not after all the `fast` web components have been initialized.

philippjfr:
I'm trying to upgrade to latest fast-components version so will likely fix this separately but suspect these changes will still be needed.

philippjfr:
Applied these fixes in https://github.com/holoviz/panel/pull/4326

PR #4448, opened 2023-02-15
Update LaTeX models to support Bokeh 3

Hoxbro:
- [x] Work with Katex
- [x] Work(ish) with MathJax

If CSS is needed for the MathJax model, I will rename the CSS file to equation.css.

``` python
import panel as pn

pn.extension("katex", "mathjax")
msg = r"The LaTeX pane supports two delimiters: $LaTeX$ and \(LaTeX\)"

latex = pn.pane.LaTeX(msg).servable()
latex = pn.pane.LaTeX(msg, renderer="mathjax").servable()
```

image


Hoxbro:
MathJax also works now:

image

`\(\)` was fixed by  changes @philippjfr made over the last couple of days. 

I couldn't get `$` to work in TS as the conversion happens [here in Bokeh](https://github.com/bokeh/bokeh/blob/9f00012c81ac0410562a3aaf5d04ccb92b3e8fc9/bokehjs/src/lib/models/text/mathjax/index.ts) and could not figure out how to [configure](https://docs.mathjax.org/en/latest/web/configuration.html#configuring-mathjax) the inline display. So I convert the `$` to the method which works. 

@mattpap, have I missed something?

philippjfr (approved):
Happy to merge this for now and find a way to configure the valid escape sequences later.

mattpap:
I don't remember why single `$` doesn't work, but here are the delimiters that bokehjs supports:
```ts
const delimiters: Delimiter[] = [
  {start: "$$", end: "$$", inline: false},
  {start: "\\[", end: "\\]", inline: false},
  {start: "\\(", end: "\\)", inline: true},
]
```
As defined [in](https://github.com/bokeh/bokeh/blob/78dc3e238194302b5daadf11d47bdd69bd07a2b8/bokehjs/src/lib/models/text/utils.ts#L11-L15).

Issue #4430, opened 2023-02-09
The x-axis is cut off in Bokeh 3 if no label is present


Hoxbro:
The x-axis is cut off in Bokeh 3 if no label is present:

``` python
import holoviews as hv
hv.extension("bokeh")
hv.Curve([]).opts(xlabel="")
```

![image](https://user-images.githubusercontent.com/19758978/217808246-8dd25cf7-8895-4470-8f2a-458ace6eb6c7.png)

This could also be a Panel issue.

PR #4463, opened 2023-02-21
Ensure Pane layout takes margin into account

philippjfr:
Fixes https://github.com/holoviz/panel/issues/4430


philippjfr:
pre-commit.ci autofix

PR #4467, opened 2023-02-22
Update TextEditor for Bokeh 3

Hoxbro:
This tries to fix two things, but to be honest, I'm not sure about the implementation for any of them. And more putting it up here to have something to talk out from.

``` python
import panel as pn

pn.extension("texteditor")

pn.widgets.TextEditor(value="Highlight me to edit formatting").servable()
```

With the change in patch_stylesheet, my output looks like this:
image

<br>
Without the selection addition:

https://user-images.githubusercontent.com/19758978/220684150-9f69420f-3aa8-47b8-abdb-59bad1052713.mp4

<br>
With the selection addition:

https://user-images.githubusercontent.com/19758978/220683337-e3790878-c6e1-4603-ba31-fe238e9aeef0.mp4

<br>

[Reference for the selection](https://stackoverflow.com/questions/67914657/quill-editor-inside-shadow-dom)

### DIFF
--- a/panel/io/resources.py
+++ b/panel/io/resources.py
@@ -183,7 +183,11 @@ def loading_css():
     }}
     """
 
-def patch_stylesheet(stylesheet, dist_url):
+def patch_stylesheet(stylesheet, dist_url) -> None:
+    global RESOURCE_MODE
+    if RESOURCE_MODE == "server":
+        return
+
     url = stylesheet.url
     if not url.startswith('http') and not url.startswith(dist_url):
         patched_url = f'{dist_url}{url}'
--- a/panel/io/resources.py
+++ b/panel/io/resources.py
@@ -183,11 +183,7 @@ def loading_css():
     }}
     """
 
-def patch_stylesheet(stylesheet, dist_url) -> None:
-    global RESOURCE_MODE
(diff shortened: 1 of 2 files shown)


panel/io/resources.py
```diff
@@ -183,7 +183,11 @@ def loading_css():
     }}
     """
 
-def patch_stylesheet(stylesheet, dist_url):
+def patch_stylesheet(stylesheet, dist_url) -> None:
+    global RESOURCE_MODE
+    if RESOURCE_MODE == "server":
+        return
```
philippjfr, line 189: After merging latest changes from main this should not be needed.

Hoxbro: You are right. I have reverted the change.

philippjfr:
Sadly it does indeed look like Quill does not yet fully support the shadow DOM, here's a more complete polyfill I found in the Quill repo: https://github.com/quilljs/quill/issues/2021#issuecomment-864096352

philippjfr:
Similar polyfill (would have to test if this works in recent Safari version): https://stackoverflow.com/questions/67914657/quill-editor-inside-shadow-dom/67944380#67944380

PR #4479, opened 2023-02-28
Bump bokeh to 3.1.0dev4

philippjfr:


panel/models/html.ts
```diff
@@ -5,8 +5,10 @@ import {Markup} from "@bokehjs/models/widgets/markup"
 import {PanelMarkupView} from "./layout";
 import {serializeEvent} from "./event-to-object";
 
+
 export class DOMEvent extends ModelEvent {
   event_name: string = "dom_event"
+  publish: boolean = true
```
mattpap, line 11: Both should be `prototype` members, not instance members, though that also works. In bokehjs we use a decorator to make sure this is set up properly:
```ts
import {event} from "@bokehjs/core/bokeh_events"

@event("dom_event")
export class DOMEvent extends ModelEvent {
  ...
}
```
Alternatively one can use this:
```ts
export class DOMEvent extends ModelEvent {
  static {
    this.prototype.event_name = "dom_event"
    this.prototype.publish = true
  }
}
```
I will make a change, so that `publish = true` is actually the default, regardless how a event is defined.

philippjfr: Will fix this in another PR.

mattpap: Fixed in https://github.com/bokeh/bokeh/pull/12871.

PR #4481, opened 2023-03-01
Small bug fixes 

Hoxbro:
This fixes two things:
- Make `nan=nan` comparison to True, as I don't see a reason why it shouldn't be for our case.
- Use cache from previous sessions when using `to_disk`


philippjfr (approved):
Looks good!

PR #4482, opened 2023-03-01
Update Tabulator to 5.4 and optimize rendering

philippjfr:
- Upgrades Tabulator.js to 5.4.4
- Avoids unnecessary redraw and re-render cycles 
- Fixes resource handling on initial rendering


panel/models/tabulator.ts
```diff
@@ -374,38 +386,35 @@ export class DataTabulatorView extends HTMLBoxView {
...
     super.after_layout()
     if (this.tabulator != null && this._initializing)
       this.redraw()
+    this._initializing = false
   }
 
   render(): void {
     super.render()
     this._initializing = true
     const container = div({style: "display: contents;"})
```
Hoxbro, line 410:
```suggestion
    const container = div({style: "display: contents"})
```
Very nit, but for consistency with changes below.

panel/tests/ui/widgets/test_tabulator.py
```diff
@@ -1041,10 +1041,10 @@ def test_tabulator_frozen_rows(page, port):
     expected_text = """
     index
     col
-    8
-    Y
     1
     X
+    8
+    Y
```
Hoxbro, line 1047: Why is this changed?

panel/tests/ui/widgets/test_tabulator.py
```diff
@@ -1071,9 +1071,6 @@ def test_tabulator_frozen_rows(page, port):
     X_bb = page.locator('text="X"').bounding_box()
     Y_bb = page.locator('text="Y"').bounding_box()
 
-    # Check that the Y row is rendered before the X column
-    assert Y_bb['y'] < X_bb['y']
```
Hoxbro, line 1075: Why is this removed?

panel/models/tabulator.ts
```diff
@@ -478,19 +482,14 @@ export class DataTabulatorView extends HTMLBoxView {
...
     this._building = false
-    this.setHidden()
     this.setSelection()
-
-    // For remote pagination initialize on tableBuilt
-    this.setMaxPage()
-    this.tabulator.setPage(this.model.page)
-    this.setStyles()
     this.renderChildren()
-    this.setGroupBy()
```
Hoxbro, line 489: Why is this `this.setGroupBy` removed?

panel/models/tabulator.ts
```diff
@@ -478,19 +482,14 @@ export class DataTabulatorView extends HTMLBoxView {
 
   tableBuilt(): void {
     this._building = false
-    this.setHidden()
```
Hoxbro, line 481: Is this removal related to the change further down (`columns.push({field: '_index', frozen: true, visible: false})`)?

philippjfr: Yes, the aim is to configure as much as possible in the table configuration rather than setting the options after initialization.

panel/models/tabulator.ts
philippjfr: Same as setHidden, it's better to configure this in the configuration rather than setting it after rendering.

panel/models/tabulator.ts
Hoxbro: The one right below:
`const el = div({class: "pnx-tabulator", style: "width: 100%; height: 100%"})`

panel/tests/ui/widgets/test_tabulator.py
philippjfr: I've decided to change the behavior, a) because it's much simpler and b) frozen rows should respect the original row order. While I could see an argument that the order in which the frozen row were defined should determine their ordering at the top of the table the actual implementation only did that correctly in the simplest case. So I've decided to switch to a consistent and simpler implementation.

panel/models/tabulator.ts
philippjfr: Ah, I'd probably rather change that one.

panel/tests/ui/widgets/test_tabulator.py
philippjfr: Will record this change in the CHANGELOG.

Issue #4472, opened 2023-02-24
`panel examples` does not find examples


domoritz:
I have a conda environment (and installed conda with miniforge) and installed panel with `conda install -c pyviz panel`. When I run `panel examples`, I get this error

I am on macOS 13.3. conda 22.11.1. Python 3.11.0. 

```
Traceback (most recent call last):
  File "/opt/homebrew/Caskroom/miniforge/base/envs/ramsch/bin/panel", line 10, in <module>
    sys.exit(main())
             ^^^^^^
  File "/opt/homebrew/Caskroom/miniforge/base/envs/ramsch/lib/python3.11/site-packages/panel/command/__init__.py", line 112, in main
    pyct.cmd.substitute_main('panel', cmds=pyct_commands, args=args)
  File "/opt/homebrew/Caskroom/miniforge/base/envs/ramsch/lib/python3.11/site-packages/pyct/cmd.py", line 455, in substitute_main
    args.func(args) if hasattr(args,'func') else parser.error("must supply command to run")
    ^^^^^^^^^^^^^^^
  File "/opt/homebrew/Caskroom/miniforge/base/envs/ramsch/lib/python3.11/site-packages/pyct/cmd.py", line 394, in <lambda>
    parser.set_defaults(func=lambda args: fn(name, **{k: getattr(args,k) for k in vars(args) if k!='func'} ))
                                          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/homebrew/Caskroom/miniforge/base/envs/ramsch/lib/python3.11/site-packages/pyct/cmd.py", line 38, in examples
    copy_examples(name, path, verbose, force)
  File "/opt/homebrew/Caskroom/miniforge/base/envs/ramsch/lib/python3.11/site-packages/pyct/cmd.py", line 44, in copy_examples
    source = _find_examples(name)
             ^^^^^^^^^^^^^^^^^^^^
  File "/opt/homebrew/Caskroom/miniforge/base/envs/ramsch/lib/python3.11/site-packages/pyct/cmd.py", line 28, in _find_examples
    raise ValueError("Could not find examples for %s at any of %s"%(name,candidates))
[...]

PR #3481, opened 2022-05-02
Add support for converting Panel apps to pyscript/pyodide

philippjfr:
Implements conversion of a standard Panel app (in a .py or .ipynb file) to a standalone PyScript/Pyodide application.

Depends on https://github.com/pyscript/pyscript/pull/130

- [x] Add pure pyodide support
- [x] Add pyodide in a web worker support 
- [x] Add CLI

### Example

```bash
panel convert ./examples/gallery/simple/random_number_generator.ipynb --to pyscript
```

Output:

panel_auto

PR #4484, opened 2023-03-01
Add `examples` folder dynamically added to the distrib

maximlt:
Fixes https://github.com/holoviz/panel/issues/4472

Grafting `panel/examples` was done before but removed in https://github.com/holoviz/panel/pull/3481/commits/79e05e630c61c099639ebec42365338f922343f9, not sure why.

I think adding it back is the right change to fix the reported issue.

The way the examples are distributed is a bit confusing (to me at least!) and I've started to track that here https://github.com/holoviz/holoviews/pull/5632. Understanding that is necessary anyway for us to modernize the build process and hopefully move to `pyproject.toml`.

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -27,6 +27,7 @@ global-exclude *.py[co]
 global-exclude *~
 global-exclude *.ipynb_checkpoints/*
 global-exclude *.whl
+graft panel/examples
 graft examples
 graft doc/
 graft panel/dist


philippjfr:
Indeed, this has been very confusing. I'm in favor of restoring this and then figuring out what to do when we switch to a new build system.

philippjfr:
Will merge, failing tests have been resolved on main.

PR #4493, opened 2023-03-03
Use dev package for holoviews install

Hoxbro:
After the dev release of Panel, it should be possible to install Holoviews directly with conda.


philippjfr (commented):
Looks good, thanks!

philippjfr (approved):

PR #4495, opened 2023-03-03
Implement support for passing parameters by reference

philippjfr:
Supersedes https://github.com/holoviz/panel/pull/4042


philippjfr:
I'm going ahead and merging this to be able to evaluate it as I re-write various examples and applications. How-tos around reactivity and parameter binding are already lacking so I'm going to add as many of those in one go as I can.

PR #4499, opened 2023-03-03
Bump bokeh version to 3.1.0dev5

philippjfr:
Minor changes to adapt to the `styles` method being renamed and correctly defining the `event_name`.

### DIFF
--- a/panel/models/html.ts
+++ b/panel/models/html.ts
@@ -7,15 +7,16 @@ import {serializeEvent} from "./event-to-object";
 
 
 export class DOMEvent extends ModelEvent {
-  event_name: string = "dom_event"
-  publish: boolean = true
-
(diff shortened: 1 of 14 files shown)


mattpap:
Looks good to me.

panel/models/html.ts
```diff
@@ -17,6 +17,7 @@ export class DOMEvent extends ModelEvent {
 
   static {
     this.prototype.event_name = "dom_event"
+    this.prototype.publish = true
```
mattpap, line 20: Setting `publish` shouldn't be needed since https://github.com/bokeh/bokeh/pull/12871.

philippjfr: Yep, I had just made a copy-paste error in defining one of the event names making me think the publish default wasn't working.

PR #4510, opened 2023-03-07
Update asyncio/Tornado 

Hoxbro:
`make_current` deprecation was reverted in Python and will not give a warning in Tornado 6.3 and forward (https://github.com/tornadoweb/tornado/issues/3216).

The change in `conftest.py` is to avoid this warning, though I'm unsure if it is the right way. 
``` python
  /home/shh/miniconda3/envs/holoviz/lib/python3.10/site-packages/tornado/ioloop.py:265: DeprecationWarning: There is no current event loop
    loop = asyncio.get_event_loop()
```

### DIFF
--- a/panel/io/server.py
+++ b/panel/io/server.py
@@ -919,7 +919,7 @@ def get_server(
 
     opts = dict(kwargs)
     if loop:
-        loop.make_current()
+        asyncio.set_event_loop(loop.asyncio_loop)
         opts['io_loop'] = loop
     elif opts.get('num_procs', 1) == 1:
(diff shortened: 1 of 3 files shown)


philippjfr:
So I'm not 100% convinced we should actually make this change. Each thread should get its own event loop so it makes sense to me that we should use `make_current=False` in the IOLoop constructor and then have the thread start the event loop. Then again my mental model of this may not be 100% correct.

philippjfr:
For the time being I'm going to keep the current approach. It seems more correct to me for the thread to start the event loop.

PR #4531, opened 2023-03-14
Update homepage and remove bokeh

Hoxbro:
- Remove some Bokeh workaround.
- Add python to a codeblock
- Add tags back to the homepage.ipynb


philippjfr (approved):
Thanks!

PR #4533, opened 2023-03-14
Add BrowserInfo model to expose browser window and navigator APIs

philippjfr:
Provides various pieces information about the browser.

- `dark_mode`: Whether the user prefers dark mode.
- `device_pixel_ratio`: Provides the ratio of the resolution in physical pixels to the resolution in CSS pixels for the current display device.
- `language`: The preferred language of the user, usually the language of the browser UI.
- `timezone`: The timezone configured as the local timezone of the user.
- `timezone_offset`: The time offset from UTC in minutes.
- `webdriver`: Indicates whether the user agent is controlled by automation.

### DIFF
--- a/panel/config.py
+++ b/panel/config.py
@@ -16,6 +16,7 @@ from weakref import WeakKeyDictionary
 
 import param
 
+from bokeh.document import Document
 from bokeh.settings import settings as bk_settings
 from pyviz_comms import (
     JupyterCommManager as _JupyterCommManager, extension as _pyviz_extension,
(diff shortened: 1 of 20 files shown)


MarcSkovMadsen:
Just for info. Dash just added a Geolocation component.

panel/config.py
```diff
@@ -115,6 +116,9 @@ class _config(_base_config):
     autoreload = param.Boolean(default=False, doc="""
         Whether to autoreload server when script changes.""")
 
+    browser_info = param.Boolean(default=True, doc="""
```
Hoxbro, line 119: I think it would be better to be off by default.

philippjfr: Can see that argument but can you explain your reasoning? One of the main reasons we want to use it is for things like the device pixel ratio to determine the resolution of datashader plots and hardly anyone will know to enable this.

Hoxbro: The reason for the suggestion was that there was no reason in this PR why it should be on by default. 

You gave me two reasons.

PR #4547, opened 2023-03-21
Remove pkg_resources

### LINKED ISSUE #3282: Replace `pkg_resources` with `importlib_metadata`
Currently `pkg_resources` is used in a few places (in particular `iter_entry_points`). Would be good to replace this with [`importlib_metadata`]( https://github.com/python/importlib_metadata ) ([`importlib.metadata`]( https://docs.python.org/3/library/importlib.metadata.html ) added in Python 3.8 & stable in Python 3.10), which supersedes `pkg_resources`. In particular usages of `iter_entry_points` can be replaced by [`entry_points`]( https://docs.python.org/3.8/library/importlib.metadata.html#entry-points ).

This also would allow `setuptools` to be dropped as a runtime requirement.

Hoxbro:
Fixes #3282 

`pkg_resources.parse_requirements` was replaced with `from packaging.requirements import Requirement` and was tested on this:
``` python
requirements = """
panel  # this is panel
 #2
datashader >=0.11.1; python_version < '3.11'
  
"""

import pkg_resources
parsed = pkg_resources.parse_requirements(requirements)
[str(requirement) for requirement in parsed]

from packaging.requirements import Requirement
[r2 for r in requirements.split("\n") if (r2 := r.split("#")[0].strip()) and Requirement(r2)]

```

image

### DIFF
--- a/panel/auth.py
+++ b/panel/auth.py
@@ -7,7 +7,6 @@ import re
 import urllib.parse as urlparse
 import uuid
 
-from importlib.metadata import entry_points
 from typing import List, Tuple, Type
 
 import tornado
@@ -19,6 +18,7 @@ from tornado.httputil import url_concat
 from tornado.web import RequestHandler
 
 from .config import config
+from .entry_points import entry_points_for
 from .io import state
 from .io.resources import ERROR_TEMPLATE, _env
 from .util import base64url_decode, base64url_encode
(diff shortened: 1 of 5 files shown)


philippjfr:
This looks significantly better, thanks!

philippjfr (approved):
Looks good, thanks!

PR #4551, opened 2023-03-21
Add Image pane reference

Hoxbro:
Adds a references pane for the base Image class.

### DIFF
--- a/examples/reference/panes/Image.ipynb
+++ b/examples/reference/panes/Image.ipynb
@@ -103,7 +103,7 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "Note that by default the aspect ratio of the image is fixed, to override this behavior set `fixed_aspect=False` or provide fixed `width` and `height` values."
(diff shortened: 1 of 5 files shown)


examples/reference/panes/Image.ipynb
```diff
@@ -0,0 +1,118 @@
...
+    "pn.pane.Image(\n",
+    "    'https://assets.holoviz.org/panel/samples/png_sample2.png', sizing_mode='scale_width'\n",
+    ")"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "Note that by default the aspect ratio of the image is fixed, to override this behavior set `fixed_aspect=False` or provide fixed `width` and `height` values."
```
jbednar, line 106:
```suggestion
    "Note that by default the aspect ratio of the image is fixed, and so there may be a gap beside or below the image even in responsive sizing modes. To override this behavior set `fixed_aspect=False` or provide fixed `width` and `height` values."
```

PR #4555, opened 2023-03-24
Describe on_session_destroyed callback signature

MarcSkovMadsen:
Victor was struggling with the callback signature in https://discourse.holoviz.org/t/how-to-use-pn-state-on-session-destroyed/5121. It was not clear to him the callback must have a `session_context` argument.

I add a description of this to the docs.


MarcSkovMadsen:
Fails on macos. But I assume this is not related to this change.

philippjfr:
Thanks @MarcSkovMadsen!

PR #4569, opened 2023-03-29
Add functional size to pn.widgets.Select for size > 1

### LINKED ISSUE #4568: pn.widgets.Select does not work with size.
``` python
import panel as pn

pn.extension()

size = 10
pn.widgets.Select(options=list(range(size)), size=size)
```

image

This seems to be because `pn.widgets.Select._rename` is not sent to the Bokeh model.
image

I'm unsure if the fix is to remove `size` from `_rename` or if I'm overlooking something.

Hoxbro:
Fixes #4568


philippjfr (approved):
Looks good!

PR #4595, opened 2023-04-04
Add automation to deploy gallery

philippjfr:
Automatically re-uploads the gallery to `pyviz.demo.anaconda.com` and starts a deployment.


Hoxbro:
I think this will redeploy every weekend when the docs test is run. 

I don't know if you want this, but I want to mention it.

Did you consider having a separate yaml file for this?

philippjfr:
Tbh I want to redeploy nightly but yes I will probably end up making this a new file.

PR #4602, opened 2023-04-05
Add support for writing applications in Markdown

philippjfr:
- [x] Add ability to `panel serve`, `pn.serve()` and `panel convert` .md files
- [x] Add tests
- [x] Add docs


MarcSkovMadsen:
If you have time please test that a panel app .md file renders nicely on GitHub.

Then you have an easy way deploying documentation.

Please consider whether the python tag could/ should be supported. Maybe it's better than the Panel tag. The Python tag is recognized and will render nicely on GitHub and in .md based docs or blog posts.

philippjfr:
>If you have time please test that a panel app .md file renders nicely on GitHub.

Works fine, see https://github.com/holoviz/panel/blob/5593ff79585ea4fc1ef6e29d1611fe677e950c5b/doc/how_to/display/examples/hello_world.md

>Please consider whether the python tag could/ should be supported. Maybe it's better than the Panel tag. The Python tag is recognized and will render nicely on GitHub and in .md based docs or blog posts.

>IDEs like Pycharm or VS code might also be able to provide help inside the code section if the Python tag is used?

These are good points and originally I had support for python tags but my concern was that it made it impossible to actually render Python code inline as Markdown.

philippjfr:
Just discovered that you can use the file extension as a language tag, e.g. `.py` works. So I'm tempted to change it.

PR #4614, opened 2023-04-09
Fix VTK rendering

philippjfr:
Resolves serialization and rendering issues with bokeh 3.x


philippjfr:
@xavArtley If you have a minute to look at this that would be great. I tried to update to latest vtk.js but encountered some issues with the volume renderer (something about the zBuffer not being defined deep inside of the rendering pass of vtk.js). Otherwise I updated some of the property definitions and serialization, the main change being that we no longer base64 encode the binary data but we also no longer zip the bytes so that could be a bit of a performance regression (I did this because latest vtk no longer ships with decompression utilities).

Hoxbro:
Should we put a lower pin on VTK corresponding to the version you have used to update the VTK model? 

What about pyvista?  That seems to be related to the VTK. Does that work with these changes?  Right now it has an upper pin. Can this be removed?
https://github.com/holoviz/panel/blob/e179a2d043c97491863d9581767a65a959edd01e/setup.py#L176

The numpy upper pin should also be removed (or the comment should be updated) 
https://github.com/holoviz/panel/blob/e179a2d043c97491863d9581767a65a959edd01e/setup.py#L154

philippjfr:
Yes, I'll remove the pins, in fact they seem to be breaking the build.

xavArtley:
Hello, I tried to update the vtk.js version a while ago, but I struggled with the zBuffer and abandoned it. I didn't have much bandwidth with one child, but now I have two of them, so I'm sorry I won't be able to dig into this issue.

PR #4619, opened 2023-04-10
Various documentation improvements before 1.0b10

### LINKED ISSUE #4607: Add ability to open custom notebook in Panelite.
The Jupyter lite extension https://github.com/jupyterlab-contrib/jupyterlab-open-url-parameter enables you to open a notebook from an url in Jupyterlite.

This would be really powerful to include in the build of Panelite as we can the start to share links to notebooks that opens quickly for the user with a working environment.

philippjfr:
- Add jupyterlite-open-url extension (fixes #4607)

### DIFF
--- a/lite/requirements.txt
+++ b/lite/requirements.txt
@@ -7,3 +7,4 @@ jupyterlite
 nbformat
 pkginfo
 pyviz_comms
+jupyterlite-open-url-parameter
--- a/.github/workflows/docs.yaml
+++ b/.github/workflows/docs.yaml
@@ -173,12 +173,11 @@ jobs:
           conda activate test
           conda install -c ae5-admin anaconda-project ae5-tools
       - name: Upload and deploy gallery
-        continue-on-error: true
         run: |
           conda activate test
           python scripts/collect_thumbnails.py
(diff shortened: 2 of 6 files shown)


lite/requirements.txt
```diff
@@ -7,3 +7,4 @@ jupyterlite
 nbformat
 pkginfo
 pyviz_comms
+jupyterlite-open-url-parameter
```
jtpio, line 10: Looks like this should be `jupyterlab-open-url-parameter` here.

philippjfr: Indeed, good catch thanks.

Hoxbro: I assume you followed this: https://github.com/jupyterlite/jupyterlite/blob/76cd0496add7494bec9a749b2044e28074f60781/docs/howto/content/open-url-parameter.md.

@jtpio Does that mean that the document should be updated?

jtpio: Probably this link points to an old version of the docs?

The docs look good here: https://jupyterlite.readthedocs.io/en/latest/howto/content/open-url-parameter.html

Issue #4617, opened 2023-04-10
Tooltip symbol shown when no description/ tooltip provided


MarcSkovMadsen:
I was working on the current `master` branch. I noticed that the new tooltip icon is shown by default and even though no description/ tooltip has been provided.

I believe this provides a poor user experience. It must be a bug.

A user would expect some info to be available. Try to click and hover the symbol, but nothing happens. This will confuse the user.

Please don't show tooltip symbol unless it provides a tooltip.

```python
import panel as pn

pn.extension()

pn.Column(
    pn.widgets.TextInput(name='Text')
).servable()
```

![image](https://user-images.githubusercontent.com/42288570/230959082-eea4687e-b514-491d-aeee-e128ecf75008.png)

PR #4621, opened 2023-04-11
Set default for description to None

Hoxbro:
Fixes https://github.com/holoviz/panel/issues/4617

image


philippjfr:
Great, thanks @Hoxbro. Could you maybe create an issue with a table that summarizes which widget do and do not support tooltips and link it with the equivalent issue for widget titles (which I am having trouble finding right now)?

philippjfr (approved):
Thanks @Hoxbro, looks good.

Issue #4406, opened 2023-02-05
panel/widgets/tables.py:790: SettingWithCopyWarning: 


MarcSkovMadsen:
Panelite=0.14.3

I'm trying the notebook reference/widgets/Tabulator.ipynb. In the *Patching* section I get the `SettingWithCopyWarning`.

![image](https://user-images.githubusercontent.com/42288570/216839970-81efece9-4df0-4251-8091-2b36f296fc86.png)

```bash
/lib/python3.10/site-packages/panel/widgets/tables.py:790: SettingWithCopyWarning: 
A value is trying to be set on a copy of a slice from a DataFrame

See the caveats in the documentation: https://pandas.pydata.org/pandas-docs/stable/user_guide/indexing.html#returning-a-view-versus-a-copy
  self.value.loc[ind, k] = value
```

PR #4414, opened 2023-02-07
Update Tabulator to fix some Pandas warnings

maximlt:
Use `.loc` to set values and call `.copy()` on the filtered dataframe to avoid passing around views.

PR #4623, opened 2023-04-11
Reduce number of Pandas warnings emitted when interacting with the Tabulator widget

maximlt:
Supersedes https://github.com/holoviz/panel/pull/4414, in which I went down the route of creating copies of the filtered/sorted dataframes to avoid `SettingWithCopyWarning`. It happens that it actually matters that `_processed` is a view of `value`, for patching in particular which updates directly `value`, patches being then also applied to `_processed` as it's a view. So for now I just decided to silence these warnings.

Fixes https://github.com/holoviz/panel/issues/4406

### DIFF
--- a/examples/reference/widgets/Tabulator.ipynb
+++ b/examples/reference/widgets/Tabulator.ipynb
@@ -3,9 +3,7 @@
   {
    "cell_type": "code",
    "execution_count": null,
-   "metadata": {
-    "tags": []
-   },
+   "metadata": {},
    "outputs": [],
    "source": [
     "import datetime as dt\n",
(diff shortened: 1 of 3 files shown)


examples/reference/widgets/Tabulator.ipynb
```diff
@@ -3,7 +3,9 @@
   {
    "cell_type": "code",
    "execution_count": null,
-   "metadata": {},
+   "metadata": {
+    "tags": []
```
Hoxbro, line 7: Uff, these are annoying to look at. Did you run this in Jupyterlab?

maximlt: Oh yes indeed, that's not nice. I will clear that. I'm not sure how it got there, I opened it both in JLab and VSCode.

philippjfr (approved):
Thanks for the fixes, seems reasonable to me. Test failures are unrelated (I've got some fixes that will make them more robust again coming).

Issue #4460, opened 2023-02-21
Public API renaming


philippjfr:
In general we should have the principle that components should be named not after the implementation they use but what they do. Therefore I'd like to propose renaming of a few of our existing components:

- `Ace` -> `CodeEditor`
- `Tabulator` -> `Table`
- `ReactTemplate` -> `GridTemplate`
- `FastListTemplate` -> `FastTemplate`
- `VanillaTemplate` -> ???

@holoviz/panel-maintainers please chime in

maximlt:
Sorry for the lack of response here! That's a tricky one, probably because I've got used to these names.

But I do agree that a renaming is in order and with the general principle of object names not making any reference to the underlying implementation. I think the main reason for that is that Panel users are not expected to be front-end developers and they should not be exposed to that kind of stuff, they want a code editor, they use `CodeEditor`.

In practice though the objects leak details from the implementation, sometimes even intentionally with for instance the Tabulator doc reference linking users to tabulator.info to see what type of parameters they can pass. Maybe having a better and more general name will force us to define more general interfaces :)

> Ace -> CodeEditor

+1

> Tabulator -> Table

I wanted to ask, do you intend at some point that there will be a dataframe/table **pane** depending on a Tabulator implementation? If so, would that still be the `DataFrame` pane or would there be a new one? If there's a new one, it might be called `Table`, in which case the table widget might be better called `TableEditor` which would also align it with the `TextEditor`, `CodeEditor`, etc.

I find renaming the templates trickier. For the non-advanced Panel users, I would guess that the template names including Bootstrap, Material, Fast, etc. doesn't really bring them any value. They're certainly more interested in the functionalities/UX that a template offers. Advanced users will be more familiar with the UI frameworks, maybe they even have a requirement to use a particular UI framework to align with their internal corporate design, so they may be pretty happy to see Panel provide e.g. a Material-based template. Right now the template names are more tailored to that second class of user.

(That makes me think that a 'How to choose a template' could be useful.)

MarcSkovMadsen:
Its ok with me.

Hoxbro:
I feel like a generic name can be used if the underlying library is easily swappable. This seems to be the case for `Ace -> CodeEditor` but not `Tabulator -> Table`. Another thing is there _could_ be more table widgets like we see for plotting libraries, e.g., Bokeh, Plotly, Matplotlib, etc. 

[...]

PR #4627, opened 2023-04-11
Rename Ace to CodeEditor

philippjfr:
As discussed in https://github.com/holoviz/panel/issues/4460


panel/widgets/codeeditor.py
```diff
@@ -80,3 +81,5 @@ def _update_disabled(self, *events: param.parameterized.Event):
                 self.readonly = event.new
             elif event.name == 'readonly':
                 self.disabled = event.new
+
+Ace = CodeEditor
```
Hoxbro, line 85: Maybe give a deprecation warning here:

``` python
from panel.util.warning import deprecated

...

class Ace(CodeEditor):
    def __init__(self, **params):
        deprecated("1.1", "Ace", "CodeEditor")
        super().__init__(**params)
```

PR #4638, opened 2023-04-13
Various documentation fixes for 1.0.0b13

Hoxbro:


setup.py
```diff
@@ -173,7 +173,7 @@ def run(self):
         'jupyter_bokeh >=3.0.7',
         'django <4',
         'channels',
-        'pyvista <0.33',
+        'pyvista',
```
philippjfr, line 176: Readd pin to determine if it causes the tests to stall:

```suggestion
        'pyvista <0.33',
```

PR #4640, opened 2023-04-13
Add min_width to sync properties for Accordion

Hoxbro:
This PR adds `min_width` to be synced between for pn.Accordion.

A case could be made for having a default value for it. Maybe 300?

image

image

``` python
pn.Accordion(
    ("Test", "Hello"),
    ("Test2", 2 * "Hello"),
    min_width=300,
)
```

### DIFF
--- a/panel/layout/accordion.py
+++ b/panel/layout/accordion.py
@@ -72,7 +72,8 @@ class Accordion(NamedListPanel):
 
     _synced_properties = [
         'active_header_background', 'header_background', 'width',
-        'sizing_mode', 'width_policy', 'height_policy', 'header_color'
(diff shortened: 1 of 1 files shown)


panel/layout/accordion.py
```diff
@@ -72,7 +72,8 @@ class Accordion(NamedListPanel):
 
     _synced_properties = [
         'active_header_background', 'header_background', 'width',
-        'sizing_mode', 'width_policy', 'height_policy', 'header_color'
+        'sizing_mode', 'width_policy', 'height_policy', 'header_color',
+        'min_width',
```
philippjfr, line 76:
```suggestion
        'min_width', 'max_width'
 ```

PR #4643, opened 2023-04-14
Add description to more widgets

Hoxbro:
I observed during my investigation of which widget supports the new description in Bokeh 3, that some had the functionality but were not enabled on the Panel side.

(I will open an issue with my findings later today)


philippjfr (approved):
Perfect, thank you!

PR #4645, opened 2023-04-14
Make it clear that download of pyodide dependencies only happens once

MarcSkovMadsen:
It was not very clear to me that when clicking the Pyodide run button the download only happens once across the site. So I was a bit reluctant to do this - especially when using it on mobile without wifi.

Therefore I've tried to make it more clear that the download is cached.


philippjfr (approved):
Yeah, that's an good thing to point out!

PR #4649, opened 2023-04-15
use panelite word

MarcSkovMadsen:
I tried searching the `dev` docs for Panelite without luck. I wanted to know how to how to build it. But there was a miss

image

So I figured I would add/ use the Panelite word in the docs.


philippjfr (approved):
Looks good, thanks!

Issue #4651, opened 2023-04-15
Help the user understand the code a bit more in the introduction/ reference example


MarcSkovMadsen:
Looking at the `dev` docs the code provided in the intro example is a bit short

![image](https://user-images.githubusercontent.com/42288570/232183297-a1cfb8c1-4ece-4745-92f4-5a0371c21748.png)

I believe new users might not know what `pn` is.

Please add

```python
import panel as pn

pn.extension()
```

to the top of the code.

PR #4652, opened 2023-04-15
Improve intro example

MarcSkovMadsen:
Closes https://github.com/holoviz/panel/issues/4651 and fixes some bugs.


philippjfr (approved):
Yep, makes sense, thanks!

Issue #4611, opened 2023-04-09
Panel 1.0 Documentation Review


philippjfr:
# Documentation Review

- [x] Remove or fix warning about non-interactivity
- [x] Add JupyterLite links everywhere (https://github.com/pyviz-dev/nbsite/pull/270)
- [ ] Add links to deployed versions of how-to examples
- [x] Migrate non-featured gallery examples to reference and how-to materials
- [x] Rewrite gallery index page by hand
- [ ] Add links to pyodide converted apps
- [ ] Fix conversion of apps with global template

## Getting Started

- [x] Installation
  - [x] Move section about developing in different environments into How To (@droumis)
  - [x] Link to how-to sections but assume Jupyter(Lab)
- [ ] Get started in 2 minutes (@philippjfr)
   - Display Markdown + Chart or Table
   - Run in website, launch JupyterLite or locally
- [ ] Getting started with interactivity (@philippjfr)
   - [ ] Reactive based

## How-to

- [x] how_to/display
   - [x] JupyterLab Preview 
- [x] how_to/apis
  - [x] Find better way to guide people to correct API
    - [x] Function vs. Class
    - [x] Callback vs. Reactive
  - [x] Remove interact (and all mentions of `interact` from documentation)
- [x] how_to/reactive (@Hoxbro):
  - [x] Add section on pn.bind
  - [ ] Add section on binding parameters on components
  - [ ] Add section on ParamFunction/ParamMethod
- [ ] how_to/callbacks/session:
  - [ ] Missing on_session_created
- [ ] how_to/templates/
  - [ ] Update screenshots
- [ ] how_to/performance
  - [ ] Add section on on_load
  - [ ] Add section on defer load
  - [ ] 
- how_to/concurrency/load_balancing
  - [ ] Fix image
- [ ] how_to/integrations/flask
  - [ ] Missing
- [ ] how_to/template/template_theme
  - [ ] Remove

## Background

- [x] Organize into sections
- [x] Various items missing in TOC
- [ ] Panel and Param
  - [ ] Missing
- [ ] Panel Communications
  - [ ] Missing
- [ ] Templates
  - [ ] Update

## Reference Gallery

- [x] Ensure TOC tree has sections (https://github.com/pyviz-dev/nbsite/pull/267)
- [x] Make code pyodide executable (https://github.com/pyviz-dev/nbsite/pull/270)
- [x] Add Image and Vizzu pane thumbnails
- [ ] Review Template notebooks

## Discovered issues with components

[...]

PR #4664, opened 2023-04-17
Organize Background

droumis:
Organizing the background. Addresses some of #4611

background-org


droumis:
note, I'm using externally hosted images for the other library logos, as I don't think I have access to add to assets.holoviz.org

droumis:
update.. now using HoloViz-hosted assets

philippjfr (approved):
This looks great, thank you!!!

PR #4671, opened 2023-04-19
Documentation updates for 1.0.0b16

philippjfr:


panel/template/base.py
```diff
@@ -535,8 +535,11 @@ def __init__(self, **params):
             params['modal'] = ListLike()
         else:
             params['modal'] = self._get_params(params['modal'], self.param.modal.class_)
-        if 'theme' in params and isinstance(params['theme'], str):
-            params['theme'] = THEMES[params['theme']]
+        if 'theme' in params:
+            if isinstance(params['theme'], str):
+                params['theme'] = THEMES[params['theme']]
```
Hoxbro, line 540:
```suggestion
        if 'theme' in params and isinstance(params['theme'], str):
            params['theme'] = THEMES[params['theme']]
```

philippjfr: This breaks the else branch.

Issue #2575, opened 2021-07-26
Simple login/password auth example without oauth


andhuang-CLGX:
I want to use bokeh's auth module:

Directory structure
```
- test/
-     test/
-         main.py
-         auth.py
-         login.html
```

main.py
```
import panel as pn

pn.pane.Markdown("Login was successful!").servable()
```

auth.py: https://github.com/bokeh/bokeh/blob/branch-2.4/examples/howto/server_auth/auth.py

login.html: https://github.com/bokeh/bokeh/blob/branch-2.4/examples/howto/server_auth/login.html

`panel serve test/ --port 1234 --allow-websocket-origin="*" --auth-module=test/auth.py`

So if I login to `localhost:1234/login` and enter the proper username/password, it redirects fine.

But I can also bypass the login page by going to `localhost:1234/test`. I thought by passing --auth-module, panel would automatically redirect to the /login endpoint if `pn.state.cookies["user"]` did not exist.

Any ideas on how I can fix this?

orangeSi:
> I want to use bokeh's auth module:
> 
> Directory structure
> 
> ```
> - test/
> -     test/
> -         main.py
> -         auth.py
> -         login.html
> ```
> 
> main.py
> 
> ```
> import panel as pn
> 
> pn.pane.Markdown("Login was successful!").servable()
> ```
> 
> auth.py: https://github.com/bokeh/bokeh/blob/branch-2.4/examples/howto/server_auth/auth.py
> 
> login.html: https://github.com/bokeh/bokeh/blob/branch-2.4/examples/howto/server_auth/login.html
> 
> `panel serve test/ --port 1234 --allow-websocket-origin="*" --auth-module=test/auth.py`
> 
> So if I login to `localhost:1234/login` and enter the proper username/password, it redirects fine.
> 
> But I can also bypass the login page by going to `localhost:1234/test`. I thought by passing --auth-module, panel would automatically redirect to the /login endpoint if `pn.state.cookies["user"]` did not exist.
> 
> Any ideas on how I can fix this?

I have the same question now.  Before really resolve it, I use this https://github.com/holoviz/panel/issues/4113 as the temporary resolution.

PR #4684, opened 2023-04-23
Add DummyProvider for quick password based auth

MridulS:
This hopefully gives a solution of quick password based authentication. No need to setup a full fledged OAuth app.

This fixes https://github.com/holoviz/panel/issues/2575 and also addresses https://discourse.holoviz.org/t/tips-guide-on-creating-a-simple-username-password-auth-for-panel-app/2582/13.

The PR adds 2 config options: `--dummy-auth` and `--dummy-login-template`. With `--dummy-auth` user can set the password and with `--dummy-login-template` they can override the default user login page (which I stole from https://github.com/bokeh/bokeh/tree/branch-3.2/examples/server/app/server_auth).

An example how to use this:

```
$ panel serve --dummy-auth "panel" --cookie-secret "SECRET" .......
```

This should open up something like:

Screenshot 2023-04-24 at 1 29 32 AM

I also changed couple of things in auth.py let me know if I should make another PR instead:

- The endpoints code doesn't seem like it is required? https://github.com/holoviz/panel/pull/3656 introduced it but it doesn't seem to fix the issue.

- For pathlib, `is_file()` is the equivalent to `os.path.isfile`, trying to inject templates was erroring out on my machine.

TODO:
- [x] Add to Authentication docs
- [x] Maybe add an example test for dummy auth?


philippjfr:
Love this, thanks @MridulS! Some docs would indeed be nice, on the naming side I think we can name this something like `SimpleAuth`. Some other thoughts:

- Might be nice to allow the user to specify a json or yaml file with username: password pairs

Adding tests would also be nice, think the best way is probably just a UI/integration test (using playwright) that starts a server and then goes through the login flow.

MarcSkovMadsen:
BasicAuth could be an alternative to SimpleAuth. In user forums people use the term "basic" rather than "simple" when searching for this type of auth.

philippjfr:
Thanks @MridulS!

PR #4689, opened 2023-04-25
Move dev in diff envs to how-to and revamp install page

droumis:
Addressed one task of #4611 

Moved info about developing in different editors from the installation page to separate How-to guides. 

I've also reworked the installation page. There are two big changes that require input:
- Can we recommend installing from the defaults channel now?
- Can I remove the mention of installing examples?
- **UPDATE**: from HoloViz meeting on April 25, the answer was yes to both of these questions. 

_builtdocs_getting_started_installation html

builtdocs_how_to_display_editor html

builtdocs_how_to_display_jupyterlab html


philippjfr (approved):
Thank you, this looks good. I'll remove `panel examples` in a subsequent PR.

PR #4690, opened 2023-04-25
Additional layout compatibility

philippjfr:
Reintroduces a `config.layout_compatibility` flag which toggles between warnings and errors.

- We now warn/error if a responsive sizing_mode is set but you also provide a fixed width/height along that dimension
- We **always** compute known widths/heights inside a responsive container to ensure that we have at least as much size available as the largest object with a known width/height.

### DIFF
--- a/panel/config.py
+++ b/panel/config.py
@@ -135,6 +135,12 @@ class _config(_base_config):
     global_loading_spinner = param.Boolean(default=False, doc="""
         Whether to add a global loading spinner for the whole application.""")
 
+    layout_compatibility = param.Boolean(default=True, doc="""
+        Provide compatibility for older layout specifications. Incompatible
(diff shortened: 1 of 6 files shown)


MarcSkovMadsen:
Would it be an idea to use a more specific name like fail_instead_warn ? Compatibility layer - what does that mean or refer to?

philippjfr:
>Would it be an idea to use a more specific name like fail_instead_warn ?

fail_instead_warn for what? That seems even less clear to me. The 1.0 migration guide will very specifically reference this compatibility layer.

philippjfr:
I'll merge but if you feel strongly about the name please ping me again.

Hoxbro:
Maybe instead of being set by a boolean, it could be a string with the options of `warning` and `error`. 

Then people would not need to read the docstring.

Issue #4681, opened 2023-04-22
Reorder and Rename top entries in toc tree


MarcSkovMadsen:
Based on my experience with Panel we should

- Rename the "Gallery" to "App Gallery" and move it to the right (After Reference Gallery)
- Rename "Reference Gallery" to "Component Gallery". The word might not cover 100% but users would know what to expect here. In my experience the "Component Gallery" is used very often when a user has learned the basics. Component Gallery is also what every javascript framework has. So I think this is the right word to use.
- Consider renaming "Background" to "Explanation". Diataxis and Matplotlib uses this term. This will help users recognize what this is.

![image](https://user-images.githubusercontent.com/42288570/233802039-3d31a7d6-3f86-4b50-9be1-2d84e60c432a.png)

MarcSkovMadsen:
I'm experimenting with these changes in https://github.com/holoviz/panel/pull/4674

MarcSkovMadsen:
Ahhh. Makes so much more sense.

![image](https://user-images.githubusercontent.com/42288570/233805264-e35069c3-1598-495a-8734-5b9f27e809a3.png)

philippjfr:
+1 on all of this. Would love to hear @droumis' thoughts but we were never quite happy with "Background" so I'm fully on board.

droumis:
+1 on the Gallery renaming.

I dislike 'Background' more than 'Explanation' and I can't think of anything better so +1 from me on this as well!

maximlt:
Should not forget to set redirect links together with these changes.

Issue #4670, opened 2023-04-19
Issues with `dev` Palmer Penguins app


MarcSkovMadsen:
I'm visiting the `dev` Palmer Penguins app at https://panel-gallery-dev.pyviz.demo.anaconda.com/penguin_crossfilter

- [x] Not taking theme query arg into account
- [x] Starts repeating update cycle when you reset the plot
- [x] The number of selected penguins has black text color. Should be white
- [x] No clear separation between sidebar and main area

## Not taking theme query arg into account

Try opening https://panel-gallery-dev.pyviz.demo.anaconda.com/penguin_crossfilter?theme=dark and you will see

![image](https://user-images.githubusercontent.com/42288570/233150417-3b8046bd-dc4b-44f6-98e3-581926ed5d47.png)

## Starts repeating update cycle when you reset the plot

- Open https://panel-gallery-dev.pyviz.demo.anaconda.com/penguin_crossfilter
- Try Box Selecting some points in the Scatter Plot
- Reset the plots

https://user-images.githubusercontent.com/42288570/233151500-81daba1d-9097-4bdc-ab8e-ff0429a1cb46.mp4

## The number of selected penguins has black text color. Should be white

Text in the header should be white. I also believe this text should out as it is something that updates dynamically.

![image](https://user-images.githubusercontent.com/42288570/233151758-de2d0d38-48e4-4ec8-868f-9e6e6aad50ad.png)

## No clear separation between sidebar and main area

Maybe its on purpose, but there is no longer any styling to clearly separate the `sidebar` and the `main` area.

Maybe its on purpose? Maybe it simplifies things and makes them better?

![image](https://user-images.githubusercontent.com/42288570/233152810-1c3cf86b-72b6-4e7e-ac86-f99856c71c6d.png)

philippjfr:
@MarcSkovMadsen Please do not create issues for each app. Create one issue for all apps.

>Maybe its on purpose, but there is no longer any styling to clearly separate the sidebar and the main area.

No I think we should restore this for all templates.

philippjfr:
https://github.com/holoviz/panel/pull/4671 fixes the theming issue and black text color. The weird selection behavior probably has to be resolved in HoloViews.

Issue #4657, opened 2023-04-16
Gallery examples


philippjfr:
The new Panel gallery is manually curated and should only feature more advanced/impressive examples. Right now to fill it out a little more there are a few examples that do not fully fill that brief, e.g. [iris-kmeans](https://panel-gallery-dev.pyviz.demo.anaconda.com/iris_kmeans) or [altair brushing](https://panel-gallery-dev.pyviz.demo.anaconda.com/altair_brushing). I'm not sure where these should move yet so if you have ideas please chime in. However I also want to feature examples that aren't hosted by us, i.e. impressive examples from the community. One example that I can think of is awesome-panel.org, however we cannot ask external folks to bear the load of directing people to their hosted app. Therefore I think it's also fine just to link to a repo that at least features a video/image of the app and ideally the source code.

If you can think of any examples that should be added please post them here. I definitely want to link to @MarcSkovMadsen's awesome-panel.org and also @sophiamyang's recent ML apps and Youtube videos. If you guys could chime in what form you'd like to see these featured that would be greatly appreciated. Next I plan to review the showcase on Discourse.

Issue #4647, opened 2023-04-15
Search Feature is hard to find and use in dev docs


MarcSkovMadsen:
When using the `dev` docs I had a hard time finding the search feature. First of all because its moved :-) But also because 

- its not very central on the page
- has a color which does not stand out
- Its *cramped* together (small margin) with other icons in the top, right corner.

![image](https://user-images.githubusercontent.com/42288570/232181337-39417b73-3277-40e0-869a-7f768ddb1952.png)

Furthermore it requires more clicks to use it. Before you needed 1 click to start writing your search. 

![image](https://user-images.githubusercontent.com/42288570/232181359-ed4d15af-6ffd-41d8-87e4-7a1ffdc7fbb7.png)

Now it requires two clicks. A 100% increase.

![image](https://user-images.githubusercontent.com/42288570/232181371-96feff35-0f27-453a-b2fb-27e4ae1e51cc.png)

I think the most important Feature is the navigation in menus, and that has been improved. But search is very close in importance and that has not been improved. Lots of users could misleadingly believe there was no search. And then the site is very hard to use.

MarcSkovMadsen:
Just visited Jupyterlite. And I can see our problem is that we have a header with a custom dark color. And the icon color for search and light/ dark theme have not been adjusted accordingly.

I would actually recommend using the same header color as Jupyterlite. It has a much more fresh look and feel :-)

https://user-images.githubusercontent.com/42288570/232197567-66db6ae7-970d-4766-bae3-32e04f39a796.mp4

philippjfr:
>I would actually recommend trying out the same header color as Jupyterlite. It has a fresh look and feel :-)

I'll try it out, but I'm not sure I agree with that.

PR #4693, opened 2023-04-26
Various documentation improvements and upgrade guide

philippjfr:
Fixes https://github.com/holoviz/panel/issues/4681
Fixes https://github.com/holoviz/panel/issues/4670
Fixes https://github.com/holoviz/panel/issues/4647
Fixes https://github.com/holoviz/panel/issues/4615
Fixes https://github.com/holoviz/panel/issues/4501
Fixes https://github.com/holoviz/panel/issues/4657


doc/index.md
```diff
@@ -81,37 +81,37 @@ Panel makes it simple to:
...
+:link: background/index
 :link-type: doc
 
-A more hands on tour taking you through the process of loading some data, displaying it and then building an application around it with some of the rich features that Panel supports.
+Introduces you to some of the core concepts behind Panel and some of the advanced features that make Panel such a powerful library.
 :::
 
-:::{grid-item-card} {octicon}`book;2.5em;sd-mr-1` How-to
-:link: how_to/index
+:::{grid-item-card} {octicon}`book;2.5em;sd-mr-1` Reference
```
MarcSkovMadsen, line 105: Rename to "API Reference"

doc/upgrade.md
```diff
@@ -0,0 +1,135 @@
...
+
+#### `sizing_mode`
+
+The sizing_mode has been the primary way to configure responsiveness of components in Panel and that has not changed. However, previously the layout engine was quite forgiving in the way it interpreted combinations of the `sizing_mode` and explicit `width`/`height` settings that really did not make sense. Before we dive into some of these differences let's briefly go through the different `sizing_mode` options and what they actually mean:
+
+![Container sizing_mode](sizing_modes.svg)
+
+- `stretch_width`: Stretches content/container in width while height is determined by size of contents.
+- `stretch_height`: Stretches content/container in height while width is determined by size of contents.
+- `stretch_both`: Stretches content/container in both width and height.
```
MarcSkovMadsen, line 29: What does `content` and `container` mean here?

doc/upgrade.md
```diff
@@ -0,0 +1,135 @@
...
+
+#### Ambiguous configurations
+
+In the past Panel would happily accept ambiguous layout configurations, e.g. providing a fixed width while also setting a responsive `sizing_mode` along the same dimension. As an example take the following:
+
+```python
+pn.pane.Image(..., sizing_mode='stretch_both', width=500)
+```
+
+This specification is ambiguous. Did you want it to stretch the width or did you want to set a fixed width? In the past this would be resolved in favor of the fixed width, i.e. fixed sizing would take precedence over responsiveness, but in certain cases the behavior was completely undefined. This meant it was quite unpredictable what behavior you would actually get. In Panel 1.0 we will warn that this is not a supported configuration and turn the `width` into a `min_width`. In future this conflicting specification will simply error. To toggle these errors on instead of warning set `layout_compatibility` to `'error'`:
```
MarcSkovMadsen: Are you sure. My impression was that the `width` was ignored.

doc/upgrade.md
```diff
@@ -0,0 +1,135 @@
...
+
+The sizing_mode has been the primary way to configure responsiveness of components in Panel and that has not changed. However, previously the layout engine was quite forgiving in the way it interpreted combinations of the `sizing_mode` and explicit `width`/`height` settings that really did not make sense. Before we dive into some of these differences let's briefly go through the different `sizing_mode` options and what they actually mean:
+
+![Container sizing_mode](sizing_modes.svg)
+
+- `stretch_width`: Stretches content/container in width while height is determined by size of contents.
+- `stretch_height`: Stretches content/container in height while width is determined by size of contents.
+- `stretch_both`: Stretches content/container in both width and height.
+
+Now let us explore what changed in this release. In the past any responsive content would force a container to also become responsive. In this release this is no longer guaranteed, i.e. you should generally set the `sizing_mode` on each container if you want your application to be responsive.
```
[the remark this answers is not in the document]
philippjfr: Setting up an environment now, but in the past it definitely would have expanded the container. The behavior was certainly not very well defined though.

doc/upgrade.md
MarcSkovMadsen: Maybe that was for `stretch_width`?

philippjfr: Nope, just double checked you are correct, although I found one or two exceptions.

MarcSkovMadsen:
Consider updating the `doc/index.md` header.

I like the `The powerful data app and dashboarding solution for Python`.

But anything that can help explain and set Panel a bit more apart than *A flexible, high-level app and dashboarding solution for Python*. We should not just be one of many but have some unique identity/ selling point.

doc/upgrade.md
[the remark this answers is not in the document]
philippjfr: Ugh, it's worse, here's an image with `stretch_both` in a Row container without any sizing_mode set:

Screen Shot 2023-04-26 at 18 46 15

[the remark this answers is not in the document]
philippjfr: I've mainly configured it to auto-expand containers based on the contents for backward compatibility and in my testing this works quite well but I'm having a tough time claiming that it is guaranteed to work.

doc/upgrade.md
philippjfr: Thanks for the sanity check

doc/upgrade.md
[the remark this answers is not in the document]
philippjfr: I have decided to very explicitly write out the rules of container expansion.

PR #4701, opened 2023-04-26
Logo refresh

philippjfr:
This is a minimal logo refresh with two relatively small changes:

- Aligns the logo more closely with the default theme (colorwise)
- Rounds the corners

We can still consider a complete overhaul at a later stage but this refresh is merely meant to indicate that a) we have paid some attention to styling and UI and b) align colors with our default theme which uses this blue (HEX #0072b5).


maximlt:
Looks good to me (blue's my favorite color...). I've got no marketing hat and no clue whether it's a good idea to make a small change to the logo before an eventual bigger change. It seems like the 1.0 release allows you to do so, and a logo update can pretty much happen any time later.

Hoxbro:
I don't have any objection to the changes to the logo. 

A point for 1.0 is to tell the world that Panel is stable, and multiple logo changes over a short period do not seem very stable to me.

droumis:
I am not a fan of Panel-green, but I'm not sure this blue is enough of an improvement to counterbalance the potential downsides of multiple obvious logo changes in a short period.

How about just rounding the corners for now and saving more obvious changes for a complete overhaul? Small tweaks like rounding the corners would give the sense that things are more modern without potentially confusing people or making resources with the old logo appear outdated.

Either way, when we are ready to make noticeable changes to the Panel logo, we should update the logo everywhere.

philippjfr:
Not a bad suggestion to just do the rounding, just concerned that the logo already looks out of place on the website.

Also to be clear while I want to leave the option to overhaul the logo open there's a good chance that won't happen for quite a while.

jlstevens:
For what it is worth, I think this is a nice improvement appropriate for 1.0 and that coming up with a whole new logo will be a while: I wouldn't worry too much about confusing people in this regard.

philippjfr:
Thanks everyone, I tend to agree. I think these minor tweaks are appropriate for a 1.0 release. A more thorough update won't materialize any time soon, so on balance I feel like now is a good time.

PR #4717, opened 2023-04-27
Minor improvements to home page

MarcSkovMadsen:
Minor suggestion after Philipps update

### DIFF
--- a/doc/index.md
+++ b/doc/index.md
@@ -10,7 +10,7 @@ width: 125
 
 # Overview
 
-<h2 style="margin-top: 0.3em;">The powerful data exploration and application framework for Python</h2>
+<h1 style="margin-top: 0.3em;">The powerful data exploration and application framework for Python</h1>
 
 ::::{grid} 2 2 5 5
 :gutter: 1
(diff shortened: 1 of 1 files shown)


doc/index.md
```diff
@@ -10,7 +10,7 @@ width: 125
 
 # Overview
 
-<h2 style="margin-top: 0.3em;">The powerful data exploration and application framework for Python</h2>
+<h1 style="margin-top: 0.3em;">The powerful data exploration and application framework for Python</h1>
```
MarcSkovMadsen, line 13: I think the `h2` makes the header small which to me is strange that the first header is not the largest in the page.

But maybe its because you want it to fill up just one line?

doc/index.md
```diff
@@ -10,7 +10,7 @@ width: 125
...
```
MarcSkovMadsen, line 14: An alternative title (easier to read?) could be

"The powerful framework for data exploration and application development in Python"

doc/index.md
```diff
@@ -57,7 +57,7 @@ width: 125
...
+Panel is an [open-source](https://github.com/holoviz/panel/blob/main/LICENSE.txt) Python library that lets you **easily build powerful tools, dashboards and complex applications entirely in Python**. It has a batteries-included philosophy, putting the PyData ecosystem, powerful data tables and much more at your fingertips. High-level reactive APIs and lower-level callback based APIs ensure you can quickly build exploratory applications, but you aren't limited if you build complex, multi-page apps with rich interactivity. Panel is a member of the [HoloViz](https://holoviz.org/) ecosystem, your gateway into a connected ecosystem of data exploration tools.
```
MarcSkovMadsen, line 60: I think easily is more important at a first glance to many users than richly interactive.

MarcSkovMadsen: And tools is another use case than dashboards and complex applications which at least is one of my very frequent use cases.

doc/index.md
MarcSkovMadsen: I reverted it.

Issue #4718, opened 2023-04-27
vtk_slicer demo app issues


MarcSkovMadsen:
https://panel-gallery-dev.pyviz.demo.anaconda.com/vtk_slicer

[] Its not using the Panel Icon

![image](https://user-images.githubusercontent.com/42288570/234919024-45a1c520-72fb-4ce1-ba5a-c3510c38d883.png)

[] Widgets in the sidebar are not responsive. So when the window width is decreased they get cut off.

![image](https://user-images.githubusercontent.com/42288570/234919514-9220442c-db68-44d8-b57d-a674bf498bf7.png)

[] There is no loading indicator. For example when changing the dataset the app is inactive for a longer period of time.

philippjfr:
> Its not using the Panel Icon

Yeah there's something off about template inheritance, so it inherits from the Bokeh base template, not ours.

> Widgets in the sidebar are not responsive. So when the window width is decreased they get cut off.

Easy fix, want to submit a PR? 😄 

>There is no loading indicator. For example when changing the dataset the app is inactive for a longer period of time.

Also seems like a simple fix, same PR? 😁

MarcSkovMadsen:
ok

PR #4722, opened 2023-04-27
Fix vtk_slicer

MarcSkovMadsen:
Fixes 2 of 3 issues in https://github.com/holoviz/panel/issues/4718

- Makes controls on left responsive
- Adds loading indication to the controls Column. I added it specifically for when the Dataset changes as that is what takes a long time.

Did not fix the `favicon` issue.


philippjfr:
Perfect, I'll look at template inheritance separately.

Issue #1904, opened 2021-01-13
Scrollable log text viewer - implementation


GilShoshan94:
#### Is your feature request related to a problem? Please describe.
Hi, I needed a scrollable log text viewer to see the output of a server communicating over socket (I used ZMQ) to the python server Panel runs. I found on Discourse https://discourse.holoviz.org/t/scrollable-log-text-viewer/317. 

The solution there gave me the idea to improve even further. Now I want to contribute back to Panel. 
But being a noob in web technology (until last week did not know anything in JS, CSS, HTML, nor Browser stuff...), I prefer to share my implementation using Panel available tools and let the expert add it where it should in the library.

#### Describe the solution you'd like
A text box like a terminal, print each line and auto scroll down unless the user scroll up, but get back to auto scroll if the user scroll back down. So the scrolling is "smart" and "sticky".
Also there is a `max_history` argument to not overflow the web page (it keeps only the last `max_history` logs)

Here an example of the "smart" and "sticky" scroll:
<img src="https://user-images.githubusercontent.com/29143941/104458199-5217eb80-55ab-11eb-9167-c862607b90ff.gif" width="350">

Here an example of the `max_history` (set to 70 here):
<img src="https://user-images.githubusercontent.com/29143941/104458335-84c1e400-55ab-11eb-90c9-48e1b72542e5.gif" width="350">

#### Code:
I use a 2 HTML pane, one (`html_pane`) to render the logviewer itself and the second (`html_tmp`), hidden, just to write to the first one with JS. In the beginning I used a jscallback and trigger it by writting to `html_tmp.object`. 
Then I realize I can do a direct call to JS by writting to `html_tmp.object = f"""<script>{js_code}</script>"""`.
I also paid attention that if you write twice in a row identical content to `html_tmp.object`, nothing happen, no jscallback  nor direct callback. I guess the DOM doesn't update if the content doesn't change. So I clear `html_tmp.object` with an empty string after each write, and have to wait a little bit (`time.sleep(0.000001)`) otherwise there is a race a and the update miss most of the time.
Also I wanted to return a Panel Object to work with, so I use 2 classes, one private and one public that inherit from `pn.Column`.

Enough explanation. If you want/need more, just ask me and I will answer.
Here is my finale working code (I tested a lot of input, also UTF-8, and let the user to write text or html).
#### Python:
```python
from html import escape
from pathlib import Path

import panel as pn


def setup_panel_extention() -> Path:
[...]

PR #4723, opened 2023-04-27
Adds the ChatBox widget

ahuang11:
```
import panel as pn
pn.extension()

dummy_icon_link = "https://cdn2.iconfinder.com/data/icons/business-hr-and-recruitment/100/account_blank_face_dummy_human_mannequin_profile_user_-1024.png"

chatbox = pn.widgets.ChatBox(
    user_messages=[
        {"You": "Hello!"},
        {"Dummy": "Greetings!"},
        {"You": "Question for you..."},
        {"You": "What is the meaning of life?"},
        {"Dummy": "I don't know! Ask ChatGPT!"},
        {"Bot": "Beep boop beep! Just testing..."}
    ],
    user_icons={
        "Dummy": dummy_icon_link,
    },
    max_width=1000,
).servable()
```

image


panel/widgets/chatbox.py
```diff
@@ -0,0 +1,266 @@
...
+            )
+        elif isinstance(icon, str):
+            pane_type = PNG
+            if icon.endswith(".jpg"):
+                pane_type = JPG
+            elif icon.endswith(".svg"):
+                pane_type = SVG
+            elif icon.endswith(".gif"):
+                pane_type = GIF
+            self._icon = pane_type(icon, **icon_params)
```
philippjfr, line 102: In 1.0 (and on `main`) you can just use `Image` here.

panel/widgets/chatbox.py
```diff
@@ -0,0 +1,266 @@
...
+            # maybe a bytes stream?
+            self._icon = panel(icon, **icon_params)
+
+        # create the message bubble
+        box_width = message_layout.get("width", 100)
+        text_width = int(box_width / 2)  # make text start a new line
+        wrapped_text = "\n".join(wrap(value, width=text_width))
+        self._bubble = StaticText(
+            value=wrapped_text,
+            style=style,
```
philippjfr, line 113: In 1.0 this will be `styles` and may not work as-is, i.e. you might have to provide `stylesheets` instead.

ahuang11: Yup! Seems broken. I couldn't simply replace styles -> stylesheets. I also tried to use `stylesheets=[styles]` but I encountered
```
ValueError: List parameter 'stylesheets' must be a list, not an object of type <class 'dict'>.
```

And then without wrapping a [] around it:
```
ValueError: failed to validate Div(id='p1130', ...).stylesheets: expected an element of List(Either(Instance(StyleSheet), String, Dict(String, Either(Dict(String, Nullable(String)), Instance(Styles))))), got seq with invalid items [{'color': 'grey'}]
```
 `List(Dict(String, Either(Dict(String, Nullable(String)))` should work

image

panel/widgets/__init__.py
```diff
@@ -34,6 +34,7 @@
 """
 from .base import CompositeWidget, Widget  # noqa
 from .button import Button, MenuButton, Toggle  # noqa
+from .chatbox import ChatBox  # noqa
```
ahuang11, line 37: Are there any other places I need to add an import?

When I try to import it outside the widgets directory, I get:
```
  File "/Users/airbook/test.py", line 4, in <module>
    pn.widgets.ChatBox().servable()
AttributeError: module 'panel.widgets' has no attribute 'ChatBox'
```

panel/widgets/chatbox.py
```diff
@@ -0,0 +1,271 @@
...
+        self._chat_log = Column(**chat_layout)
+        self._user_input = TextInput(
+            name="Type your message",
+            placeholder="Press Enter to send",
+            sizing_mode="stretch_width",
+            width_policy="max",
+        )
+
+        # this doesn't work:
+        # self.param.watch(self._display_all_messages, "user_messages")
```
ahuang11, line 202: Is there another way of doing this? I thought if `user_messages` changes, it would trigger `self._display_all_messages` but it doesn't react.

panel/widgets/__init__.py
Hoxbro: It seems to work fine for me:

image

Are you in the correct environment?

panel/widgets/chatbox.py
Hoxbro: Try to take a look at this:  https://pyviz-dev.github.io/panel/upgrade.html#component-stylesheets

panel/widgets/chatbox.py
Hoxbro: Appending to a list does not change the object and does not trigger the method.

You can run this `self.param.trigger("user_messages")` after you have appended something to the list.

panel/widgets/__init__.py
ahuang11: Yeah; perhaps my environment is corrupted. I'll try recreating later.
```
(panel_dev) airbook@AHs-MacBook-Pro ~ % panel serve test.py
2023-04-28 09:50:39,763 Starting Bokeh server version 3.1.0 (running on Tornado 6.3)
2023-04-28 09:50:39,763 User authentication hooks NOT provided (default user enabled)
2023-04-28 09:50:39,765 Bokeh app running at: http://localhost:5006/test
2023-04-28 09:50:39,765 Starting Bokeh server with process id: 42681
2023-04-28 09:50:41,196 Error running application handler <bokeh.application.handlers.script.ScriptHandler object at 0x119d68ee0>: module 'panel.widgets' has no attribute 'ChatBox'
File 'test.py', line 4, in <module>:
pn.widgets.ChatBox().servable() Traceback (most recent call last):
  File "/Users/airbook/miniforge3/envs/panel_dev/lib/python3.9/site-packages/bokeh/application/handlers/code_runner.py", line 231, in run
    exec(self._code, module.__dict__)
  File "/Users/airbook/test.py", line 4, in <module>
    pn.widgets.ChatBox().servable()
AttributeError: module 'panel.widgets' has no attribute 'ChatBox'

2023-04-28 09:50:41,438 WebSocket connection opened
2023-04-28 09:50:41,439 ServerConnection created
2023-04-28 09:50:42,364 WebSocket connection closed: code=1001, reason=None
^C
Interrupted, shutting down
(panel_dev) airbook@AHs-MacBook-Pro ~ % which panel
/Users/airbook/miniforge3/envs/panel_dev/bin/panel
```

panel/widgets/chatbox.py
ahuang11: Thanks, that was helpful! I think using `styles` is more suitable in this case.

ahuang11:
I was testing this and I discovered that users cannot append / replace chat messages programmatically, e.g.
```
import panel as pn
from chatbox import ChatBox
pn.extension()

chatbox = ChatBox(user_messages=[{"You": "Hello"}, {"Bot": "Hey"}])
chatbox.user_messages[:] = [{"You": "How are you?"}]
chatbox.servable()
```
Any ideas on how to address this?

ahuang11:
Interestingly, if I use `param.ClassSelector` it works (but not append)

```
import panel as pn
from chatbox import ChatBox
pn.extension()

chatbox = ChatBox(user_messages=[{"You": "Hello"}, {"Bot": "Hey"}])
chatbox.user_messages = [*chatbox.user_messages, {"You": "How are you?"}]
chatbox.servable()
```

Hoxbro:
I think the problems you are seeing are the same. You are not changing the list, only updating the existing one. You can check this by using `id`. Maybe add a method like `chatbox.submit(msg)`, which appends to the list and triggers the update? 

``` python
import panel as pn
pn.extension()

chatbox = pn.widgets.ChatBox(user_messages=[{"You": "Hello"}, {"Bot": "Hey"}])
print(id(chatbox.user_messages))

chatbox.user_messages[:] = [{"You": "How are you?"}]
print(id(chatbox.user_messages))

chatbox.user_messages.append({"Bot": "Beep boop"})
print(id(chatbox.user_messages))

chatbox.user_messages = [*chatbox.user_messages, {"You": "How are you?"}]
print(id(chatbox.user_messages))
```

image

ahuang11:
Right, I understand that it doesn't change the underlying object container.

I was thinking how I could implement something similar to `pn.Column`.

Specifically, how `pn.Column().append` works (seems like it's just overriding the `append` method and setting `self.objects = new_objects`.

Later I realized, doing that is not equivalent to what I'm doing. The equivalent would be `pn.Column().objects.append` which also does not replace the container and thus, does not update.

So yeah I should have a public submit / send method.

ahuang11:
Okay, I added append/extend/clear, but not sure if I should rename it since it doesn't support all the list-like methods.

I think it's in a decent position now, and I can start adding some tests later.

ahuang11:
image
image
image
image

panel/widgets/chatbox.py
```diff
@@ -0,0 +1,359 @@
...
+        # add interactivity
+        self.param.watch(self._refresh_log, "value")
+        self.param.trigger("value")
+
+
+    def _generate_dark_color(self) -> str:
+        """
+        Generate a random dark color in hexadecimal format.
+        """
+        r, g, b = random.randint(0, 127), random.randint(0, 127), random.randint(0, 127)
```
Hoxbro, line 223: Maybe this could use the username to generate the color, so it is the same color for the same username each time you instantiate the chatbox.

Hoxbro (commented):
A great contribution! I have left some very small comments. :)

MarcSkovMadsen:
It looks like a great contribution.

I have one request on the fundamental level. Please make sure that the api supports or can be extend to support Panel components in general. I envision that we want the bot to be able to answer with more than text. For example nicely formatted markdown, plots, code, tables and even small working Panel applications.

My POC in the original issue would have allowed that.

MarcSkovMadsen:
Another comment is. Please consider if it could be possible to support feedback to responses from the bot. For example ChatGPT allows you to give a thumbs up or down.

Then a user that would like a UI to easily provide feedback to his AI can construct this using Panel.

MarcSkovMadsen:
Another comment is that the users might want to provide a specific input widget. For example if the conversation is about code they would like to provide the `Ace`  `CodeEditor` widget. Some might want `TextInput`, `TextAreaInput`, or another other widget in fact. Would be nice if it was possible to set the input widget on construction and maybe even dynamically.

panel/widgets/chatbox.py
```diff
@@ -0,0 +1,359 @@
...
+    show_name = param.Boolean(
+        default=True,
+        doc="""Whether to show the name of the user""",
+    )
+
+    _composite_type: ClassVar[Type[ListPanel]] = Column
+
+    def __init__(
+        self,
+        value: AnyStr,
```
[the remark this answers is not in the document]
ahuang11: I initially thought of using AnyStr for numpy string / pandas string items, but perhaps not needed.

Maybe later I'll change it back to AnyStr to address' Marc's interest in posting images.

panel/widgets/chatbox.py
```diff
@@ -0,0 +1,359 @@
...
+        if color in self.user_colors.values():
+            return self._generate_dark_color()
+        return color
+
+    @staticmethod
+    def _get_name(dict_: Dict[str, str]) -> str:
+        """
+        Get the name of the user who sent the message.
+        """
+        return list(dict_.keys())[0]
```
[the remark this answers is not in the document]
ahuang11: I like it! :) I'm all for these nits.

panel/widgets/chatbox.py
ahuang11: Brilliant idea!

ahuang11:
Thanks for all the feedback!

In regards to "For example nicely formatted markdown, plots, code" kind of have this now, but not sure why the markdown loses its color on the second line; would need to play around with it.
image

I am not sure how to approach this. Can you elaborate? Would ChatGPT produce a Markdown table and somehow the code detects this to create a tabulator?
"""
Tabulator tables and even small working Panel applications
"""

Interesting idea~
"""
Please consider if it could be possible to support feedback to responses from the bot. For example ChatGPT allows you to give a thumbs up or down.

Then a user that would like a UI to easily provide feedback to his AI can construct this using Panel.
"""

I am hesitant about this and whether this should be included in this PR. Does this mean storing the states of the output?
"""
Maybe this could be thought together with my first request. If its possible to use any Panel component as a message, then we can create message components that are editable. In this way a user can edit a response from a bot if they think it should be improved and the updated message can be used to further train the model.
"""

ahuang11:
Okay I followed your approach of using a `param.List` to support arbitrary objects.
```
    value = param.List(
        doc="""The objects to display""",
    )

    panel_type = param.ClassSelector(
        default=None,
        class_=(Viewable,),
        doc="""
        The type of Panel object to use if an item in value is not
        already rendered as a Panel object; if None, uses the
        panel function to render a displayable Panel object
        """,
    )

...


        # create the chat bubble
        bubble_objects = [
            panel_type(obj) if not isinstance(obj, Viewable) else obj for obj in value
        ]
        self._bubble = Column(
            *bubble_objects,
            margin=12,
            styles=bubble_styles,
            width_policy="max",
        )
```

image

ahuang11:
Feels a bit tacky to have the heart in a toggle button. Let me know if there's a better way to implement clickable icons in Panel.
image

ahuang11:
I think all the review comments have been addressed now.

MarcSkovMadsen:
Great work. I plan to do my review on the train tonight.

Here is inspiration for chat interfaces. https://wattenberger.com/thoughts/boo-chatbots

My main take away is that we should keep flexibility about how we receive input and visualize the output.

MarcSkovMadsen:
## Isssue: Horizontal Scrollbars

When I serve the example

```python
import panel as pn
pn.extension()

dummy_icon_link = "https://cdn2.iconfinder.com/data/icons/business-hr-and-recruitment/100/account_blank_face_dummy_human_mannequin_profile_user_-1024.png"

chatbox = pn.widgets.ChatBox(
    value=[
        {"You": "Hello!"},
        {"Dummy": "Greetings!"},
        {"You": "Question for you..."},
        {"You": "What is the meaning of life?"},
        {"Dummy": "I don't know! Ask ChatGPT!"},
        {"Bot": "Beep boop beep! Just testing..."}
    ],
    user_icons={
        "Dummy": dummy_icon_link,
    },
    max_width=1000,
).servable()
```

image

I get scrollbars. I don't think the horizontal scrollbars are needed and would like to avoid them.

ahuang11:
Thanks Marc for the thorough review and insights!

#1 - I am not sure how to control whether the horizontal / vertical scrollbars show up. I simply set scroll=True in Column; does anyone have an idea on how to fix this?

#2 - I think the text input is at the very bottom of the page, likely because of `box_objects.extend([VSpacer(height_policy="min"), self._input_message])`
image

#3 - Perhaps adding stretch_width to the panel obj could fix this.

#4 - I considered using `TextAreaInput` but users can't just press "Enter" to submit. There would have to be a separate submit button; I suppose that's okay, but for most chat message programs, I'm used to submitting by pressing enter. Are you imagining changing the input box interactively through a radio box button, or simply as an argument before it's instantiated?

#5 Right, seems like I missed adding that to the docs since this feature was patched on late.

panel/widgets/chatbox.py
```diff
@@ -0,0 +1,435 @@
...
+from panel.pane.base import PaneBase, panel as _panel
+from panel.pane.image import Image
+from panel.pane.markup import Markdown
+from panel.viewable import Layoutable, Viewable
+from panel.widgets.base import CompositeWidget
+from panel.widgets.button import Toggle
+from panel.widgets.input import StaticText, TextInput
+
+
+class ChatRow(CompositeWidget):
```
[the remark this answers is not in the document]
ahuang11: I suppose I need it for ChatBox too.

panel/widgets/chatbox.py
```diff
@@ -0,0 +1,435 @@
...
+        """
+        self.liked = event.new
+        if self.liked:
+            event.obj.name = "❤️"
+        else:
+            event.obj.name = "♡"
+
+
+class ChatBox(CompositeWidget):
+    value = param.List(
```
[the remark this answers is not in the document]
ahuang11: Can you clarify this? I'm not sure I understand.

panel/widgets/chatbox.py
```diff
@@ -0,0 +1,435 @@
...
+        Convert an object to a Panel object.
+        """
+        if isinstance(obj, Viewable):
+            return obj
+
+        try:
+            if default_panel is None or issubclass(default_panel, PaneBase):
+                panel_obj = (default_panel or _panel)(obj)
+            else:
+                panel_obj = default_panel(value=obj)
```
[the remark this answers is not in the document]
ahuang11: I think it's already supported since default_panel is any Callable, so perhaps just renaming it to `default_callable`

ahuang11:
Ok, did not expect this PR to become so big, but I'm happy that I was able to address all of your concerns.

I think perhaps in a separate PR, the `ListPanel` class can expand beyond just `scroll=True` and use `scroll="x"` or `scroll="y"`. Right now I'm doing:
```
        if "overflow-wrap" not in panel_obj.styles:
            panel_obj.styles.update({"overflow-wrap": "break-word"})
```

image

panel/widgets/chatbox.py
```diff
@@ -0,0 +1,520 @@
...
+            self._input_button.on_click(self._enter_message)
+
+            if len(self._message_inputs) > 1:
+                input_items = Tabs(
+                    *zip(self._message_inputs.keys(), self._message_inputs.values())
+                )
+            else:
+                input_items = list(self._message_inputs.values())[0]
+            input_row = Row(input_items, self._input_button)
+            box_objects.extend([VSpacer(height_policy="min"), input_row])
```
ahuang11, line 348: This part is getting large. Should probably separate into a method.

panel/widgets/chatbox.py
```diff
@@ -0,0 +1,520 @@
...
+                default_message_callable, PaneBase
+            ):
+                panel_obj = (default_message_callable or _panel)(obj)
+            else:
+                panel_obj = default_message_callable(value=obj)
+        except ValueError:
+            panel_obj = _panel(obj)
+
+        if "overflow-wrap" not in panel_obj.styles:
+            panel_obj.styles.update({"overflow-wrap": "break-word"})
```
ahuang11, line 199: This should probably be built-in into Markdown object in a separate PR.

panel/widgets/chatbox.py
```diff
@@ -0,0 +1,520 @@
...
+        # Set up layout
+        layout = {
+            p: getattr(self, p)
+            for p in Layoutable.param
+            if p not in ("name", "height", "styles", "margin")
+            and getattr(self, p) is not None
+        }
+        chat_layout = dict(
+            sizing_mode="stretch_both",
+            styles={"overflow-y": "auto", "overflow-x": "auto"},
```
ahuang11, line 311: This should be added to the ListPanel `scroll` capabilities

ahuang11:
This should be ready now; let me know if you have any other suggestions or feedback!

MarcSkovMadsen:
I plan to review tomorrow.

My plan is to combine it with PandasApi to test its flexibility and then Maybe contribute that as a Gallery example. I will probably Mock the PandasApi such that we do not get an additional dependency.

ahuang11:
Sounds good~ hopefully this can be released alongside v1 too!

MarcSkovMadsen:
# Issues - Round 2

## 1. Cannot set `sizing_mode`.

```python
import panel as pn
pn.extension(template="fast")

dummy_icon_link = "https://cdn2.iconfinder.com/data/icons/business-hr-and-recruitment/100/account_blank_face_dummy_human_mannequin_profile_user_-1024.png"

chatbox = pn.widgets.ChatBox(
    value=[
        {"You": "Hello!"},
        {"Dummy": "Greetings!"},
        {"You": "Question for you..."},
        {"You": "What is the meaning of life?"},
        {"Dummy": "I don't know! Ask ChatGPT!"},
        {"Bot": "Beep boop beep! Just testing..."}
    ],
    message_icons={
        "Dummy": dummy_icon_link,
    },
    sizing_mode="stretch_width"
).servable()
```

image

## 2 Message Panel should scroll

After having written a few messages the message panel expands and the text input goes out of the visible window. This is not a nice user experience.

```python
import panel as pn
pn.extension(template="fast")

dummy_icon_link = "https://cdn2.iconfinder.com/data/icons/business-hr-and-recruitment/100/account_blank_face_dummy_human_mannequin_profile_user_-1024.png"

chatbox = pn.widgets.ChatBox(
    value=[
        {"You": "Hello!"},
        {"Dummy": "Greetings!"},
        {"You": "Question for you..."},
        {"You": "What is the meaning of life?"},
        {"Dummy": "I don't know! Ask ChatGPT!"},
        {"Bot": "Beep boop beep! Just testing..."}
    ],
    message_icons={
        "Dummy": dummy_icon_link,
    },
).servable()
```

https://user-images.githubusercontent.com/42288570/236658290-17c7f09a-d764-4542-a2c1-b391f0c70752.mp4

## 3 Does not show the latest message

If I try to fix 2 by setting `height=800` then the message panel does not scroll to show the latest message.

```python
import panel as pn
pn.extension(template="fast")

dummy_icon_link = "https://cdn2.iconfinder.com/data/icons/business-hr-and-recruitment/100/account_blank_face_dummy_human_mannequin_profile_user_-1024.png"

chatbox = pn.widgets.ChatBox(
    value=[
        {"You": "Hello!"},
        {"Dummy": "Greetings!"},
        {"You": "Question for you..."},
        {"You": "What is the meaning of life?"},
        {"Dummy": "I don't know! Ask ChatGPT!"},
        {"Bot": "Beep boop beep! Just testing..."}
    ],
    message_icons={
        "Dummy": dummy_icon_link,
    },
    height=800
).servable()
```

image

ahuang11:
Thanks for testing. Isn't 2 and 3 the same issue that it does not scroll to the latest message?

Might have to implement something similar to https://github.com/holoviz/panel/issues/1904, but let me know if there's a simpler implementation.

I suppose the simpler implementation is moving the text input on the top and prepending messages to the top and let the messages disappear at the bottom.

MarcSkovMadsen:
Maybe 2 and 3 are the same issue and solution.

MarcSkovMadsen:
## Overflow Issue

ChatGPT will sometimes return long (markdown based) text messages. They currently overflow

[script.csv](https://github.com/holoviz/panel/files/11414048/script.csv) (rename to script.py)

image

ahuang11:
I think I addressed most of the issues; I found a work around for 2, but I can't fix 3. Changing the scroll top of the bokeh container does nothing, even in the browser console:
image

Also, I can't reproduce 6, but I did make some changes before testing it.
image

ahuang11:
Well a minimal example seems to have worked...
```

import panel as pn
pn.extension()

col = pn.Column(styles={"overflow-y": "auto"}, sizing_mode="stretch_width", max_height=300)
for i in range(0, 100):
    col.append(pn.pane.Markdown("Hello"))
col.servable()
```

image

ahuang11:
I added scroll to top button, but I can't figure out a way to auto-trigger it on start.

image

`self._scroll_button.param.trigger("clicks")` has no effect, since it only triggers the Python callbacks, not Javascript.

MarcSkovMadsen:
Just for inspiration. We could hook up the chatbox to HugChat as in https://github.com/dataprofessor/hugchat/blob/master/streamlit_app.py.

MarcSkovMadsen:
yes

panel/widgets/chatbox.py
```diff
@@ -0,0 +1,560 @@
...
+        box_objects.insert(0, input_items)
+
+    def _generate_pastel_color(self, string: str) -> str:
+        """
+        Generate a random pastel color in hexadecimal format.
+        """
+        seed = sum([ord(c) for c in string])
+        random.seed(seed)
+
+        rgb_range = (150, 185)
```
Hoxbro, line 402: Maybe have this as a parameter? It could be nice to customize this range.

Hoxbro (commented):
Some small comments. 

The bind in the notebook does not work for me. It needs to be changed to something like this `pn.bind(lambda liked: print(liked), row.param.liked, watch=True)`.

If I `allow_likes=True` in the second cell and run `row.liked = True` in the seventh cell, the like does not update in the chatbox.

It could be nice to have a way of setting the location of the `message_input_widget` to the top or the bottom. Also, a setting to change the message flow from `top-to-bottom` to `bottom-to-top`.

panel/widgets/chatbox.py
ahuang11: That's a good suggestion!

panel/widgets/chatbox.py
```diff
@@ -0,0 +1,560 @@
...
+        else:
+            background = self._generate_pastel_color(string=user)
+            self.message_colors[user] = background
+
+        # try to get input icon
+        message_icon = self.message_icons.get(user, None)
+
+        align = "start" if user != self.primary_name else "end"
+        if not isinstance(message_contents, List):
+            message_contents = [message_contents]
```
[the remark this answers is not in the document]
ahuang11: I think I implemented this later to allow a list of messages, e.g.

`{"user": ["message 1", picture_obj, "another_message"]}` so it's slightly more convenient to write than `[{"user": "message 1"}, {"user": picture_obj}, {"user": "another_message"}]`

ahuang11:
Thanks for the thorough review!

I think I've addressed them all:
1. You're right; I swapped the function / and param and left out the param
2. Fixed by linking the `liked` param to the `_like` Toggle bidirectionally
3. I added "ascending"
image

MarcSkovMadsen:
My remaining feedback are two concerns

- That the current implementation will be harder to add more complex styling and `js` based features do. I imagine that one day this would be reimplemented in ReactiveHTML or as a Bokeh model.
- The default colors are not my favourites. But I don't have a better suggestion right now.

I think both my concerns should be blocking approval as they can be addressed later without user impact.

MarcSkovMadsen:
For future improvements the Shiny chat might also give some inspiration.

ahuang11:
I'm going to guess TextInput can't be instantiated in default.
```

          File "<string>", line 2, in <module>
          File "<pip-setuptools-caller>", line 34, in <module>
          File "/home/runner/work/panel/panel/setup.py", line 317, in <module>
            setup(**setup_args)
          File "/usr/share/miniconda3/envs/test-environment/lib/python3.8/site-packages/setuptools/__init__.py", line 107, in setup
            return distutils.core.setup(**attrs)
          File "/usr/share/miniconda3/envs/test-environment/lib/python3.8/site-packages/setuptools/_distutils/core.py", line 185, in setup
            return run_commands(dist)
          File "/usr/share/miniconda3/envs/test-environment/lib/python3.8/site-packages/setuptools/_distutils/core.py", line 201, in run_commands
            dist.run_commands()
          File "/usr/share/miniconda3/envs/test-environment/lib/python3.8/site-packages/setuptools/_distutils/dist.py", line 969, in run_commands
            self.run_command(cmd)
          File "/usr/share/miniconda3/envs/test-environment/lib/python3.8/site-packages/setuptools/dist.py", line 1244, in run_command
            super().run_command(command)
          File "/usr/share/miniconda3/envs/test-environment/lib/python3.8/site-packages/setuptools/_distutils/dist.py", line 988, in run_command
            cmd_obj.run()
          File "/home/runner/work/panel/panel/setup.py", line 56, in run
            _build_paneljs()
          File "/home/runner/work/panel/panel/setup.py", line 38, in _build_paneljs
            from panel.compiler import bundle_resources
          File "/home/runner/work/panel/panel/panel/__init__.py", line 50, in <module>
            from . import pane  # noqa
          File "/home/runner/work/panel/panel/panel/pane/__init__.py", line 37, in <module>
            from .holoviews import HoloViews, Interactive  # noqa
          File "/home/runner/work/panel/panel/panel/pane/holoviews.py", line 26, in <module>
            from ..widgets import Player
          File "/home/runner/work/panel/panel/panel/widgets/__init__.py", line 37, in <module>
            from .chatbox import ChatBox  # noqa
          File "/home/runner/work/panel/panel/panel/widgets/chatbox.py", line 223, in <module>
            class ChatBox(CompositeWidget):
          File "/home/runner/work/panel/panel/panel/widgets/chatbox.py", line 301, in ChatBox
            TextInput(placeholder="Enter or click send when complete."),
          File "/home/runner/work/panel/panel/panel/widgets/base.py", line 73, in __init__
            super().__init__(**params)
          File "/home/runner/work/panel/panel/panel/reactive.py", line 538, in __init__
            super().__init__(**params)
          File "/home/runner/work/panel/panel/panel/reactive.py", line 115, in __init__
            super().__init__(**params)
          File "/home/runner/work/panel/panel/panel/viewable.py", line 679, in __init__
            self._update_design()
          File "/home/runner/work/panel/panel/panel/viewable.py", line 687, in _update_design
            from .theme import Design
          File "/home/runner/work/panel/panel/panel/theme/__init__.py", line 12, in <module>
            from .bootstrap import Bootstrap
          File "/home/runner/work/panel/panel/panel/theme/bootstrap.py", line 10, in <module>
            from ..widgets import Number, Tabulator
        ImportError: cannot import name 'Number' from partially initialized module 'panel.widgets' (most likely due to a circular import) (/home/runner/work/panel/panel/panel/widgets/__init__.py)
        [end of output]
```

ahuang11:
> That the current implementation will be hard to add more complex styling and js based features to. I imagine that one day this would be reimplemented in ReactiveHTML or as a Bokeh model.

Wondering what complex styling do you have in mind?

> The default colors are not my favourites. But I don't have a better suggestion right now.

Yeah I tried to pick a range where it worked in both light and dark mode; but users can change the message_rgb_range to their likings.

ahuang11:
I was watching https://www.youtube.com/watch?v=JSURzPQnkl0 and decided to go with HSL instead of RGB range.

Hoxbro (approved):

philippjfr:
Thanks, this is looking good. I do want to spend some time tweaking the CSS here as there's some funkyness in a few places with things being weirdly sized (not visually but when you inspect the CSS there's definitely some oddness). From a styling perspective I think I'd like to see:

1. Would like to distinguish between the user and the message more clearly. Probably by making the user box circular.
2. I think randomly generated colors have a high chance of being pretty ugly, I'd prefer having a list of ~10 default, well curated (probably pastel) colors to pick from.
3. I'd probably tweak the chat box styling to have some box-shadow.

Sorry for being nitpicky here but I think this is really close to being really awesome.

ahuang11:
No problem; thanks for the feedback! I think it looks nicer now. Let me know if you want anything else changed.

image
image

MarcSkovMadsen:
Just inspiration for additional work later. Not in this PR.

A new chat data app framework https://github.com/Chainlit/chainlit

ahuang11:
I like the styling of it; maybe the buttons in this PR can use icons from
https://panel.holoviz.org/reference/widgets/Button.html#icons

MarcSkovMadsen:
Here is another example we can compare the chatbox to. 

- Can we support the same use case?
- Does our chat bot look and feel just as great. Or even better?

https://twitter.com/andfanilo/status/1664199858285326337?s=20

[image](https://twitter.com/andfanilo/status/1664199858285326337?s=20)

philippjfr:
Hope you don't mind but I've made a number of changes here that I think simplify the UI a bit:

1. By default there are only two colors:
2. More padding around the text
3. Remove the abbreviated user text "icon"
4. Various other small CSS tweaks

Screen Shot 2023-06-05 at 15 29 14

ahuang11:
Thanks, that's fine by me. Looks slick, I'll have to take a deeper look at the CSS changes later to learn more!

PR #4738, opened 2023-04-29
Fix mimetype issue

### LINKED ISSUE #4721: Failed to load module script
I'm testing out panel==1.0.0rc1 (and current `main`) branch.

I again get the MIMETYPE problem as in https://github.com/holoviz/panel/issues/4080

```python
import numpy as np
import panel as pn

# import mimetypes
# mimetypes.add_type("application/javascript", ".js")

pn.extension(template='fast')

freq = pn.widgets.FloatSlider(
    name='Frequency', start=0, end=10, value=5
).servable(target='sidebar')

ampl = pn.widgets.FloatSlider(
    name='Amplitude', start=0, end=1, value=0.5
).servable(target='sidebar')

def plot(freq, ampl):
    fig = plt.figure()
    ax = fig.add_subplot(111)
    xs = np.linspace(0, 1)
    ys = np.sin(xs*freq)*ampl
    ax.plot(xs, ys)
    return fig

mpl = pn.pane.Matplotlib(
    pn.bind(plot, freq, ampl)
)

pn.Column(
    '# Sine curve', mpl
).servable(target='main')
```

image

```bash
Failed to load module script: Expected a JavaScript module script but the server responded with a MIME type of "text/plain". Strict MIME type checking is enforced for module scripts per HTML spec.
fast-components.js:1  Failed to load module script: Expected a JavaScript module script but the server responded with a MIME type of "text/plain". Strict MIME type checking is enforced for module scripts per HTML spec.
script?theme=dark:74  Uncaught TypeError: window.fastDesignProvider is not a constructor
    at HTMLDocument.<anonymous> (script?theme=dark:74:25)
```

If I add 

```python
import mimetypes
mimetypes.add_type("application/javascript", ".js")
```

It works.

I'm on Windows 10 on a new laptop I received two weeks ago.

MarcSkovMadsen:
Closes #4721

### DIFF
--- a/panel/io/resources.py
+++ b/panel/io/resources.py
@@ -122,9 +122,7 @@ JS_URLS = {
 
 extension_dirs['panel'] = str(DIST_DIR)
 
-mimetypes.init()
-if mimetypes.types_map.get('.js') != "application/javascript":
-    mimetypes.add_type("application/javascript", ".js")
+mimetypes.add_type("application/javascript", ".js")
 
 @contextmanager
 def set_resource_mode(mode):
--- a/panel/io/resources.py
+++ b/panel/io/resources.py
@@ -122,6 +122,7 @@ JS_URLS = {
 
 extension_dirs['panel'] = str(DIST_DIR)
 
+mimetypes.init()
 if mimetypes.types_map.get('.js') == 'text/plain':
     logger.warn(
         "Mimetype declaration for .js files was set to 'text/plain'. "
--- a/panel/io/resources.py
+++ b/panel/io/resources.py
@@ -123,13 +123,7 @@ JS_URLS = {
 extension_dirs['panel'] = str(DIST_DIR)
 
 mimetypes.init()
-if mimetypes.types_map.get('.js') == 'text/plain':
-    logger.warn(
-        "Mimetype declaration for .js files was set to 'text/plain'. "
-        "This will interefere with serving JS modules. Ensure your system "
-        "does not override the mimetype declaration and/or contact your "
-        "system administrator. "
-    )
+if mimetypes.types_map.get('.js') != "application/javascript":
     mimetypes.add_type("application/javascript", ".js")
 
 @contextmanager


panel/io/resources.py
```diff
@@ -122,6 +122,7 @@ def conffilter(value):
 
 extension_dirs['panel'] = str(DIST_DIR)
 
+mimetypes.init()
```
MarcSkovMadsen, line 125: To me it looks like the function is idempotent and can be run multiple times and thus its ok for us to trigger it.

panel/io/resources.py
```diff
@@ -122,13 +122,8 @@ def conffilter(value):
...
 
-if mimetypes.types_map.get('.js') == 'text/plain':
-    logger.warn(
-        "Mimetype declaration for .js files was set to 'text/plain'. "
-        "This will interefere with serving JS modules. Ensure your system "
-        "does not override the mimetype declaration and/or contact your "
-        "system administrator. "
-    )
+mimetypes.init()
+if mimetypes.types_map.get('.js') != "application/javascript":
```
MarcSkovMadsen, line 126: I removed the logger.warn for many reasons including 1) logger.warn is deprecated and 2) it only gives the user problems.

- A newbie would not know what a `Mimetype` declaration or JS Modules is. In my experience they definitely don't like to contact a system administrator.
- An experience, grumpie man like me would have to spend time on explaining this warning to my colleagues. And if I have the energy spend time on reporting this to IT and spend weeks on explaining to them why they have to change the windows registry for 7000 people.

Why warn when everything will work after the code has been executed?

MarcSkovMadsen:
It might be simpler and faster to not `.init` and check. But just run `mimetypes.add_type("application/javascript", ".js")` always. Even though the name is `add_type` if is really and `add_or_update_type` function.

panel/io/resources.py
philippjfr: This is true but I also worry that it might override someone else's overrides, from the docs:

>If files is None the internal data structure is completely rebuilt to its initial default value. This is a stable operation and will produce the same results when called multiple times.

panel/io/resources.py
philippjfr: Agreed.

philippjfr:
>It might be simpler and faster to not .init and check. But just run mimetypes.add_type("application/javascript", ".js") always. Even though the name is add_type if is really an add_or_update_type function.

+1 this makes sense to me.

MarcSkovMadsen:
Thanks @philippjfr . I've just pushed the simplification and the PR should be ready.

PR #4739, opened 2023-04-29
Improve vizzu reference

MarcSkovMadsen:
- The links to the vizzu docs where broken. I fixed them.
- When testing in the dev docs, there was not a lot of interactivity to try out. I added another cell to enable the user to animate from circle to area and back.

### DIFF
--- a/examples/reference/panes/Vizzu.ipynb
+++ b/examples/reference/panes/Vizzu.ipynb
@@ -3,7 +3,9 @@
   {
    "cell_type": "code",
    "execution_count": null,
-   "metadata": {},
+   "metadata": {
+    "tags": []
+   },
    "outputs": [],
    "source": [
     "import numpy as np\n",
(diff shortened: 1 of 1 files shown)


philippjfr:
>Why a separate Panel implementation? Why not just support Ipyvizzu? Whats the benefit?

I consider the ipywidgets support to be extremely inefficient as it requires a multi-MB bundle so a native implementation is almost always preferable. 

>How do I stream and patch? There is no example

We should have a general `stream` and `patch` guide, because it all works the same (I think you wrote most of that code for Tabulator).

philippjfr:
You do hit on a good point though, it was a conscious decision not to try to re-create the ipyvizzu API. I did that because my feeling the ipyvizzu API doesn't really fit the Panel model, I want to lean heavily into the reactive APIs and binding parameters because it allows users to use existing APIs (i.e. pandas) to achieve their tasks. ipyvizzu introduced their own APIs for data manipulation and I'd rather users lean into pandas + bind.

MarcSkovMadsen:
I understand But it also means that ipyvizzu users would not see that Panel offers anything for them, we cannot rely ipyvizzu community examples and we will have to keep things updated towards vizzu.

droumis:
re: lambda, except in the simplest of cases, I think they negatively impact readability. I'd prefer we not use them unless there's a really good reason and they are accompanied by a clear explanation

philippjfr:
I'll merge, can follow up by removing the lambda separately.

PR #4741, opened 2023-04-29
Add Design and Theme explanation documentation

philippjfr:
@MarcSkovMadsen This is definitely far from perfect but I'm hoping it'll give you an idea of how to contribute to improve the Fast designs (and restore some of the styling that didn't get ported).


doc/explanation/styling/design.md
```diff
@@ -0,0 +1,107 @@
+# Designs and Theming
+
+:::{versionadded} 1.0.0
+Designs and theming allow controlling the look and feel of individual components or your entire application.
```
MarcSkovMadsen, line 4: Maybe connect to *template* too. For me I want to correct a template, so I'm looking for an explanation for that.

So I probably need to be taught that the starting point is a *design*. A design can have different *themes*. And a template can use the design (and theme).

Maybe even visualize how these things are connect via a graph. Or via an app wrapped in a nice template that uses some desing.

doc/explanation/styling/design.md
```diff
@@ -0,0 +1,107 @@
...
+:::
+
+## `Design` class
+
+The `Design` class has the ability to apply styling and theming to components via three mechanisms:
+
+1. Loading external JS and CSS `resources`.
+2. Applying `modifiers` that directly override (or combine with) component parameter settings (including CSS stylesheets).
+3.
```
MarcSkovMadsen, line 13: 3.?

doc/explanation/styling/design.md
```diff
@@ -0,0 +1,107 @@
...
+3. Apply a `css` with specific CSS overrides for that theme.
+4. Apply its own set of `modifiers`.
+
+### Example: Bootstrap
+
+Let us work through the definition of the `Bootstrap` design step-by-step.
+
+#### Resources
+
+First of all, the `Bootstrap` design defines CSS and JS `resources` which load the bootstrap5 library:
```
MarcSkovMadsen, line 28: Here it would be helpful with a link to the actual bootstrap Design class to make this very explicit.

doc/explanation/styling/design.md
```diff
@@ -0,0 +1,107 @@
...
+```
+
+#### Modifiers
+
+Next we get to the `modifiers` which is where the meat of the styling definition lives:
+
+```python
+    modifiers = {
+        Accordion: {
+            'active_header_background': 'var(--bs-surface-bg)'
```
MarcSkovMadsen, line 48: We should make it clear that we are changing parameters of the component. Not directly some stylesheet. I thought we did the latter because the value is ? var(--bs-surface-bg)` which looks like we are changing css (which we are, but via the parameter).

doc/explanation/styling/design.md
```diff
@@ -0,0 +1,107 @@
...
+        }
+		...
+        Tabulator: {
+            'theme': 'bootstrap4'
+        },
+        Viewable: {
+            'stylesheets': [Inherit, f'{CDN_DIST}bundled/theme/bootstrap.css']
```
MarcSkovMadsen, line 55: Would I put the Bokeh plots css fixes into fast.css or into some fast_holoviews.css file to use in the `stylesheets` for the `HoloViews` pane?

doc/explanation/styling/design.md
```diff
@@ -0,0 +1,107 @@
...
+Here we define specific parameters to override or combine with, e.g. we set the `active_header_background` of an `Accordion` to inherit from the `--bs-surface-bg` CSS variable. In this way we can control settings that are not controlled by CSS. However, most crucially we also declare that all `Viewable` components should inherit a bundled stylesheet: `bundled/theme/bootstrap.css`.
+
+The `bootstrap.css` file includes styling overrides for many different components and will be loaded by all `Viewable` components.
+
+#### CSS stylesheet
+
+All these design stylesheets open with CSS variable definitions which map the global color definitions to specific variables used by a particular design system:
+
+```css
+:host, :root {
```
MarcSkovMadsen, line 69: You will need to explain what :host and :root refer to.

doc/explanation/styling/design.md
```diff
@@ -0,0 +1,107 @@
...
+    --bs-primary-color: var(--jp-ui-font-color0, var(--panel-on-primary-color));
```
MarcSkovMadsen, line 70: Probably explain that you are trying to make the design work in both jupyter notebooks (via --jp) and server (via --panel).

doc/explanation/styling/design.md
```diff
@@ -0,0 +1,107 @@
...
+
+```python
+    _themes = {
+        'dark': BootstrapDarkTheme,
+        'default': BootstrapDefaultTheme,
+    }
+```
+
+Currently only `default` (light) and `dark` themes are supported. The implementation of these themes subclasses from the `DefaultTheme` and `DarkTheme` base classes and applies specific overrides.
+
```
MarcSkovMadsen, line 98: What about the css for the template. Somethings like the *card* background and the *reset grid button* are not viewables. Where do you put that css?

MarcSkovMadsen:
Its a good starting point for me. I added my questions and comments. I hope we can use that to improve the PR.

doc/explanation/styling/design.md
philippjfr: >Would I put the Bokeh plots css fixes into fast.css or into some fast_holoviews.css file to use in the stylesheets for the HoloViews pane?

For now I put everything in one file but as this grows splitting them up makes sense.

doc/explanation/styling/design.md
philippjfr: Anything related to the template styling lives in, `templates/<template-name>/<template-name>.css`, e.g. for fast that's `templates/fast/fast.css`.

philippjfr:
Indeed, only a start, all good points in your review.

doc/explanation/styling/design.md
```diff
@@ -0,0 +1,107 @@
...
+## Developing an internal theme
+
+If you want to help contribute improvements to a `Design` or `Theme` simply edit the CSS definitions in `panel/theme/css` and periodically re-bundle the resources with:
+
+```bash
+panel bundle --themes --only-local
+```
+
+On UNIX you can simply use `watch -n 5 panel bundle --themes --only-local` to re-bundle these resources every few seconds.
```
MarcSkovMadsen, line 107: But as far as I can see the panel server does not provides the updates `fast.css` if I (hard) refresh the browser (on windows). As far as I can see, I need to restart the server. Is that correct @philippjfr ?

MarcSkovMadsen:
On thing I need to do is to change the `border-left-color` below to `var(--accent-foreground-active)`. I've been trying for an hour now, and I simply don't know how to do this. How can I do this @philippjfr , @mattpap ?

image

philippjfr:
Any component in the shadow DOM cannot be styled from the template CSS file. As laid out in the document those rules have to go in panel/theme/css/fast.css

doc/explanation/styling/design.md
philippjfr: I can get it working with hard reload in the browser.

mattpap:
`--active-tool-highlight` is a local CSS variable, thus the only way to override is to add a stylesheet to the component (or a class of components). To theme a component (i.e. to override CSS variables from global scope), component's CSS variables have to be set up like this:
```css
--active-tool-highlight: var(--theme-active-tool-highlight, #26aae1);
```
where `theme` suffix is discretionary (i.e. whatever as long as there's distinction between local and global variables). The global variable by default is unset, which makes the local variable to use the literal default value. Note you can chain multiple CSS variables for more advanced themes. Most bokehjs' CSS variables are local, and only a few so far as set for theming. To override the local variable in a theme, one would use:
```css
:root {
  --theme-active-tool-highlight: red;  /* or var(--another-themed-variable) */
}
```
Note that `:root` pseudo selector only works in the global scope and usually is equivalent`html` selector, just has higher specificity (pretty useful to avoid `!important` in some cases). `:host` pseudo selector only works in shadow DOM and refers to the shadow root element. `:root` and `:host` should not be used together.

philippjfr:
@mattpap Thanks for the explanation, unfortunately in this case I can't see how we'd even go about injecting anything into the shadow DOM of the tool button, as far as I can tell the tool button does not have a corresponding Python model that we could inject the stylesheets on.

MarcSkovMadsen:
Is it correctly understood that currently we cannot style the tooltip and toolbar buttons?

philippjfr:
Waiting to hear back from @mattpap but to me it looks like that. The tool buttons are internal models so can't easily be controlled from Python. IMO all tool related CSS variables should therefore be set at the toolbar level not at the tool button level.

mattpap:
> The tool buttons are internal models so can't easily be controlled from Python. 

Each button is an individual component and then toolbar is another component. Tool buttons are exposed in bokehjs, but not it Python, because I couldn't settle on a stable API before 3.1 release.

philippjfr:
Yes, absolutely. That said I'm not entirely sure what shape that would take. I don't really see how we can expose `ToolButton` at the Python level or if that is even desirable. I'll let @mattpap suggest an approach but maybe let's start by making an issue that simply outlines the problem which is that the `--active-tool-highlight` default on `ToolButton` cannot currently be overridden.

philippjfr:
I haven't addressed all review comments but this is already much better than nothing.

MarcSkovMadsen:
> I haven't addressed all review comments but this is already much better than nothing.

I feel the same for the `ReactiveHTML` docs PR 😄

PR #4751, opened 2023-04-30
Unify pyodide dependencies across JupyterLite, convert and Sphinx docs

philippjfr:
Panel examples and docs can be consumed in one of three ways:

1. JupyterLite
2. `panel convert` apps
3. Sphinx Documentation using the pyodide extension

To ensure that dependencies are correctly specified across all three without duplicating work their dependencies are now all determined by the `doc/pyodide_dependencies.json` file.


panel/io/convert.py
```diff
@@ -259,7 +259,7 @@ def script_to_html(
     if any('holoviews' in req for req in reqs):
         reqs = ['holoviews>=1.16.0b5' if 'holoviews' else req in req for req in reqs]
     elif any('hvplot' in req for req in reqs):
-        reqs.append('holoviews>=1.16.0b5')
+        reqs.insert(2, 'holoviews>=1.16.0b5')
```
Hoxbro, line 262:
```suggestion
        reqs.insert(2, 'holoviews>=1.16.0a7')
```

No beta release of holoviews 1.16

philippjfr: Ah, thanks :)

PR #4757, opened 2023-05-02
Minor changes to prepare for Param 2.0

maximlt:
This PR fixes calls to Param's main warning, as in Param 2.0 the `warning` method is going to be removed from the Parameterized namespace and will only be available in the `.param` namespace.

This PR also removes an unused kwarg passed in a test to the `HoloViews` pane, spotted by Simon in https://github.com/holoviz/param/pull/729.


philippjfr (approved):
Looks good, thanks!

Issue #4695, opened 2023-04-26
Release checklist for Version 1.0


philippjfr:
- [ ] JupyterLite
  - [ ] Ensure automated test by Marc can run successfully such that the right top cells (green/ red) can be set. [Marc, need help from @philippjfr to get started due to major refactor]
  - [x] #4530 [Marc]
  - [x] Ensure all examples run (or have banner warning of issues) [Marc]
  - [x] Add links throughout documentation
    - [x] Component gallery
    - [x] App Gallery
    - [x] How-to guides
- [x] Gallery
  - [x] Re-run every example and fix any remaining issues
  - [x] Fix links (currently each card adds a link to the source notebook but since the entire card is clickable you can't navigate there)
- [ ] Documentation
  - [x] Fix Styling Issues #4647
  - [x] Reorganize Navigation #4681
  - [x] Ensure homepage not outdated #4214 
  - [x] Ensure all pages can be run in pyodide
  - [x] Ensure 1.0 api is used and not some older api. For example pn.bind, pn.state.served (@maximlt )
  - [x] Migration guide complete
  - [ ] Getting Started
    - [ ] Updated 2-minute getting started
    - [ ] Rewrite build-an-app guide based on latest reactive APIs
  - How-to
    - Prepare to develop
      - [x] how_to/display
        - [x] JupyterLab Preview (@maximlt)
        - [x] Break this out into notebook and editor/IDE sections (@droumis)
      - [x] how_to/apis (@droumis)
        - [x] Move this to background
    - Build Apps
        - [x] how_to/interactivity (@Hoxbro)
          - [x] Rename how_to/reactive to how_to/interactivity and make it third item in Build Apps section
          - [x] Add section on pn.bind as component (put param docs in "Related Resources")
          - [x] Add section on binding parameters and bound function on components `pn.pane.Markdown(pn.bind(...))` 
          - [x] Add section on binding `hvplot.interactive` (if you need inspiration look at Portfolio Optimizer)
        - [x] Remove how_to/interact (@droumis )
        - [x] how_to/templates/ (@maximlt)
          - [x] Move this up from "Prepare to share" 
          - [x] Update screenshots
    - Create specialized UIs (Rename this to something like "Specialized UIs and APIs")
      - how_to/links (move this here) (@droumis )
      - how_to/param (move this here)  (@droumis )
      - Find way to add multi-page app guide
    - Prepare to share
      - [ ] how_to/performance
        - [ ] defer_load (@philippjfr)
[...]

PR #4759, opened 2023-05-02
Reorganize API guides

droumis:
- moves API pros and cons (context) to the explanation from how-to
- removes interact API how-to guides and mention of it in the others
- moves declarative and callbacks API to the specialized how-to section
- rename Param API to Declarative API

- addresses part of #4695


philippjfr (approved):
Thanks, reorganization all looks good. Will merge to be part of the RC2 docs build. Can iterate from there if needed.

PR #4760, opened 2023-05-02
split dev in nb vs editor into diff subsections

droumis:
- addresses part of #4695

- draft, waiting on #4759 to be merged first


philippjfr (approved):
Perfect!

PR #4763, opened 2023-05-03
Update async.md

Hoxbro:
Removed `aiohttp` for pyodide_dependencies.json and removed warning about python 3.8.


philippjfr:
Wait, actually I'm confused, doesn't that document actually require aiohttp?

Hoxbro:
To be honest, I’m also confused about what I am supposed to do. Adding a dependency which does not work with pyodide.

philippjfr:
Suspect we either need to find an alternative to aiohttp or disable pyodide on this page for now.

MarcSkovMadsen:
Please align whatever you do with Panelite. There we displayed a red alert and link to the github issue.

philippjfr:
I do not think docs pages should link to GitHub issues and I also don't think they should show red alerts. If we can't get it to work we should disable it, not tell users about all the ways it is broken.

Issue #4735, opened 2023-04-29
Dev Documentation Issues


MarcSkovMadsen:
I've been asked on Discord to post all `dev` Docs  issues identified pre the 1.0 release in one Bug Report. So here goes.

## Migrated Issues

For details see

- [ ] App Thumbnails not looking nice (see below)
- [ ] Twitter not using new logo, tagline and description
- [ ] Discord Invite link is invalid (see below)
- [ ] #4732
- [ ] Old Jupyter Panel Preview icon shown
- [ ] CodeEditor Control not shown component page
- [ ] JS Linked Controls in reference docs not working
- [ ] VTK Interactive app demo app never loads (see below)
- [ ] Panelite version not in sync with docs (see below)
- [ ] Improve app sign buttons
- [x] #4734
- [x] #4733
- [x] #4731
- [x] #4730
- [x] #4729
- [x] #4728
- [x] Debugger issues (see below)
- [x] Favicon on demo apps needs update (see below)
- [x] source link in Gallery points to app not code. See https://twitter.com/gypsyvirus/status/1654212248737394688?s=46&t=QCeLnYcdhHHHwawHQ7y8xA

philippjfr:
Thank you!

philippjfr:
Most of these were easy fixes, not sure about the Bokeh one yet.

MarcSkovMadsen:
## Debugger Issues

When I go to https://pyviz-dev.github.io/panel/reference/widgets/Debugger.html, I see 

### 1 - Wird button

![image](https://user-images.githubusercontent.com/42288570/235829993-0f6bfbd1-893c-492e-850f-3af0d44f26b6.png)

The play symbol confuses me. Should it be there. 1 min later I realize its a collapsed accordion. I think either the button should be vertically alligned such that the header looks like a button to be clicked or that in the example the debugger should be expanded by default.

### 2 - Too little container

The example below is put in a very small container with a scrollbar which makes it not very useful

![image](https://user-images.githubusercontent.com/42288570/235830259-be42bb32-3acf-422a-b813-144036d4959c.png)

### 3 - Missing Icon

![image](https://user-images.githubusercontent.com/42288570/235830341-a9382f68-d352-4be4-a530-a714af71fab9.png)

MarcSkovMadsen:
## Favicon on demo apps needs update

For example https://panel-gallery-dev.pyviz.demo.anaconda.com/vtk_interactive

[...]

PR #4764, opened 2023-05-03
Various fixes for Debugger

philippjfr:
Partial fix for https://github.com/holoviz/panel/issues/4735

### DIFF
--- a/examples/reference/widgets/Debugger.ipynb
+++ b/examples/reference/widgets/Debugger.ipynb
@@ -3,33 +3,32 @@
   {
    "cell_type": "code",
    "execution_count": null,
+   "id": "b15a6cf0",
    "metadata": {},
    "outputs": [],
    "source": [
+    "import logging\n",
+    "\n",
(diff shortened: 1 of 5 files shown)


examples/reference/widgets/Debugger.ipynb
```diff
@@ -3,33 +3,32 @@
   {
    "cell_type": "code",
    "execution_count": null,
+   "id": "b15a6cf0",
```
maximlt, line 6: @Hoxbro do we allow `id`s now in cells metadata? Just to know.

philippjfr: We don't really use the shell script anymore, I rely exclusively on Simon's excellent pre-commit hook. But I also thought that removed ids.

maximlt: I still have to use it on `examples` as there are notebooks that need to be pushed with their cell output, unfortunately :)

PR #4770, opened 2023-05-04
Fix/fast design

MarcSkovMadsen:
Note: I'm using the command `panel bundle --themes --only-local` to bundle the template resources. This command is not documented anywhere and I'm in doubt whether its covered by running `python setup.py develop`.


panel/theme/fast.py
```diff
@@ -170,10 +170,10 @@ class FastWrapper(ReactiveHTML):
 DEFAULT_STYLE = FastStyle()
 
 DARK_STYLE = FastStyle(
-    background_color="#181818",
+    background_color="#181818", #242424
```
MarcSkovMadsen, line 173: The Background and luminance changes the fundamental background colors to the original from 0.14.4

image

I can see on the `Fast` homepage they don't recommend using #181818 anymore. They recommend using #242424 or others. If I get the time before the 1.0 release I would like to 1) Create a FastStyler widget that can interactively change the basic design settings of a Fast template 2) Experiment with and determine the best settings.

But I believe this change improves the style a lot and is a safe bet.

panel/theme/css/fast.css
```diff
@@ -24,9 +24,9 @@
 /* Separator */
 
 hr {
```
MarcSkovMadsen, line 26: Fixes the styling of the `Divider`.

philippjfr:
>This command is not documented anywhere and I'm in doubt whether its covered by running python setup.py develop.

It is but until we switch to hatch it only works by setting `SETUPTOOLS_ENABLE_FEATURES="legacy-editable"`.

panel/theme/fast.py
philippjfr: So which is it, should we switch to #242424? I think it looks better.

philippjfr: Hmm, actually I don't think that looks good. Setting #242424 makes it very hard to distinguish the background and the card.

philippjfr (approved):
Changes look good, thanks.

Issue #4385, opened 2023-02-03
ColorPicker issues on the branch 1.0


maximlt:
- [ ] Unnamed ColorPicker has an "inner" zero width
![image](https://user-images.githubusercontent.com/35924738/216478983-09f33d97-bf90-4575-aab2-e74c8f269478.png)

Might be an issue shared with other components?

philippjfr:
I've generally had to set some default width. For other components I've just set `param.Integer(default=300)` (which should all be reviewed and cleaned up). I'm not sure whether a colorpicker really needs that wide a default width.

maximlt:
300px wide feels a little to wide indeed. Apparently a width of 56px makes the colored part square.

![image](https://user-images.githubusercontent.com/35924738/216985968-bc0c3829-e5f0-4f31-a29a-7a714ba8e833.png)

An alternative would be not to have any padding around the colored part. If so, I guess such a change would make more sense if applied directly in Bokeh.

 > For other components I've just set param.Integer(default=300) (which should all be reviewed and cleaned up).
 
 Yes indeed reviewing that seems important, I'll open an issue to keep track of it, too easy to forget! Would it make sense to make that width configurable at a global level?

PR #4788, opened 2023-05-05
Minor fixes for components

philippjfr:
- Remove unnecessary width/height handling on `Matplotlib`
- Fixes https://github.com/holoviz/panel/issues/4385

### DIFF
--- a/panel/io/jupyter_executor.py
+++ b/panel/io/jupyter_executor.py
@@ -77,7 +77,8 @@ class PanelExecutor(WSHandler):
         try:
             self.session = self._create_server_session()
             self.connection = ServerConnection(self.protocol, self, None, self.session)
-        except Exception:
+        except Exception as e:
+            self.exception = e
             self.session = None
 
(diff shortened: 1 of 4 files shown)


panel/io/jupyter_executor.py
```diff
@@ -77,7 +77,8 @@ def __init__(self, path, token, root_url):
         try:
             self.session = self._create_server_session()
             self.connection = ServerConnection(self.protocol, self, None, self.session)
-        except Exception:
+        except Exception as e:
+            self.exception = e
```
Hoxbro, line 81: Nit: I would have `str(e)´ here to not have the error object saved.

philippjfr: Agreed, doesn't matter a ton in this case because if there's an exception the kernel this is in will get terminated pretty much immediately.

Hoxbro: Yep, that is why it is a nit.

PR #4793, opened 2023-05-06
Various theming fixes

philippjfr:
Various small tweaks to Bootstrap, Material and Fast designs and themes.

### DIFF
--- a/panel/dist/bundled/datatabulator/tabulator-tables@5.4.4/dist/css/tabulator_fast.min.css
+++ b/panel/dist/bundled/datatabulator/tabulator-tables@5.4.4/dist/css/tabulator_fast.min.css
@@ -20,7 +20,7 @@
     position: relative;
     box-sizing: border-box;
     width: 100%;
-    border-bottom: 1px solid var(--neutral-stroke-divider-rest);
+    border-bottom: 2px solid var(--accent-fill-rest);
     background-color: var(--fill-color);
     color: var(--neutral-foreground-rest);
     font-weight: bold;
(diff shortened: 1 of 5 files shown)


MarcSkovMadsen:
For the `Fast` buttons you will have to choose if you select to 0.14.4 design which exposes the different types of Fast styles or a Bootstrap like style where buttons in different colors are exposed. I've seen users ask for both.

For me the original 0.14.4 design is most valuable. I don't think the bootstrap colors goes well with the fast design.

## 0.14.4. Design

### Default

image

### Primary

image

### Success

image

I think this style works particularly well when applied to the `RadioButtonGroup`. So it would be a big loss for me if not available.

### Warning

image

(This has nothing to do with a warning. But it exposes a style of button exposed by Fast).

### Danger

image

(This has nothing to do with a warning. But it exposes the link button exposed by Fast).

### Light

image

philippjfr:
That overloading of the names is pretty weird. I think we should add another parameter that allows switching between outline, accent and stealth styling.

philippjfr:
For now I've got solid and outline:

button

philippjfr:
And for the RadioButtonGroup:

Screen Shot 2023-05-07 at 12 09 49

MarcSkovMadsen:
> For now I've got solid and outline:
> 
> 
> 
> button
> 
> 

If "primary" is using the accent color, then that looks great to me. If it does not easy way to apply the accent color would be appreciated.

philippjfr:
It does indeed.

PR #4798, opened 2023-05-07
Implement reveal.js based SlideTemplate

philippjfr:
Working on a talk and decided that I no longer wanted to use [RISE](https://rise.readthedocs.io/en/stable/), so I decided to implement a slides template for Panel instead.


philippjfr:
Will merge and add docs later. Want to get a last RC out.

PR #4809, opened 2023-05-08
Update set_param to update and fix a related bug

Hoxbro:
Found a small bug in `_recursive_update` by running this:

``` python
import pandas as pd
import panel as pn

pn.extension("tabulator")

n = pn.widgets.IntSlider(name="Frequency", start=0, end=10, value=2)


@pn.depends(n)
def f(n):
    a = pn.widgets.Tabulator(pd.DataFrame(range(n)))
    return pn.Row(a)


pn.Column(n, pn.Row(f)).servable()
```

And was informed by @maximlt that `set_param` is deprecated so I updated it to `update`.


philippjfr (approved):
Thanks!

PR #4811, opened 2023-05-08
Remove holoviews version pins

philippjfr:
No longer need a lower bound version pin on holoviews and ipywidget_bokeh.

To be merged when holoviews 1.16.0 is released.

### DIFF
--- a/scripts/panelite/generate_panelite_content.py
+++ b/scripts/panelite/generate_panelite_content.py
@@ -38,7 +38,7 @@ def _get_dependencies(nbpath: pathlib.Path):
         return []
     for name, min_version in MINIMUM_VERSIONS.items():
         if any(name in req for req in dependencies):
-            deps = [f'{name}>={min_version}' if name in req else req for req in dependencies]
+            dependencies = [f'{name}>={min_version}' if name in req else req for req in dependencies]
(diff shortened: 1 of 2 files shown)


Hoxbro:
Right now, this page is failing: https://pyviz-dev.github.io/panel/how_to/interactivity/hvplot_interactive.html

Looking at the console, it seems to have added two `>=>=`, so I think it will stop warning with this PR. Though, there is something wrong _somewhere_.

image

philippjfr:
@Hoxbro Could you merge my laptop won't allow me to for some reason.

Hoxbro:
`deps` in pyodide should be renamed to `dependencies`. I can't add a suggestion right now because of Github.

scripts/panelite/generate_panelite_content.py
```diff
@@ -40,8 +39,6 @@ def _get_dependencies(nbpath: pathlib.Path):
     for name, min_version in MINIMUM_VERSIONS.items():
         if any(name in req for req in dependencies):
             deps = [f'{name}>={min_version}' if name in req else req for req in dependencies]
```
Hoxbro, line 41:
```suggestion
            dependencies = [f'{name}>={min_version}' if name in req else req for req in dependencies]
```

PR #4826, opened 2023-05-12
Fix Jupyter Panel Preview in Windows

### LINKED ISSUE #4819: Jupyter Panel Preview not working on `main`
I am on the main branch. I would like to take and contribute a few screenshots I think are missing in the documentation.

I would like one using the Panel Jupyter Preview. But it does not work for me. I'm on the `main` branch of Panel. I'm running inside my `panel_dev` python environment. And that is also where Jupyterlab is installed and started from.

image

```bash
404 : Not Found
You are requesting a page that does not exist!
```

My version of Jupyterlab is 3.6.3.

Hoxbro:
Fixes #4819

The path string was not raw, which meant on Windows, the path would give a Unicode error. I also added more logging if the msg_type is an error.

image

### DIFF
--- a/panel/io/jupyter_server_extension.py
+++ b/panel/io/jupyter_server_extension.py
@@ -111,7 +111,7 @@ import os
 import pathlib
 import sys
 
-app = '{{ path }}'
+app = r'{{ path }}'
 os.chdir(str(pathlib.Path(app).parent))
 sys.path = [os.getcwd()] + sys.path[1:]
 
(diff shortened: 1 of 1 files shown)


MarcSkovMadsen:
I'm so happy for

- the fix
- that it was not me speaking out very loudly about a problem that was on my side 😄 

Thanks a lot @Hoxbro . My hero 🦸

panel/io/jupyter_server_extension.py
```diff
@@ -186,6 +186,8 @@ async def _get_info(self, msg_id, timeout=KERNEL_TIMEOUT):
                 comm_id = msg['content']['comm_id']
             elif msg_type == 'stream' and msg['content']['name'] == 'stderr':
                 logger.error(msg['content']['text'])
+            elif msg_type == "error":
+                logger.error(msg['content'])
```
philippjfr, line 190:
```suggestion
                logger.error(msg['content']['traceback'])
```

philippjfr:
Thank you both, glad we were able to track this down!

MarcSkovMadsen:
I tried the fix locally and can confirm it works for me

image

image

PR #4849, opened 2023-05-15
Adds support to Audio for more numpy dtypes as well as Torch objects

MarcSkovMadsen:


panel/pane/media.py
```diff
@@ -143,8 +165,11 @@ class Audio(_MediaBase):
 
     @classmethod
     def applies(cls, obj: Any) -> float | bool | None:
+        if isinstance(obj, TorchLike):
+            obj = obj.numpy()
```
philippjfr, line 169: This scares me as it can potentially make in-memory copies of huge array, just to check if their dtype matches.

MarcSkovMadsen: 😁i Think i can ind a better implementation. We should just check if it Torchlike and then the dimension

MarcSkovMadsen: I believe your concern has now been adressed.

MarcSkovMadsen:
Ready for review. I don't plan to do more before I get feedback.

----

On my side it does not fail the pre-commit. Its strange?

image

panel/pane/media.py
```diff
@@ -19,6 +19,25 @@
...
+        numpy_attr = getattr(instance, "numpy", "")
+        dim_attr = getattr(instance, "dim", "")
+        return (
+            bool(numpy_attr) and
+            callable(numpy_attr) and
+            callable(dim_attr) and
+            hasattr(instance, "dtype")
+        )
+
+# When support for Python 3.7 is dropped, Torchlike should be implemented as a typing.Protocol
```
philippjfr, line 34:
```suggestion
# When support for Python 3.7 is dropped, Torchlike should be implemented as a typing.Protocol
```

Panel 1.0 does not support 3.7

philippjfr: Honestly don't know enough about it to say. I'll merge because I'm about to tag 1.0 but please follow up on that.

Issue #4869, opened 2023-05-17
Add support for header_max_width


ahuang11:
I would like the header to be aligned with the main body, similar to this
https://developer.mozilla.org/en-US/docs/Web/CSS/clamp
<img width="3008" alt="image" src="https://github.com/holoviz/panel/assets/15331990/d192aa2e-1191-4588-a6ec-9fc7be1121c0">

At the moment, the title is at the far top left and the buttons are on the far top right.
<img width="3005" alt="image" src="https://github.com/holoviz/panel/assets/15331990/12a0bde1-8822-421c-a516-06447a129573">

If accepted, I'm happy to contribute a PR.

philippjfr:
Makes sense to me, would love to see a PR.

PR #4871, opened 2023-05-18
Start adding max header width

ahuang11:
Closes https://github.com/holoviz/panel/issues/4869

Start adding max_header_width, but before continuing, wanted to make sure if this is the desired way of injecting styles inside jinja HTML templates instead of css.

I also removed the shadows (or glow)
image

Because it looked weird.
image

I guess I could keep the shadow if I created another div?

### DIFF
--- a/panel/template/fast/list/fast_list_template.html
+++ b/panel/template/fast/list/fast_list_template.html
@@ -84,7 +84,7 @@ document.addEventListener('DOMContentLoaded', (event) => {
 <fast-design-system-provider id="body-design-provider" use-defaults>
   <div id="container">
     <fast-design-system-provider id="header-design-provider">
-      <nav id="header" {{ 'class="shadow"' if style.shadow else '' }} >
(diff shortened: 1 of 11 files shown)


panel/template/fast/list/fast_list_template.html
```diff
@@ -84,7 +84,7 @@
 <fast-design-system-provider id="body-design-provider" use-defaults>
   <div id="container">
     <fast-design-system-provider id="header-design-provider">
-      <nav id="header" {{ 'class="shadow"' if style.shadow else '' }} >
+      <nav id="header" {% if header_max_width %}style="margin-left: auto; margin-right: auto; max-width: {{ header_max_width }}"{% endif %} >
```
philippjfr, line 87: Would indeed prefer a CSS class.

ahuang11: Should I also migrate the `main_max_width` or perhaps just header_max_width?

ahuang11:
Okay, I couldn't add it to an actual .css file, but I was able to add it to styles.

ahuang11:
image

image

image

image

image

image

image

ahuang11:
Very stale at this point.

PR #4873, opened 2023-05-18
Adding description to Button, CheckButtonGroup, and RadioButtonGroup

Hoxbro:
Mostly a proof of concept of having tooltips for buttons. This should maybe be upstreamed to Bokeh itself.

https://github.com/holoviz/panel/assets/19758978/12fef0ea-3732-4599-a0c1-6e93af55f971

``` python
import panel as pn
from bokeh.models import Tooltip

pn.widgets.Button(
    name="Button", button_type="primary", width=100, description="default"
).servable()

pn.widgets.Button(name="Button", button_type="primary", width=100).servable()

tooltip = Tooltip(content="Example", position="bottom")
pn.widgets.Button(
    name="Button", button_type="primary", width=100, description=tooltip
).servable()

```


maximlt:
The less Typescript we have the better! Definitely voting for having this in Bokeh :)

philippjfr:
Indeed, let's make sure we upstream this soon. Will merge for now.

PR #4875, opened 2023-05-18
Replace user guide links

droumis:
This is in progress, there are many more to fix. :(

### DIFF
--- a/doc/how_to/state/url.md
+++ b/doc/how_to/state/url.md
@@ -12,16 +12,16 @@ When starting a server session Panel will attach a `Location` component which ca
 
 ### Core
 
-* **``pathname``** (string): pathname part of the url, e.g. '/user_guide/Interact.html'.
+* **``pathname``** (string): pathname part of the url, e.g. '/how_to/layout/spacing.html'.
 * **``search``** (string): search part of the url e.g. '?color=blue'.
-* **``hash_``** (string): hash part of the url e.g. '#interact'.
+* **``hash_``** (string): hash part of the url e.g. '#margin-parameter'.
 * **``reload``** (bool): Whether or not to reload the page when the url is updated.
     - For independent apps this should be set to True.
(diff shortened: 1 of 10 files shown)


droumis:
@philippjfr, for updates to accessing/manipulating the url (how_to/state/url), I put the port number after the domain `.org`. Is that incorrect? Before it was after the file path `.html`

doc/how_to/state/url.md
```diff
@@ -12,16 +12,16 @@ When starting a server session Panel will attach a `Location` component which ca
...
-* **``hash_``** (string): hash part of the url e.g. '#interact'.
+* **``hash_``** (string): hash part of the url e.g. '#margin-parameter'.
 * **``reload``** (bool): Whether or not to reload the page when the url is updated.
     - For independent apps this should be set to True.
     - For integrated or single page apps this should be set to False.
 
 ### Readonly
 
-* **``href``** (string): The full url, e.g. 'https://panel.holoviz.org/user_guide/Interact.html:80?color=blue#interact'.
+* **``href``** (string): The full url, e.g. 'https://panel.holoviz.org:443/how_to/layout/spacing.html?color=blue#margin-parameter'.
```
philippjfr, line 24: Yes, port after hostname, although in this case I'd just leave it out

```suggestion
* **``href``** (string): The full url, e.g. 'https://panel.holoviz.org/how_to/layout/spacing.html?color=blue#margin-parameter'.
```

philippjfr:
Note I made a lot of fixes here: https://github.com/holoviz/panel/pull/4872

doc/how_to/state/url.md
droumis: right, in the default port case, one would typically leave it out, but we want to demonstrate specifying/accessing the port, so I think leave it in here

philippjfr (approved):
Thanks!

PR #4882, opened 2023-05-18
Update reactive example

Hoxbro:
Currently, it is not working: https://panel.holoviz.org/explanation/components/components_custom.html#child-templates

Before: 
image

After:
image

Changing: `${option}` to `{{option}}` does the trick, as I don't completely understand why I have removed the mention of it in the text.


philippjfr:
Thanks, this indeed slipped through the cracks. We should fix the actual issue though.

philippjfr:
Okay, seems like `<select>` and `<option>` nodes just don't like rendering things that are inside the shadow DOM. Best approach is writing a better example.

philippjfr:
Changed the example in https://github.com/holoviz/panel/pull/4884.

Issue #4887, opened 2023-05-18
Tooltip issues


jbednar:
The new tooltips on widgets in Panel 1.0 are great! However, the UX isn't quite right, because hovering over the (?) takes so long that I reflexively click on the (?), which then stops the hover text from appearing. I think the delay should be maybe half what it is, and clicking on it should make the tooltip first appear, then (if clicked again) disappear, rather than suppress it silently without ever having shown it. That's probably a Bokeh issue rather than Panel, but maybe Panel sets the defaults.

Also, I think Panel or Param should strip out leading spaces that appear either on all lines of the doc or on all but the first line, so that docstrings can be formatted naturally in the code but also display naturally in the UI:

<img width="691" alt="image" src="https://github.com/holoviz/panel/assets/1695496/1c557c03-e9d1-449b-960a-83b64febd58d">

philippjfr:
Most of this is a request for Bokeh. I certainly agree that the timeout is too long and clicking shouldn't have any effect.

mattpap:
> I certainly agree that the timeout is too long and clicking shouldn't have any effect.

Note that plain text (string) description uses HTML's `title` attribute, so we don't have any control over how this behaves. If you want full control, then those descriptions have to be wrapped in an explicit `Tooltip(content=...)`. Thus in bokeh:
```py
input_widget.description = "string"
```
and
```py
input_widget.description = Tooltip(content="string")
```
are not equivalent.

philippjfr:
Maybe I don't know how to use it correctly, but this doesn't work for me at all:

```
input_widget.description = Tooltip(content="string")
```

No tooltip is ever shown.

philippjfr:
Nevermind, had to provide a position. I'll probably switch to explicit tooltips immediately.

philippjfr:
Was able to resolve that by setting `syncable=False` on the `Tooltip` model but still seems like an issue.

mattpap:
> Was able to resolve that by setting `syncable=False` on the Tooltip model but still seems like an issue.

Setting `syncable=False` in such cases is a generally good idea. The serialization error is a bug and I thought I already fixed it. Or maybe it was reported somewhere and got forgotten.

mattpap:
> Nevermind, had to provide a position. I'll probably switch to explicit tooltips immediately.

The API of tooltips is not well refined. Setting `position` is one of those issue, which I trip over every time I use `Tooltip`.

kcpevey:
[...]

PR #4890, opened 2023-05-19
Improve tooltip rendering

philippjfr:
- `lstrip` followed by dedent does not work as desired because it strips indentation on the first line which then breaks dedenting on subsequent lines.
- Switches from `title` property based tooltip to explicit HTML tooltip rendered using markdown

tooltips

Fix https://github.com/holoviz/panel/issues/4887

### DIFF
--- a/panel/widgets/base.py
+++ b/panel/widgets/base.py
@@ -117,7 +117,7 @@ class Widget(Reactive):
             html = parser.render(params['description'])
             params['description'] = Tooltip(
                 content=HTML(html), position='right',
-                stylesheets=[':host { white-space: initial; max-width: 250px; }'],
+                stylesheets=[':host { white-space: initial; max-width: 300px; }'],
                 syncable=False
             )
         return params
--- a/panel/widgets/base.py
+++ b/panel/widgets/base.py
@@ -14,7 +14,8 @@ from typing import (
 
 import param  # type: ignore
 
-from bokeh.models import ImportedStyleSheet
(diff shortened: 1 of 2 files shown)


Hoxbro:
I assume this comes from our docstring starting like this:
```
    param.String("HELLO", doc="""Line Zero
         Line One
         Line Two
         """
````

Maybe we should ignore `dedent` on line zero.

**Comment by philippjfr** [comment]
We should probably handle this, but personally I think writing the docstrings like this is bad:

```
    param.String("HELLO", doc="""Line Zero
         Line One
         Line Two
         """
```

I try to consistently use this style:

```

    align = Align(default='start', doc="""
        Whether the object should be aligned with the start, end or
        center of its container. If set as a tuple it will declare
        (vertical, horizontal) alignment.""")
```

**Comment by Hoxbro** [comment]
I can see that it is what is used in our repos. Though, we can't force someone writing a custom parameterized class to do this.  

I would maybe also use `.strip` instead of `.lstrip`.

File: panel/widgets/base.py [hunk 1]
```diff
@@ -96,13 +97,29 @@ def from_param(cls, parameter: param.Parameter, **params) -> Self:
         )
         return layout[0]
 
+    @property
+    def _linked_properties(self) -> Tuple[str]:
```
mattpap, line 101: The type should be `Tuple[str, ...]` unless the expectation is for a single value tuple.

PR #4895, opened 2023-05-19
Do not load mathjax bundle in notebook by default

philippjfr:
In the notebook we were always loading the (huge) bokeh-mathjax bundle even though we already told users that they would have to explicitly request the bundle (with `pn.extension('mathjax')`). This change will necessitate HoloViews to add a flag to load the bundle.

### DIFF
--- a/panel/io/resources.py
+++ b/panel/io/resources.py
@@ -346,10 +346,8 @@ def bundle_resources(roots, resources, notebook=False, reloading=False, enable_m
     css_files = []
     css_raw = []
 
-    if enable_mathjax == True:
-        use_mathjax = True
-    elif not enable_mathjax:
-        use_mathjax = False
+    if isinstance(enable_mathjax, bool):
+        use_mathjax = enable_mathjax
     else:
         use_mathjax = (_use_mathjax(roots) or 'mathjax' in ext._loaded_extensions) if roots else False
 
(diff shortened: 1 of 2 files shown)


panel/io/resources.py
```diff
@@ -347,21 +346,19 @@ def bundle_resources(roots, resources, notebook=False, reloading=False):
     css_files = []
     css_raw = []
 
-    use_mathjax = (_use_mathjax(roots) or 'mathjax' in ext._loaded_extensions) if roots else True
+    if enable_mathjax == True:
+        use_mathjax = True
+    elif not enable_mathjax:
+        use_mathjax = False
```
philippjfr, line 352:
```suggestion
    if isinstance(enable_mathjax, bool):
        use_mathjax = enable_mathjax
```

panel/io/resources.py
```diff
@@ -347,21 +346,17 @@ def bundle_resources(roots, resources, notebook=False, reloading=False):
     css_files = []
     css_raw = []
 
-    use_mathjax = (_use_mathjax(roots) or 'mathjax' in ext._loaded_extensions) if roots else True
+    if isinstance(enable_mathjax, bool):
+        use_mathjax = enable_mathjax
+    else:
+        use_mathjax = (_use_mathjax(roots) or 'mathjax' in ext._loaded_extensions) if roots else False
```
philippjfr, line 352:
```suggestion
    elif roots:
        use_mathjax = _use_mathjax(roots) or 'mathjax' in ext._loaded_extensions
    else:
        use_mathjax = False     
```

philippjfr:
pre-commit.ci autofix

PR #4903, opened 2023-05-19
Make tests more robust

philippjfr:
- Switch to playwright `expect` API in more places
- Mark some tests as flaky
- Improve location comparison


Hoxbro:
I ran the last commit twice because I saw two errors. One is a terminal test, which kills the process with an exit code of 143. The second is  `test_tabulator_row_content`, already marked as flaky with `max_runs=3`, which we could bump to 5. 

My personal opinion is we should merge this as is (maybe with the bump) and have a look at the failing terminal test in another PR. 

https://github.com/holoviz/panel/actions/runs/5369608585/attempts/1
image
image

PR #4904, opened 2023-05-20
Update JPEG filetype from jpg to jpeg.

n3011:
Pillow library does not have the `_repr_jpg_` method. Recently the `_repr_jpeg_` method is added https://github.com/python-pillow/Pillow/pull/7135

### DIFF
--- a/panel/pane/image.py
+++ b/panel/pane/image.py
@@ -357,7 +357,7 @@ class JPG(ImageBase):
     ... )
     """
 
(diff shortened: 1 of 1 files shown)


philippjfr:
Thanks `_repr_jpeg_` is indeed the canonical form: https://ipython.readthedocs.io/en/stable/config/integrating.html#custom-methods

philippjfr (approved):
Thank you! Test failures are unrelated.

PR #4917, opened 2023-05-22
Fix dunder method in alert

### LINKED ISSUE #4914: Alerts show dunder methods as bold
image

```
import panel as pn

pn.extension()

def __test__():
    raise ValueError("Test")

pn.panel(__test__).servable()
```

Hoxbro:
Fixes #4914

image

### DIFF
--- a/panel/io/server.py
+++ b/panel/io/server.py
@@ -723,7 +723,7 @@ def modify_document(self, doc: 'Document'):
             # Serve error
             e_msg = str(e).replace('\033[1m', '<b>').replace('\033[0m', '</b>')
             Alert(
-                f'<b>{type(e).__name__}</b>: {e_msg}</br><pre style="overflow-y: auto">{tb}</pre>',
+                f'<b>{type(e).__name__}</b>: {e_msg}\n<pre style="overflow-y: auto">{tb}</pre>',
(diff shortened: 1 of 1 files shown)


philippjfr (approved):
Looks good!

panel/io/server.py
```diff
@@ -723,7 +723,7 @@ def handle_exception(handler, e):
             # Serve error
             e_msg = str(e).replace('\033[1m', '<b>').replace('\033[0m', '</b>')
             Alert(
-                f'<b>{type(e).__name__}</b>: {e_msg}</br><pre style="overflow-y: auto">{tb}</pre>',
+                f'<b>{type(e).__name__}</b>: {e_msg}\n<pre style="overflow-y: auto">{tb}</pre>',
```
mattpap, line 726: I presume this assumes that `Alert` always uses preformatted text. Otherwise newlines have no bearing on text flow. Perhaps a better fix would be to use `<br>` (no `/` at all; `<br>` is self closing).

Hoxbro: The string is passed through a Markdown parser.

mattpap: OK, so this is not HTML, it's actually Markdown. I would somehow indicate this, because I'm pretty sure the next reader of this code will be as confused as I was.

Issue #4910, opened 2023-05-22
Widgets does not show up correctly in VS Code


Hoxbro:
![image](https://github.com/holoviz/panel/assets/19758978/0416ae75-0500-479a-97a3-59354bdfc786)

``` python
import panel as pn
pn.extension()

pn.widgets.FloatInput()
```

(The theme I'm using is Dark Modern)

PR #4923, opened 2023-05-24
Ensure VSCode does not apply white background

philippjfr:
Fixes https://github.com/holoviz/panel/issues/4910

### DIFF
--- a/panel/_templates/jupyter.css
+++ b/panel/_templates/jupyter.css
@@ -7,6 +7,9 @@
 }
 
 /* Override VSCode background color */
+.cell-output-ipywidget-background:has(> .cell-output-ipywidget-background
+    > .lm-Widget
+    > *[data-root-id]),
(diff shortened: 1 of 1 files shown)


panel/_templates/jupyter.css
```diff
@@ -3,5 +3,13 @@
   box-sizing: border-box;
   font-family: var(--jp-ui-font-family);
   font-size: var(--jp-ui-font-size1);
-  color: var(--jp-ui-font-color1);
+  color: var(--vscode-editor-foreground, var(--jp-ui-font-color1));
+}
+
+/* Override VSCode background color */
+.cell-output-ipywidget-background:has(> .cell-output-ipywidget-background
```
mattpap, line 10: `:has()` pseudo selector is not available in Firefox yet. Given this change is related to vscode, then maybe this is not a concern.

Issue #4933, opened 2023-05-25
Static configuration not working on Tabulator object when created from param


mayonnaisecolouredbenz7:
#### ALL software version info
panel              1.0.2
param              1.13.0
holoviews          1.16.0
tornado            6.2
pyviz-comms        2.2.1

#### Description of expected behavior and the observed behavior
Attempting to disable header sort.  It works when creating a tabulator object normally as is shown in the documentation.
https://panel.holoviz.org/reference/widgets/Tabulator.html#static-configuration
Does not seem to work when created from a param object.

#### Complete, minimal, self-contained example code that reproduces the issue
Does not work:
```
import pandas as pd
import panel as pn
import param
pn.extension('tabulator')

class TestClass(param.Parameterized):
    df = pd.DataFrame({
    'int': [1, 2, 3],
    'float': [3.14, 6.28, 9.42],
        }, index=[1, 2, 3])
    
    param_df=param.DataFrame(df)

c=TestClass()

pn.widgets.Tabulator.from_param(c.param.param_df,configuration={
    'columnDefaults': {
        'headerSort': False,
    },})

```
Works:
```
df = pd.DataFrame({
    'int': [1, 2, 3],
    'float': [3.14, 6.28, 9.42],
}, index=[1, 2, 3])


pn.widgets.Tabulator(df,configuration={
    'columnDefaults': {
        'headerSort': False,
    },})
```
#### Stack traceback and/or browser JavaScript console output
None
#### Screenshots or screencasts of the bug in action
Param:
![image](https://github.com/holoviz/panel/assets/38597661/2dd3cb80-49d2-4dab-bb6f-ca43d15d4596)
Non Param:
![image](https://github.com/holoviz/panel/assets/38597661/5c00fb67-aa94-46e1-826b-2de4512429c5)

mayonnaisecolouredbenz7:
Work around that has worked for me in the mean time:
```
tab_obj = pn.widgets.Tabulator.from_param(...)
```
```
tab_obj._configuration["columnDefaults"] = {"headerSort": False}
```

PR #4935, opened 2023-05-25
Warn about widget parameters that could not be applied by Param

philippjfr:
Add warning for cases like https://github.com/holoviz/panel/issues/4933

### DIFF
--- a/panel/param.py
+++ b/panel/param.py
@@ -467,6 +467,13 @@ class Param(PaneBase):
             kw['description'] = textwrap.dedent(p_obj.doc).strip()
 
         # Update kwargs
+        ignored_kws = [repr(k) for k in kw_widget if k not in widget_class.param]
+        if ignored_kws:
+            self.param.warning(
+                f'Param pane was given unknown keyword argument(s) for {p_name!r} '
+                f'parameter with a widget of type {widget_class!r}. The following '
+                f'keyword arguments could not be applied: {", ".join(ignored_kws)}.'
+            )
         kw.update(kw_widget)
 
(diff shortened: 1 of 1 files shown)


panel/param.py
```diff
@@ -467,6 +467,13 @@ def widget(self, p_name):
             kw['description'] = textwrap.dedent(p_obj.doc).strip()
 
         # Update kwargs
+        ignored_kws = [repr(k) for k in kw_widget if k not in widget_class.param]
+        if ignored_kws:
+            self.param.warning(
+                f'Param pane was given unknown keyword argument(s) for {p_name!r} '
+                f'parameter with a widget of type {widget_class!r}. The following '
```
Hoxbro, line 474:
```suggestion
                f'parameter with a widget of type {widget_class.__name___!r}. The following '
```

Hoxbro (commented):
Is there a way to let `ignored_kws` be passed to the widget at instantiation and then let the widget handle it? Like passing them in here: 
https://github.com/holoviz/panel/blob/3b588a31d1af515ba8271a76ead7e9ad4b078dc2/panel/param.py#LL487C1-L487C44

philippjfr:
In future Param will error if you give it unexpected kwargs so no we can't just pass it in but for situations like this it may indeed be nice to have a non-strict mode.

Hoxbro:
I don't see a problem in erroring if the input is wrong. Passing in `configuration` into Tabulator would still work in future versions of Param.

philippjfr:
Okay, I agree. Can't make that change in a micro release though.

PR #4939, opened 2023-05-25
Fix to remove call to registered_names which no longer exists in bokeh

alexmilowski:
I'm really sure this is the correct change. It appears the IPyWidget is available like other things via Bokeh.Panel.IPyWidget 

The issue is the use of registered_names generates a lot of JavaScript exceptions when we use things like the Tabulator.

### DIFF
--- a/panel/_templates/doc_nb_js.js
+++ b/panel/_templates/doc_nb_js.js
@@ -35,7 +35,7 @@
   }
   function is_loaded(root) {
     var Bokeh = get_bokeh(root)
-    return (Bokeh != null && Bokeh.Panel !== undefined{% for reqs in requirements %} && ({% for req in reqs %}{% if loop.index0 > 0 %}||{% endif %} root['{{ req }}'] !== undefined{% endfor %}){% endfor %}{% if ipywidget %}&& Bokeh.Panel.IPyWidget !== undefined {% endif %})
+    return (Bokeh != null && Bokeh.Panel !== undefined{% for reqs in requirements %} && ({% for req in reqs %}{% if loop.index0 > 0 %}||{% endif %} root['{{ req }}'] !== undefined{% endfor %}){% endfor %}{% if ipywidget %}&& Bokeh.Models._known_models.has("ipywidgets_bokeh.widget.IPyWidget") !== undefined {% endif %})
   }
   if (is_loaded(root)) {
     embed_document(root);
--- a/panel/_templates/doc_nb_js.js
+++ b/panel/_templates/doc_nb_js.js
@@ -35,7 +35,7 @@
   }
   function is_loaded(root) {
     var Bokeh = get_bokeh(root)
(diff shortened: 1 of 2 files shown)


panel/_templates/doc_nb_js.js
```diff
@@ -35,7 +35,7 @@
   }
   function is_loaded(root) {
     var Bokeh = get_bokeh(root)
-    return (Bokeh != null && Bokeh.Panel !== undefined{% for reqs in requirements %} && ({% for req in reqs %}{% if loop.index0 > 0 %}||{% endif %} root['{{ req }}'] !== undefined{% endfor %}){% endfor %}{% if ipywidget %}&& (Bokeh.Models.registered_names().indexOf("ipywidgets_bokeh.widget.IPyWidget") > -1){% endif %})
+    return (Bokeh != null && Bokeh.Panel !== undefined{% for reqs in requirements %} && ({% for req in reqs %}{% if loop.index0 > 0 %}||{% endif %} root['{{ req }}'] !== undefined{% endfor %}){% endfor %}{% if ipywidget %}&& Bokeh.Panel.IPyWidget !== undefined {% endif %})
```
philippjfr, line 38:
```suggestion
    return (Bokeh != null && Bokeh.Panel !== undefined{% for reqs in requirements %} && ({% for req in reqs %}{% if loop.index0 > 0 %}||{% endif %} root['{{ req }}'] !== undefined{% endfor %}){% endfor %}{% if ipywidget %}&& Bokeh.Models._known_models.has("ipywidgets_bokeh.widget.IPyWidget"){% endif %})
```

philippjfr:
Thanks for the fix. Have to think about this a little more. To give a bit of background, Panel has two approaches for rendering IPyWidgets:

1. When you're inside a notebook environment it uses the notebooks WidgetManager and will render itself using the Panel ipywidget model.
2. When running on a server or when embedded in a website it instead uses the `ipywidgets_bokeh` implementation of a renderer.

So really this code should only check for the `ipywidgets_bokeh` model if we are in scenario 2.

panel/_templates/doc_nb_js.js
```diff
@@ -35,7 +35,7 @@
   }
   function is_loaded(root) {
     var Bokeh = get_bokeh(root)
-    return (Bokeh != null && Bokeh.Panel !== undefined{% for reqs in requirements %} && ({% for req in reqs %}{% if loop.index0 > 0 %}||{% endif %} root['{{ req }}'] !== undefined{% endfor %}){% endfor %}{% if ipywidget %}&& (Bokeh.Models.registered_names().indexOf("ipywidgets_bokeh.widget.IPyWidget") > -1){% endif %})
+    return (Bokeh != null && Bokeh.Panel !== undefined{% for reqs in requirements %} && ({% for req in reqs %}{% if loop.index0 > 0 %}||{% endif %} root['{{ req }}'] !== undefined{% endfor %}){% endfor %}{% if ipywidget %}&& Bokeh.Models._known_models.has("ipywidgets_bokeh.widget.IPyWidget") !== undefined {% endif %})
```
philippjfr, line 38:
```suggestion
    return (Bokeh != null && Bokeh.Panel !== undefined{% for reqs in requirements %} && ({% for req in reqs %}{% if loop.index0 > 0 %}||{% endif %} root['{{ req }}'] !== undefined{% endfor %}){% endfor %}{% if ipywidget %}&& Bokeh.Models._known_models.has("ipywidgets_bokeh.widget.IPyWidget") {% endif %})
```

PR #4954, opened 2023-05-27
Add support-Panel-with-a-star shield to side nav

droumis:
Just an idea to help address the discrepancy between the number of active users and # of stars on github.

Feedback welcome. I'm not thrilled about the UX, but maybe it's fine...

image

image

### DIFF
--- a/doc/_static/css/custom.css
+++ b/doc/_static/css/custom.css
@@ -72,3 +72,8 @@ button.copybtn:hover {
 .o-tooltip--left:after {
   background: none;
 }
+
+.github-star-badge {
+  text-align: center;
+  margin-bottom: 10px;
+}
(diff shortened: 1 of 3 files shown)


philippjfr:
Love the idea but wondering if it should go in the sidebar footer or maybe in the top right like other badges.

MarcSkovMadsen:
Top right would be nice if it can fit

droumis:
I haven't found an aesthetically reasonable solution to putting it in the top right navbar. 

Here is my compromise, to put it as a close-to-top-right secondary sidebar item:

image

Is this good enough for now? Any change to the text?

Here is what it looks like on pages with other badges in that spot:

image

philippjfr:
Sorry if that wasn't clear but that's exactly what I meant, thank you!

PR #4955, opened 2023-05-27
Fixed vizzu data series type conversion for datetime.

simzer:
Data series with 'datetime' type were passed as 'measure' to Vizzu. However, they should be passed as dimensions; otherwise Vizzu won't handle them correctly.


philippjfr:
UI test failure is transient, I'll merge.

PR #4978, opened 2023-05-30
Resolving design modifiers based on the theme type

maximlt:
Found this while trying out [objgraph](https://pypi.org/project/objgraph/) on a dummy app (to find a memory leak on a bigger app) and it proved useful as it detected that opening a new session of the same app generated some new `NativeDefaultTheme` objects that weren't cleaned up. This happened because `_resolve_modifiers` - which is decorated with `@functools.lru_cache` - was passed a new instance of the theme instead of its type, while I believe passing the type is enough (I haven't followed all the *modifiers* related code though).

### DIFF
--- a/panel/theme/base.py
+++ b/panel/theme/base.py
@@ -185,7 +185,7 @@ class Design(param.Parameterized, ResourceComponent):
         from ..io.resources import (
             CDN_DIST, component_resource_path, resolve_custom_path,
         )
-        modifiers, child_modifiers = cls._resolve_modifiers(type(viewable), theme)
+        modifiers, child_modifiers = cls._resolve_modifiers(type(viewable), type(theme))
         modifiers = dict(modifiers)
         if 'stylesheets' in modifiers:
             if isolated:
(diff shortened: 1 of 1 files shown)


maximlt:
Well I should have ran the tests locally first! It seems like this change broke a few things. Marking the PR as draft.

philippjfr:
Suspect sometimes the theme is passed in as the type and sometimes as the instance, so it just needs an isinstance check.

panel/theme/base.py
```diff
@@ -179,7 +179,7 @@ def _get_modifiers(
         from ..io.resources import (
             CDN_DIST, component_resource_path, resolve_custom_path,
         )
-        modifiers, child_modifiers = cls._resolve_modifiers(type(viewable), theme)
+        modifiers, child_modifiers = cls._resolve_modifiers(type(viewable), type(theme))
```
philippjfr, line 182:
```suggestion
        theme_type  = type(theme) if isinstance(theme, Theme) else theme 
        modifiers, child_modifiers = cls._resolve_modifiers(type(viewable), them_type)
```

philippjfr (approved):
Looks good to me and thanks for catching this!

philippjfr (approved):
Looks good to me and thanks for catching this!

philippjfr (approved):

PR #5007, opened 2023-06-02
Ignore reference already known warning in VSCode

Hoxbro:
Fixes https://github.com/holoviz/holoviews/issues/5731

Another case:
image

I have been conservative and only ignored it in VSCode.

### DIFF
--- a/panel/config.py
+++ b/panel/config.py
@@ -9,6 +9,7 @@ import importlib
 import inspect
 import os
 import sys
+import warnings
 
 from concurrent.futures import ThreadPoolExecutor
 from contextlib import contextmanager
@@ -791,6 +792,7 @@ class panel_extension(_pyviz_extension):
 
         if "VSCODE_PID" in os.environ:
             config.comms = "vscode"
+            self._ignore_bokeh_warnings()
             return
 
     def _apply_signatures(self):
@@ -842,6 +844,11 @@ class panel_extension(_pyviz_extension):
         """
         from .entry_points import load_entry_points
         load_entry_points('panel.extension')
+
+    def _ignore_bokeh_warnings(self):
+        from bokeh.util.warnings import BokehUserWarning
+        warnings.filterwarnings("ignore", category=BokehUserWarning, message="reference already known")
+
 
 #---------------------------------------------------------------------
 # Private API


philippjfr (approved):
LGTM

giladpn:
@philippjfr @Hoxbro I am using version 1.1.1 (from last week?) but this fix, it seems to me, is not in there. I get e.g.
```
BokehUserWarning: reference already known '6bcaff44-9f70-4597-b9dd-c75214adbc30'
```
Versions used:
Bokeh 3.1.1
jupyter_bokeh 3.0.7
Panel 1.1.1
vscode 1.79.2

Some code to reproduce, omitting irrelevant details, is:
```
from math import floor
import bokeh.plotting as bkp
from bokeh.events import Tap
from bokeh.io import output_notebook
import panel as pn
pn.extension()
pn.config.comms = 'vscode'
output_notebook()

def tap_callback(event):
            an_index = floor(event.y)
            a_name = event.model.y_range.factors[an_index]
            new_value =  ... # ommitted irrelevant details

            # create new dict of all the data
            new_data = {}
            for r in event.model.renderers:
                if hasattr(r, 'glyph'):
                    new_data = r.data_source.data.copy()
                    new_data['relevant column name'][-1-an_index] = new_value
                    r.data_source.data = new_data
                    break

def app():
   '''
   this function should return a Bokeh model, such as a plot, column or gridplot, 
   which shows your stuff
   '''
   ... # details of creating data and the bokeh_model
   fig = bkp.figure(....details...)
   fig.on_event(Tap, tap_callback)
   return fig

bokeh_pane = pn.pane.Bokeh(app())
bokeh_pane
```
Am I missing something? Should 1.1.1 have this fix? Is there some config I should be using?
Thanks
GPN

Hoxbro:
The change in this PR is only enabled if no `comms` has been set and `pn.extension()` is called. Try removing `pn.config.comms` and restarting the kernel. 

Please post [minimal, reproducible example (MRE)](https://stackoverflow.com/help/minimal-reproducible-example) if this problem still occurs.

giladpn:
Thanks @Hoxbro however if I remove the line
```
pn.config.comms = 'vscode'
```
then events no longer propagate properly to *Python* callbacks (I double checked and tested).

I am working of course in vscode so it is perhaps not surprising that if I need events to propagate to python callbacks - I must enable vscode comms. Is there a way to get rid of these warnings even though I do enable vscode comms?

As always, many thanks for the support

GPN

Hoxbro:
`pn.config.comms = 'vscode'` should be automatic set, when calling `pn.extension`. Can you call `pn.extension` from a clean notebook and see what the value of `pn.config.comms` is after it?

giladpn:
Thanks @Hoxbro - I just did that and without setting the comms I get 
```
pn.config.comms = default
```

Here is a minimal reproducible example. Note that the `comms` line has been commented out.
```
import bokeh.plotting as bkp
from bokeh.events import Tap
from bokeh.io import output_notebook
import panel as pn
pn.extension()
# pn.config.comms = 'vscode'
print(f'pn.config.comms = {pn.config.comms}')
output_notebook()

fruits = ['Apples', 'Pears', 'Nectarines', 'Plums', 'Grapes', 'Strawberries']
counts = [5, 3, 4, 2, 4, 6]

def tap_callback(event):
    print(f'Here I am! tapped at ({event.x},{event.y})')

def app():
    fig = bkp.figure(x_range=fruits, height=350, title="Fruit Counts",
            toolbar_location=None, tools="")

    fig.vbar(x=fruits, top=counts, width=0.9)

    fig.xgrid.grid_line_color = None
    fig.y_range.start = 0
    fig.on_event(Tap, tap_callback)
    return fig

bokeh_pane = pn.pane.Bokeh(app())
bokeh_pane
```
Versions used:
Bokeh 3.1.1
jupyter_bokeh 3.0.7
Panel 1.1.1
vscode 1.79.2

In this MRE tapping the plot does *not* print out "Here I am". What you get is 
```
pn.config.comms = default
```

However, if you uncomment the comms line, restart the python kernel, and run the MRE - then you will get an output such as:
```
pn.config.comms = vscode
Here I am! tapped at (2.431304347826087,2.163636363636364)
BokehUserWarning: reference already known '97415e03-8d28-485f-be51-92beb2414b4b'
Here I am! tapped at (1.6382608695652174,1.506060606060606)
BokehUserWarning: reference already known '97415e03-8d28-485f-be51-92beb2414b4b'
```
so with the comms enabled, the event does propagate to the python callback.

Many thanks

GPN

Hoxbro:
I'm getting `pn.config.comms = vscode` when running your example above. Are you running vscode locally or ssh into an instance? Can you check if you have `"VSCODE_PID"` in your environment?

giladpn:
I ran
```
printenv | grep VSCODE
```
(is this what you wanted when you asked I check `VSCODE_PID`? please guide me if I misunderstood) 
in two environments:
+ WSL with Ubuntu 22.04
+ a docker container with RHEL9

The result for both environments if there is no such env variable `VSCODE_PID`. Did I understand you correctly?

Thanks 

GPN

Hoxbro:
I can see `VSCODE_PID` is not set in a WSL notebook. I will make a new issue around this.

PR #5009, opened 2023-06-02
Handle height responsiveness of HoloViews layout correctly

philippjfr:
Since the Bokeh 3.x layout engine changes layouts have to precisely reflect the sizing_mode of their contents. So while previously we got away with not setting height responsiveness on a container with a responsive HoloViews plot this now results in the plot collapsing in height. Therefore the HoloViews pane now does two things:

1. Correctly computes the height responsiveness
2. If it turns out that the layout it generated is actually incorrect (because information about responsiveness may not be known until the plot is rendered) and it has to invalidate the whole layout the `RerenderError` now has the capability to propagate the invalidation all the way up to the layout that actually needs to be re-rendered.

To expand on this, when a HoloViews pane is instantiated the following happens:

1. A layout is generated on instantiation
2. During rendering this layout is instantiated and the layout models are generated before the HoloViews plot is generated
3. When the HoloViews plot is generated it may turn out that the layout generated during instantiation is actually incorrect.
4. In such a case we need to propagate that fact all the way up to the root layout to tell it that the layouts it generated are actually wrong and it needs to restart the rendering process.

This is a convoluted approach to rendering but the process is relatively cheap and it avoids eagerly evaluating `DynamicMap` before they are actually rendered.

### DIFF
--- a/panel/pane/holoviews.py
+++ b/panel/pane/holoviews.py
@@ -172,13 +172,14 @@ class HoloViews(PaneBase):
             smode = 'stretch_height'
         else:
             smode = None
+        layout_smode = 'stretch_width' if not smode and center else smode
         if not len(self.widget_box):
             if center:
                 components = [HSpacer(), self, HSpacer()]
             else:
                 components = [self]
             self.layout[:] = components
-            self.layout.sizing_mode = smode
+            self.layout.sizing_mode = layout_smode
             return
 
         items = (self.widget_box, self) if widget_first else (self, self.widget_box)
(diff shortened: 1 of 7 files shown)


panel/pane/holoviews.py
```diff
@@ -163,24 +164,38 @@ def _update_layout(self):
...
                 components = [self]
             self.layout[:] = components
+            self.layout.sizing_mode = smode
             return
 
         items = (self.widget_box, self) if widget_first else (self, self.widget_box)
+        kwargs = {'sizing_mode': smode}
         if not center:
-            components = [layout(*items)]
+            if False:#self.default_layout is layout:
```
Hoxbro, line 187: This looks like a debug thing

philippjfr: Indeed, thanks for catching, may re-enable.

Hoxbro:
I can't seem to run a simple example:
image

philippjfr:
Thanks and yes I see why will have a fix and some extra tests after lunch.

Issue #5016, opened 2023-06-02
Add example comparing a simple Streamlit app to a Panel app.


MarcSkovMadsen:
Its great to have the section [Compare Panel and Streamlit](https://panel.holoviz.org/explanation/comparisons/compare_streamlit.html) but its very text heavy and does not indicate in any way how to code and apps compare.

Please add an end to end example comparing the two. And example that will also help the user migrate from Streamlit to Panel as I saw in a request for on Discourse. The user deleted his post. But came back later with this post https://discourse.holoviz.org/t/cannot-execute-script-with-panel/5464.

MarcSkovMadsen:
It could for example be this example suggested on Discourse

## Streamlit

https://github.com/holoviz/panel/assets/42288570/58512dbd-2f2e-4e70-9fd0-688ac6c8314c

```python
import time

import streamlit as st


def calculation_a():
    time.sleep(1.5)


def calculation_b():
    time.sleep(1.5)


st.write("# Calculation Runner")

option = st.radio("Which Calculation would you like to perform?", ("A", "B"))

st.write("You chosed: ", option)
if option == "A":
    if st.button("Press to run calculation"):
        with st.spinner("running... Please wait!"):
            time_start = time.perf_counter()
            calculation_a()
            st.write("Done!")
            time_end = time.perf_counter()
            st.write(f"The function took {time_end - time_start:1.1f} seconds to complete")

    else:
        st.write("Calculation A did not run yet")

elif option == "B":
    if st.button("Press to run calculation"):
        with st.spinner("running... Please wait!"):
            time_start = time.perf_counter()
            calculation_b()
            st.write("Done!")
            time_end = time.perf_counter()
            st.write(f"The function took {time_end - time_start:1.1f} seconds to complete")
    else:
        st.write("Calculation B did not run yet")
```

COMING UP: PANEL CODE

philippjfr:
Just had a thought, I think if we supported generator this could be rewritten roughly like this:

```python
import time
import panel as pn

pn.extension(sizing_mode="stretch_width", template='bootstrap')

calculation = pn.widgets.RadioBoxGroup(
    name="Calculation",
    options=["Calculation A", "Calculation B"],
)
run = pn.widgets.Button(name="Press to run calculation")

def calculation_a():
    time.sleep(2.5)

def calculation_b():
    time.sleep(1.5)

def runner(calculation, running):
    if not running:
        return f"Calculation {calculation} did not run yet"
    yield pn.Column(
[...]

PR #5019, opened 2023-06-03
Implement support for reactive generator functions

philippjfr:
In looking at https://github.com/holoviz/panel/issues/5016 I realized that the only way to write reactive expressions with multiple steps in a natural way would be using generators. So I had a go at implementing support for them and now I'm in love, here's a simplified version of the app in #5016:

### Asynchronous Generator

```python
import asyncio
import panel as pn

pn.extension(sizing_mode="stretch_width", template='bootstrap')

run = pn.widgets.Button(name="Press to run calculation")

async def runner(running):
    if not running:
        yield "Calculation did not run yet"
        return
    yield pn.Column(
        pn.indicators.LoadingSpinner(value=True, width=25, height=25),
        "Running...Please wait!",
    )
    time_start = time.perf_counter()
    await asyncio.sleep(1.5)
    time_end = time.perf_counter()
    yield f"""Done!
    The function took {time_end - time_start:1.1f} seconds to complete"""

pn.Column(
    "# Calculation Runner", run, pn.bind(runner, run)
).servable()
```

### Classic Generator

```python
import time
import panel as pn

pn.extension(sizing_mode="stretch_width", template='bootstrap')

run = pn.widgets.Button(name="Press to run calculation")

def runner(running):
    if not running:
        yield "Calculation did not run yet"
        return
    yield pn.Column(
        pn.indicators.LoadingSpinner(value=True, width=25, height=25),
        "Running...Please wait!",
    )
    time_start = time.perf_counter()
    time.sleep(1.5)
    time_end = time.perf_counter()
    yield f"""Done!
    The function took {time_end - time_start:1.1f} seconds to complete"""

pn.Column(
    "# Calculation Runner", run, pn.bind(runner, run)
).servable()
```

generator

## Streaming example

```python
import asyncio
import panel as pn
import random

pn.extension()

@pn.depends()
async def stream():
    while True:
        await asyncio.sleep(0.1)
        yield random.randint(0, 100)
        
pn.indicators.Number(value=stream, name='Random')
```

random_generator

### DIFF
--- a/panel/depends.py
+++ b/panel/depends.py
@@ -1,5 +1,6 @@
-import inspect
 import sys
+
+from inspect import isasyncgenfunction
 
 import param
 
@@ -223,7 +224,7 @@ def _param_bind(function, *args, watch=False, **kwargs):
                 fn = eval_function(p)
         return fn
 
-    if inspect.isasyncgenfunction(function):
+    if isasyncgenfunction(function):
         async def wrapped(*wargs, **wkwargs):
             combined_args, combined_kwargs = combine_arguments(
                 wargs, wkwargs, asynchronous=True
--- a/panel/depends.py
+++ b/panel/depends.py
@@ -1,3 +1,4 @@
+import inspect
 import sys
 
 import param
@@ -183,7 +184,7 @@ def _param_bind(function, *args, watch=False, **kwargs):
         elif isinstance(v, param.Parameter):
             dependencies[kw] = v
 
-    def combine_arguments(wargs, wkwargs):
+    def combine_arguments(wargs, wkwargs, asynchronous=False):
         combined_args = []
         for arg in args:
             if hasattr(arg, '_dinfo'):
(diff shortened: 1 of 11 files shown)


MarcSkovMadsen:
The streaming example is extremely powerful. The Async and Classic generators too, but here we have to teach our users `yield`.

MarcSkovMadsen:
Does `pn.bind(runner, run)` run twice when I click the `run` button? Has it always done that?

panel/depends.py
```diff
@@ -216,12 +223,24 @@ def eval_fn():
                 fn = eval_function(p)
         return fn
 
-    if iscoroutinefunction(function):
+    if inspect.isasyncgenfunction(function):
```
MarcSkovMadsen, line 226: use `from inspect import isasyncgenfunction` instead. You will save an attribute lookup each time this line is run which could be thousands or millions of times.

MarcSkovMadsen:
I added another use case for generator functions in Panel that I believe should be supported as well. Please consider this in the context of this PR. Thanks. https://github.com/holoviz/panel/issues/5020

philippjfr:
>Does pn.bind(runner, run) run twice when I click the run button? Has it always done that?

It runs once but if you `bind` a Button to a function it'll run the first time with a value of False and on every click it'll run with a value of `True`.

>Does this work with other widgets like an IntSlider?

Absolutely, it works just like a regular function with `bind`.

>Does the streaming example perform nicely if you have 100 streaming generators? I'm asking because I've found that its not performant to setup 100 period callbacks - its better to update 100 times in one. It would be so great not to have to couple the streaming components.

Absolutely, this should be significantly more lightweight than periodic callbacks, here's 100 instances of the streaming generator

many_generators

panel/depends.py
philippjfr: I'll do that but note that this is not inside the core loop and won't be called on each update.

philippjfr:
Okay, feature-wise this is now done. Last item was implementing synchronization primitives to ensure the async generators were updated when a dependency changes, e.g. we make the timeout dynamic:

```python
import asyncio
import panel as pn
import random

pn.extension()

timeout = pn.widgets.FloatSlider(start=0.1, end=1.0, value=1, name='Timeout')

@pn.depends(timeout)
async def stream(timeout):
    while True:
        await asyncio.sleep(timeout)
        yield random.randint(0, 100)

pn.Column(
    timeout,
    pn.FlexBox(*(pn.indicators.Number(value=stream, name='Random', font_size='12pt') for _ in range(200)))
).servable()
```

async_gen_deps

Issue #5050, opened 2023-06-06
ChatBox `liked` disappears after sending message


ahuang11:
To reproduce, run this:

```python
import panel as pn

pn.extension(design="fast", theme="dark")

chat_box = pn.widgets.ChatBox(
    value=[
        {"You": "Hello bot!"},
        {"Bot": pn.pane.GIF("https://thumbs.gfycat.com/PepperyGrizzledClownanemonefish-max-1mb.gif")},
    ],
    allow_likes=True,
    ascending=True,
    message_hue=200,
    width=900,
)

chat_box.servable()
```

Then click the heart and send a message.

The heart reverts back to null.

PR #5065, opened 2023-06-08
Keeps like in ChatBox and prevent duplicate sends in notebook

ahuang11:
Fixes: https://github.com/holoviz/panel/issues/5050

Also fixes a strange issue where there were duplicate sends in notebook:
image


philippjfr:
Looks good, thanks!

PR #5089, opened 2023-06-11
Allow to clear enabled_dates in DatetimePicker

### LINKED ISSUE #4949: Make it possible to override enabled_dates by setting disabled_dates in DatePicker
Suppose you have set a list of valid dates with enabled_dates and want to make all dates available again. Reading the docs you could imagine that setting disabled_dates to an empty list would do the trick but in fact it cannot override enabled_dates (the reverse works). However in 0.14.4, updating enabled_dates to an empty list did that.

The behaviour has changed with 1.0 and an empty enabled_dates makes all dates unavailable. This is more intuitive but now to my knowledge there is no way to reinitialise the available dates. I've also tried setting enabled_dates to None which yields an error in javascript. See the following code:

```python
import panel as pn
pn.extension()


def update(event):
    if event.obj == button1:
        picker.enabled_dates = [] # Used to work in 0.14.4
    elif event.obj == button2:
        picker.enabled_dates = None # Error in browser console
    elif event.obj == button3:
        picker.disabled_dates = []
    elif event.obj == button4:
        picker.disabled_dates = None


picker = pn.widgets.DatePicker(enabled_dates=['2023-05-26'])
button1 = pn.widgets.Button(name='enabled_dates=[]')
button2 = pn.widgets.Button(name='enabled_dates=None')
button3 = pn.widgets.Button(name='disabled_dates=[]')
button4 = pn.widgets.Button(name='disabled_dates=None')

button1.on_click(update)
button2.on_click(update)
button3.on_click(update)
button3.on_click(update)

pn.Column(picker, pn.Row(button1, button2, button3, button4)).servable()
```

From a user perspective it would be nice to make both parameters able to override the other?

It may be possible, albeit not ideal, to reinstate the original behaviour.

Hoxbro:
Fixes #4949

Implementation is like the changes made upstream https://github.com/bokeh/bokeh/pull/13194 to fix the problem.

### DIFF
--- a/panel/models/datetime_picker.ts
+++ b/panel/models/datetime_picker.ts
@@ -41,11 +41,20 @@ export class DatetimePickerView extends InputWidgetView {
     this.connect(max_date.change, () => this._picker?.set("maxDate", this.model.max_date))
     this.connect(disabled_dates.change, () => {
       const {disabled_dates} = this.model
-      this._picker?.set("disable", disabled_dates != null ? _convert_date_list(disabled_dates) : undefined)
+      this._picker?.set("disable", disabled_dates != null ? _convert_date_list(disabled_dates) : [])
     })
     this.connect(enabled_dates.change, () => {
       const {enabled_dates} = this.model
-      this._picker?.set("enable", enabled_dates != null ? _convert_date_list(enabled_dates) : undefined)
+      if (enabled_dates != null) {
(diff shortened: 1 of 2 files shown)


philippjfr (changes requested):
Any chance we could add a quick UI test?

philippjfr (approved):
Looks good, thanks!

Issue #5100, opened 2023-06-13
ChatBox Send button stretches too far.


ahuang11:
```
import panel as pn
from langchain.llms import OpenAI
from langchain.chains import ConversationChain
from langchain.memory import ConversationBufferMemory

pn.extension()


def chat(history):
    chat_box.append({"AI": "HELLO"})
    # response = conversation.predict(input=history[-1])
    # chat_box.append(response)

# llm = OpenAI(temperature=2)
# conversation = ConversationChain(llm=llm, memory=ConversationBufferMemory())

chat_box = pn.widgets.ChatBox()
pn.bind(chat, history=chat_box, watch=True)
chat_box.servable()
```

<img width="494" alt="image" src="https://github.com/holoviz/panel/assets/15331990/b7e7c033-60f5-402f-9c53-61c378ca7ef8">

ahuang11:
Can't tell if this is a general Panel issue or specific to ChatBox because

- send_button is stretch_both
- row is stretch_width
```

        self._send_button = Button(
            name="Send",
            button_type="default",
            sizing_mode="stretch_both",
            max_width=100,
        )
...
        row_layout = layout.copy()
        row_layout.pop("width", None)
        row_layout.pop("height", None)
        row_layout["sizing_mode"] = "stretch_width"
...
            message_row = Row(message_input, self._send_button.clone(), **row_layout)
            input_items.append((message_input_name, message_row))

```

Issue #5099, opened 2023-06-13
ChatBox cannot be disabled


ahuang11:
```
import panel as pn

pn.extension()

chat_box = pn.widgets.ChatBox(disabled=True)
chat_box.servable()
```

Still can submit.

PR #5101, opened 2023-06-13
Fix ChatBox issues

ahuang11:
While trying to use it for an actual use case, I found some problems with it, namely:
1. Send button height stretches too far https://github.com/holoviz/panel/issues/5100
2. ChatBox inputs cannot be disabled https://github.com/holoviz/panel/issues/5099
3. pn.bind didn't work with `append` because previously, `append` doesn't actually trigger value in the latest patch
4. If `append` now trigger values, it does a whole refresh and `liked` again gets lost so I added logic to match up the messages along with index

### DIFF
--- a/panel/widgets/chatbox.py
+++ b/panel/widgets/chatbox.py
@@ -47,37 +47,37 @@ class ChatRow(CompositeWidget):
         """,
     )
 
-    icon = param.String(
-        default=None, doc="""The icon to display adjacent to the value"""
-    )
+    icon = param.String(default=None, doc="""The icon to display adjacent to the value""")
 
-    liked = param.Boolean(default=False, doc="""Whether a user liked the message""")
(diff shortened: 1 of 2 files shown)


panel/widgets/chatbox.py
```diff
@@ -205,47 +222,79 @@ class ChatBox(CompositeWidget):
     >>> ChatBox(value=[{"You": "Hello!"}, {"Bot": ["How may I help?", "I'm a bot."]}])
     """
 
-    value = param.List(default=[], item_type=Dict, doc="""
+    value = param.List(
```
philippjfr, line 225: I realize you probably use auto-formatting and this is annoying but at least for the time being I would like to keep param definitions consistent in their formatting.

ahuang11: Yeah I do, but no worries!

Issue #5098, opened 2023-06-12
Styling does not properly render (server-side) with Tabulator filters in version 1.1.0


enismaxim1:
Version: 1.1.0

Similar to issue #5085, styling in Tabulator does not properly render when run in a server in Panel version 1.1.0. The styling does, however, work properly when embedded in a Jupyter notebook.

Minimal code:
```
import panel as pn
import pandas as pd

df = pd.DataFrame({
    'a': [-1,0,1]
})

select = pn.widgets.Select(options = [None, -1, 0, 1])
table = pn.widgets.Tabulator(df)
table.add_filter(select, 'a')
table.style.applymap(lambda x: 'background-color: red' if x >= 0 else '')
pn.Column(select, table).servable()
```

After running a server on the aforementioned code, filtering by -1, then clearing the filter, we see that the positive table elements lose their styling.

enismaxim1:
Resolved in the PR

PR #5110, opened 2023-06-14
[Bug Fix] Fix styling with filters in tabulator

enismaxim1:
This PR fixes the issue identified in #5098, where Tabulator styling would not render after filtering in panel version 1.1.0. 

Previously, there was a race condition causing the table to usually set styling before new data is processed into the table, meaning the table was rerendered without the requisite CSS. The solution is to reapply styling after the rerendering.

The PR also adds a test to check that styling is preserved after filtering, which passes due to the fix in this PR.

### DIFF
--- a/panel/tests/ui/widgets/test_tabulator.py
+++ b/panel/tests/ui/widgets/test_tabulator.py
@@ -2467,7 +2467,6 @@ def test_tabulator_filters_and_styling(page, port, df_mixed):
     page.goto(f"http://localhost:{port}")
 
     # Filtering to one field and then clicking None again should display all data, with styling
-    print(page.locator('option'))
     page.locator('option').nth(1).click()
     page.locator('option').nth(0).click()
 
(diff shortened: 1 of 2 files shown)


philippjfr:
pre-commit.ci autofix

panel/tests/ui/widgets/test_tabulator.py
```diff
@@ -2450,6 +2452,31 @@ def test_tabulator_patching_and_styling(page, port, df_mixed):
...
+    layout = Column(select, table)
+
+    serve(layout, port=port, threaded=True, show=False)
+
+    time.sleep(0.2)
+
+    page.goto(f"http://localhost:{port}")
+
+    # Filtering to one field and then clicking None again should display all data, with styling
+    print(page.locator('option'))
```
philippjfr, line 2470:
```suggestion
```

philippjfr:
Thank you @enismaxim1, really appreciate the test as well.

PR #5118, opened 2023-06-15
Add ChatBox methods and replacing ChatRow values

ahuang11:
As I was building my [AI chat box](https://discourse.holoviz.org/t/openai-conversation-using-chatbox/5548) I felt like `pn.widgets.ChatBox` was missing some core functionality so this PR adds:

- Allows updating row message
- Adds insert, pop, replace list methods


ahuang11:
Actually dropping support for `Allows updating row message` because it gets out of sync with ChatBox.

philippjfr (approved):
Looking great!

PR #5125, opened 2023-06-16
Bug fix in table filter

TBym:
List with a single value gets converted to the value itself, however Pandas isin() function is expecting a list.

### DIFF
--- a/panel/widgets/tables.py
+++ b/panel/widgets/tables.py
@@ -483,6 +483,7 @@ class BaseTable(ReactiveData, Widget):
             elif op == '<=':
                 filters.append(col <= val)
             elif op == 'in':
+                if not isinstance(val, list): val = [val]
                 filters.append(col.isin(val))
             elif op == 'like':
                 filters.append(col.str.contains(val, case=False, regex=False))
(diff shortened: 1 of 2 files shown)


Hoxbro:
Can you add a test in https://github.com/holoviz/panel/blob/main/panel/tests/widgets/test_tables.py. 

Which did not work before this fix and will work after?

Hoxbro:
Can you add a test in https://github.com/holoviz/panel/blob/main/panel/tests/widgets/test_tables.py. 

Which did not work before this fix and will work after?

TBym:
Added function `def test_tabulator_constant_single_element_list_filter_client_side(document, comm)` to test_table.py

Hoxbro:
Your changes breaks panel/tests/widgets/test_tables.py::test_tabulator_constant_list_filter_client_side https://github.com/holoviz/panel/actions/runs/5288423386/jobs/9570931880#step:6:5062

Can you take a look at why this is the case?

TBym:
it should be fine now.

panel/widgets/tables.py
```diff
@@ -483,6 +483,7 @@ def _get_header_filters(self, df):
             elif op == '<=':
                 filters.append(col <= val)
             elif op == 'in':
+                if not isinstance(val, (list, np.ndarray)): val = [val]
```
Hoxbro, line 486: Can you add a comment referencing the issue above this if-statement?

philippjfr (approved):
Don't want to hold this up further. Very much appreciate the added test!

PR #5129, opened 2023-06-16
Add append mode for reactive generator output

philippjfr:
Allows generator output to append to output rather than simply replacing it:

```python
import random
import time
import panel as pn

def model():
    time.sleep(1)
    return random.randint(0, 100)

def results(running):
    if not running:
        yield "The model has not run yet"
        return

    for i in range(0, 10):
        yield f"Result {i}: {model()}"

run_input = pn.widgets.Button(name="Run model")
pn.Column(
    run_input,
    pn.panel(pn.bind(results, run_input), loading_indicator=True, generator_mode='append'),
).servable()
```

gen_append

### DIFF
--- a/panel/pane/base.py
+++ b/panel/pane/base.py
@@ -24,7 +24,7 @@ from ..io.document import create_doc_if_none_exists, unlocked
 from ..io.notebook import push
 from ..io.state import state
 from ..layout.base import (
-    ListPanel, NamedListPanel, Panel, Row,
+    Column, ListPanel, NamedListPanel, Panel, Row,
 )
 from ..links import Link
 from ..models import ReactiveHTML as _BkReactiveHTML
@@ -559,7 +559,7 @@ class ReplacementPane(PaneBase):
         super().__init__(object, **params)
         self._pane = panel(None)
         self._internal = True
-        self._inner_layout = Row(self._pane, **{k: v for k, v in params.items() if k in Row.param})
+        self._inner_layout = Column(self._pane, **{k: v for k, v in params.items() if k in Column.param})
         self._internal_callbacks.append(
             self.param.watch(self._update_inner_layout, list(Layoutable.param))
         )
@@ -706,7 +706,7 @@ class ReplacementPane(PaneBase):
             return
 
         self._pane = new_pane
-        self._inner_layout[0] = self._pane
+        self._inner_layout[:] = [self._pane]
         self._internal = internal
 
     def _cleanup(self, root: Model | None = None) -> None:
--- a/panel/param.py
+++ b/panel/param.py
@@ -823,12 +823,11 @@ class ParamMethod(ReplacementPane):
         try:
             if isinstance(awaitable, types.AsyncGeneratorType):
                 append_mode = self.generator_mode == 'append'
-                async for i, new_obj in enumerate(awaitable):
+                if append_mode:
+                    self._inner_layout[:] = []
+                async for new_obj in awaitable:
                     if append_mode:
-                        if i == 0:
-                            self._inner_layout[:] = [new_obj]
-                        else:
-                            self._inner_layout.append(new_obj)
+                        self._inner_layout.append(new_obj)
                         self._pane = self._inner_layout[-1]
                     else:
                         self._update_inner(new_obj)
@@ -858,12 +857,11 @@ class ParamMethod(ReplacementPane):
                 return
             elif isinstance(new_object, Generator):
                 append_mode = self.generator_mode == 'append'
-                for i, new_obj in enumerate(new_object):
+                if append_mode:
+                    self._inner_layout[:] = []
+                for new_obj in new_object:
                     if append_mode:
-                        if i == 0:
-                            self._inner_layout[:] = [new_obj]
-                        else:
-                            self._inner_layout.append(new_obj)
+                        self._inner_layout.append(new_obj)
                         self._pane = self._inner_layout[-1]
                     else:
                         self._update_inner(new_obj)
--- a/panel/param.py
+++ b/panel/param.py
@@ -764,6 +764,9 @@ class ParamMethod(ReplacementPane):
         Whether to defer load until after the page is rendered.
         Can be set as parameter or by setting panel.config.defer_load.""")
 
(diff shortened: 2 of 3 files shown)


MarcSkovMadsen:
One issue with your example. The appended results are disabled until all results are ready.

If for example my generative AI model streams multiple video files, audio files, tables or similar back, I want to be able to interact with them as soon as they are ready. Not wait several minutes until the final result is returned 😄

MarcSkovMadsen:
Please also consider if `generator_mode` should be inside `pn.bind`/ `pn.depends` instead? This would save users the friction of having to put it in a `pn.panel`.

philippjfr:
```python
import random
import time
import panel as pn

def model():
    time.sleep(1)
    return random.randint(0, 100)

def results(running):
    if not running:
        yield "The model has not run yet"
        return

    for i in range(0, 10):
        result = pn.panel('Running model...', loading=True)
        yield result
        result.param.update(object=f"Result {i}: {model()}", loading=False)

run_input = pn.widgets.Button(name="Run model")
pn.Column(
    run_input,
    pn.panel(pn.bind(results, run_input), generator_mode='append'),
).servable()
```

gen_load

MarcSkovMadsen:
A `Column` is probably the most used layout. But consider if users would like to use other layouts like `Row` or some grid. Maybe this could be solved by a `generator_layout´ that accepts some *list* like interface.

philippjfr:
>Please also consider if generator_mode should be inside pn.bind/ pn.depends instead? This would save users the friction of having to put it in a pn.panel.

I've been thinking about this a lot but haven't yet come to any good solutions. In `hvplot.bind()` we have allowed using `hvplot.bind(...).interactive()`, wonder if we should consider something like that here or whether there is some other solution. I'm not super excited about adding special keywords to the `pn.bind` signature since it's meant to be a generic thing like `functools.partial` and adding reserved keyword feels dirty.

philippjfr:
I do find `pn.panel(pn.bind(...), ...)` really ugly so I'm very motivated to find something better.

MarcSkovMadsen:
```python
result = pn.panel('Running model...', loading=True)
yield result
result.param.update(object=f"Result {i}: {model()}", loading=False)
```

Please consider whether this can be simplified to two lines. For example you can temporarily `append` and the `pop` a loading spinner. Or `Spacer` with a loading_indicator.

philippjfr:
>Please consider whether this can be simplified to two lines. For example you can temporarily append and the pop a loading spinner. Or Spacer with a loading_indicator.

That approach just falls naturally out of the functionality and is pretty powerful. Anything else like you're suggesting would require injecting arguments into the generator or set up some special control flow signals, which seems overly complex. It would be neat if there was a clean way to temporarily switch between append and replace modes, so you can yield something that is appended and then subsequently yield a replacement.

philippjfr:
Only thing I can imagine is some sentinel value on the replacement:

```python
def results(running):
    if not running:
        yield "The model has not run yet"
        return

    for i in range(0, 10):
        yield pn.panel('Running model...', loading=True)
        yield pn.panel(f"Result {i}: {model()}", tags=['replace'])
```

MarcSkovMadsen:
> Only thing I can imagine is some sentinel value on the replacement:
> 
> ```python
> def results(running):
>     if not running:
>         yield "The model has not run yet"
>         return
> 
>     for i in range(0, 10):
>         yield pn.panel('Running model...', loading=True)
>         yield pn.panel(f"Result {i}: {model()}", tags=['replace'])
> ```

At a first glance it does not "feel" right. Its like an "agreed", "convention" that you need to remember. Then I probably prefer you 3 lines and then clear documentation.

philippjfr:
Indeed, the `tags` approach definitely feels too much like a very ad-hoc convention but it was really mostly to demonstrate the idea. That said I can't think of anything that's actually better.

philippjfr:
Only other suggestion I have and it's definitely in weird convention territory again:

```python
def results(running):
    if not running:
        yield "The model has not run yet"
        return

    for i in range(0, 10):
        yield pn.panel('Running model...', loading=True), ...
        yield f"Result {i}: {model()}"
```

Hoxbro:
I don't think there is a great solutions. The convention seems non-ideal, I don't like adding the tag, and the `pn.bind(...).panel` in my mind is a small rewrite and not a big enough change. 

I would not say the solution I have thought of is great either. But the thinking is making custom bind functions where you set the panel arguments, so you define them once and reuse that custom bind as needed. 

``` python
import random
import time

import panel as pn

def model():
    time.sleep(1)
    return random.randint(0, 100)


def results(running):
    if not running:
        yield "The model has not run yet"
        return

    for i in range(0, 10):
        yield f"Result {i}: {model()}"


# Only reason for two buttons is so they don't block each other
run_input1 = pn.widgets.Button(name="Run model")
run_input2 = pn.widgets.Button(name="Run model")

custom_bind = pn.bind.create(loading_indicator=True, generator_mode="append")

pn.Column(
    pn.Row(run_input1, run_input2),
    pn.Row(
        custom_bind(results, run_input1),
        pn.bind(results, run_input2),
    ),
).servable()
```

https://github.com/holoviz/panel/assets/19758978/0ba9c2e3-8d19-49cd-833b-1c923d652407

<details>
  <summary>The hack to get it to work :upside_down_face:  </summary>

``` python
import panel as pn
from panel.pane import PaneBase


def create(**pn_kwargs):
    def func(*args, **kwargs):
        bound = pn.bind(*args, **kwargs)
        bound.__panel__ = lambda: PaneBase.get_pane_type(bound, **pn_kwargs)(
            bound, **pn_kwargs
        )
        return bound

    return func


pn.bind.create = create
````


</details>

jbednar:
I'm having trouble really wrapping my head around this enough to have a strong opinion, but it seems like `pn.bind.create(loading_indicator=True, generator_mode="append")` is precisely what a `param.ParameterizedFunction` `.instance()` call does, i.e. allows you to configure a "function" with parameter settings that then gets called as normal. I agree with Philipp's reluctance to pollute the `pn.bind` argument namespace, and maybe just .instance vs. a regular call would do it?

philippjfr:
Merging this, adding an API like `pn.bind.create` or `pn.bind(...).panel()` really is a separate discussion.

PR #5138, opened 2023-06-19
Rewrite interactive API without holoviews dependency

philippjfr:
This came out of some discussions with @jbednar on Friday. With the advent of more React-like APIs in Python we are in a place where we don't have a good way to declare standalone reactive variables, such as in this Solara example:

FxpgKt6XgAAipl-

Initially I was thinking that we should be able to declare unbound param.Parameter instances which we could pass around but then I noted that we already have a very powerful reactive data structure in our ecosystem, namely the `hvplot` `interactive` API, which does not only support reactivity but also allows operator chaining and much more.

One of the main obstacles with `.interactive` is that it depended on `dim` expressions in HoloViews as a consequence of which it was specialized to support only objects that a `holoviews.Dataset` can wrap. Since I always had planned to rewrite `.interactive` to be independent of HoloViews AND to allow it to actually chain and store intermediates.

Unlike the `hvplot` implementation this implementation does not store the full chain of operations in `hv.dim` objects, instead we build up a chain of `interactive` objects each of which references the previous one. This means that intermediate results can be cached.

## Migration plan

The plan is to implement a generic version of `interactive` here but allow it to be extended in all the ways that are required for hvPlot to handle plotting and add special handling for DataFrame and xarray Dataset/DataArray types.

## Examples

Counter app:

```python
count = pn.expression(0)
pn.Column(
    pn.widgets.Button(name='Click me!', on_click=lambda e: count.set(count.eval()+1)),
    pn.expression('Click count is {}').format(count)
)
```

## ToDo

- [x] Ensure all types of reactive objects are supported
  - [x] `param.Parameter`
  - [x] `panel.widgets.Widget`
  - [x] Bound functions and methods
  - [x] Other `interactive` objects 
- [x] Decide on top-level API, i.e. is it `pn.ref`
- [x] Add tests
- [ ] Add documentation


philippjfr:
I have now split out an `interactive_base` which effectively only uses APIs that will be available in param 2.0 and `interactive` extends the base class adding all the Panel based display machinery. This means, in principle, `interactive_base` could live in param.

panel/interactive.py
```diff
@@ -0,0 +1,815 @@
...
+
+    def __new__(cls, obj, **kwargs):
+        wrapper = None
+        if 'fn' in kwargs:
+            fn = kwargs.pop('fn')
+        elif isinstance(obj, (FunctionType, MethodType)):
+            fn = panel(obj, lazy=True)
+            obj = fn.eval(obj)
+        elif isinstance(obj, param.Parameter):
+            fn = panel(bind(lambda obj: obj, obj), lazy=True)
```
philippjfr, line 188: These functions should not be wrapped by panel.

panel/depends.py
```diff
@@ -32,6 +32,9 @@ def transform_dependency(arg):
         arg = transform(arg)
     return arg
 
+# Alias for backward compatibility
+param_value_if_widget = transform_dependency
```
Hoxbro, line 36:
```suggestion
def param_value_if_widget(*args, **kwargs):
    deprecated("1.3", "param_value_if_widget", "transform_dependency")
    return transform_dependency(*args, **kwargs)
```

+ An import on top. 

Note: I just chose a version.

panel/react.py
```diff
@@ -0,0 +1,963 @@
...
+        from matplotlib.pyplot import Figure
+        self.figure = fig = Figure()
+        FigureCanvas(fig)
+        return fig.subplots()
+
+
+class react_base:
+    """
+    `react` allows wrapping objects and then operating on them
+    reactly while recording any operations applied to them.  By
```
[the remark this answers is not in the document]
philippjfr: Haha, nice catch. Wasn't careful in my search and replace.

panel/react.py
```diff
@@ -239,7 +252,9 @@ class react_base:
     _method_handlers: dict[str, Callable] = {}
 
     @classmethod
-    def register_accessor(cls, name: str, accessor: Callable[react, Any]):
+    def register_accessor(
+        cls, name: str, accessor: Callable[[react], Any], predicate = Optional[Callable[[Any], bool]]
```
Hoxbro, line 256:
```suggestion
        cls, name: str, accessor: Callable[[react], Any], predicate: Optional[Callable[[Any], bool]] = None
```

panel/expression.py
```diff
@@ -1,30 +1,30 @@
...
+A `expression` instance watches what operations are applied to the object
 and records these on each instance, which are then strung together
 into a chain.
 
-The original input to a `ref` is stored in a mutable list and can be
+The original input to a `expression` is stored in a mutable list and can be
 accessed via the `_obj` property. The shared mutable data structure
 ensures that all `reactive` instances created from the same object can
-hold a shared reference that can be updated, e.g. via the `.set`
+hold a shared expressionerence that can be updated, e.g. via the `.set`
```
Hoxbro, line 15: find and replace gone wrong

Issue #5135, opened 2023-06-18
ChatBox Issues and Feedback


MarcSkovMadsen:
panel==1.1.0

I'm trying out the new `ChatBox` widget which is great we now have. Here is my feedback and some issues.

I WILL UPDATE THIS LIST AS I EXPLORE THE `ChatBox` in a real example.

### Does not work well with `sizing_mode="stretch_width"`

I always set `sizing_mode="stretch_width"` in `pn.extension` because I want and need responsive behaviour. 

```python
import panel as pn

pn.extension(sizing_mode="stretch_width", template="fast")

chat_box = pn.widgets.ChatBox(value=[{"bot": "Welcome"}], primary_name="user")

pn.Column(chat_box).servable()
```

I would expect the chat message to be left or right aligned as in the reference guide. Not stretching to full width, which makes it harder to understand the flow.

![image](https://github.com/holoviz/panel/assets/42288570/532f8255-d727-432c-bbfa-9aa1e355d8ba)

### Default colors do not work well with Fast Template

I almost always use the `Fast` templates as their styling is the most mature. But the message colors does not work well with this template as you can see in the example above. The `Chatbox` blue does not work well with the Panel Blue and the `gray` color of the bot is the same as the card background color. This again makes it hard to follow the flow.

### Overflow when setting fixed height with Fast Template

I think its a very natural thing to do to set a fixed height such that there is is space in your application.

```python
import panel as pn

pn.extension(sizing_mode="stretch_width", template="fast")

chat_box = pn.widgets.ChatBox(value=[{"System": "You are now a welcoming assistant"}, {"Assistant": "Welcome"}, {"User": "Thanks"}]*30, primary_name="User", message_colors={
    "User": "#f0cce2", "Assistant": "#cce2f0", "System": "lightgray", 
}, ascending=True, height=600)

pn.Column(chat_box).servable()
```

![image](https://github.com/holoviz/panel/assets/42288570/e4313b64-6a81-4a6e-908e-49335e7a9541)

### name in a confusing position when `ascending=True`

If I set `ascending=True` to get the chat to work like chatGPT and other chats the name of the user is positioned closer to the previous message. I have a feeling that it would also be better to position it below the message it belongs to.

![image](https://github.com/holoviz/panel/assets/42288570/7b5b2219-34cf-4cfe-982a-8d293ef1bd8c)

```python
import panel as pn

pn.extension(sizing_mode="stretch_width", template="fast")

[...]

PR #5139, opened 2023-06-19
Fix chatbox layout

ahuang11:
Partially addresses 
https://github.com/holoviz/panel/issues/5135

Mainly fixes overflow issues by setting fixed with on all
image


panel/widgets/chatbox.py
```diff
@@ -332,6 +340,7 @@ def _add_scroll_callback(self, obj: Viewable, what: str):
         else:
             code += "\ncolumn.scrollTop = -column.scrollHeight"
 
+        code = "window.scrollTo(0, document.body.scrollHeight);"
```
ahuang11, line 343:
```suggestion
```

Issue #5142, opened 2023-06-19
Provide clear warnings or exceptions if extension is not loaded


MarcSkovMadsen:
I'm on Panel 1.1.0 and trying to help a user with a Plotly based app in 

I simply could not understand why the below app was not working. The server showed no issues. And the errors in the browser console did not give me any hints. After banging my head at the screen for 20 mins and trying upgrading/ downgrading Panel and Plotly I figured I've forgot to add `"plotly"` to `pn.extension`.

Please raise an exception if I've forgot to load my extension. Any new user to Panel would have been unable to solve this issue.

```python
import panel as pn

import plotly.express as px
import pandas as pd

pn.extension(sizing_mode="stretch_width")

data = pd.DataFrame({
    'Stat1': [5, 8, 3, 6],
    'Stat2': [7, 10, 4, 8],
    'Stat3': [9, 12, 6, 10],
    'Stat4': [4, 6, 2, 5]
})

def get_plot(graph_option1, graph_option2, data=data):
    fig = px.bar(data, x=graph_option1, y=graph_option2, title=f'{graph_option1}, {graph_option2}')
    fig.layout.autosize=True
    return fig

graph_options = list(data)
graph_option1 = pn.widgets.Select(name="x-axis Stat:", 
                options=graph_options, value="Stat1")
graph_option2 = pn.widgets.Select(name="y-axis Stat:", 
                options=graph_options, value="Stat2")

bound_plot = pn.bind(get_plot, graph_option1=graph_option1, graph_option2=graph_option2)

plot_pane = pn.pane.Plotly(bound_plot)

pn.template.FastListTemplate(
    title="Reacting to Plotly events",
    sidebar=[graph_option1, graph_option2],
    main=[plot_pane]
).servable()
```

![image](https://github.com/holoviz/panel/assets/42288570/d05a345b-06a4-48ed-8901-e777f6f84cd0)

philippjfr:
Yeah, this is bad. It definitely should have printed a big fat warning to the console.

philippjfr:
Two things I'll do:

1. As long as the app is still starting up, we can in fact load the extensions dynamically.
2. If the component is added after initial load we fix the warning

One question is whether the app itself should display an error somehow. Not 100% clear to me how that should happen.

MarcSkovMadsen:
I would really, really prefer a clear explanatory exception was raised.

Its much simpler to explain, understand and work with.

philippjfr:
[...]

PR #5143, opened 2023-06-20
Allow automatic loading of extensions during launch

philippjfr:
The logic warning users if they correctly loaded an extension was broken in the case where the user called `pn.extension()` without loading any actual extensions. So in this PR we do two things:

1. If the application is launching we can actually allow loading extensions automatically. So if you instantiate a component that requires an extension that wasn't explicitly loaded we now automatically add it to the list of loaded extensions.
2. If the application is fully loaded (and we therefore cannot load any more resources), we correctly warn the user that they forgot to load the extension explicitly. 

Fixes https://github.com/holoviz/panel/issues/5142

### DIFF
--- a/panel/io/state.py
+++ b/panel/io/state.py
@@ -241,6 +241,13 @@ class _state(param.Parameterized):
         return thread_id
 
     @property
+    def _is_launching(self) -> bool:
+        curdoc = self.curdoc
+        if not curdoc or not curdoc.session_context:
+            return False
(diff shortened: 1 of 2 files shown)


Hoxbro (commented):
Looks good. 

One small thing could be to give another warning about an extension not being initialized with `pn.extension` during the `is_launching`  if statement, and it has been done automatically for you, but it is advised to do it regardless.

Hoxbro (approved):

PR #5146, opened 2023-06-20
Correctly check the version for deprecation

Hoxbro:
I was wondering why I still got a warning when running something like  `pn.pane.HTML("", background="red")`

Let's see how the CI likes this. I would suggest we update the last supported version to 1.2 if it was 1.1.


philippjfr:
Indeed, I may even push them back a little further since it hasn't been that long since these were added. I think we didn't anticipate that we would be moving through versions a lot faster.

Hoxbro:
Agree, the pace has been very fast.

philippjfr (approved):
Looks good!

PR #5166, opened 2023-06-22
Unpin numpy

Hoxbro:


panel/pane/vizzu.py
```diff
@@ -120,7 +120,7 @@ def _get_columns(self):
                         columns.append({'name': col, 'type': 'datetime'})
                     elif isinstance(value, str):
                         columns.append({'name': col, 'type': 'dimension'})
-                    elif isinstance(value, (float, np.float, int, np.int)):
+                    elif isinstance(value, (float, np.float_, np.int_, int)):
```
Hoxbro, line 123: `np.float` and `np.int` was aliasing  `float` and `int` and were removed in numpy 1.24.

philippjfr (approved):

PR #5169, opened 2023-06-23
Param 2.0 updates: rename style to styles

maximlt:
Noted the test suite partly failed in the downstream tests after releasing Param 2.0.0a1.


philippjfr (approved):
Thanks!

PR #5175, opened 2023-06-26
fix bundle_resources

sdc50:
Ensure that only the first instance of the `root_url` is replaced, so when `root_url` is `/` that it doesn't replace all `/` in the whole url.


philippjfr:
That fix looks good to me, thanks for the catch!

philippjfr (approved):

PR #5178, opened 2023-06-28
Add ChatBox export method

ahuang11:
Partially addresses https://github.com/holoviz/panel/issues/5135

To convert into this format:
```
[
        {"role": "system", "content": "You are a helpful assistant."},
        {"role": "user", "content": "Who won the world series in 2020?"},
        {"role": "assistant", "content": "The Los Angeles Dodgers won the World Series in 2020."},
        {"role": "user", "content": "Where was it played?"}
    ]
```

However, I'm not sure what the best way to handle widgets / panes. 

@MarcSkovMadsen would love your thoughts on this

### DIFF
--- a/panel/widgets/chatbox.py
+++ b/panel/widgets/chatbox.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 from collections.abc import Iterable
+from enum import Enum
 from typing import (
     Any, ClassVar, Dict, List, Optional, Tuple, Type, Union,
 )
@@ -18,6 +19,17 @@ from ..viewable import Layoutable, Viewable
 from .base import CompositeWidget
 from .button import Button, Toggle
 from .input import StaticText, TextInput
+
+
+class ChatExportFormat(Enum):
+    """
+    The ChatExportFormat enum allows specifying the format to export
+    the chat log in.
+    """
+
+    JSON = "json"
+    TUPLE = "tuple"
+    OPENAI = "openai"
 
 
 class ChatRow(CompositeWidget):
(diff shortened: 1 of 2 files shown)


philippjfr:
I don't like this approach much. We should simply allow both formats as valid input.

philippjfr:
I think I misunderstood what this is meant to achieve. Could you explain what the idea behind this PR is?

ahuang11:
Initially, I think the goal was probably what you had in mind for "We should simply allow both formats as valid input."

However, I think it might be useful to serialize the widgets' values too (and maybe attach the name as context).

philippjfr:
Personally don't see major utility in the `export` functionality, particularly because of the question you already brought up: "how do you "export" rich output and panel components"?

ahuang11:
I think it could be useful; an example off the top of my head:
1. I want to create a plot with OpenAI API
2. It gets a plot
3. I don't like the color but I don't know the color; **somehow** the OpenAI API provides the ColorPicker widget
4. I select the color
5. It creates a plot with that color.

panel/widgets/chatbox.py
```diff
@@ -611,5 +612,37 @@ def clear(self) -> None:
...
+            return "\n".join(self._serialize_message(m) for m in message)
+
+        if hasattr(message, "value"):
+            message = message.value
+        elif hasattr(message, "object"):
+            message = message.object
+        message = str(message)
+        return message
+
+    def export(self, serialize: bool = True) -> List[Dict[str, Union[List[Any], Any]]]:
```
philippjfr, line 626: I think we should add `format` to this and then let users select between OpenAI, tuple and the old dictionary format.

philippjfr:
Left one comment. Happy to finally merge this if it becomes a general way to export/convert to any of the formats.

philippjfr:
Looks great, thank you!

philippjfr (approved):

MarcSkovMadsen:
I think this on is related to my question in https://github.com/holoviz/panel/issues/5276#issuecomment-1641387568.

A ChatBox message can represent many things.

- The raw (string) result returned by the AI
- The raw (string) result + any intermediate thinking, planning etc returned by the AI
- The way we display/ view things returned by the AI

We need to think deeply about what the ChatBox should keep track of.

My current thinking is that we should use a an openai `dictionary` format. I think we should keep it in the format `{"user": "Andrew", "value": "What do you think about the ChatBox?"}` format. The `value` can be strings, list of strings, Panel components etc. The `value` is what we display. If the user needs to keep track of the `context` for openai he/ she just adds the `context` key and sets its value. I.e. we should not start creating an `.export` function that I believe will hard to get right because the objects in the `value` key can be anything.

philippjfr:
I agree with everything you said but also think the `export` functionality will always be helpful, even if the precise internals of the current implementation change significantly because there will always be different conventions used by various APIs and having the ability to convert the output to different formats will be helpful. I do think we should probably punt on implementing any conversion for custom objects for now until we have collected more use cases and perhaps even have a longer term vision for figuring out how we can build prompts and or parse the output from an LLM to be able to render rich outputs.

philippjfr:
My only concern now is that `'json'` is probably not the right name for the current format. I'd rather call it 'legacy' or something because I think we can all agree that we should deprecate it and either settle on one more closely aligned with OpenAI or something like you're suggesting @MarcSkovMadsen.

ahuang11:
Okay I updated JSON -> legacy and replaced JSON as a "user" / "value" dict.

I also agree we should have an export method.

panel/widgets/chatbox.py
```diff
@@ -611,5 +625,52 @@ def clear(self) -> None:
...
+        elif hasattr(message, "object"):
+            message = message.object
+        message = str(message)
+        return message
+
+    def export(
+            self,
+            serialize: bool = True,
+            format: Union[ChatExportFormat, str] = ChatExportFormat.JSON
+        ) -> List[Union[Dict[str, Any], Tuple[str, Any]]]:
```
philippjfr, line 643:
```suggestion
    def export(
        self,
        serialize: bool = True,
        format: Union[ChatExportFormat, str] = ChatExportFormat.JSON
    ) -> List[Union[Dict[str, Any], Tuple[str, Any]]]:
```

ahuang11:
Closing in favor of https://github.com/holoviz/panel/pull/5333

PR #5179, opened 2023-06-28
Make unit and UI tests more robust

philippjfr:
As the title says, this PR attempts to avoid timeouts in UI tests where possible. Timeouts are nasty and either cause flakiness or slow down tests unncessarily.


philippjfr:
pre-commit.ci autofix

PR #5180, opened 2023-06-28
ChatBox support for role/content format input

ahuang11:
To support easily accessing user and message, and directly input into OpenAI API
https://help.openai.com/en/articles/7042661-chatgpt-api-transition-guide


philippjfr (approved):
Happy with this change, we can still think about other formats.

PR #5198, opened 2023-07-02
Param private name space: handle `initialized`

maximlt:
Parameterized objects are going to have a private namespace starting from Param 2.0 (https://github.com/holoviz/param/pull/766). This PR handles the `initialized` attribute that is moving to this namespace.

@philippjfr let me know when you make a dev release including this PR as it will unblock https://github.com/holoviz/param/pull/766.


philippjfr (approved):
Looks good, thank you!

PR #5201, opened 2023-07-03
Improve notebook detection

### LINKED ISSUE #5200: Automatic vscode detection is not always working
Continuation of #5007. 

When running VSCode inside a virtualization/containerized environment, for example, WSL on Windows, the system environment variable we use `VSCODE_PID` to detect if we are running inside a vs code notebook is not set. Below is an example of the `VSCODE_` environment variables set in a notebook environment on WSL. 

```
import os
sorted(k for k in os.environ if "VSCODE" in k)
```

image

image

Hoxbro:
Fixes #5200

As far as I can see, `VSCODE_CWD` is always set in a VSCode Notebook.

### DIFF
--- a/panel/config.py
+++ b/panel/config.py
@@ -787,12 +787,9 @@ class panel_extension(_pyviz_extension):
             return
 
         # Try to detect environment so that we can enable comms
-        try:
-            import google.colab  # noqa
+        if "google.colab" in sys.modules:
             config.comms = "colab"
             return
-        except ImportError:
-            pass
 
         if "VSCODE_CWD" in os.environ or "VSCODE_PID" in os.environ:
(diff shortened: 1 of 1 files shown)


Hoxbro:
@philippjfr, when we are in the part of the codebase, I was thinking if we should change the colab detection from:

``` python
try:
    import google.colab  # noqa
    config.comms = "colab"
    return
except ImportError:
    pass
```
To:
``` python
if "google.colab" in sys.modules:
    config.comms = "colab"
    return
```

Timing in a normal notebook:
image

Check in a colab notebook:
image

philippjfr:
Yep makes sense, didn't know it was automatically imported.

PR #5211, opened 2023-07-05
Add the `editable` attribute to `pn.pane.Perspective`

### LINKED ISSUE #5192: `pn.pane.Perspective` doesn't have the `editable` attribute
The [panel/models/perspective.py](https://github.com/holoviz/panel/blob/main/panel/models/perspective.py#L39) file has an `editable` attribute that defaults to `True`. This attribute overrides the `editable` field in the `plugin_config` of the [panel/models/perspective.ts](https://github.com/holoviz/panel/blob/main/panel/models/perspective.ts#L138-L142) file. However, [panel/pane/perspective.py](https://github.com/holoviz/panel/blob/main/panel/pane/perspective.py#L254) does not include the editable attribute. As a result, developers are unable to change the value of `editable` for the Perspective pane.

xiaoyang-sde:
fixes #5192

This pull request adds the `editable` attribute to `pn.pane.Perspective`, which allows the user to indicate whether items are editable. Otherwise, users are unable to control this behavior for the `Perspective` pane.

perspective

### DIFF
--- a/examples/reference/panes/Perspective.ipynb
+++ b/examples/reference/panes/Perspective.ipynb
@@ -18,7 +18,6 @@
    ]
   },
   {
-   "attachments": {},
    "cell_type": "markdown",
    "metadata": {},
    "source": [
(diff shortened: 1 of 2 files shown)


philippjfr:
Thank you @xiaoyang-sde!

Issue #5137, opened 2023-06-19
Naviation events on a Column


MarcSkovMadsen:
With Panel 1.x we can layout more complex apps which means that a Column could end up containing a lot of `objects` and we are adding them dynamically.

A natural request is then to be able to either `go_to_top`, `go_to_bottom`, `scroll_to_top`, or `scroll_to_bottom`.

Please add.

## Additional context

I very much need this functionality to develop chat interfaces. Panel `ChatBox` interface has a button to scroll to top/ bottom. But in my use cases it does not work. I think the implementation is only tested on the use case where the `ChatBox` is the only component adding a `Column` to the app.

MarcSkovMadsen:
## Hacky Workaround

```python
import panel as pn

pn.extension(sizing_mode="stretch_width", template="fast")

INITIAL_SCRIPT = """
<img src onerror="(()=>{
this.onerror = null;// prevent endless loop if function generates an error
var column = null
if (window.el){
    column = window.el;
} else {
    const element = this.getRootNode().host;
    column = element.getRootNode().host;
    window.el=column;
}
column.scrollTop = column.scrollHeight;
})()">
"""
script = pn.pane.HTML(INITIAL_SCRIPT, sizing_mode="fixed", width=0, height=0, margin=0)

chat_rows = pn.Column(script, scroll=True, height=600, css_classes=["chat_rows"])
message = pn.widgets.TextInput(placeholder="Submit a task", name="Press ENTER to submit the task")


count={"index": 0}

def trigger_scroll():
   
   script.object=str({count["index"]}) + INITIAL_SCRIPT

@pn.depends(message, watch=True)
def add_message(value, user="User"):
    if value:
        index = count["index"]
        count["index"]+=1
        row = pn.Row(str(index) + " | user: " + value)
        chat_rows.append(row)
        message.value=""
        trigger_scroll()

for index in range(0,20):
    add_message(f"this is message {index}")

pn.Column(
    chat_rows, message
).servable()

pn.widgets.ChatBox
```

MarcSkovMadsen:
Streamlit is currently adding scroll to bottom feature. They look to https://github.com/compulim/react-scroll-to-bottom for inspiration.

ahuang11:
This seems to work by connecting Python callbacks to Javascript callback:
https://discourse.holoviz.org/t/panel-jscallback-function/3865/7
[...]

PR #5212, opened 2023-07-05
Add scroll log

ahuang11:
Closes https://github.com/holoviz/panel/issues/5137 and addresses https://github.com/holoviz/panel/pull/5177#discussion_r1246773395

Not sure if this is the best name for it, but pretty happy that I learned how to use ReactiveHTML-ish~

```python
import panel as pn
pn.extension(template="fast")

def add_item(event):
    scroll_log.objects = [*scroll_log.objects, str(len(scroll_log.objects))]

scroll_log = pn.ScrollLog(
    *[str(i) for i in range(0, 5)]
)
button = pn.widgets.Button(name="Add")
button.on_click(add_item)

pn.Column(scroll_log, button).servable()
```

https://github.com/holoviz/panel/assets/15331990/16ad9369-f897-4cf4-9287-7bdb9dab6022

### DIFF
--- a/panel/__init__.py
+++ b/panel/__init__.py
@@ -61,7 +61,7 @@ from .io import (  # noqa
 )
 from .layout import (  # noqa
     Accordion, Card, Column, FlexBox, FloatPanel, GridBox, GridSpec, GridStack,
-    HSpacer, Row, Spacer, Swipe, Tabs, VSpacer, WidgetBox,
+    HSpacer, Row, ScrollLog, Spacer, Swipe, Tabs, VSpacer, WidgetBox,
 )
 from .pane import panel  # noqa
 from .param import Param  # noqa
@@ -82,6 +82,7 @@ __all__ = (
     "Param",
     "Row",
     "Spacer",
+    "ScrollLog",
     "Tabs",
     "Template",
     "VSpacer",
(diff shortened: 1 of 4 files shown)


MarcSkovMadsen:
Awesome

Somehow we are missing a Concept for adding "actions" like AutoScroll, FullScreen, Sort, etc. to a Column without having to implement a new container component each time @philippjfr

philippjfr:
While this functionality is great, I too do not think we should implement a new container component every time we want to add some functionality like this.

ahuang11:
Sure. So as a way forward, should Column should be refactored with ReactiveHTML with an `actions` parameter (like `tools` to Bokeh plots)?

philippjfr:
I think a component as fundamental as Row and Column should still be implemented as a Bokeh model, so we'd have to subclass `Column`.

ahuang11:
Suggestions on what it would be called? `ReactiveColumn`

philippjfr:
Sorry, if I wasn't clear, `pn.layout.Column` would remain the same, the change would be to implement a Typescript subclass of the Bokeh `Column` that adds the features we want.

ahuang11:
Any docs/resources on how create a typescript file that subscripts Bokeh Column?

ahuang11:
Okay, I attempted to convert to Typescript, but I would like some pointers on how to connect all the pieces together.
https://github.com/holoviz/panel/pull/5237/files

philippjfr:
Nice, maybe let's do a session together tomorrow or Tuesday.

ahuang11:
Closed in favor of https://github.com/holoviz/panel/pull/5245

PR #5220, opened 2023-07-05
Fix structure and organization of streamlit migration guide

philippjfr:


MarcSkovMadsen:
The change is fine. Could you explain why you want it?

philippjfr:
Because the TOC doesn't work with the organization you had.

PR #5221, opened 2023-07-05
Rename `reactive.py` to `react.py`

Hoxbro:
PR to see if we can rename `reactive.py` to `react.py` as `panel.reactive` is a prime estate when it comes to names. `react.py` is only chosen because it is not `reactive.py`. The removal version is also not set in stone.

With this PR the following will now give warnings:
- `python -c "import panel as pn; pn.reactive.ReactiveHTML"`
- ` python -c "from panel.reactive import ReactiveHTML"`

image

Running the test local give the following warnings:
``` python-traceback
../lumen/lumen/panel.py:8: 24 warnings
  /home/shh/Repos/holoviz/lumen/lumen/panel.py:8: PanelDeprecationWarning: 'panel.reactive.ReactiveHTML' is deprecated and will be removed in version 1.4, use 'panel.react.ReactiveHTML' instead.
```

Before merging: 
- examples should also be updated.
- Add test which confirm these warning are emitted.


MarcSkovMadsen:
I believe the react.py name will give wrong associations to users. They will Think about the react framework.

Could we call it reactive_html.py?

Hoxbro:
The new filename is definitely not locked, just needed something to show that a rename was possible. 

Some other suggestions could be:
- `component` / `component`
- `builder`

philippjfr:
We could also start thinking about moving some of the core baseclasses into a `core` subpackage.

maximlt:
> deprecated("1.4", "panel.reactive", "panel.react")

I'm still a bit surprised that we decide to remove public APIs in minor versions. If we go down that route, I'd strongly suggest we move to CalVer.

philippjfr:
>I'm still a bit surprised that we decide to remove public APIs in minor versions. If we go down that route, I'd strongly suggest we move to CalVer.

Don't really agree, a well advertised change with sufficient notice to ensure folks can migrate should not have to wait on a major release. We don't actually follow SemVer, (almost) no one does. Real SemVer is basically unachievable given our API surface and I think major releases should be reserved for changes that are actually major and require significant effort to adapt to, not a simple rename that someone can fix in a couple of minutes at most. I get that we could say we keep all deprecated APIs around until such a time we decide to make a major release but realistically that's either a really long time or will cause version inflation where we end up making major releases just because.

maximlt:
> We don't actually follow SemVer, (almost) no one does

I don't entirely disagree with that, I disagree though with the implicit assumption that no one tries to follow SemVer, I actually think it's quite the opposite, to me most projects that adapt this scheme try to follow it, sometimes they fail.

I think we should align on this, as I was pretty sure we tried hard to follow SemVer. Things changed maybe? For instance, I wonder now why it has been decided that the clean up of the Parameterized namespace had to happen for Param 2, if instead it could have been deprecated it and removed a few minor versions later?

Also I think I'd have expected more stability from Panel after majoring from 0.x to 1.x.

philippjfr:
>I don't entirely disagree with that, I disagree though with the implicit assumption that no one tries to follow SemVer, I actually think it's quite the opposite, to me most projects that adapt this scheme try to follow it, sometimes they fail.

Certainly we follow some kind of notion of SemVer but have never been particularly strict about enforcing it. With an API surface as large as ours that would require some automation that actually compares API signatures, this is perhaps worth looking into. So the only thing we roughly follow is the notion that micro releases being for patches, minor releases for enhancements and major releases for major changes including API changes. That loose notion is not sufficient  to satisfy SemVer in any real sense though and even then we (semi-)frequently expand APIs or add minor features in micro releases.  

>I think we should align on this, as I was pretty sure we tried hard to follow SemVer. Things changed maybe? For instance, I wonder now why it has been decided that the clean up of the Parameterized namespace had to happen for Param 2, if instead it could have been deprecated and removed a few minor versions later?

Those aren't really minor changes imo, i.e. we weren't merely renaming but also removing methods entirely and adding new methods which diverged in behavior.

>Also I think I'd have expected more stability from Panel after majoring from 0.x to 1.x.

This is a fair point, technically according to SemVer `0.x` releases can change anything at any time, so I suppose that means thus far we have followed `SemVer`. Also to be fully clear, @Hoxbro submitted this PR mostly as an experiment. I decided against including it in 1.2.0 and have not committed to merging this at all yet. Mostly I'm exploring options.

panel/__init__.py
```diff
@@ -68,6 +68,14 @@
...
 from .widgets import indicators, widget  # noqa
 
+
+def __getattr__(name):
+    if name == "reactive":
+        from .util.warnings import deprecated
+        deprecated("1.4", "panel.reactive", "panel.react")
+        return react
+    raise AttributeError(f"module {__name__!r} has no attribute {name!r}")
+
```
jbednar, line 78: Just thinking out loud here, but if the intention is to help people who have done `import panel as pn ; pn.reactive.ReactiveHTML(...)` once ReactiveHTML is on pn.react and pn.reactive is a callable rather than a module, can't we instead set the `ReactiveHTML` attribute on the `pn.reactive` callable to `pn.react.ReactiveHTML` (etc.)? (Or to a proxy that warns and then returns ReactiveHTML?)

jbednar:
> I always told myself that this was due to the background of the core team in science where reproducibility is very important

We do have a very strong focus on reproducibility, but I would say that is separate from SemVer or any other versioning scheme, because for reproducibility we strongly recommend pinning to _exact_ versions using anaconda-project or conda-lock or similar tools. No matter what versioning policy a project uses, only full pinning will give reproducibility, because there can always be unintended implications from any change in code.

I believe true SemVer is achievable for a tiny tool like one of the GNU utilities or for very, very slowly changing codebases, and it's useful only when any particular user is exposed to a large fraction of the tool's API. For something highly configurable like HoloViews or Panel, the API surface is much larger than most users will ever encounter, and so most "breaking" API changes won't actually break most users, making the strict semantics of SemVer not  be applicable to them. HoloViz developers try to minimize pointless API changes, but if you don't ever want any breaks, don't upgrade!

E.g. how many Panel users actually use ReactiveHTML directly?  1%? 0.005%? Only for those users is a change in the path to that component a breaking change, and needlessly alarming every other user by calling it Panel 3.0 or soon 56.0 or 7856.0 isn't achieving any semantics to the version number, and indeed may as well be CalVer at that point. 

Personally, I appreciate that a human has decided that a 1.3 release is _primarily_ a minor update on 1.2, without completely overhauling what most users use, and to be more wary when 1.3 goes to 2.0. That sort of information is meaningful to me, and I wouldn't want to lose that by moving to CalVer. Anyone can already simply act like the pseudo-SemVer versions they see in reality are like CalVer if they like, and I think actually achieving precise SemVer is reserved for trivial utilities and not expansive libraries like the HoloViz tools, which would be on version 10034 by true SemVer. So I'm happy with how we try to balance things, but there's definitely no way to make _everyone_ happy.

maximlt:
Maybe I shouldn't have mentioned *semver*, *calver* or *reproducibility* as that wasn't really my point (except for semver, I'll come back to that soon), I was more thinking in terms of _backwards compatibility_

I'll come back again to a practical example. Param 1.12.0 deprecated a bunch of methods, just in the docs so without run-time deprecation warnings. Working on the release of Param 2.0, we realized that some of these methods were still used across HoloViz projects, which was evidence that Param users didn't really get the message. That pushed us to keep these methods around for longer, now equipped with run-time deprecation warnings. In this case, despite going for a major release, we went down the route of backwards compatibility. These methods are going to be removed eventually, and you'll find in Param's source code a lot of `# PARAM3_DEPRECATION` comments that mark them. I wrote Param 3 as that's the latest version until which they are meant to be around. Additionally, I **clearly** remember us talking about respecting _semver_ in that case, i.e. not removing them in a 2.X release. As I felt that this was a decision that could still be changed in flight, I chose not to indicate in which version they're planned to be removed (Numpy does that too).

Indeed in our ecosystem semver is not followed closely and in my opinion it's fine if we follow that approach too. That being said, I believe our policy, per project and even more so at the HoloViz level, is very loose and doesn't help anyone, users and maintainers (e.g. should/can we remove these methods from Param in a 2.x release?). Questions like the following ones should have answers:

- Should deprecating an API come together with run-time deprecation warnings? Deprecation message in the API docs?
- How hard should we try to be backwards compatible between minor releases?
- When an API is deprecated, how long should it stick around before removal? Should deprecation messages announce when the API is planned to be removed?
- ...

It's not uncommon for projects to give details about their approach to versioning, deprecation policy, backwards compatibility, etc., see for instance:
- https://docs.djangoproject.com/en/dev/internals/release-process
- https://numpy.org/neps/nep-0023-backwards-compatibility.html

I'd see a lot of value in the HoloViz projects following the same approach and us documenting and applying it. Of course I realize that e.g. Panel and Param are quite different in their nature and one may argue they should be handled differently, except that the truth is that in this particular case most Param users are likely to be Panel users too, and that having a common approach across HoloViz can help the users who leverage more than one tool (which we hope they do!) and help the maintainers working across projects.

To those involved in the discussion here, let me know if you'd be happy with us defining better our policy, in which case I could work on a HEP.

---

> because for reproducibility we strongly recommend pinning to _exact_ versions using anaconda-project or conda-lock or similar tools

To be fair this isn't quite true, I haven't found a single mention of conda-lock across the projects, anaconda-project is used on holoviz.org and examples.holoviz.org but these aren't the most visited sites, and I don't think there's any mention of the tools people are more likely to use like poetry, pip-tools and others.

> E.g. how many Panel users actually use ReactiveHTML directly? 1%? 0.005%?

That's maybe about right, I don't know, what I know though is that it's affecting the majority of the apps I have contributed to for customers :) I agree anyway with you that it wouldn't be great for any backwards incompatible change to force a major release (even if I think you exaggerate a little with version 10034!).

jbednar:
Thanks for the very thoughtful commentary, @maximlt ; I think we are largely in agreement. 

>  was more thinking in terms of _backwards compatibility_

Param has an extremely strong focus on backwards compatibility, because of its position as being downstream to just about every other library. Historically the Param code has changed very, very slowly, so we've been able to keep compatibility high across versions. However, the result has been a lot of cruft buildup, and 2.x is finally the chance to remove this cruft (according to SemVer, whether strict or not).

> Questions like the following ones should have answers:
> - Should deprecating an API come together with run-time deprecation warnings? Deprecation message in the API docs?
> - How hard should we try to be backwards compatible between minor releases?
> - When an API is deprecated, how long should it stick around before removal? Should deprecation messages announce when the API is planned to be removed?

I agree; we should formalize and document that. The answers will likely be different between colorcet and param on the one hand (favoring backwards compatibility nearly always) and other repos on the other (which have much less solid APIs that we are gradually working out).

> I'd see a lot of value in the HoloViz projects following the same approach and us documenting and applying it.

Agreed!

> most Param users are likely to be Panel users too

Depends what you mean by "user"; I'd agree if you mean people who learn Param's API are probably mostly doing so because they use Panel. But Param is used internally in all our projects and thus any breakages affect e.g. hvPlot users even if they do not use Panel, hence wanting a more conservative approach for Param than Panel.

> To those involved in the discussion here, let me know if you'd be happy with us defining better our policy, in which case I could work on a HEP.

I'd love to see such a HEP, and if you e.g. focus it on Panel, then Param can simply exceed it by adding additional constraints.

> because for reproducibility we strongly recommend pinning to _exact_ versions using anaconda-project or conda-lock or similar tools

> To be fair this isn't quite true, I haven't found a single mention of conda-lock across the projects, anaconda-project is used on holoviz.org and examples.holoviz.org but these aren't the most visited sites, and I don't think there's any mention of the tools people are more likely to use like poetry, pip-tools and others.

By "we" I mean me, I guess: :-) https://www.anaconda.com/blog/8-levels-of-reproducibility 

I steer anyone there if they mention reproducibility, which is a feature of environments rather than packages and thus not something that individual package web sites would necessarily cover. Definitely something we should consider addressing more explicitly at holoviz.org, though.

> I agree anyway with you that it wouldn't be great for any backwards incompatible change to force a major release (even if I think you exaggerate a little with version 10034!).

For HoloViews I count 273 releases, of which 75 are non-dev releases, of which 19 are minor or major releases. Let's assume we have succeeded in never making a breaking change in a patch release (we certainly shouldn't!), while also assuming that minor releases will generally have enough code changes in them that _some_ aspect of the API will change, in which case we'd be on version 19 by SemVer. And essentially what will have happened is that we lose any distinction between minor releases and major ones; we'd just have patch vs. major releases.

Basically, what it comes down to is that SemVer uses minor releases exclusively for the case of adding rather than changing functionality, and I'm not sure that case comes up in our libraries often enough to be meaningful, so we use minor release in a different way. So instead I'd say our policy is:

- **patch** releases should never intentionally and rarely, if ever, unintentionally break API. Should be safe to upgrade to the latest patch release if you encounter any problem.
- **minor** releases should not have API changes that affect most users. API changes in minor releases should be rare, but are not unheard of, particularly for recently added functionality whose API is still being refined, or for bugs found in older code that can't be fixed without changing API at least slightly.
- **major** releases will typically break APIs, but should normally include significant new functionality to motivate updating your code. Major breaking API changes should be postponed to a major release and paired with significant user-visible advantages.

I'd be interested in whether other team members think this description is accurate.

philippjfr:
Given that we will likely released `param.rx` and therefore don't expect a naming clash I'll close. I still likely want to clean up some of these core modules but that will happen later.

PR #5232, opened 2023-07-07
Update python version to 3.9 in build

Hoxbro:
Fixes (for future releases) https://discourse.holoviz.org/t/is-there-a-bokeh-3-2-wheel-to-use-with-panel-1-2-0-in-pyscript/5661

Because Bokeh 3.2 is only supported on Python>=3.9, the pyodide Bokeh wheel used version 3.1.1.

EDIT: I have done a manual upload of Bokeh 3.2 to CDN for Panel version 1.2: https://cdn.holoviz.org/panel/1.2.0/dist/wheels/bokeh-3.2.0-py3-none-any.whl


philippjfr (approved):
Thank you!

Issue #5205, opened 2023-07-04
FloatInput widget shows a NaN value when its input is cleared in PyScript


owenlamont:
#### ALL software version info

**PyScript**
Browser: Vivaldi 6.1.3035.111 (Stable channel) (64-bit) 
Panel 1.1.1
Pyscript 2023.05.01
Pyodide 0.23.3
Bokeh 3.1.1

#### Description of expected behavior and the observed behavior

Create a FloatInput widget then clear/delete its value. In PyScript it will now show the value NaN but when its value is queried it returns None - not NaN. In a regular Python environment its displayed value just remains clear rather than showing NaN (which I assume is the correct behavior).

#### Complete, minimal, self-contained example code that reproduces the issue

**index.thml**
```html
<!DOCTYPE html>
<html lang="en">
<head>
    <title>Panel FloatInput</title>
    <meta charset="UTF-8">
    <meta name="viewport" content="width=device-width,initial-scale=1">
    <script
            type="text/javascript"
            src="https://cdn.bokeh.org/bokeh/release/bokeh-3.1.1.js"
    ></script>
    <script
            type="text/javascript"
            src="https://cdn.bokeh.org/bokeh/release/bokeh-widgets-3.1.1.min.js"
    ></script>
    <script
            type="text/javascript"
            src="https://cdn.bokeh.org/bokeh/release/bokeh-tables-3.1.1.min.js"
    ></script>
    <script
            type="text/javascript"
            src="https://cdn.jsdelivr.net/npm/@holoviz/panel@1.1.1/dist/panel.min.js"
    ></script>
    <link rel="stylesheet" href="https://pyscript.net/releases/2023.05.1/pyscript.css"/>
    <script defer src="https://pyscript.net/releases/2023.05.1/pyscript.js"></script>
</head>
<body>
    <py-config src="./pyscript.toml"></py-config>
    <py-script src="./main.py"></py-script>
    <div id="plot"></div>
</body>
</html>
```

**main.py**
```python
import panel as pn
import math

float_input = pn.widgets.FloatInput()

float_input.servable(target="plot")

def float_input_callback(event):
    print(event.new)
    print(event.new is None)

float_input_watcher = float_input.param.watch(
    float_input_callback, "value"
)
```

**pyscript.toml**
```toml
packages = [
    "https://cdn.holoviz.org/panel/1.1.1/dist/wheels/bokeh-3.1.1-py3-none-any.whl",
[...]

PR #5249, opened 2023-07-12
Clarify FloatInput value/value_thottled behavior in docs

owenlamont:
Thought  I'd tweak this documentation that read a little funny to me when I was raising [issue 5205](https://github.com/holoviz/panel/issues/5205). I think the word spinner threw me which I interpreted as busy spinner but maybe that meant spinbox.

I did some testing and found the value and value_throttled behaved seemingly identically in terms of updates. Also couldn't get the mouse wheel to change the value at all as the documentation implies it should.

Happy to have this PR rejected if I'm being too verbose or am writing too inconsistently with the other docs - I'm trying to slowly wrap my head around the preferred writing style and constraints.

Took the liberty of tweaking the .gitignore too as I was using PyCharm at the time and saw its config wasn't ignored while VS Code's was.


owenlamont:
Damn, I don't see how my docs update could've broken any of these tests...  Do these intermittently fail often or is something wrong on the main branch?

Hoxbro:
Don't worry, it is not related to your changes.

The macos-latest unit tests also fail on main and the UI tests are flaky.

philippjfr (approved):
Thanks!

PR #5256, opened 2023-07-14
Do not set Bokeh properties that are readonly

philippjfr:
When testing against Bokeh developer releases I encountered some issues related to setting readonly properties.


Hoxbro:
I vaguely remember removing something like this in one of my first prs. Can you check if this does not break this:

``` python
import panel as pn
pn.extension(throttled=True)
w = pn.widgets.IntSlider()
pn.Column(w, pn.bind(lambda x: x, w))
```

philippjfr:
It does break it.

Issue #4256, opened 2023-01-06
Binder conda installs broken?


cdeil:
I tried to launch a Binder using the link at https://github.com/holoviz/panel/blob/main/README.md which is this one:
https://mybinder.org/v2/gh/holoviz/panel/v0.14.1?urlpath=lab/tree/examples

It errors like this:
```
Step 42/55 : RUN ${KERNEL_PYTHON_PREFIX}/bin/pip install --no-cache-dir -r "binder/requirements.txt"
 ---> Running in 8442f1cd67b1
Obtaining file:///home/jovyan (from -r binder/requirements.txt (line 1))
  Installing build dependencies: started
  Installing build dependencies: finished with status 'done'
  WARNING: Missing build requirements in pyproject.toml for file:///home/jovyan (from -r binder/requirements.txt (line 1)).
  WARNING: The project does not specify a build backend, and pip cannot fall back to setuptools without 'wheel'.
  Checking if build backend supports build_editable: started
  Checking if build backend supports build_editable: finished with status 'done'
  Getting requirements to build editable: started
  Getting requirements to build editable: finished with status 'done'
  Installing backend dependencies: started
  Installing backend dependencies: finished with status 'done'
  Preparing editable metadata (pyproject.toml): started
  Preparing editable metadata (pyproject.toml): finished with status 'done'
Obtaining file:///home/jovyan/binder/jupyter-panel-apps-server (from -r binder/requirements.txt (line 2))
  Preparing metadata (setup.py): started
  Preparing metadata (setup.py): finished with status 'done'
Obtaining file:///home/jovyan/binder/jupyter-voila-docs-server (from -r binder/requirements.txt (line 3))
  Preparing metadata (setup.py): started
  Preparing metadata (setup.py): finished with status 'done'
Collecting pydeck
  Downloading pydeck-0.8.0-py2.py3-none-any.whl (4.7 MB)
     ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 4.7/4.7 MB 24.4 MB/s eta 0:00:00
Collecting pyecharts
  Downloading pyecharts-2.0.0-py3-none-any.whl (146 kB)
     ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 146.0/146.0 kB 168.9 MB/s eta 0:00:00
Collecting idom
  Downloading idom-0.42.0-py2.py3-none-any.whl (89 kB)
[...]

PR #5257, opened 2023-07-14
Fix binder

MarcSkovMadsen:
Makes binder work again

Try it out [https://mybinder.org/v2/gh/holoviz/panel/fix/binder?urlpath=lab/tree/examples](https://mybinder.org/v2/gh/holoviz/panel/fix/binder?urlpath=lab/tree/examples).

Closes

- #4256


.gitignore
```diff
@@ -139,3 +139,13 @@ doc/reference/*
 !/panel/dist/css
 !/panel/dist/css/*.css
 .venv2/
+.npm/*
```
MarcSkovMadsen, line 142: The files below are added to the root of the Panel project on binder. So to not show them in VS code git changes I remove them.

philippjfr:
pre-commit.ci autofix

philippjfr:
@maximlt How much of a pain would this rebase be for you if I merged it right now?

binder/requirements.txt
```diff
@@ -1,12 +1,7 @@
--e .['examples']
+-e .[all_pip,build]
```
MarcSkovMadsen, line 1: binder did not work because non-pip dependencies where added to the `examples`.

Personally I think the packages in `extras_require` should be `pip` installable. Conda and other packages should be somewhere else.  `extras_require` is a `setuptools`/ `pip` thing and should be used for that.

maximlt:
No problem go for it.

binder/requirements.txt
maximlt: > Personally I think the packages in extras_require should be pip installable. Conda and other packages should be somewhere else. extras_require is a setuptools/ pip thing and should be used for that only.

Yeah that's a bug.

philippjfr: Once the hatch stuff is fully worked out they won't have to be listed at all because we will automatically map `pip` dependencies to `conda`.

maximlt: I tracked down the graphviz stuff today and decided to remove all of it anyway in the hatch PR. The graphviz example is gone and the inheritance diagrams in the API docs aren't very useful. I just hope it's not used in other places, I'll see what breaks if it does.

This is where `python-graphviz` was added to the extras: https://github.com/holoviz/panel/pull/2732/files#diff-60f61ab7a8d1910d86d9fda2261620314edcae5894d5aaa236b821c7256badd7R157

philippjfr: Great!

MarcSkovMadsen:
Thanks guys. I'm just checking once more on Binder to be 100% certain this works.

MarcSkovMadsen:
With the Fix Binder PR at least I have one place to point users to a working environment. It might not be optimal and well performing. But its there.

At work we can have a new employee up and running on our Jupyter hub with tools, repositories, python environments, data etc. in 15 minutes. When we install repositories for development we can always just do pip install -e .[dev]. When we want to build the docs we can just do mt build.docs etc. So its very, very easy for people not familiar with coding, git etc. to get started.

I would really, really love to get closer to a situation like that for this project too.

philippjfr:
Indeed, that's the goal of the `hatch` work. For the doc builds to become more tractable we'll have to find ways of doing only partial builds because there's no way getting around the fact that even on a good machine a full docs build will take over an hour.

MarcSkovMadsen:
We also have lots of non-python dependencies. But they are pre-installed on the docker notebook/ vs code images we provide to users.

MarcSkovMadsen:
This PR has been tested in practice with success. 

- `pytests panel/tests` works
- vscode works
- panel apps works

image
image

maximlt:
> At work we can have a new employee up and running on our Jupyter hub with tools, repositories, python environments, data etc. in 15 minutes. When we install repositories for development we can always just do pip install -e .[dev]. When we want to build the docs we can just do mt build.docs etc. So its very, very easy for people not familiar with coding, git etc. to get started.

Oh I'm sure you're not comparing the capabilities of a large energy company with those of an OSS project! Good job anyway, of course not all large companies have such a good set up process, it must be great to work at your place.

> We also have lots of non-python dependencies. But they are pre-installed on the docker notebook/ vs code images we provide to developers.

I have never had to use a docker image to contribute to a Python project. I can easily imagine that would be a barrier to contribution.

---

Overall what we have is:

- complex and large environments to build, there are not many Python projects like Panel or HoloViews that integrate with so many other libraries
- many projects to manage, making sure it's done in a consistent way
- deep and widespread tech debt to handle with pyctdev, nbsite, etc.
- a complicated python packaging/project management landscape, we're going to try `hatch` and to be honest it's hard to know whether it's a good choice yet. It might well be that `pdm` becomes the most popular one and in two years `hatch` is unmaintained and we're back to square one.
- unsurprisingly contributors don't fancy that much helping with infra work

There's no magic trick to improve all of that, you just have to put in a lot of hard work. What would help a lot would be dedicated funding, maybe there are special funding for helping OSS projects with managing their infrastructure and improving their contributor's experience. I doubt any company would fund that work, but who knows!

MarcSkovMadsen:
Just to be sure. Users dont care about or use Docker. They just log into jupyter or vs code on the web.

Its the same that binder and codespaces provides. Ours is just more powerful and with resources and data integrated.

philippjfr (approved):
Looks good, thanks!

Issue #5255, opened 2023-07-14
Vizzu Chart Tooltip


derrickmcs:
Vizzu chart tooltip would be very useful for analysis. Can you please add this as part Vizzu pane?

Example:
https://lib.vizzuhq.com/latest/examples/analytical_operations/drilldown/Area_Area/
`chart.feature('tooltip', true);
`

MarcSkovMadsen:
Hi @derrickmcs 

The link shows 404 error to me. Could you provide the working one?

MarcSkovMadsen:
I believe this is the correct link https://lib.vizzuhq.com/latest/examples/analytical_operations/drilldown/area/

The doumentation for the `chart.feature` can be found here https://lib.vizzuhq.com/latest/reference/classes/Vizzu/?h=feature#feature

derrickmcs:
@MarcSkovMadsen yes, that's correct. Thank you. Sorry about the link. I had the page opened and didn't realize it had changed.

philippjfr:
Support was merged.

PR #5258, opened 2023-07-14
Add Vizzu tooltip support

philippjfr:
Implements https://github.com/holoviz/panel/issues/5255


MarcSkovMadsen:
There vizzu docs have been updated and a few links in our reference notebook do not work any more. Let me document them below.

MarcSkovMadsen:
- vizzu.Anim is now https://lib.vizzuhq.com/latest/reference/modules/Anim/
- vizzu.Config.Chart is now https://lib.vizzuhq.com/latest/reference/interfaces/Config.Chart/

MarcSkovMadsen:
The links in the param docstrings are also invalid.

philippjfr:
Thank you! I was also finally able to remove the forced re-render when the data is updated, which means animations now work when streaming, patching and updating data.

Issue #5115, opened 2023-06-14
pn.Fade, like pn.Swipe, but changes opacity


ahuang11:
I was wondering if it's straightforward to stack two panel objects and have a slider widget to lower the opacity of one to see the underlying one.

philippjfr:
I would prefer to generalize this to a Slideshow/Carousel component with configurable controls and transitions.

philippjfr:
I realize this isn't the same, but `Fade` is imo quite niche.

PR #5261, opened 2023-07-15
Add Fade layout that adjusts opacity of the stacked object

ahuang11:
Closes https://github.com/holoviz/panel/issues/5115

Implementing the scroll to latest taught me a lot about using ReactiveHTML + Typescript for Panel and made me realize it's kind of straightforward. So I implemented Fade, which is similar to Swipe, but instead of clipping, changes the opacity of the "after" object.

```python
import panel as pn
import xarray as xr
import hvplot.xarray

pn.extension()


ds = xr.tutorial.open_dataset("air_temperature")["air"]
pn.Fade(ds.isel(time=0).hvplot(clim=(230, 300)), ds.isel(time=-1).hvplot(clim=(230, 300)))
```

https://github.com/holoviz/panel/assets/15331990/cd9dec47-40a5-465c-9a74-2c77657c7202

Just needs docs.


panel/layout/fade.py
```diff
@@ -0,0 +1,163 @@
...
+from ..reactive import ReactiveHTML
+from .base import ListLike
+
+
+class Fade(ListLike, ReactiveHTML):
+    """
+    The Fade layout enables you to quickly compare two panels laid
+    out on top of each other with a part of the *before* panel shown
+    on one side of a slider and a part of the *after* panel shown on
+    the other side.
```
ahuang11, line 20:
```suggestion
    The Fade layout enables you to quickly compare two panels laid
    out on top of each other; moving the slider to the left will fade out
    the top panel.
```

panel/layout/swipe.py
```diff
@@ -23,7 +23,7 @@ class Swipe(ListLike, ReactiveHTML):
     objects = param.List(default=[], bounds=(0, 2), doc="""
         The list of child objects that make up the layout.""", precedence=-1)
 
-    slider_width = param.Integer(default=5, bounds=(0, 25), doc="""
+    slider_width = param.Integer(default=2, bounds=(0, 25), doc="""
```
ahuang11, line 26:
```suggestion
    slider_width = param.Integer(default=5, bounds=(0, 25), doc="""
```

ahuang11:
Now supports n objects!
https://github.com/holoviz/panel/assets/15331990/db6dda83-47d4-400c-9d76-6fa3710a757b

ahuang11:
Wondering what other transitions ideas you have? Swipe is one, but I don't think it can support more than two objects.

MarcSkovMadsen:
And would it be an idea to make this a Widget with the same Signature/Api as Select, such that it can represent a Select parameter?

ahuang11:
I've been wondering a lot about what makes a widget, a widget, instead of a pane or layout. I imagine this is similar to Tabs and Fade.

Another example would `ChatInterface` be a layout or a widget? Or what about `ChatRow`? Is that a pane or layout?

examples/reference/layouts/Fade.ipynb
```diff
@@ -0,0 +1,122 @@
...
+ "cells": [
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "ed7c1d01-d655-4878-a27d-c6d0e330097b",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "import panel as pn\n",
+    "import pandas as pd\n",
```
MarcSkovMadsen, line 11: `import pandas` can be removed as its not accessed.

MarcSkovMadsen:
@ahuang11 

When I run this in the notebook in VS Code what I see is

image

I.e. the slider is not shown and the heigh of `Fade` is proportional to the number of `objects`.

examples/reference/layouts/Fade.ipynb
```diff
@@ -0,0 +1,122 @@
...
+   "id": "c6187cb3-21a1-4900-be93-02ce3e480ff4",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "import hvplot.pandas\n",
+    "\n",
+    "penguins = pd.read_csv('https://datasets.holoviz.org/penguins/v1/penguins.csv')\n",
+    "\n",
+    "pn.Fade(\n",
+    "    penguins[penguins.species=='Chinstrap'].hvplot.violin('bill_length_mm', color='#00cde1'),\n",
```
MarcSkovMadsen, line 107: The violin plot requires scipy to be installed. It might be an idea to use a plot type that does not require scipy installed as

- Some users might get a `ImportError: univariate_kde operation requires SciPy to be installed.`
- It would make the notebook much faster to load on  Panelite.

panel/layout/fade.py
```diff
@@ -0,0 +1,165 @@
...
+    fade between object 0 to object 1 to object 2.
+    """
+
+    objects = param.List(
+        default=[],
+        doc="The list of child objects that make up the layout.",
+        precedence=-1,
+    )
+
+    slider_width = param.Integer(
```
MarcSkovMadsen, line 33: I don' think that `Fade` should include a slider. I think `Fade` should be a low level interface that can be combined into other elements. For example I might want to have a periodic callback changing the `value`. Or I might want to use the `Player` widget.

Its the same challenge as the `ChatBox`. Do we build the low level components that users can combine? Do we build opinionated, high level interfaces that serves specific use cases? Or do we build both.

So far I believe we've been building the low level components.

I need some guidance from you @philippjfr here on what are our core principles.

panel/layout/fade.py
```diff
@@ -0,0 +1,165 @@
...
+
+    @param.depends("objects", watch=True, on_init=True)
+    def _update_layout(self):
+        for i, obj in enumerate(self.objects):
+            if i < len(self._objects):
+                self._objects[i] = obj
+            else:
+                self._objects.append(obj)
+
+    def select(self, selector=None):
```
MarcSkovMadsen, line 147: Why do we need this method? (Any method added is costly as it needs documentation, maintenance and support on discourse).

Would a user not just set the `objects` to the objects he/ she wants to display?

MarcSkovMadsen:
> Wondering what other transitions ideas you have? Swipe is one, but I don't think it can support more than two objects.

Looking at the api of `Fade` I can see its not a `Gallery` layout. For a `Gallery` layout (or widget) the `value` would be the index of the current object displayed. There can be different transitions when you change the current object. For example as shown in https://support.google.com/webdesigner/answer/9348598?hl=en.

panel/layout/fade.py
ahuang11: Not sure actually. Lots of this is migrated from `Swipe`
https://github.com/holoviz/panel/blob/main/panel/layout/swipe.py#L129

panel/layout/fade.py
ahuang11: This was based off [`Swipe`](https://github.com/holoviz/panel/blob/main/panel/layout/swipe.py#L129) but I can understand linking it to player. Maybe as a compromise it can have a `show_slider` param?

philippjfr:
Will close this particular PR, I think Fade in its current form is a relatively niche functionality. I'd prefer having something like a Slideshow/Carousel component that has configurable transitions.

PR #5266, opened 2023-07-16
Document fast full screen mode

MarcSkovMadsen:
Panel 1.x changes how cards works in the `FastListTemplate`. Before their height was restricted. Now they are `stretch_both`.

Its not documented how to restrict the height of cards while keeping the full screen mode useful. This makes it hard for users.

This PR addresses this issue by adding documentation.


MarcSkovMadsen:
I believe it would be nice to have full screen mode on a `Column` instead one day. That would make it much more general and useful.

philippjfr (approved):

PR #5268, opened 2023-07-16
Fix dom events description

MarcSkovMadsen:
I noticed a bug in the description while working on https://discourse.holoviz.org/t/bootstrap-styling-on-panel-reactivehtml/5705. This fixes it.


philippjfr (approved):
Thanks!

Issue #5213, opened 2023-07-05
Getting per interaction metrics and stats about Panel servers


nishikantparmariam:
## Is your feature request related to a problem? Please describe.

Our use-case is to log and store how Panel servers are performing (cpu, memory), stats (like total and active sessions), how much time an interaction took e.g. when a button is clicked in frontend which triggers some callback in Python, we would like to log how much time it took and do this preferably for each such interaction/widget. 

#### Describe the solution you'd like

I'd prefer a solution which would allow me to hook into Panel and intercept requests to time interactions for logging.

### Patching 

This request is similar to https://discourse.holoviz.org/t/performance-metrics-logging-with-panel/4355 and I've gone through the recommendations there. I patched these two functions in a starter script (that will be run before `panel serve`) - https://github.com/holoviz/panel/blob/f7f9ea921e9f62890fea2d4f6039c05ce1415fd9/panel/reactive.py#L408 and https://github.com/holoviz/panel/blob/f7f9ea921e9f62890fea2d4f6039c05ce1415fd9/panel/reactive.py#L441 like this -

```python3
from panel.reactive import Syncable

original_change_event = Syncable._change_event

def _change_event(*args, **kwargs):
  start = time.time()
  res = original_change_event(*args, **kwargs)
  print(f'End change event : {time.time() - start}s')
  # Do logging stuff etc.
  return res

Syncable._change_event = _change_event
```

Using this I was able to log timings for button clicks, select etc. and this also included the time for any registered callback to complete e.g. `on_click` on Button. Ultimately I'd like to standardize these kind of logs and store them in a fixed format.

Questions -
- Are these methods enough if I want to intercept an interaction or request? 
- Should I be looking into some other methods?
- Is there a cleaner way to do this rather than patching private functions e.g. if Panel can accept callback similar to `pn.state.on_session_created`.

### Admin Panel

[Admin panel](https://panel.holoviz.org/how_to/profiling/admin.html) seems great here but it becomes problematic when using [nginx as reverse proxy with multiple Panel servers](https://docs.bokeh.org/en/latest/docs/user_guide/server/deploy.html#load-balancing) architecture - which we are using.  This makes the `/admin` request go to one of the Panel servers every time you open it and stats seem not that meaningful. Is there a way I can tell `admin` page to show combined stats across Panel servers? 

I was able to mimic similar problem with `num-procs` in Panel where admin page shows stats for different Panel processes (I think?) after multiple refreshes (notice total and active sessions) - 

[...]

Issue #1294, opened 2020-04-23
Issue with name parameter of some widgets


rithwikjc:
#### ALL software version info
OS: Windows 10
Python 3.6.5
pyviz              0.10.3
panel              0.9.5
jupyterlab         2.1.0

#### Description of expected behavior and the observed behavior
Some of the widgets, even though have a `name` parameter, don't display one when the widgets are rendered.
I've seen this with the following widgets:

- checkboxgroup
- checkbuttongroup
- radioboxgroup
- radiobuttongroup

Another issue that I noticed was that the `StaticText` widget- if the `value` parameter is unspecified, the widget's `name` does not get displayed. This is unfortunate, as it removes the possibility of having a bold title displayed without any value specified. However the reverse works fine, that is without a `name` if the `value` is specified it does gets displayed.

#### Complete, minimal, self-contained example code that reproduces the issue

```
import panel as pn
pn.extension()
axes = ['A', 'B', 'C', 'D']
x_text = pn.widgets.StaticText(name='X-Axis') # Does not display anything
x_axis = pn.widgets.RadioButtonGroup(name='X-axis', value=axes[0:1], options=axes)
y_text = pn.widgets.StaticText(value='Y-Axis') # Works fine
y_axis = pn.widgets.RadioButtonGroup(name='Y-axis', value=axes[3:4], options=axes)
pn.Column(text, x_axis, y_axis)
```

jbednar:
I've noticed this too, and been annoyed by the inconsistency.  It's not entirely arbitrary, as the various items in each `group` do have such a label, but I'd personally vote for the name to appear as the label by default, but for the name to be able to be suppressed with an optional setting if people prefer the old behavior.

rithwikjc:
That would be the best solution I imagine. Also it would be best to have such an option across all widgets, not only making it consistent but also allowing flexibility.

I guess instead of having an optional setting/variable to control this behavior it could just be controlled by whether `name` is specified or not, i.e. making it an optional parameter.
Although it would indeed be confusing to have sliders and some other widgets that don't have a `name`, but the trade-off being not having to use another parameter to specify to control something trivial. Do you think that's a good idea?

I'll take a look at the code and see if I can make heads or tails of it, to see if I can make the required changes and put up a pull request for the changes.

jbednar:
[...]

PR #5273, opened 2023-07-17
Add support for middleware for component level interactions & custom name for logging purposes

nishikantparmariam:
Solves https://github.com/holoviz/panel/issues/5213


panel/io/state.py
```diff
@@ -209,6 +212,10 @@ class _state(param.Parameterized):
     _sessions = {}
     _session_key_funcs = {}
 
+    # Middlewares
+    _bokeh_event_middlewares: ClassVar[List[BokehEventMiddleware]] = []
```
philippjfr, line 216: Should these be at the class-level or the instance level? Right now you define them at the class-level but require the user to register them on an instance.

nishikantparmariam:
@philippjfr let me know what else would be required here e.g. tests, more documentation etc. Please point to any references. Thanks!

nishikantparmariam:
@philippjfr please let me know any review comments here. Thanks!!

nishikantparmariam:
Hi, @philippjfr @MarcSkovMadsen please add your review comments / thoughts here so we can push these changes. Thanks!

philippjfr:
I'll do a quick review shortly but also wanted to hear from @maximlt since he's been thinking about using this for some logging.

panel/viewable.py
```diff
@@ -631,6 +633,18 @@ def __repr__(self, depth: int = 0) -> str:
         return '{cls}({params})'.format(cls=type(self).__name__,
                                         params=', '.join(param_reprs(self)))
 
+    @property
+    def component_name(self) -> str:
```
philippjfr, line 637: Would definitely much prefer not adding this. Can you explain what was wrong with simply using the `.name` parameter value?

nishikantparmariam: I totally agree that `.name` can be used but it has a problem - For some components `name` appears in the UI e.g. button. Now consider a page that has two forms. Each form has a button with name `Submit` in the UI. Now, when user clicks one of the button there is no way to know from the logs (logged using this middleware feature) which button was clicked. This property gives developers an option to add custom business specific name that can be used for logging purposes, e.g. two buttons in same page can be `pn.widgets.Button(name='Submit', component_name='SUBMIT_FORM_X')` and `pn.widgets.Button(name='Submit', component_name='SUBMIT_FORM_Y')`.

Button is just one example, idea was to provide users a way so they can choose different name for what appears in UI v/s what is logged and both can be different. Let me know your thoughts.

philippjfr: Yeah, it was an unfortunate mistake of mine in the early design of Panel to use the `name` as the title/label for various widgets.

nishikantparmariam: To make it less confusing with `name`, we can name this property to `logging_name`, `loggable_name`, `component_logging_name`, `log_tag`, etc.

nishikantparmariam: @philippjfr Given above, should I completely remove `component_name` for now, if we do not plan to re-name it to say `loggable_name` or something? 

I'll note another problem with `name` is that by default each widget takes arbitrary name which are not meaningful - https://github.com/holoviz/panel/issues/1294#issuecomment-619232730

maximlt: I agree with you @nishikantparmariam and I share the same use case with multiple Button widgets having `'Submit'` for their `name` value. But I also share the same concerns of Philipp adding a new argument to the constructor. @philippjfr how about abusing `tags` with e.g. `Button(name='Submit', tags=[{'debug_name': 'MyComponent'}])`?

maximlt (commented):
Hi @nishikantparmariam and sorry for the late reply. I have some interest in some similar functionality so here are some questions:

- would you have an example use case for watching property changes?
- do you think the `preprocess`/`postprocess` model is going to be convenient enough to for instance time how long it took to process an event? I see for instance that FastAPI has a different model for setting up middlewares so just checking we're going down the right path 
(https://fastapi.tiangolo.com/tutorial/middleware/#create-a-middleware)

I'm going to do more testing on a larger app that would require some middleware logging and will report back when done.

panel/viewable.py
nishikantparmariam: > how about abusing tags with e.g. `Button(name='Submit', tags=[{'debug_name': 'MyComponent'}])`?

To me at least, it seems great! I wasn't aware of `tags`, as it is a recent addition.

jbednar: Don't we have a distinction between `name` and `label`, where the name is the _default_ label, but an explicit label can override it? Seems like that could apply here, just as it does in other HoloViz contexts.

philippjfr: That seems like a nice way to begin migrating away from using `name` as the parameter controlling widget titles.

nishikantparmariam: Removed `component_name`, I'll use `tags` for my use-case.

nishikantparmariam:
Hi @maximlt,
> would you have an example use case for watching property changes?

Yes, there are many e.g. I want to log if user changed tab of `pn.Tabs` (`active` property would change), user interacted with an `pn.widgets.IntSlider` (`value` property will change). Also, if my components have attached callback (added by `slider.param.watch(my_callback, ['value'])`) which does some computation, I am able to time all of those using the middleware. The `postprocess` part gets executed only after all callbacks have finished processing. One use-case here would be to log the timing of each such UI interaction, and find out which is the slowest of them to improve UX. 


> do you think the preprocess/postprocess model is going to be convenient enough to for instance time how long it took to process an event? 

Thanks for pointing this out! 

<details>
<summary>Here is how one may write a timing middleware:</summary>


```python
class TimingMiddleware(PropertyChangeEventMiddleware):
    def __init__(self):
         """Do any initialization"""

    def preprocess(self, syncable, doc, events):
        self.start_time = time.time()
        self.syncable = syncable # Save for postprocess        

    def postprocess(self):
        process_time = time.time() - self.start_time
        # Do something with self.syncable now

pn.state.add_property_change_event_middleware(TimingMiddleware())
```

</details>

I also explored [django](https://docs.djangoproject.com/en/4.2/topics/http/middleware/) and [expressjs](https://expressjs.com/en/guide/writing-middleware.html) middlewares. 

Then, I realized that in pre/postprocess model, for more advanced cases there is an issue:  If one wants to run pre of middleware1 before pre of middleware2, but want that post of middleware1 runs after post of middleware2. 

<details>
<summary>See Explanation</summary>

Preprocess/postprocess model (current implementation):

```
pre of M1
pre of M2
pre of M3
...main processing...
post of M1
post of M2
post of M3
```

Whereas the other models e.g. FastAPI, django works like this ([**Reference**](https://docs.djangoproject.com/en/4.2/topics/http/middleware/#:~:text=The%20get_response%20callable%20provided%20by%20Django%20might%20be%20the%20actual%20view%20(if%20this%20is%20the%20last%20listed%20middleware)%20or%20it%20might%20be%20the%20next%20middleware%20in%20the%20chain.%20The%20current%20middleware%20doesn%E2%80%99t%20need%20to%20know%20or%20care%20what%20exactly%20it%20is%2C%20just%20that%20it%20represents%20whatever%20comes%20next.)) -

```
pre of M1
  call_next 
  > pre of M2
     call_next
     > pre of M3
         call_next 
             > ...main processing...
     > post of M3
  > post of M2
post of M1
```

</details>

To fix this in pre/postprocesss model, it seems correct thing would be to run `postprocess` in reverse order:

```python
for bokeh_event_middleware in state._bokeh_event_middlewares[::-1]:
      bokeh_event_middleware.postprocess()
```

And, we should mention in the docs that postprocess of current middleware will run after the postprocess of the next added middleware. 

After this, the pre/postprocess model seems as powerful as the other models. Implementation-wise, I found pre/postprocess to be quite easy and intuitive. Let me know your thoughts, if you feel we should for go other type of model, I'll implement that. 

Thanks!

ndmlny-qs:
@nishikantparmariam I'm looking at the middleware you have proposed, and I've run into a few issues that would be solved with an minimum example using your changes. Do you mind posting one?

nishikantparmariam:
@maximlt @philippjfr let me know your comments on https://github.com/holoviz/panel/pull/5273#issuecomment-1704293261 so we can push this forward. I happy to go with either model.

-----

@ndmlny-qs you can try this example and notice that we can find out processing time of button click callback.

```python
import time
import panel as pn
pn.extension()

class TimingMiddleware(BokehEventMiddleware):
    def preprocess(self, syncable, doc, event):
        self.start_time = time.time()      

    def postprocess(self):
        process_time = time.time() - self.start_time
        print('Took :', process_time)

pn.state.add_bokeh_event_middleware(TimingMiddleware())

btn = pn.widgets.Button(name='Foo')

def callback(event):
    # Some computation
    time.sleep(2)

btn.on_click(callback)

pn.Row(btn).servable()
```

ndmlny-qs:
@nishikantparmariam I can see button timings showing up in my terminal. My mouse isn't visible for some reason, but the timings are showing up. Thank you very much for a MRE, it was very helpful.

[button-timing.webm](https://github.com/holoviz/panel/assets/60523020/f530c8f7-ee49-451f-aba1-fc807b9b6c49)

If this feature looks good, and it's a feature Panel would like to have, then let me know and I will write documentation and examples for it.

ndmlny-qs:
I've got a playwright test for this feature. Still need to make a few more tests, and documentation around it.

ndmlny-qs:
I rebased against the main branch and added playwright tests for the middleware components. The new tests are passing locally and in CI.

@philippjfr I can add documentation around using this feature. As @nishikantparmariam mentioned, there are some caveats to how the pre/post processing work that will need to be captured in examples and documentation.

Issue #5290, opened 2023-07-20
Video Pane not respecting height setting


MarcSkovMadsen:
I'm on the current `main` branch of Panel

```python
import panel as pn

pn.extension(sizing_mode="stretch_width")

pn.Column(
    "# Hello",
    pn.pane.Video('https://download.samplelib.com/mp4/sample-5s.mp4', height=300),
    "# By by",
).servable()
```

![image](https://github.com/holoviz/panel/assets/42288570/6f8c7739-5b24-4d7c-a959-4e528c797460)

## Potential Cause

Inspecting the html I can see that the `video` element `min-height` and `min-width` are set to `100%`. Changing `min-height` to `height` manually seems to fix the issue.

MarcSkovMadsen:
H'mm. There seems to be an issue with the `max_height` too.

```python
import panel as pn

pn.extension(sizing_mode="stretch_width")

pn.Column(
    "# Hello",
    pn.pane.Video('https://download.samplelib.com/mp4/sample-5s.mp4', max_height=400),
    "# By by",
).servable()
```

![image](https://github.com/holoviz/panel/assets/42288570/4d46047d-7aa5-44ec-95a8-51c9bbeb1a61)

PR #5296, opened 2023-07-21
fix video max-height

MarcSkovMadsen:
Finalizes the PR #5290

## Test Example

```python
import panel as pn
import requests
from pathlib import Path

def local_file(url):
    name = url.split("/")[-1]
    if not Path(name).exists():
        r = requests.get(url, allow_redirects=True)
        open(name, 'wb').write(r.content)
    return name


FILE = "https://download.samplelib.com/mp4/sample-5s.mp4"
CONFIGS = [
    dict(sizing_mode="fixed"),
    dict(sizing_mode="fixed", height=300, width=300),
    dict(sizing_mode="fixed", max_height=300, width=300),
    dict(sizing_mode="stretch_width"),
    dict(sizing_mode="stretch_width", height=300),
    dict(sizing_mode="stretch_width", max_height=300),
]
layout = pn.Column()
for config in CONFIGS:
    layout.append(f"# {config}")
    layout.append(pn.panel(FILE, **config))
layout.servable()
```

### DIFF
--- a/panel/models/video.ts
+++ b/panel/models/video.ts
@@ -5,6 +5,7 @@ import {HTMLBox, HTMLBoxView} from "./layout"
 export class VideoView extends HTMLBoxView {
   model: Video
   protected videoEl: HTMLVideoElement
+  protected containerEl: HTMLElement
   protected dialogEl: HTMLElement
   private _blocked: boolean
   private _time: any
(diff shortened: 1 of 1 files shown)


MarcSkovMadsen:
I have tested, but then I wanted to merge with `main` and now I cannot build any more because tabulator.ts cannot be built

```bash
panel/models/tabulator.ts:621:33 - error TS4073: Parameter 'resize' of public method from exported class has or is using private name 'bool'.

621   _render_row(row: any, resize: bool = true): void 
{
```

I believe this is not something I've caused.

MarcSkovMadsen:
One question. Is the cause of the `max-height` not working that its not being set on the host/ parent element outside the shadowRoot? Could this be fixed somewhere?

philippjfr:
Indeed build failure is my fault. Will fix shortly. 

max-height is being set on the host via a stylesheet on the component, it just didn't seem to do anything.

PR #5305, opened 2023-07-22
Improve developer guide to meet first time contributors

MarcSkovMadsen:
This would have helped a lot in the EuroPython sprints.

The document can be previewed [here](https://github.com/holoviz/panel/blob/enhancements/first-time-contributor/doc/developer_guide/index.md).


doc/developer_guide/index.md
```diff
@@ -2,6 +2,10 @@
 
 The Panel library is a complex project which provides a wide range of data interfaces and an extensible set of plotting backends, which means the development and testing process involves a wide set of libraries.
 
+This guide describes how to install and configure the development environment fully.
+
+If you have any problems with the steps here, please reach out in the `dev` channel on [Discord](https://discord.gg/rb6gPXbdAr).
```
ahuang11, line 7: What about discourse instead?

MarcSkovMadsen: I would argue that the `dev` channel on Discord is the most *welcoming* place to go for new contributors. Its faster to get help. You can start chatting. Its *extra friendly* to me someone live in the other end. You become a member of the *dev community*.

The only downside I can see is that the history on Discourse is much more discoverable.

ahuang11: I suppose discord you need to make an account which not everyone has.

doc/developer_guide/index.md
```diff
@@ -2,6 +2,10 @@
...
+
 ## Preliminaries
```
philippjfr, line 9: The organization doesn't really make sense anymore, since we now have a Preliminaries section with just one subsections. I'd probably move the pip and conda sections up under preliminaries but am open to other suggestions.

MarcSkovMadsen: I've reorganized as you suggested

philippjfr (approved):
Definitely an improvment and hoping to revisit this asap anyway.

philippjfr (approved):

Issue #5303, opened 2023-07-22
Speed up Pyodide apps by using 'pyc' file


MarcSkovMadsen:
I talked to the Pyodide team at the EuroPython Sprints. And they said we can speed up loading time by using the `/pyc/` file instead of the `/full/` file. The file size is the the same. But its much faster to get the interpreter started.

Maybe it should be configurable via a flag which file to use

This could maybe also be relevant for Panelite. But we would loose the understandable stack traces

PR #5309, opened 2023-07-23
Update pyodide and pyscript and switch to pyc build

philippjfr:
Implements https://github.com/holoviz/panel/issues/5303


MarcSkovMadsen:
Have you tried the latest PyScript with Panel, for me it raised an exception on a simple example that worked with previous version

philippjfr:
We have multiple UI tests that seem to be passing.

PR #5344, opened 2023-07-28
Fix race condition in periodic callback resulting in additional callback executions

### LINKED ISSUE #5339: Race Condition with threaded Periodic Callback can result in additional executions

panel==1.2.1

Looks like due to a race condition, by scheduling a "heavy" periodic callback, it can be executed more than the maximum count when letting panel handle the multithreading. This is because, I believe, the counter is incremented after the callback is finished executing, and with multiple threads many instances can be scheduled while waiting for the count to be met.

In the below example, when you toggle on the periodic with a 0 delay it runs as expected, but increasing the delay beyond the period will result in additional executions of the callback. The larger the delay the worse the effect.

```python
# app.py
import time
import panel as pn
pn.extension(nthreads=0)  # or > 1

def heavy():
    print(cb.counter)
    time.sleep(delay.value)  # simulate lots of work

delay = pn.widgets.IntInput(name='Delay', value=0, start=0, end=5, step=1)

cb = pn.state.add_periodic_callback(heavy, 200, count=5, start=False)
toggle = pn.widgets.Toggle(name='Run', value=False)
toggle.link(cb, bidirectional=True, value='running')

pn.Column(delay, toggle).servable()
```

Run with:
```shell
panel serve app.py
```

N/A

N/A

Lnk2past:
So I am not convinced I properly ran all of the tests locally, but here is a fix that seems to be working as expected. If there is a clearer path to victory then please let me know. Fixes #5339

Not only did the location of the increment need to change, but checking the counter against the count and signaling a stop also needed to happen *before* the callback is executed to prevent executions from being scheduled. This now appears to work as expected for any number of threads.

The logic works as follows:

1. first we check if the periodic is running; this needs to happen to prevent the counter from being incremented after a stop has already been signaled (otherwise, subsequent adds of the periodic do not start at 0). This is important for the case were the count is specified as some number greater than the number of threads allocated.
2. increment the counter
3. check if we need to stop
4. execute the callback if the periodic is still running

A side effect of this is that during execution of the callback, the counter has already been incremented.

### DIFF
--- a/panel/io/callbacks.py
+++ b/panel/io/callbacks.py
@@ -82,7 +82,11 @@ class PeriodicCallback(param.Parameterized):
         from .state import set_curdoc
         try:
             with set_curdoc(self._doc):
-                cb = self.callback()
+                if self.running:
+                    self.counter += 1
+                    if self.counter > self.count:
+                        self.stop()
+                cb = self.callback() if self.running else None
         except Exception:
             cb = None
         if post:
@@ -98,7 +102,6 @@ class PeriodicCallback(param.Parameterized):
         if not self._background:
             with edit_readonly(state):
                 state._busy_counter -= 1
-        self.counter += 1
         if self.timeout is not None:
             dt = (time.time() - self._start_time) * 1000
             if dt > self.timeout:


Lnk2past:
I'll add the necessary tests to cover this code some time good weekend.

philippjfr:
Thanks for this, I think incrementing before execution is fine.

Lnk2past:
Sorry I dropped the ball on finishing this out! Started a new job and everything is chaos :)

el-abcd:
I'm wondering if there is an issue with callback timers in version 1.2.2 (if the "count" parameter is left blank). 
I have a UI with a counter that ticks up every second, and it seems like it stopped working with panel version 1.2.2 (I reverted to 0.14.3, 1.1.0, and 1.2.1 and see it working well there).  
I DID see the counter start working if I set the "count" parameter ...

periodic_cb = pn.state.add_periodic_callback(server_tick_increment, start=True, period=1000, count=1000)
instead of:
periodic_cb = pn.state.add_periodic_callback(server_tick_increment, start=True, period=1000)

This looks like the only change in callbacks.py in release 1.2.2 (IIUC).  
I'm not sure, but I'm wondering if perhaps self.count = None if no value is passed in, and some of the comparisons (like "if self.counter > self.count:") might have errors in that case?  
If so, perhaps check if self.count = None before that and handle that case differently?  

Regards,
Eric

Hoxbro:
@el-abcd, you are completely correct. This is why I have opened a PR https://github.com/holoviz/panel/pull/5490, which should fix this regression. 

We will make a new release at the start of next week.

Issue #5348, opened 2023-07-29
Mention updating in place in ReactiveHTML's scripts does not sync Python's variable


ahuang11:
Here, `_log` does not trigger because data is modified in place, e.g. `data.index += 1`:
```
from panel.reactive import ReactiveHTML
import param

class JSSlideshow(ReactiveHTML):

    index = param.Integer(default=0)

    _template = """<img id="slideshow" src="https://picsum.photos/800/300?image=${index}" onclick="${script('click')}"></img>"""

    _scripts = {'click': 'data.index += 1'}

    @param.depends("index", watch=True)
    def _log(self):
        print(self.index)

ss = JSSlideshow(width=800, height=300)
ss
```

To make it work, `_scripts = {'click': 'data.index = data.index + 1'}`

ahuang11:
Fixed now
```
from panel.reactive import ReactiveHTML
import param

class JSSlideshow(ReactiveHTML):

    index = param.Integer(default=0)

    _template = """<img id="slideshow" src="https://picsum.photos/800/300?image=${index}" onclick="${script('click')}"></img>"""

    _scripts = {'click': 'data.index += 1'}

    @param.depends("index", watch=True)
    def _log(self, *args):
        print(args)
        print(self.index)

ss = JSSlideshow(width=800, height=300)
ss
```

But I think it triggers three times.

PR #5349, opened 2023-07-29
Add blurb about in place variable modification in ReactiveHTML

ahuang11:
Closes https://github.com/holoviz/panel/issues/5348

I updated the examples to not use `v += 1` in favor of `v = v + 1`


MarcSkovMadsen:
I really Think v+=1 not working is a bug and should be fixed. Changing the example just hides the problem.

philippjfr:
Yes, indeed, this is a bug.

ahuang11:
I'm wondering, if this is considered a bug, whether the following is too?

```python
import panel as pn
pn.extension()

def log(event):
    print(event.new)

x = pn.widgets.Select(options=["a", "b", "c"], value="a")
x.param.watch(log, "options")

x.options += ["c", "d"]
```

philippjfr:
That scenario isn't really equivalent, modifying mutable containers inplace vs. assigning to a non-mutable integer.

PR #5360, opened 2023-08-01
Ensure DataModel updates are correctly scheduled

philippjfr:
DataModel updates were executed even if the Document was locked. Here we correctly schedule these updates.

### DIFF
--- a/panel/io/datamodel.py
+++ b/panel/io/datamodel.py
@@ -1,5 +1,7 @@
 import weakref
 
+from functools import partial
+
 import bokeh
 import bokeh.core.properties as bp
 import param as pm
@@ -8,7 +10,9 @@ from bokeh.model import DataModel
 from bokeh.models import ColumnDataSource
 
 from ..reactive import Syncable
-from .notebook import push_on_root
+from .document import unlocked
+from .notebook import push
+from .state import state
 
 
 class Parameterized(bokeh.core.property.bases.Property):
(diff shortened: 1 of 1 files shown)


panel/io/datamodel.py
```diff
@@ -192,13 +196,30 @@ def cb_param(*events):
...
             if root:
-                push_on_root(root.ref['id'])
+                ref = root.ref['id']
             elif tags:
                 ref = tags[0].split('__ref:')[-1]
-                push_on_root(ref)
+            else:
+                ref = None
+
+            if ref:
```
philippjfr, line 208:
```suggestion
            if ref and ref in state._views:
```

ahuang11:
I think something might be one off:

```python
async def echo_callback(
    value: Any, user: str, chat_interface: pn.widgets.ChatInterface
):
    await asyncio.sleep(2)

    message_echo = ""
    for character in value:
        message_echo += character
        await asyncio.sleep(0.1)
        yield {
            "avatar": "https://user-images.githubusercontent.com/42288570/246667324-5cf26789-765f-4f76-a8bf-49309d2ae84f.png",
            "user": "OpenAI",
            "value": message_echo,
        }


chat_card = pn.widgets.ChatCard(callback=echo_callback)
chat_interface = pn.widgets.ChatInterface(value=chat_card)
chat_interface.servable()
```

https://github.com/holoviz/panel/assets/15331990/a519b8fc-579c-477c-afe7-5bd54172968c

philippjfr:
I'm not convinced the issue you're reporting is related to this PR. I'll merge for now and then do some more testing.

PR #5369, opened 2023-08-04
Patch Cannot read properties of undefined for Column

ahuang11:
Actually fixes #5365

Previously:
Cannot read properties of undefined (reading 'classList')
image

After:
image

Scroll arrow still works:
image


philippjfr (approved):
Thanks, looks good! Failures are unrelated/transient.

Issue #5375, opened 2023-08-08
Single newline `\n` in pn.pane.Markdown doesn't work


ahuang11:
```python
import panel as pn
pn.extension()
pn.pane.Markdown("Hello\nSomething")
```
I expect Something to be rendered on a newline.
<img width="376" alt="image" src="https://github.com/holoviz/panel/assets/15331990/b299a4c6-8a5c-4ac2-9028-66d94f48a388">

```python
import panel as pn
pn.extension()
pn.pane.Markdown("Hello<br>Something")
```
works as expected

ahuang11:
printing the output HTML in transform_objects; seems like <br> is missing.

<img width="466" alt="image" src="https://github.com/holoviz/panel/assets/15331990/647537df-1d54-4950-b5d1-e6256a35c90f">

Adding two newlines separates into two tags.
<img width="463" alt="image" src="https://github.com/holoviz/panel/assets/15331990/b140901e-ad1b-422d-ae54-d3bb1170b381">

MarcSkovMadsen:
I think its because by default `markdown-it` (and `markdown`) parser does not convert `\n` to a break.

See https://github.com/markdown-it/markdown-it#init-with-presets-and-options

![image](https://github.com/holoviz/panel/assets/42288570/df9951a4-378e-4728-9726-7e7b7ce08815)

MarcSkovMadsen:
If you add `options_update` in the Panel code it works

![image](https://github.com/holoviz/panel/assets/42288570/440fba0e-1c6e-4dee-a9ea-4503b263c741)

![image](https://github.com/holoviz/panel/assets/42288570/118a908d-d6cb-4462-b31f-6a77d899854a)

I think we should support an options (dict) parameter. That can  be used to set the `MarkdownIt` `options_update`.

PR #5376, opened 2023-08-08
Fixes displaying single newlines in output markdown and add `renderer_options`

ahuang11:
Fixes https://github.com/holoviz/panel/issues/5375
Before
image

After
image

Primary use case atm:
Langchain outputs newlines as `\n` so I have to do this:
```python

    def on_llm_new_token(self, token: str, **kwargs) -> None:
        self._entry = self.chat_interface.stream(
            token.replace("\n", "<br>"), self._entry
        )
        return super().on_llm_new_token(token, **kwargs)
```


philippjfr:
Looks good, but could you please also update the docs in the reference notebook?

ahuang11:
Added

philippjfr:
Thank you!

philippjfr (approved):

PR #5380, opened 2023-08-09
Updating pytest-playwright install instructions to use conda

owenlamont:
With a lot of help from @mxschmitt I created a conda recipe for [pytest-playwright](https://anaconda.org/microsoft/pytest-playwright) which is now available on conda as of [release 0.4.2](https://github.com/microsoft/playwright-pytest/releases/tag/v0.4.2) from the Microsoft channel ([playwright-python](https://github.com/microsoft/playwright-python) already had a [conda recipe](https://anaconda.org/Microsoft/playwright)).

As such I've updated the developer instructions to install pytest-playwright via conda instead of pip in the development environment (this should install playwright-python as a dependency).

I have tested I can still run the UI tests with these conda packages.

### DIFF
--- a/doc/developer_guide/index.md
+++ b/doc/developer_guide/index.md
@@ -58,9 +58,20 @@ doit develop_install -c pyviz/label/dev -c conda-forge -c bokeh -o build -o test
 
 The above command installs Panel's dependencies using conda, then performs a pip editable install of Panel. If it fails, `nodejs>=14.0.0` may be missing from your environment, fix it with `conda install -c conda-forge nodejs` then rerun above command.
 
-If you also want to run the UI tests run the following:
+If you also want to run the UI tests you'll need to install pytest-playwright with Conda:
 ``` bash
 conda install pytest-playwright -c microsoft -c conda-forge
+```
+
+or with PyPi:
+
+``` bash
+pip install pytest-playwright
+```
+
+then run:
+
+``` bash
 playwright install chromium
 ```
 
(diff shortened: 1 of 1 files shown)


Hoxbro:
Thank you, and great initiative in getting the package on to conda!

owenlamont:
😞 I realise I created the conda package with the wrong name. It's supposed to be playwright-pytest, not pytest-playwright. Consider this a work in progress, I'll see if the Microsoft channel maintainers want to fix my mistake.

Hoxbro:
I think pytest-playwright is correct - https://pypi.org/project/pytest-playwright/

owenlamont:
Cheers, yeah @mxschmitt confirmed pytest-playwright as the package name is correct. Sorry when I noticed the repo and package names were reversed I thought I'd messed up.

owenlamont:
Now I look at the existing code closer I can see maybe you could conda install pytest-playright just by tweaking the doit line call:

```
doit develop_install -c pyviz/label/dev -c conda-forge -c bokeh -o build -o tests -o recommended
```

 I noticed later pytest-playwright is already a dependency in setup.py in the _ui_ dependencies. If you added _-c microsoft_ and _-o ui_ to that call would that install those ui dependencies? I'm afraid I'm not familiar with _pyctdev_ but can test that later if that was a better way to install it.

Hoxbro:
We plan to overhaul the installation procedure soon, so I think we should keep the changes here on a documentation level. 

I still would like the option of `pip install` to be mentioned in docs along with the conda install option.

owenlamont:
Sorry for the slow follow up, let me know if this reads better to you @Hoxbro

philippjfr:
Looks good, we're hopefully overhauling these docs again soon.

philippjfr (approved):

PR #5386, opened 2023-08-10
Authorization checks in server.py

### LINKED ISSUE #3179: Enable serving admin site at custom and secure url
The new admin page looks really, really nice.

But as far as I can see everybody can access it if enabled. I would like to be able to restrict access.

For example via a username and password. Or alternatively at a custom "token" url by specifying something like `--admin=879B996862F69D8561D1B369F7E44` and then the admin page is available at `/879B996862F69D8561D1B369F7E44`.

My use case would be for example awesome-panel.org where I finally could start getting some insights :-)

ndmlny-qs:
This commit adds a single parameter used in the
`config.authorize_callback` that allows the user supplied method to check if an app user is authorized to view the requested app at the given path.

Resolves #3179

### DIFF
--- a/panel/io/server.py
+++ b/panel/io/server.py
@@ -8,6 +8,7 @@ import datetime as dt
 import gc
 import html
 import importlib
+import inspect
 import logging
 import os
 import pathlib
@@ -488,7 +489,6 @@ class DocHandler(BkDocHandler, SessionPrefixHandler):
                 resources = Resources.from_bokeh(self.application.resources())
                 auth_cb = config.authorize_callback
                 if auth_cb:
-                    import inspect
                     auth_cb = config.authorize_callback
                     auth_params = inspect.signature(auth_cb).parameters
                     auth_args = (state.user_info,)


philippjfr:
Thanks, this makes sense but it's also not backward compatible. Could we maybe add a try/except or explicitly check the signature of the function to make sure the old signature is still supported?

Also we should update the docstring of the config. authorize_callback parameter.

ndmlny-qs:
thanks @philippjfr I'll make those changes and update the docstring. If it makes sense, I'll also add a  blurb to the documentation page with an example.

doc/how_to/authentication/user_info.md
```diff
@@ -38,3 +38,29 @@ The auth template must be a valid Jinja2 template and accepts a number of argume
...
+    "user2": ["/app1"],
+}
+
+def authorize(user_info, request_path):
+    current_user = user_info['username']
+    current_path = parse.urlparse(request_path).path
+    with open('users.txt') as f:
+        valid_users = f.readlines()
+    user_paths = authorized_user_paths[current_user]
+    if current_user in valid_users and current_path in user_paths:
```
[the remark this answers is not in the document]
ndmlny-qs: makes sense to return as fast as possible, thanks for the suggestion

panel/io/server.py
```diff
@@ -485,7 +485,15 @@ async def get(self, *args, **kwargs):
                 token = session.token
             logger.info(LOG_SESSION_CREATED, id(session.document))
             with set_curdoc(session.document):
-                if config.authorize_callback and not config.authorize_callback(state.user_info):
+                import inspect
+                authorize_callback_signature = inspect.signature(config.authorize_callback)
```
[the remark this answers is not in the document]
ndmlny-qs: excellent catch. i moved the check before the import of `inspect`

panel/io/server.py
```diff
@@ -485,27 +485,28 @@ async def get(self, *args, **kwargs):
...
+                if config.authorize_callback:
+                    import inspect
+                    authorize_callback_signature = inspect.signature(config.authorize_callback)
+                    authorize_callback_parameters = authorize_callback_signature.parameters
+                    authorized = False
+                    if len(authorize_callback_parameters) == 1:
+                        authorized = config.authorize_callback(state.user_info)
+                    if len(authorize_callback_parameters) == 2:
+                        authorized = config.authorize_callback(state.user_info, self.request.path)
+                    if not authorized:
```
philippjfr, line 497: This branching doesn't look right, if `authorized` is true it won't hit the else branch below.

ndmlny-qs: yup, will fix...was attempting to juggle too many things atm

ndmlny-qs: I can move the creation of the `page` object above the `if config.authorize_callback`, but this always assumes the request and requester are authorized. If a user supplies a `config.authorize_callback` and they are authorized, then the initial `page` creation is returned, however, if they are not authorized then the orginal `page` object is updated to be the `ERROR_TEMPLATE` object.

I'll make a new commit to see if you think this is okay or not. If not, then I'm happy to write it as a `try/except/finally` as you suggested above.

philippjfr: Yeah, this is still not correct. I'll push a fix.

el-abcd:
Hi, 

I was trying to add oath to a small panel app (to run in "google cloud run").  
It seems like with panel==1.2.2 that the authorize() method MUST have BOTH parameters?   

def authorize(user_info, request_path):

i.e. when I tried 
def authorize(user_info):
which is working for panel=1.2.1 I see:

###
2023-09-11 14:07:44,242 Uncaught exception GET /220514_01_awesome_panel (::1)
HTTPServerRequest(protocol='http', host='localhost:5006', method='GET', uri='/220514_01_awesome_panel', version='HTTP/1.1', remote_ip='::1')
  File "/Users/ericlee/.pyenv/versions/3.10.4/lib/python3.10/site-packages/tornado/web.py", line 1704, in _execute
    result = await result
  File "/Users/ericlee/.pyenv/versions/3.10.4/lib/python3.10/site-packages/panel/io/server.py", line 517, in get
    raise RuntimeError(
RuntimeError: Authorization callback must accept either one or two arguments.
###

Perhaps the check here:
https://github.com/holoviz/panel/commit/b6d6b347472fd5af3d867cf20e7db5d3f9151589#diff-aaf5dc2e57ff38e7660dbe2c7e08cd2b8a34242b382d40f049987b95f605e628R513
should be changed from 
if len(auth_params) == 2:
to something like 
if len(auth_params) in [1,2]:
?

Regards,
Eric

maximlt:
Hi @el-abcd and thanks for reporting that. Indeed it seems there's a bug! I believe your suggestion not to be quite the right one, when `len(auth_params) == 1` you do not want to expand to expand `auth_args`. The logic could be rewritten as:

```python
                    if len(auth_params) == 1:
                        auth_args = (state.user_info,)
                    elif len(auth_params) == 2:
                        auth_args == (state.user_info, self.request.path)
                    else:
                        raise RuntimeError(
                            'Authorization callback must accept either one or two arguments.'
                        )
```

Would you be interested in making that change in a PR?

el-abcd:
Hi Maxime,

Yes, I can give that a try.  
It might take me a couple days to get a development environment setup, etc.  but I will see if I can make the PR, etc.  

Eric

ndmlny-qs:
@el-abcd I see the problem, and can make a fix tomorrow. thanks for the report

PR #5390, opened 2023-08-13
ChatFeed notebook improvements

MarcSkovMadsen:
Improvements based on review of the `ChatFeed` notebook.

Overall the notebook is great. I've made some changes to make it easier to see what goes on when I run the notebook interactively.

- Put `.send` in its own cell
- Add `await sleep` in some places

To make the Feed more fun and engaging

- I've changed the user emoji from 🧑 to 👤. The 🧑 represents a genderless person. The color and not least smile helps put your into a positive mood.
- I've changed the `callback` emoji from 🤖 to the below avatar. 

I think this avatar better signals positivity and fun. If you agree we should get the `.png` rescaled and uploaded to the holoviz cdn for optimal performance.

## Assets

I've made some gifs to also display the OPENAI examples

openai-acreate

bot-conversation

These should be uploaded to our CDN.


examples/reference/widgets/ChatFeed.ipynb
```diff
@@ -258,14 +274,48 @@
...
+    "chat_feed"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# TODO: IT'S LIKE THE its not sleeping but just displaying right away. The spinner is not displayed either. Andrew should look at this. `async` functions with `return` should also work.\n",
```
MarcSkovMadsen, line 309: It like this example with an `async` function with a `return` statement does not work correctly. Its like the `await sleep` is never run. And the `LoadingIndicator` is not triggered either.

Would you take a look @ahuang11 ? Thanks.

ahuang11:
>I've changed the user emoji from 🧑 to 👤.
I think you mean from 👤 to 🧑*; I actually prefer the shadowy figure since a similar image is used as the default in most social media, but I don't really have strong feelings for this.

In regards to the callback avatar, I prefer not to use a URL for an avatar since it requires the user to have an internet connection. Also, it's using a Ubuntu(?) mascot(?) which I don't think should be associated with Panel.

ahuang11:
I will take a closer look tonight and hopefully merge so there's not so many diverging PRs at once.

examples/reference/widgets/ChatFeed.ipynb
ahuang11: I'm a bit confused with this request. Since `respond=False`, the callback will not execute.

ahuang11 (approved):

ahuang11:
I appreciate your comments and thanks for adding the GIFs; I made tweaks based on your suggestions and merging now.

PR #5392, opened 2023-08-13
ChatEntry notebook improvements

MarcSkovMadsen:
FYI @ahuang11

### DIFF
--- a/examples/reference/widgets/ChatEntry.ipynb
+++ b/examples/reference/widgets/ChatEntry.ipynb
@@ -99,7 +99,7 @@
     "    ChatEntry(\n",
     "        value=fig,\n",
     "    ),\n",
-    ")\n"
+    ")"
    ]
   },
   {
@@ -140,7 +140,10 @@
     "        value=\"Yes. I want to hear some beat boxing\", user=\"Music Lover\", avatar=\"🎸\"\n",
     "    ),\n",
     "    ChatEntry(\n",
-    "        value=pn.Column(\"Here goes. Hope you like this one?\", pn.pane.Audio('https://upload.wikimedia.org/wikipedia/commons/d/d3/Beatboxset1_pepouni.ogg')),\n",
+    "        value=pn.Column(\n",
+    "            \"Here goes. Hope you like this one?\",\n",
+    "            pn.pane.Audio('https://upload.wikimedia.org/wikipedia/commons/d/d3/Beatboxset1_pepouni.ogg')\n",
+    "        ),\n",
     "        user=\"Beat Boxer\",\n",
     "        avatar=\"🎶\"\n",
     "    )\n",
--- a/examples/reference/widgets/ChatEntry.ipynb
+++ b/examples/reference/widgets/ChatEntry.ipynb
@@ -59,14 +59,18 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "pn.widgets.ChatEntry(value=\"Good to see you!\", user=\"Happy User\", avatar=\"😀\")"
+    "from panel.widgets import ChatEntry\n",
+    "\n",
+    "ChatEntry(value=\"Hi!\")"
    ]
   },
   {
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "Although `ChatEntry` is commonly used for `value`s of `str` type, it can support Panel panes, widgets, and layouts too!"
+    "Although `ChatEntry` is commonly used for `value`s of `str` type, it can support any Python object that Panel can display!\n",
+    "\n",
+    "# Todo: Andrew I can see that currently the `value` is restricted. I believe it should support any Python object that panel supports. Including async generator functions. Do you agree? I also think you should consider whether `value` or `object` is the correct parameter name. We use `value` for widgets  and this is not a widget. We use `object` for components that display element which is what the `ChatEntry` does."
    ]
   },
   {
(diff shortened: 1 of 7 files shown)


MarcSkovMadsen:
I've change the `value` parameter type back to `Parameter` such that the `ChatEntry` can display any object that panel can display. Not just a subset.

```python
import panel as pn
from panel.widgets import ChatEntry
import numpy as np
from asyncio import sleep

def mpl_figure():
    fig = plt.figure()
    ax = fig.add_subplot(111)
    ax.plot(np.random.rand(10, 2))
    plt.close(fig)
    return fig

async def async_func():
    return "hello"

async def async_generator():
    await sleep(1)
    yield "hello"
    await sleep(2)
    yield "hello world"

pn.Column(
    ChatEntry(value=mpl_figure()),
    ChatEntry(value=async_func),
    ChatEntry(value=async_generator),
).servable()
```

examples/reference/widgets/ChatEntry.ipynb
```diff
@@ -59,14 +59,16 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "pn.widgets.ChatEntry(value=\"Good to see you!\", user=\"Happy User\", avatar=\"😀\")"
+    "from panel.widgets import ChatEntry\n",
```
MarcSkovMadsen, line 62: I import `ChatEntry` to make the code shorter and easier to read.

examples/reference/widgets/ChatEntry.ipynb
```diff
@@ -75,8 +77,73 @@
...
+    "dataframe = pd.DataFrame(data, columns=list(\"AB\"))\n",
+    "\n",
+    "fig = plt.figure(figsize=(5,2.5))\n",
+    "ax = fig.add_subplot(111)\n",
+    "ax.plot(data)\n",
+    "plt.close(fig)\n",
+    "\n",
+    "pn.Column(\n",
+    "    ChatEntry(\n",
+    "        value=pn.widgets.Button(name=\"CLICK ME\", button_type=\"primary\"),\n",
```
MarcSkovMadsen, line 94: This is the example where the `Button` displays as `False`?

panel/widgets/chat.py
```diff
@@ -204,8 +204,8 @@ class ChatEntry(CompositeWidget):
         Whether to display the timestamp of the message.
     """
 
-    value = param.ClassSelector(class_=(Viewable, str, int, float, _FileInputMessage), doc="""
-        The message contents. Can be a string, pane, widget, layout, etc.""")
+    value = param.Parameter(doc="""
```
MarcSkovMadsen, line 207: I want to be able to display any object that Panel can display. Not only a limited subset 😄

examples/reference/widgets/ChatEntry.ipynb
ahuang11: I changed it from Parameter because of this; Parameter casts widgets into Str

panel/widgets/chat.py
ahuang11: If its Parameter most widgets input are broken that is why I changed it to limited subset

examples/reference/widgets/ChatEntry.ipynb
MarcSkovMadsen: Is that a bug @philippjfr ?

panel/widgets/chat.py
MarcSkovMadsen: I will try a workaround now by wrapping it in a bound function.

MarcSkovMadsen: I can see its already wrapped. The below is a minimum reproducible, test example 

```python
import panel as pn
import param

class Entry(pn.widgets.base.Reactive):
    value = param.Parameter()

entry = Entry(
    value=pn.widgets.Button(name="CLICK ME", button_type="primary"),
)
assert isinstance(entry.value, pn.widgets.Button)
```

Would you consider this a bug @philippjfr ?

MarcSkovMadsen: When debugging, I can see that there is an attribute `_ignored_refs` that can be set to `('value',)`. I believe that will solve the problem.

MarcSkovMadsen: It works. I'm closing this one.

examples/reference/widgets/ChatEntry.ipynb
MarcSkovMadsen: I've solved this by setting `ChatEntry._ignored_refs=('value',)`. Closing this one.

panel/pane/base.py
```diff
@@ -545,7 +545,7 @@ class ReplacementPane(PaneBase):
 
     _pane = param.ClassSelector(class_=Viewable)
 
-    _ignored_refs: ClassVar[Tuple[str]] = ['object']
+    _ignored_refs: ClassVar[Tuple[str,...]] = ('object',)
```
MarcSkovMadsen, line 548: Please note the types and values of `_ignore_refs` were incorrect. I believe I've fixed them correctly in the places where needed @philippjfr

MarcSkovMadsen:
I've fixed the issue with the `Button`s being turned to into `str`s.

panel/widgets/chat.py
```diff
@@ -203,9 +203,10 @@ class ChatEntry(CompositeWidget):
     show_timestamp : bool
         Whether to display the timestamp of the message.
     """
+    _ignored_refs: ClassVar[Tuple[str,...]] = ('value',)
```
MarcSkovMadsen, line 206: This solved the issue with `Button`s and other widgets being turned into `str`s @ahuang11 and @philippjfr .

I believe its the right solution. But I'm not familiar with this `attribute` and why its there?

panel/pipeline.py
```diff
@@ -165,7 +165,7 @@ class Pipeline(Viewer):
 
     previous = param.Action(default=lambda x: x.param.trigger('previous'))
 
-    _ignored_refs: ClassVar[Tuple[str]] = ('next_parameter', 'ready_parameter')
+    _ignored_refs: ClassVar[Tuple[str,...]] = ('next_parameter', 'ready_parameter')
```
ahuang11, line 168:
```suggestion
    _ignored_refs: ClassVar[Tuple[str, ...]] = ('next_parameter', 'ready_parameter')
```

panel/pane/holoviews.py
```diff
@@ -678,7 +678,7 @@ class Interactive(PaneBase):
 
     priority: ClassVar[float | bool | None] = None
 
-    _ignored_refs: ClassVar[Tuple[str]] = ['object']
+    _ignored_refs: ClassVar[Tuple[str,...]] = ('object',)
```
ahuang11, line 681:
```suggestion
    _ignored_refs: ClassVar[Tuple[str, ...]] = ('object',)
```

examples/reference/widgets/ChatEntry.ipynb
```diff
@@ -75,8 +77,76 @@
...
+    "import pandas as pd\n",
+    "import numpy as np\n",
+    "\n",
+    "data = np.random.rand(5, 2)\n",
+    "dataframe = pd.DataFrame(data, columns=list(\"AB\"))\n",
+    "\n",
+    "fig = plt.figure(figsize=(5,2.5))\n",
+    "ax = fig.add_subplot(111)\n",
+    "ax.plot(data)\n",
+    "plt.close(fig)\n",
```
ahuang11, line 90: I prefer less unrelated code for reference gallery; perhaps something like (haven't tested):
```python

figs, axes = plt.subplots(1, 1)
axes[0].plot([1, 2, 3])
```
would suffice to demo support for Matplotlib?

ahuang11 (changes requested):
Thanks for expanding it; some thoughts.

ahuang11 (approved):
Going to merge this now and fix the pn.bind in the main branch!

PR #5333, opened 2023-07-27
Implement Chat* composable components

ahuang11:
https://github.com/ahuang11/panel-chat-examples

Diagram:
image

Overview:
- `ChatMessage` is essentially a dataclass, and holds the content of the users data and metadata
- `ChatEntry` is the rendering of the `ChatMessage` (pane?)
- `ChatFeed` is the container to hold all the `ChatEntry`(s) (layout?) with the ability to attach a callback (e.g. AI response)
- `ChatInterface` is the highest level interface, composing the `ChatFeed` and `Tabs` of `Widget`(s)

Todo:
- [x] [callback](https://github.com/holoviz/panel/pull/5317)
- [x] [auto scroll + scroll button](https://github.com/holoviz/panel/pull/5310)
- [x] [practical and friendly ux](https://github.com/holoviz/panel/pull/5279)
- [x] langchain agents + streaming
- [x] reference gallery + examples
- [x] fix binding on reactions
- [x] fix layout issues
- [x] tests

PR #5393, opened 2023-08-13
Improve ChatInterface notebook

MarcSkovMadsen:
Builds on top of #5333. So review that one first.

FYI. @ahuang11 . Started reviewing the `ChatInterface` via this PR.


panel/widgets/chat.py
```diff
@@ -30,6 +31,14 @@
 from .input import FileInput, TextInput
 
 
+def dataframe_renderer(value):
+    layout = Column(f"{len(value)} rows was uploaded")
+    if len(value)>1000:
+        value = value.sample(1000)
+        layout.insert(-1, "The data was downsampled to 1000 rows")
+    # Todo: We need to figure out how to tell users to add "perspective" to pn.extension
+    return layout.insert(-1, Perspective(object=value, sizing_mode="stretch_width", max_height=500))
```
ahuang11, line 40: Can you explain why Perspective over pn.pane.DataFrame?

MarcSkovMadsen: Pn.pane.DataFrame is so difficult when you and the user don't know the data uploaded.

It cannot really handle a lot of data nicely. Adding scroll bars and a max height would help.

The user cannot really explore the data. Filter, sort, pivot and Chart.

ahuang11: How big is the `perspective` extension?

ahuang11:
Working on this PR.

ahuang11:
Will merge now; feel free to review the main branch after

PR #5395, opened 2023-08-14
Fixes f-string in how-to guide for generators

thomasjpfan:
This PR adds a f-string, such that the counter appears correctly in the example.


Hoxbro:
Thank you for finding and fixing it :+1:

PR #5396, opened 2023-08-14
Correct return types from threading.Thread to panel.io.server.StoppableThread

dogbunny:
Related to https://github.com/holoviz/hvplot/pull/1121


philippjfr (approved):
Thank you!

PR #5401, opened 2023-08-16
update ChatInterface notebook

MarcSkovMadsen:
FYI. @ahuang11

### DIFF
--- a/examples/reference/widgets/ChatInterface.ipynb
+++ b/examples/reference/widgets/ChatInterface.ipynb
@@ -162,7 +162,7 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "Widgets other than `TextInput` and `TextAreaInput` will require the user to manually click the `Send` button, unless the type is specified in `auto_send_types`."
+    "Widgets other than `TextInput` will require the user to manually click the `Send` button, unless the type is specified in `auto_send_types`."
    ]
   },
   {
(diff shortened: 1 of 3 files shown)


examples/reference/widgets/ChatInterface.ipynb
```diff
@@ -160,7 +162,7 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "Widgets other than `TextInput` will require the user to manually click the `Send` button, unless the type is specified in `auto_send_types`."
+    "Widgets other than `TextInput` and `TextAreaInput` will require the user to manually click the `Send` button, unless the type is specified in `auto_send_types`."
```
MarcSkovMadsen, line 165: I don't know if it works. But I believe it should be easy to replace with the `TextAreaInput`. Maybe also with other text inputs like `TextEditor` and `CodeEditor`.

ahuang11, line 165:
```suggestion
    "Widgets other than `TextInput` will require the user to manually click the `Send` button, unless the type is specified in `auto_send_types`."
```

panel/widgets/chat.py
```diff
@@ -1060,6 +1060,8 @@ class ChatInterface(ChatFeed):
     def __init__(self, **params):
         if params.get("widgets") is None:
             params["widgets"] = [TextInput(placeholder="Send a message")]
+        active_tab = params.pop("active_tab", None)
```
ahuang11, line 1063: If we're doing this, might as well make it a param.

Issue #5394, opened 2023-08-14
Scroll to top of pane


mayonnaisecolouredbenz7:
#### Is your feature request related to a problem? Please describe.
In my app I created some buttons to navigate between tabs
![image](https://github.com/holoviz/panel/assets/38597661/350dba32-ba63-4bff-a6e8-d21223aff6b4)

By changing the index of `tabs.active` on button click.  However. If the current tab has lots of content (the buttons are at the bottom), then when changing the active tab the new tab is out of focus, ie.  below the content of the new tab.

#### Describe the solution you'd like
It would be amazing if there was some scroll to top method on panes.  Could even be used in a _Go To Top_ button for example

#### Describe alternatives you've considered
Since it is effectively a single web page app.  Using HTML such as ` href="#top"` or `href="#"` does not work in this situation.  As technically the web page is up the top always.

#### Additional context
I'm sure this isn't high on the priority list. But i thought I would put it out there and see also if anyone else had come across this as an issue that has been thought about.  Just one of those things that as the app gets bigger and more content is shown.  Would be an amazing thing from a user perspective to have.

ahuang11:
I recently implemented the a scroll to bottom (view_latest), but I think it'd be nice to have a param to control the `scroll_height`, e.g.

`column.scroll_height = 0` would bring it to the top.

MarcSkovMadsen:
Is the specific problem not that the scrollbar is on the Tab component and not on each individual tab page?

mayonnaisecolouredbenz7:
Hi @MarcSkovMadsen , I guess in terms of my situation specifically.  Changing the tabs.active, when you are at the bottom of a tab with content that means you no longer can see the tabs at the top, ends up leaving you in that location on the tab you have changed to. If that makes sense?  

So lets say tab 1 has lots of content, and tab 2 has one line of content.  If I am at the bottom of tab 1 and change active tab.  Then I am way below the content of tab 2.

ahuang11:
Also, I noticed if you set `pn.Tabs(dynamic=True)`, the height is unlinked across tabs.

PR #5403, opened 2023-08-16
Add scroll position to control the location of Column's scroll

ahuang11:
Closes https://github.com/holoviz/panel/issues/5394

"* **``scroll_position``** (int): Current scroll position of the Column. Setting this value will update the scroll position of the Column. Setting to 0 will scroll to the top.\"\"\"\n",


doc/reference/layouts/Column.md
```diff
@@ -0,0 +1,86 @@
+# Column
```
philippjfr, line 1: Remove this file.

PR #5408, opened 2023-08-17
Fix chat issues reported by static analysis

MarcSkovMadsen:
My editor showed me a bunch of (typing) issues. I've solved all but 3.

The remaining ones are

image

I don't plan to fix those.


panel/widgets/button.py
```diff
@@ -243,17 +243,26 @@ def _process_event(self, event: param.parameterized.Event) -> None:
         self.clicks += 1
 
     def on_click(
-        self, callback: Callable[[param.parameterized.Event], None]
+        self, callback: Callable[[param.parameterized.Event], None | Awaitable[None]]
```
MarcSkovMadsen, line 246: This change is important for you to notice @philippjfr . Async functions need special `Awaitable` types.

One day we should run `mypy` on the codebase. Would identify a bunch of potential issues, that if fixed would avoid exceptions or help provide better information/ auto-completion in editors.

panel/widgets/chat.py
```diff
@@ -963,10 +963,11 @@ def stream(
...
         entry = self._build_entry(entry or value, user=user, avatar=avatar)
-        if replace:
-            self._replace_placeholder(entry)
-        else:
-            self._stream(value, entry)
+        if entry:
+            if replace:
+                self._replace_placeholder(entry)
+            else:
+                self._stream(value, entry)
```
ahuang11, line 970: This s a bit confusing to me; it should never be None

MarcSkovMadsen: It can be None due to the line below

image

MarcSkovMadsen: We have seen use cases where the callback should not respond anything. We do this by returning None. For example in the example I had where a user called Philipp was ignored. He would not get responded to by the callback 😄

MarcSkovMadsen: That is maybe not the best example. But  you could imagine setting up the ChatInterface to provide multiple assistants that will respond to each other. And the logic might be easier to implement if you always "respond" to any new message. But depending on the message and user you want to "break out" by returning None.

ahuang11: Thanks for clarifying.

ahuang11 (approved):

MarcSkovMadsen:
If you run ` mypy panel/widgets/chat.py | grep chat.py` it will still identify some potential issues

I don't plan to fix these. But potentially these could cause errors.

panel/widgets/chat.py:17: error: Skipping analyzing "param": module is installed, but missing library stubs or py.typed marker  [import]
panel/widgets/chat.py:17: note: See https://mypy.readthedocs.io/en/stable/running_mypy.html#missing-imports
panel/widgets/chat.py:316: error: Item "None" of "ExitStack | None" has no attribute "enter_context"  [union-attr]
panel/widgets/chat.py:319: error: Incompatible types in assignment (expression has type "type[FileBase]", variable has type "Callable[[Any, KwArg(Any)], Viewable]")  [assignment]
panel/widgets/chat.py:321: error: Item "None" of "ExitStack | None" has no attribute "enter_context"  [union-attr]
panel/widgets/chat.py:327: error: Incompatible types in assignment (expression has type "type[PaneBase]", variable has type "Callable[[Any, KwArg(Any)], Viewable]")  [assignment]
panel/widgets/chat.py:329: error: Item "None" of "ExitStack | None" has no attribute "enter_context"  [union-attr]
panel/widgets/chat.py:332: error: Incompatible types in assignment (expression has type "type[PaneBase]", variable has type "Callable[[Any, KwArg(Any)], Viewable]")  [assignment]
panel/widgets/chat.py:334: error: Item "None" of "ExitStack | None" has no attribute "enter_context"  [union-attr]
panel/widgets/chat.py:337: error: Incompatible types in assignment (expression has type "type[FileBase]", variable has type "Callable[[Any, KwArg(Any)], Viewable]")  [assignment]
panel/widgets/chat.py:339: error: Library stubs not installed for "pandas"  [import]
panel/widgets/chat.py:339: note: Hint: "python3 -m pip install pandas-stubs"
panel/widgets/chat.py:339: note: (or run "mypy --install-types" to install all missing stub packages)
panel/widgets/chat.py:342: error: Incompatible types in assignment (expression has type "type[DataFrame]", variable has type "Callable[[Any, KwArg(Any)], Viewable]")  [assignment]
panel/widgets/chat.py:392: error: Cannot find implementation or library stub for module named "magic"  [import]
panel/widgets/chat.py:741: error: Incompatible types in assignment (expression has type "str | BinaryIO", target has type "str")  [assignment]
panel/widgets/chat.py:779: error: Item "None" of "ChatEntry | None" has no attribute "param"  [union-attr]
panel/widgets/chat.py:1000: error: Unsupported operand type for unary - ("int | None")  [operator]
panel/widgets/chat.py:1001: error: Unsupported operand type for unary - ("int | None")  [operator]
panel/widgets/chat.py:1152: error: Argument 2 to "link" of "Reactive" has incompatible type "**dict[str, str]"; expected "dict[str, str | Callable[..., Any]] | None"  [arg-type]
panel/widgets/chat.py:1152: note: "Dict" is invariant -- see https://mypy.readthedocs.io/en/stable/common_issues.html#variance
panel/widgets/chat.py:1152: note: Consider using "Mapping" instead, which is covariant in the value type
panel/widgets/chat.py:1152: error: Argument 2 to "link" of "Reactive" has incompatible type "**dict[str, str]"; expected "bool"  [arg-type]
panel/widgets/chat.py:1361: error: Item "list[Viewable]" of "Viewable | list[Viewable]" has no attribute

MarcSkovMadsen:
I asked chatGPT to do a review of the code. It liked the code a lot 😄 . It identified a few things I fixed.

panel/widgets/chat.py
```diff
@@ -273,9 +277,9 @@ def __init__(self, **params):
         self.reaction_icons.link(self, value="reactions", bidirectional=True)
         self.param.trigger("reactions")
 
-        render_kwargs = dict(
-            inplace=True, stylesheets=self._stylesheets
-        )
+        render_kwargs = {
+            "inplace": True, "stylesheets": self._stylesheets
```
MarcSkovMadsen, line 281: Using `{...}` is faster than `dict(...)` as you avoid a function call. In practice I don't know how much it matters.

panel/widgets/chat.py
```diff
@@ -1370,8 +1374,7 @@ def active(self) -> int:
         """
         if isinstance(self._input_layout, Tabs):
             return self._input_layout.active
-        else:
-            return -1
+        return -1
```
MarcSkovMadsen, line 1377: This removes one line from the code base which makes it faster to read for users and for python.

MarcSkovMadsen:
There is One test failing. But I've checked and its not introduced by these changes. It is also in the starting branch.

panel/widgets/chat.py
```diff
@@ -720,12 +724,12 @@ def _build_entry(
             value: Union[ChatEntry, dict, Any],
             user: Optional[str] = None,
             avatar: Optional[Union[str, BinaryIO]] = None,
-        ) -> ChatEntry:
+        ) -> Optional[ChatEntry]:
```
Hoxbro, line 727:
```suggestion
        ) -> ChatEntry | None:
```

I would call `Optional` soft deprecated, as this style is preferred.

panel/widgets/chat.py
```diff
@@ -880,14 +884,14 @@ def _stream(self, token: str, entry: ChatEntry) -> ChatEntry:
...
+                value = getattr(value, "objects")[i]
                 i = -1
             elif hasattr(value, "object"):
                 attr = "object"
-                value = value.object
+                value = getattr(value, "object")
             elif hasattr(value, "value"):
                 attr = "value"
-                value = value.value
+                value = getattr(value, "value")
```
Hoxbro, line 894: This makes the code more obscure and has the same effect as the previous code.

panel/widgets/chat.py
```diff
@@ -346,9 +350,9 @@ def _set_default_attrs(self, obj):
         Set the sizing mode and height of the object.
         """
         if hasattr(obj, "objects"):
-            for obj in obj.objects:
-                self._set_default_attrs(obj)
-            return
+            for _obj in obj.objects:
+                self._set_default_attrs(_obj)
```
Hoxbro, line 354:
```suggestion
            for subobj in obj.objects:
                self._set_default_attrs(subobj)
```

This is very nit.

panel/widgets/chat.py
```diff
@@ -211,7 +215,7 @@ class ChatEntry(CompositeWidget):
     show_timestamp : bool
         Whether to display the timestamp of the message.
     """
-    _ignored_refs: ClassVar[Tuple[str,...]] = ('value',)
+    _ignored_refs: ClassVar[List[str]] = ['value']
```
Hoxbro, line 218: I can see the value of having a class variable as a non-mutable object.

panel/widgets/chat.py
MarcSkovMadsen: Agree. It was just the One used already

panel/widgets/chat.py
MarcSkovMadsen: Ah. That is a good suggestion.

panel/widgets/chat.py
MarcSkovMadsen: I totally agree. I previously had a pr to correct this in the entire codebase. But it never came in. Dont know the reason.

panel/widgets/chat.py
MarcSkovMadsen: I sort of agree. The alternative would be to silende static Analysis on that line. I prefer to avoid silencing when possible. But its a Trade off.

panel/widgets/button.py
philippjfr: Thanks, and yes, I do want to add mypy to the automated code checks eventually.

panel/widgets/chat.py
Hoxbro: The functionality code-wise is identical. What is it complaining about?

panel/widgets/chat.py
philippjfr: I think we should indeed prefer `Tuple` everywhere.

panel/widgets/chat.py
philippjfr: We should go through and update this consistently everywhere. Here's all the changes I'd like to make:

- Remove all usage of `Optional`
- Remove all usage of `Union`
- Always use actual types rather than referring to the type as a string

panel/widgets/chat.py
philippjfr: Much prefer silencing here, changing runtime code because of a typing issue is an anti-pattern imo.

philippjfr: Well that's not quite true, changing runtime code because of a typing issue often indicates some actual issue with the code but in this case it's an issue with Python typing being dumb.

panel/widgets/chat.py
philippjfr: Doesn't matter much but I'm happy to change it.

panel/widgets/chat.py
Hoxbro: It should be easy to do with Ruff: `ruff --select=UP --fix --target-version=py38 panel`. I chose all of the settings as this will also convert to f-strings if possible. For a more specific selection, see [here](https://beta.ruff.rs/docs/rules/#pyupgrade-up).

Always having a future import with isort at the start of the file, should also be done.

MarcSkovMadsen: Fixed in the chat.py file. Changing it anywhere else is out of scope for this PR.

panel/widgets/chat.py
MarcSkovMadsen: Yes. But in order to be able to get value from static analysis, you need to remove `False` positives. Otherwise you and the next developer will be investigating the issue over and over again.

And you can argue that disabling static analysis in a line creates a sort of debt because you never know whether its silenced as a "quick fix" by someone or whether the silencing is correct.

That is why I have a tendency to find a way of writing the code such that I don't need to silence the static analysis.

panel/widgets/chat.py
MarcSkovMadsen: I've changed to `Tuple` everywhere and also fixed for `_linkable_properties` properties. But the scope here is `chat.py` so I don't plan to do more in this PR.

MarcSkovMadsen:
Will also merge with `main`. I did a `git merge origin chat_components` and then I got `main` too which seems not to be in `chat_components`!?

MarcSkovMadsen:
Thanks for the comments and helpful suggestions. I've implemented them. I think this one is ready.

panel/widgets/chat.py
```diff
@@ -800,20 +806,20 @@ def _replace_placeholder(self, entry: Optional[ChatEntry] = None) -> None:
...
     def _build_entry(
             self,
-            value: Union[ChatEntry, dict, Any],
-            user: Optional[str] = None,
-            avatar: Optional[Union[str, BinaryIO]] = None,
-        ) -> ChatEntry:
+            value: ChatEntry |  dict | Any,
+            user: str | None = None,
+            avatar: str | BinaryIO | None = None,
+        ) -> ChatEntry | None:
```
ahuang11, line 812: What's the earliest Python version Panel supports? If it's 3.9, I don't think `|` is supported yet:
https://peps.python.org/pep-0604/

panel/widgets/chat.py
```diff
@@ -331,7 +337,7 @@ class ChatEntry(CompositeWidget):
         f"{CDN_DIST}css/chat_entry.css"
     ]
 
-    _exit_stack: ExitStack = None
+    _exit_stack: ExitStack | None = None
```
philippjfr, line 340:
```suggestion
    _exit_stack: ClassVar[ExitStack | None] = None
```

panel/widgets/chat.py
ahuang11: Nevermind, Philipp mentioned from future import annotations should allow this.

PR #5414, opened 2023-08-17
Add cache busting to server CSS resources

philippjfr:
When using server resources the URL for the CSS resources does not change between different versions of Panel. This means that it will always use the same cached resources even when a deployed application updates the Panel version it runs on. To address this we add a query parameter to CSS resources containing the current Panel.js version, which will ensure that on each version upgrade the cache is busted.


MarcSkovMadsen:
Why is this only an issue for css files? What about js, svg etc files?

philippjfr:
Bundled files are mirrored from npm and reflect the full path which generally contains a version number.

PR #5420, opened 2023-08-20
Move user/ default avatars to ChatEntry

MarcSkovMadsen:
I believe it makes sense to move the `user_avatars` to the `ChatEntry` such that if I go low-level and use the `ChatEntry` as well there is still a mechanism for default avatars. I believe its the right place for it to sit.

I've been back and forth but suggest in this PR that

- `user_avatars` are renamed to `default_avatars`.
- We don't also expose `default_avatars` on the `ChatFeed`/ `ChatInterface`. As in `ChatFeed.default_avatars["test"]="X"`. We could. That would avoid an import of `ChatEntry`. But I think there should be "1 way and only 1 way". 
- We don't use the clean up (lower case plus remove space, "-" etc.) functionality originally suggested and implemented. Sorry.
  - I felt it was hard to explain and remember that when configuring the `ChatEntry.default_avatars` you should be using lower cased names without space, "-" etc.
  - Furthermore not using the clean functionality would enable to distinguish between user `AHuang` and `a-huang`.

### DIFF
--- a/panel/tests/widgets/test_chat.py
+++ b/panel/tests/widgets/test_chat.py
@@ -39,7 +39,7 @@ class TestChatReactionIcons:
         assert svg.margin == 0
         svg_text = svg.object
         assert 'alt="favorite"' in svg_text
-        assert 'icon-tabler-heart' in svg_text
+        assert "icon-tabler-heart" in svg_text
 
         assert icons._reactions == ["favorite"]
 
@@ -60,23 +60,24 @@ class TestChatReactionIcons:
 
         svg = icons._svgs[0]
         svg_text = svg.object
-        assert 'icon-tabler-heart-fill' in svg_text
+        assert "icon-tabler-heart-fill" in svg_text
 
     def test_active_icons(self):
         icons = ChatReactionIcons(
             options={"dislike": "thumb-up"},
             active_icons={"dislike": "thumb-down"},
-            value=["dislike"])
+            value=["dislike"],
+        )
         assert icons.options == {"dislike": "thumb-up"}
 
         svg = icons._svgs[0]
         svg_text = svg.object
-        assert 'icon-tabler-thumb-down' in svg_text
+        assert "icon-tabler-thumb-down" in svg_text
 
         icons.value = []
         svg = icons._svgs[0]
         svg_text = svg.object
-        assert 'icon-tabler-thumb-up' in svg_text
+        assert "icon-tabler-thumb-up" in svg_text
 
     def test_width_height(self):
         icons = ChatReactionIcons(width=50, height=50)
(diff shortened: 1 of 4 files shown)


panel/tests/widgets/test_chat.py
```diff
@@ -251,6 +251,23 @@ def callback(reactions):
...
+        assert isinstance(ChatEntry.default_avatars, dict)
+        assert ChatEntry(user="Assistant").avatar==ChatEntry(user="assistant").avatar=="🤖"
+        assert ChatEntry(value="Hello", user="NoDefaultUserAvatar").avatar==ChatEntry.param.avatar.default
+        ChatEntry.default_avatars["test1"]="1"
+        ChatEntry.default_avatars["test2"]="2"
+
+        entry = ChatEntry(value="Hello", user="test1")
+        assert entry.avatar=="1"
+        entry.user="test2"
+        assert entry.avatar=="1"
```
MarcSkovMadsen, line 264: I'm in doubt here if we want to `depends` on `user`?

MarcSkovMadsen:
We could simplify the user interface, code and explanations by removing the `callback_avatar` parameter. That would simplify things. If you think that is a good idea @ahuang11 then let me know and I will remove it.

ahuang11:
> I felt it was hard to explain and remember that when configuring the ChatEntry.default_avatars you should be using lower cased names without space, "-" etc

I believe you misunderstood the purpose of that cleanup code. Users do not have to use lower case names without spaces--in fact they can have ChatGPT, chatgpt, chat-gpt, and chat gpt, all match to the default avatar.

I recommend that the original functionality is added back.

The rest of the PR is fine with me.

ahuang11:
> I felt it was hard to explain and remember that when configuring the ChatEntry.default_avatars you should be using lower cased names without space, "-" etc

I believe you misunderstood the purpose of that cleanup code. Users do not have to use lower case names without spaces--in fact they can have ChatGPT, chatgpt, chat-gpt, and chat gpt, all match to the default avatar.

I recommend that the original functionality is added back.

The rest of the PR is fine with me.

I suppose I misunderstood, but I still prefer this functionality; else there's too many variations one can label GPT and the defaults become useless. If they really need to differentiate they can pass an explicit avatar.

ahuang11:
Yes removing callback avatar with this is fine with me.

MarcSkovMadsen:
> I believe you misunderstood the purpose of that cleanup code. Users do not have to use lower case names without spaces--in fact they can have ChatGPT, chatgpt, chat-gpt, and chat gpt, all match to the default avatar.

I don't think I understood that. But I was in doubt whether `Ahuang` and `a-huang` should be given the same avatar? Maybe the name `default_avatars` is confusing. Because its really a database of user names and avatars that can be maintained.

MarcSkovMadsen:
I've implemented the proposals. Please take a look @ahuang11 . Thanks.

ahuang11:
At the moment, Ahuang and a-huang should be given the same **default** avatar, but users can explicitly specify an avatar to override that if needed. The default being only alpha-numeric sensitive is much more useful, and also cleaner internally, than having the user remember whether the default avatar was keyed as ChatGPT, Chat GPT, Chat-GPT.

Anyhow, I think this PR is now in a good state and I'm going to merge. Besides that, I believe we should start feature freezing that PR unless something is identified as critical because that PR is getting way way way too large.

PR #5423, opened 2023-08-22
Various fixes for rendering of Tabulator and Column

philippjfr:
The rendering of Tabulator children had a few issues related to race conditions of populating the children property, some issues related to serialization of Map/Object types in Bokeh and JS and then some missing guards to ensure things were waiting for everything to be ready.

Additionally the new `Column` implementation @ahuang11 added had a render method that was wrong in a number of ways:

1. It called `super.render()` resulting in two separate render cycles.
2. It called `after_render` on all the children, something the LayoutDOM model explicitly does not do.

I think by simply calling super and then adding the scroll button we are doing things correctly but @ahuang11 please confirm that the scrolling still works with this change.

### DIFF
--- a/panel/models/column.ts
+++ b/panel/models/column.ts
@@ -47,28 +47,14 @@ export class ColumnView extends BkColumnView {
 
   render(): void {
     super.render()
-    this.empty()
-    this._update_stylesheets()
-    this._update_css_classes()
-    this._apply_styles()
-    this._apply_visible()
-
-    this.class_list.add(...this.css_classes())
     this.scroll_down_button_el = DOM.createElement('div', { class: 'scroll-button' });
     this.shadow_el.appendChild(this.scroll_down_button_el);
-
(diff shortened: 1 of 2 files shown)


ahuang11:
I can confirm that all the functionality still works with these changes:
```python
import panel as pn

pn.extension()


def append(event):
    col.append(len(col))


col = pn.Column(
    *list(range(0, 1000)),
    height=500,
    view_latest=True,
    scroll_button_threshold=20,
    auto_scroll_limit=100,
    sizing_mode="stretch_width"
)


button = pn.widgets.Button(name="Append", button_type="primary", on_click=append)
pn.Row(col, button).servable()
```
image


Curious if Card also needs to be refactored too since I took a lot of those lines from that:
https://github.com/holoviz/panel/blob/main/panel/models/card.ts#L73-L78

ahuang11 (approved):

PR #5427, opened 2023-08-24
Fix DeckGL TextLayer issues

### LINKED ISSUE #5346: Serialization Error When Visualizing PyDeck TextLayer
PyDeck 0.8, Panel 1.2

I recently coded up a variation of the PyDeck TextLayer example [from the PyDeck website](https://deckgl.readthedocs.io/en/latest/gallery/text_layer.html) and then attempted to integrate that into a Panel application. Panel is having a serialization issue with the PyDeck String type. 

[This Google Colab notebook](https://colab.research.google.com/drive/13ziK2OrM7n6zmqQDKJQ1_fvrFCCj2oGf?usp=sharing) shows the visualization running without issue in PyDeck and then shows my attempt to render that same visualization via Panel.

MarcSkovMadsen:
Closes #5346

### DIFF
--- a/panel/models/data.ts
+++ b/panel/models/data.ts
@@ -1,10 +1,11 @@
 import {ColumnDataSource} from "@bokehjs/models/sources/column_data_source";
 
+
 export function transform_cds_to_records(cds: ColumnDataSource, addId: boolean = false, start: number = 0): any {
   const data: any = []
   const columns = cds.columns()
   const cdsLength = cds.get_length()
-  if (columns.length === 0||cdsLength === null)
+  if (columns.length === 0 || cdsLength === null)
     return []
 
   for (let i = start; i < cdsLength; i++) {
@@ -14,13 +15,17 @@ export function transform_cds_to_records(cds: ColumnDataSource, addId: boolean =
       const shape = (array[0] == null || array[0].shape == null) ? null : array[0].shape;
       if ((shape != null) && (shape.length > 1) && (typeof shape[0] == "number"))
         item[column] = array.slice(i*shape[1], i*shape[1]+shape[1])
-      else
+      else if (array.length != cdsLength && (array.length % cdsLength === 0)) {
+        const offset = array.length / cdsLength
+	item[column] = array.slice(i*offset, i*offset+offset)
+      } else
         item[column] = array[i]
     }
     if (addId)
       item['_index'] = i
     data.push(item)
   }
+  console.log(data)
   return data
 }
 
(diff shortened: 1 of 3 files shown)


MarcSkovMadsen:
The last issue I need to fix is how to handle the fact that DeckGL supports data rows containing lists like the `coordinates` value below

The DeckGL pane or `ColumnDataSource` used to transfer the data does not support the `coordinates` list. Only `-122.418466` ends up in the DeckGL `data` on the js side. `37.752254` is dropped along the way.

```python
data = {
        "layers": [
            {
                "@@type": "TextLayer",
                "data": [
                            {
                                "name": "24th St. Mission (24TH)",
                                "code": "24",
                                "address": "2800 Mission Street, San Francisco CA 94110",
                                "entries": 12817,
                                "exits": 12529,
                                "coordinates": [-122.418466, 37.752254]
                            }
                ],
            }
        ],
    }
    pane = DeckGL(data)
```

This can be handled in many ways

- 1: Enable using `list`/ `array` values.
  - That seems difficult as we assume data is columnar and we want to transfer it efficiently using ColumnDataSource.
- 2: Add validation to the `DeckGL.object`.
  - This could be done. But it could slow things down because there could be lots of rows to test if you want to test all the `.data` types supported. One place it could be done is when the `ColumnDataSource` is created. If we try to convert to a Pandas dataframe it raises an exception

```python
cds.to_df()
*** ValueError: Per-column arrays must each be 1-dimensional
```

But again this could be a costly operation.
- 3: We could document this issue in the DeckGL reference notebook and provide an example where we on the python side convert the `coordinates` to two `lat` and `long` coordinates.

Please advice @philippjfr. Thanks.

philippjfr:
This example does not yet work for me with this PR: https://deckgl.readthedocs.io/en/latest/gallery/text_layer.html

philippjfr:
Okay, fixed, i.e. implemented your suggestion 1.



This isn't actually difficult because the data is already serialized and sent correctly it just had to be correctly unpacked in JS.

MarcSkovMadsen:
Thanks so much for finalizing this.

panel/models/data.ts
```diff
@@ -14,13 +15,17 @@ export function transform_cds_to_records(cds: ColumnDataSource, addId: boolean =
...
+        const offset = array.length / cdsLength
+	item[column] = array.slice(i*offset, i*offset+offset)
+      } else
         item[column] = array[i]
     }
     if (addId)
       item['_index'] = i
     data.push(item)
   }
+  console.log(data)
```
MarcSkovMadsen: Believe this one should be removed @philippjfr 😸

philippjfr: It was.

PR #5416, opened 2023-08-18
feat: perspective click events

MarcSkovMadsen:
I'm not the author. Just created this to make it easier to see what the suggestion is. See #5074

PR #5430, opened 2023-08-24
Add Perspective click events

philippjfr:
Supersedes https://github.com/holoviz/panel/pull/5416

### DIFF
--- a/examples/reference/panes/Perspective.ipynb
+++ b/examples/reference/panes/Perspective.ipynb
@@ -44,6 +44,10 @@
     "* **``theme``** (str): The theme of the viewer, available options include `'material'`, `'material-dark'`, `'monokai'`, `'solarized'`, `'solarized-dark'` and `'vaporwave'`\n",
     "* **``toggle_config``** (bool): Whether to show the config menu. Default is True.\n",
     "\n",
+    "##### Callbacks\n",
+    "\n",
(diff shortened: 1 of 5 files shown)


examples/reference/panes/Perspective.ipynb
```diff
@@ -44,6 +44,10 @@
     "* **``theme``** (str): The theme of the viewer, available options include `'material'`, `'material-dark'`, `'monokai'`, `'solarized'`, `'solarized-dark'` and `'vaporwave'`\n",
     "* **``toggle_config``** (bool): Whether to show the config menu. Default is True.\n",
     "\n",
+    "##### Callbacks\n",
```
MarcSkovMadsen, line 47: What is really the strategy for events?

- Should we always implement it in a way such that we can `pn.bind` to events?
- Do we at all want to expose call back method?

Are there other events we could/ should add. I believe the Perspective Viewer can be editable. Are there any events signaling a cell update? What I've been able to find is

image

image

philippjfr: >Should we always implement it in a way such that we can pn.bind to events?

Yes, I've got a half-finished implementation of an EventBus object which can be added to an object, which exposes events as parameters. I'll target this for the 1.3 release.

>Are there other events we could/ should add. I believe the Perspective Viewer can be editable. Are there any events signaling a cell update? What I've been able to find is

I also looked at this but haven't really worked out which event would signal an edit.

philippjfr:
Will merge this as is and then improve event handling across the board in 1.3.

Issue #5425, opened 2023-08-23
FileDownload widget is only clickable within Text/label


xnsde:
#### ALL software version info
Panel 1.2.1
Bokeh 3.1.1

#### Description of expected behavior and the observed behavior
When having a wide FileDownload button, clicking outside of the Label causes it to not trigger the download. I looked at the WS socket and no message is sent to the backend, so it's presumably in the javascript code.
Nothing in logs on either python or javascript.

I would expect being able to click anywhere on the download button to download the file. (It's at least very confusing to our users)

#### Complete, minimal, self-contained example code that reproduces the issue

```
from io import BytesIO

import panel as pn


def callback():
    print("click")
    f = BytesIO()
    f.write(b"Hello!")
    return f


if __name__.startswith("bokeh_app"):
    pn.Row(pn.widgets.FileDownload(filename="moo.csv", callback=callback, width=500)).servable()
# Same behaviour whether it's explicit width or sizing_mode="stretch_width"

```

MarcSkovMadsen:
I can reproduce. I only download when clicking the text

![image](https://github.com/holoviz/panel/assets/42288570/30ce74d3-d3d5-498f-80b8-514085acf82a)

MarcSkovMadsen:
The problem seems to be that the `click` event is attached to an anchor `<a>` inside the button `<button>`. And the anchor does not fill the button.

## Workaround

The below works

```python
from io import BytesIO

import panel as pn

STYLE_SHEET = """
button.bk-btn {
    padding: 0px;
}
a {
    display: inline-block;
    height: 100%;
    width: 100%;
    padding: 6px;
}
"""

def callback():
    print("click")
    f = BytesIO()
    f.write(b"Hello!")
    return f


if __name__.startswith("bokeh_app"):
    download=pn.widgets.FileDownload(filename="moo.csv", callback=callback, width=500, stylesheets=[STYLE_SHEET])
    pn.Row(download).servable()
```

## Fix

I don't know if this should be fixed in the file_download.ts file or via some css file?

MarcSkovMadsen:
Try

```python
pn.widgets.FileDownload.stylesheets.append(STYLE_SHEET)
# or if it fails
pn.widgets.FileDownload.stylesheets=[STYLE_SHEET]
```

before you instantiate a `FileDownload`.

PR #5431, opened 2023-08-24
Ensure FileDownload button can be clicked anywhere

philippjfr:
Fixes https://github.com/holoviz/panel/issues/5425

### DIFF
--- a/panel/dist/css/button.css
+++ b/panel/dist/css/button.css
@@ -136,3 +136,10 @@
   background-color: var(--light-bg-color);
   color: black;
 }
+
+.bk-btn a {
+  display: inline-block;
+  height: 100%;
+  width: 100%;
+  padding: 6px;
+}


panel/dist/css/button.css
```diff
@@ -136,3 +136,10 @@
   background-color: var(--light-bg-color);
   color: black;
 }
+
+.bk-btn a {
```
MarcSkovMadsen, line 140: You also need to remove the padding from the button. I.e. set it to 0px.

I think you need to make this more specific targetting the specifically FileDownload?

MarcSkovMadsen: @philippjfr

philippjfr: Wouldn't want to delete the padding from the general Button that is true, the `.bk-btn a` rule is fine because a regular Button doesn't support links inside it.

philippjfr: Is the padding on `a` needed at all?

MarcSkovMadsen: It was for me. It aligns the text center vertically. That is also why it was on the button in the first place.

PR #5435, opened 2023-08-24
Fix #5424 : update echarts when removing series

### LINKED ISSUE #5424: Echarts not updated when removing a serie
#### Software version info
- python 3.10.12
- panel 1.2.1
- bokeh 3.2.1
- MacOS Ventura 13.5
- Chrome Version 115.0.5790.170 arm64

#### Description 
I have a param `ListSelector`, used to build a panel `CheckBoxGroup`, and an `echart`
The number of series of the echart varies according to the number of values selected in the `CheckBoxGroup`.

When I click new values, a new serie is added, and the plot is properly updated.
But when I unclick an existing value, the plot isn't updated.

More details : 
- in both cases, the echart series are updated. see the reproduction code below.
- having a look at the messages passing through the Websocket, I know that "ModelChanged" messages are received by the browser, *with updated series*. Check screenshots below.

- It's really only *removing* a serie that is failing. Adding series always work. This is illustrated in the video below. For instance, running my reproduction code, run the following sequence : 
  - check 2 (it's updated)
  - check 4 (it's updated)
  - uncheck 4 (it's not updated)
  - check8 (it's updated as expected)

```python 
import panel as pn
import param

class DemoEchartsBug(pn.viewable.Viewer):

    numbers = param.ListSelector(default=[],
                                objects=[2, 4, 8])

    def __init__(self, **params):
        super().__init__(**params)

        self.cb_group = pn.widgets.CheckBoxGroup.from_param(self.param.numbers)
    
    @param.depends("numbers")
    def plot(self):

        echart_index = list(range(10))
        
        echart_series = [
            {
                "type": "bar",
                "name": str(num),
                "data": [i * num for i in range(10)],
            }
            for num in self.numbers
        ]

        echart_config = {
            "xAxis": [{"type": "category",  "data": echart_index}],
            "yAxis": [{"type": "value"}],
            "series": echart_series,
        }

        echart_pane = pn.pane.ECharts(echart_config,
                                      width=800,
                                      height=400)

        return pn.Column(
                pn.pane.Markdown(f"# numbers : {self.numbers}"),
                echart_pane
        )

    def __panel__(self):

        return pn.Column(
            pn.pane.Markdown(" # Bug with ECharts"),
            self.cb_group,
            pn.layout.Divider(),
            self.plot
        )
```

No console output

#### Screen recording 

https://github.com/holoviz/panel/assets/756464/9b4cfdf5-1414-4b1f-ad7f-c9f88981e0bd

#### Screenshots of the messages in the websocket

- Step 1 : Before removing a serie. There are two series in the echart config
step 1 - before removing a serie

- Step 2 : After removing a serie. There is one serie in the echart config, as it should be. 
step 2 - after removing a serie

pierrotsmnrd:
This PR fixes issue #5424.

The fix is very simple and is based on a workaround described in the [original issue on Apache Echarts ](https://github.com/apache/echarts/issues/6202) (thanks @MarcSkovMadsen for pointing it out)

### DIFF
--- a/panel/models/echarts.py
+++ b/panel/models/echarts.py
@@ -54,7 +54,7 @@ class ECharts(LayoutDOM):
 
     data = Nullable(Dict(String, Any))
 
-    options_kwargs = Nullable(Dict(String, Any))
+    options = Nullable(Dict(String, Any))
 
     event_config = Dict(String, Any)
 
--- a/panel/models/echarts.py
+++ b/panel/models/echarts.py
@@ -54,6 +54,8 @@ class ECharts(LayoutDOM):
 
     data = Nullable(Dict(String, Any))
 
+    options_kwargs = Nullable(Dict(String, Any))
+
     event_config = Dict(String, Any)
 
     js_events = Dict(String, List(Any))
--- a/panel/models/echarts.ts
+++ b/panel/models/echarts.ts
@@ -80,7 +80,7 @@ export class EChartsView extends HTMLBoxView {
   _plot(): void {
     if ((window as any).echarts == null)
       return
-    this._chart.setOption(this.model.data);
+    this._chart.setOption(this.model.data, true);
   }
 
   _resize(): void {
--- a/panel/models/echarts.ts
+++ b/panel/models/echarts.ts
@@ -161,7 +161,7 @@ export class ECharts extends HTMLBox {
 
     this.define<ECharts.Props>(({ Any, String }) => ({
       data:          [ Any,           {} ],
-      options:[ Any,           {} ],
+      options:       [ Any,           {} ],
       event_config:  [ Any,           {} ],
       js_events:     [ Any,           {} ],
       theme:         [ String,  "default"],
--- a/panel/models/echarts.ts
+++ b/panel/models/echarts.ts
@@ -80,7 +80,7 @@ export class EChartsView extends HTMLBoxView {
   _plot(): void {
     if ((window as any).echarts == null)
       return
-    this._chart.setOption(this.model.data, true);
+    this._chart.setOption(this.model.data, this.model.options_kwargs );
   }
 
   _resize(): void {
@@ -137,6 +137,7 @@ export namespace ECharts {
   export type Attrs = p.AttrsOf<Props>
   export type Props = HTMLBox.Props & {
     data: p.Property<any>
+    options_kwargs: p.Property<any>
     event_config: p.Property<any>
     js_events: p.Property<any>
     renderer: p.Property<string>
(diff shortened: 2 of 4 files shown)


panel/models/echarts.ts
```diff
@@ -80,7 +80,7 @@ export class EChartsView extends HTMLBoxView {
   _plot(): void {
     if ((window as any).echarts == null)
       return
-    this._chart.setOption(this.model.data);
+    this._chart.setOption(this.model.data, true);
```
MarcSkovMadsen, line 83: How do we know

- That it will work across the use cases we need to support?
  - I believe we loose the nice transitions that Echarts provides today. I would recommend having a parameter holding the `true`/ `false` value. `false` is default and provides nice transitions. `true` if you need the workaround for removing series?
- That it  does not have performance implications to rerender from scratch every time?

MarcSkovMadsen: In my experiments Echarts is on top when streaming data compared to bokeh, holoviews, plotly and altair. That we might risk destroying?

pierrotsmnrd: Thanks for the feedback Marc :) 
I have pushed a new commit.  Now, instead of passing true by default, a dict is passed. It's named `options_kwargs` and is added in the TS model as well as the python model and echarts pane.

So new, we can either decide to leave the behavior as is, or we can do : 
```py 
pn.pane.ECharts(my_data,
        options_kwargs={ "replaceMerge": ['series'] }) # similar to true
```

Solution based on [this comment](https://github.com/apache/echarts/issues/6202#issuecomment-974761211) by a core contributor to Echarts.

panel/models/echarts.py
```diff
@@ -54,6 +54,8 @@ def __js_skip__(cls):
 
     data = Nullable(Dict(String, Any))
 
+    options_kwargs = Nullable(Dict(String, Any))
```
MarcSkovMadsen, line 57: Personally I would call it `options` instead of `options_kwargs`. Everyone knows what `options` are. But `kwargs` is a *scary* word to newbies.

panel/models/echarts.py
```diff
@@ -54,6 +54,8 @@ def __js_skip__(cls):
```
MarcSkovMadsen, line 54: The new parameter should also be added to the Reference Notebook documentation. We need the parameter described in the parameter section. I think it could be helpful if the parameter description contained "... for example `options_kwargs={ "replaceMerge": ['series'] })`".

Maybe a link to the apache echart `setOption` options would be helpful?

panel/pane/echarts.py
```diff
@@ -38,6 +38,10 @@ class ECharts(ModelPane):
     object = param.Parameter(default=None, doc="""
         The Echarts object being wrapped. Can be an Echarts dictionary or a pyecharts chart""")
 
+    options_kwargs = param.Parameter(default=None, doc="""
+        An optional dict of options passed to Echarts.setOption. Allows to fine-tune the rendering behavior.
```
MarcSkovMadsen, line 42: It might be useful to add "For example if you want to update your chart to a chart containing a smaller number of series setting `options_kwargs={ "replaceMerge": ['series'] })` is needed".

panel/models/echarts.py
pierrotsmnrd: Done in my latest commit, with a link to the documentation of setOption :)

MarcSkovMadsen:
Thanks for the contribution @pierrotsmnrd .I don't have more to add.

(I've been reading the PR. Not testing it out).

philippjfr (approved):

philippjfr:
Great work and congrats on your first PR @pierrotsmnrd!

philippjfr (approved):

PR #5439, opened 2023-08-25
Add more avatars

MarcSkovMadsen:
Inspired by a recent release of an alternative framework.


panel/widgets/chat.py
```diff
@@ -35,18 +35,42 @@
 from .indicators import LoadingSpinner
 from .input import FileInput, TextInput
 
+ASSISTANT_LOGO = "🤖"
 GPT_3_LOGO = "https://upload.wikimedia.org/wikipedia/commons/thumb/0/04/ChatGPT_logo.svg/1024px-ChatGPT_logo.svg.png?20230318122128"
 GPT_4_LOGO = "https://upload.wikimedia.org/wikipedia/commons/a/a4/GPT-4.png"
+SYSTEM_LOGO = "⚙️"
+USER_LOGO = "🧑"
```
MarcSkovMadsen, line 42: We can also leave it with the smiling emoji.

This emoji of a genderless person also smiles. But then it signals a human as well. I think that is a positive. So at least I would like us to consider if its a beter emoji.

Let me know. Thanks.

panel/widgets/chat.py
```diff
@@ -35,18 +35,42 @@
...
 WOLFRAM_LOGO = "https://upload.wikimedia.org/wikipedia/commons/thumb/e/eb/WolframCorporateLogo.svg/1920px-WolframCorporateLogo.svg.png"
 
 DEFAULT_USER_AVATARS = {
-    "user": "😊",
-    "assistant": "🤖",
-    "system": "⚙️",
+    # User
```
MarcSkovMadsen, line 46: I primarily added synonyms for user and assistant.

ahuang11 (approved):
Looks good to me

ahuang11:
Another idea is to directly use emoji names.

panel/widgets/chat.py
```diff
@@ -44,15 +44,41 @@
...
 GPT_3_LOGO = "https://upload.wikimedia.org/wikipedia/commons/thumb/0/04/ChatGPT_logo.svg/1024px-ChatGPT_logo.svg.png?20230318122128"
 GPT_4_LOGO = "https://upload.wikimedia.org/wikipedia/commons/a/a4/GPT-4.png"
 WOLFRAM_LOGO = "https://upload.wikimedia.org/wikipedia/commons/thumb/e/eb/WolframCorporateLogo.svg/1920px-WolframCorporateLogo.svg.png"
 
-DEFAULT_AVATARS: AvatarDict = {
-    "user": "😊",
-    "bot": "🤖",
-    "assistant": "🤖",
-    "system": "⚙️",
+DEFAULT_USER_AVATARS = {
```
ahuang11, line 54:
```suggestion
DEFAULT_AVATARS = {
```

Issue #5411, opened 2023-08-17
Logout Button


simonaay:
I'm frustrated on how to develop a logout button. Basically, I tried to force an href "logout" to the pathname so I can sign out the user from the authenticated generic provider. Regardless of the provider's type, I can't seem to be able to logout even after learning  the LogoutHandler class found under auth.py file. 

What I'd like to have a button widget to be created and use its callback function to have the user click on it when they need to logout.

I found 2 alternatives where both of which are not helping me or are rather misused.:
- first: https://discourse.holoviz.org/t/forcing-oauth2-logout/4936
- second: https://discourse.holoviz.org/t/tips-guide-on-creating-a-simple-username-password-auth-for-panel-app/2582/3

 Would appreciate the help on this. 

As additional context, I have a dashboard to display using MaterialTemplate and I managed to configure its authentication and authorization steps but I am left with no logout button to develop. All suggestions did not work with me. It would be great if we can simplify the process of logging out with a button widget found either in header or sidebar. Thanks.

philippjfr:
Can you provide some more detail, e.g. which auth providers you tried and the exact details on how you implemented the logout button? I just tried it with something like this and that seemed to work fine:

```
import panel as pn

b = pn.widgets.Button(name='Logout')

b.js_on_click(code="""window.location.href = '/logout'""")

b.servable()
```

MarcSkovMadsen:
Hi @simonaay 

I've updated the `basic` docs in #5441 . Please take a look. Thanks.

simonaay:
> Can you provide some more detail, e.g. which auth providers you tried and the exact details on how you implemented the logout button? I just tried it with something like this and that seemed to work fine:
> 
> ```
> import panel as pn
> 
> b = pn.widgets.Button(name='Logout')
> 
> b.js_on_click(code="""window.location.href = '/logout'""")
> 
> b.servable()
> ```

Hi, yes I happened to have tried this but it's not working. I'll get back to you.

PR #5441, opened 2023-08-25
Enhance Basic Auth

### LINKED ISSUE #5440: Basic Auth does not use Panel favicon or color
I'm on the current Panel `main` branch. I wanted to try out basic auth.

```
panel serve script.py --basic-auth my_password --cookie-secret my_super_safe_cookie_secret
```

But it does not use the Panel favicon or blue color for the Button. This makes it look like a half finished solution.

Personally I don't understand why you want to expose a link to what for the users of your app is complicated documentation of auth. That is just confusing and information overload.

image

MarcSkovMadsen:
Closes #5440

Addresses #5411

You can preview the updated docs and form [here](https://github.com/holoviz/panel/blob/fix/basic-auth/doc/how_to/authentication/basic.md)


panel/auth.py
```diff
@@ -863,7 +863,7 @@ def post(self):
             self.set_current_user(username)
             self.redirect("/")
         else:
-            error_msg = "?error=" + tornado.escape.url_escape("Login incorrect")
+            error_msg = "?error=" + tornado.escape.url_escape("Invalid username or password!")
```
MarcSkovMadsen, line 866: I believe this is the text you normally see.

doc/how_to/authentication/basic.md
```diff
@@ -45,32 +69,4 @@ When serving an application dynamically using `pn.serve` you can also provide a
...
-    <!--Login Button-->
-    <div class="form-group">
-        <button class="form-button" type="submit">Login</button>
-    </div>
-    <div><small>
-    <p><a href="https://panel.holoviz.org/how_to/authentication/index.html">See the documentation</a> for a full discussion.</p>
-    </small></div>
-</form>
-```
+If you want to customize the authentication template you can provide a custom template with the `--basic-login-template` CLI argument. The template needs to submit `username` and `password` to the `/login` endpoint of the Panel server, e.g. the form of the default template looks like this. Check out the default template [here](https://github.com/holoviz/panel/blob/main/panel/_templates/basic_login.html) for inspiration.
```
MarcSkovMadsen, line 72: It helps me more to see the entire template than a part of it. And it also requires work to keep the subset here updated. So replaced it with a link instead.

panel/_templates/basic_login.html
```diff
@@ -2,7 +2,11 @@
 <!DOCTYPE html>
 <html>
 <head>
-    <title>Panel App with BasicAuth</title>
+    <meta charset="utf-8">
+    <meta content="width=device-width, initial-scale=1.0" name="viewport">
+    <meta content="en" name="docsearch:language">
+    <title>Panel App | Login</title>
+    <link rel="icon" type="image/x-icon" href="https://panel.holoviz.org/_static/favicon.ico">
```
MarcSkovMadsen, line 9: I tried `href="_static/favicon.ico"`, but it did not find the favicon.

panel/_templates/basic_login.html
```diff
@@ -2,7 +2,11 @@
 <!DOCTYPE html>
 <html>
 <head>
```
MarcSkovMadsen, line 4: A use does not care about the type of authentication applied. But its nice to know you are on a login page.

panel/_templates/basic_login.html
MarcSkovMadsen: Can you fix this @philippjfr ?

MarcSkovMadsen:
I will also try to fix issue with `/` vs `./` when Panel is not served at `/`. For example when serving in the terminal of a Jupyter Hub that with `jupyter-server-proxy`.

MarcSkovMadsen:
Solving the Jupyterhub issue is a seperate issue. I've described it in #5446.

philippjfr:
Merging this for now.

philippjfr (approved):
Looks great, thank you!

PR #5445, opened 2023-08-25
Adapt tests to Param 2 changes

maximlt:
Adapting tests to:

- updated validation error messages
- updated signature

`panel_extension._apply_signatures` can't yet be removed as Param only updates the signature of a Parameterized class when its `__init__` signature hasn't changed at all compared to `Parameterized.__init__`, while `_apply_signatures` can handle updating Panel objects (e.g. with `def __init__(self, object, ...)`).

### DIFF
--- a/panel/tests/test_param.py
+++ b/panel/tests/test_param.py
@@ -1563,7 +1563,7 @@ def test_set_widget_autocompleteinput(document, comm):
 
     autocompleteinput = model.children[1]
     assert isinstance(autocompleteinput, BkAutocompleteInput)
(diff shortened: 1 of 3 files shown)


philippjfr:
Tests seem quite unhappy!

philippjfr:
Also can you replace usage of `._param_watchers`?

maximlt:
> Also can you replace usage of `._param_watchers`?

This was already done in https://github.com/holoviz/panel/pull/5350 :)

philippjfr:
> This was already done in #5350 :)

Not quite, `param.config` L388 still has a usage of it.

Issue #5195, opened 2023-07-01
custom ReactiveHTML unable to render and process MouseEvent correctly


derrickmcs:
panel version 1.1.1 - [Cytoscape.js](https://github.com/cytoscape/cytoscape.js/issues?q=milestone%3A3.25.0+is%3Aclosed) ver. 3.25.0 rendered using ReactiveHTML, MacOS12.1, Safari 15.2

When using Panel v0.14.1, the component rendered correctly. Mouse clicks were  functional and on target.
When using Panel v1.1.1, the component can only render on certain layouts and mouse click did not work until manually resize the window. However, when the window scrolled away from original position, the mouse clicks were off target (usually by the same amount of pixels from the original position.

#### Complete, minimal, self-contained example code that reproduces the issue

```
from json.tool import main
import param
import panel as pn
from panel.reactive import ReactiveHTML

layouts = {'cose': {'name': 'cose'},
    'grid': {'name': 'grid'},
    'circle': {'name': 'circle'},
    'concentric': {'name': 'concentric'},
    'breadthfirst': {'name': 'breadthfirst'},
    'random': {'name': 'random'},
    'preset': {'name': 'preset'}
}

class Cytoscape(ReactiveHTML):
    layout_names = param.ObjectSelector(default='cose', doc="Layout Options to choose from.", objects=list(layouts.keys()))
    graph_layouts = param.Dict(default=layouts)
    style = param.List(doc="Use to set the styles of the nodes/edges")
    data = param.List(doc="Use to send node's data/attributes to Cytoscape")

    _template = """
        <div id="cy" style="width: 100%; height: 100%; position: relative; border: 1px solid"></div>
    """
    __javascript__ = ['assets/cytoscape.min.js']

    _scripts = {
        'render': """
            state.cy = cytoscape({
              container: cy,
              layout: {name: 'cose'}, //these layout will fail: grid, circle, concentric
              elements: [{data:{id:'A', label:'A'}},{data:{id:'B', label:'B'}}, {data:{source:'A', target:'B'}}],
              zoom: 1,
              pan: {x: 0, y: 0},
            });
            state.cy.on('click', function (evt) {
                console.log(evt.originalEvent, evt)
            });
        """,
        'remove': """
        delete state.cy
        """,
        'layout_names': """
            const layout = state.cy.layout(data.graph_layouts[data.layout_names])
            layout.run()
        """
    }

    _extension_name = 'cytoscape'

pn.extension('cytoscape', template='fast')

[...]

PR #5448, opened 2023-08-26
Refactor ReactiveHTML docs

MarcSkovMadsen:
The Panel 1.x refactor of our docs left the `ReactiveHTML´ documentation in a confusing state. Its not clear where to start or where to find what.

This PR tries to address that.

You can preview the documentation via https://github.com/holoviz/panel/tree/enhancement/reactive-html-docs/doc/how_to/custom_components/reactive_html/index.md and the navigation tree on the left.

## Related Issues

- #5451
- #5452
- #5476
- #5499
- #5500
- #5541
- #5550
- #5551
- #5586
- #5587

### Docs

- [how-to](https://panel.holoviz.org/how_to/custom_components/custom_reactiveHTML.html)

### Discourse

- [5857: user friendly indicator with the consideration of automatic sizing](https://discourse.holoviz.org/t/implement-a-real-and-user-friendly-indicator-with-the-consideration-of-automaic-sizing-for-all-the-indicators-features/5857)


doc/how_to/custom_components/reactive_html/reactive_html_layout.md
```diff
@@ -0,0 +1,292 @@
...
+```{pyodide}
+import panel as pn
+import param
+
+pn.extension()
+
+class LayoutSingleObject(pn.reactive.ReactiveHTML):
+    object = param.Parameter()
+
+    _ignored_refs = ("object",)
```
MarcSkovMadsen, line 22: If I don't add `object` to `_ignored_refs` its shown as a literal value. Is that a bug?

doc/how_to/custom_components/reactive_html/reactive_html_layout.md
```diff
@@ -0,0 +1,292 @@
...
+rendered as markdown! That is correct.
+
+Before inserting the value of a parameter in the
+`_template`, Panel transforms the value using `pn.panel`. And for a string value `pn.panel` returns
+a `Markdown` pane.
+
+Let's verify this.
+
+```{pyodide}
+print(type(layout.object1), type(layout.object2))
```
MarcSkovMadsen, line 97: Is this a regression/ bug. I've never noticed before that my parameter values a converted. This makes it hard to use the parameters for other use cases like binding.

doc/how_to/custom_components/reactive_html/reactive_html_layout.md
```diff
@@ -0,0 +1,292 @@
...
+    styles={"border": "2px solid lightgray"},
+)
+```
+
+## A layout of literal `str` values
+
+If you want to show the *literal* `str` value of your parameter instead of the `pn.panel` return
+value you can configure that via the `_child_config` attribute.
+
+```{pyodide}
```
MarcSkovMadsen, line 115: I thought it was also possible to insert literal values via `{{ object1 }}`. I've tried different variations of the below code. But they all show me a blank page.

```python
pn.extension()
class LayoutLiteralValues(pn.reactive.ReactiveHTML):
    object1 = param.Parameter()
    object2 = param.Parameter()
    _ignored_refs = ("object1", "object2")
    _child_config = {"object2": "literal"}
    _template = """
    <style>
    .pn-container {height: 100%;width: 100%;}
    </style>
    <div class="pn-container">
        <h1>Object 1</h1>
        <div id="object1">{{object1}}</div>
        <h1>Object 2</h1>
        <div id="object2">${object2}</div>
    </div>
"""
layout = LayoutLiteralValues(
    object1="This is the **value** of `object1`", object2="This is the **value** of `object2`",
    styles={"border": "2px solid lightgray"},
)
layout.servable()
```

image

doc/how_to/custom_components/reactive_html/reactive_html_layout.md
```diff
@@ -0,0 +1,292 @@
...
+pn.extension()
+
+class LayoutMultipleValues(pn.reactive.ReactiveHTML):
+    object1 = param.Parameter()
+    object2 = param.Parameter()
+
+    _ignored_refs = ("object1", "object2")
+
+    _template = """
+    <style>
```
MarcSkovMadsen, line 69: Instead of inlining the CSS I need guidance on the best practice for adding CSS @philippjfr . I can't see how this should be done.

I can of course add CSS via the `stylesheets` parameter. But that will leave a risk that it gets overridden by users if the set the `stylesheets` parameter on construction or after.

doc/how_to/custom_components/reactive_html/reactive_html_layout.md
```diff
@@ -0,0 +1,292 @@
...
+
+pn.extension()
+
+class LayoutSingleObject(pn.reactive.ReactiveHTML):
+    object = param.Parameter()
+
+    _ignored_refs = ("object",)
+
+    _template = """
+    <div style="height:100%;width:100%">
```
MarcSkovMadsen, line 25: Why is it that I always have to add a wrapper div and set the `height` and `width` to 100% @philippjfr . And why do I have to explain users this as well. This seems like friction. Could it be done automatically?

doc/how_to/custom_components/reactive_html/reactive_html_layout.md
```diff
@@ -0,0 +1,292 @@
...
+- We can insert the `key` as a literal value using `{{ key }}`.
+- We must not give the HTML element containing `{{ key }}` an `id`. If we do, an exception will be
+raised.
+
+## A Grid like layout
+
+If you want to create a grid like layout similar to the [`GridSpec`](https://panel.holoviz.org/reference/layouts/GridSpec.html)
+you can ???
+
+Todo: Figure out how to do this. Looking at the `GridSpec` it seems there is no *mixin* class I can use.
```
MarcSkovMadsen, line 292: I've looked at `GridSpect` it does not seem like there is any easy to use *mixin* class to create a grid like layout.

Maybe I should just drop this example?

doc/how_to/custom_components/reactive_html/reactive_html_layout.md
ahuang11: I suspect it's because it's because you're using `param.Parameter`

Have you tried it with your ignored_refs patch in `chat_components`?

MarcSkovMadsen: No. I dont understand What ignored_refs has to do with ReactiveHTML.

I've never seen it mentioned in that context before?

ahuang11: Not sure either; just a random guess since you're using param.Parameter, and the https://github.com/holoviz/panel/pull/5408 seemed to fix rendering.

doc/how_to/custom_components/reactive_html/reactive_html_widgets.md
```diff
@@ -0,0 +1,183 @@
...
+    .style {border: 2px dashed lightgray;border-radius:20px}
+    </style>
+    <select id="select" class="pn-container style" value="${value}">
+      {% for option in options %}
+      <option id="option">{{option}}</option>
+      {% endfor %}
+    </select>
+    """
+
+    _dom_events = {'select': ['change']}
```
MarcSkovMadsen, line 160: I don't really understand the `_dom_events`

1. What does it really do?
  - Without it the `${value}` is not updated. I believe I have created widgets that update value if the are just in the `_template`?
2. What if I had 2 `select` tags in my `_template` but I only wanted to bind to one of them? It seems to me like I lack some control?

ahuang11: I believe select is based off ID, not the widget

MarcSkovMadsen: Thanks. That makes much more sense. Changing it to `select_id` seems to also work.

```python
import panel as pn
import param

pn.extension()

class Select(pn.reactive.ReactiveHTML):

    options = param.List(doc="Options to choose from.")

    value = param.String(doc="Current selected option")

    _template = """
    <select id="select_id" value="${value}" style="width: ${model.width}px">
      {% for option in options %}
      <option id="option">${option}</option>
      {% endfor %}
    </select>
    """

    _dom_events = {'select_id': ['change']}

select = Select(options=['A', 'B', 'C'])
select

pn.Column(select, select.param.value).servable()
```

MarcSkovMadsen: The `Select` example has some issues though which I've reported in #5452

doc/how_to/custom_components/reactive_html/reactive_html_widgets.md
```diff
@@ -0,0 +1,275 @@
...
+          state.ctx.moveTo(state.start.offsetX, state.start.offsetY)""",
+        "draw": """
+          if (state.start == null)
+            return
+          state.ctx.lineTo(event.offsetX, event.offsetY)
+          state.ctx.stroke()""",
+        "end": "delete state.start",
+        "save": """
+          data.value = canvas_el.toDataURL()
+          data.save=false""",
```
MarcSkovMadsen, line 228: Need confirmation from @philippjfr that its OK and good practice to use `param.Event` parameters in the way I do with `save` and `reset`. I.e. be setting `data.save=false` in the script to enable triggering/ clicking it again.

doc/how_to/custom_components/reactive_html/reactive_html_dataframe.md
```diff
@@ -0,0 +1,66 @@
+# DataFrames and ReactiveHTML
+
+In this guide you will learn how to efficiently implement `ReactiveHTML` components with a
+`DataFrame` parameter.
+
+## Creating a Table Pane
```
MarcSkovMadsen, line 6: Would it help users more if the example was based on Tabulator? That would enable users to implement advanced Tabulator features if needed.

philippjfr: I'd say no, I don't think implementing your own Tabulator wrapper will ever be a good route to go.

philippjfr:
Starting to think we should just turn this guide and the streamlit migration guide into tutorials.

doc/how_to/custom_components/reactive_html/reactive_html_dataframe.md
MarcSkovMadsen: It would not be implementing everything. Instead it would be implementing a specific use case that Panels Tabular does not support. Maybe you need some special JavaScript event handling or images in the cells or ....

MarcSkovMadsen: As I see ReactiveHTML, it is both for reusable components and for one offs.

MarcSkovMadsen:
My plan was to create a tutorial highly inspired by AnyWidget then I would not have to invent things and it would be easy to compare to AnyWidget.

doc/how_to/custom_components/reactive_html/reactive_html_panes.md
```diff
@@ -0,0 +1,150 @@
...
+                zoom: data.zoom,
+                pan: data.pan,
+                });
+                state.cy.on('select unselect', function (evt) {
+                  data.selected_nodes = state.cy.elements('node:selected').map((el)=>{return el.id()})
+                  data.selected_edges = state.cy.elements('edge:selected').map((el)=>{return el.id()})
+                });
+                self.style()
+                const mainEle = document.querySelector("body")
+                mainEle.addEventListener("scrollend", (event) => {state.cy.resize().fit()})
```
MarcSkovMadsen, line 117: Don't we lack a life cycle function after `after_layout` @philippjfr ? Why is a hack like listening for `scrollend` or adding a `timeOut` sometimes needed?

MarcSkovMadsen: In https://github.com/holoviz/panel/issues/5195#issuecomment-1693935047 @derrikmcs also notes that `scrollend` event is not supported by all browsers.

doc/explanation/components/reactive_html_components.md
```diff
@@ -1,12 +1,171 @@
...
+The *blueprint* for how your custom component should look and behave. The *_template* is defined using HTML. You can use [Jinja2](https://jinja.palletsprojects.com) syntax as well as *template variables* `${...}` to make the template *dynamic*.
+
+### `_child_config` (Optional[dict])**
+
+Controls how template variables ${...} will be rendered when inserted as children into the HTML
+elements
+
+- `model` (default): Create child and render as Panel component.
+- `literal`: Create child and insert the string value as `innerHTML`
+- `template`: Create child and insert the string value as `innerText`
```
MarcSkovMadsen, line 60: I need help to define these @philippjfr . I've seen different definitions and they are not consistently aligned with what I experience when using them. See for example examples and questions below. I'm so confused.

- Can you confirm the are all rendered on `js` side. Sometimes I'm thinking either `literal` or `template` is rendered on python side using Jinaj2.
- I've tried to look at the .py and .ts code many times. But I cannot find the logic that distinguishes between `literal` and `template`. So I cannot learn from the code. Where is that code?
- What does the names `literal` and `template` refer too? Please give me something so that I can remember the difference.
- Why does the *escaping* not work consistently for `literal` and `template` values? Is it a feature or a bug? It makes it so hard to understand and trust that this is robust. See examples below.
- Why can I set the `v_model` parameter value below to a markdown string that is rendered as a Markdown pane. But if I try setting it to an `svg` string then I get an error because a `String` parameter cannot be assigned a `SVG` pane? Is is a bug or a feature. I'm so confused here.
- Why does `ReactiveHTML` even try to change parameter values to their Panel pane? It makes it so hard to use `DataFrame` etc. parameters.

MarcSkovMadsen: This is the example that confuses me so much

```python
import panel as pn
import param

pn.extension()

class CustomComponent(ReactiveHTML):
    v_model = param.String(default="I'm a **model** value. <em>emphasize</em>")
    v_literal = param.String(default="I'm a **literal** value. <em>emphasize</em>")
    v_template = param.String(default="I'm a **template** value. <em>emphasize</em>")

    _child_config = {
        "v_model": "model",
        "v_literal": "literal",
        "v_template": "template",
    }

    _template = """
<div id="el_model">${v_model}</div>
<div id="el_literal">${v_literal}</div>
<div id="el_template">${v_template}</div>
"""


component = CustomComponent(width=500, height=200).servable()
component.v_model=component.v_model.replace("**", "*")
component.v_literal=component.v_model.replace("**", "*")
component.v_template=component.v_model.replace("**", "*")
print(component.v_model)

svg = """<svg style="stroke: #e62f63;" width="18" height="18" viewBox="0 0 18 18" fill="none" xmlns="http://www.w3.org/2000/svg" slot="collapsed-icon">
<path d="M15.2222 1H2.77778C1.79594 1 1 1.79594 1 2.77778V15.2222C1 16.2041 1.79594 17 2.77778 17H15.2222C16.2041 17 17 16.2041 17 15.2222V2.77778C17 1.79594 16.2041 1 15.2222 1Z" stroke-linecap="round" stroke-linejoin="round"></path>
<path d="M9 5.44446V12.5556" stroke-linecap="round" stroke-linejoin="round"></path>
<path d="M5.44446 9H12.5556" stroke-linecap="round" stroke-linejoin="round"></path></svg>"""
CustomComponent(v_literal=svg, v_template=svg, width=500, height=200).servable()
```

image

doc/explanation/components/reactive_html_components.md
```diff
@@ -1,12 +1,260 @@
...
+#### What can I use Jinja2 templating for?
+
+You can use Jinja2 syntax to layout your template. When using Jinja2 syntax you can refer to parameters using `{{ ... }}` syntax. This will insert your parameter values as a literal string values. In addition you can use the following context variables:
+
+- `param`: The param namespace object allows templating parameter names, labels, docstrings and other attributes.
+- `__doc__`: The class docstring
+
+Here is an illustrative example
+
+```{pyodide}
```
MarcSkovMadsen, line 86: This example is important. I spent hours finding some combination that works.

One problem is that `ReactiveHTML` has issues inserting `{{ ... }}` values that contains tilde

- https://github.com/holoviz/panel/issues/5499
- https://github.com/holoviz/panel/issues/5500

Another is that I could not find a working way to just loop `param`. Instead I had to use `param.params().values()` which is cumbersome.

Another is that I could not find an easy way to get the current value of a parameter in the loop. So I ended up having to do `object.owner[object.name]`.

philippjfr: > One problem is that ReactiveHTML has issues inserting {{ ... }} values that contains tilde

Will look at fixing that.

> Another is that I could not find a working way to just loop param. Instead I had to use param.params().values() which is cumbersome.

`.param` can be looped directly.

> Another is that I could not find an easy way to get the current value of a parameter in the loop. So I ended up having to do object.owner[object.name].

This is a good point, `.param.values()[name]` seems awkward.

doc/how_to/custom_components/reactive_html/reactive_html_layout.md
philippjfr: Can you try using `_stylesheets`? It was designed for other components but **should** work with ReactiveHTML as well.

doc/how_to/custom_components/reactive_html/reactive_html_widgets.md
philippjfr: You technically should have to do that, but I guess since the value isn't synced when it is reset to false it won't trigger the next time. We should fix this and ensure it does reset automatically.

doc/explanation/components/reactive_html_components.md
MarcSkovMadsen: You can loop `.param` directly. It gives you a name of a parameter. But `.param[name]` does not seem to work. Thus you cannot get the actual parameter to work with.

doc/how_to/custom_components/reactive_html/reactive_html_layout.md
MarcSkovMadsen: The problem was the ` in the string. Panel had a bug that Philipp has fixed.

panel/io/resources.py
```diff
@@ -268,7 +272,7 @@ def resolve_stylesheet(cls, stylesheet: str, attribute: str | None = None):
         The stylesheet definition
     """
     stylesheet = str(stylesheet)
-    if not stylesheet.startswith('http') and attribute and (custom_path:= resolve_custom_path(cls, stylesheet)):
+    if not stylesheet.startswith('http') and attribute and _is_file_path(stylesheet) and (custom_path:= resolve_custom_path(cls, stylesheet)):
```
MarcSkovMadsen, line 275: I tried running the below example. But it raised an error because it thought the `_stylesheets` value was a path.

```python
import param


class CounterWidget(ReactiveHTML):
    value = param.Integer(default=0)

    _template = """  
    <button id="button_el" class="styled-button" onclick="${script('click_handler')}"></button>  
    """

    _scripts = {
        "render": "self.value()",
        "click_handler": "data.value+=1",
        "value": "button_el.innerHTML = `count is ${data.value}`"
    }
    
    _stylesheets = ["""\
.styled-button {  
    display: inline-block;
    padding: 10px 20px;
    font-size: 16px;
    font-weight: bold;
    text-align: center;
    text-decoration: none;
    background-color: #4CAF50;
    color: white;
    border: none;
    border-radius: 5px;
    cursor: pointer;
    transition: background-color 0.3s;
}  

.styled-button:hover {
    background-color: #45a049;  
}  
"""]  

CounterWidget().servable()
```

doc/how_to/custom_components/reactive_html/reactive_html_layout.md
MarcSkovMadsen: I will drop this comment. There is enough to review already.

doc/explanation/components/reactive_html_components.md
```diff
@@ -0,0 +1,781 @@
+# Building `ReactiveHTML` Components
```
MarcSkovMadsen, line 1: The section headers are too long. Either they should be shortened or the width of the sidebar should be changed.

image

doc/how_to/custom_components/index.md
```diff
@@ -2,6 +2,10 @@
```
MarcSkovMadsen, line 2: I feel like this "big" topic of custom components and `ReactiveHTML` is very much hidden because its several clicks down

image

Also I believe that it would make navigation easier of the sections and cards where aligned as much as possible between `how-to` and `explanation` pages. That would make navigation easier as you don't have to remember seperately where to find ReactiveHTML components material for how-to guides and for explanation.

image

MarcSkovMadsen: I would recommend restructing the `how-to` navigation by having a `Components` section and moving the following cards there

- construct components
- arrange components
- style components
- build custom components

MarcSkovMadsen:
I've reviewed a version of the docs deployed to dev site and reviewed. Among other things I've fixed some broken navigation.

I've triggered the docs dev build again for a second review.

doc/how_to/custom_components/index.md
```diff
@@ -12,20 +16,109 @@ These How-to pages provide solutions for common tasks related to extending Panel
 How to build custom components that are combinations of existing components.
 :::
 
-:::{grid-item-card} {octicon}`plus-circle;2.5em;sd-mr-1 sd-animate-grow50` Build Components from Scratch
-:link: custom_reactiveHTML
+::::
+
+## `ReactiveHTML` Components
```
MarcSkovMadsen, line 21: This section is currently broken

image

doc/explanation/components/reactive_html_components.md
```diff
@@ -0,0 +1,635 @@
...
+
+- [Shoelace](https://shoelace.style/): A large, mature collection of components that integrate well with the Bootstrap design.
+- [Fast](https://explore.fast.design/): A relatively large and mature collection of components that integrate well with the Fast design.
+- [Material](https://github.com/material-components/material-web): A growing collection of components that integrate well with the Material design.
+
+For more inspiration check out the [awesome-web-components](https://github.com/web-padawan/awesome-web-components#components) list.
+
+### React, Preact and Vue
+
+`ReactiveHTML` can be used with [React](https://react.dev/), [Preact](https://preactjs.com/) and [Vue](https://vuejs.org/).
```
MarcSkovMadsen, line 571: I'm no longer sure `ReactiveHTML` works with React (and MaterialUI). In https://github.com/mui/material-ui/issues/17473 they claim React and Material does not and do not want to support ShadowRoot.

There is a minimal example here that I tried to get working but could not https://discourse.holoviz.org/t/help-with-panel-react-materialui/5204/3?u=marc

MarcSkovMadsen:
Can I do something to move this forward and be easier to review+accept @philippjfr ? I think it will be a big step forward for `ReactiveHTML` and custom components in Panel.

philippjfr:
I think you've really done a great job here. It's really on me to review and see if we can fix some of the issues you raised.

MarcSkovMadsen:
We could postpone the "hard to solve" issues, fix the "easy ones" and release it in a first step. Then take a second step to solve the harder issues and maybe improve `ReactiveHTML` too.

doc/how_to/custom_components/reactive_html/reactive_html_layout.md
```diff
@@ -0,0 +1,272 @@
...
+- wrap the `{% for object in objects %}` loop in an HTML element with an `id`. Here it is wrapped
+with `<div id="container">...</div>`.
+- close all HTML tags! `<hr>` is valid HTML, but not valid with `ReactiveHTML`. You must close it
+as `<hr/>`.
+
+Please note you can optionally
+
+- get the index of the `{% for object in objects %}` loop via `{{ loop.index0 }}`.
+
+## Create a list like layout
```
tomascsantos, line 194: I notice that you're using jinja template, but when I tested this and updated the objects parameter, this actually updated--which is awesome. I guess that's some magic from NamedListLike? Might be worth mentioning this is the case though

tomascsantos (commented):
This PR is a work of art. thanks so much @MarcSkovMadsen for putting it together, I had no idea ReactiveHTML was so POWERFUL!

One last open ended question that I'm wondering about...:
what is the difference between reactiveHTML and templates? In other words, why don't templates inherit from reactive HTML?

MarcSkovMadsen:
> This PR is a work of art. thanks so much @MarcSkovMadsen for putting it together, I had no idea ReactiveHTML was so POWERFUL!
> 
> One last open ended question that I'm wondering about...: what is the difference between reactiveHTML and templates? In other words, why don't templates inherit from reactive HTML?

Hi @tomascsantos 

I think the answer is history.

- Bokeh comes with support for jinja templates. These has been used to build the `FastListTemplate` and similar templates. The jinja templates can be used to layout Bokeh/ Panel components in a HTML template.
- Bokeh comes with *bokeh models* to build advanced Bokeh/ Panel components using typecript and build tools
- Later there was a need for more easily creating custom components with bi-directional communication and javascript without build tools. The `ReactiveHTML` was introduced. In principle `ReactiveHTML` can replace the layout functionality of templates. I don't think it would ever replace the basic templating things like specifying the `favicon` etc. in the header of a template. But if I should start over creating the `FastListTemplate` etc. one day, then I would probably have done that by building up sub components in ReactiveHTML and combining them.
- There is also the future. And `ReactiveHTML` does not support ESM modules which Javascript developing is moving towards. So some day this will be supported in Panel as well. See https://github.com/holoviz/panel/pull/5593.

doc/explanation/components/reactive_html_components.md
```diff
@@ -0,0 +1,635 @@
...
+
+## The Name
+
+`ReactiveHTML` is named for its ability to enable reactive programming in HTML. Unlike the static HTML content that the [`HTML`](../../../examples/reference/panes/HTML.ipynb) pane displays, `ReactiveHTML` components can update their view dynamically in response to changes in parameter values and other events.
+
+We could also have called the `ReactiveHTML` class for example `BaseComponent`, `HTMLComponent`, `SimpleComponent` or `AnyComponent` to give you the right associations.
+
+It's worth noting that the name `ReactiveHTML` is not related to the JavaScript framework [React](https://react.dev/), although you can still use React with `ReactiveHTML` components.
```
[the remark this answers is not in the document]
MarcSkovMadsen: You can use `React`. No problem. But I was not able to really get the MaterialUI react components working.

doc/explanation/components/reactive_html_components.md
```diff
@@ -0,0 +1,635 @@
...
+
+The required attribute is:
+
+- `_template` (str): This defines the blueprint for how your custom component should look and behave. It is defined using HTML and can use Javascript *template variables* `${...}` as well as Python [Jinja2](https://jinja.palletsprojects.com) syntax to make the template *dynamic*.
+
+You can also declare the following optional attributes:
+
+- `_child_config` (dict): This is a mapping that controls how children are rendered.
+- `_ignored_refs` (tuple[str]): This is tuple of parameter names. Use this to render Panel components as Panel components and not their value or object.
+- `_dom_events` (dict): This is a mapping of node IDs to DOM events to add event listeners to.
```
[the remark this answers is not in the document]
MarcSkovMadsen: Yes. Which section are you referring to?

doc/explanation/components/reactive_html_components.md
```diff
@@ -0,0 +1,635 @@
...
+When creating a `ReactiveHTML` component, there are several class attributes that you can declare to customize its behavior.
+
+The required attribute is:
+
+- `_template` (str): This defines the blueprint for how your custom component should look and behave. It is defined using HTML and can use Javascript *template variables* `${...}` as well as Python [Jinja2](https://jinja.palletsprojects.com) syntax to make the template *dynamic*.
+
+You can also declare the following optional attributes:
+
+- `_child_config` (dict): This is a mapping that controls how children are rendered.
+- `_ignored_refs` (tuple[str]): This is tuple of parameter names. Use this to render Panel components as Panel components and not their value or object.
```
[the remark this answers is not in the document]
MarcSkovMadsen: I think its because I don't understand the `_ignored_refs`. I could just see I had to use it to get some examples to work. Could you please elaborate @philippjfr . Thanks.

doc/explanation/components/reactive_html_components.md
```diff
@@ -0,0 +1,635 @@
...
+
+The `_scripts` attribute is used to define two JavaScript callbacks: `increment` and `count`. The `increment` callback increases the count value by 1, while the `count` callback resets the `count` value to 0 if the `count` value reaches 3 or higher. Each callback can access the `data` object, which holds the current values of synced parameters (in this case, the `count` parameter).
+
+When the *Increment* button is clicked, the corresponding JavaScript callback is invoked, updating the count value.
+
+When the `count` value is changed, the corresponding `count` JavaScript callback is invoked, optionally resetting the `count` value to zero.
+
+By combining Python callbacks and JavaScript callbacks, you can create dynamic and interactive components that respond to user interactions.
+
+### Parameter callbacks
```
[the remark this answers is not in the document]
MarcSkovMadsen: Its a subsection of the section above. The intention is to systematically go through the different kind of scripts. And there is some overlap with the above.

Any ideas to how I can make this clearer or better?

doc/how_to/custom_components/reactive_html/reactive_html_layout.md
MarcSkovMadsen: I've tried to mention this.

doc/explanation/components/reactive_html_components.md
```diff
@@ -0,0 +1,635 @@
...
+The required attribute is:
+
+- `_template` (str): This defines the blueprint for how your custom component should look and behave. It is defined using HTML and can use Javascript *template variables* `${...}` as well as Python [Jinja2](https://jinja.palletsprojects.com) syntax to make the template *dynamic*.
+
+You can also declare the following optional attributes:
+
+- `_child_config` (dict): This is a mapping that controls how children are rendered.
+- `_ignored_refs` (tuple[str]): This is tuple of parameter names. Use this to render Panel components as Panel components and not their value or object.
+- `_dom_events` (dict): This is a mapping of node IDs to DOM events to add event listeners to.
+- `_extension_name` (str): This is the name used to import external CSS and JS dependencies via `pn.extension(_extension_name)`.
```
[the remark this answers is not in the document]
MarcSkovMadsen: included

doc/how_to/custom_components/index.md
philippjfr: Sounds reasonable to me.

doc/explanation/components/reactive_html_components.md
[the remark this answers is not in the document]
tomascsantos: I recommend linking all the sections... I just reviewed this reading through github markdown, not sure if when pushed to the website this would add links automatically

doc/explanation/components/reactive_html_components.md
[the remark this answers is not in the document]
tomascsantos: You can take this with a grain of salt but,

If you feel you need this section, I might change the example code so that there is no overlap, and then include a callback to the previous example:

> If the key in the `_scripts` dictionary matches one of the parameters declared on the class the callback will automatically fire whenever the synced parameter value changes. This is how counts changes in the example above. Here's another example let's say we have a class...

doc/how_to/custom_components/reactive_html/reactive_html_widgets.md
philippjfr: Fixed in https://github.com/holoviz/panel/pull/6247

doc/how_to/custom_components/reactive_html/reactive_html_panes.md
philippjfr: There sadly is no event after `after_layout` as that's the last step in the Bokeh rendering pipeline.

philippjfr:
pre-commit.ci autofix

philippjfr:
I will merge as is and if further fixes/cleanup is needed will address that as part of the ESM docs in https://github.com/holoviz/panel/pull/5593

Issue #5459, opened 2023-08-31
Duplication bug while trying to render native ipywidgets


aditya3434:
#### ALL software version info
Panel version 1.2.1
Bokeh version 3.2.2
Jupyter Lab version 3.6.2

#### Description of expected behavior and the observed behavior

Expected behaviour : Widget renders slider value and its doubled value correctly as slider moves.

Actual behaviour : The previous slider value is not deleted and the new values get rendered after it, resulting in duplicate value messages.

Only happens while using native ipywidgets. Works fine with panel components.

#### Complete, minimal, self-contained example code that reproduces the issue

```
import panel as pn
import ipywidgets as ipw
pn.extension()

def f(m=10):
    return pn.Column(ipw.HTML("Value : "+str(m)), ipw.HTML("Double Value : "+str(2*m)))

pn.interact(f)

```

#### Stack traceback and/or browser JavaScript console output
No stack traceback or console output

#### Screenshots or screencasts of the bug in action

![Animation](https://github.com/holoviz/panel/assets/56246494/400b54f4-c2ac-4287-a661-1b0028140088)

PR #5462, opened 2023-08-31
Fix race conditions when initializing and rendering IPyWidgets

philippjfr:
When rendering IPyWidgets in a notebook there was a race condition which meant that because lazy_initialize was slow you could end up with multiple Views for the same model.

@mattpap While I realize I was inappropriately using lazy_initialize here I think this revealed a bokeh bug where there could be a race condition where `this._child_views` was reset while a view was still initializing meaning that it was never cleaned up (or reused).  

Fixes https://github.com/holoviz/panel/issues/5459

### DIFF
--- a/panel/models/ipywidget.ts
+++ b/panel/models/ipywidget.ts
@@ -11,8 +11,9 @@ export class IPyWidgetView extends HTMLBoxView {
   private ipychildren: any[]
   private manager: any
 
-  override async lazy_initialize(): Promise<void> {
-    await super.lazy_initialize()
+
+  initialize(): void {
+    super.initialize()
     let manager: any
     if ((Jupyter != null) && (Jupyter.notebook != null))
       manager = Jupyter.notebook.kernel.widget_manager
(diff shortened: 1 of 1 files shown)


panel/models/ipywidget.ts
```diff
@@ -59,13 +54,26 @@ export class IPyWidgetView extends HTMLBoxView {
...
-        child.trigger('displayed', child)
-      this.invalidate_layout()
-    }
+    const {spec, state} = this.model.bundle
+    this.manager.set_state(state).then(async (models: any) => {
+      const model = models.find((item: any) => item.model_id == spec.model_id)
+      if (model == null)
+	return
+
+      this.manager.create_view(model, {el: this.el}).then(async (view: any) => {
```
mattpap, line 63: Given the enclosing function is `async`, you can simply use `const view = await this.manager.create_view ...`.

philippjfr: Indeed.

PR #5470, opened 2023-09-01
Statically type globals

mattpap:
This PR removes `window as any` and switches the code base using type declarations for globals and declared types. For now only one module was changed `ace.ts`, which ended up being a simple change, yet addressing all important points regarding typing globals. I will do some more next week.

Unrelated, but recently I switched bokehjs to using isolated modules and verbatim module syntax, which means there are restrictions being placed on re-exports of types and that types must be imported explicitly and separately from values (using `import type` syntax), so I went ahead and applied relevant changes to panel. This change only affects development of panel and is effectively future proofing for when the new build system for bokehjs will arrive.


philippjfr:
Will merge since it's a start that we can build on.

PR #5482, opened 2023-09-07
FileInput seek 0

MarcSkovMadsen:
The documentation states

image

It does not mention that every user has to run `.seek(0)`. This is a major pain and friction.

This PR tries to address this.

### DIFF
--- a/panel/widgets/input.py
+++ b/panel/widgets/input.py
@@ -227,8 +227,8 @@ class FileInput(Widget):
                     f.write(val)
             else:
                 fn.write(val)
-                if hasattr(val, "seek"):
-                    val.seek(0)
+                if hasattr(fn, "seek"):
+                    fn.seek(0)
 
 
 class StaticText(Widget):
--- a/panel/widgets/input.py
+++ b/panel/widgets/input.py
@@ -227,6 +227,8 @@ class FileInput(Widget):
                     f.write(val)
             else:
                 fn.write(val)
+                if hasattr(val, "seek"):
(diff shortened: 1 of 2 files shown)


MarcSkovMadsen:
I also added an example to the `FileInput` reference motivated by https://discourse.holoviz.org/t/how-to-handle-the-scenario-is-which-a-figure-updates-based-on-a-selected-data-set-but-no-data-has-been-chosen-yet/5986. I've seen users ask many times how to use the `FileInput` to upload files, convert them and also handle the situation where no file has been uploaded yet.

panel/widgets/input.py
```diff
@@ -227,6 +227,8 @@ def save(self, filename):
                     f.write(val)
             else:
                 fn.write(val)
+                if hasattr(fn, "seek"):
```
MarcSkovMadsen, line 230: I've only seen `BytestIO` as `fn`. But maybe others need to be supported as well?

philippjfr:
While I can see the argument for this I don't think it's idiomatic. Writing to a file like object generally writes and does not then seek to start of the file, e.g. take matplotlib `Figure.savefig` or Altair's `Chart.save()`.

MarcSkovMadsen:
I did not know that. What i know is that the current functionality caused pain for users. Any other suggestions for Solvang that?

philippjfr:
My only other solution would be to add a helper to return a BytesIO object. I'm -1 on this specific change because its the idiomatic behavior.

PR #5490, opened 2023-09-08
Fix autoreload

### LINKED ISSUE #5489: --autoreload does not work in Panel 1.2.2
Run this code with `panel serve --autoreload tmp.py` and change the end to 2 the 

``` python
# tmp.py
import panel as pn

pn.widgets.IntSlider(end=1).servable()
```

A git bisect indicates https://github.com/holoviz/panel/pull/5344 is the culprit.

Hoxbro:
Fixes  #5489

`self.count` has a default of None (meaning unlimited), which will always give an exception out when comparing to `self.counter`

### DIFF
--- a/panel/io/callbacks.py
+++ b/panel/io/callbacks.py
@@ -84,7 +84,7 @@ class PeriodicCallback(param.Parameterized):
             with set_curdoc(self._doc):
                 if self.running:
                     self.counter += 1
-                    if self.counter > self.count:
+                    if self.count is not None and self.counter > self.count:
                         self.stop()
                 cb = self.callback() if self.running else None
         except Exception:
--- a/panel/tests/ui/command/test_serve.py
+++ b/panel/tests/ui/command/test_serve.py
@@ -29,7 +29,8 @@ def test_autoreload_app(py_file, port, page):
         page.goto(f"http://localhost:{port}/{app_name}")
         expect(page.locator(".markdown")).to_have_text("Example 1")
 
-        time.sleep(0.5)
+        # Timeout to ensure websocket is initialized
+        time.sleep(1.0)
 
         write_file(app2, py_file.file)
         expect(page.locator(".markdown")).to_have_text('Example 2')
--- a/panel/tests/ui/command/test_serve.py
+++ b/panel/tests/ui/command/test_serve.py
@@ -1,0 +1,33 @@
+import os
+import time
+
+import pytest
+
+from panel.tests.util import (
+    run_panel_serve, unix_only, wait_for_port, write_file,
+)
+
+try:
+    from playwright.sync_api import expect
+    pytestmark = pytest.mark.ui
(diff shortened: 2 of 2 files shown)


maximlt:
Is it possible to test this? Breaking `--autoreload` is a pretty bad regression :(

Hoxbro:
> Is it possible to test this? 

I quickly tried to do something with `subprocess` but found it difficult to get up and running. Having a type checker running (which works with Param) could have detected this error. 

> Breaking `--autoreload` is a pretty bad regression :(

Yep, and it is all PeriodicCallback, which does not have a `count` set that does not work. I would suggest we release a micro-release at the start of next week with this fix in it.

philippjfr:
Yes, let's release 1.2.3 asap. Definitely need a test here.

Hoxbro:
I have been trying to get a test to run with the following, which works fine if I use something like: `pytest panel/tests --ui -k test_autoreload_app --slowmo` but not when I remove `--slowmo`.



``` python
import os

import pytest
import time

from panel.tests.util import (
    run_panel_serve, unix_only, wait_for_port, write_file
)

try:
    from playwright.sync_api import expect
    pytestmark = pytest.mark.ui
except ImportError:
    pytestmark = pytest.mark.skip('playwright not available')


@unix_only
def test_autoreload_app(py_file, port, page):
    app = "import panel as pn; pn.Row('Example 1').servable()"
    app2 = "import panel as pn; pn.Row('Example 2').servable()"
    write_file(app, py_file.file)

    app_name = os.path.basename(py_file.name)[:-3]

    with run_panel_serve(["--port", str(port), '--autoreload', py_file.name]) as p:
        port = wait_for_port(p.stdout)
        time.sleep(0.2)

        page.goto(f"http://localhost:{port}/{app_name}")
        expect(page.locator(".markdown")).to_have_text("Example 1")

        write_file(app2, py_file.file)
        expect(page.locator(".markdown")).to_have_text('Example 2')
```

philippjfr:
Can you add the test to the PR, then I'll iterate.

Hoxbro:
Done. I want to note we can't use `page.goto` after `write_file` as this also worked without this fix.

philippjfr:
Added a timeout, which isn't elegant but should work. The problem here is that I guess is that the file is changed before the websocket is fully open, which means it never receives the event.

Wonder if maybe we should put a `Document.hold()` while the app is initializing and then unhold when the `DocumentReady` event fires.

Hoxbro:
I feel like I tried all combinations of `time.sleep` and `page.wait_until_timeout` to get this to work. Apparently, I did not...

Thank you for the solution :+1:

philippjfr:
If this passes I'll merge as is, because the .hold() + .unhold() thing seems like a good idea but has the potential to break a ton of stuff, so should be tackled separately.

Issue #5494, opened 2023-09-08
MyST renderer does not work with Panel 1.2.2


monodera:
#### ALL software version info
(this library, plus any other relevant software, e.g. bokeh, python, notebook, OS, browser, etc)

panel 1.2.2
bokeh 3.2.2
myst-parser 2.0.0
Markdown 3.4.4
markdown-it-py 3.0.0
Python 3.11.5

macOS 13.5.2
Firefox 117.0

##### Expected behavior

Markdown pane is rendered correctly. 

##### Observed behavior

Got following error message.

```
TypeError: MdParserConfig.init() got an unexpected keyword argument 'breaks'

Traceback (most recent call last):
  File "/Users/monodera/tmp/test_panel/.venv/lib/python3.11/site-packages/bokeh/application/handlers/code_runner.py", line 229, in run
    exec(self._code, module.__dict__)
  File "/Users/monodera/tmp/test_panel/markdown_test.py", line 5, in <module>
    pn.Row(pn.pane.Markdown("# Markdown Text", renderer="myst")).servable()
  File "/Users/monodera/tmp/test_panel/.venv/lib/python3.11/site-packages/panel/viewable.py", line 392, in servable
    self.server_doc(title=title, location=location) # type: ignore
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/monodera/tmp/test_panel/.venv/lib/python3.11/site-packages/panel/viewable.py", line 1024, in server_doc
    model = self.get_root(doc)
            ^^^^^^^^^^^^^^^^^^
  File "/Users/monodera/tmp/test_panel/.venv/lib/python3.11/site-packages/panel/layout/base.py", line 306, in get_root
    root = super().get_root(doc, comm, preprocess)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/monodera/tmp/test_panel/.venv/lib/python3.11/site-packages/panel/viewable.py", line 666, in get_root
    root = self._get_model(doc, comm=comm)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/monodera/tmp/test_panel/.venv/lib/python3.11/site-packages/panel/layout/base.py", line 174, in _get_model
    objects, _ = self._get_objects(model, [], doc, root, comm)
[...]

PR #5497, opened 2023-09-09
Fix markdown rendering with the MyST parser (#5494)

monodera:
MyST parser does not work with Panel 1.2.2 (#5494) and the commit fixes it.


monodera:
BTW, the fix was suggested by @ahuang11 and I just followed it.

philippjfr:
Thanks @monodera, appreciate the contribution!

PR #5511, opened 2023-09-14
Ensure Quill correctly detects selections and renders HTML

philippjfr:
Some fixes for Quill rendering inside the shadow DOM.

### DIFF
--- a/panel/models/quill.py
+++ b/panel/models/quill.py
@@ -13,12 +13,12 @@ class QuillInput(HTMLBox):
     """
 
     __css_raw__ = [
-        'https://cdn.quilljs.com/1.3.6/quill.bubble.css',
-        'https://cdn.quilljs.com/1.3.6/quill.snow.css'
+        'https://cdn.quilljs.com/1.3.7/quill.bubble.css',
+        'https://cdn.quilljs.com/1.3.7/quill.snow.css'
     ]
 
     __javascript_raw__ = [
-        'https://cdn.quilljs.com/1.3.6/quill.js',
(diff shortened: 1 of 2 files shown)


maximlt:
This PR was motivated by a few issues. The first one being that in some situations, re-rendering the widget would affect its content (not the model though). This can be reproduced with the following code, adding a bullet list in the editor, switching to the next view and back to the editor, where you should see the bullet lists are gone.

```python
import panel as pn

pn.extension('texteditor')

contents = [pn.widgets.TextEditor(), pn.pane.HTML('<h1>Bob</h1>')]
contents = [pn.Column(c) for c in contents]
w_radio = pn.widgets.RadioButtonGroup(options=[0, 1])

placeholder = pn.Column(contents[0])


@pn.depends(w_radio, watch=True)
def update_view(i):
    placeholder[:] = [contents[i]]


pn.Column(w_radio, placeholder).servable()
```

As it currently stands, this PR **fixes** that issue.

The other issue I saw with the editor was that clicking on one of the tools, like *B*(old), the cursor would always move up to the first position in the editor. This is **fixed** too. **However**, there is still an issue with the tools. When I click on one of them like *B*(old), and type something, the text isn't in bold. The tool doesn't get activated (or it does but very shortly). I can set some bit of text in bold if I select it and then hit on tool.

philippjfr:
Unfortunately I wasn't able to find a fix for the last issue yet (and no one on StackOverflow or the Quill issue tracker seems to have done so either). I'll merge for now so I can cut a dev release.

Issue #5492, opened 2023-09-08
OAuth login includes prefix twice in the login url


kuykendall-ben:
Minor bug related to https://github.com/holoviz/panel/pull/5422. The login endpoint configured with OAuth is now `/<prefix>/<prefix>/login` which appears to work, but can't possibly be the intended path.

#### ALL software version info
Panel: 1.2.2
Bokeh: 3.2.0
Tornado: 6.2
Python: 3.10.13
OAuth provider: GitHub
Platform: macOS Ventura 13.5.1

#### Description of expected behavior and the observed behavior
Visiting `http://localhost:5006/prefix/app` should redirect to `http://localhost:5006/prefix/login` to start the authentication flow. Instead it redirects to `http://localhost:5006/prefix/prefix//login`. Login does proceed correctly.

#### Complete, minimal, self-contained example code that reproduces the issue
Create an OAuth app in GitHub per https://docs.github.com/en/apps/oauth-apps/building-oauth-apps/creating-an-oauth-app. Set the "authorization callback URL" to `http://localhost:5006/prefix/`. Generate a "client secret", and save the key and secret for use below.

Write a minimal `app.py`
```
import panel as pn
pn.Column('test').servable()
```

Start the panel server with OAuth parameters and a prefix. Make sure to include the prefix (including the trailing `/`) in the OAuth redirect uri
```
panel serve app.py --port 5006  --cookie-secret `panel secret` \
  --oauth-provider github --oauth-key $KEY --oauth-secret $SECRET \
  --oauth-redirect-uri http://localhost:5006/prefix/ \
  --prefix prefix
```
Visit `http://localhost:5006/prefix/app`. It will redirect to `http://localhost:5006/prefix/prefix//login?next=/prefix/app`. The authentication flow will start; optionally sign in with GitHub and eventually get redirected to `http://localhost:5006/prefix/app`. Verify the page renders.

#### Screenshots or screencasts of the bug in action
Included are screenshots of the first two requests in the authentication flow.

The initial `GET /prefix/app` request returns with a redirect to `./prefix//login?next=/prefix/app` 

<img width="998" alt="Screenshot 2023-09-08 at 11 32 46 AM" src="https://github.com/holoviz/panel/assets/3002978/b3137bd7-029e-4ddd-bf6f-252d06fab5af">

This resolves to `GET /prefix/prefix//login` (?!) which redirects to the OAuth provider as desired

[...]

PR #5514, opened 2023-09-15
Allow configuring the logout page

philippjfr:
Instead of immediately directing a user back to the login page this PR ensures that the `/logout` endpoint serves a page that confirms logging out and allows the user to click a button to log back in. It also makes this template configurable and adds documentation.

Additionally this PR fixes some issues with redirects, particularly when a prefix is provided.

Fixes https://github.com/holoviz/panel/issues/5492


panel/tests/ui/test_auth.py
```diff
@@ -35,3 +35,37 @@ def test_basic_auth(py_file, page, prefix):
         page.get_by_role("button").click(force=True)
 
         expect(page.locator('.markdown')).to_have_text('test_user', timeout=10000)
+
+
+#@unix_only
```
philippjfr, line 40:
```suggestion
@unix_only
```

PR #5517, opened 2023-09-16
Param 2: adapt one test to updated error message

maximlt:
Adapting one test to an improved error message changed in https://github.com/holoviz/param/pull/824 that will be in Param 2.0.0a3 (not yet release)


philippjfr (approved):
Thank you!

PR #5518, opened 2023-09-16
Param 2: skip custom logic for the private namespace when instantiating the config object

maximlt:
Without this change https://github.com/holoviz/param/pull/834 would make Panel fail on import when instantiating its config object.


philippjfr (approved):
Thank you!

PR #5525, opened 2023-09-20
Add Panel tutorials

philippjfr:
- [x] 01. **Param**: Basics for Panel [@philippjfr]
  - What is a Parameter?
  - Getting and setting the value
  - Parameters as references (convering indirect references via `bind` and eventually reactive expressions)
  - Writing parameterized classes with `depends`, using  the `pn.Param` pane and `Widget.from_param`
- [x] 02. **Components**: Core Components [@maximlt]
  - What are core component types?
    - Panes: Display some object, can be updated by setting `.object`
    - Widget: Allow getting user input, act as a proxy for their `.value` parameter
    - Layout: Two main types `ListLike` and `GridLike`, contents can be replaced
  - Briefly go over some of the main parameters shared by all components, i.e. `margin`, `loading`, `width`, `height`, `styles` etc.
- [x] 03. **Interactivity**: Effectively using APIs for adding interactivity and reactivity [@philippjfr]
  - Recap parameters as references
  - Go over reactive vs. imperative
  - Go over function vs. class based 
- [x] 04. **Efficient Development**: Debugging and effective development in VSCode and JupyterLab [@maximlt]
  - JupyterLab
    - Using the panel-preview button and endpoint
  - VSCode
    - Using `--autoreload` on the commandline
    - Using the debugger
    - Maybe just add screenshots (will try to do this live as part of the training)  
- [x] 05. **Layouts**: Achieving responsive layouts [@philippjfr]
  - Explain inherent sizing
  - Explain sizing modes and width/height responsiveness
  - Explain approaches to real screen responsiveness 
- [x] 06. **Structuring your application code**: Classes & Design Patterns [@philippjfr]
  - Recap function vs. class based approaches
  - Begin structuring application with multiple Parameterized classes
  - Use central DataStore pattern that drives the different views
- [x] 07. **Styling**: How to leverage designs, stylesheets and CSS [@philippjfr]
- [x] 08. **Multi-page Apps**: Different approaches for designing multi-page applications [@maximlt]
- [ ] 09. **Scalability**: Async, threading, profiling and caching 
- [ ] 10. **Server**: Effectively working with the server [@philippjfr]
  - Quick explanation of Server, Sessions and how global state is stored
  - Compare CLI vs `pn.serve` (i.e. latter has more control over routing)
  - Discuss serving extra assets and other endpoints 
- [x] 11. **Deep Linking**: Adding deep links and persistent sessions [@ahuang11]
  - Introduce `pn.state.location` and `pn.state.session_args`
  - Start building example that syncs Parameter state
  - Add example of a second app and then sync/unsync depending on the currently selected app
  - Discuss example where there is significant session state and create a unique user ID based on `pn.state.user` (assuming Auth has been configured)   
- [ ] 12. **Authentication and Authorization**: Configuring and effectively using OAuth [@philippjfr]


doc/tutorials/01_Param.ipynb
```diff
@@ -0,0 +1,338 @@
...
+   "source": [
+    "Notice that we passed in the widget object directly instead of the `.param.value`. This is possible because widgets are treated as a proxy of their `value` parameter just like a `Parameter` is treated as a proxy for current value."
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "55f2a3fe-771f-4aff-b2eb-b107e373b735",
+   "metadata": {},
+   "source": [
+    "### Indirect Mapping\n",
```
maximlt, line 278: I thought that `pn.bind` and Parameters as reference belonged to the *Interactivity* guide.

doc/tutorials/01_Param.ipynb
```diff
@@ -0,0 +1,338 @@
...
+  {
+   "cell_type": "markdown",
+   "id": "fdfc0ac9-b516-49b6-a21b-e173b5d2de11",
+   "metadata": {},
+   "source": [
+    "# Param Basics for Panel\n",
+    "\n",
+    "Panel and other projects in the HoloViz ecosystem all build on Param. Param provides a framework to add validation, documentation and interactivity to a project. It is similar to more modern projects such as Pydantic but focuses primarily on providing APIs that make it easy to express complex dependencies and UI interactions.\n",
+    "\n",
+    "Here we won't focus on how param works but instead try to get you to understand the fundamentals that will allow you to effectively use Panel, express interactivity and reactivity in idiomatic ways and lastly how to structure your code to make it easily reusable and avoid callback hell."
```
maximlt, line 36: It seems to me this is going a little too fast over basics Param. If this guide is intended to build complex apps, we'll want to nudge people into writing more Parameterized classes. This guide doesn't really explain how as it currently is.

doc/tutorials/01_Param.ipynb
philippjfr: Parameters as references and pn.bind (and eventually expressions) is definitely core Param material. The Interactivity guide will go significantly beyond these basic concepts. It may turn out that this is not good material as beginner docs, we are definitely aiming to deliver quite advanced material that goes significantly beyond "basics".

MarcSkovMadsen:
Are you sure these are tutorials?

If I look at https://diataxis.fr/tutorials/, tutorials should Lead the user through a series of successfull steps. Not explain anything.

philippjfr:
In all likelihood they will not be added to the docs in their current form, I mainly needed somewhere to put them for now. So no, these are not tutorials in the documentation sense, at least not yet.

>Not explain anything.

This is, *I think*, a mischaracterization, the point Diataxis makes is the explanation is not what they should take away from the material but rather just a means to get them to do something practical. The important thing is to get across **how**  to achieve something, it is imo okay to explain plain facts, i.e. the **that** but only as a means to get them to apply the knowledge. What absolutely shouldn't be in a tutorial is the **why**. 

Even for the current purpose my current material is just a start and I still need to make sure that the existing material:

1. Clearly states what they are going to learn in each section
2. Doing a better job of using the knowledge to get them to build something practical

This material will be the basis of a training course and once I've delivered it I'll see if and how we can adapt it to our docs or whether it'll be completely separate. Likely I will try to split each section into theory and practice and the practice part is what fits the Diataxis tutorial definition.

philippjfr:
That all being said I also struggle a lot here, my goal for this tutorial is to get someone to a relatively deep understanding of Panel in a relatively small amount of time. I realize this won't happen in one big session but will require revisiting these materials a few times. At the same time I struggle to see how a purely practical tutorial without anything stage-setting will be enough to get any understanding or proficiency.

ahuang11:
Perhaps, we should focus on calling these course materials instead of tutorials and we shouldn't try to create both in one go.

maximlt:
We currently have a pretty major problem in Panel's docs, for a new user there's a big hole after going through the Getting Started, that is definitely too short to learn how to build any kind of app that is not trivial. The User Guides we used to have served that purpose somehow, you could go through one after the other. With the How-To's, we've gained content that is useful for when you're working on an app and have already some knowledge, or when you look for something specific, but we've lost that learning experience, going through all the how-to's doesn't feel very natural to me (how-to's are also a better way to scale docs compared to the user guides, which is good!).

There's really one thing you can do with Panel: build an app. The Getting Started covers that quickly, my vision is that we're building an advanced tutorial that users will follow when they're done with the Getting Started, decide to commit with Panel and/or embark on building a more complex app. It's going to be generic, because Panel is a generic framework and that's the HoloViz philosophy. In the future we can still add new tutorials, e.g. build a BI app, build a data app, build a machine learning app, etc.

To have a nice advanced tutorial I believe we will need (not for this PR!):

- an app that we build step by step
- to link to other places in the docs when required, I'm sure we can do a much better job a linking our docs!

We also need to figure out what exactly should go in the `Explanation` category.

philippjfr:
>an app that we build step by step

Are you suggesting that each tutorial section builds up one app step by step growing in complexity each time?

philippjfr:
I've gone back and forth on that idea a lot, my concern is that it'll be very hard to do that well. It's very hard to slowly build up an app in a way that makes sense in terms of a pedagogical progression, i.e. the concepts that you want to teach first may not necessarily overlap with the first steps you do when building an app. It may be easier to simply build the full app and then pull out bits and pieces to allow the user to recreate those through exercises but I'm honestly unsure about that too.

MarcSkovMadsen:
What about taking inspiration from some that are famous for good docs. Like Django, FastApi or Streamlit?

MarcSkovMadsen:
What about taking inspiration from the Streamlit Guide. They start by introducing simple things. It could be layouts, Panes and templates. Dynamic layouts and Streaming with generators are also powerful and relatively simple. They can bring a lot of value already to users. Then later widgets, parameters, bind etc.

cdeil:
My 2 cents: I think this will be super! :-)

My main problem with https://panel.holoviz.org/ to learn Panel was that there's so much content that it's almost impossible to read it all, and it's very difficult to figure out a useful order in which to read pages under "how to" and "explanation" and the galleries. I got lost.

What I like is how here I have 1, 2, 3, ..., 12 with a clear path. I can read through it all and learn Panel thoroughly in a weekend, or of course also omit topics if I don't need them. But it's finite and well-structured. Other docs I like for that reason (clear order and structure) is e.g. https://doc.rust-lang.org/book/ or https://shiny.posit.co/

I think teaching Panel by building a complex app and refactoring it over and over and learning new concepts is also fantastic. I would love to do such a course. But it's very hard to put together. I remember reading https://www.flaskbook.com/ years ago where this is done in Part II, and Part I is more like this tutorial introducing topics one by one with several small apps. So it could come later as Part II for Panel as well?

doc/tutorials/03_Interactivity.ipynb
```diff
@@ -0,0 +1,311 @@
...
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "b0cedaaa-c66d-4ea9-83de-09b8a1e4bc6c",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "turbines = pd.read_parquet('./windturbines.parq')\n",
```
cdeil, line 65: This tutorial doesn't work for me because I don't have windturbines.parq

Maybe you could choose example data for the tutorials that can be readily loaded (e.g. https://docs.bokeh.org/en/latest/docs/examples/basic/data/transform_markers.html)?

Or alternatively link to an online URL or bundle it so that the notebook executes?

philippjfr: I'm probably either going to bundle it or load it from our cdn.

maximlt:
I can only agree with you @cdeil :) We have certainly neglected tutorials a little bit too much (for good reasons! as we improved a lot the remaining content). I think an critical point is that when you have good tutorial/learning content, advanced/interested Panel users can own it more easily which enables them teach their peers better. For instance, if we have some good tutorials, I wouldn't be surprised to see some people leveraging them to produce videos for Youtube.

MarcSkovMadsen:
One question I have is whether the *getting started* section is something separate. This is a tutorial focusing on getting started.

The there is a separate *tutorials* page with other tutorials. Could be Panel topic specific tutorials, domain specific tutorials and maybe even links to some of the best external tutorials shared by the community as blog posts and videos?

maximlt:
> The there is a separate _tutorials_ page with other tutorials. Could be Panel topic specific tutorials, domain specific tutorials and maybe even links to some of the best external tutorials shared by the community as blog posts and videos? These tutorials are read when you need to start learning about some more complex topic. Could be ReactiveHTML, could be AI Chat interfaces, could be using Panel in Finance etc.

I've wanted a `Tutorials` section for a while, that's the only place so far where we diverge completely from Diataxis. I believe the Getting Started should be the first tutorial presented on that page, and that it should be linked from the docs homepage and from the future landing page.

MarcSkovMadsen:
As inspiration for learning about a web frameworks features see https://eugenkiss.github.io/7guis/tasks

cdeil:
Any chance to get this merged and available in the docs?

It's inactive since over 2 months.

Could it just go in as-is and then see incremental improvements moving forward?
Or is some big refactor or streamlining or small improvements still needed here?

Let me know if there's anything specific I could help with - would love to see this in the docs and take some (admittedly limited) time to help with edits or review here.

doc/tutorials/11_Deep_Linking.ipynb
```diff
@@ -0,0 +1,339 @@
...
+    "[Click here to check if your modifications work!](/panel-preview/render/exercise_5.py)\n",
+    "\n",
+    "For an additional challenge, picture a scenario where you manage a trading platform. The platform's users have their portfolios, which consist of lists of stock tickers. Your challenge is to display only the tickers that a user owns, based on `pn.state.user`."
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## Conclusion\n",
```
cdeil, line 294: I would recommend to remove this section from this deep linking tutorial.

It contains a list of 15 data-driven app ideas or or classes of apps that could be nicely built in Panel, but would each take a day or a week or a month depending on skill level.

To me this feel it's not really related to this tutorial specifically, and personally I like short and to the point tutorials that end with one working example and exercise - like this one perfectly does with the stock explorer - and then want to move on to the next tutorial.  :-)

philippjfr:
I'm merging tutorials 1-7 from above and will continue iterating on them.

doc/tutorials/index.md
```diff
@@ -0,0 +1,73 @@
...
+:link-type: doc
+
+In this section we will take you through the process of structuring a more complex application, discussing different approaches for managing reactivity and how to create a composable architecture for your applications.
+:::
+
+:::{grid-item-card} {octicon}`paintbrush;2.5em;sd-mr-1` Styling
+:link: styling
+:link-type: doc
+
+In this section we will take you through the process of structuring a more complex application, discussing different approaches for managing reactivity and how to create a composable architecture for your applications.
```
cdeil, line 56: I think here is a copy & paste error - card description should be for styling tutorial?

PR #5527, opened 2023-09-22
Fix a typo in the FileInput notebook

monodera:
Conversion from MB to bytes was done by multiplying MB by 1024*1014. This commit fix it to 1024*1024.


philippjfr:
Thank you!

PR #5531, opened 2023-09-24
Improved doc on deploying with GCP

pierrotsmnrd:
Following up an experience I had with GCP, I'm updating the docs to share an important caveat.

When deploying with Cloud Run, it's important to tell it to extend the default timeout for long-running HTTP requests.
Otherwise it kills the websocket to Bokeh after the default 5 minutes, leading to an unresponsive UI, forcing the user to refresh the page.


philippjfr (approved):
Perfect, thank you @pierrotsmnrd!

PR #5532, opened 2023-09-24
Fixes for Column invisible scroll_button taking space

philippjfr:
Columns showed a bogus scrollbar due to the scroll button.

### DIFF
--- a/panel/dist/css/listpanel.css
+++ b/panel/dist/css/listpanel.css
@@ -13,8 +13,8 @@
 .scroll-button {
   /* For location */
   position: sticky;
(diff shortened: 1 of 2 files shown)


ahuang11:
I think the arrow is now at the top right rather than bottom right with this change.
image

ahuang11:
I think re-setting `top: calc(100% - 38px);` fixes it.

philippjfr:
Cannot reproduce that, still at the bottom right for me.

ahuang11:
Weird, I was using Google Chrome. I also deleted my cache.

PR #5547, opened 2023-09-27
Add authorization_code and password based OAuth login handlers

philippjfr:
So far our OAuth login handlers all used authorization flows that required a `client_secret`. Our new workflow works by using the code challenge auth flow, which redirects the user to a log in page along with a SHA encoded code challenge that is then verified against the verification code that is stored in a cookie.

We also add a handler for the password based grant types, which is simple to set up for development but should generally not be used.

- [x] Add docs
- [ ] Add tests


philippjfr:
@cdeil Would love it if you guys could try this one out on your end.

armaaar:
> If its easy to make it mandatory for specific pages or optional the it would be really valuable.

Hello @philippjfr , Can you please share with us how can we test this PR with our app?

doc/how_to/authentication/configuration.md
```diff
@@ -12,6 +12,9 @@ The first step in configuring a OAuth is to specify a specific OAuth provider. P
 * `gitlab`: GitLab
 * `google`: Google
 * `okta`: Okta
+* `generic`: Generic OAuth Provider with configurable endpoints
+* `password`: Generic password grant based OAuth Provider with configurable endpoints
+* `code`: Generic code challenge grant based OAuth Provider with configurable endpoints
```
ratanasov, line 17: `code` might be somewhat ambiguous name, since there is also a device code flow. What about `auth_code`? Right, @armaaar ?

armaaar: I agree. There is also an OAuth flow called Device Code flow: https://oauth.net/2/grant-types/device-code/
This name might cause confusion

doc/how_to/authentication/providers.md
```diff
@@ -32,6 +32,16 @@ Google provides a guide about [configuring a OAuth application](https://develope
...
 
+## **Generic**/**Password**/**Code**
+
+The `'generic'`, `'password'` and `'code'` OAuth providers allows you to provide custom authentication endpoints using the `--oauth-extra-param` or using environment variables. Specifically you must provide:
+
+- `AUTHORIZE_URL`: The authorization endpoint of the authentication server, may also be provided using the `PANEL_OAUTH_AUTHORIZE_URL` environment variable.
+- `TOKEN_URL`: The token endpoint of the authentication server, may also be provided using the `PANEL_OAUTH_TOKEN_URL` environment variable.
+- `USER_URL`: The user information endpoint of the authentication server, may also be provided using the `PANEL_OAUTH_USER_URL` environment variable.
+
+The difference between these three providers is the authentication flow they perform. The `generic` provider uses the standard authentication flow, which requests authorization using the client secret, while the `password` based workflow lets the user log in via a form served on the server (only recommended for testing and development), and the `code` uses a code challenge based auth flow.
```
armaaar, line 43: There is nothing call "standard authentication flow", I guess you mean client credentials flow: https://oauth.net/2/grant-types/client-credentials/

panel/auth.py
```diff
@@ -891,26 +985,94 @@ def set_current_user(self, user):
...
+
+class OAuthProvider(BasicAuthProvider):
+    """
+    An AuthProvider using specific OAuth implementation selected via
+    the global config.oauth_provider configuration.
+    """
+
+    @property
+    def login_handler(self):
+        handler = AUTH_PROVIDERS[config.oauth_provider]
```
armaaar, line 1069: If we Do:
```
pn.serve(..., oauth_provider="code")
```

We get this error
```
 File "***/serve_app.py", line 135, in serve_app
    pn.serve(
  File "***/panel/io/server.py", line 922, in serve
    return get_server(panels, **kwargs)
  File "***/panel/io/server.py", line 1175, in get_server
    opts['auth_provider'] = provider(
  File "***/panel/auth.py", line 1032, in __init__
    super().__init__()
  File "***/lib/python3.10/site-packages/bokeh/server/auth_provider.py", line 70, in __init__
    self._validate()
  File "***/lib/python3.10/site-packages/bokeh/server/auth_provider.py", line 179, in _validate
    if self.login_handler and self.get_login_url:
  File "***/panel/auth.py", line 1069, in login_handler
    handler = AUTH_PROVIDERS[config.oauth_provider]
KeyError: None
```

This is because you don't set `config.oauth_provider` in `panel/io/server.py` the same way you set it in `panel/command/serve.py`

panel/auth.py
```diff
@@ -106,6 +110,8 @@ class OAuthLoginHandler(tornado.web.RequestHandler):
 
     _error_template = ERROR_TEMPLATE
 
+    _login_endpoint = '/login'
```
armaaar, line 113: You are hard coding the login endpoint here. if we use:
```
pn.serve(..., login_endpoint="/auth")
```
This will result a redirect to `/login` with a 404 error:
```
404 GET /login (::1) 1.10ms
```

panel/auth.py
```diff
@@ -891,26 +985,94 @@ def set_current_user(self, user):
         self.set_secure_cookie('id_token', id_token, expires_days=config.oauth_expiry)
 
 
+class LogoutHandler(tornado.web.RequestHandler):
+
+    _login_endpoint = '/login'
```
armaaar, line 990: Same here about hardcoding login endpoint

panel/auth.py
```diff
@@ -354,15 +377,22 @@ def _on_error(self, response, body=None):
...
+        self.write(self._error_template.render(
+            npm_cdn=config.npm_cdn,
+            title='Panel: Authentication Error',
+            error_type='Authentication Error',
+            error=body.get('error', 'Unknown Error'),
+            error_msg=body.get('error_description', body)
+        ))
 
 
 class GenericLoginHandler(OAuthLoginHandler, OAuth2Mixin):
```
armaaar, line 390: Isn't there a way to update the scope other than defining `PANEL_OAUTH_SCOPE`? we would like to have it as a config option the same the URLs are

panel/io/server.py
armaaar, line 1: On line 1206:
You are hard coding the login endpoint here. if we use:
```
pn.serve(..., show=True, login_endpoint="/auth")
```
This will result a redirect to `/login` with a 404 error:
```
404 GET /login (::1) 1.10ms
```

panel/auth.py
armaaar, line 1: We get this error
```
Uncaught exception GET /auth (::1)
HTTPServerRequest(protocol='http', host='localhost:5005', method='GET', uri='/auth', version='HTTP/1.1', remote_ip='::1')
  File "***/lib/python3.10/site-packages/tornado/web.py", line 1784, in _execute
    result = method(*self.path_args, **self.path_kwargs)
  File "***/lib/python3.10/site-packages/tornado/web.py", line 3276, in wrapper
    if not self.current_user:
  File "***/lib/python3.10/site-packages/tornado/web.py", line 1420, in current_user
    self._current_user = self.get_current_user()
  File "***/lib/python3.10/site-packages/bokeh/server/views/auth_request_handler.py", line 71, in get_current_user
    return self.application.auth_provider.get_user(self)
  File "***/panel/auth.py", line 1038, in get_user
    return request_handler.get_secure_cookie("user", max_age_days=config.oauth_expiry)
  File "***/lib/python3.10/site-packages/tornado/web.py", line 836, in get_signed_cookie
    self.require_setting("cookie_secret", "secure cookies")
  File "***/lib/python3.10/site-packages/tornado/web.py", line 1669, in require_setting
    raise Exception(
Exception: You must define the 'cookie_secret' setting in your application to use secure cookies
500 GET /auth (::1) 22.92ms
```
I thought you create this by default. Do we need to always supply it? if so, shouldn't it be documented that this is required?

armaaar:
The solution doesn't work for us. It results in man
Screenshot 2023-09-28 at 10 54 43 AM
y redirects:

armaaar:
We have a huge problem with not being able to logout.
Currently there is no logout handler for `OAuthProvider`. That means that once a user login, they can't logout.
This is not the only problem, the current `LogoutHandler` that is used in `BasicAuthProvider` can't be used in our use case as well. To fully understand why, take this into consideration:

1. In order to fully logout, we end our session and revoke our token from the identity server. If not, we will always be logged in.
2. The `LogoutHandler` only removes cookies, but doesn't end the session or revoke the token
3. Panel would always redirect to the Auth provider portal if not authenticated

Now if we assume that there is a logout button, have a look into this scenario:
1. You visit Panel app, you visit panel login page
2. you are redirected to auth portal
3. you login in portal, and get redirected back to panel app
4. you logout, cookies are cleared, you are redirected to login page
5. you are redirected to auth portal again
6. Since you are still logged in in the portal, you are immediately logged in again and redirected to panel app
7. you are not logged out as intended

Just clearing cookies, and not allowing viewing any page without being authenticated, leads to always being logged in after the first login attempt. without the ability to ever logout. This is unacceptable for us. It also prevents people with multiple accounts to view anything as they are locked to the first account they logged in with

panel/auth.py
```diff
@@ -388,6 +418,126 @@ def _USER_KEY(self):
...
+        if self._access_token_header:
+            user_url = self._OAUTH_USER_URL
+            user_headers['Authorization'] = self._access_token_header.format(
+                body['access_token']
+            )
+        else:
+            user_url = '{}{}'.format(self._OAUTH_USER_URL, body['access_token'])
+
+        user_response = await http.fetch(user_url, headers=user_headers)
+        user = decode_response_body(user_response)
```
armaaar, line 484: In our use case, The user url is a custom URL which  uses one of the decoded properties of the access token as one of its path parameters. We can't supply a fixed url.
We actually don't care about the user info so much, either Make this optional, or provide a way (like a hook) to set the user info ourselves

doc/how_to/authentication/configuration.md
philippjfr: Makes sense to me, thanks!

panel/auth.py
philippjfr: Indeed, should expose it this way.

panel/auth.py
philippjfr: I'll have to retest this but when I tried it with your standard endpoint I got back a valid response containing my user information.

philippjfr:
Appreciate the review comments, all reasonable requests.

doc/how_to/authentication/providers.md
philippjfr: Probably going to split this guide up into separate pages for the different grant types to clarify this.

panel/auth.py
philippjfr: Did you set `openid` as one of the scopes?

panel/auth.py
philippjfr: Thanks, fixed.

panel/auth.py
armaaar: I couldn't reach that point due to the infinite redirects bug

doc/how_to/authentication/configuration.md
cdeil: Maybe link from the docs here or elsewhere to the API docs to make it easy to explore the details where reading this?

With search it's possible to find e.g. [panel.auth.GenericLoginHandler](https://panel.holoviz.org/api/panel.auth.html#panel.auth.GenericLoginHandler) and other relevant classes, but at least for me having a link to click when reading here would be nice to have.

philippjfr:
Going to merge this PR and then continue improving logout and access_token refresh handling as well as the docs in another PR.

PR #5557, opened 2023-09-29
Fix authorize_info in user_info.md

cdeil:
Fix docs typo: `authorization_callback` -> `authorize_callback`.

Also the docs say it can be configured via `pn.config`  and `pn.extension`, but that's not mentioned in their docs:

* https://holoviz-dev.github.io/panel/api/config.html
* https://holoviz-dev.github.io/panel/api/panel.config.html#panel.config.panel_extension

If someone knows how to add document the option there, please do.

### DIFF
--- a/doc/how_to/authentication/user_info.md
+++ b/doc/how_to/authentication/user_info.md
@@ -11,9 +11,9 @@ Once a user is authorized with the chosen OAuth provider certain user informatio
 
 ## Authorization callbacks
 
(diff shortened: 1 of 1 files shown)


philippjfr:
Thank you, I'll make sure to update those when I add support for defining redirects in the authorize callback.

philippjfr (approved):
Looks good, thanks!

PR #5565, opened 2023-10-02
Replace _ignored_refs with explicit allow_refs

philippjfr:
Use `allow_refs` definitions as ultimate source of truth for enabling/disabling allow_refs.


philippjfr:
Remaining failures are resolved in https://github.com/holoviz/holoviews/pull/5918

Issue #5544, opened 2023-09-27
Unnecessary scroll bar when using Column


aditya3434:
When rendering panel column with multiple row in Jupyter Lab/Notebook, a redundant scrollbar shows up even though there appears to be no overflow. 

#### ALL software version info
Panel version 1.2.x and Bokeh version 3.x

#### Description of expected behavior and the observed behavior
Expected behavior : All components rendered without scrollbar
Actual behavior: A scrollbar is displayed even though there is no overflow

#### Complete, minimal, self-contained example code that reproduces the issue

```
import panel as pn
pn.extension()

pn.Column(pn.Row(pn.pane.HTML("My panel")), pn.Row(pn.widgets.Button(name='Click me', button_type='primary'), pn.widgets.IntSlider(start=0, end=10, value=5)))
```

#### Stack traceback and/or browser JavaScript console output
No stack traceback or console error

#### Screenshots or screencasts of the bug in action
![Animation](https://github.com/holoviz/panel/assets/56246494/d29f0405-ea4d-48b0-936e-5cb09b2553aa)

- [ ] I may be interested in making a pull request to address this

maximlt:
Hi @aditya3434 and thanks for this clear bug report 👍 

I believe this is problem was due to the invisible scroll button and was fixed in https://github.com/holoviz/panel/pull/5532.
![image](https://github.com/holoviz/panel/assets/35924738/4fc424c5-7be5-44bc-bd32-db8748446c52)

I'm going to close this issue, feel free to comment in the issue if it's not fixed in the next release.

PR #5566, opened 2023-10-02
Make scroll button more robust

ahuang11:
Patches https://github.com/holoviz/panel/pull/5532 where the scroll_button landed on top right:
image

Now:
image

Also, it still fixes https://github.com/holoviz/panel/issues/5544
image

### DIFF
--- a/panel/models/column.ts
+++ b/panel/models/column.ts
@@ -49,7 +49,6 @@ export class ColumnView extends BkColumnView {
   toggle_scroll_button(): void {
     const threshold = this.model.scroll_button_threshold
     const exceeds_threshold = this.distance_from_latest >= threshold
-    // if scroll_down_button_el is not null
     if (this.scroll_down_button_el) {
       this.scroll_down_button_el.classList.toggle(
         "visible", threshold !== 0 && exceeds_threshold
--- a/panel/models/column.ts
+++ b/panel/models/column.ts
@@ -43,7 +43,7 @@ export class ColumnView extends BkColumnView {
   }
 
   record_scroll_position(): void {
-    this.model.scroll_position = this.el.scrollTop;
+    this.model.scroll_position = Math.round(this.el.scrollTop);
   }
 
   toggle_scroll_button(): void {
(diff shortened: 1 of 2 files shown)


philippjfr:
I still cannot reproduce your issue here, but I'd like to make two other changes: 

1. Change it to `bottom: 10px`
2. Change the unicode arrow to an inlined base64 encoded SVG arrow because rendering is otherwise inconsistent across browsers.

panel/models/column.ts
```diff
@@ -43,7 +43,7 @@ export class ColumnView extends BkColumnView {
   }
 
   record_scroll_position(): void {
-    this.model.scroll_position = this.el.scrollTop;
+    this.model.scroll_position = Math.round(this.el.scrollTop);
```
ahuang11, line 46: scrollTop can be a float; scroll_position is defined as param.Integer so this caused a lot of console errors.

ahuang11:
```python
import panel as pn

pn.extension()


def append(event):
    col.append(len(col))


col = pn.Column(
    *list(range(0, 1000)),
    height=500,
    view_latest=True,
    scroll_button_threshold=20,
    auto_scroll_limit=100,
    sizing_mode="stretch_width"
)


button = pn.widgets.Button(name="Append", button_type="primary", on_click=append)
pn.Row(col, button).servable()

```

image

panel/models/column.ts
```diff
@@ -43,15 +43,18 @@ export class ColumnView extends BkColumnView {
...
   }
 
   toggle_scroll_button(): void {
     const threshold = this.model.scroll_button_threshold
     const exceeds_threshold = this.distance_from_latest >= threshold
-    this.scroll_down_button_el.classList.toggle(
-      "visible", threshold !== 0 && exceeds_threshold
-    )
+    // if scroll_down_button_el is not null
```
ahuang11, line 52:
```suggestion
```

panel/models/column.ts
```diff
@@ -43,15 +43,17 @@ export class ColumnView extends BkColumnView {
...
+    this.model.scroll_position = Math.round(this.el.scrollTop);
   }
 
   toggle_scroll_button(): void {
     const threshold = this.model.scroll_button_threshold
     const exceeds_threshold = this.distance_from_latest >= threshold
-    this.scroll_down_button_el.classList.toggle(
-      "visible", threshold !== 0 && exceeds_threshold
-    )
+    if (this.scroll_down_button_el) {
```
ahuang11, line 52: This fixes https://github.com/holoviz-topics/panel-chat-examples/issues/37

philippjfr (approved):
Looks good!

PR #5578, opened 2023-10-04
Add ChatTextArea that shrinks/grows

ahuang11:
grows/shrinks based on input value

https://github.com/holoviz/panel/assets/15331990/fecb3f42-5a46-4190-a532-1c38589d3159

- [ ] limit to max height
- [ ] figure out how to use stretch_width instead of defining in styles
- [ ] press enter/return to send (if only one row) else enter/return is used to start new line (no idea if this is possible in panel)


MarcSkovMadsen:
I love this idea

Please consider making it work like ChatGPT TextArea. It will make it much easier for users.

- always send on Enter
- expand down in-sted of up
- new Line is shift enter

I would love if the ChatFeed (?) also expanded. An expanded down. Just like ChatGPT.

When coming back to panel chat components after a while I thought they where not working. Simply because I overlooked the TextInput at the bottom of the window. I have a Big trader screen. And its not natural that there is such a Big distance between ChatRow and text input.

ahuang11:
> always send on Enter

ChatGPT actually isn't always send on enter; if you enter the second line, it doesn't send

> expand down in-sted of up; I would love if the ChatFeed (?) also expanded. An expanded down. Just like ChatGPT.

I don't follow; I copied ChatGPT's textarea behavior here.

> Simply because I overlooked the TextInput at the bottom of the window. I have a Big trader screen. And its not natural that there is such a Big distance between ChatRow and text input.

Isn't that also ChatGPT's behavior; the text input all the way at the bottom?
image

philippjfr:
I think we should probably just extend bokeh's TextAreaInput as a Bokeh model so that the functionality of an autosizing text area can be made available to everyone and we don't ship yet another version of a slightly different widget.

ahuang11:
Okay, I tried doing that, but I couldn't figure out the error:
https://github.com/holoviz/panel/pull/5591/files

I might be missing something really simple...

PR #5582, opened 2023-10-04
Allow refs on Matplotlib and PaneBase

Hoxbro:
I can see the Gaminder example fails because they are False. 

https://github.com/holoviz/panel/actions/runs/6403251613/job/17381429742?pr=5572


philippjfr (approved):
Thanks, good catch!

PR #5584, opened 2023-10-04
Fix Select widget client side selection

### LINKED ISSUE #5583: Select widget of size>1 does not select correct value on client side
I'm on Panel 1.2.3. Trying to build an app using `pn.widgets.Select` with `size>1`. I can see that on the client side it does not select the correct value. Furthermore if I use it in tabs the selected value on the client side is reset when navigating between tabs.

## Reproduce

```python
import panel as pn

pn.extension()

select1 = pn.widgets.Select(value=2, options=[1,2], size=4, name="Select 1")
select2 = pn.widgets.Select(value=2, options=[1,2], size=4, name="Select 2")

pn.Tabs(
    select1, select2
).servable()
```

When I open the app it looks like

image

I would expect 2 to be selected.

If I select 2. Navigate to the second tab and back to the first tab. Then again option value 1 is selected again!

MarcSkovMadsen:
Closes #5583

### DIFF
--- a/panel/models/singleselect.ts
+++ b/panel/models/singleselect.ts
@@ -51,7 +51,7 @@ export class SingleSelectView extends InputWidgetView {
   render_selection(): void {
     const selected = this.model.value
 
-    for (const el of this.el.querySelectorAll('option'))
+    for (const el of this.input_el.querySelectorAll('option'))
       if (el.value === selected)
         el.selected = true
 


philippjfr:
Could we get a UI test for this?

panel/tests/ui/widgets/test_select.py
```diff
@@ -14,3 +14,10 @@ def test_select_with_size(page):
     page.locator('option').nth(1).click()
 
     wait_until(lambda: select.value == 'B')
+
+def test_select_option(page):
+    select = Select(value='B', options=['A', 'B', 'C'], size=4)
+
+    serve_component(page, select)
+
+    assert page.locator('select').evaluate("(sel)=>sel.value")=='B'
```
philippjfr, line 23:
```suggestion
    wait_until(lambda: page.locator('select').evaluate("(sel)=>sel.value") == 'B', page)
```

philippjfr:
All test failures are transient.

philippjfr (approved):
Thank you!

panel/tests/ui/widgets/test_select.py
```diff
@@ -20,4 +20,4 @@ def test_select_option(page):
 
     serve_component(page, select)
 
-    assert page.locator('select').evaluate("(sel)=>sel.value")=='B'
+    wait_until(lambda: page.locator('select').evaluate("(sel)=>sel.value") == 'B', page)
```
MarcSkovMadsen, line 23: Hmmm. I thought "tricks" like these where not nescessary with Playwright @philippjfr ?

philippjfr: If you use the expect API it isn't but I don't think that has a evaluate method.

Issue #5600, opened 2023-10-09
`TextEditor` widget doesn't render on FireFox


maximlt:
Very likely due to changes made in https://github.com/holoviz/panel/pull/5511. The error displayed in Firefox's console being:
```
Error rendering Bokeh items: TypeError: this.shadow_el.getSelection is not a function
    getNativeRange http://localhost:5007/static/extensions/panel/panel.min.js?v=fb7e691caef7a59ca25d1fd9580b7a7d13146fd012a16bf047f631f7bf8a1326:135
    getRange http://localhost:5007/static/extensions/panel/bundled/quillinput/1.3.7/quill.js:3040
    Toolbar http://localhost:5007/static/extensions/panel/bundled/quillinput/1.3.7/quill.js:9401
    emit http://localhost:5007/static/extensions/panel/bundled/quillinput/1.3.7/quill.js:8664
    emit http://localhost:5007/static/extensions/panel/bundled/quillinput/1.3.7/quill.js:1943
    optimize http://localhost:5007/static/extensions/panel/bundled/quillinput/1.3.7/quill.js:4381
    update http://localhost:5007/static/extensions/panel/bundled/quillinput/1.3.7/quill.js:7331
    update http://localhost:5007/static/extensions/panel/bundled/quillinput/1.3.7/quill.js:4403
    deleteAt http://localhost:5007/static/extensions/panel/bundled/quillinput/1.3.7/quill.js:7202
    deleteAt http://localhost:5007/static/extensions/panel/bundled/quillinput/1.3.7/quill.js:4266
    deleteText http://localhost:5007/static/extensions/panel/bundled/quillinput/1.3.7/quill.js:2556
    setContents http://localhost:5007/static/extensions/panel/bundled/quillinput/1.3.7/quill.js:1508
    modify http://localhost:5007/static/extensions/panel/bundled/quillinput/1.3.7/quill.js:1660
    setContents http://localhost:5007/static/extensions/panel/bundled/quillinput/1.3.7/quill.js:1505
    render http://localhost:5007/static/extensions/panel/panel.min.js?v=fb7e691caef7a59ca25d1fd9580b7a7d13146fd012a16bf047f631f7bf8a1326:135
    render http://localhost:5007/static/js/bokeh.min.js?v=c8d721b75531cff043feef626ae0d6e29a26031fae26bba93c1baa962ecfb1b43f211027df3a0cf7772f30afed8c38198d70677b8967e581ca7e605358c68465:547
[...]

Issue #5545, opened 2023-09-27
Pasting text in the `TextEditor` no longer works


maximlt:
Pasting text in the `TextEditor` no longer works. That's the error logged in the browser console when you try to paste in the widget:

```
quill.js:3065 Uncaught TypeError: Cannot read properties of null (reading 'offset')
    at quill.js:3065:26
    at Array.map (<anonymous>)
    at Selection.normalizedToRange (quill.js:3059:31)
    at Selection.getRange (quill.js:3042:24)
    at Selection.update (quill.js:3229:28)
    at HTMLDocument.<anonymous> (panel.min.js?v=c47f5b18b8a430e698b9fe15e51f6119984e78334bcf3f45e210d30c37ef2f9e:135:2379)
```

maximlt:
The `TextEditor` widget is implemented based on QuillJS, which hasn't been updated in a while and doesn't yet support the shadow DOM. A good summary of this can be found in:
- https://github.com/quilljs/quill/issues/2961
- https://github.com/robrez/quill/pull/1

After Panel moved to Bokeh 3, the `TextEditor` widget was patched twice to improve its support in the shadow DOM:
- https://github.com/holoviz/panel/pull/4467
- https://github.com/holoviz/panel/pull/5511

Both patches were inspired from https://stackoverflow.com/questions/67914657/quill-editor-inside-shadow-dom/67944380#67944380, the author of this answer can also be found on Quill's Github https://github.com/quilljs/quill/issues/2961#issuecomment-859955012.

The copy/paste issue has also been reported on Quill's issue tracker a few times:
- https://github.com/quilljs/quill/issues/2021#issuecomment-1479055139
- https://github.com/quilljs/quill/issues/2961#issuecomment-1195538461

I haven't yet seen a solution that fixes the issue, I haven't tried them all though and I could easily miss it 🙃 
---

I've tried a few things to fix the issue, without luck so far. It seems the issue doesn't occur on Firefox.

maximlt:
[...]

PR #5609, opened 2023-10-10
TextEditor: hacks to re-enable rendering on Firefox and copy/paste

maximlt:
Fixes https://github.com/holoviz/panel/issues/5545
Fixes https://github.com/holoviz/panel/issues/5600

"Fixes" is a bit exaggerated!

Basically I've noticed that the selection parsed by `normalizeNative` contained a node the code didn't expect so I tried to return `null` in that case and it worked (even if there's still an error in the console...).

On rendering in Firefox, browsers have different level of support for the Selection API and patching it for Firefox made things worse obviously, so I've disabled that and the widget can be rendered again and interacted with (there are still some weird behaviors).

Until Quill supports being rendered in the Shadow DOM, or we find a proper way to patch it, the TextEditor widget is going to behave strangely in some cases.

### DIFF
--- a/panel/models/quill.ts
+++ b/panel/models/quill.ts
@@ -13,6 +13,11 @@ const normalizeNative = (nativeRange: any) => {
 
     const range = nativeRange;
 
+    // // HACK: To allow pasting
+    if (range.baseNode?.classList?.value === 'ql-clipboard') {
+      return null
+    }
+
     if (range.baseNode) {
       range.startContainer = nativeRange.baseNode;
       range.endContainer = nativeRange.focusNode;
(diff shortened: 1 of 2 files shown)


panel/models/quill.ts
```diff
@@ -101,13 +106,17 @@ export class QuillInputView extends HTMLBoxView {
...
 
-    // Hack Quill and replace document.getSelection with shadow.getSelection
-    // see https://stackoverflow.com/questions/67914657/quill-editor-inside-shadow-dom/67944380#67944380
-    this.quill.selection.getNativeRange = () => {
-      const selection = (this.shadow_el as any).getSelection();
-      const range = normalizeNative(selection);
-      return range;
-    };
+    // Apply only with getSelection() is defined (e.g. undefined on Firefox)
+    if (typeof this.quill.root.getRootNode().getSelection !== 'undefined') {
```
Hoxbro, line 110: Shouldn't this be inside the function? Or is `getSelection` never defined in FireFox?

maximlt: Not sure about the correct language to use there, `getSelection` isn't "supported" in the shadow DOM on Firefox, see https://developer.mozilla.org/en-US/docs/Web/API/ShadowRoot#instance_methods.

I think that then when `document.getSelection()` is called on Firefox, it "pierces" through the shadow DOM. Of course, I have no idea what I'm talking about.

philippjfr: >I have no idea what I'm talking about.

Fake til you make it 😄 That certainly sounded convincing.

philippjfr (approved):
Until Quill itself fixes thing this seems like hacky but reasonable fix.

PR #5613, opened 2023-10-10
Update `TextAreaInput` rows on value change when auto_grow

ahuang11:
Watching `input` events only watches user key strokes, and does not watch `textarea.value = ""`

This PR watches `value` so that rows will also be updated when value is changed programmatically.

### DIFF
--- a/panel/models/textarea_input.ts
+++ b/panel/models/textarea_input.ts
@@ -9,8 +9,7 @@ export class TextAreaInputView extends BkTextAreaInputView {
 
     const { value, max_rows } = this.model.properties;
 
-    this.on_change(value, () => this.update_rows());  // for manually setting value
-    this.on_change(max_rows, () => this.update_rows());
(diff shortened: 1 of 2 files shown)


panel/models/textarea_input.ts
```diff
@@ -4,6 +4,15 @@ import * as p from "@bokehjs/core/properties";
 export class TextAreaInputView extends BkTextAreaInputView {
   model: TextAreaInput;
 
+  connect_signals(): void {
+    super.connect_signals();
+
+    const { value, max_rows } = this.model.properties;
+
+    this.on_change(value, () => this.update_rows());  // for manually setting value
+    this.on_change(max_rows, () => this.update_rows());
```
philippjfr, line 13:
```suggestion
    this.on_change([max_rows, value], () => this.update_rows());
```

philippjfr (approved):
Looks good!

PR #5627, opened 2023-10-12
Unify OAuth implementations and refresh access_token

philippjfr:
- [x] Deduplicate implementations of different auth providers
- [x] Adding support for automatically refreshing the `access_token`
- [x] Add tests
    - [x] Okta
    - [x] Azure
    - [x] Auth0


panel/tests/ui/test_auth.py
```diff
@@ -39,6 +39,100 @@ def test_basic_auth(py_file, page, prefix):
...
+
+        page.locator('input[name="username"]').fill(okta_user)
+        page.locator('input[name="password"]').fill(okta_password)
+        page.locator('input[type="submit"]').click(force=True)
+
+        expect(page.locator('.markdown')).to_have_text(okta_user, timeout=10000)
+
+
+@unix_only
+@pytest.mark.skipif('AZURE_OAUTH_KEY' not in os.environ, reason='Azure credentials not available')
```
Hoxbro, line 74: These checks seem common enough for people to actually have in their environment. 

Maybe have an environment variable called `PANEL_CI_TESTS` which is used instead.

philippjfr: I can prefix them with `PANEL_`

Hoxbro: That is also a way to do it :slightly_smiling_face:

panel/auth.py
```diff
@@ -1033,12 +861,26 @@ def __init__(
                 self._login_template = _env.from_string(f.read())
         self._login_endpoint = login_endpoint or '/login'
         self._logout_endpoint = logout_endpoint or '/logout'
+
+        state.on_session_destroyed(self._remove_user)
         super().__init__()
 
+    def _remove_user(self, session_context):
```
Hoxbro, line 868: Is there a `remove_user` property somewhere to match ho `get_user` works?

panel/auth.py
```diff
@@ -1078,6 +984,60 @@ def login_handler(self):
...
+            if user in state._oauth_user_overrides:
+                del state._oauth_user_overrides[user]
+
+    def _schedule_refresh(self, expiry_ts, user, refresh_token, application, request):
+        if not state._active_users.get(user):
+            return
+        now_ts = dt.datetime.now(dt.timezone.utc).timestamp()
+        expiry_seconds = expiry_ts - now_ts - 10
+        log.debug("%s scheduling token refresh in %d seconds", type(self).__name__, expiry_seconds)
+        expiry_date = dt.datetime.now() + dt.timedelta(seconds=expiry_seconds)
```
Hoxbro, line 1002: Shouldn't this also use `dt.timezone.utc`?

panel/auth.py
```diff
@@ -216,14 +249,24 @@ async def _fetch_access_token(self, code, redirect_uri, client_id, client_secret
...
         if not user:
-            return
-
-        log.debug("%s received user information.", type(self).__name__)
-        return self._on_auth(user, body['access_token'], body.get('refresh_token'))
+            log.debug("%s could not obtain user_info, falling back to decoding access_token.", type(self).__name__)
+            try:
+                user = decode_token(body['access_token'])
+            except Exception as e:
+                raise e
```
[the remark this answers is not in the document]
Hoxbro: (See lines below)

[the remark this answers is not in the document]
philippjfr: Indeed, already working on fixing this.

panel/auth.py
philippjfr: No, `schedule_task` works in the local timezone.

Hoxbro: This would be nice to have as a small comment, I think.

panel/auth.py
```diff
@@ -1069,6 +911,70 @@ class OAuthProvider(BasicAuthProvider):
...
+                log.debug("Fully authenticated and access_token still valid.")
+                return user
+
+            if refresh_token:
+                try:
+                    refresh_json = decode_token(refresh_token)
+                    if refresh_json['exp'] < now_ts:
+                        refresh_token = None
+                except (ValueError, UnicodeDecodeError):
+                    pass
```
Hoxbro, line 967: Shouldn't `refresh_token` be set to None here if something goes wrong? With a warning

philippjfr: No, I'll add a comment.  If the refresh_token is not a valid JWT token (which is what would trigger this error) then according to the spec it will not expire at all. I'll comment to that effect.

panel/tests/ui/test_auth.py
philippjfr: Honestly, I don't think it's very likely and what's the worst case? The test fails, that's honestly fine.

panel/io/server.py
```diff
@@ -1046,15 +1048,19 @@ def get_server(
...
       Additional information for the OAuth provider
     oauth_error_template: str (optional, default=None)
       Jinja2 template used when displaying authentication errors.
     cookie_secret: str (optional, default=None)
       A random secret string to sign cookies (required for OAuth)
-    oauth_encryption_key: str (optional, default=False)
+    oauth_encryption_key: str (optional, default=None)
       A random encryption key used for encrypting OAuth user
       information and access tokens.
+    oauth_refresh_tokens: str (optional, default=None)
```
cdeil, line 1062: Is `str` type correct here for `oauth_refresh_tokens` or should it be `bool`?

doc/how_to/server/commandline.md
```diff
@@ -109,6 +109,8 @@ options:
                         A random string used to encode the user information.
   --oauth-expiry-days OAUTH_EXPIRY_DAYS
                         Expiry off the OAuth cookie in number of days.
+  --oauth-token-refresh OAUTH_TOKEN_REFRESH
```
cdeil, line 112: @philippjfr  - I think this is the name of the option?

```
--oauth-refresh-tokens OAUTH_REFRESH_TOKENS
```

panel/auth.py
```diff
@@ -1078,6 +1000,65 @@ def login_handler(self):
...
+            return
+        now_ts = dt.datetime.now(dt.timezone.utc).timestamp()
+        expiry_seconds = expiry_ts - now_ts - 10
+        log.debug("%s scheduling token refresh in %d seconds", type(self).__name__, expiry_seconds)
+        expiry_date = dt.datetime.now() + dt.timedelta(seconds=expiry_seconds) # schedule_task is in local TZ
+        refresh_cb = partial(self._scheduled_refresh, user, refresh_token, application, request)
+        if expiry_seconds <= 0:
+            refresh_cb()
+            return
+        task = f'{user}-access-token-refresh'
```
cdeil, line 1023: I think here it doesn't matter, but for consistency with the other code you could choose `refresh-tokens` instead of `token-refresh` here.

Issue #5625, opened 2023-10-12
Move Chat Components to their own namespace


MarcSkovMadsen:
I think the chat components should be moved to their own namespace `pn.chat` for the reasons below.

- They are not all widgets. For example they don't all have a parameter `value`.
- To keep it simple. The `ChatInterface` can be used almost stand alone for users who does not know a lot about Param and Panel.
- To give it focus. The chat components would get their own section in the Component Gallery.
- To enable chat components to grow. I would expect more components to be added later. For example the `EnvironmentWidget` that makes it easy to use a combination of environment variables and user input for api keys etc.
- To make it easier to carve out. Maybe one day this is too big or specialized for Panel and should be in its own `panel-chat` package.

ahuang11:
> To enable chat components to grow. I would expect more components to be added later. For example the EnvironmentWidget

I don't think is specific to Chat though.

philippjfr:
+1 on moving them into their own namespace. These aren't widgets as Panel defines them even if they have a `value`. Widgets are inputs or controls, while these components are used for bi-directional communication and rendering of rich outputs.

MarcSkovMadsen:
For the `ChatEntry` might have a `.value` parameter. But the `value` is not something the user can input. Instead its actually displayed as an output.

philippjfr:
Yep, indeed. I do wonder if it should even subclass `CompositeWidget` but that's not too important.

ahuang11:
Will add this before release.

PR #5633, opened 2023-10-13
Move chat components into its own namespace

ahuang11:
Closes https://github.com/holoviz/panel/issues/5625

It's probably a good idea to move these chat components into its own namespace before release to avoid future deprecations and confusion.

I believe that's everything, but I may be forgetting something so let me know if you think of something that needs to be moved too.


ahuang11:
Oh I need to actually update the imports.

MarcSkovMadsen:
> Now that it's in its own namespace, would it make sense to drop the "Chat" prefix from ChatInterface, ChatFeed, ChatEntry, etc so that it's `pn.chat.Interface`, `pn.chat.Feed`, and `pn.chat.Entry` or is that too generic?

I believe that is too generic.

examples/reference/chat/ChatEntry.ipynb
```diff
@@ -9,7 +9,7 @@
     "import asyncio\n",
     "\n",
```
MarcSkovMadsen, line 10: I think you need to move the Chat components to their own chat components section in the Component Gallery too?

ahuang11: I'm a bit confused; these Jupyter notebooks are the component gallery items, right? Once the docs are built, it'll create its own section automatically?

panel/chat/langchain.py
```diff
@@ -9,7 +9,7 @@
     AgentFinish = None
```
MarcSkovMadsen, line 9: You also need to update the links to the reference guides in the docstrings. For example https://panel.holoviz.org/reference/widgets/ChatInterface.html

examples/reference/chat/ChatEntry.ipynb
MarcSkovMadsen: Ok. I did not know. I thought they where defined in some TocTree somewhere.

panel/chat/langchain.py
ahuang11: Thanks for catching that.

panel/widgets/chatbox.py
```diff
@@ -207,7 +207,7 @@ class ChatBox(CompositeWidget):
     The ChatBox widget displays a conversation between multiple users
     composed of users' icons, names, messages, and likes.
 
-    Reference: https://panel.holoviz.org/reference/widgets/ChatBox.html
+    Reference: https://panel.holoviz.org/reference/chat/ChatBox.html
```
ahuang11, line 210:
```suggestion
    Reference: https://panel.holoviz.org/reference/widgets/ChatBox.html
```

ahuang11:
Decided to separate each component into its own module since they're all pretty big.

examples/reference/chat/ChatEntry.ipynb
ahuang11: I'm not 100% sure either; leaving this unresolved for Philipp to see.

MarcSkovMadsen (approved):

examples/reference/chat/ChatEntry.ipynb
philippjfr: You need to add a new section in the `nbsite_gallery_conf` definition of `doc/conf.py`

ahuang11: Okay added

PR #5643, opened 2023-10-16
Add examples/conftest.py

Hoxbro:
This is just moving the `ignore_glob` into a file.  

But this will give the option to selectively ignore files based on the Python version, a package version or the OS instead of removing it from all.

### DIFF
--- a/examples/conftest.py
+++ b/examples/conftest.py
@@ -2,10 +2,8 @@ collect_ignore_glob = [
     "apps/",
     "developer_guide/",
     "homepage.ipynb",
-    "*VTK*",
-    "*vtk*",
+    "*VTK*.ipynb",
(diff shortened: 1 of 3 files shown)


examples/conftest.py
```diff
@@ -0,0 +1,11 @@
+collect_ignore_glob = [
+    "apps/",
+    "developer_guide/",
+    "homepage.ipynb",
+    "*VTK*",
+    "*vtk*",
+    "*Vega.ipynb",
+    "*DeckGL*",
+    "*deckgl*",
```
Hoxbro, line 9: I will make a test run without `*deckgl*` and `*vtk*` and see if this will run.

philippjfr (approved):
Test failures appear transient, merging.

Issue #5554, opened 2023-09-28
Patching fails with DataFrame that don't have a monotonic range


maximlt:
Patching a Tabulator widget with a DataFrame whose index isn't the usual monotonic integer 0, 1, 2, etc. isn't supported.

```python
import numpy as np
import pandas as pd
import panel as pn

pn.extension('tabulator')

df = pd.DataFrame(
    index=[10, 50, 100],
    data=dict(
        A=list(range(3)),
        B=list('abc'),
))
w = pn.widgets.Tabulator(df)
w.patch(df)
```

Traceback:
```
ValueError                                Traceback (most recent call last)
Cell In[10], line 1
----> 1 w.patch(df)

File ~/dev/panel/panel/widgets/tables.py:796, in BaseTable.patch(self, patch_value, as_index)
    792 if isinstance(patch_value, pd.DataFrame):
    793     patch_value_dict = {
    794         column: list(patch_value[column].items()) for column in patch_value.columns
    795     }
--> 796     self.patch(patch_value_dict, as_index=as_index)
    797 elif isinstance(patch_value, pd.Series):
    798     if "index" in patch_value:  # Series orient is row

File ~/dev/panel/panel/widgets/tables.py:816, in BaseTable.patch(self, patch_value, as_index)
    814             else:
    815                 self.value.iloc[ind, columns.index(k)] = value
--> 816     self._patch(patch_value)
    817 else:
    818     raise ValueError(
    819         f"Patching with a patch_value of type {type(patch_value).__name__} "
    820         "is not supported. Please provide a DataFrame, Series or Dict."
    821     )

File ~/dev/panel/panel/util/__init__.py:380, in updating.<locals>.wrapped(self, *args, **kwargs)
    378 self._updating = True
    379 try:
--> 380     fn(self, *args, **kwargs)
    381 finally:
    382     self._updating = updating

File ~/dev/panel/panel/widgets/tables.py:1476, in Tabulator._patch(self, patch)
   1474 if not patch:
   1475     return
-> 1476 super()._patch(patch)
   1477 self._update_style()
   1478 self._update_selectable()

File ~/dev/panel/panel/util/__init__.py:380, in updating.<locals>.wrapped(self, *args, **kwargs)
    378 self._updating = True
    379 try:
--> 380     fn(self, *args, **kwargs)
    381 finally:
    382     self._updating = updating

File ~/dev/panel/panel/reactive.py:954, in SyncableData._patch(self, patch)
[...]

PR #5645, opened 2023-10-16
Fix patching a table with a DataFrame with a custom index

maximlt:
Fixes https://github.com/holoviz/panel/issues/5554

### DIFF
--- a/examples/reference/widgets/DataFrame.ipynb
+++ b/examples/reference/widgets/DataFrame.ipynb
@@ -3,7 +3,9 @@
   {
    "cell_type": "code",
    "execution_count": null,
-   "metadata": {},
+   "metadata": {
+    "tags": []
+   },
    "outputs": [],
    "source": [
     "import pandas as pd\n",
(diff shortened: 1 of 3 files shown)


maximlt:
The examples tests are failing on the `reference/widgets/DataFrame.ipynb` notebook. `.patch` is exerted in this notebook in the following way:

```python
import pandas as pd
import panel as pn

pn.extension()

df = pd.DataFrame({'int': [1, 2, 3], 'float': [3.14, 6.28, 9.42], 'str': ['A', 'B', 'C'], 'bool': [True, False, True]}, index=[1, 2, 3])
table = pn.widgets.DataFrame(df)
table.patch({
    'bool': [
        (0, False),
        (2, False)
    ],
    'int': [
        (slice(0, 2), [3, 2])
    ]
})
```

Note that the index of the DataFrame is `[1, 2, 3]` so not the default DataFrame index.

The default of `.patch()` is `as_index=True` which means that the indexes should be treated as DataFrame indexes, not as simple integer indexes. I think the example in the notebook is leveraging a bug so I'll update it.

philippjfr (approved):
This looks great, thank you for adding the tests!

philippjfr (approved):

PR #5646, opened 2023-10-16
Add Python 3.12 and drop 3.8

Hoxbro:


panel/tests/test_util.py
```diff
@@ -84,8 +86,13 @@ def test_abbreviated_repr_list():
 
 
 def test_abbreviated_repr_ordereddict():
-    assert (abbreviated_repr(OrderedDict([('key', 'some really, really long string')]))
-            == "OrderedDict([('key', ...])")
+    if sys.version_info >= (3, 12):
+        expected = "OrderedDict({'key': 'some ...])"
```
Hoxbro, line 90: I applied this change without thinking... this does not look right 🙃

philippjfr: Appears to be passing though.

philippjfr (approved):

PR #5647, opened 2023-10-16
Add ColorMap widget

philippjfr:
Adds ColorMap widget which was recently added to Bokeh.

- [x] Implements https://github.com/holoviz/panel/issues/5154
- [x] Add docs
- [x] Add tests
- [x] Add reference gallery image

Screen Shot 2023-10-16 at 12 37 45

### DIFF
--- a/examples/reference/widgets/ColorMap.ipynb
+++ b/examples/reference/widgets/ColorMap.ipynb
@@ -1,0 +1,109 @@
+{
+ "cells": [
+  {
+   "cell_type": "code",
+   "execution_count": null,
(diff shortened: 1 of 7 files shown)


tupui:
Oh nice 👏 I actually needed something like that and went the route of generating a colormap from a single color. This is way simpler and complete as you really take advantage of existing colormaps. Still in case you are wondering: 

https://github.com/Simulation-Decomposition/simdec-python/blob/256afac0a8b9d03dc72398533274cd10bd8446f6/src/simdec/visualization.py#L37-L70

Issue #4236, opened 2022-12-24
pn.cache not working with async code: RuntimeError: cannot reuse already awaited coroutine


MarcSkovMadsen:
I'm trying to contribute a guide on Panel+Dask in #4234.

I assume its best to use Dask `asynchronously`, because then you can off load computation to the dask cluster while still keeping your application responsive.

Then I need a way to instantiate and share the async Dask client that work both in a notebook when people are experimenting and on the server. I assume its best to instantiate the client once and then share it across sessions.

In the notebook you can just do `client = await Client(asynchronously)` in a cell. But in a script running on the server you cannot use `await` outside of a function. If I do I get I get an error saying this is not possible.

So instead I think I can just use `pn.cache` to instantiate once and then share. But this does not work.

```python
import param

from dask.distributed import Client

import panel as pn

pn.extension()

@pn.cache
async def get_client():
    return await Client(asynchronous=True)

submit_button = pn.widgets.Button(name="Submit").servable()

@pn.depends(submit_button, watch=True)
async def run(_):
    print("submit")
    client = await get_client()
```

```bash
panel serve script.py
```

The second time you click the `Submit` button you get

```bash
$ panel serve script.py
2022-12-24 05:59:38,361 Starting Bokeh server version 2.4.3 (running on Tornado 6.1)
2022-12-24 05:59:38,362 User authentication hooks NOT provided (default user enabled)
2022-12-24 05:59:38,366 Bokeh app running at: http://localhost:5006/script
2022-12-24 05:59:38,366 Starting Bokeh server with process id: 2564
submit
submit
2022-12-24 05:59:47,087 - tornado.application - ERROR - Exception in callback functools.partial(<bound method IOLoop._discard_future_result of <tornado.platform.asyncio.AsyncIOMainLoop object at 0x7f5e6f6b0070>>, <Task finished name='Task-1415' coro=<async_execute.<locals>.wrapper() done, defined at /opt/conda/lib/python3.10/site-packages/panel/io/server.py:155> exception=RuntimeError('cannot reuse already awaited coroutine')>)
Traceback (most recent call last):
  File "/opt/conda/lib/python3.10/site-packages/tornado/ioloop.py", line 741, in _run_callback
    ret = callback()
  File "/opt/conda/lib/python3.10/site-packages/tornado/ioloop.py", line 765, in _discard_future_result
    future.result()
  File "/opt/conda/lib/python3.10/site-packages/panel/io/server.py", line 161, in wrapper
[...]

PR #5649, opened 2023-10-16
Add support for decorating async functions with pn.io.cache

philippjfr:
Fixes https://github.com/holoviz/panel/issues/4236

### DIFF
--- a/panel/io/cache.py
+++ b/panel/io/cache.py
@@ -20,7 +20,7 @@ from contextlib import contextmanager
 
 import param
 
-from param._utils import iscoroutinefunction
+from param.parameterized import iscoroutinefunction
 
 from .state import state
 
--- a/panel/io/cache.py
+++ b/panel/io/cache.py
@@ -20,6 +20,8 @@ from contextlib import contextmanager
 
 import param
 
+from param._utils import iscoroutinefunction
+
 from .state import state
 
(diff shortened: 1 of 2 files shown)


panel/io/cache.py
```diff
@@ -20,6 +20,8 @@
 
 import param
 
+from param._utils import iscoroutinefunction
```
maximlt, line 23: :)

Maybe the utils module of Param should be made public and its functions private, except those that you use in Panel/HoloViews? Doesn't have to happen now, just brainstorming.

philippjfr: I'm using this and a few other functions quite routinely so I'd be in favor. That said, I'll probably switch to `from param.parameterized import iscoroutinefunction` for now like I've done elsewhere in the codebase.

maximlt: You're certainly the one in the best place to decide where you want to import these from and how 👍 

I'm definitely in favor of being more explicit in Param about its public API and documenting whatever is public. I got bitten by this recently when refactoring `__init__.py`, it turns out some package (Panel maybe, I don't remember) relied on `_is_number` being a top level function :D

https://github.com/holoviz/param/blob/31f71c304f81ab14aec52d96f92653fbf6a8cf8e/param/__init__.py#L70

PR #5652, opened 2023-10-16
Fix disabling show_button_name with width >= 400

ahuang11:
image


panel/chat/interface.py
```diff
@@ -207,7 +207,8 @@ def _update_input_width(self):
         """
         Update the input width.
         """
-        self.show_button_name = self.width is None or self.width >= 400
+        if self.show_button_name is None:
+            self.show_button_name = self.width is None or self.width >= 400
```
philippjfr, line 211: Should this run `on_init` or does it already?

ahuang11: It's run under init. I think I tried using `on_init` but there was an issue so I dumped it in init
https://github.com/holoviz/panel/blob/357eb332e31b4b83270dd81d373ee23f96e828c1/panel/chat/interface.py#L185C10-L186C35

panel/tests/chat/test_interface.py
```diff
@@ -146,6 +146,11 @@ def test_show_button_name_width(self, chat_interface):
         chat_interface.width = 200
         assert not chat_interface.show_button_name
```
ahuang11, line 147:
```suggestion
        assert chat_interface.show_button_name
```

philippjfr (approved):

PR #5660, opened 2023-10-17
Move API removals from version 1.3 to 1.4

maximlt:
The CI is currently failing as `deprecated` raises an error if the version announced for the removal has a beta or RC release.

I checked some of these deprecations and believe they were set before Panel 1.1.1 was released, on the 21th of June. In my opinion 4 months is too tight, so I bumped it arbitrarily to 1.4. Feel free to discard this PR and resolve the issue in another manner.


philippjfr:
I'm fine with bumping them.

philippjfr (approved):

PR #5661, opened 2023-10-18
Make `_display_globals` a staticmethod

Hoxbro:
This will make it possible to call it in HoloViews without initializing `pn.extension`.

### DIFF
--- a/panel/config.py
+++ b/panel/config.py
@@ -834,8 +834,8 @@ class panel_extension(_pyviz_extension):
         if not published:
             self._display_globals()
 
-    @classmethod
-    def _display_globals(self):
+    @staticmethod
+    def _display_globals():
         if config.browser_info and state.browser_info:
             doc = Document()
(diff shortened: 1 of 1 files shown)


panel/config.py
```diff
@@ -834,6 +834,7 @@ def __call__(self, *args, **params):
         if not published:
             self._display_globals()
 
+    @classmethod
     def _display_globals(self):
```
maximlt, line 838: Shouldn't it be a `staticmethod` if it doesn't depend on `self` at all?

Hoxbro: I want to call it without initializing `pn.extension`, see https://github.com/holoviz/holoviews/pull/5947.

If I change to a `staticmethod` I get:
image

maximlt: Of course you need to remove `self` from the signature.

Hoxbro: This should be done.

philippjfr (approved):

PR #5667, opened 2023-10-18
Skip tests broken with latest playwright

maximlt:
Two date tests broke with the latest version of Playwright (1.39.0) on Chromium (working of Firefox!). They should probably be re-written as they're a little hacky anyway.

### DIFF
--- a/panel/tests/ui/widgets/test_tabulator.py
+++ b/panel/tests/ui/widgets/test_tabulator.py
@@ -567,6 +567,7 @@ def test_tabulator_editors_bokeh_select(page, df_mixed):
     expect(page.locator('text="option1"')).to_have_count(1)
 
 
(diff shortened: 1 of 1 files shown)


philippjfr:
Thanks, should we just pin playwright for now?

maximlt:
> Thanks, should we just pin playwright for now?

playwright is installed by the holoviz install action:

https://github.com/holoviz-dev/holoviz_tasks/blob/f2afba15f34b5b09b4f800ed747a38d31ec155c0/install/action.yaml#L167

We could either:

- pin it there, release, bump the version
- set `playwright: false` here and add a step here to install it with a pin

I went for the simplest approach in this PR. Pick which one you prefer :)

philippjfr:
I'll merge for now, maybe playwright will fix this in a followup release.

philippjfr (approved):

PR #5671, opened 2023-10-18
Add playwright-version

Hoxbro:


setup.py
```diff
@@ -1,7 +1,6 @@
 #!/usr/bin/env python
 import json
 import os
-import shutil
```
Hoxbro, line 4: Not used in `setup.py`, so used it to reset the cache :slightly_smiling_face:

PR #5675, opened 2023-10-18
Fix patching a table with `pd.NaT` values

maximlt:
I got this error while trying to patch a table with a `pd.NaT` value.

```python
        for name, patch in patches.items():
            for ind, value in patch:
                if isinstance(ind, (int, slice)):
>                   self[name][ind] = value
E                   TypeError: float() argument must be a string or a real number, not 'NaTType'

../../miniconda3/envs/panel310/lib/python3.10/site-packages/bokeh/core/property/wrappers.py:517: TypeError
```


philippjfr (approved):
Looks good, thanks!

Issue #5672, opened 2023-10-18
`FileDownload` does not align with other `Button`


tupui:
There is an alignment issue if in a `Row`:

```python
import panel as pn
pn.extension(template="material")
pn.Row(pn.widgets.FileDownload(), pn.widgets.Button(name='Test')).servable(area="header")
```

<img width="534" alt="Screenshot 2023-10-18 at 14 31 18" src="https://github.com/holoviz/panel/assets/23188539/39eb6b7d-3290-49ef-9514-4fe6ecf9a19a">

_Originally posted by @tupui in https://github.com/holoviz/panel/issues/1853#issuecomment-1768353271_

Besides the alignment issue, the defaults for font icon sizing seem to be different.

PR #5676, opened 2023-10-18
Small tweaks for Material template and theme

philippjfr:
Fixes https://github.com/holoviz/panel/issues/5672

### DIFF
--- a/panel/theme/css/material.css
+++ b/panel/theme/css/material.css
@@ -375,7 +375,8 @@ select[multiple].bk-input option:checked {
 
 /* Button widgets */
 
-.bk-btn-group > button {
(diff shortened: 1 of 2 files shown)


panel/theme/css/material.css
```diff
@@ -228,6 +228,10 @@ button.mdc-button.mdc-card-button {
   margin-top: 0.5rem;
```
ahuang11, line 228: Wondering if this is necessary? I also noticed it making the buttons not aligned.

image

ahuang11: If I remove the margin-top, it's aligned
image

tupui:
Awesome thanks 🙏

PR #5677, opened 2023-10-18
Fix issues with ChatInterface overlap and style

philippjfr:
Screenshot 2023-10-18 at 16 59 11

### DIFF
--- a/panel/dist/css/chat_interface.css
+++ b/panel/dist/css/chat_interface.css
@@ -1,5 +1,14 @@
 @import url('chat_feed.css');
 
+:host {
+  overflow-y: hidden;
+}
+
 .chat-feed-log {
-  max-height: 89%;
+  max-height: 90%;
+}
+
+.chat-interface-input-container {
+  position: sticky;
+  bottom: 0;
 }
--- a/panel/dist/css/chat_interface.css
+++ b/panel/dist/css/chat_interface.css
@@ -1,7 +1,7 @@
 @import url('chat_feed.css');
 
-:host {
-  max-height: none;
+.chat-feed-log {
+  max-height: 89vh;
 }
 
 .chat-interface {
--- a/panel/dist/css/chat_interface.css
+++ b/panel/dist/css/chat_interface.css
@@ -3,7 +3,3 @@
 .chat-feed-log {
   max-height: 89%;
 }
-
-.chat-interface {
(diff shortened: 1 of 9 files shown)


panel/dist/css/chat_interface.css
```diff
@@ -4,14 +4,6 @@
   max-height: none;
```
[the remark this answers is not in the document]
ahuang11: Or else, if the ChatMessages is super long, the chat inputs are all the way at the bottom.

ahuang11:
with 89vh (instead of 90vh because it shifts a bit with 90)

https://github.com/holoviz/panel/assets/15331990/627a7edb-86e9-483b-be59-e9ea25f148f5

ahuang11:
Actually, not sure how to fix all cases: PDF example
https://holoviz-topics.github.io/panel-chat-examples/langchain/#pdf-assistant

image

ahuang11:
I suspect it's because of the template header.

ahuang11:
Use of max-height instead of height and % instead of vh seems to work best.

MarcSkovMadsen:
If the issue you are trying to solve is for the input at the bottom, then try different cases like TextInput, TextAreaInput, FileInput and CodeEditor.

ahuang11:
Thanks for pointing that out; here's the latest changes.

image

panel/dist/css/chat_message.css
```diff
@@ -9,7 +9,6 @@
 
 :host {
   max-height: none;
-  max-width: 99%; /* Prevent horizontal scroll bar */
```
ahuang11, line 12: Slightly worried that this is important to prevent it from creating an ugly horizontal scroll bar.

philippjfr: Yeah text isn't the issue with the max_width on the text but I see what you mean. I just don't think this can be on `:host` since you apply this stylesheet to a bunch of things. It's fine for a max-width of 99% on the message itself.

ahuang11:
No more horizontal scrollbars and double vertical scrollbars

https://github.com/holoviz/panel/assets/15331990/152de5aa-b4a2-4c27-87f8-5823d08b4c05


https://github.com/holoviz/panel/assets/15331990/ae41a4ce-26c8-4a48-8a0d-47ffeb27b4d1

PR #5689, opened 2023-10-20
Make ChatBox deprecation very clear

MarcSkovMadsen:
The `ChatBox` is deprecated. This shows up after some time in the output cells when a user reads the `ChatBox` reference notebook.

I want to make it really clear before the user starts reading that the `ChatBox` is deprecated by adding an alert at the top.

image


philippjfr (approved):
Thanks!

Issue #5622, opened 2023-10-12
Chat Docs issues


MarcSkovMadsen:
As the new chat interface is in the `dev` docs I will start reviewing and add issues here

Dev Docs: https://holoviz-dev.github.io/panel/reference/index.html#widgets

## Issues

- [x] Spelling errors in Chat component gallery thumbnails. I've updated the ones in [post below](https://github.com/holoviz/panel/issues/5622#issuecomment-1760088855). Philipp should upload to CDN.
- [x] Component Gallery now has many categories without clear ordering. To be fixed by #5692
- [x] No mention of and link to panel-chat-examples.
  - [x] Should be in App Gallery https://holoviz-dev.github.io/panel/gallery/index.html. To be fixed by #5690 
  - [x] Should be in ChatInterface reference guide. Could be in ChatEntry and ChatFeed guides
- [x] Images missing from component gallery
- [x] Cannot run dev examples in Pyodide (We now can but with issues listed below)
- [x] #5625

### ChatMessage

- [x] Design specification image not found
  - [x] Remember to change ChatEntry to ChatMessage
- [x] Cannot import name `document` from `js` error when running Matplotlib in pyodide. See https://github.com/pyodide/pyodide/issues/561. To be worked around via #5702.
- [ ] "Reference ... isn't known" in console when running in Pyodide.
- [x] Beat boxing cannot play. To be fixed by #5688.
- [x] Both Streamlit and LangChain has the concept of message(s). For example `st.chat_message`. Consider if `ChatMessage` is a better name than `ChatEntry` @ahuang11 .

### ChatFeed

- [x] Design specification image not found
- [x] Verbose mentioned but not set. To be fixed by #5691 
- [x] Feed without entries is totally flat (Is not an issue, but a feature).
- [x] Await outside function

### ChatInterface

- [x] Design specification image not found
- [x] #5695. To be solved by #5701
- [x] #5708. To be solved by #5712
- [x] #5705. To be solved by #5706
- [x] #5709
- [x] #5716
- [x] `callback_user` not described in Parameters section.
- [x] ChatInterface Perspective examples not working. To be fixed by #5696.
- [x] The TextInput widget does not stretch when `show_button_name=False`. Fixed by #5697
- [x] show_button_name not working

### PanelCallbackHandler

- [x] Images not showing
- [x] #5707. Fixed by #5710
- [x] No reference guide
- [x] #5629
[...]

PR #5692, opened 2023-10-20
Order Component Gallery alphabetically

MarcSkovMadsen:
Addresses #5622 by ordering the Component Gallery categories alphabetically.

## Motivation

The Component Gallery has grown big and now contains many categories. 

image

The order of the categories is probably originally motivated by the order in which you should learn or use them ??

But now there are so many categories it takes up some of your muscle memory to remember where they are located. And to new users this order might not be clear at all.

With the addition of the `chat` category this issue has become very clear. Why are they located at the bottom? Where should they really be located?

I suggest solving this and preparing for more future categories by using alphabetical sort.

### DIFF
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -96,19 +96,21 @@ nbsite_gallery_conf = {
         'reference': {
             'title': 'Component Gallery',
             'sections': [
+                'panes',
+                'widgets',
+                'layouts',
+                # 3 most important by expected usage. Rest alphabetically
                 'chat',
                 'global',
                 'indicators',
-                'layouts',
-                'panes',
                 'templates',
-                'widgets',
             ],
             'titles': {
                 'Vega': 'Altair & Vega',
                 'DeckGL': 'PyDeck & Deck.gl',
                 'ECharts': 'PyEcharts & ECharts',
-                'IPyWidget': 'ipywidgets'
+                'IPyWidget': 'ipywidgets',
+                'PanelCallbackHandler': 'LangChain CallbackHandler',
             },
             'as_pyodide': True,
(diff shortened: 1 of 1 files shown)


philippjfr:
I'm on the fence about this having chat, global and indicators as the first three items doesn't feel quite right.

MarcSkovMadsen:
I can sort of follow you, but what is the order and why is chat at bottom?

Is it possible to have Layout, Pane, Widgets as the 3 first because they are most important. Then signal change by having divider and the add rest alphabetically?

philippjfr:
>Is it possible to have Layout, Pane, Widgets as the 3 first because they are most important. Then signal change by having divider and the add rest alphabetically?

I'd favor that but don't see a way to achieve it without changing nbsite since the index is auto-generated.

philippjfr:
Chat is last merely because @ahuang11 added it there. Let's just do what you suggest without the separator.

MarcSkovMadsen:
Ok. I changed the sort order as discussed.

I've also tried to sneak in a more userfriendly title for the `PanelCallbackHandler`

## Before

image


## After

image

philippjfr (approved):
Thank you!

doc/conf.py
```diff
@@ -97,18 +97,20 @@
             'title': 'Component Gallery',
             'sections': [
                 'panes',
```
MarcSkovMadsen, line 99: I propose this order because I believe Panes and widgets are most used. The Panes on top because they can quickly give you a visual impression of what Panel supports if you are a new user. Layouts at bottom because I believe they are lesser used?

PR #5696, opened 2023-10-20
Fix ChatInterface Perspective examples

MarcSkovMadsen:
As reported in #5622 the Perspective examples do not work in the ChatInterface reference notebook on the `dev` site. The output cell just displays as blank.

image

I believe the issue is that the `"perspective"` extension is not loaded at the top of the notebook.

I've manually changed the code in the docs to test out the hypothesis and then 1) clicked the run pyodide button 2) tested the example. It works 👍 

image


philippjfr (approved):
Thanks!

PR #5702, opened 2023-10-20
Workaround ChatMessage Matplotlib Pyodide issue

MarcSkovMadsen:
Addresses #5622

The `ChatMessage` reference notebook does not work in Pyodide because Matplotlib cannot run in Pyodide with service workers. See https://github.com/pyodide/pyodide/issues/561.

I work around the issue by replacing Matplotlib with Vega.

The result looks like

image

### DIFF
--- a/examples/reference/chat/ChatMessage.ipynb
+++ b/examples/reference/chat/ChatMessage.ipynb
@@ -11,7 +11,7 @@
     "\n",
     "from panel.chat import ChatMessage\n",
     "\n",
-    "pn.extension()"
+    "pn.extension(\"vega\")"
    ]
   },
   {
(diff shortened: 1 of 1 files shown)


examples/reference/chat/ChatMessage.ipynb
```diff
@@ -85,19 +85,27 @@
...
+    "}\n",
+    "vgl_pane = pn.pane.Vega(vegalite, height=240)\n",
     "\n",
     "pn.Column(\n",
     "    ChatMessage(pn.widgets.Button(name=\"Click\")),\n",
     "    ChatMessage(df),\n",
-    "    ChatMessage(fig),\n",
-    ")"
+    "    ChatMessage(vgl_pane),\n",
+    ").servable()"
```
ahuang11, line 108: Do we need servable here?

examples/reference/chat/ChatMessage.ipynb
```diff
@@ -85,19 +85,27 @@
    "outputs": [],
    "source": [
     "import pandas as pd\n",
```
ahuang11, line 87: Should we put imports at the top?

MarcSkovMadsen: I've just moved it up.

ahuang11 (approved):

PR #5706, opened 2023-10-21
Fix chat interface background

### LINKED ISSUE #5705: ChatInterface taskbar not working with Fast dark theme
I'm on the current `main` branch of Panel working on contributing a Pandas Agent to panel-chat-examples

image

(To be minimized later)

```python
"""
Demonstrates how to use the `ChatInterface` and `PanelCallbackHandler` to create a
chatbot to talk to your Pandas DataFrame.
"""
# Reference: https://github.com/langchain-ai/streamlit-agent/blob/main/streamlit_agent/chat_pandas_df.py
from langchain.agents import AgentType
from langchain.agents import create_pandas_dataframe_agent
from langchain.chat_models import ChatOpenAI
import streamlit as st
import pandas as pd
import os
import panel as pn
from panel_chat_examples import EnvironmentWidgetBase
import param

pn.extension("perspective")

file_formats = {
    "csv": pd.read_csv,
    "xls": pd.read_excel,
    "xlsx": pd.read_excel,
    "xlsm": pd.read_excel,
    "xlsb": pd.read_excel,
}


class Environment(EnvironmentWidgetBase):
    OPENAI_API_KEY = param.String()


class AppState(param.Parameterized):
    environ = param.ClassSelector(class_=Environment, constant=True)
    llm = param.Parameter(constant=True)
    data = param.DataFrame()
    pandas_df_agent = param.Parameter(constant=True)

    def __init__(self, environ: Environment|None=None):
        if not environ:
            environ=Environment()
        super().__init__(environ=environ)

    @param.depends("environ.OPENAI_API_KEY", on_init=True, watch=True)
    def _update_llm(self):
        with param.edit_constant(self):
            if self.environ.OPENAI_API_KEY:
                self.llm = ChatOpenAI(
                    temperature=0,
                    model="gpt-3.5-turbo-0613",
                    api_key=self.environ.OPENAI_API_KEY,
                    streaming=True,
                )
            else:
                self.llm = None

    @param.depends("llm", "data", on_init=True, watch=True)
    def _update_pandas_df_agent(self):
        with param.edit_constant(self):
            if self.llm and self.data:
                self.pandas_df_agent = create_pandas_dataframe_agent(
                    self.llm,
                    self.data,
                    verbose=True,
                    agent_type=AgentType.OPENAI_FUNCTIONS,
                    handle_parsing_errors=True,
                )
            else:
                self.pandas_df_agent = None

    @property
    def error_message(self):
        if not self.llm and self.data is None:
            return "Please provide your `OPENAI_API_KEY` and upload some data."
        if not self.llm:
            return "Please provide your `OPENAI_API_KEY`."
        if self.data is None:
            return "Please upload some data."
        return ""

    @property
    def welcome_message(self):
        return ("I'm your Pandas Agent." + " " + state.error_message).strip()

    async def callback(self, contents, user, instance):
        if self.error_message:
            return self.error_message
        
        pn_cb = pn.chat.langchain.PanelCallbackHandler(instance=instance)
        response = state.pandas_df_agent.run(st.session_state.messages, callbacks=[pn_cb])
        print(response)
        


@pn.cache(ttl=2 * 60 * 60)
def load_data(file, file_name):
    try:
        ext = os.path.splitext(file_name)[1][1:].lower()
    except:
        ext = file_name.split(".")[-1]
    if ext in file_formats:
        return file_formats[ext](file)
    else:
        st.error(f"Unsupported file format: {ext}")
        return None


state = AppState()

    widgets=[pn.widgets.FileInput(name="Upload"), pn.widgets.TextInput(name="Message")],
    callback=state.callback)
chat_interface.send(state.welcome_message, user="Pandas Agent", avatar="🐼", respond=False)



template = pn.template.FastListTemplate(
    title="🦜 LangChain: Chat with pandas DataFrame",
    main=[chat_interface],
)

if state.environ.variables_not_set:
    template.sidebar.append(state.environ)

template.servable()
```

MarcSkovMadsen:
Closes #5705

I've not been able to get Panel to serve this.

But if I change this manually via the browser console in a working app. It works for the below example

```python
import panel as pn

component = pn.chat.ChatInterface(
    widgets=[pn.widgets.FileInput(name="Upload"), pn.widgets.TextInput(name="Message")]
)
pn.template.FastListTemplate(main=[component]).servable()
```

I can manually verify it works for this example

image

### DIFF
--- a/panel/dist/css/chat_interface.css
+++ b/panel/dist/css/chat_interface.css
@@ -5,5 +5,5 @@
 }
 
 .chat-interface-input-tabs {
-  background: white;
+  background: var(--neutral-fill-rest, var(--background-color, white));
 }
--- a/panel/dist/css/chat_interface.css
+++ b/panel/dist/css/chat_interface.css
@@ -5,5 +5,5 @@
 }
 
 .chat-interface-input-tabs {
-  background: var(--neutral-fill-rest, var(--background-color, white));
+  background: var(--fill-color, var(--background-color, white));
 }


ahuang11 (approved):
I think you may need to just clear the cache

MarcSkovMadsen:
> I think you may need to just clear the cache

I have tried :-)

ahuang11:
I had to re-run
`SETUPTOOLS_ENABLE_FEATURES=legacy-editable pip install -e .` in panel repo

image

panel/dist/css/chat_interface.css
```diff
@@ -5,5 +5,5 @@
 }
 
 .chat-interface-input-tabs {
-  background: white;
+  background: var(--fill-color, var(--background-color, white));
```
ahuang11, line 8:
```suggestion
  background: var(--background-color, white);
```

MarcSkovMadsen: The `--background-color` is static and managed by Panel. It refers to some background?

image

The `--fill-color` is dynamic and managed by the *fast design* framework. It refers to the background of the wrapping element, currently a card. But if I use `fast` elements as wrappers in custom `ReactiveHTML` components. That might be different.

image

So the `--fill-color` and `--background-color` are not the same in theory or practice.

MarcSkovMadsen: Thanks for the suggestion. But believe the right thing is to keep my suggestion.

ahuang11 (approved):

panel/dist/css/chat_interface.css
philippjfr: Specific design related variables should really never leak into the generic CSS definitions. Unfortunately this is currently not implemented very well, so I'll have to see if there's an alternative.

MarcSkovMadsen: You can use @ahuang11 suggestion right now if that is better. Then keep the issue open and we can find solution for next release.

panel/theme/css/fast_variables.css
```diff
@@ -1,5 +1,6 @@
 :host {
+  --background-color: var(--fill_color);
   --bokeh-base-font: Open Sans, Roboto, sans-serif;
   color: var(--neutral-foreground-rest);
-  background-color: var(--fill-color);
+  background-color: var(--background-color);
```
MarcSkovMadsen, line 5: This change I would say is not 100% correct. The `--fill-color` is the correct one to use for fast.

MarcSkovMadsen: Other wise you should define `--background-color: var(--fill-color)`?

philippjfr: I did. Look above.

PR #5710, opened 2023-10-21
Revert PanelCallbackHandler change

### LINKED ISSUE #5707: PanelCallbackHandler design principle
yesterday I changed the PanelCallBackHandler handler to also output the final response from the agent.

image

The change is marked below.

image

I can see that Streamlits Callback handler does not write the 'output'. The user/ developer have to manually do this. 

image

I'm thinking that maybe it was a mistake by me to change this. What do you think @ahuang11 ?

MarcSkovMadsen:
Closes #5707 by letting the PanelCallbackHandler focus on outputting the intermediate output/ *chain of thought*. Not the final response.


panel/chat/langchain.py
```diff
@@ -68,12 +69,13 @@ def _update_active(self, avatar: str, label: str):
             self._active_user = f"{self._active_user} - {label}"
 
     def _stream(self, message: str):
-        return self.instance.stream(
-            message,
-            user=self._active_user,
-            avatar=self._active_avatar,
-            message=self._message,
-        )
+        if message:
```
MarcSkovMadsen, line 72: I added the `if message` because the Pandas Agent llm streams an empty token which leads to an empty buble.

panel/chat/langchain.py
```diff
@@ -117,6 +119,7 @@ def on_tool_start(
         self, serialized: Dict[str, Any], input_str: str, *args, **kwargs
     ):
         self._update_active(DEFAULT_AVATARS["tool"], serialized["name"])
+        self._stream(input_str)
```
MarcSkovMadsen, line 122: I added `self._stream(input_str)` because its really useful for debugging/ learning to learn what input the tool takes.

For example for the Pandas Agent.

image

ahuang11: To clarify to the user what it is, maybe it should be prefixed and wrapped in ``?

panel/tests/chat/test_langchain.py
```diff
@@ -30,7 +30,7 @@ async def callback(contents, user, instance):
         tools, llm, agent=AgentType.ZERO_SHOT_REACT_DESCRIPTION, callbacks=[callback_handler]
     )
     instance.send("2 + 2")
-    assert len(instance.objects) == 4
+    assert len(instance.objects) == 3
     assert instance.objects[1].object == "Action: Python REPL\nAction Input: print(2 + 2)"
     assert instance.objects[2].object == "Final Answer: 4"
-    assert instance.objects[3].object == "4"
+    assert instance.disabled==False
```
ahuang11, line 36:
```suggestion
    assert not instance.disabled
```

panel/chat/langchain.py
MarcSkovMadsen: Maybe. But I don't know if it always makes sense. All I know about input_str is that it is a string?

Would it make sense to do for all tools?

ahuang11: Okay maybe let's hold off on ``, but perhaps: `f"Tool input: {input_str}"`?

ahuang11, line 122:
```suggestion
        self._stream(f"Tool input: {input_str}")
```

panel/chat/langchain.py
ahuang11, line 72: Maybe?
```suggestion
        if message.strip():
```

panel/chat/langchain.py
MarcSkovMadsen: To be consistent then you should do something similar with output `f"Tool output: {...}"`?

Personally I don't think its needed currently. But I believe that when we have learned more about agents and callback usecases we will do a second iteration of the PanelCallbackHandler. Also comparing it systematically to what Streamlit, and Chainlit does. Probably also implementing a second callback handler - for example for HayStack - to understand how a generative AI callback handler should work in general.

MarcSkovMadsen: Should I add it @ahuang11 ?

MarcSkovMadsen: Ok. I comitted your suggestion below.

MarcSkovMadsen:
Thanks for the feedback @ahuang11 . I've applied your suggestions. Let me know if there is more I should do. Thanks.

panel/chat/langchain.py
```diff
@@ -68,12 +69,13 @@ def _update_active(self, avatar: str, label: str):
...
-            avatar=self._active_avatar,
-            message=self._message,
-        )
+        if message.strip():
+            return self.instance.stream(
+                message,
+                user=self._active_user,
+                avatar=self._active_avatar,
+                message=self._message,
+            )
```
ahuang11, line 79: Should this return self._message?

MarcSkovMadsen: I would not know. All I can say is that in the different examples I have tried this works.

MarcSkovMadsen: Please propose what you see fit or resolve the conversation. Thanks.

ahuang11: I created https://github.com/holoviz/panel/pull/5718/

MarcSkovMadsen: Thx

ahuang11 (approved):

examples/reference/chat/PanelCallbackHandler.ipynb
```diff
@@ -129,7 +129,8 @@
     "\n",
     "async def callback(contents, user, instance):\n",
     "    callback_handler = pn.chat.langchain.PanelCallbackHandler(instance)\n",
-    "    await agent.arun(contents, callbacks=[callback_handler])\n",
+    "    response = await agent.arun(contents, callbacks=[callback_handler])\n",
```
ahuang11, line 132: Can you confirm this works for you?

I crash into a NotImplemented error when using async with ddg.

MarcSkovMadsen: Could you be more specific? How can I reproduce this?

ahuang11: Try running this example and see if it successfully runs.

philippjfr: I can reproduce.

philippjfr: Seems like an issue with the DuckDuckGo tool rather than with our code. Just need to change the example to use `agent.run`

ahuang11: Fixed.

examples/reference/chat/PanelCallbackHandler.ipynb
```diff
@@ -127,9 +127,9 @@
     "    tools, llm, agent=AgentType.ZERO_SHOT_REACT_DESCRIPTION, verbose=True, \n",
     ")\n",
     "\n",
-    "async def callback(contents, user, instance):\n",
+    "def callback(contents, user, instance):\n",
     "    callback_handler = pn.chat.langchain.PanelCallbackHandler(instance)\n",
-    "    await agent.arun(contents, callbacks=[callback_handler])\n",
+    "    agent.run(contents, callbacks=[callback_handler])\n",
```
ahuang11, line 132:
```suggestion
    "    return agent.run(contents, callbacks=[callback_handler])\n",
```

Issue #5708, opened 2023-10-21
ChatInterface callback cannot return ChatMessage


MarcSkovMadsen:
I'm on the `main` branch of Panel trying to create a LangChain Pandas DataFrame agent. I'm seeing some issues with the `ChatInterface`.

- [ ] `callback` with `ChatMessage` `return` value is not working.
- [ ] `callback` with `ChatMessage` `yield` value raises exception.

MarcSkovMadsen:
## Async `callback` with `return` value is not working.

```python
import panel as pn
import pandas as pd

agent_user="Pandas Agent"
agent_avatar="🐼"

pn.extension("perspective")

def _create_message(message):
    return pn.chat.ChatMessage("I'm your LangChain Pandas Agent!", user="Pandas Agent", avatar="🐼")

async def callback(self, contents, user, instance):
    if isinstance(contents, pd.DataFrame):
        instance.active=1
        return _create_message("You uploaded a dataframe. Ask me anything about it")

    return _create_message("That is a great question. I don't know.")

chat_interface = pn.chat.ChatInterface(
    widgets=[pn.widgets.FileInput(name="Upload"), pn.widgets.TextInput(name="Message")],
    renderers=pn.pane.Perspective,
    
    callback_exception="verbose",
)
    
chat_interface.send("I'm your LangChain Pandas Agent!", user="Pandas Agent", avatar="🐼", respond=False)
chat_interface.servable()
```

After upload of the file I would expect the `ChatInterface` to add the message "You uploaded a dataframe. Ask me anything about it". But it does not.

![image](https://github.com/holoviz/panel/assets/42288570/8b97d57b-6369-48d2-aef4-d0c04c01dde5)

You can use the file [penguins (1).csv](https://github.com/holoviz/panel/files/13060163/penguins.1.csv) for testing

MarcSkovMadsen:
## Async callback with yield value raises exception.

If I replace `return` with `yield` in the callback I get an exception instead.

```python
import panel as pn
import pandas as pd

agent_user="Pandas Agent"
agent_avatar="🐼"

pn.extension("perspective")

def _create_message(message):
    return pn.chat.ChatMessage("I'm your LangChain Pandas Agent!", user="Pandas Agent", avatar="🐼")

async def callback(contents, user, instance):
    if isinstance(contents, pd.DataFrame):
        instance.active=1
        print("I was here")
        yield _create_message("You uploaded a dataframe. Ask me anything about it")
        

[...]

PR #5712, opened 2023-10-21
Fix returning/yielding chat message from callback

ahuang11:
Closes https://github.com/holoviz/panel/issues/5708

### DIFF
--- a/panel/chat/feed.py
+++ b/panel/chat/feed.py
@@ -338,6 +338,12 @@ class ChatFeed(ListPanel):
 
         if message is not None:
             # ChatMessage is already created; updating existing ChatMessage
+            if isinstance(value, ChatMessage):
+                # Cannot set user or avatar when explicitly sending
+                # a ChatMessage; need to set them directly on the ChatMessage.
+                user = value.user
+                avatar = value.avatar
+                value = value.object
             message.update(value, user=user, avatar=avatar)
             return message
         elif isinstance(value, ChatMessage):
--- a/panel/chat/feed.py
+++ b/panel/chat/feed.py
@@ -335,12 +335,17 @@ class ChatFeed(ListPanel):
         if isinstance(value, dict):
             user = value.get("user", user)
             avatar = value.get("avatar")
+
         if message is not None:
+            # ChatMessage is already created; updating existing ChatMessage
             message.update(value, user=user, avatar=avatar)
             return message
         elif isinstance(value, ChatMessage):
+            # ChatMessage is not created yet, but a ChatMessage is passed; use it
+            self._replace_placeholder(value)
             return value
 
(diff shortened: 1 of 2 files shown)


panel/chat/feed.py
```diff
@@ -335,12 +335,17 @@ def _upsert_message(
         if isinstance(value, dict):
             user = value.get("user", user)
             avatar = value.get("avatar")
+
         if message is not None:
```
MarcSkovMadsen, line 339: Could there be a situation where `message is not None and isinstance(value, ChatMessage)`? And how should that be handled?

ahuang11: ChatMessage overrides it I think in

        elif isinstance(value, ChatMessage):
            # ChatMessage is not created yet, but a ChatMessage is passed; use it
            self._replace_placeholder(value)
            return value

MarcSkovMadsen:
I've tested that `async` and `return` works.

But `async` and `yield` does not seem to work.

```python
import panel as pn
import pandas as pd

agent_user="Pandas Agent"
agent_avatar="🐼"

pn.extension("perspective")

def _create_message(message):
    return pn.chat.ChatMessage(message, user="Pandas Agent", avatar="🐼")

async def callback(contents, user, instance):
    if isinstance(contents, pd.DataFrame):
        instance.active=1
        print("I was here")
        yield _create_message("You uploaded a dataframe. Ask me anything about it")
        

    yield _create_message("That is a great question. I don't know.")

    widgets=[pn.widgets.FileInput(name="Upload"), pn.widgets.TextInput(name="Message")],
    renderers=pn.pane.Perspective,
    callback=callback,
    callback_exception="verbose",
)
    
chat_interface.send("I'm your LangChain Pandas Agent!", user="Pandas Agent", avatar="🐼", respond=False)
chat_interface.servable()
```

image

panel/chat/feed.py
MarcSkovMadsen: I can't see that. Where do you see that?

image

ahuang11: :) Ok, perhaps not ChatMessage, but dict works as expected.

ahuang11:
Should be fixed now.

panel/chat/feed.py
MarcSkovMadsen: Should you add something like?

```python
...
elif isinstance(value, ChatMessage):
    user = value.user
    avatar = value.avatar
    value = value.value
```

MarcSkovMadsen:
> Should be fixed now.

I can confirm its fixed.Thanks.

panel/chat/feed.py
ahuang11: No, we don't need to extract the values if it's a ChatMessage and message is None.

It does look neater though.

philippjfr (approved):

Issue #5704, opened 2023-10-20
Add on_chat_model_start to PanelCallbackHandler


ahuang11:
I was using `ChatOpenAI` and I think it mentioned that the method was not implemented

on_chat_model_start(serialized, messages, *, ...) | Run when a chat model starts running.
-- | --

[on_chat_model_start](https://api.python.langchain.com/en/latest/callbacks/langchain.callbacks.base.BaseCallbackHandler.html#langchain.callbacks.base.BaseCallbackHandler.on_chat_model_start)(serialized, messages, *, ...)

Run when a chat model starts running.

PR #5714, opened 2023-10-22
Add on_chat_model_start

ahuang11:
Closes https://github.com/holoviz/panel/issues/5704

### DIFF
--- a/panel/chat/langchain.py
+++ b/panel/chat/langchain.py
@@ -172,6 +172,6 @@ class PanelCallbackHandler(BaseCallbackHandler):
         """
         Does not do anything here; the other methods will handle output,
         primarily on_llm_new_token and on_llm_end. Note, this method
-        explicitly does not call super() to prevent crashing into
+        explicitly does nothing and does not call super() to prevent crashing into
         a NotImplementedError from the inherited class.
         """
--- a/panel/chat/langchain.py
+++ b/panel/chat/langchain.py
@@ -2,7 +2,9 @@
 
 from __future__ import annotations
 
-from typing import Any, Dict, List, Union
+from typing import (
+    Any, Dict, List, Union,
+)
 
 try:
     from langchain.callbacks.base import BaseCallbackHandler
--- a/panel/chat/langchain.py
+++ b/panel/chat/langchain.py
@@ -170,8 +170,8 @@ class PanelCallbackHandler(BaseCallbackHandler):
         **kwargs: Any
     ) -> None:
         """
-        Does not do anything here; the other methods will handle output,
-        primarily on_llm_new_token and on_llm_end. Note, this method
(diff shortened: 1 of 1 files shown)


panel/chat/langchain.py
```diff
@@ -162,3 +162,16 @@ def on_retriever_end(self, documents, **kwargs: Any) -> Any:
     def on_text(self, text: str, **kwargs: Any):
         """Run when text is received."""
         return super().on_text(text, **kwargs)
+
+    def on_chat_model_start(
```
MarcSkovMadsen, line 166: What is the purpose of adding this method? As far as I can see it does not do anything?

ahuang11: See the docstring.

panel/chat/langchain.py
```diff
@@ -162,3 +162,16 @@ def on_retriever_end(self, documents, **kwargs: Any) -> Any:
...
+        self,
+        serialized: Dict[str, Any],
+        messages: List,
+        **kwargs: Any
+    ) -> None:
+        """
+        Does not do anything here; the other methods will handle output,
+        primarily on_llm_new_token and on_llm_end. Note, this method
+        explicitly does not call super() to prevent crashing into
```
ahuang11, line 175:
```suggestion
        explicitly does nothing and does not call super() to prevent crashing into
```

panel/chat/langchain.py
ahuang11: Updated the docstring to be clearer.

panel/chat/langchain.py
```diff
@@ -162,3 +162,16 @@ def on_retriever_end(self, documents, **kwargs: Any) -> Any:
...
+        self,
+        serialized: Dict[str, Any],
+        messages: List,
+        **kwargs: Any
+    ) -> None:
+        """
+        Does not do anything here; the other methods will handle output,
+        primarily on_llm_new_token and on_llm_end. Note, this method
+        explicitly does nothing and does not call super() to prevent crashing into
+        a NotImplementedError from the inherited class.
```
ahuang11, line 176:
```suggestion
        Does not do anything to prevent the inherited class from raising
        NotImplementedError, and thus will not call super() here;
        the other methods will handle output, primarily
        `on_llm_new_token` and `on_llm_end`.
```

panel/chat/langchain.py
```diff
@@ -2,7 +2,7 @@
 
 from __future__ import annotations
 
-from typing import Any, Dict, Union
+from typing import Any, Dict, List, Union
```
ahuang11, line 5:
```suggestion
from typing import (
    Any, Dict, List, Union,
)
```

MarcSkovMadsen (approved):

philippjfr (approved):

Issue #5694, opened 2023-10-20
The ChatInterface should clear any widget after "send"


MarcSkovMadsen:
Currently the `ChatInterface` clears the `TextInput.value` when it is "sent". This makes it easy to type/ input the next value.

I don't see why this is only done for the `TextInput`? Should also apply to other widgets like `TextAreaInput`, `CodeEditor` and `FileInput`. But currently it does not as you can see from the screenshot below from the reference notebook. This makes it more cumbersome to use the `ChatInterface`.

![image](https://github.com/holoviz/panel/assets/42288570/6656a3b8-bd57-42e3-9df6-e1f42c84b37d)

ahuang11:
There's `reset_on_send` param. I think I did this so that `FileInput` doesn't clear to access it's value, but perhaps with the [PDF assistant](https://github.com/holoviz-topics/panel-chat-examples/blob/main/docs/examples/langchain/langchain_pdf_assistant.py) refactored, it's no longer necessary.

PR #5715, opened 2023-10-22
Default to reset_on_send=True for ChatInterface

ahuang11:
Closes https://github.com/holoviz/panel/issues/5694

### DIFF
--- a/panel/chat/interface.py
+++ b/panel/chat/interface.py
@@ -319,7 +319,6 @@ class ChatInterface(ChatFeed):
                     file_name=active_widget.filename,
                 )
             # don't use isinstance here; TextAreaInput subclasses TextInput
-            print(self.reset_on_send)
             if type(active_widget) is TextInput or self.reset_on_send:
                 updates = {"value": ""}
                 if hasattr(active_widget, "value_input"):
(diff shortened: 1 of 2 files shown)


panel/chat/interface.py
```diff
@@ -80,7 +80,7 @@ class ChatInterface(ChatFeed):
     )
 
     reset_on_send = param.Boolean(
-        default=False,
+        default=True,
```
MarcSkovMadsen, line 83: I believe the reference guide should be updated too then.

ahuang11: I think the reference guide doesn't mention the default.

panel/chat/interface.py
```diff
@@ -319,6 +319,7 @@ def _click_send(self, _: param.parameterized.Event | None = None) -> None:
                     file_name=active_widget.filename,
                 )
             # don't use isinstance here; TextAreaInput subclasses TextInput
+            print(self.reset_on_send)
```
MarcSkovMadsen, line 322: Should be removed.

philippjfr, line 322:
```suggestion
```

philippjfr (approved):

Issue #5727, opened 2023-10-23
PanelCallbackHandler example return response before LLM Callback Value


MarcSkovMadsen:
I'm on the current `main` branch of Panel.

I would expect the two answers below to be in the opposite order.

![image](https://github.com/holoviz/panel/assets/42288570/cf867d39-2d0e-4552-868f-ba3c26327f0d)

```python
import panel as pn
from langchain.llms import OpenAI

pn.extension()

llm = OpenAI(temperature=0)

def callback(contents, user, instance):
    callback_handler = pn.chat.langchain.PanelCallbackHandler(instance)    
    return llm.predict(contents, callbacks=[callback_handler])

pn.chat.ChatInterface(callback=callback).servable()
```

MarcSkovMadsen:
## Correct Order

The correct order should be as below. Which the streaming version produces.

ahuang11:
Related to this line, but not sure how to fix.
https://github.com/holoviz/panel/blob/main/panel/chat/langchain.py#L103-L110

ahuang11:
Okay nevermind, found a fix https://github.com/holoviz/panel/pull/5729

PR #5728, opened 2023-10-23
Fix callback examples

MarcSkovMadsen:
Depends on issue #5727 being fixed via #5729

The callbacks in the `PanelCallbackHandler` examples should `return` the response. This PR does that.


ahuang11:
Wondering why would it be necessary to return if the output is streamed already?

MarcSkovMadsen:
> Wondering why would it be necessary to return if the output is streamed already?

It is not not necessary.

What I am trying to achieve is educational. You need to be able to see what the `callback` does and what the `return` value does.

An example without a return value would indicate that you could replace `return` value with `callback` in general. And that you cannot and should not.

PR #5729, opened 2023-10-23
Fix ordering of Langchain output and ChatModel

ahuang11:
Don't exactly understand why send didn't work, but this fixes it:

Closes https://github.com/holoviz/panel/issues/5727

image

### DIFF
--- a/panel/chat/langchain.py
+++ b/panel/chat/langchain.py
@@ -104,7 +104,6 @@ class PanelCallbackHandler(BaseCallbackHandler):
 
     def on_llm_end(self, response: LLMResult, *args, **kwargs):
         if not self._is_streaming:
-            print("HELLO")
             # on_llm_new_token does not get called if not streaming
             self._stream(response.generations[0][0].text)
 
--- a/panel/chat/langchain.py
+++ b/panel/chat/langchain.py
@@ -75,7 +75,7 @@ class PanelCallbackHandler(BaseCallbackHandler):
         self._active_avatar = self._input_avatar
         self._message = None
 
-    def _on_start(self, kwargs, serialized):
+    def _on_start(self, serialized, kwargs):
         model = kwargs.get("invocation_params", {}).get("model_name", "")
         self._is_streaming = serialized.get("kwargs", {}).get("streaming")
         messages = self.instance.objects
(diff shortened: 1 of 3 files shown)


ahuang11:
Actually there's now a duplication issue:
image

ahuang11:
Ok all fixed 
image

panel/chat/langchain.py
```diff
@@ -75,8 +75,17 @@ def _reset_active(self):
...
+        model = kwargs.get("invocation_params", {}).get("model_name", "")
+        self._is_streaming = serialized.get("kwargs", {}).get("streaming")
+        messages = self.instance.objects
+        if messages[-1].user != self._active_user:
+            self._message = None
+        if self._active_user and model not in self._active_user:
+            self._active_user = f"{self._active_user} ({model})"
+
     def _stream(self, message: str):
-        if message.strip():
```
ahuang11, line 79: If token == " ", it will not output the space which leads to this:
image

panel/chat/langchain.py
```diff
@@ -101,13 +104,9 @@ def on_llm_new_token(self, token: str, **kwargs) -> None:
 
     def on_llm_end(self, response: LLMResult, *args, **kwargs):
         if not self._is_streaming:
+            print("HELLO")
```
ahuang11, line 107:
```suggestion
```

philippjfr:
Looks good to me, can we catch regressions here earlier?

ahuang11:
Yeah, I can add a test.

ahuang11:
Actually maybe not, because it's dependent on a Chat model class, which requires OpenAI key.

philippjfr:
You can't mock it?

ahuang11:
I can try, I'm not very familiar with the Langchain codebase, and whether they have built-in mocks that I can use/copy. Looking here https://github.com/langchain-ai/langchain/blob/8c150ad7f6e9073f3508187ce79ea4715b74e105/libs/langchain/langchain/chat_models/openai.py#L342

ahuang11:
Okay added two tests to cover these changes.

panel/chat/langchain.py
```diff
@@ -178,8 +176,7 @@ def on_chat_model_start(
         **kwargs: Any
     ) -> None:
         """
-        Does not do anything to prevent the inherited class from raising
-        NotImplementedError, and thus will not call super() here;
-        the other methods will handle output, primarily
-        `on_llm_new_token` and `on_llm_end`.
+        To prevent the inherited class from raising
+        NotImplementedError, and thus will not call super() here.
```
ahuang11, line 180:
```suggestion
        NotImplementedError, will not call super() here.
```

MarcSkovMadsen (approved):
I've looked at the code and also tested it on all the examples in the notebooks.

Looks fine to me now.

PR #5734, opened 2023-10-24
fix refreshed access_token access

### LINKED ISSUE #5733: `state.access_token` ignores new tokens acquired from refresh
(this library, plus any other relevant software, e.g. bokeh, python, notebook, OS, browser, etc)

Version 1.2.3

Expected behavior: 

When a user connects with an expired oauth access token, a new access token is should be acquired using the refresh token and
made available via `state.access_token`.

Observed behavior:

When a user connects with an expired oauth access token, a new access token is acquired using the refresh token. However,
when attempting to access this access token via `state.access_token`, `None` is returned, as the code is only looking at
the original expired access token, not the new one that was refreshed.

#### 

polivbr:
fixes #5733

### DIFF
--- a/panel/io/state.py
+++ b/panel/io/state.py
@@ -904,7 +904,7 @@ class _state(param.Parameterized):
         """
         Returns the OAuth access_token if enabled.
         """
-        if self.user in self._oauth_user_overrides and 'refresh_token' in self._oauth_user_overrides:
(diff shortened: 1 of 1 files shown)


philippjfr:
Thanks for catching this so quickly!

philippjfr:
Seems like it isn't setting the OAuth secrets as part of tests. I'll merge this and then see if that's just due to this being a PR from an external contributor. Thanks again @polivbr!

philippjfr (approved):

PR #5735, opened 2023-10-25
Apply ruff format

philippjfr:
Just testing this for now.


panel/_param.py
```diff
@@ -64,18 +60,13 @@ def _validate_value(self, val, allow_None):
...
+            raise ValueError(f"Margin parameter {self.name!r} only takes integer and " f"tuple values, not values of not {type(val)!r}.")
 
     def _validate_length(self, val):
         if not isinstance(val, tuple) or len(val) in (2, 4):
             return
         raise ValueError(
-            f'Margin parameter {self.name!r} only takes integer and '
-            f'tuple values of length 2 (vertical, horizontal) or 4 '
-            '(top, right, bottom, left).'
+            f"Margin parameter {self.name!r} only takes integer and " f"tuple values of length 2 (vertical, horizontal) or 4 " "(top, right, bottom, left)."
```
philippjfr, line 69: Hmm, this really does not make me trust the formatter.

Hoxbro: It looks weird, but isn't it still a valid code?

I think this happens because of the Panel's `line-length` of  160 (?)

philippjfr: Yes valid, but I would expect the formatter to correctly join and then wrap the strings.

MarcSkovMadsen: For me `black` creates issues like these and `pylint` identifies them. Them I fix and don't see them again.

panel/chat/feed.py
```diff
@@ -120,81 +120,135 @@ class ChatFeed(ListPanel):
...
+        default=False,
+        readonly=True,
+        doc="""
+        Whether the Card should be expandable and collapsible.""",
+    )
+
+    disabled = param.Boolean(
+        default=False,
+        doc="""
+       Whether the feed is disabled.""",
```
ahuang11, line 175: One space off.

philippjfr:
Massively outdated at this point.

PR #5743, opened 2023-10-25
Add OAuth Guest for selected endpoints

tupui:
This PR adds the possibility to set `PANEL_OAUTH_GUEST_ENDPOINTS` (or `--oauth-guest-endpoints`) to register endpoints where we skip the OAuth workflow (if not authenticated) and return a guest user.

Once we have a guest, we can use `authorize_callback` and for instance detect that we have a guest and show the `oauth-template`. I am using this to present a landing page before OAuth.

See some background discussion in https://discord.com/channels/1075331058024861767/1166025760138600478


panel/io/state.py
tupui, line 1: I am not sure how to properly register a guest besides returning `"guest"` in `auth.py`, hence here `id_token` would be None. This works but there might be a better way.

philippjfr: Honestly, I think returning None here is fine, there is no `id_token` since the auth flow hasn't happened. We should however make sure that `pn.state.user` returns 'guest'.

philippjfr: >If we return None, then we don't it would be None for authorize_callback too as it gets state.user_info, self.request.path. Is this wanted?

Ah, that's a good point!

>To update pn.state.user I edited the cookie, not sure if this is the proper way.

I don't think we necessarily want that as that will persist and potentially authenticate the user on pages that shouldn't be accessible to a guest.

panel/auth.py
```diff
@@ -879,10 +881,18 @@ def _remove_user(self, session_context):
...
+        return True if uri in self._guest_endpoints else False
+
     @property
     def get_user(self):
         def get_user(request_handler):
             user = request_handler.get_secure_cookie("user", max_age_days=config.oauth_expiry)
+
+            if user is None and self._is_guest(request_handler.request.uri):
+                user = "guest".encode('utf-8')
+                state.cookies["user"] = "guest"
```
philippjfr, line 894: Ah I see now. I'm not 100% sure this is safe.

panel/io/state.py
tupui: Ok, so what about instead of returning None by default in both `state.user_info, state.user` we return guest info?

philippjfr: Sounds reasonable but wonder if that might have backwards compatibility implications for developers that used `pn.state.user` as a proxy for whether auth is enabled in their app or not.

tupui:
Not sure what is happening but now I get this error locally:

```python
2023-10-26 13:57:34,675 Uncaught exception GET /app/ws (::1)
HTTPServerRequest(protocol='http', host='localhost:5006', method='GET', uri='/app/ws', version='HTTP/1.1', remote_ip='::1')
  File "/opt/homebrew/Caskroom/mambaforge/base/envs/panel/lib/python3.10/site-packages/tornado/websocket.py", line 942, in _accept_connection
    open_result = handler.open(*handler.open_args, **handler.open_kwargs)
  File "/opt/homebrew/Caskroom/mambaforge/base/envs/panel/lib/python3.10/site-packages/tornado/web.py", line 3205, in wrapper
    self.redirect(url)
  File "/opt/homebrew/Caskroom/mambaforge/base/envs/panel/lib/python3.10/site-packages/tornado/websocket.py", line 614, in _raise_not_supported_for_websockets
    raise RuntimeError("Method not supported for Web Sockets")
RuntimeError: Method not supported for Web Sockets
```

philippjfr:
I'll try to play around.

philippjfr:
@tupui I made a bunch of changes and added some docs. If you get a chance please review.

doc/how_to/authentication/guest_users.md
tupui, line 1: This makes a lot of sense 👍 I initially was thinking about using the Guest endpoint (what I added) feature as to have a login page. But using instead what you added with `--oauth-optional` sounds like a better idea.

Can we somehow link that to `basic-login-template`??

panel/io/state.py
```diff
@@ -1093,7 +1097,12 @@ def user_info(self) -> Dict[str, Any] | None:
         """
         Returns the OAuth user information if enabled.
         """
+        is_guest = self.cookies.get('is_guest')
+        if is_guest:
+            return {"user": "guest"}
```
tupui, line 1102: I had also `username` as this is what the `authorize_callback` template needs. Otherwise it's None.

doc/how_to/authentication/guest_users.md
```diff
@@ -0,0 +1,43 @@
...
+```python
+import panel as pn
+
+pn.extension(template='material')
+
+pn.state.template.title = 'Optional Auth'
+
+if pn.state.user == 'guest':
+    button = pn.widgets.Button(name='Login').servable(target='header')
+    button.js_on_click(code='window.location.href="/login"')
```
tupui, line 24: I am not sure I understand how that works. What if I want to hide everything, shall I just put the `template.servable()` in the `else` (my attempt did not work. Also during my tries I got this error: `2023-10-27 15:24:49,470 DocumentLifeCycleHandler on_session_destroyed callback <bound method OAuthProvider._remove_user of <panel.auth.OAuthProvider object at 0x168aaf220>> failed with following error: 'NoneType' object has no attribute 'decode'
`)? Or do I need to check the user on every pane?

tupui (commented):
Thanks for working on that @philippjfr 🚀 I see I had a few missing bits 😅

doc/how_to/authentication/guest_users.md
philippjfr: That should work fine.

```python
def authorize(user_info, path):
    if user_info['user'] == 'guest' and path == '/admin':
        return '/login'
    return True
```

doc/how_to/authentication/guest_users.md
philippjfr: That should work fine and does when I try:

```python
import panel as pn

pn.extension(template='material')

pn.state.template.title = 'Optional Auth'


pn.Column(f'# Hello {pn.state.user}!', pn.state.user_info).servable()

if pn.state.user == 'guest':
    button = pn.widgets.Button(name='Login').servable(target='header')
    button.js_on_click(code='window.location.href="/login"')
else:
    pn.Column('My super secret content that I can only see if I am logged in.').servable()
```

Screenshot 2023-10-30 at 12 05 17

philippjfr (approved):

tupui:
Thanks @philippjfr 🚀

doc/how_to/authentication/guest_users.md
tupui: I might be doing something wrong on my side, but I still cannot make it work.

```python
print(pn.state.user)
if pn.state.user == 'guest':
    prompt = pn.pane.Markdown("Login to access App")

    login_button = pn.widgets.Button(name="Login", width=100)
    login_button.js_on_click(code="""window.location.href = './login'""")

    template.main[:, :] = pn.Column(prompt, login_button)

    template.servable()
else:
    template.header.append(...)
    template.main[:, :] = ...
    template.servable()
    ...
```

Here `print(pn.state.user)` starts with None, then goes to "guest" and after I click the button and authenticate with Google, I am back on the same page and it again prints "guest". Did I miss something on the OAuth setup part?

Also when a connection is closed, I get the following:

`2023-10-31 21:41:12,873 DocumentLifeCycleHandler on_session_destroyed callback <bound method OAuthProvider._remove_user of <panel.auth.OAuthProvider object at 0x17cdebd10>> failed with following error: 'NoneType' object has no attribute 'decode'
`

philippjfr: Is there any chance you still have an older development version sticking around?

tupui: I thought that and re-created an env from scratch but got the same results. My env and PATH looks normal too.

philippjfr: Thanks, will look into it.

philippjfr: Do you have a more complete example? I just tried to reproduce again and couldn't.

tupui: I will try to make a more minimal example tomorrow.

My app is here otherwise: https://github.com/Simulation-Decomposition/simdec-python/blob/main/panel/app.py

There is a `make serve-oauth` that I modified like so:

```
	PANEL_OAUTH_SCOPE=email panel serve panel/app.py \
		--show \
		--cookie-secret panel_cookie_secret_oauth \
		--oauth-optional \
		--basic-login-template panel/login.html \
		--logout-template panel/logout.html \
		--oauth-provider google \
		--static-dirs _static=docs/_static \
		--reuse-sessions --warm
```

And on the app itself I was trying to add the if/else as above.

tupui: I confirm that even your example does not work on my side. I am on a fresh mamba env with 1.3.1. The user remains guest.

philippjfr: Sorry I haven't checked back in here, could you file an issue so this doesn't fall through the cracks?

tupui: Sure no problem, done 😃

PR #5747, opened 2023-10-26
Return `None` user info if no `id_token`

maximlt:
To fix an error I've seen in a test suite after upgrading to Panel 1.3.0

```
[2023-10-26T10:12:07.721Z] /usr/local/lib/python3.10/site-packages/panel/util/__init__.py:282: in decode_token
[2023-10-26T10:12:07.721Z]     if signed and "." in token:
[2023-10-26T10:12:07.721Z] E   TypeError: argument of type 'NoneType' is not iterable
```


philippjfr (approved):
Thanks!

Issue #5567, opened 2023-10-02
Add download conversation / upload conversation


ahuang11:
I don't want to start the conversation from scratch so ideally, I can save a history of chat messages and re-upload it.

Or perhaps an easy way for users to add custom buttons to ChatInterface without dropping to ChatFeed, or even templates.

ahuang11:
Prior to this, I believe there should be an export / method method on ChatFeed that supports two formats:

1. panel kwargs {"user": ..., "value": ..., "avatar": ...}
2. openai / langchain style {"role": "system", "content": "You are a helpful bot."}

For use like this
```
from transformers import AutoTokenizer

tokenizer = AutoTokenizer.from_pretrained("mistralai/Mistral-7B-v0.1")
chat = [
  {"role": "system", "content": "You are a helpful bot."},
  {"role": "user", "content": "Hello, how are you?"},
  {"role": "assistant", "content": "I'm doing great. How can I help you today?"},
  {"role": "user", "content": "I'd like to show off how chat templating works!"},
]

print(tokenizer.apply_chat_template(chat, tokenize=False))
```

ahuang11:
cc: @MarcSkovMadsen for feedback

PR #5764, opened 2023-10-27
Add serialize methods for ChatMessage + ChatFeed

ahuang11:
I'm trying to make ChatInterface easy to use to fine tune LLMs. The first step is easy exporting of it, ready to use for transformers.

Kind of addresses https://github.com/holoviz/panel/issues/5567, but specifically for use with fine tuning transformers https://huggingface.co/docs/transformers/main/chat_templating, not ChatInterface.

I was initially thinking of combining these it like in the ChatBox export method https://github.com/holoviz/panel/pull/5178, but decided against it since the logic got long.

There's a lot of ways to serialize the widgets / panes / etc, so I added `format_func` as an escape hatch if users don't agree with the default.

Would like some early thoughts before I add tests.

image

```python
import panel as pn
import numpy as np
import pandas as pd

pn.extension()

msg = pn.chat.ChatMessage(
    pn.Row(
        pn.widgets.FloatSlider(value=10),
        pn.widgets.Checkbox(name="Check me!"),
        pn.pane.HTML("Hello World!"),
        pn.pane.DataFrame(
            pd.DataFrame(np.random.randn(10, 3), columns=list("ABC"))
        ),
    ),
)
print(str(msg))
```

- [x] Add tests
- [x] Use `user` from ChatInterface in `user_names`

### DIFF
--- a/panel/chat/feed.py
+++ b/panel/chat/feed.py
@@ -591,7 +591,7 @@ class ChatFeed(ListPanel):
         self._chat_log.clear()
         return cleared_entries
 
-    def serialize_for_transformers(
+    def _serialize_for_transformers(
         self,
         role_names: Dict[str, str | List[str]] | None = None,
         default_role: str | None = "assistant",
@@ -599,26 +599,6 @@ class ChatFeed(ListPanel):
     ) -> List[Dict[str, Any]]:
         """
         Exports the chat log for use with transformers.
-
-        Arguments
-        ---------
-        role_names : dict(str, str | list(str)) | None
-            A dictionary mapping the role to the ChatMessage's user name.
-            Defaults to `{"user": ["user"], "assistant": [self.callback_user]}`
-            if not set. The keys and values are case insensitive as the strings
-            will all be lowercased. The values can be a string or a list of strings,
-            e.g. `{"user": "user", "assistant": ["executor", "langchain"]}`.
-        default_role : str
-            The default role to use if the user name is not found in role_names.
-            If this is set to None, raises a ValueError if the user name is not found.
-        custom_serializer : callable
-            A custom function to format the ChatMessage's object. The function must
-            accept one positional argument and return a string. If not provided,
-            uses the serialize method on ChatMessage.
-
-        Returns
-        -------
-        A list of dictionaries with a role and content keys.
         """
         if role_names is None:
             role_names = {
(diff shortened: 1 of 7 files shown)


panel/chat/interface.py
```diff
@@ -409,3 +411,39 @@ def active(self, index: int) -> None:
...
+        Returns
+        -------
+        A list of dictionaries with a role and content keys.
+        """
+        if role_names is None:
+            role_names = {
+                "user": [self.user],
+                "assistant": [self.callback_user],
+            }
+        super().serialize_for_transformers(role_names, default_role, format_func)
```
ahuang11, line 449:
```suggestion
        return super().serialize_for_transformers(role_names, default_role, format_func)
```

ahuang11:
Just tried this out and it works well.
```python
import panel as pn
from transformers import AutoTokenizer
from ctransformers import AutoConfig, AutoModelForCausalLM, Config

pn.extension(design="material")

BASE_MODEL = "mistralai/Mistral-7B-Instruct-v0.1"
QUANTIZED_MODEL = "TheBloke/Mistral-7B-Instruct-v0.1-GGUF"
QUANTIZED_FILE = "mistral-7b-instruct-v0.1.Q4_K_M.gguf"


    conversation = instance.serialize_for_transformers()
    formatted_conversation = tokenizer.apply_chat_template(conversation, tokenize=False)
    response = llm(formatted_conversation, stream=True)
    message = ""
    for token in response:
        message += token
        yield message

config = AutoConfig(
    config=Config(
        temperature=0.5, max_new_tokens=2048, context_length=2048, gpu_layers=1
    ),
)
llm = AutoModelForCausalLM.from_pretrained(
    QUANTIZED_MODEL,
    model_file=QUANTIZED_FILE,
    config=config,
)
tokenizer = AutoTokenizer.from_pretrained(BASE_MODEL)

    callback=callback,
    show_send=False,
    show_rerun=False,
    show_undo=False,
    show_clear=False,
    show_button_name=False,
    callback_exception="verbose"
)
pn.template.FastListTemplate(
    main=[chat_interface],
    sidebar_width=400,
).show()
```

panel/chat/feed.py
```diff
@@ -591,6 +591,98 @@ def clear(self) -> List[Any]:
...
+            - transformers
+            role_names : dict(str, str | list(str)) | None
+                A dictionary mapping the role to the ChatMessage's user name.
+                Defaults to `{"user": ["user"], "assistant": [self.callback_user]}`
+                if not set. The keys and values are case insensitive as the strings
+                will all be lowercased. The values can be a string or a list of strings,
+                e.g. `{"user": "user", "assistant": ["executor", "langchain"]}`.
+            default_role : str
+                The default role to use if the user name is not found in role_names.
+                If this is set to None, raises a ValueError if the user name is not found.
```
philippjfr, line 674:
```suggestion
        **serialize_kwargs
            Additional keyword arguments to use for the specified format.

            - format="transformers"
              role_names : dict(str, str | list(str)) | None
                  A dictionary mapping the role to the ChatMessage's user name.
                  Defaults to `{"user": ["user"], "assistant": [self.callback_user]}`
                  if not set. The keys and values are case insensitive as the strings
                  will all be lowercased. The values can be a string or a list of strings,
                  e.g. `{"user": "user", "assistant": ["executor", "langchain"]}`.
              default_role : str
                  The default role to use if the user name is not found in role_names.
                  If this is set to None, raises a ValueError if the user name is not found.
```

philippjfr (approved):

Issue #5768, opened 2023-10-28
Add Panel Playground


MarcSkovMadsen:
Make it really easy for users to share Panel Apps via Pyodide.

- I've tried at awesome-panel.org/sharing. But it would need improvements
  - Github auth is optional. So I had to implement my own auth which is not perfect.
  - The user has to manually trigger `panel convert`
    - This is a slow process compared to for example stlite or gradio-lite
    - I can only convert apps for which I have the packages installed on the server
- Stlite, Gradio, Shiny Live all provide this

Some complication is that 

- We use jinja templates which are hard to support in pyodide
- We load extensions like Tabulator on demand which also needs to be handled.

I could do this on my own. But I think it would be stronger if Panel provided this.

MarcSkovMadsen:
This is a nice example of a Sympy playground https://teachyourselfsystems.com/playground?.

Issue #5506, opened 2023-09-13
ValueError when embedding Panel Server App in Iframe


peterroelants:
#### Description of expected behavior and the observed behavior

The [Location regex `"^$|[\/].*$"`](https://github.com/holoviz/panel/blob/ff43d36e14ff84238e0450bd56c7541ca5a0147a/panel/io/location.py#L41) fails when injecting a bokeh app in an iframe using `iframe.srcdoc`

I'm getting the following error in my Python logs:
```python-traceback
ValueError: String parameter 'pathname' value 'srcdoc' does not match regex '^$|[\\/].*$'.
```
However, the panel app seems to be rendered ok.

#### Complete, minimal, self-contained example code that reproduces the issue

- Webpage with iframe to hold panel/bokeh app containing the following:
    ```html
  <!DOCTYPE html>
  <html>
      <head>
          <title>Test iframe</title>
      </head>
      <body>
          <p>This is a web app.</p>
          <iframe id="myIframe" frameborder="10" marginwidth="10" marginheight="0" scrolling="NO" width="100%" height="500px" sandbox="allow-scripts"></iframe>
          <script>
              async function getSrc(iframe, url) {
                  const resp = await fetch(url, {
                      method: 'GET',
                      headers: {
                      'Accept': 'application/json',
                      'ContentType': 'application/json',
                      }
  
                  });
                  console.log("resp", resp)
                  const app_html = await resp.text();
                  console.log("app_html", app_html)
                  iframe.srcdoc = app_html
                  }
  
              const load = () => {
                  var iframe = document.getElementById("myIframe");
                  var url = "http://localhost:8000/bokeh-app";
                  getSrc(iframe, url)
              }
              load()
          </script>
          <p>This is the footer of the web app.</p>
      </body>
  </html>
    ```
- [FastAPI Server wrapping around Panel Server](https://panel.holoviz.org/how_to/integrations/FastAPI.html) in backend:
  ```python
  import panel as pn
  from bokeh.embed import server_document
  from fastapi import FastAPI, Request
  from fastapi.templating import Jinja2Templates
  
  app = FastAPI()
  templates = Jinja2Templates(directory="templates")
  
  
  @app.get("/bokeh-app")
  async def bkapp_page(request: Request):
      script = server_document(
          "http://127.0.0.1:5000/app",
      )
      return templates.TemplateResponse(
          "base.html", {"request": request, "script": script}
      )
  
  def createApp():
[...]

PR #5769, opened 2023-10-28
POC of Panel Playground

MarcSkovMadsen:
Proof of concept for #5768 to aid the discussion.

This code shows that its doable to build a simple Panel Playground using Panel and Pyodide.

Is this something for Panel to build and manage or should I build on my own @philippjfr ?

**THE APPS ON THE RIGHT ARE RUN IN PYODIDE. THEY RUN SMOOTHLY, FAST. AND THE LIVE RELOAD IS VERY FAST.**

https://github.com/holoviz/panel/assets/42288570/62703adb-e633-44f8-9cd7-9d51660e6256

image

THERE ARE PROBABLY LOTS OF ISSUES TO SOLVE FOR A REAL APPLICATION THAT CAN SERVE USERS

## Requirements

- [ ] Standalone js or webcomponents that can easily be embedded on website
  - [ ] web component to easily display panel app also inside `ReactiveHTML` and with *hot reload*.
    - [ ] Custom `<panel-lite>` like `<gradio-lite>` or
    - [ ] `<py-script>` if below was solved
      - [ ] https://github.com/pyscript/pyscript/issues/1820
      - [ ] https://github.com/pyscript/pyscript/discussions/1822
  - [ ] web component playground to easily edit and show Panel apps with hot reload of packages and code
    - [ ] Custom `<panel-lite-editor>` like the Gradio Lite Playground or
    - [ ] General `<py-script-editor>` as requested here https://github.com/pyscript/pyscript/discussions/1823

## Roadmap

- [ ] Figure out if we should build on Panel or make a standalone playground in JS?
- [ ] Figure out if I can use panel convert to generate the *template* and embed that inside an iframe
  - [x] https://github.com/holoviz/panel/issues/5506
- [ ] Figure out if this can run safely outside iframe (as now) or should run inside iframe
- [ ] Add support for `async` javascript to the `ReactiveHTML` `_script` such that this can run properly using async/ await syntax.
- [ ] Carve our `Pyodide` pane and make it generally available. We really want two versions. One that can display Panel components. Another that can run Python code in a safe sandbox environment and return the result - for example code generate generated by llms.
- [ ] Support optional packages (via requirements.txt tab)
- [ ] Support optional template (via index.html tab)
- [ ] Support extension that the user provides in `pn.extension`
- [ ] Support design that the user provides in `pn.extension`
- [ ] Support writing to multiple `target`s
- [ ] Support export/ download
- [ ] Maybe use service workers.
- [ ] Wrap this in a nice template with "splitters" or similar inspired by for example Gradio Playground.
- [ ] Add examples
- [ ] Add optional (github?) auth for users such that apps can be persisted under that user
- [ ] Enable users to share their apps
- [ ] Enable users to update their apps
- [ ] Enable users to easily share at pyscript.com, Hugging Face spaces or similar.
- [ ] Add a gallery for discovery and appraisal of apps
- [ ] Add functionality to take screenshots of apps to show in gallery
- [ ] Figure out how *versions* of Panel, Bokeh, Tabulator etc. should be managed.
- [ ] Figure out what it requires to main this. Should some code run automatically once in a while to rebuild apps using latest versions of Panel, Bokeh etc.?
- [ ] Lots of testing of real world application development. How does the live reload work when we start to develop larger applications that load a lot of data.
- [ ] Should we actually enable developing in the (jupyterlite) notebook with live reload rendering on the right?
- [ ] Fix workaround where `#main` div is moved outside shadowroot and then back in after panel app has rendered.

### DIFF
--- a/examples/gallery/playground/playground.py
+++ b/examples/gallery/playground/playground.py
@@ -14,6 +14,37 @@ pn.extension("terminal")
 pn.panel("Hello World").servable()
 """
 
+TRANSFORMERS_EXAMPLE = '''\
+import panel as pn
+
+MODEL = "sentiment-analysis"
+pn.chat.ChatMessage.default_avatars["hugging face"] = "🤗"
+
+pn.extension(design="material")
+
+@pn.cache
+async def _get_pipeline(model):
+    from transformers_js import import_transformers_js
+    transformers = await import_transformers_js()
+    return await transformers.pipeline(model)
+
+async def callback(contents: str, user: str, instance: pn.chat.ChatInterface):
+    pipe = await _get_pipeline(MODEL)
+    response = await pipe(contents)
+    label, score = response[0]["label"], round(response[0]["score"], 2)
+    return f"""I feel a {label} vibe here (score: {score})"""
+
+welcome_message = pn.chat.ChatMessage(
+    f"I'm a Hugging Face Transformers `{MODEL}` model.\\n\\nPlease *send a message*!",
+    user="Hugging Face",
+)
+
+pn.chat.ChatInterface(
+    welcome_message, placeholder_text="Loading the model ...",
+    callback=callback, callback_user="Hugging Face",
+).servable()
+'''
+
 EXAMPLES = {
     "Hello World": HELLO_WORLD_EXAMPLE,
     "Reactive Expressions": """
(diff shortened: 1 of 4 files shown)


MarcSkovMadsen:
I've tried to reuse pyodide outside, inside and across iframes. I've been given hope in https://github.com/pyodide/pyodide/issues/4279 that it is possible. But currently I'm stuck. I've reported some reproducible examples in https://github.com/pyodide/pyodide/issues/4279#issuecomment-1793416640.

MarcSkovMadsen:
Replaced by #5812

PR #5772, opened 2023-10-29
Autofind transformers_js/ transformers-js-py

MarcSkovMadsen:
Addresses https://github.com/whitphx/transformers.js.py/issues/26 as discussed in https://github.com/whitphx/transformers.js.py/pull/23#discussion_r1372601075

I renamed from `find_imports` to `find_requirements` because it find the package requirements and not the *imports* (?) as previously described by the docstring.


philippjfr (approved):
Looks good!

Issue #5775, opened 2023-10-30
404 error for some css resources for deep links


armaaar:
#### ALL software version info

Python 3.10.13
Panel: 1.3.0
Bokeh: 3.3.0
Param: 2.0.0

#### Description of expected behavior and the observed behavior

When notifications are enabled and you are serving a page from a path that is at least 2 levels deep, you get 404 errors

#### Complete, minimal, self-contained example code that reproduces the issue

```python
import panel as pn

pn.extension(
    notifications=True,
)

pn.serve({"deep/link": "# Hello"})

```

#### Stack traceback and/or browser JavaScript console output

```
python issues.py
Launching server at http://localhost:52363
404 GET /deep/static/extensions/panel/bundled/notificationarea/notyf@3/notyf.min.css (::1) 3.33ms
404 GET /deep/static/extensions/panel/bundled/font-awesome/css/all.min.css (::1) 0.25ms
```

#### Screenshots or screencasts of the bug in action

![Screenshot 2023-10-30 at 10 39 26 AM](https://github.com/holoviz/panel/assets/25823409/ca068bbc-d4ac-49b8-bbfe-02015298e924)

- [ ] I may be interested in making a pull request to address this

PR #5779, opened 2023-10-30
Adjust ReactiveHTML css resources for relative paths

philippjfr:
Fixes https://github.com/holoviz/panel/issues/5775


armaaar:
@philippjfr Can I ask in which version will this fix be available?

PR #5781, opened 2023-10-30
Ensure invalid query parameters warn instead of erroring

philippjfr:
An invalid query parameter should not be able to break an app.

### DIFF
--- a/panel/util/__init__.py
+++ b/panel/util/__init__.py
@@ -8,6 +8,7 @@ import base64
 import datetime as dt
 import inspect
 import json
+import logging
 import numbers
 import os
 import pathlib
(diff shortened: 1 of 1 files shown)


panel/util/__init__.py
```diff
@@ -247,7 +248,13 @@ def parse_query(query: str) -> dict[str, Any]:
             try:
                 parsed_query[k] = json.loads(v)
             except Exception:
-                parsed_query[k] = ast.literal_eval(v)
+                try:
+                    parsed_query[k] = ast.literal_eval(v)
+                except Exception:
+                    warn(
```
Hoxbro, line 254: Thinking more about this, I think it would make sense to log it over a warning.

PR #5791, opened 2023-10-31
Add nested select

ahuang11:
```python
import panel as pn
import param

pn.extension()

d = {
    "Andrew": {
        "temp": [1000, 925, 700, 500, 300],
        "vorticity": [500, 300],
    },
    "Ben": {
        "temp": [500, 300],
        "windspeed": [700, 500, 300],
    },
}

from panel.widgets import NestedSelect

NestedSelect(options=d, level_names=["product", "variable", "level"]).show()
```

Have selects depend on each other; worked on this together with @benbarn313.

https://github.com/holoviz/panel/assets/15331990/d54973cb-5c9f-4a8f-8fbe-eccf17f98903

- [ ] tests
- [ ] docs

Wondering if we should support this use case, where the nested levels are not even.
```
d = {
    "Andrew": {
        "temp": [1000, 925, 700, 500, 300],
        "vorticity": [500, 300],
    },
    "Ben": {
        ["temp", "windspeed"],
    },
}

```


ahuang11:
I believe this is ready to add tests + docs

https://github.com/holoviz/panel/assets/15331990/ccb6888f-0b3e-4314-83e2-6acd9edb7d70

ahuang11:
I added variable depth support

https://github.com/holoviz/panel/assets/15331990/d4e83c97-667a-4274-a1b9-559405a3d899

ahuang11:
Talked with Philipp. We decided to first rollback variable level nesting, e.g. the following needs more design and should not be allowed for now due to its complexity.
```
options = {
    "Ben": ["A", "B"],
    "Andrew": {
        "temp": [1000, 925, 700, 500, 300],
        "vorticity": [500, 300],
    },
}
select = NestedSelect(options=options, value=["Ben", "B"])
```

However, given uniform level nesting, we're still debating whether `value` should be a `tuple` vs `dict`, e.g. given:

```
options = {
    "Andrew": {
        "temp": [1000, 925, 700, 500, 300],
        "vorticity": [500, 300],
    },
    "Ben": {
        "temp": [500, 300],
        "windspeed": [700, 500, 300],
    },
}

levels = [{'name': 'Name', 'type': MultiChoice, 'default': 'Ben'}, {'name': 'Variable', 'default': "temp"}, {'name': 'Pressure', 'type': DiscreteSlider, 'default': 300}]

select = NestedSelect(options=options, levels=levels)
```

Should `select.value == ("Ben", "temp", 300)` or `select.value == {"Name": "Ben", "Variable": "temp", "Pressure": 300}`?

jbednar:
It seems to me that variable nested levels will come up often, particularly during development. The dict format sounds better to me. Is there any future for dynamic updating with this approach? E.g. instead of a list for a particular entry can you imagine accepting a callable? If so, what would the arguments to the callable be?  It's fine to say that more complex cases like that should be handled by nested Parameterized classes; just seeing if they are _explicitly_ out of scope or only not yet considered.

ahuang11:
> instead of a list for a particular entry can you imagine accepting a callable

Good point! I think it might be out of scope for now, but I imagine it can be a feature enhancement in the near future, when we think of the design a little more...

Here's an example:
```python
options = {
    "Andrew": {
        "temp": get_temp_levels,
        "windspeed": get_wspd_levels",
    },
    "Ben": {
        "temp": get_temp_levels,
        "windspeed": get_wspd_levels",
    },
}
```

However, I don't know the input kwargs either. Perhaps it's just the current nested select `value`

philippjfr:
> It seems to me that variable nested levels will come up often, particularly during development.

Can we gather some use cases for this? I wasn't able to come up with many good ones.

jbednar:
During development, if you have to add the same number of items to every sublevel, it seems tricky to try things out...

philippjfr:
Variable nesting refers to the number of levels not the number of items at each level of I understand things correctly.

jbednar:
I think we're all saying the same thing, but to be specific, I'm referring to filling out suboptions here for one option but not yet the others:

- Category A:
    -  Option 1
    -  Option 2
         - Suboption 1
         - Suboption 2

The question is whether Option 1 is required to have suboptions here once you add them for Option 2. I believe the answer is yes right now, but ideally we wouldn't have to add them at every level if we add them at one level.

ahuang11:
> The question is whether Option 1 is required to have suboptions here once you add them for Option 2. I believe the answer is yes right now

That is correct. Maybe in a future PR when callbacks are supported, we can also have dynamic variable nesting as well.

philippjfr:
>I think we're all saying the same thing,

I don't think we are (or maybe I'm still misunderstanding) but I think I get your request now. Effectively you want a partial specification to be valid so you don't have to fill out the full tree immediately just to get something working. That's not the same as having a variable number of levels though. I fully agree that what you're saying should be supported but we should make a decision what the behavior of such a partial specification should be, i.e. levels that are not fully specified should either:

- Display an empty select widget
- Hide the widget for a level if no options are provided

philippjfr:
So to be clear, I think we can and should allow what you described without actually supporting variable levels of nesting as I understand it based on Andrew's initial description of it.

panel/widgets/select.py
```diff
@@ -319,6 +319,258 @@ def values(self):
...
+        # little optimization to avoid looping through all the
+        # widgets and updating their value
+        for start_i, select in enumerate(self._widgets):
+            if select is event.obj:
+                break
+
+        options = self.options.copy()
+        # batch watch to prevent continuously triggering
+        # this function when updating the select widgets
+        with param.batch_watch(self):
```
philippjfr, line 500: Nice!

philippjfr (approved):
Looks great!

PR #5792, opened 2023-11-01
Adds VS Code simple browser to the docs

MarcSkovMadsen:
I recently discovered the VS Code *simple browser*. And its made my developer experience much smoother, so I would like to add it to the VS Code guide.

You can preview the updated guide [here](https://github.com/holoviz/panel/blob/enhancement/vscode-simple-browser/doc/how_to/editor/vscode_configure.md#simple-browser).


maximlt:
It seems it comes with the *Live Preview* extension, that's what its description says:

> This extension is most useful for projects where a server is not already created (e.g. not for apps using React, Angular, etc.). To work with these, feel free to run the Simple Browser: Show command that is already built-in with VS Code.

So it might be wise saying that you need to have this extension installed and enabled.

philippjfr:
>So it might be wise saying that you need to have this extension installed and enabled.

The way I understand it, VSCode ships with Simple Browser by default, Live Preview is only needed to start a web server, which we don't need.

PR #5802, opened 2023-11-02
Fall back to rendering cell outputs if notebook does not define servable

philippjfr:
This PR addresses a usability issue of Panel notebook deployments. Specifically it allows users that do not know anything about Panel to render their notebooks as apps and even re-arrange the outputs on a page. This works by adding a fallback case, when there are no servables published in the notebook, that will automatically populate the app with all cell outputs and put them inside the newly added `EditableTemplate` which will allow re-arranging the components on the page.

drag_drop (1)

- [x] Add support for fallback option without changing behavior for notebooks that do export servable components
- [x] Add `EditableTemplate` that allows users to drag, resize and delete components
- [x] Add support for persisting positions of InteractJS components in notebook metadata
- [x] Add ability to hide/remove specific cell outputs 
- [x] Add documentation
- [x] Add tests


jbednar:
Looks great! Can cell outputs that are marked resizable in the original notebook also resize automatically in the template (expanding to the full width available), or do they become fixed size when dragged around?

jbednar:
What happens to Markdown text cells?

philippjfr:
Currently nothing but we could certainly have a report mode.

jbednar:
If it becomes easy to delete outputs (no idea how to achieve that), then I'd vote for keeping the Markdown output too.

MarcSkovMadsen:
Dash released app studio which turns any notebook into a dashboard. It does more than you try to do here @philippjfr 

https://www.youtube.com/watch?v=9PH9MtKVf6U

maximlt:
It's a paid product, right?

jbednar:
Yes, it's paid, and also apparently tied to a server: "You must be on Dash Enterprise 5.2 to access Plotly App Studio".  So immediately it would seem to be locking up the app that you create to be usable only in a tiny, locked-down range of scenarios.

I'd have to use it for a while to see whether it's something I'd actually consider a good idea. It's definitely carving things up in a different way than we have been. Seems like it's a mix of pn.rx, the layout mechanism in this PR, and Lumen Builder, and whether that makes sense for Plotly users depends on how well it fits in with the rest of the Dash/Plotly ecosystem. 

Like the mechanism in this PR, it can take a notebook you wrote and make it into an app without further coding. Then it goes further by giving you a GUI for some aspects of the app, but from that demo I can't tell whether it actually does much more than what the drag-and-drop interface in this PR already provides (in what I consider a more natural interface already). What feels constraining to me is that when you reach the limits of that GUI, I didn't see any option for programmatic control over layout and look and feel, which seems like it would irritate people who by definition can program in Python at least a bit (or they wouldn't be authoring notebooks).  So apart from whether it's OSS or not, it isn't clear to me that it would be a good experience beyond the initial thrill of making an app from nothing.

Plus it seems to embody a mistaken notion that there are notebooks on the one hand and apps on the other, such that the notebook can define a reactive attribute but widgets for manipulating that appear only in the separately rendered app. I think that's completely the wrong approach; even a single widget with a single output is enormously powerful in the notebook as a compact way to present and explore large numbers of options and large quantities of data. Widgets aren't just for deployed server apps! Ugh.

"Tens, hundreds, sometimes thousands of lines of code to create an app like this" -- Maybe for JavaScript, but it's hard to imagine a Python app framework taking thousands of lines to make what looked like a trivially simple app. In fact, the only framework that I'd think would come close is Dash. :-)

Coderambling:
Nice! After dragging around and re-arranging elements, is there a way to copy / paste / save the arrangement (as code)?

MarcSkovMadsen:
> Nice! After dragging around and re-arranging elements, is there a way to copy / paste / save the arrangement (as code)?

That was also my first thought. How can I better persist this and use it with `.servable()` approach?

Coderambling:
> > Nice! After dragging around and re-arranging elements, is there a way to copy / paste / save the arrangement (as code)?
> 
> +1. That was also my first thought. How can I better persist this and use it with `.servable()` approach?

In Discord, @jbednar pointed out that it is saved in the metadata of the Notebook. But I don't think that info can be extracted?

Maybe it is possible to use the mechanism that stores it there to also export or expose the information in a variable?

Kind of like code_repr (or is it obj_repr ?) does with .controls, expressing the .controls names and values as (almost) Python code.

Would that be a viable approach at all?

dwr-psandhu:
Is there a plan to support full screen toggles here. I am using @MarcSkovMadsen fullscreen.py for components ?

Coderambling:
@philippjfr has explained further details about the implementation in Discord [here](https://discord.com/channels/1075331058024861767/1088149441988526241/1223914324633063464) (https://discord.com/channels/1075331058024861767/1088149441988526241/1223914324633063464) and in the rest of that thread.

Issue #5801, opened 2023-11-02
illogical order of execution for `authorize_callback`, causing errors and performance degradation


armaaar:
#### ALL software version info
Python 3.10.13
Panel: 1.3.1
Bokeh: 3.3.0
Param: 2.0.0

#### Description of expected behavior and the observed behavior

We noticed that our panel pages are evaluated even if the current user is not authorized to view them, Causing errors and performance degradation. We expect to first evaluate if the user can access the page before actually call the page content.

Expected behavior:

1. User visit page in browser
2. `authorize_callback` is called to determine if the user have access to the page
3. if False, don't show or evaluate the page content
4. if True, Evaluate and show the page content

Observed behavior:

1. User visit page in browser
2. page content is evaluated
3. `authorize_callback` is called to determine if the user have access to the page
4. if False, don't show the page content
5. if True, show the page content

#### Complete, minimal, self-contained example code that reproduces the issue

```
import time

import panel as pn


def fast():
    print("Evaluating fast page")
    return "# Hello fast"


def slow():
    print("Evaluating slow page")
    time.sleep(10)
    return "# Hello slow"


def authorize(user_info, request_path):
    print("Authorizing")
    return False


pn.config.authorize_callback = authorize

pn.serve(panels={"fast": fast, "slow": slow})


```

#### Stack traceback and/or browser JavaScript console output

Running the server and visiting `http://localhost:53253/slow` 

```
python issue.py
Launching server at http://localhost:53253
Evaluating slow page
Authorizing
```

#### Screenshots or screencasts of the bug in action

- [ ] I may be interested in making a pull request to address this

PR #5821, opened 2023-11-06
Ensure global authorization callbacks are run before session creation

philippjfr:
Previously the `authorize_callback` was always run **after** a session was created. This is required when the user configures the `authorize_callback` inside their app, but you can also set a global `authorize_callback`, e.g. as part of a `--setup` script or when dynamically starting a server with `pn.serve`. In this scenario we can run the `authorize_callback` without running the session and therefore avoid pointless computation.

Fixes https://github.com/holoviz/panel/issues/5801

### DIFF
--- a/panel/io/server.py
+++ b/panel/io/server.py
@@ -538,15 +538,13 @@ class DocHandler(LoginUrlMixin, BkDocHandler, SessionPrefixHandler):
                 return None, None
             elif not authorized:
                 auth_error = (
-                    f'Authorization callback errored. Could not validate user name "{state.user}" '
-                    f'for the given app "{self.request.path}".'
+                    f'Access denied! User {state.user!r} is not authorized '
+                    f'for the given app {self.request.path!r}.'
                 )
-                if authorized:
-                    auth_error = None
+            if authorized:
+                auth_error = None
         except Exception:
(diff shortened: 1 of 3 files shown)


panel/io/server.py
```diff
@@ -510,8 +511,76 @@ def _token_payload(self):
...
+            error_type='Authorization Error',
+            error='User is not authorized.',
+            error_msg=auth_error
+        )
+
     @authenticated
     async def get(self, *args, **kwargs):
+        # Run global authorization callback
+        payload = self._generate_token_payload()
+        print('!!!!!!!')
```
Hoxbro, line 570:
```suggestion

```

panel/io/server.py
```diff
@@ -510,8 +511,75 @@ def _token_payload(self):
...
     async def get(self, *args, **kwargs):
+        # Run global authorization callback
+        payload = self._generate_token_payload()
+        if config.authorize_callback:
+            temp_session = generate_session(
+                self.application, self.request, payload, initialize=False
+            )
+            with set_curdoc(temp_session.document):
+                authorized, auth_error = self._authorize()
+            print('>>>>', authorized)
```
philippjfr, line 574:
```suggestion
```

Issue #5815, opened 2023-11-05
Issue with oauth2 authentication, for google oauth and panel version above 1.3.0: "Some requested scopes were invalid ... invalid=[offline_access]"


el-abcd:
SUMMARY:
From some googling, I __think__ that google oauth does NOT like the URL sent to it to contain: "&scope=...+offline_access", instead it wants to have a "&access_type=offline&" segment in the URL.  
IIUC, using "&scope=...+offline_access" might be the "openid standard way" but google oauth might not be following the standard for this field?

References:
* https://developers.google.com/identity/protocols/oauth2/web-server#creatingclient
  * only documents using the "access_type" parameter
* https://github.com/ory/kratos/pull/3088
  * Sounds like a very similar issue in another open-source package.
* https://community.qlik.com/t5/Official-Support-Articles/quot-Error-400-Invalid-Scope-invalid-offline-access-quot-when/ta-p/1808663
  * Also sounds like the same issue

DETAILS:

#### ALL software version info
panel version 1.3.0 and 1.3.1.  
I am running this on a docker container in google cloud run.  Based on my debugging so far I think it is very likely a change in panel 1.3.0 (this is working in version 1.2.3)
Most likely in this change:
Unify OAuth implementations and refresh access_token (https://github.com/holoviz/panel/pull/5627)

#### Description of expected behavior and the observed behavior
I am running this in a docker container running in google cloud run.  
When using panel==1.2.3 the oauth login page "works as expected", I see the prompt asking me to authorize.  

When running with panel==1.3.0 (and 1.3.1) I see an error screen: 
<img width="378" alt="image" src="https://github.com/holoviz/panel/assets/39447831/4ead4531-c182-40b2-afe9-0c64028716e5">

I am launching my code in a docker container, using a command line like this...
```
CMD panel serve 220514_01_awesome_panel.py --address 0.0.0.0 --port 8080 --allow-websocket-origin="*" \
--oauth-provider=google \
--oauth-key=<my key> \
--oauth-redirect-uri=<my redirect url>
```

From inspecting the chrome developer tools I think the issue is the "+offline_access" after the &scope=:
* BAD: 
  * https://accounts.google.com/o/oauth2/v2/auth?response_type=code&redirect_uri=https....googleusercontent.com&state=...&scope=openid+email+profile+offline_access
* It WORKS if I edit the URL to use &access_type=offline instead:
  * &scope=openid+email+profile&access_type=offline&
[...]

PR #5823, opened 2023-11-06
Fix Google OAuth default scopes

philippjfr:
Fixes https://github.com/holoviz/panel/issues/5815

### DIFF
--- a/panel/auth.py
+++ b/panel/auth.py
@@ -94,11 +94,13 @@ class OAuthLoginHandler(tornado.web.RequestHandler, OAuth2Mixin):
         'User-Agent': 'Tornado OAuth'
     }
 
-    _access_token_header = None
+    _DEFAULT_SCOPES = ['openid', 'email', 'profile', 'offline_access']
 
     _EXTRA_TOKEN_PARAMS = {
         'grant_type':    'authorization_code'
     }
+
+    _access_token_header = None
 
     _state_cookie = None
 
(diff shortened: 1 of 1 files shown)


el-abcd:
FWIW:
* I looked at the code change and it LGTM.
* I tried building / deploying from the "tip of main" and the oauth flow worked.


Details:
* I used this in my requirements.txt (to install tip of main):  
  * git+https://github.com/holoviz/panel.git@main#egg=panel
* Tweaked my dockerfile to install git + node.js
  * RUN apt-get update && apt-get install -y git
  * RUN apt-get update && apt-get install -y nodejs

Then I had to fiddle around a bit to figure out how to "reset" so I would see the initial oauth login.  I think the final piece was removing the id_token and access_token cookies.  
Then the google oauth flow "worked as expected".  

NOTE:
* The tip of main DID have an issue for me (when run on docker+google cloud run), I see the "Panel Application" header and a blank sidebar, but no controls.  
* And I see a 404 error for a "panel.min.js" URL.  
  * https://<MY APP URL>/static/extensions/panel/panel.min.js?v=05c64cb6a9476411f5c1be2445cb7e163b575262cf76c1c116348335d4f9686e"

But, that is clearly a separate issue, and maybe running against the "tip of main" isn't supported / a great idea?  
And "tip of main" DID work for me locally (mac osx).  

If useful I could file that as a separate issue.  

BUT, this one looks like it is fixed! 👍

petegordon:
I just want to say you guys are amazing! Thank you!

I hit this bug and was wrestling with it on panel 1.3.1 this morning thinking I should be using extra_oauth_params for the offline access. I upgraded panel and it fixed it no problem. Thank you!

Issue #5351, opened 2023-07-31
enable multiselect in Tabulator header_filter


dslemusp:
Hi guys. I am trying to configure a `list` header filter enabling [`multiselect`](https://tabulator.info/docs/5.5/edit#editor-list) but it does not seem to work.

```python
table_filters = {
    "experiment_type": {"type": "list","multiselect":True},
}
pn_table = pn.widgets.Tabulator(
    experiments_df[["experiment_type"]], 
    pagination="remote", 
    page_size=30, 
    sizing_mode="stretch_width", 
    header_filters=table_filters
)
pn_table.show()
```

![image](https://github.com/holoviz/panel/assets/6799189/29c0e77a-0951-4b9f-ae8a-97e36b87eab2)

compared to the default beahviour

```python
table_filters = {
    "experiment_type": {"type": "list"},
}
pn_table = pn.widgets.Tabulator(
    experiments_df[["experiment_type"]], 
    pagination="remote", 
    page_size=30, 
    sizing_mode="stretch_width", 
    header_filters=table_filters
)
pn_table.show()
```

![image](https://github.com/holoviz/panel/assets/6799189/1efefa3a-2e87-49b5-a022-c29bc6e5000b)

_Originally posted by @dslemusp in https://github.com/holoviz/panel/issues/2815#issuecomment-1649988136_

maximlt:
Hi @dslemusp,

This is actually supported by Panel, you need:
- to set `'func'` to `'in'` when `multiselect` is `True`
- to provide the values, either by setting `valuesLookup` to `True` or by passing a list of values to `values`

An example is added in https://github.com/holoviz/panel/issues/5351.

PR #5825, opened 2023-11-06
Tabulator: document how to set a *multiselect* header filter

maximlt:
Addresses https://github.com/holoviz/panel/issues/5351 by adding an example of a *multiselect* header filter. I've also added a test to make sure this is supported correctly.


philippjfr (approved):
Looks good, thanks!

Issue #3117, opened 2022-01-19
Add a Search widget


maximlt:
Panel has the `AutocompleteInput` widget that helps application users to select an item among a large list of options available. However it only works from left-to-right. A full-blown Search widget would be a very nice addition to Panel.

xavArtley:
Some times ago I implemented this widget:
![ms_widget](https://user-images.githubusercontent.com/18531147/150354330-09f3ed88-52ee-4a5f-b42e-64ab2e3588fd.gif)

It needs some polish but it can be a base
The code can be found in this repository
https://github.com/xavArtley/BokehParallelCoordinates.git
https://github.com/xavArtley/BokehParallelCoordinates/blob/7a743e5d92e72072fa647bbb225dd122a1bf280f/pcp/models/multiselect.py
https://github.com/xavArtley/BokehParallelCoordinates/blob/7a743e5d92e72072fa647bbb225dd122a1bf280f/pcp/models/multiselect.ts

maximlt:
Thanks @xavArtley for sharing the code of this widget which seems very powerful. Sort of a mix of a Search widget/Select/Multichoice. Btw I also recently opened an issue for a Select widget with a select-all feature, a la Excel or Tableau (https://github.com/holoviz/panel/issues/3018).

alon-sht:
This widget would be very useful!!
Some kind of a mix between MultiChoice and Select widgets. 
All the above suggestions sound great:
- Option to Select All or Dis-select all
- Search bar that filters the shown dropdown - this can be either in a dedicated search bar or in the widget itself. 
- Maybe also an option to "select all" that selects just the filtered results (after filtering by the search bar).

maximlt:
Thanks for chiming in @hyamanieu, what you say totally makes sense to me.

> I would focus this issue on changing the search method of `AutocompleteInput` to search within the whole string rather than just the beginning. What do you think?

I am not sure it's a good idea. The semantics of auto-completing is different from searching. Also if the search functionality is embedded in the auto-complete widget it makes it harder for users to find it. I'd be more in favor of dedicated Search widget.

PR #5832, opened 2023-11-07
`AutocompleteInput`: expose and document `search_strategy`

maximlt:
Addresses https://github.com/holoviz/panel/issues/3117

Very very happy to find out that Bokeh added in https://github.com/bokeh/bokeh/pull/13032 a `search_strategy` property to `AutocompleteInput` that when set to `'includes'` allows to to search through the whole string and not just the start.

In https://github.com/holoviz/panel/issues/3117 I was asking specifically for a Search widget but I think I had a too narrow vision of what autocomplete means, I think it can closed as  https://github.com/bokeh/bokeh/pull/13032 implemented the feature I was after.

### DIFF
--- a/examples/reference/widgets/AutocompleteInput.ipynb
+++ b/examples/reference/widgets/AutocompleteInput.ipynb
@@ -3,7 +3,9 @@
   {
    "cell_type": "code",
    "execution_count": null,
-   "metadata": {},
+   "metadata": {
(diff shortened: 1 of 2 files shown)


maximlt:
Test failures look unrelated.

philippjfr (approved):
Neat, looks good!

Issue #5711, opened 2023-10-21
ChatInterface: Expose Actions


MarcSkovMadsen:
I'm on the `main` branch of Panel working on contributing a LangChain Pandas DataFrame Agent. https://github.com/holoviz-topics/panel-chat-examples/pull/83

I would like to be able to reset the *agent*. For that we have the *reset* button. But this button resets the `objects` of the `ChatInterface` only. And I don't see any `reset` action/ event exposed that I can bind or depends to.

Please expose a `reset` event.

Similar for other actions like `revert`, `undo` and `rerun`. When I use these, I will sometimes have to add custom logic to handle  what my model or agent should do.

On the same note. Please also expose the method to let me clear the `ChatInterface` programmatically.

On the same note. Please let me add custom buttons to the task bar. Like a custom `Clear` button.

## Workaround

I can add the `reset_button` manually and somewhere else via 

```python
def _reset(_=None):
    state.reset()
    chat_interface.objects=[] # Clear Chat Interface
    chat_interface.send(
        state.welcome_message, user="Pandas Agent", avatar="🐼", respond=False
    )

_reset()
reset_button = pn.widgets.Button(name="Restart", on_click=_reset, button_type="primary", button_style="outline")
```

MarcSkovMadsen:
On the same note. Please also expose the method to let me clear the `ChatInterface` programmatically.

MarcSkovMadsen:
On the same note. Please let me add custom buttons to the task bar. Like a custom `Clear` button.

ahuang11:
Right, I had this in mind down the line https://github.com/holoviz/panel/issues/5567#issue-1922999520

Associate additional hooks to the existing buttons, and/or allowing users to add buttons with custom functionality easily.

PR #5839, opened 2023-11-09
Add custom callbacks to ChatInterface buttons

ahuang11:
Closes https://github.com/holoviz/panel/issues/5711

So users can add additional functionality to existing buttons, or add additional buttons easily.

I can't figure out how to drop event here though.

```python
import panel as pn

pn.extension()

pn.chat.ChatInterface(
    custom_callbacks={
        ("Random", "activity"): lambda instance, event: instance.send(
            "Here have a random number"
        ),
        "send": lambda instance, event: instance.send("Hi another send"),
    }
)
```

image


ahuang11:
@MarcSkovMadsen with this PR, you can do something like:

```python
def reset_welcome(instance, event):
    instance.send(
        state.welcome_message, user="Pandas Agent", avatar="🐼", respond=False
    )
...
ChatInterface(..., button_properties={"reset": {"post_callback": reset_welcome}})
```

panel/chat/interface.py
```diff
@@ -341,7 +420,11 @@ def _click_rerun(self, _):
             return
         self.send(value=messages[0], respond=True)
 
-    def _click_undo(self, _):
+    def _click_undo(
+            self,
+            event: param.parameterized.Event | None = None,
+            instance: Self | None = None
+        ) -> None:
```
philippjfr, line 427:
```suggestion
    def _click_undo(
        self,
        event: param.parameterized.Event | None = None,
        instance: Self | None = None
    ) -> None:
```

panel/chat/interface.py
```diff
@@ -330,7 +405,11 @@ def _reset_button_data(self):
             button_data.objects.clear()
             self._toggle_revert(button_data, False)
 
-    def _click_rerun(self, _):
+    def _click_rerun(
+            self,
+            event: param.parameterized.Event | None = None,
+            instance: Self | None = None
+        ) -> None:
```
philippjfr, line 412:
```suggestion
    def _click_rerun(
        self,
        event: param.parameterized.Event | None = None,
        instance: Self | None = None
    ) -> None:
```

panel/chat/interface.py
```diff
@@ -259,7 +302,39 @@ def _init_widgets(self):
...
+                    finally:
+                        self.disabled = False
+            return wrapper
+        return decorate
+
+    def _click_send(
+            self,
+            event: param.parameterized.Event | None = None,
+            instance: Self | None = None
+        ) -> None:
```
philippjfr, line 337:
```suggestion
    def _click_send(
        self,
        event: param.parameterized.Event | None = None,
        instance: Self | None = None
    ) -> None:
```

panel/chat/interface.py
```diff
@@ -353,12 +436,19 @@ def _click_undo(self, _):
...
         else:
             self.extend(undo_objects)
             self._reset_button_data()
 
-    def _click_clear(self, _):
+    def _click_clear(
+            self,
+            event: param.parameterized.Event | None = None,
+            instance: Self | None = None
+        ) -> None:
```
philippjfr, line 451:
```suggestion
    def _click_clear(
        self,
        event: param.parameterized.Event | None = None,
        instance: Self | None = None
    ) -> None:
```

panel/chat/interface.py
```diff
@@ -100,6 +106,17 @@ class ChatInterface(ChatFeed):
...
 
+    button_properties = param.Dict(default={}, doc="""
+        A mapping of button names to its properties (`callback`, `post_callback`, `icon`).
+        This can be used to update default buttons' properties, or create new buttons
+        to be added to the ChatInterface input row. The callback signature must accept two
+        positional args `interface` and `event`, where `instance` is the ChatInterface instance
+        and event is the button click event. Button names that overlap with
+        the default buttons (`send`, `rerun`, `undo`, `clear`) will override the default icon
+        if specified, but does not overwrite the default callback--only prepends to it.
+        Keys that do not overlap will create new buttons, and each require callback or post_callback.
```
philippjfr, line 117:
```suggestion
        Allows overriding or adding buttons by providing a mapping from the button name to
        a dictionary containing the `icon`, `callback` and `post_callback`. Specifying button
        names that match one of the default buttons (`send`, `rerun`, `undo`, `clear`) allows
        overriding the default icon but does not overwrite the default callback, instead
        prepending the callback. Specifying button names as keys that do not match one of the
        existing names will create new buttons and must provide a callback or post_callback.
        The signature of provided callbacks must accept two positional arguments: `instance`
        and `event`, where `instance` is the ChatInterface instance and event is the button click
        event. 
```

philippjfr:
Needs rebasing after the merge of the serialize PR also `typing_extensions` is no longer a dependency so the Self import has to be removed.

philippjfr (approved):

PR #5848, opened 2023-11-11
Speed up Pane.clone with 50%

MarcSkovMadsen:
Pane.clone did the same look up three times which was costly.

This changes to one lookup and also makes the code more readable.

## Measurement 1

Running pytest on below code shows ~3% speed up consistently.

```python
import panel as pn
item = pn.pane.Markdown("Hello World")

def test_html_clone():
    for i in range(20000):
        item.clone()
```

The reason why I found this is that I can see in apps `.clone` loops over `self.param.values().items()`. And loops over `self.param.values().items()` takes up +5% of the time of loading apps in my tests. My hypothesis is that that one could be improved significantly.

### DIFF
--- a/panel/viewable.py
+++ b/panel/viewable.py
@@ -45,7 +45,7 @@ from .io.notebook import (
 from .io.save import save
 from .io.state import curdoc_locked, set_curdoc, state
 from .util import escape, param_reprs
-from .util.param import get_params_to_inherit
+from .util.parameters import get_params_to_inherit
 from .util.warnings import deprecated
 
 if TYPE_CHECKING:
--- a/panel/viewable.py
+++ b/panel/viewable.py
@@ -55,6 +55,15 @@ if TYPE_CHECKING:
     from .io.location import Location
     from .io.server import StoppableThread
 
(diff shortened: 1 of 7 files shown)


philippjfr (approved):

ahuang11:
Wow super excited for all the speed gains you're discovering!

philippjfr:
I've done this a few times now so most low hanging fruit are likely gone but after param 2.0 some performance characteristics have likely changed. We should really add a basic asv based performance profiling suite, even if it only covers a small number of test cases initially.

panel/viewable.py
```diff
@@ -55,6 +55,15 @@
     from .io.location import Location
     from .io.server import StoppableThread
 
+def _should_inherit(self, p, v):
+        _p = self.param[p]
+        return v is not _p.default and not _p.readonly and (v is not None or _p.allow_None)
+
+def _get_items_to_inherit(self: param.Parameterized)->Dict:
```
philippjfr, line 62: Should probably be moved to `panel/util/__init__.py`

PR #5851, opened 2023-11-11
Speed up config.__getattr__ with a factor 10 corresponding to 5-10% app speed up.

### LINKED ISSUE #5850: Speed up initial load by 5-10% by making look up of config values efficient
I've been profiling different apps. What I see is that `config.__getattribute__` is mentioned all over the place taking up more than 5% of the time. Looking up the `loading_indicator` value on `pn.config` should not be costly 😄 

## Reproduce

```python
import panel as pn

N=10

pn.Column(
    *[pn.Row(
        *[pn.pane.Markdown("Hello World") for i in range(N)]
    ) for i in range(N)]
).servable()
```

```bash
panel serve script.py --autoreload --index script --admin --profiler pyinstrument
```

Open the app 5 times and visit the pyinstrument report.

image

For me `config.__get_attribute__` adds up to 11.5% of the time spent!.

MarcSkovMadsen:
Closes #5850

## Pytest

I've run the below test on the `main` branch and `this` branch.

```python
from pathlib import Path
import pytest
from pyinstrument import Profiler
import panel as pn

TESTS_ROOT = Path.cwd()

@pytest.fixture(autouse=True)
def auto_profile(request):
    PROFILE_ROOT = (TESTS_ROOT / ".profiles")
    # Turn profiling on
    profiler = Profiler()
    profiler.start()

    yield  # Run test

    profiler.stop()
    PROFILE_ROOT.mkdir(exist_ok=True)
    results_file = PROFILE_ROOT / f"{request.node.name}.html"
    profiler.write_html(results_file)

item = pn.pane.Markdown("Hello World")

def method1():
    pn.config.loading_indicator
    pn.config.npm_cdn
    pn.config.sizing_mode
    pn.config.admin

def test_performance():
    N = 100000
    for i in range(N):
        method1()
```

I've achieved a 75% speed up. Left is `main` branch, Right is `this` branch

image

## App

Compared to the profiling of the app in the reported issue, it looks much better. `config.__getattribute__` is only in one place with a smaller proportion of 1% compared to the 11% before. Still I think that is too much.

image

### DIFF
--- a/panel/config.py
+++ b/panel/config.py
@@ -336,7 +336,7 @@ class _config(_base_config):
         super().__init__(**params)
         self._validating = False
         self._parameter_set = set(self.param)
-        for p in self.param:
+        for p in self._parameter_set:
             if p.startswith('_') and p[1:] not in _config._globals:
                 setattr(self, p+'_', None)
         if self.log_level:
@@ -372,7 +372,7 @@ class _config(_base_config):
     def set(self, **kwargs):
         values = [(k, v) for k, v in self.param.values().items() if k != 'name']
         overrides = [
-            (k, getattr(self, k+'_')) for k in self.param
+            (k, getattr(self, k+'_')) for k in self._parameter_set
             if k.startswith('_') and k[1:] not in _config._globals
         ]
         for k, v in kwargs.items():
@@ -398,9 +398,9 @@ class _config(_base_config):
         if attr in _config._globals or self.param._TRIGGER:
             super().__setattr__(attr if attr in self.param else f'_{attr}', value)
         elif state.curdoc is not None:
-            if attr in self.param:
+            if attr in self._parameter_set:
                 validate_config(self, attr, value)
-            elif f'_{attr}' in self.param:
+            elif f'_{attr}' in self._parameter_set:
                 validate_config(self, f'_{attr}', value)
             else:
                 raise AttributeError(f'{attr!r} is not a valid config parameter.')
@@ -410,7 +410,7 @@ class _config(_base_config):
             watchers = param_watchers(self).get(attr, {}).get('value', [])
             for w in watchers:
                 w.fn()
-        elif f'_{attr}' in self.param and hasattr(self, f'_{attr}_'):
+        elif f'_{attr}' in self._parameter_set and hasattr(self, f'_{attr}_'):
             validate_config(self, f'_{attr}', value)
             super().__setattr__(f'_{attr}_', value)
         else:
--- a/panel/config.py
+++ b/panel/config.py
@@ -37,13 +37,8 @@ __version__ = str(param.version.Version(
 
 _LOCAL_DEV_VERSION = any(v in __version__ for v in ('post', 'dirty')) and not state._is_pyodide
 
-#---------------------------------------------------------------------
-# Public API
-#---------------------------------------------------------------------
-
-_PATH = os.path.abspath(os.path.dirname(__file__))
-
-GLOBALS = {
+_CONFIG_PARAMS = set() # To be populated later
+_CONFIG_GLOBAL_PARAMS = {
     'admin_plugins', 'autoreload', 'comms', 'cookie_secret',
     'nthreads', 'oauth_provider', 'oauth_expiry', 'oauth_key',
     'oauth_secret', 'oauth_jwt_user', 'oauth_redirect_uri',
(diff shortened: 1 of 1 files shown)


panel/config.py
```diff
@@ -394,7 +398,7 @@ def __setattr__(self, attr, value):
         if not init or (attr.startswith('_') and attr.endswith('_')) or attr == '_validating':
             return super().__setattr__(attr, value)
         value = getattr(self, f'_{attr}_hook', lambda x: x)(value)
-        if attr in self._globals or self.param._TRIGGER:
```
MarcSkovMadsen, line 397: This is a major speed impediment because

- It recursively looks up something that is a constant in my understanding.
- Later we run `__contains__` which is slow because the value is a list.

panel/config.py
```diff
@@ -433,19 +437,29 @@ def __getattribute__(self, attr):
...
+
+        if attr=="param":
+            return super().__getattribute__(attr)
+
+        global _PARAMS
+        if not _PARAMS and init:
+            _PARAMS = set(super().__getattribute__('param'))
+
         if init and not attr.startswith('__'):
-            params = super().__getattribute__('param')
```
MarcSkovMadsen, line 446: `super().__getattribute__('param')` is a major speed impediment because

- It looks up something that is a constant in my understanding.
- Later we run `__contains__` which is slow because the value is a list.

panel/config.py
```diff
@@ -433,19 +437,29 @@ def __getattribute__(self, attr):
         """
         from .io.state import state
 
-        if attr == '_param__private':
-            return super().__getattribute__('_param__private')
+        # ATTRS[attr]+=1
+        # print(ATTRS)
+        if attr in ('_param__private', '__class__', 'npm_cdn', 'param'):
```
MarcSkovMadsen, line 442: `npm_cdn` and `param` are asked for very frequently. So I move them to the top. I don't understand if that is a problem.

panel/config.py
```diff
@@ -454,11 +459,11 @@ def __getattribute__(self, attr):
...
             setattr(self, attr, new_obj)
-        if attr in global_params or attr == 'theme':
+        if attr in _CONFIG_GLOBAL_PARAMS or attr == 'theme':
             return super().__getattribute__(attr)
         elif curdoc and curdoc in session_config and attr in session_config[curdoc]:
             return session_config[curdoc][attr]
-        elif f'_{attr}' in params and getattr(self, f'_{attr}_') is not None:
+        elif f'_{attr}' in _params and getattr(self, f'_{attr}_') is not None:
             return super().__getattribute__(f'_{attr}_')
         return super().__getattribute__(attr)
```
MarcSkovMadsen, line 469: If noticed a huge, huge speed up if I change from `self._admin` to `_config.admin` in the property.

image

I don't understand why and I did not change it.

But you can see the decrease below

image

MarcSkovMadsen:
I cannot do more now. The implementation is quite complicated. I am not aware I've broken anything and the tests pass.

I hope you will review and finalize. This needs your knowledge @philippjfr . Thanks.

philippjfr:
Thanks for looking at this, refactoring `config` has been on my to-do list for a while. Will review the details and see if we can squeeze out a little more.

panel/config.py
```diff
@@ -433,19 +433,24 @@ def __getattribute__(self, attr):
...
         try:
             init = super().__getattribute__('_param__private').initialized
         except AttributeError:
             init = super().__getattribute__('initialized')
-        global_params = super().__getattribute__('_globals')
+
+        global _CONFIG_PARAMS
+        if not _CONFIG_PARAMS and init:
+            _CONFIG_PARAMS = set(super().__getattribute__('param'))
+
```
MarcSkovMadsen, line 448: I don't understant if I need to handle `attr.startswith('__')` as a special case. Or if I could just throws `params` way and refer to `_PARAMS`.

panel/config.py
```diff
@@ -433,19 +433,24 @@ def __getattribute__(self, attr):
...
         # _param__private added in Param 2
         try:
             init = super().__getattribute__('_param__private').initialized
         except AttributeError:
             init = super().__getattribute__('initialized')
-        global_params = super().__getattribute__('_globals')
+
+        global _CONFIG_PARAMS
+        if not _CONFIG_PARAMS and init:
+            _CONFIG_PARAMS = set(super().__getattribute__('param'))
```
philippjfr, line 447: May just move this in the constructor.

panel/config.py
philippjfr: Have to think about `npm_cdn` but everything else makes sense here for sure.

panel/config.py
philippjfr: `admin` should have been declared as a global config option.

maximlt:
Certainly `config` was the most difficult piece of code out there I had to deal with when working on Param 2.0.

Issue #5845, opened 2023-11-10
Speed up initial load of Hello World app by 50% by caching jinja template environment


MarcSkovMadsen:
I was looking into the initial load of a Panel app.

I can see that a "Hello World" Panel app uses 53% (190ms of 352ms) of its time in `_env.from_string(template)`.

![image](https://github.com/holoviz/panel/assets/42288570/41890f3c-9290-42a3-8f38-d610ac76bdd2)

![image](https://github.com/holoviz/panel/assets/42288570/81008e51-0a54-4bdf-937b-bf6435b17128)

My understanding is that the `template` string is not dynamic. Its the same again and again and again. We should be able to remove this time using caching.

The jinja docs even avoids against Environments from strings.

![image](https://github.com/holoviz/panel/assets/42288570/4b6c75a0-5787-4921-b4fc-16c4ec08f7b0)

Maybe we can continue providing a string. Maybe we should save those strings to files. But we should implement some caching.

## Hello World App

```python
import panel as pn
import param
from datetime import date

pn.extension(template="fast")

pn.panel("hello").servable()
```

```bash
panel serve script.py --admin --profiler pyinstrument
```

philippjfr:
Great catch, let's prioritize this for 1.3.2

philippjfr:
Not that we shouldn't do this but I'm getting very different timings than you are:

<img width="341" alt="Screenshot 2023-11-11 at 14 04 19" src="https://github.com/holoviz/panel/assets/1550771/4a7c82fd-4818-499f-8ea7-b91554c7c329">

MarcSkovMadsen:
Im showing percentages. You are showing absolute numbers?

philippjfr:
True, although my percentages are **much** smaller too.

ahuang11:
Probably difference between Mac and Windows?

MarcSkovMadsen:
I'm on Linux on a cloud instance

PR #5854, opened 2023-11-11
Cache templates loaded from string

philippjfr:
Fixes https://github.com/holoviz/panel/issues/5845


panel/template/base.py
```diff
@@ -15,6 +15,7 @@
     Type,
 )
 
+import jinja2
```
Hoxbro, line 18: Maybe move this into the method. I think we need to run this file when `pip install -e .` because of `panel bundle`

philippjfr: jinja2 is a core dependency of bokeh so we already satisfy the dependency in the build env.

Issue #5862, opened 2023-11-13
Multithreading: Run multiple deferred functions on the thread pool


ratanasov:
#### Is your feature request related to a problem? Please describe.
With `pn.config.defer_load = True` and multithreading enabled (`pn.config.nthreads`), multiple parts of the page still cannot take advantage of the thread pool, to load concurrently.

#### Describe the solution you'd like
Run multiple deferred functions in the same page load concurrently on the thread pool, when multithreading is enabled. (The deferred functions do run concurrently between multiple concurrent page loads.)

#### Describe alternatives you've considered
This feature request is part of an evaluation of Panel concurrency options, including deferring, multiprocessing and async.

#### Additional context
@philippjfr said in a call this sounds like a reasonable feature request.

#### Code Sample
```py
import panel as pn
import time
import threading

def load(what):
    tid = threading.get_ident()

    print(f"[{time.time():.3f}] loading {what} on {tid} ...")
    time.sleep(5)
    print(f"[{time.time():.3f}] done loading {what} on {tid} ...")

    return f"# {what}"

def app():
    def sidebar():
        return load("sidebar")
    def top():
        return load("top")
    def bottom():
        return load("bottom")

    return pn.Row(sidebar, pn.Column(top, bottom))

if __name__ == "__main__":
    pn.config.defer_load = True
    pn.config.nthreads = 4

    pn.serve(app)
```

The output during initial page load shows that the three `load` calls are run sequentially on the same thread:
```
[1699875623.686] loading sidebar on 6261272576 ...
[1699875628.688] done loading sidebar on 6261272576 ...
[1699875628.710] loading top on 6261272576 ...
[1699875633.713] done loading top on 6261272576 ...
[1699875633.737] loading bottom on 6261272576 ...
[1699875638.743] done loading bottom on 6261272576 ...
```

The output during (almost) simultaneous loads in two browser tabs shows that the three `load` calls are run concurrently on two different threads (one per page?):
```
[1699875680.264] loading sidebar on 6261272576 ...
[1699875680.655] loading sidebar on 6278098944 ...
[1699875685.267] done loading sidebar on 6261272576 ...
[1699875685.286] loading top on 6261272576 ...
[1699875685.660] done loading sidebar on 6278098944 ...
[1699875685.674] loading top on 6278098944 ...
[1699875690.289] done loading top on 6261272576 ...
[1699875690.305] loading bottom on 6261272576 ...
[1699875690.680] done loading top on 6278098944 ...
[...]

Issue #5819, opened 2023-11-06
`pn.state.onload` doesn't work with threads


armaaar:
#### ALL software version info
Python 3.10.13
Panel: 1.3.1
Bokeh: 3.3.0
Param: 2.0.0

#### Description of expected behavior and the observed behavior

By enabling multi-threading (using `pn.config.nthreads`), I expect differnet callbacks registered with `pn.state.onload` to run in parallel on different threads. But currently they all run on the main process only.

#### Complete, minimal, self-contained example code that reproduces the issue

```
import os
import threading
import time

import panel as pn
from panel.template import MaterialTemplate
from panel.viewable import Viewer


def slow_task():
    pid = os.getpid()
    tid = threading.get_ident()

    print(time.time(), f"{pid=}, {tid=} working slowly ...")
    time.sleep(5)
    print(time.time(), f"{pid=}, {tid=} done.")


class Page(Viewer):
    sidebar_content = pn.Column("Loading...")
    main_content = pn.Column("Loading...")

    def __init__(self, **params):
        super().__init__(**params)

        print("curdoc:", pn.state.curdoc)

        def sidebar_cb():
            print("Sidebar CB")
            slow_task()
            self.sidebar_content.clear()
            self.sidebar_content.append("# Sidebar")

        pn.state.onload(sidebar_cb)

        def main_cb():
            print("Main CB")
            slow_task()
            self.main_content.clear()
            self.main_content.append("# Main")

        pn.state.onload(main_cb)

    def __panel__(self):
        template = MaterialTemplate()

        template.sidebar.append(self.sidebar_content)
        template.main.append(self.main_content)

        return template


pn.config.nthreads = os.cpu_count()

print("nthreads:", pn.config.nthreads)
pn.serve(Page)


```

#### Stack traceback and/or browser JavaScript console output

```
> python issue.py
nthreads: 10
Launching server at http://localhost:53940
curdoc: <bokeh.document.document.Document object at 0x142cc9270>
Sidebar CB
1699268606.199587 pid=33641, tid=6225473536 working slowly ...
1699268611.2046762 pid=33641, tid=6225473536 done.
Main CB
1699268611.22032 pid=33641, tid=6225473536 working slowly ...
1699268616.2236311 pid=33641, tid=6225473536 done.
```

#### Screenshots or screencasts of the bug in action
[...]

PR #5865, opened 2023-11-13
Allow running onload and defer_load tasks on threads

philippjfr:
Currently `onload` callbacks and components that use `defer_load` run consecutively when in many cases they could be running in parallel on the configured thread pool. This PR does a few things to enable this:

- Allow `pn.state.execute(cb, schedule='thread')` as an option to schedule a callback on a thread
- Add support for `pn.state.onload(cb, threaded=True)` to register an onload callback that can be run on a thread 
- Components which have `defer_load=True` will automatically be scheduled on the thread pool (if available)

Tasks:

- [x] Adds support for https://github.com/holoviz/panel/issues/5862, https://github.com/holoviz/panel/issues/5819
- [x] Add tests
- [x] Add docs

### DIFF
--- a/panel/io/state.py
+++ b/panel/io/state.py
@@ -5,7 +5,6 @@ from __future__ import annotations
 
 import asyncio
 import datetime as dt
-import functools
 import inspect
 import logging
 import shutil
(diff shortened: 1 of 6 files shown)


panel/io/state.py
```diff
@@ -589,15 +597,17 @@ def execute(
...
+          or execute them immediately.
         """
-        cb = callback
-        while isinstance(cb, functools.partial):
-            cb = cb.func
         doc = self.curdoc
-        if param.parameterized.iscoroutinefunction(cb):
+        if schedule == 'thread':
+            if not state._thread_pool:
+                raise RuntimeError('Cannot execute callback on thread')
```
philippjfr, line 607:
```suggestion
                raise RuntimeError(
                    'Cannot execute callback on thread, set up thread pool by setting '
                    'config.num_threads.'
                )
```

MarcSkovMadsen:
Why id it nescessary to specific it. Could this not be automatic if  the threadpool is available?

philippjfr:
Does not seem safe, on_load callbacks are sometimes dependent on being run in a specific order so this would be a backward compatibility nightmare. I'm open to changing the behavior at some point but not without communicating the change well in advance.

philippjfr:
UI test failures unrelated and hopefully fixed in https://github.com/holoviz/panel/pull/5867

PR #5867, opened 2023-11-13
Stabilize test suite

Hoxbro:


panel/tests/ui/io/test_jupyter_server_extension.py
```diff
@@ -9,7 +9,7 @@
 pytestmark = [pytest.mark.jupyter, pytest.mark.ui]
 
 
-@pytest.mark.flaky(max_runs=3)
+@pytest.mark.flaky(reruns=3)
```
philippjfr, line 12: What's the difference?

Hoxbro: I am trying to change from `flaky` to `pytest-rerunfailures`. Because `flaky` hasn't been updated in 3 years and pytest-rerunfailures is under the [pytest-dev org](https://github.com/pytest-dev/pytest-rerunfailures). They share the same marker.

For this, it is only to align it with how flaky does it.  I wanted to add `@pytest.mark.flaky(reruns=3, reruns_delay=10)` to the convert test in [Mark convert test flaky](https://github.com/holoviz/panel/pull/5867/commits/deb16a029e4fd11f037eb03d096c3416f181f7c9).

philippjfr (approved):

philippjfr:
It's a big hammer indeed, but for now this seems like the best approach.

PR #5870, opened 2023-11-14
Allow image panes to be used for avatars

ahuang11:
```python
import panel as pn
pn.extension()

chat = pn.chat.ChatInterface(avatar=pn.pane.Image("https://panel.holoviz.org/_static/logo_horizontal.png"))
chat
```

### DIFF
--- a/examples/reference/chat/ChatFeed.ipynb
+++ b/examples/reference/chat/ChatFeed.ipynb
@@ -50,7 +50,7 @@
     "\n",
     "* **`header`** (Any): The header of the chat feed; commonly used for the title. Can be a string, pane, or widget.\n",
     "* **`callback_user`** (str): The default user name to use for the message provided by the callback.\n",
-    "* **`callback_avatar`** (str | BytesIO | pn.pane.ImageBase): The avatar to use for the user. Can be a single character text, an emoji, or anything supported by `pn.pane.Image`. If not set, uses the first character of the name.\n",
+    "* **`callback_avatar`** (str | bytes | BytesIO | pn.pane.ImageBase): The avatar to use for the user. Can be a single character text, an emoji, or anything supported by `pn.pane.Image`. If not set, uses the first character of the name.\n",
     "* **`placeholder_text`** (any): If placeholder is the default LoadingSpinner, the text to display next to it.\n",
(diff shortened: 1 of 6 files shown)


AlbertDeFusco:
Maybe this should be `ImageBase` like it is for Feed?

https://github.com/holoviz/panel/blob/34df686ba2ed3b3efbeb2b97999981bd7cf86ffb/panel/chat/message.py#L39

ahuang11:
Thanks @AlbertDeFusco for reviewing. I just removed it since it's not used anywhere and param cannot accept `typing` types.

panel/chat/feed.py
```diff
@@ -461,7 +458,7 @@ def send(
         self,
         value: ChatMessage | dict | Any,
         user: str | None = None,
-        avatar: str | BinaryIO | None = None,
+        avatar: str | BytesIO | None = None,
```
AlbertDeFusco, line 461: should all of these type declarations include `bytes` as well?

ahuang11: Yup! Thanks for catching that. I just pushed another commit

examples/reference/chat/ChatFeed.ipynb
```diff
@@ -50,7 +50,7 @@
     "\n",
     "* **`header`** (Any): The header of the chat feed; commonly used for the title. Can be a string, pane, or widget.\n",
     "* **`callback_user`** (str): The default user name to use for the message provided by the callback.\n",
-    "* **`callback_avatar`** (str | BinaryIO | pn.pane.ImageBase): The avatar to use for the user. Can be a single character text, an emoji, or anything supported by `pn.pane.Image`. If not set, uses the first character of the name.\n",
+    "* **`callback_avatar`** (str | BytesIO | pn.pane.ImageBase): The avatar to use for the user. Can be a single character text, an emoji, or anything supported by `pn.pane.Image`. If not set, uses the first character of the name.\n",
```
ahuang11, line 53:
```suggestion
    "* **`callback_avatar`** (str | bytes | BytesIO | pn.pane.ImageBase): The avatar to use for the user. Can be a single character text, an emoji, or anything supported by `pn.pane.Image`. If not set, uses the first character of the name.\n",
```

AlbertDeFusco (approved):
This looks correct to me!

philippjfr (approved):
Looks great, thanks for providing review @AlbertDeFusco !

PR #5875, opened 2023-11-15
Fix fast FileDownload embed=True style

### LINKED ISSUE #5874: FileDownloadButton with embed=True not styled
I'm on the latest Panel branch using edge browser. I you use the `FileDownload` and set `embed=True` the original `color` is used. Not the `fast` one.

```python
import panel as pn

pn.extension(template="fast")

layout =pn.Row(
    *(pn.Column(*(pn.widgets.FileDownload("script.py", embed=True, button_type=p, button_style=bs) for p in pn.widgets.Button.param.button_type.objects))
    for bs in pn.widgets.Button.param.button_style.objects)
)

layout.servable()
```

image

If you set `embed=False` they look like

image

MarcSkovMadsen:
Closes #5874

The example from the issue looks like

image

### DIFF
--- a/panel/theme/css/fast.css
+++ b/panel/theme/css/fast.css
@@ -473,6 +473,11 @@ input[type='file']:active {
   );
 }
 
+.bk-btn a {
+  text-decoration: none;
+  color: inherit;
+}
+
 :host(.solid) .bk-btn.bk-btn-default {
   background-color: var(--neutral-fill-rest);
   color: var(--neutral-foreground-rest);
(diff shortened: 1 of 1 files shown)


MarcSkovMadsen:
The failing tests are ui tests of the chat interface that time out. I don't think its related.

Hoxbro:
> The failing tests are ui tests of the chat interface that time out. I don't think its related.

It is unrelated, but the tests are in the _unit_ test, not the UI test. I thought I fixed these yesterday. There could be a more underlying issue with the chat components. But all of this has nothing to do with this PR. @ ahuang11, can you look at why the chat tests are flaky and open a new issue / PR with your findings?

philippjfr (approved):

PR #5880, opened 2023-11-15
Only set HoloViews backend in first extension call

philippjfr:
`hv.Store.set_current_backend` can be a fairly expensive operation (around 40ms) and does not have to be run each time. So unless the backend does not match the current backend we do not rerun it each time.

### DIFF
--- a/panel/config.py
+++ b/panel/config.py
@@ -798,7 +798,7 @@ class panel_extension(_pyviz_extension):
                 backend = hv.Store.current_backend
             else:
                 backend = 'bokeh'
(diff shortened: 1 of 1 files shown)


panel/config.py
```diff
@@ -794,7 +794,7 @@ def __call__(self, *args, **params):
                 backend = hv.Store.current_backend
             else:
                 backend = 'bokeh'
-            if hasattr(hv.Store, 'set_current_backend'):
+            if not loaded and backend != hv.Store.current_backend and hasattr(hv.Store, 'set_current_backend'):
```
Hoxbro, line 797: Shouldn't this be used for both parts of the if statements?

PR #5894, opened 2023-11-19
Use watchfiles library for detecting changed files

philippjfr:
Using our own periodic callback handler sucked and turned out to be somewhat slow/unreliable. The [watchfiles library](https://github.com/samuelcolvin/watchfiles) on the other hand works exceptionally well, is very fast (since it is written in Rust) and simplifies our code. I think the dependency is worth it on balance, particularly since I have another usecase in mind.

@MarcSkovMadsen You may want to experiment.

### DIFF
--- a/panel/io/reload.py
+++ b/panel/io/reload.py
@@ -1,12 +1,12 @@
 import asyncio
 import fnmatch
+import logging
 import os
 import sys
 import types
 import warnings
 
 from contextlib import contextmanager
-from functools import partial
 
 try:
     from watchfiles import awatch
@@ -26,6 +26,8 @@ except Exception:
 
 from ..util import fullpath
 from .state import state
+
+_reload_logger = logging.getLogger('panel.io.reload')
 
 _watched_files = set()
 _modules = set()
@@ -81,7 +83,7 @@ def file_is_in_folder_glob(filepath, folderpath_glob):
     file_dir = os.path.dirname(filepath) + "/"
     return fnmatch.fnmatch(file_dir, folderpath_glob)
 
-async def async_file_watcher(curdoc):
+async def async_file_watcher(stop_event=None):
     files = list(_watched_files)
     modules = {}
     for module_name in _modules:
@@ -102,25 +104,15 @@ async def async_file_watcher(curdoc):
         modules[path] = module_name
         files.append(path)
 
-    stop_event = asyncio.Event()
-
-    async def stop_on_destroy():
-        await asyncio.sleep(1)
-        if not curdoc.session_context:
-            stop_event.set()
-
-    # Keep handle on task to avoid weakref from being collected
-    stop_on_task = asyncio.create_task(stop_on_destroy())  # noqa
-
     async for changes in awatch(*files, stop_event=stop_event):
         for _, path in changes:
             if path in modules:
                 module = modules[path]
                 if module in sys.modules:
                     del sys.modules[module]
-        _reload()
+        _reload(changes)
 
-def setup_autoreload_watcher():
+async def setup_autoreload_watcher(stop_event=None):
     """
     Installs a periodic callback which checks for changes in watched
     files and sys.modules.
@@ -134,9 +126,8 @@ def setup_autoreload_watcher():
             'watchfiles being installed. Since it provides a much better '
             'user experience consider installing it today.', FutureWarning
         )
-    if not state.curdoc or not state.curdoc.session_context.server_context:
-        return
-    state.execute(partial(async_file_watcher, state.curdoc))
+    _reload_logger.debug('Setting up global autoreload watcher.')
+    await async_file_watcher(stop_event=stop_event)
 
 def watch(filename):
     """
@@ -179,12 +170,17 @@ def record_modules():
         except Exception:
             continue
 
-def _reload():
-    if state.location is not None:
-        # In case session has been cleaned up
-        state.location.reload = True
-    for loc in state._locations.values():
-        loc.reload = True
+def _reload(changes):
+    _reload_logger.debug('Changes detected by autoreload watcher, reloading sessions.')
+    for doc, loc in state._locations.items():
+        if not doc.session_context:
+            continue
+        elif state._loaded.get(doc):
+            loc.reload = True
+            continue
+        def reload_session(event):
+            loc.reload = True
+        doc.on_event('document_ready', reload_session)
 
 def _check_file(path, modify_times):
     """
--- a/panel/io/reload.py
+++ b/panel/io/reload.py
@@ -124,7 +124,8 @@ async def setup_autoreload_watcher(stop_event=None):
             '--autoreload functionality now depends on the watchfiles '
             'library. In future versions autoreload will not work without '
             'watchfiles being installed. Since it provides a much better '
-            'user experience consider installing it today.', FutureWarning
+            'user experience consider installing it today.', FutureWarning,
+            stacklevel=0
         )
     _reload_logger.debug('Setting up global autoreload watcher.')
     await async_file_watcher(stop_event=stop_event)
@@ -178,7 +179,7 @@ def _reload(changes):
         elif state._loaded.get(doc):
             loc.reload = True
             continue
-        def reload_session(event):
+        def reload_session(event, loc=loc):
             loc.reload = True
         doc.on_event('document_ready', reload_session)
 
--- a/panel/io/reload.py
+++ b/panel/io/reload.py
@@ -10,7 +10,6 @@ from functools import partial
 
 try:
     from watchfiles import awatch
-    raise ValueError
 except Exception:
     async def awatch(*files, stop_event=None):
         modify_times = {}
@@ -18,7 +17,7 @@ except Exception:
         while not stop_event.is_set():
             changes = set()
             for path in files:
-                change = _check_file(modify_times, path)
+                change = _check_file(path, modify_times)
                 if change:
                     changes.add((change, path))
             if changes:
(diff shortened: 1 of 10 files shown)


MarcSkovMadsen:
There are some other issues we should try to solve as well.

- If the code has an error already when the server is started a blank screen is (always)? shown to the user and the reload is not configured. This is friction.
- My experience is that sometimes later on the reload can also get reconnected. This is also friction.

Maybe they are already solved?

MarcSkovMadsen:
This is such a nice DevEx improvement

## Works on Linux Jupyterhub

https://github.com/holoviz/panel/assets/42288570/8893e0b5-2e24-4e2d-8c9b-cbdb36e736ec

MarcSkovMadsen:
## Issue: Does not pick up change in other file.

If I import from another file it does not pick up the change. It only reloads.

https://github.com/holoviz/panel/assets/42288570/af0ba4ef-5437-4841-ae76-b576e9c5c687

```python
import panel as pn
import pandas as pd
import datetime as dt
from other import spacer

pn.extension("tabulator", design="material", css_files=["https://cdnjs.cloudflare.com/ajax/libs/font-awesome/6.4.2/css/all.min.css"])
df = pd.DataFrame({
    'int': [1, 2, 34],
    'float': [3.14, 6.28, 9.42],
    'str': ['A', 'B', 'C'],
    'bool': [True, False, True],
    'date': [dt.date(2019, 1, 1), dt.date(2020, 1, 1), dt.date(2020, 1, 10)],
    'datetime': [dt.datetime(2019, 1, 1, 10), dt.datetime(2020, 1, 1, 12), dt.datetime(2020, 1, 10, 13)]
}, index=[1, 2, 3])

button_table = pn.widgets.Tabulator(df, buttons={
    'print': '<i class="fa fa-print"></i>',
    'check': '<i class="fa fa-check"></i>'
})

pn.Column(button_table, spacer).servable()
```

```python
import panel as pn

spacer = pn.Spacer(height=200, width=200, styles={"background": "salmon"})
```

MarcSkovMadsen:
## Issue: Does not pick up change in library installed with for editing

Most times I'm working on a package which contains one or more apps. I my package is installed with `pip install -e .`. Its very unproductive if the Panel autoreload does not pick up changes in my package.

If I try this with Panel it does not seem to work. It would expect the page to reload when I hard code the Tabulator width.

https://github.com/holoviz/panel/assets/42288570/95594140-a0c4-4d9c-8d1a-f58d75be7ac2

philippjfr:
The logic was a little off, so the above issues should be fixed now.

Hoxbro:
I will be somewhat careful with adding `watchfiles` as a required dependency, as it is not absolutely needed for running Panel.

Also, as it is a rust app, it can mean we can't support new minor versions of Python before a new app release has been compiled with the ABI. As Panel has become a dependency of many of our other packages, it will be harder to test other packages.

It will also take time before it is published on the `main` / `default` conda channel. As far as I can see [here](https://anaconda.org/main/watchfiles/files), the latest version only supports up to Python 3.10.

Lastly (though very minor). This adds an extra 1.2 MB, which is unnecessary for production apps in a Docker container or running an app in Pyoidide.  

> another usecase in mind

My points above can change depending on the other use cases.

MarcSkovMadsen:
Good considerations @Hoxbro . Maybe its worth considering whether the built in autoreload can be kept but watchfiles used if it is installed?

philippjfr:
Thanks @Hoxbro, we should definitely carefully evaluate all new dependencies. Overall none of that concerns me overly much.

>Also, as it is a rust app, it can mean we can't support new minor versions of Python before a new app release has been compiled with the ABI. 

For PyPI this doesn't seem a major issue, `watchfiles` is a very popular package with 250k downloads a day and is used by various major libraries and it's used heavily at Pydantic (the company) so I trust Samuel to stay on top of it. 

>It will also take time before it is published on the main / default conda channel. As far as I can see here, the latest version only supports up to Python 3.10.

We have control over this, we can ask for the recipe to be updated if/when we needed so this doesn't concern me very much. In fact Anaconda in general should make sure that we update Samuel's projects in a timely manner because he has a strong voice in the community. It is true that this is yet another thing for us to track though.

>Lastly (though very minor). This adds an extra 1.2 MB, which is unnecessary for production apps in a Docker container or running an app in Pyoidide.

1.2 MB is a non-issue from my perspective.

On balance I'd rather offer users a uniform experience but let's mull it over a little bit.

MarcSkovMadsen:
Is it possible to White or blacklist dependencies for Pyodide?

setup.py
```diff
@@ -113,6 +113,7 @@ def run(self):
     'bleach',
     'typing_extensions',
     'pandas >=1.2',
+    'watchfiles'
```
maximlt, line 116: Add a min pin maybe? This avoids solvers picking up an old version we never tested against.

samuelcolvin:
Hi, author of watchfiles here, pleased to hear you're thinking of using watchfiles.

Watchfiles is on conda forge as well as anaconda proper, bit it looks like the pr for the most recent release is failing, https://github.com/conda-forge/watchfiles-feedstock/pull/23 I'll try to get this merged asap.

As a company we use watchfiles (via uvicorn) ever day, so we'll definitely continue to maintain the package. It's pretty stable so all it needs is updates for new versions of python.

philippjfr:
Thanks @samuelcolvin, really appreciate you chiming in. Really appreciate all your work. I'll start the process of getting `watchfiles` updated on conda defaults independently of any decision we make here.

philippjfr:
Quick update here, thanks to the great conda-forge folks watchiles 0.21 is now uploaded there and I was able to get an update on conda defaults merged as well. For now I will make watchfiles optional and issue a FutureWarning to indicate that eventually we will drop the old autoreload code and rely entirely on `watchfiles`.

setup.py
```diff
@@ -113,6 +113,7 @@ def run(self):
     'bleach',
     'typing_extensions',
     'pandas >=1.2',
+    'watchfiles >=0.20'
```
maximlt, line 116: It's not yet optional in the setup.py.

philippjfr: Sheesh, give me a minute here 😉

maximlt:
I looked a little on Github and `watchfiles` is definitely a direct dependency of a number of frameworks. As Panel users are used to install extras to get some additional functionalities (probably because we rely on conda a lot and conda doesn't the exact same feature and setting something equivalent is a bit of a pain), I agree a warning is sensible right now and hope we can soon make `watchfiles` a direct dependency.

panel/io/reload.py
```diff
@@ -117,48 +179,24 @@ def record_modules():
...
-def _check_file(modify_times, path, module=None):
+def _check_file(modify_times, path):
     try:
         modified = os.stat(path).st_mtime
+    except FileNotFoundError:
+        if path in modify_times:
+            return 3
     except Exception:
-        return
+        return 0
```
Hoxbro, line 196: Can you make a comment or reference to what the different return numbers mean?

panel/io/reload.py
```diff
@@ -65,16 +82,61 @@ def file_is_in_folder_glob(filepath, folderpath_glob):
...
+        # module.
+        if module_name not in sys.modules:
+            continue
+        module = sys.modules[module_name]
+        if not isinstance(module, types.ModuleType):
+            continue
+        path = getattr(module, "__file__", None)
+        if not path:
+            continue
+        if path.endswith(".pyc") or path.endswith(".pyo"):
```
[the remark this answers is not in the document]
philippjfr: Will be resolved by rebasing against main.

panel/io/reload.py
```diff
@@ -65,16 +82,61 @@ def file_is_in_folder_glob(filepath, folderpath_glob):
...
+        files.append(path)
+
+    stop_event = asyncio.Event()
+
+    async def stop_on_destroy():
+        await asyncio.sleep(1)
+        if not curdoc.session_context:
+            stop_event.set()
+
+    stop_on_task = asyncio.create_task(stop_on_destroy())  # noqa
```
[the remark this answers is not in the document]
philippjfr: Indeed, I'll leave a comment.

panel/io/reload.py
philippjfr: Will do, effectively this is just roughly mirroring watchfiles codes:

- 0: No change
- 2: File modified
- 3: File deleted

philippjfr:
pre-commit.ci autofix

philippjfr:
Took me weeks to figure out but I finally realized that when you shut down a server you briefly have to restart the event loop so that watchfiles can correctly catch the stop event and shut down the thread, otherwise you end up with a zombie thread. This should now finally be ready to merge.

MarcSkovMadsen:
I noticed on windows with panel==1.3.6 the jupyter panel preview is really slow to reload. I hope this PR solves the issue.

https://github.com/holoviz/panel/assets/42288570/f7df4e34-0df8-4a74-b3a1-74999bb394c8

MarcSkovMadsen:
Its actually quite slow on linux too.


https://github.com/holoviz/panel/assets/42288570/107bf9e5-dbd2-4858-a7b0-e16abe94a604

(I don't remember it like that ??)

philippjfr:
It's been relatively slow for a while. To make it robust we have to start a new kernel for each session. We might be able to look into reusing a kernel for reloads though.

MarcSkovMadsen:
Could you use autoreload by default but add a button to restart the kernel if something goes wrong?

philippjfr:
Will have to look into what's feasible, but note that this is entirely separate from "autoreload" functionality. The autoreload is triggered directly by JupyterLab file content manager events.

philippjfr:
Seems very likely that this is possible, could you open an issue?

PR #5896, opened 2023-11-20
Add Panel Modal

MarcSkovMadsen:
This PR migrates the Panel-Modal to Panel as agreed with Philipp. Having a standalone modal provides more flexibility than the modal built into templates. For example it also works in notebooks. The user experience is also better.

### DIFF
--- a/panel/layout/modal.py
+++ b/panel/layout/modal.py
@@ -1,0 +1,187 @@
+"""The `Modal` is an element that displays in front of and deactivates all other page content."""
+import param
+
+from ..reactive import ReactiveHTML
+from .base import NamedListLike
+
+# See https://a11y-dialog.netlify.app/
+JS_FILE = "https://cdn.jsdelivr.net/npm/a11y-dialog@7/dist/a11y-dialog.min.js"
+
+STYLE = """
+.dialog-container,
+.dialog-overlay {
+  position: fixed;
+  top: 0;
+  right: 0;
+  bottom: 0;
+  left: 0;
+}
+.dialog-container {
+  z-index: 100002;
+  display: flex;
+}
+.dialog-container[aria-hidden='true'] {
+  display: none;
+}
+.dialog-overlay {
+  z-index: 100001;
+  background-color: rgb(43 46 56 / 0.9);
+}
+.dialog-content {
+  margin: auto;
+  z-index: 100002;
+  position: relative;
+  background-color: white;
+  border-radius: 2px;
+  padding: 10px;
+  padding-bottom: 20px;
+}
+fast-design-system-provider .dialog-content {
+  background-color: var(--background-color);
+  border-radius:  calc(var(--corner-radius) * 1px);
+}
+@keyframes fade-in {
+  from {
+    opacity: 0;
+  }
+}
+@keyframes slide-up {
+  from {
+    transform: translateY(10%);
+  }
+}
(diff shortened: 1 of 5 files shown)


panel/layout/modal.py
```diff
@@ -0,0 +1,187 @@
+"""The `Modal` is an element that displays in front of and deactivates all other page content."""
+import param
+
+from ..reactive import ReactiveHTML
+from .base import NamedListLike
+
+# See https://a11y-dialog.netlify.app/
+JS_FILE = "https://cdn.jsdelivr.net/npm/a11y-dialog@7/dist/a11y-dialog.min.js"
```
MarcSkovMadsen, line 8: - [ ] I should try upgrading to v8.

panel/layout/modal.py
```diff
@@ -0,0 +1,187 @@
...
+    def _hide(self):
+        self.is_open = False
+
+    _extension_name = "modal"
+
+    __javascript__ = [JS_FILE]
+
+
+    _template = """
+<style id="pnx_dialog_style">
```
MarcSkovMadsen, line 145: I should maybe find another way to include the style. The problem is that with the fast templates this style needs to be available globally while with other templates it does not have to.

panel/layout/modal.py
```diff
@@ -0,0 +1,187 @@
...
+    >>> pn.extension("modal")
+    >>> modal = Modal(pn.panel("Hi. I am the Panel Modal!", width=200))
+    >>> pn.Column(modal.param.open, modal).servable()
+    """
+
+    is_open = param.Boolean(doc="""
+        Whether or not the modal is open. Set to True to open. Set to False to close.""")
+    show_close_button = param.Boolean(True, doc="Whether to show a close button in the modal")
+
+    open = param.Event(doc="Click here to open the modal")
```
MarcSkovMadsen, line 120: - [ ] I would like to have actions instead of events as they are more natural to use from Python. But when I do, I get serialization error.

panel/layout/modal.py
```diff
@@ -0,0 +1,187 @@
...
+}
+fast-design-system-provider .pnx-dialog-close:hover {
+  background-color: var(--neutral-fill-hover);
+}
+.lm-Widget.p-Widget.lm-TabBar.p-TabBar.lm-DockPanel-tabBar.jp-Activity {
+  z-index: -1;
+}
+"""
+
+class Modal(ReactiveHTML, NamedListLike):  # pylint: disable=too-many-ancestors
```
MarcSkovMadsen, line 95: - [ ] Should figure out if its possible to add parameter to disable the closing of the modal when clicking the back drop.

panel/layout/modal.py
MarcSkovMadsen: With v8 microsoft explorer is no longer supported. Do we support microsoft explorer @philippjfr ?

panel/layout/modal.py
philippjfr: Can you expand on this? We explicitly introduced Event because `Action` seemed weird and hacky.

MarcSkovMadsen: Because its more natural truffet an action from the Python side as a function.

MarcSkovMadsen:
Superseeded by @hoxbro in https://github.com/holoviz/panel/pull/7083

PR #5911, opened 2023-11-22
Add support for rendering inside PyScript Next

philippjfr:
For PyScript Next support we need to do a few things:

1. Detect if we are in a worker and then patch the `js.document` and `js.Bokeh` objects
2. Ensure that we act like we are on the main thread even though we are on the worker (since PyScript now transparently handles message passing between the two).

### DIFF
--- a/panel/io/pyodide.py
+++ b/panel/io/pyodide.py
@@ -14,6 +14,7 @@ from typing import (
 )
 
 import bokeh
+import js
 import param
 
 import pyodide # isort: split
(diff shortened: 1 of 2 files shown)


MarcSkovMadsen:
I believe the template also needs an update. You can use `<pyscript>` but it should be `<script type="py">` etc. now.

philippjfr:
Indeed, that'll land in 1.4.0. Wanted to get this into 1.3.2 since it was such a tiny fix. It also relies on absolute bleeding edge of pyscript core,  since we had to resolve some issues that were uncovered by the Panel Python <-> JS message passing dance.

PR #5912, opened 2023-11-22
Allow passing partial function to tabulator filter

TBym:
It is possible to pass a function as a filter to Tabulator but passing a functools.partial function currently fails. This update allows to pass both regular and partial functions.


philippjfr (approved):

philippjfr:
Thanks for your contribution @TBym!

PR #5929, opened 2023-11-28
Persist select when `pagination=remote`

Hoxbro:
**TODO:**

- [ ] I need to handle shift key logic

**Example**

https://github.com/holoviz/panel/assets/19758978/40f7da16-93b9-40a1-b619-6df5913cc188

``` python
import numpy as np
import pandas as pd
import panel as pn

pn.extension("tabulator")

data = np.arange(20) + 100
df = pd.DataFrame(data)
tab = pn.widgets.Tabulator(
    value=df,
    disabled=True,
    pagination="remote",
    page_size=10,
    header_filters=True,
)
tab.servable()
```

### DIFF
--- a/panel/models/tabulator.ts
+++ b/panel/models/tabulator.ts
@@ -44,12 +44,12 @@ export class CellClickEvent extends ModelEvent {
 }
 
 export class SelectionEvent extends ModelEvent {
-  constructor(readonly indices: number[], readonly selected: boolean) {
+  constructor(readonly indices: number[], readonly selected: boolean, readonly flush: boolean = false) {
     super()
   }
 
   protected get event_values(): Attrs {
-    return {model: this.origin, indices: this.indices, selected: this.selected}
+    return {model: this.origin, indices: this.indices, selected: this.selected, flush: this.flush}
   }
 
   static {
@@ -1070,6 +1070,16 @@ export class DataTabulatorView extends HTMLBoxView {
     let indices: number[] = []
     const selected = this.model.source.selected
     const index: number = row._row.data._index
+
+    if (this.model.pagination === 'remote') {
+      let includes = this.model.source.selected.indices.indexOf(index) == -1
+      let flush = !(e.ctrlKey || e.metaKey)
+      this._selection_updating = true
+      this.model.trigger_event(new SelectionEvent([index], includes, flush))
+      this._selection_updating = false
+      return
+    }
+
     if (e.ctrlKey || e.metaKey) {
       indices = [...this.model.source.selected.indices]
     } else if (e.shiftKey && selected.indices.length) {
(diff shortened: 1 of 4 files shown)


panel/widgets/tables.py
```diff
@@ -1629,9 +1625,10 @@ def _update_selection(self, indices: List[int] | SelectionEvent):
                 continue
             if selected:
                 ilocs.append(iloc)
-            else:
+            elif iloc in ilocs:
                 ilocs.remove(iloc)
         self.selection = list(dict.fromkeys(ilocs))
+        self.param.trigger('selection')
```
Hoxbro, line 1631: I'm not entirely sure why this is needed to trigger the update.

philippjfr: It shouldn't be unless the selection hasn't actually changed or something modified it in-place meaning that it can't detect the change.

Hoxbro: I was adding and removing from the `self.selection` list directly and therefore is the same on both sides of the equal sign on the line above.

I now copy the list of `self.selection` before adding/removing it so it registers a change.

philippjfr: 👍

panel/models/tabulator.ts
```diff
@@ -1070,6 +1070,28 @@ export class DataTabulatorView extends HTMLBoxView {
...
     const selected = this.model.source.selected
     const index: number = row._row.data._index
+
+    if (this.model.pagination === 'remote') {
+      let includes = this.model.source.selected.indices.indexOf(index) == -1
+      let flush = !(e.ctrlKey || e.metaKey || e.shiftKey)
+      if (e.shiftKey && selected.indices.length) {
+        const start = selected.indices[selected.indices.length-1]
+        if (index>start) {
+          for (let i = start; i<=index; i++)
```
Hoxbro, line 1080: In the non-remote logic, this is done with `for (let i = start; i<index; i++)`. I feel like it makes more sense to go up to the selected index and not one above/below it.

Hoxbro:
Shift behavior differs a bit for `pagination="local"` as it only looks at the current page.


https://github.com/holoviz/panel/assets/19758978/86a13e0e-0526-452a-9f66-aff6b1cb1164

Hoxbro:
Also implemented logic if selectable is an integer like `selectable=2`.


https://github.com/holoviz/panel/assets/19758978/10e4702f-c455-4faf-bac2-80282d63d92e

panel/models/tabulator.ts
philippjfr: Indeed, was that a bug?

Hoxbro: I think it is. It was added in #2230.

Hoxbro: I have updated it also to be the case for the non-remote option.

Hoxbro: It was not a bug, have reverted the change. The index is added in the following if statement:

https://github.com/holoviz/panel/blob/078591fda4150421c1601217ae54a9b4d321cf4f/panel/models/tabulator.ts#L1107-L1109

philippjfr (approved):

Issue #5897, opened 2023-11-20
Logging in with Generic Provider in V1.3+


simonaay:
Hi friends,

After unifying the access token in latest versions in Panel 1.3+, I started encountering problems with logging in with the Generic Provider feature. I'm not sure if I got to the bottom of this correctly, but what I noticed that the id_token does not contain all user info part of which is a key I pass to USER_KEY when I setup the CLI command. Keep in mind in any version of 1.0 to 1.2.3, the webapp works completely fine and successfully logs in. Would appreciate the help on this matter.

philippjfr:
Would love a little bit more detail on this part:

>what I noticed that the id_token does not contain all user info part of which is a key I pass to USER_KEY when I setup the CLI command

What is contained and what isn't? What precisely isn't working? What are the error messaes?

philippjfr:
Thanks! Could it be related to the scopes? They changed slightly and those partially determine the information available in the user token.

philippjfr:
@simonaay can you maybe list the keys that **are** present in the id_token?

simonaay:
Hi Philip,

Unfortunately, I can't reach my laptop for today. I'll see what I can share tomorrow. On another note, for security reasons I don't I can share everything. I'll see what I can do.

simonaay:
> Thanks! Could it be related to the scopes? They changed slightly and those partially determine the information available in the `id_token`.

I usually use the default parameters of scope, do you think I should reinstate them in `--oauth-extra-params`? In older versions, they were openid and email, right?

simonaay:
Quick updates.

I tried modifying in the `scope` key-value in `--oauth-extra-params` and then modifying directly in the source code. I tried multiple ways, initially I used to rely on the older version's default values, namely "openid" and "email". Following that logic, I kept passing several properties such as offline_access, UsersInfo and Identities. Result: internal server error similar to the aforementioned error.  Example below:

`   @property
    def _SCOPE(self):
        if 'scope' in config.oauth_extra_params:
            return config.oauth_extra_params['scope']
        elif 'PANEL_OAUTH_SCOPE' not in os.environ:
            return ['openid', 'email']
        return [scope for scope in os.environ['PANEL_OAUTH_SCOPE'].split(',')]`

simonaay:
Good morning @philippjfr , let me know if I can help in any way.

simonaay:
[...]

PR #5939, opened 2023-11-29
Ensure OAuth falls back to OpenID userinfo if id_token is not sufficient

philippjfr:
Before refactoring the OAuth handlers we would always either obtain our user info via the id_token **OR** via the OpenID userinfo endpoint. After the refactor we ended up skipping the request to the userinfo endpoint if a valid `id_token` was returned by the token endpoint. However in many cases that `id_token` does not return all the info needed to verify the user. Therefore we now check whether the `id_token` contains the required user key and if it doesn't we fall back to making the request to the userinfo endpoint.

Fixes https://github.com/holoviz/panel/issues/5897

### DIFF
--- a/panel/auth.py
+++ b/panel/auth.py
@@ -242,7 +242,7 @@ class OAuthLoginHandler(tornado.web.RequestHandler, OAuth2Mixin):
             except HTTPError:
                 pass
             else:
-                log.debug("%s successfully obtained tokens.", type(self).__name__)
+                log.debug("%s successfully obtained access_token and id_token.", type(self).__name__)
                 return user, access_token, refresh_token, expires_in
 
         user_headers = dict(self._API_BASE_HEADERS)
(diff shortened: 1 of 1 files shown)


panel/auth.py
```diff
@@ -237,9 +237,13 @@ async def _fetch_access_token(
...
         if id_token:= body.get('id_token'):
-            log.debug("%s successfully obtained tokens.", type(self).__name__)
-            user = self._on_auth(id_token, access_token, refresh_token, expires_in)
-            return user, access_token, refresh_token, expires_in
+            try:
+                user = self._on_auth(id_token, access_token, refresh_token, expires_in)
+            except HTTPError:
+                pass
+            else:
+                log.debug("%s successfully obtained tokens.", type(self).__name__)
```
philippjfr, line 245:
```suggestion
                log.debug("%s successfully obtained access_token and id_token.", type(self).__name__)
```

PR #5944, opened 2023-11-29
Remove print weasel

cdeil:


philippjfr (approved):
Good catch, shouldn't have rushed that release out yesterday.

PR #5946, opened 2023-11-29
Add a few missing asserts in tests

cdeil:
Are those missing asserts? PyCharm pointed them out as statements without effect.


ahuang11:
Yes! Those are missing asserts.

ahuang11 (approved):

PR #5947, opened 2023-11-29
Remove duplicate property VizzuChart.config

cdeil:
This looks like a duplicate property? It's already defined on line 44.


philippjfr (approved):
Good catch, thank you @cdeil!

philippjfr:
I'll have to find a way to skip those OAuth tests for external contributors since the secrets aren't defined in that case.

PR #5952, opened 2023-11-30
Ruff bugbear

Hoxbro:
Implements the Bugbear rules https://docs.astral.sh/ruff/rules/#flake8-bugbear-b

I'm currently unsure about if the followings errors are bugs or things to ignore:

``` python-traceback
panel/config.py:910:17: B020 Loop control variable `cls` overrides iterable it iterates
panel/io/state.py:972:17: B012 `return` inside `finally` blocks cause exceptions to be silenced
panel/links.py:582:23: B007 Loop control variable `code` not used within loop body
panel/pane/base.py:274:21: B007 Loop control variable `node` not used within loop body
panel/reactive.py:145:5: B019 Use of `functools.lru_cache` or `functools.cache` on methods can lead to memory leaks
panel/widgets/misc.py:234:13: B007 Loop control variable `mtype` not used within loop body
```

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -42,6 +42,7 @@ ignore = [
     "E701", # Multiple statements on one line
 ]
 select = [
+    "B",
     "E",
     "F",
     "W",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -41,6 +41,7 @@ ignore = [
     "W605",
     "E701", # Multiple statements on one line
     "B006", # Do not use mutable data structures for argument defaults
+    "B019", # lru_cache and cache are used deliberately
 ]
 select = [
     "B",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -40,6 +40,7 @@ ignore = [
     "E741",
     "W605",
     "E701", # Multiple statements on one line
+    "B006", # Do not use mutable data structures for argument defaults
 ]
 select = [
     "B",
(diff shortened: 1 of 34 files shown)


Hoxbro:
Fixing B023 in bb438e7 is because of [late-binding-closures](https://docs.python-guide.org/writing/gotchas/#late-binding-closures)

pyproject.toml
```diff
@@ -41,6 +41,7 @@ ignore = [
     "W605",
     "E701", # Multiple statements on one line
     "B006", # Do not use mutable data structures for argument defaults
+    "B019", # lru_cache and cache are used deliberately
```
Hoxbro, line 44: Maybe add `noqa: B019` to the deliberate lines, such that a future wrong cache is not used.

philippjfr: It may be misplaced trust in myself or others, but I think I trust that developers will recognize that a cache leaks memory.

philippjfr (approved):

pyproject.toml
Hoxbro: I'm not as afraid of the memory leak (that is expected), but the incorrect use of it for method, like this, will print 3 times:

``` python
from functools import lru_cache

class T:

    @lru_cache
    def t(self, n):
        print("printing t")

T().t(10)
T().t(10)
T().t(10)
```

https://docs.astral.sh/ruff/rules/cached-instance-method/

philippjfr: Ah indeed, missed that somehow. Will open a new PR.

Issue #5958, opened 2023-11-30
ReactiveExpr - Styled DataFrame Example is broken


MarcSkovMadsen:
I'm working my way through the ReactiveExpr Reference Notebook. 

I'm on panel=1.3.4, param=2.0.1, bokeh=3.3.1, pandas=2.1.3 python=3.11 os=linux

When I interact with the styled dataframe example it displays/ prints a lot of text when I drag the slider. And the Tabulator table is not getting styled.

![image](https://github.com/holoviz/panel/assets/42288570/3977e60f-dec9-4e41-80e6-9c0a68ea67db)

https://github.com/holoviz/panel/assets/42288570/48892eca-7471-4700-a69e-98905a9aa446

MarcSkovMadsen:
When you run the reference notebook in pyodide in the docs, the styled dataframe example is not working either.

![image](https://github.com/holoviz/panel/assets/42288570/d8ad7f03-6b86-46a5-942d-01d210b4ddbb)

Issue #5948, opened 2023-11-30
default_layout of ReactiveExpr not working


MarcSkovMadsen:
I'm on panel 1.3.4 and param 2.0.1. I cannot get the `default_layout` of `pn.ReactiveExpr` working.

I would expect the bottom component to be organized in a Column. But its not

![image](https://github.com/holoviz/panel/assets/42288570/35bf8a5e-30b3-4738-9b6a-ba65f0396371)

I believe I've seen it working before. So I marked it as a regression.

```python
import panel as pn
pn.extension()

def model(n):
    return f"Hello! {n}*2 is {n*2}"

slider = pn.widgets.IntSlider(value=30, start=1, end=50, name="n")
pn.rx(model)(n=slider)

pn.ReactiveExpr(
    pn.rx(model)(n=slider),
    default_layout=pn.Column
)
```

MarcSkovMadsen:
Looking at the implementation. I can see that the `default_layout` is not used any where. I believe it should be used instead of `Row` below.

![image](https://github.com/holoviz/panel/assets/42288570/1e9d95e9-f039-4b2c-bc50-73bc6d3974e3)

The documentation says `default_layout` is supported.

![image](https://github.com/holoviz/panel/assets/42288570/5a69b7be-dd20-4b2a-997d-87d15f69ca05)

MarcSkovMadsen:
I can see that the `widget_location` can layout things the way I expected.

![image](https://github.com/holoviz/panel/assets/42288570/c317c878-d0fa-4b72-9b33-7e47b57cf730)

I think the problem is that the `default_layout` is documented. And it should not be.

I will close this one.

PR #5960, opened 2023-11-30
Improvements to the ReactiveExpr reference guide

MarcSkovMadsen:
I had problems understanding the new `ReactiveExpr` pane from the reference guide. Especially how to layout things.

During that process I discovered that one of the problems is that `default_layout` is mentioned in the `reference guide`. And it should not be. See https://github.com/holoviz/panel/issues/5948. This PR will be Closing #5948 .

I added more small examples to help the user understand what reactive expressions are and how powerful they can be.

I added in emojis. I know some people think its too much. But I just think they communicate better when you open up the notebook during a training session, a conference talk or if you record a video for social media. I also believe that users when studying the notebook feel its more fun and engaging to study. For the same reason I used the `material` design.

## Todo

- [ ] Fix the bug https://github.com/holoviz/panel/issues/5958. I don't plan to do this. I hope Phillip will.

### DIFF
--- a/examples/reference/panes/ReactiveExpr.ipynb
+++ b/examples/reference/panes/ReactiveExpr.ipynb
@@ -9,30 +9,36 @@
    "outputs": [],
    "source": [
     "import panel as pn\n",
-    "import param\n",
     "\n",
-    "pn.extension('tabulator')"
+    "pn.extension('tabulator', design=\"material\")"
    ]
   },
   {
    "cell_type": "markdown",
    "metadata": {},
    "source": [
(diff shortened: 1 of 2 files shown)


MarcSkovMadsen:
## Assets

Here is a little bit for the social media

https://github.com/holoviz/panel/assets/42288570/ec02967b-048a-4228-b986-e9b9a477e4a3

https://github.com/holoviz/panel/assets/42288570/2002a062-28fc-4513-ad31-cf1f3d90af60

https://github.com/holoviz/panel/assets/42288570/c5da8d10-fce7-49db-bc9d-3ac1f95e6883

philippjfr (approved):
Looks good!

Issue #5699, opened 2023-10-20
ChatMessage browser timezone support


cdeil:
Currently it seems that `panel.chat.ChatMessage` shows the `timestamp` by default in `HH:MM` format in UTC.

Probably the vast majority of users wants to see times in their browser timezone and this is confusing for them?

@ahuang11  - Maybe try to support this and consider making it the default?

PR #5961, opened 2023-11-30
Add support for timestamp timezones for chat message

ahuang11:
Closes https://github.com/holoviz/panel/issues/5699

### DIFF
--- a/panel/chat/message.py
+++ b/panel/chat/message.py
@@ -16,11 +16,12 @@ from textwrap import indent
 from typing import (
     TYPE_CHECKING, Any, ClassVar, Dict, Iterable, List, Union,
 )
+from zoneinfo import ZoneInfo
 
 import param
-import pytz
 
 from ..io.resources import CDN_DIST
+from ..io.state import state
 from ..layout import Column, Row
 from ..pane.base import PaneBase, panel as _panel
 from ..pane.image import (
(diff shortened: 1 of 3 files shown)


maximlt:
Does this implement the actual request to display the time in the browser timezone?

> Probably the vast majority of users wants to see times in their browser timezone and this is confusing for them?

ahuang11:
Okay implemented zoneinfo + pn.state.browser_info.timezone

panel/chat/message.py
```diff
@@ -18,6 +18,7 @@
 )
 
 import param
+import pytz
```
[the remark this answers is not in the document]
maximlt: Generally I don't think you want to rely on transitive dependencies, i.e. Panel would break if `pandas` removed one day `pytz` from their direct dependencies.

panel/tests/chat/test_message.py
```diff
@@ -6,6 +6,7 @@
 
 import pandas as pd
 import pytest
+import pytz
```
philippjfr, line 9: Can't we use `zoneinfo` here too?

ahuang11: Oops forgot to update.

philippjfr (approved):

ahuang11:
Thanks! Fix is in https://github.com/holoviz/panel/pull/6122 which I think can go into 1.3.7 release next week.

Issue #5943, opened 2023-11-29
Interrupt Button for ChatInterface


ahuang11:
Sometimes, the LLM writes a huge wall of text, irrelevant to the prompt. It'd be nice to interrupt it midway.

PR #5962, opened 2023-12-01
Allow stopping respond callbacks midway

ahuang11:
Closes https://github.com/holoviz/panel/issues/5943

Upon sending a message, if the callback is async, the Send button gets hidden and the Stop button appears. Clicking it will cancel the callback midway.

https://github.com/holoviz/panel/assets/15331990/9cf8957e-ffd1-4c22-9b61-c642d5da0e4b

Makes it so that sync tasks are submitted to a separate thread so it's cancellable.

https://github.com/holoviz/panel/assets/15331990/94df994e-5ddc-4594-abea-1941d7f39e0c

```python
import panel as pn
from openai import OpenAI

pn.extension()


    response = client.chat.completions.create(
        model="gpt-3.5-turbo",
        stream=True,
    )
    message = None
    for chunk in response:
        if part is not None:
            message = instance.stream(part, user="OpenAI", message=message)


client = OpenAI()
chat_interface.send(
)
chat_interface.stream("Hello")
chat_interface.show()
```

Also, I think I made it more performant for sync too.

### DIFF
--- a/panel/chat/feed.py
+++ b/panel/chat/feed.py
@@ -571,7 +571,7 @@ class ChatFeed(ListPanel):
 
     def stop(self) -> bool:
         """
-        Cancels the current callback task.
+        Cancels the current callback task if possible.
 
         Returns
         -------
--- a/panel/chat/feed.py
+++ b/panel/chat/feed.py
@@ -8,9 +8,8 @@ from __future__ import annotations
 import asyncio
 import traceback
 
-from inspect import (
-    isasyncgen, isasyncgenfunction, isawaitable, isgenerator,
-)
+from functools import partial
+from inspect import isasyncgen, isawaitable, isgenerator
 from io import BytesIO
 from typing import (
     TYPE_CHECKING, Any, Callable, ClassVar, Dict, List, Literal,
@@ -92,6 +91,10 @@ PLACEHOLDER_SVG = """
         <path d="M17 12a5 5 0 1 0 -5 5"></path>
     </svg>
 """  # noqa: E501
+
+
+class StopCallback(Exception):
+    pass
 
 
 class ChatFeed(ListPanel):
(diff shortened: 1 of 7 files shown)


panel/chat/feed.py
```diff
@@ -407,13 +412,10 @@ async def _schedule_placeholder(
         if self.placeholder_threshold == 0:
             return
 
-        callable_is_async = asyncio.iscoroutinefunction(
-            self.callback
-        ) or isasyncgenfunction(self.callback)
         start = asyncio.get_event_loop().time()
         while not task.done() and num_entries == len(self._chat_log):
             duration = asyncio.get_event_loop().time() - start
-            if duration > self.placeholder_threshold or not callable_is_async:
```
ahuang11, line 416: Since it's a separate thread, it's no longer needed to always show placeholder when sync.

panel/chat/feed.py
```diff
@@ -428,16 +430,31 @@ async def _prepare_response(self, _) -> None:
...
-            await self._schedule_placeholder(task, num_entries)
-            await task
-            task.result()
+            loop = asyncio.get_event_loop()
+            contents = self._extract_contents(message)
+
+            if asyncio.iscoroutinefunction(self.callback):
+                future = loop.create_task(self._wrap_async_callback(contents, message))
+            else:
+                future = loop.run_in_executor(None, partial(self.callback, contents, message.user, self))
```
philippjfr, line 448: I think in future I want this to go via `state.execute` but that needs to be rewritten to return a `Future` and to allow forcing threaded execution even when there isn't a ThreadPoolExecutor available.

philippjfr (approved):

philippjfr:
Still seeing some errors in the tests.

MarcSkovMadsen:
Has anyone concidered whether to focus on async via something like asyncer by Tiangolo. https://asyncer.tiangolo.com/

PR #5966, opened 2023-12-01
Split build jobs into seperate jobs

Hoxbro:
Split up the job Conda job into three separate jobs. One for Conda itself, one for NPM, and one for the CDN upload.

For these 3 I have added an upload artifact which should make it easy to inspect the build if needed. 

The NPM I have used `npm pack` to be able to save the tarball as an artifact, and then I use that tarball to upload with `npm upload`. 

Also took the time to clean up the build file.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -55,9 +55,53 @@ jobs:
       - name: conda build
         run: |
           bash ./scripts/build_conda.sh
-      - name: build pyodide wheels for CDN
+      - name: conda dev deploy
+        if: (github.event_name == 'push' && (contains(steps.vars.outputs.tag, 'a') || contains(steps.vars.outputs.tag, 'b') || contains(steps.vars.outputs.tag, 'rc')))
         run: |
-          python ./scripts/build_pyodide_wheels.py
+          VERSION="$(echo "$(ls dist/*.whl)" | cut -d- -f2)"
+          FILE="$CONDA_PREFIX/conda-bld/noarch/panel-$VERSION-py_0.tar.bz2"
+          anaconda --token ${{ secrets.CONDA_UPLOAD_TOKEN }} upload --user pyviz --label=dev $FILE
+      - name: conda main deploy
+        if: (github.event_name == 'push' && !(contains(steps.vars.outputs.tag, 'a') || contains(steps.vars.outputs.tag, 'b') || contains(steps.vars.outputs.tag, 'rc')))
+        run: |
+          VERSION="$(echo "$(ls dist/*.whl)" | cut -d- -f2)"
+          FILE="$CONDA_PREFIX/conda-bld/noarch/panel-$VERSION-py_0.tar.bz2"
+          anaconda --token ${{ secrets.CONDA_UPLOAD_TOKEN }} upload --user pyviz --label=dev --label=main $FILE
+
+  npm_build:
+    name: Build NPM Packages
+    runs-on: 'ubuntu-latest'
+    defaults:
+      run:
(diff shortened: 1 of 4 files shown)


scripts/build_pyodide_wheels.py
```diff
@@ -69,7 +69,7 @@
                 for line in buffer.decode("utf-8").split("\n")
                 if not (
                     "Requires-Dist:" in line
-                    and ("tornado" in line or "contourpy" in line)
```
Hoxbro, line 72: contourpy is now available on pyodide

Hoxbro:
Need to double-check if the wheel files are correctly uploaded to the cdn.

.github/workflows/build.yaml
```diff
@@ -52,12 +49,54 @@ jobs:
...
+          FILE="$CONDA_PREFIX/conda-bld/noarch/panel-$VERSION-py_0.tar.bz2"
+          anaconda --token ${{ secrets.CONDA_UPLOAD_TOKEN }} upload --user pyviz --label=dev $FILE
+      - name: conda main deploy
+        if: (github.event_name == 'push' && !(contains(steps.vars.outputs.tag, 'a') || contains(steps.vars.outputs.tag, 'b') || contains(steps.vars.outputs.tag, 'rc')))
+        run: |
+          VERSION="$(echo "$(ls dist/*.whl)" | cut -d- -f2)"
+          FILE="$CONDA_PREFIX/conda-bld/noarch/panel-$VERSION-py_0.tar.bz2"
+          anaconda --token ${{ secrets.CONDA_UPLOAD_TOKEN }} upload --user pyviz --label=dev --label=main $FILE
+
+  npm_build:
```
philippjfr, line 75: I would like this to at least be dependent on either the conda or the pip build passing.

.github/workflows/build.yaml
```diff
@@ -67,50 +106,83 @@ jobs:
...
-          anaconda --token ${{ secrets.CONDA_UPLOAD_TOKEN }} upload --user pyviz --label=dev --label=main $FILE
       - name: npm main deploy
         if: (github.event_name == 'push' && !(contains(steps.vars.outputs.tag, 'a') || contains(steps.vars.outputs.tag, 'b') || contains(steps.vars.outputs.tag, 'rc')))
         run: |
           cd ./panel
-          npm publish --tag latest
+          npm publish --tag latest $TARBALL
           cd ..
+
+  cdn:
```
philippjfr, line 132: Same with this.

.github/workflows/build.yaml
Hoxbro: Was thinking about having deploy/publish step separated for the build, which needs approval and requires all the building steps to pass, so we can ensure that everything is ready before publishing. 

Inspiration for this could be https://timheuer.com/blog/add-approval-workflow-to-github-actions/

maximlt: > we can ensure that everything is ready before publishing

I'm still unsure why we'd need that. Isn't that the purpose of the dev releases?

Hoxbro: Things can still go wrong. The dev and final releases are fundamentally independent; e.g., a new package can be released between the dev and final releases.

This will also give extra security if a wrong tag has been pushed, you forgot to pull before tagging locally, or you forgot to merge PRs.

maximlt: Let's bring this up in a HoloViz meeting or a more general channel. We need to ensure that the repos are structured and behave consistently.

philippjfr (approved):

Issue #5971, opened 2023-12-02
Dynamically Changing height of Matplotlib pane does not change height of div inside host


MarcSkovMadsen:
I'm on the main branch of Panel using bokeh 3.3.1

I'm trying to understand why the Matplotlib pane does not support responsive layouts. I would expect I could test the responsiveness by dynamically changing the height of a Matplotlib pane. But it does not work.

## Base example

- Why does the gray matplotlib plot not adjust its height when I drag the slider?
- Why does the inner yellow bordered div not adjust the height when I drag the slider?

https://github.com/holoviz/panel/assets/42288570/44401f94-9648-4209-9001-849a09de77a2

```python
import panel as pn

from matplotlib.figure import Figure


fig = Figure(figsize=(16,4))
fig.set_facecolor("lightgray")

component = pn.pane.Matplotlib(
    fig,
    height=300,
    sizing_mode="stretch_width",
    styles={"border": "1px solid black"},
    stylesheets=["div {border: 1px solid yellow}"]
)

pn.state.location.sync(component, [
    "height",
])

pn.Column(
    pn.widgets.IntSlider.from_param(component.param.height, value=300, start=100, end=400),
    component,
).servable()
```

## Create new pane when height is changed

If I try to create a new Matplotlib pane on each change of `height` I would expect the yellow and black border to follow the gray matplotlib plot. But it does not.

You can see that

- The inner `div` stays at the initial height

This makes it hard for anything inside the inner div to adjust its height responsively.

https://github.com/holoviz/panel/assets/42288570/b6cf60c1-d98c-477f-aa0c-6fea0d5bb482

```python
import panel as pn

from matplotlib.figure import Figure


fig = Figure(figsize=(16,4))
fig.set_facecolor("lightgray")

component = pn.pane.Matplotlib(
    fig,
    height=300,
    sizing_mode="stretch_width",
)

pn.state.location.sync(component, [
    "height",
])

@pn.depends(
    height=component.param.height,
)

def _get_matplotlib(**kwargs):
    return pn.pane.Matplotlib(
        fig,
        **kwargs,
        sizing_mode="stretch_width",
        styles={"border": "1px solid black"},
        stylesheets=["div {border: 1px solid yellow}"]
    )

pn.Column(
    pn.widgets.IntSlider.from_param(component.param.height, value=300, start=100, end=400),
    _get_matplotlib,
).servable()
```

## Additional comments

[...]

Issue #5967, opened 2023-12-01
Cannot get responsive Matplotlib plots


MarcSkovMadsen:
Currently I'm not able to make matplotlib plots resize responsively (horizontally) in a useful way. I've tried all kinds of combinations of `sizing_mode`, `height`, `max_height` etc.

![image](https://github.com/holoviz/panel/assets/42288570/d05b173d-90ce-421e-ab56-1268bfe0b6f4)

```python
from pandas import read_csv
import panel as pn

from plots import first_taxi, plot_hist, plot_tips

accent = "#f7b731"

pn.extension(
    sizing_mode="stretch_width",
)

data = pn.state.as_cached(
    key="nyc-taxi", fn=read_csv, filepath_or_buffer="nyc-taxi.csv"
)
plot_hist = pn.cache(plot_hist)
plot_tips = pn.cache(plot_tips)

sample_input = pn.widgets.FloatSlider(
    value=0.1, start=0, end=1, step=0.01, name="Sample"
)
scale_input = pn.widgets.Checkbox(name="Use Log Scale", margin=(20, 10, 0, 10))

sample_data = pn.bind(data.sample, frac=sample_input)

pn.template.FastListTemplate(
    site="Panel",
    title="NYC Taxi Data",
    sidebar=[
        "## NYC Taxi Data",
        sample_input,
        scale_input,
    ],
    main=[
        pn.bind(first_taxi, sample_data),
        pn.bind(plot_tips, sample_data, scale_input, accent),
        pn.bind(plot_hist, sample_data, accent),
    ],
    theme_toggle=False,
    accent=accent,
    main_layout=None,
).servable()
```

```python
from matplotlib.pyplot import close
from plotnine import (
    aes,
    geom_histogram,
    geom_point,
    ggplot,
    scale_x_log10,
    scale_y_log10,
    theme_bw,
)


def plot_tips(sampled_data, log, color="black"):
    plot = (
        ggplot(sampled_data, aes("tip_amount", "total_amount"))
        + geom_point(color=color)
        + theme_bw()
    )
    if log:
        plot = plot + scale_x_log10() + scale_y_log10()
    fig = plot.draw()
    close()
    return fig


def plot_hist(sampled_data, color="black"):
    plot = (
        ggplot(sampled_data, aes(x="total_amount"))
        + geom_histogram(binwidth=5, color=color, fill=color)
        + theme_bw()
    )
    fig = plot.draw()
    close()
    return fig

def first_taxi(data):
[...]

PR #5973, opened 2023-12-02
fix Matplotlib Responsiveness and improve reference notebook

MarcSkovMadsen:
I believe this PR fixes the Matplotlib responsiveness issues reported in #5967. Its really hard to know because there are so many possible combinations of settings. And for example the `scale_height`, `scale_width` and `scale_both` values I've never understood how are expected to work. For me that is a big mystery and I'm not able to understand it from the Panel docs. The same goes for the `height_policy` and `width_policy`. I just never use them.

You can confirm manually that it works via 

```python
import panel as pn

from matplotlib.figure import Figure


fig = Figure(figsize=(16,4))
fig.set_facecolor("salmon")
ax = fig.add_subplot(111)  
ax.plot(
    [1, 2, 3, 4, 5],
    [10, 8, 6, 4, 2],
)

component = pn.pane.Matplotlib(
    fig,
    height=300,
    format="svg",
    encode=False,
    fixed_aspect=False,
    sizing_mode="stretch_width",
    styles={"border": "1px solid black"},
    stylesheets=["div {border: 1px solid yellow}"]
)

pn.state.location.sync(component, [
    "height",
    "format",
    "encode",
    "fixed_aspect",
    "sizing_mode",
])

settings=pn.WidgetBox(
    pn.widgets.IntSlider.from_param(component.param.height, start=100, end=400, step=25),
    component.param.format, 
    component.param.encode, 
    component.param.fixed_aspect,
    component.param.sizing_mode, 
)
pn.Row(settings, component).servable()
```

Due to bug #5971 you will have to manually refresh the page every time you change a setting.

## Todo

- [x] Add layout examples to Matplotlib reference example notebook
- [x] Generally update Matplotlib reference notebook. Its been a long time since last update and it shows.
- [x] Compare with alternative frameworks documentation and see if somethings missing.

## It works - Samples

image

image

image

image

image

### Shiny

Makes the plots width responsive by default and recommends setting a height. Wish we did the same. Its a shame everything is fixed size by default. See https://shiny.posit.co/py/api/ui.output_plot.html#shiny.ui.output_plot

### Streamlit

Streamlit is also width responsive by default.

image

### DIFF
--- a/panel/pane/image.py
+++ b/panel/pane/image.py
@@ -167,7 +167,8 @@ class ImageBase(FileBase):
         alt = f'alt={self.alt_text!r}' if self.alt_text else ''
         width = f' width: {width};' if width else ''
         height = f' height: {height};' if height else ''
-        html = f'<img src="{src}" {alt} style="max-width: 100%; max-height: 100%; object-fit: contain;{width}{height}"></img>'
+        object_fit = "contain" if self.fixed_aspect else "fill"
+        html = f'<img src="{src}" {alt} style="max-width: 100%; max-height: 100%; object-fit: {object_fit};{width}{height}"></img>'
         if self.link_url:
             html = f'<a href="{self.link_url}" target="_blank">{html}</a>'
         return escape(html)
@@ -440,13 +441,8 @@ class SVG(ImageBase):
         if self.encode:
             ws = f' width: {w};' if w else ''
             hs = f' height: {h};' if h else ''
-            data = f'<img src="{self._b64(data)}" style="max-width: 100%; max-height: 100%; object-fit: contain;{ws}{hs}"></img>'
-        elif self.width or self.height or self.sizing_mode not in (None, 'fixed'):
-            self.param.warning(
-                'SVG sizing cannot be scaled if the SVG has been embedded '
-                'but is not encoded. Either enable encoding or manipulate '
-                'the SVG itself.'
-            )
+            object_fit = "contain" if self.fixed_aspect else "fill"
+            data = f'<img src="{self._b64(data)}" style="max-width: 100%; max-height: 100%; object-fit: {object_fit};{ws}{hs}"></img>'
         if isinstance(data, bytes):
             data = data.decode('utf-8')
         return dict(width=width, height=height, text=escape(data))
--- a/panel/pane/plot.py
+++ b/panel/pane/plot.py
@@ -221,7 +221,7 @@ class Matplotlib(Image, IPyWidget):
     dpi = param.Integer(default=144, bounds=(1, None), doc="""
         Scales the dpi of the matplotlib figure.""")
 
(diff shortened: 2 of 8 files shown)


panel/pane/image.py
```diff
@@ -440,13 +441,8 @@ def _transform_object(self, obj: Any) -> Dict[str, Any]:
         if self.encode:
             ws = f' width: {w};' if w else ''
             hs = f' height: {h};' if h else ''
-            data = f'<img src="{self._b64(data)}" style="max-width: 100%; max-height: 100%; object-fit: contain;{ws}{hs}"></img>'
-        elif self.width or self.height or self.sizing_mode not in (None, 'fixed'):
```
MarcSkovMadsen, line 444: I've read the warning below 10 times and still I do not understand it :-). I don't think what ever was ment with it applies after this fix.

MarcSkovMadsen:
Ok. There are differences between browsers. Firefox on top. Edge below.

image

MarcSkovMadsen:
Changing the height from `auto` to `100%` made both edge and firefox behave well and in the same way.

MarcSkovMadsen:
It would be nice to get a quick review from @philippjfr because there is a lot of history and a lot possible ways to use this Matplotlib pane.

panel/pane/plot.py
```diff
@@ -210,7 +221,7 @@ class Matplotlib(Image, IPyWidget):
     dpi = param.Integer(default=144, bounds=(1, None), doc="""
         Scales the dpi of the matplotlib figure.""")
 
-    encode = param.Boolean(default=True, doc="""
+    encode = param.Boolean(default=False, doc="""
```
MarcSkovMadsen, line 224: I believe (?) encode was set to `True` to try to enable responsive 'svg' outputs. I believe the issues have been fixed and encoding is no longer needed. And encoding slows things down so I believe it should not be the default.

Please also note the `encode` parameter was not documented in the reference Notebook and I've never met anyone on discourse asking about it. So I believe its safe to change.

philippjfr: Agreed.

panel/pane/plot.py
```diff
@@ -356,8 +368,14 @@ def _data(self, obj):
             dpi=self.dpi,
             bbox_inches=bbox_inches
         )
-        return b.getvalue()
+        value = b.getvalue()
+
+        if self.format=="svg":
+            value = _make_matplotlib_svg_responsive(value)
+            if  not self.fixed_aspect:
```
philippjfr, line 375:
```suggestion
            if not self.fixed_aspect:
```

panel/pane/image.py
philippjfr: Yep indeed, your changes address the problem.

philippjfr (approved):

PR #5975, opened 2023-12-02
Replace whitelist blacklist with allowlist denylist

droumis:
This PR replaces whitelist, blacklist with allowlist, denylist in the interest of being a more inclusive community.

### DIFF
--- a/panel/io/reload.py
+++ b/panel/io/reload.py
@@ -15,7 +15,7 @@ _modules = set()
 _callbacks = {}
 
 # List of paths to ignore
-DEFAULT_FOLDER_BLACKLIST = [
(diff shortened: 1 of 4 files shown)


panel/io/reload.py
```diff
@@ -15,7 +15,7 @@
 _callbacks = {}
 
 # List of paths to ignore
-DEFAULT_FOLDER_BLACKLIST = [
+DEFAULT_FOLDER_denylist = [
```
philippjfr, line 18:
```suggestion
DEFAULT_FOLDER_DENYLIST = [
```

philippjfr (approved):

PR #5990, opened 2023-12-05
Allow interrupting sync calls midway for ChatInterface

ahuang11:
Makes it so that sync tasks are submitted to a separate thread so it's cancellable.

https://github.com/holoviz/panel/assets/15331990/94df994e-5ddc-4594-abea-1941d7f39e0c

```python
import panel as pn
from openai import OpenAI

pn.extension()


    response = client.chat.completions.create(
        model="gpt-3.5-turbo",
        stream=True,
    )
    message = None
    for chunk in response:
        if part is not None:
            message = instance.stream(part, user="OpenAI", message=message)


client = OpenAI()
chat_interface.send(
)
chat_interface.stream("Hello")
chat_interface.show()
```

### DIFF
--- a/panel/chat/feed.py
+++ b/panel/chat/feed.py
@@ -204,9 +204,6 @@ class ChatFeed(ListPanel):
     _callback_is_running = param.Boolean(default=False, doc="""
         Whether the callback is currently running.""")
 
-    _callback_is_async = param.Boolean(default=False, doc="""
-        Whether the callback is an async function.""")
-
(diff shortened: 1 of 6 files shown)


panel/chat/feed.py
```diff
@@ -538,6 +547,9 @@ def stream(
         -------
         The message that was updated.
         """
+        if message is not None and (self._callback_future is None or self._callback_future.cancelled()):
```
ahuang11, line 550: Not sure how to stop streaming midway.

Hoxbro:
Was this PR ready to be merged? Failing tests seem very relevant.

PR #5996, opened 2023-12-05
Update pyodide docs version to use `panel.io.convert`

Hoxbro:
Right now, I'm getting this:

image

I have started this branch from v1.3.4, so it should be possible to re-publish the docs from this PR. 

image

Will set over a docs dev build.


philippjfr (approved):

Hoxbro:
The main docs site has been updated with this change.

PR #5998, opened 2023-12-06
Fix post callback for default

ahuang11:
Fixes post callback for existing buttons.

Also, attaches callback/post_callback to auto send on value change.

```python
import panel as pn
pn.extension()

def pre_callback(instance, event):
    instance.send("This sends before send", respond=False)

def post_callback(instance, event):
    instance.send("This sends after send", respond=False)

chat_interface = pn.chat.ChatInterface(button_properties = {
    "send": {"post_callback": post_callback, "callback": pre_callback},
})
chat_interface.show()

```

https://github.com/holoviz/panel/assets/15331990/4c057283-33c2-42e0-b197-f1ba56796e44

### DIFF
--- a/panel/chat/interface.py
+++ b/panel/chat/interface.py
@@ -203,7 +203,11 @@ class ChatInterface(ChatFeed):
             if default_properties:
                 default_callback = default_properties["_default_callback"]
                 callback = (
-                    self._wrap_callbacks(callback=callback, post_callback=post_callback)(default_callback)
+                    self._wrap_callbacks(
+                        callback=callback,
+                        post_callback=post_callback,
+                        name=name,
+                    )(default_callback)
                     if callback is not None or post_callback is not None else default_callback
                 )
             elif callback is not None and post_callback is not None:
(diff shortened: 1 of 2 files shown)


panel/chat/interface.py
```diff
@@ -309,13 +314,18 @@ def _init_widgets(self):
...
+            post_callback: Callable | None = None,
+            name: str = ""
         ):
         """
         Wrap the callback and post callback around the default callback.
         """
         def decorate(default_callback: Callable):
             def wrapper(self, event: param.parameterized.Event):
+                if name == "send" and not self.active_widget.value:
+                    # don't trigger if no message to prevent duplication
```
ahuang11, line 326: This is primarily because value gets reset to "", and re-triggers the callback without this clause

philippjfr (approved):

panel/tests/chat/test_interface.py
```diff
@@ -215,6 +217,34 @@ def post_callback(instance, event):
...
+        def post_callback(instance, event):
+            instance.send("This should show", respond=False)
+
+        chat_interface.button_properties = {
+            "clear": {"post_callback": post_callback},
+        }
+        clear_button = chat_interface._input_layout[-1]
+        chat_interface.send("This shouldn't show up!", respond=False)
+        clear_button.param.trigger("clicks")
+        time.sleep(0.5)
```
philippjfr, line 230: Should have reviewed this more closely, `time.sleep` should not be used in the test suite (except for very rare cases). Use the `wait_until` utility.

PR #6008, opened 2023-12-07
Clear selection if value change from `pagination=remote`

hoxbro:
If the value of the Tabulator changes, we can no longer trust the selection; therefore, we clear it.


setup.py
```diff
@@ -176,7 +176,7 @@ def run(self):
     'jupyter_bokeh >=3.0.7',
     'django <4',
     'channels',
-    'pyvista',
+    'pyvista <0.43',  # Breaks an example
```
ahuang11, line 179: I think https://github.com/holoviz/panel/pull/6020 this fixes it

Hoxbro: Yep. I just haven't gotten around to updating this PR since the fix.

```suggestion
    'pyvista',  # Breaks an example
```

panel/widgets/tables.py
```diff
@@ -1562,6 +1562,11 @@ def _update_max_page(self):
         for ref, (model, _) in self._models.items():
             self._apply_update([], {'max_page': max_page}, model, ref)
 
+    @param.depends("value", watch=True)
+    def _clear_selection_remote_pagination(self):
```
philippjfr, line 1566: Think this may have to be a watcher that checks `if event.new is not event.old`.

philippjfr (approved):

Issue #5915, opened 2023-11-24
In the Tabulator of panel 1.3.2, using HTMLTemplateFormatter for the Index now raises an exception.


kuri-menu:
It worked fine in panel 1.3.1

#### ALL software version info
Windows 10 python 3.11.6

bokeh                     3.3.1
holoviews                 1.18.1
panel                     1.3.2
param                     2.0.1
jupyter-bokeh             3.0.7
jupyter_client            7.4.9
jupyter_core              5.3.1
jupyter-events            0.6.3
jupyter-lsp               2.2.0
jupyter_server            2.5.0
jupyter_server_terminals  0.4.4
jupyterlab                4.0.9
jupyterlab-pygments       0.2.2
jupyterlab_server         2.23.0
jupyterlab-widgets        3.0.8
ipykernel                 6.24.0

#### Complete, minimal, self-contained example code that reproduces the issue

```

from bokeh.models.widgets.tables import HTMLTemplateFormatter
import panel as pn
pn.extension('tabulator')

df = pd.DataFrame(
    {
        'A':[1,2,3],
        'B':[1,2,3],
    }
)
df = df.set_index('A')

tabu = pn.widgets.Tabulator(
    df,
    formatters=dict(
        A=HTMLTemplateFormatter(template='<a href="https://www.google.com/search?code=<%= value %>"><%= value %></a>'),
    ),
)
tabu

```

#### Stack traceback and/or browser JavaScript console output

---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
File [~\AppData\Local\Programs\Python\Python311\Lib\site-packages\IPython\core\formatters.py:974](https://untitled+.vscode-resource.vscode-cdn.net//~/AppData/Local/Programs/Python/Python311/Lib/site-packages/IPython/core/formatters.py:974), in MimeBundleFormatter.__call__(self, obj, include, exclude)
    [971](file:///c%3A/Users/nao/AppData/Local/Programs/Python/Python311/Lib/site-packages/IPython/core/formatters.py?line=970)     method = get_real_method(obj, self.print_method)
    [973](file:///c%3A/Users/nao/AppData/Local/Programs/Python/Python311/Lib/site-packages/IPython/core/formatters.py?line=972)     if method is not None:
--> [974](file:///c%3A/Users/nao/AppData/Local/Programs/Python/Python311/Lib/site-packages/IPython/core/formatters.py?line=973)         return method(include=include, exclude=exclude)
[...]

PR #6010, opened 2023-12-07
Tabulator: don't attempt to set `text_align` on Bokeh formatters that don't support it

maximlt:
I got pretty much the same error as reported in https://github.com/holoviz/panel/issues/5915 and fixed in https://github.com/holoviz/panel/pull/5922, where the code fails while trying to set `text_align` on a Bokeh formatter that doesn't support this attribute. https://github.com/holoviz/panel/pull/5922 fixed that for the index but it applies to any other column too.

```py
import pandas as pd
import panel as pn

from bokeh.models import HTMLTemplateFormatter

df = pd.DataFrame(dict(x=['a', 'b']))

pn.widgets.Tabulator(
    df,
    widths=300,
    text_align=dict(x='center'),
    formatters=dict(x=HTMLTemplateFormatter(template='<b><%= value %></b>'))
).servable()
```

### DIFF
--- a/panel/tests/widgets/test_tables.py
+++ b/panel/tests/widgets/test_tables.py
@@ -2149,13 +2149,13 @@ def test_bokeh_formatter_index_with_no_textalign():
     serve_and_request(table)
     wait_until(lambda: bool(table._models))
 
-
-def test_bokeh_formatter_column_with_no_textalign(document, comm):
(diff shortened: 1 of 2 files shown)


Hoxbro (approved):

philippjfr:
Merging, pyvista related example is failing but that was already fixed in a separate PR.

PR #6011, opened 2023-12-07
Add NestedSelect dynamic levels

ahuang11:
Apparently dynamic levels was already supported after changes for this https://github.com/holoviz/panel/pull/5791#issuecomment-1830373970

There was just a tiny fix to make to hide the subsequent options.

https://github.com/holoviz/panel/assets/15331990/cc007860-936c-47cf-89e0-78695bb957c8

### DIFF
--- a/panel/tests/widgets/test_select.py
+++ b/panel/tests/widgets/test_select.py
@@ -629,6 +629,8 @@ def test_nested_select_dynamic_levels(document, comm):
     assert select._widgets[2].options == []
     assert select._widgets[3].options == []
 
+    assert select.value == {"A": "Easy", "B": "Easy_A", "C": None, "D": None}
+
     # now update to Medium
     select.value = {"A": "Medium", "B": "Medium_C"}
     assert select._widgets[0].visible
@@ -641,6 +643,8 @@ def test_nested_select_dynamic_levels(document, comm):
     assert select._widgets[2].options == ["Medium_C_1", "Medium_C_2"]
     assert select._widgets[3].options == ["Medium_C_1_1"]
 
+    assert select.value == {"A": "Medium", "B": "Medium_C", "C": "Medium_C_1", "D": "Medium_C_1_1"}
+
     # now update to Hard
     select.value = {"A": "Hard"}
     assert select._widgets[0].visible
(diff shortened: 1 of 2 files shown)


panel/tests/widgets/test_select.py
```diff
@@ -601,6 +601,65 @@ def list_options(level, value):
...
+                    "Medium_C_1": ["Medium_C_1_1"],
+                    "Medium_C_2": ["Medium_C_2_1", "Medium_C_2_2"],
+                },
+            },
+            "Hard": {}
+        },
+        levels=["A", "B", "C", "D"],
+    )
+    select
+    select.value = {"A": "Easy", "B": "Easy_A"}
```
ahuang11, line 621:
```suggestion
```

ahuang11:
Those tests should be fixed in a separate PR now.

MarcSkovMadsen:
Nice. My use case does not always start from a well defined `dict`. Instead it starts either from

- A dataframe for example with categorical columns `Source`, `Product`, `Var`, `GPH`.
-  a set of functions `get_sources()`, `get_products(source)`, `get_vars(source, product)` and `get_gphs(source, product, var)`.

It would be really, really nice if these cases where supported out of the box or it was documented how to implemented them.

philippjfr:
Yes, I still really want to see the .from_data (naming to be decided) method implemented on widgets where it makes sense. I find myself writing code to find the min/max or the unique values in a column or index constantly.

MarcSkovMadsen:
Furthermore it would be really useful to be able to define the `widgets` used. I would for example like to sometimes be able to set the `size` of the `Select` widget to something `>1` or use a `RadioButtonGroup` instead of a `Select`.

ahuang11:
For your second to last request, it's already possible https://github.com/holoviz/panel/blob/main/examples%2Freference%2Fwidgets%2FNestedSelect.ipynb

ahuang11:
For your second to last request it's already possible https://github.com/holoviz/panel/blob/main/examples%2Freference%2Fwidgets%2FNestedSelect.ipynb

ahuang11:
To clarify, would it be possible to get this PR in before those other feature requests?

PR #6013, opened 2023-12-08
Doc: Add an announcement banner for Panel 1.4

maximlt:
This PR is motivated by the goal of getting more GitHub stars, pushing on multiple fronts around a single event, that could be the Panel 1.4 release.

I built the small test site located in nbsite's repo to see how that would look like:

image


maximlt:
@philippjfr if the blog post isn't ready when Panel 1.4 is released (which is often the case IIRC), I can update this PR to link to the [Releases](https://panel.holoviz.org/about/releases.html) page.

maximlt:
@philippjfr would you like this to be merged before releasing 1.4.0?

philippjfr:
> @philippjfr would you like this to be merged before releasing 1.4.0?

Yes, I think so. How does it work? Will it always be there? Is it on all pages or just the home page? What's the process for removing it when "Panel 1.4 has just been released" is no longer true?

maximlt:
> How does it work? Will it always be there? Is it on all pages or just the home page? What's the process for removing it when "Panel 1.4 has just been released" is no longer true?

Good questions! :)

Yes, this banner is displayed on all pages.

Initially, I thought we'd keep the banner until a future release, before which we'd just remove the `announcement` key from `conf.py`. Looking at the [theme docs](https://pydata-sphinx-theme.readthedocs.io/en/v0.13.3/user_guide/announcements.html) again, they indicate you can point to an HTML file. So I've just tried that locally:

- it fetches the banner content from the HTML file and displays it
- if the file can't be fetched, no banner is displayed

So it seems we could use that to dynamically control whether the banner is displayed or not, and its content when displayed. I guess it'd be easier if the HTML file was served from another place (e.g. S3)?

I think I still prefer the simpler option of removing the announcement in a future release, perhaps in a month or so.  But I'd be fine implementing the other approach too.

---

By the way, this is the current text:

> Panel 1.4 has just been released! Checkout the <a href='https://panel.holoviz.org/about/releases.html#version-1-4-0'>release notes</a> and support Panel by giving it a 🌟 on <a href='https://github.com/holoviz/panel'>Github</a>.

Feel free to enhance it, here's another suggestion:

> Panel 1.4 released! Checkout the <a href='https://panel.holoviz.org/about/releases.html#version-1-4-0'>new features</a> and support Panel by giving it a 🌟 on <a href='https://github.com/holoviz/panel'>Github</a>.

Interestingly, my initial intent with this PR was just to nudge people to add a star to Panel on GitHub. But I feel like announcing the new release also adds value, it can easily be missed if you're not watching closely our networks or Github.

philippjfr:
Thanks for the detailed answer. Since we are planning a 1.5 release pretty soon and there will be micro-releases we will have ample opportunity to adjust.

philippjfr (approved):

PR #6014, opened 2023-12-08
Correctly set error page to be rendered

mitulb:
Correctly sets the error page to be rendered when authorize_callback returns False.

Repro

```
import panel as pn

def authorize(user_info):
    return False

pn.config.authorize_callback = authorize

pn.Row("hello world").servable()

```

I would expect running the above with oidc provider "generic" to return the error page however I get a 500 internal server error with the following in the logs

```TypeError: write() only accepts bytes, unicode, and dict objects```


philippjfr:
Thank you!

philippjfr (approved):

PR #6023, opened 2023-12-09
Fix and improve fast templates

MarcSkovMadsen:
This PR fixes various *fast* styling issues introduced by Panel 1.0.

- Removing the bootstrap styling and adding the fast styling to Buttons.
- Giving links the accent color
- Making the triangle in the `Select` widget visible - also in dark mode.

It also improves the template styling when `main_layout=None` giving it the familiar *Streamlit Gray Sidebar* look. I think I will be using this version and more and more because I think it looks the best.

- I was thinking maybe we should add a deprecation warning that we will be changing the default value of `main_layout` to `None` in Panel 1.6 or similar?

I also add a tiny bid of left padding to the sidebar too make it look more symmetric and less cramped.

I also fix https://github.com/holoviz/panel/issues/6024.

## Test Script

I've been using the below test script

```python
import panel as pn

pn.extension(sizing_mode="stretch_width", inline=True)

ACCENT = "#da1a5f"

disabled = pn.widgets.Checkbox(name="Disabled", value=False)

@pn.depends(disabled)
def get_buttons(disabled):
       buttons = pn.Row(
       *(pn.Column(*(pn.widgets.Button(name=p, button_type=p, button_style=bs, disabled=disabled) for p in pn.widgets.Button.param.button_type.objects))
       for bs in pn.widgets.Button.param.button_style.objects)
       )
       button_radioboxes = pn.Row(
       *(pn.Column(*(pn.widgets.RadioButtonGroup(name=p, options=["AAA", "BBB"], button_type=p, button_style=bs, disabled=disabled) for p in pn.widgets.Button.param.button_type.objects))
       for bs in pn.widgets.Button.param.button_style.objects)
       )
       description="""## Header
Some text and a [link](https://panel.holoviz.org)
       
"""
       select = pn.widgets.Select(value="Option A", options=["Option A", "Option B"])
       return pn.Column(
              buttons, button_radioboxes, description, select, margin=0
       )


pn.template.FastListTemplate(
       title="Test App", 
       main=[],
       sidebar=[disabled, get_buttons],
       accent=ACCENT,
       # main_layout=None,
).servable()
```

## After

This is how things will be looking when applying this PR.

image

image

### `main_layout=None`

image

image

### DIFF
--- a/panel/template/fast/fast.css
+++ b/panel/template/fast/fast.css
@@ -146,7 +146,7 @@ img.app-logo {
   transform-origin: center left;
   height: calc(100vh - 76px);
   overflow-y: auto;
-  padding: 15px 10px 5px 0;
+  padding: 15px 10px 5px 5px;
 }
 
 #sidebar.hidden {
--- a/panel/template/fast/fast.css
+++ b/panel/template/fast/fast.css
@@ -214,6 +214,10 @@ img.app-logo {
   margin-block-end: 0;
 }
 
+div.pn-wrapper {
+  height: 100%;
+}
+
 .pn-modal-close {
   position: absolute;
   right: 10%;
--- a/panel/template/fast/list/fast_list_template.html
+++ b/panel/template/fast/list/fast_list_template.html
@@ -53,6 +53,9 @@
   #sidebar {
     min-width: var(--sidebar-width);
     max-width: var(--sidebar-width);
+	{% if not main_layout %}
+	background-color: var(--panel-surface-color);
+	{% endif %}
   }
 </style>
 
--- a/panel/template/fast/grid/fast_grid_template.html
+++ b/panel/template/fast/grid/fast_grid_template.html
@@ -63,6 +63,9 @@
   #sidebar {
     min-width: var(--sidebar-width);
     max-width: var(--sidebar-width);
+    {% if not main_layout %}
+    background-color: var(--panel-surface-color);
+    {% endif %}
   }
 </style>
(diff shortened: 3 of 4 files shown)


philippjfr (approved):

panel/template/fast/fast.css
```diff
@@ -214,6 +214,10 @@ img.app-logo {
   margin-block-end: 0;
 }
 
+div.pn-wrapper {
+  height: 100%;
```
philippjfr, line 218: Are we sure about this one?

panel/template/fast/list/fast_list_template.html
```diff
@@ -53,6 +53,9 @@
   #sidebar {
     min-width: var(--sidebar-width);
     max-width: var(--sidebar-width);
+	{% if not main_layout %}
+	background-color: var(--panel-surface-color);
+	{% endif %}
```
philippjfr, line 58:
```suggestion
    {% if not main_layout %}
    background-color: var(--panel-surface-color);
    {% endif %}
```

philippjfr (approved):
Looks good, just want to be sure that the 100% height on pn-wrapper works well in all cases.

MarcSkovMadsen:
Its a good question. I can see that the fast-card has this by it self. Isabella tested in an app with two elements in main area. And it was fine.

philippjfr:
I'll merge and will make sure to do some testing over the next few days.

Issue #6021, opened 2023-12-08
Add pagination to ChatFeed


ahuang11:
I think I was preparing to add it to ChatBox, but that got deprecated.
https://github.com/holoviz/panel/pull/5177

Based on the discussion at https://github.com/Quansight/ragna/discussions/239, I realized that it probably should be attempted to be implemented again.

The question is whether this should be a feature of ChatFeed or Column?

philippjfr:
-1 on pagination, I think it's poor UX for a chat feed and the more common and nicer UX will be to dynamically stream chat messages in and out as you scroll through the history.

philippjfr:
Based on the referenced PR it sounds like I'm taking your wording of "pagination" too literally and we're on the same page :)

ahuang11:
Do you think it should be implemented in the Column TS model or in ChatFeed?

philippjfr:
Definitely a custom ChatFeed model, it'll have to do smart stuff like watching the visible elements, fetch new ones as you're scrolling etc. which all requires a bunch of orchestration that shouldn't be enabled on Column.

hyamanieu:
If I may add a comment: I would say both performance and scrolling are linked as your now closed PR #5177 is about scrolling.

It seems for Ragna or whichever other application there is, you're focusing on scrolling to top message or scrolling to bottom. But a user could want to scroll to any message within a long conversation.

I think it's a trouble for some application that a specific message cannot be selected on python side for automatic scrolling to it. Specifically, most messaging systems have a feature to jump to specific messages, either by doing a search, or due to specific tags (has a message code? has the message a "like" mention? has the message a URL? Was I mentioned? etc.)

In a user-story style: "as a user, I want to be able to jump to a specific message below 1 second".

To me the solution to this performance shall not simply be focussing on some continuous pagination, but I think there must be a parameter "message_index" which controls what is seen on client side. And depending on the viewport height and this index, only a specific amount of message should be loaded in the client (with a buffer surrounding the viewport).

(in summary, I think this issue is somewhat linked to #6083 due to the user story I and you may have)

ahuang11:
I think what you're suggesting is complementary to pagination, so I think it can come after.

However, I am not sure whether scroll_index should live on Column or Log.

hyamanieu:
> I think what you're suggesting is complementary to pagination, so I think it can come after.
> 
> However, I am not sure whether scroll_index should live on Column or Log.

[...]

PR #6031, opened 2023-12-12
Add pn.Feed to allow buffering of feed objects

ahuang11:
Closes https://github.com/holoviz/panel/issues/6021

Sets up the plumbing to add a Logs layout, for the internal _chat_log inside `ChatFeed`.

See bottom for demo.


ahuang11:
Temporarily added the functionality to Column since I couldn't get Logs to render:

https://github.com/holoviz/panel/assets/15331990/7c7afc39-96c0-4e3d-b784-c8f5f44fb3a2

panel/models/layout.py
```diff
@@ -72,3 +71,13 @@ class Card(Column):
     hide_header = Bool(False, help="Whether to hide the Card header")
 
     tag = String("tag", help="CSS class to use for the Card as a whole.")
+
+
+class Log(Column):
+    min_visible = Int(
```
ahuang11, line 77: Add max_visible too

panel/models/column.ts
```diff
@@ -56,13 +57,47 @@ export class ColumnView extends BkColumnView {
     }
   }
 
+  load_more_entries(): void {
+    if (this.el.scrollTop < 100 && this.loaded_entries < this.model.children.length) {
```
ahuang11, line 61: Parameterize 100

panel/models/column.ts
```diff
@@ -56,13 +57,47 @@ export class ColumnView extends BkColumnView {
...
+      const entriesToAdd = Math.min(20, this.model.children.length - this.loaded_entries);
```
ahuang11, line 62: Parameterize 20

Hoxbro:
You likely need to update index.ts

ahuang11:
Hmm getting closer, but maybe I'm rebuilding it wrong or I need another command besides `panel build panel` 
image

Hoxbro:
Try running `panel bundle --all --verbose`, hard refresh the browser and if you see no errors, update

``` typescript
this.define<Log.Props>(({ Int }) =>
    min_visible: [Int, 10],
}));
```

ahuang11:
You're magical :D that worked

ahuang11:
Works pretty well now~
https://github.com/holoviz/panel/assets/15331990/afc71e44-6209-48f7-b452-7f3cf418b5b3

panel/models/column.ts
```diff
@@ -22,14 +22,14 @@ export class ColumnView extends BkColumnView {
...
-      this.el.scrollTo({top: this.model.scroll_position});
+      this.el.scrollTo({ top: this.model.scroll_position });
     });
   }
 
   scroll_to_latest(): void {
     // Waits for the child to be rendered before scrolling
     requestAnimationFrame(() => {
-      this.el.scrollTo({top: this.el.scrollHeight});
+      this.model.scroll_position = Math.round(this.el.scrollHeight);
```
ahuang11, line 32: auto_scroll_limit wasn't working without this change.

panel/models/log.ts
```diff
@@ -0,0 +1,97 @@
+import { Column, ColumnView } from "./column";
```
philippjfr, line 1: Our other .ts files use 2 space indents.

philippjfr (commented):
I'm fine with this approach, just wondering at what point it makes sense to render only the currently visible models plus some buffer and then display a loading indicator if they aren't fetched quick enough. For large numbers of messages this will significantly reduce serialization time.

ahuang11:
That sounds like a good idea. How do I subset the number of children rendered?

philippjfr:
It's probably going to be pretty involved. It'd look something like this:

1. JS watches each child to detect if it's in the viewport and updates a property with the range of currently visible children (probably throttled somehow).
2. Python observes changes in this range property and if any of the children in the buffer zone are visible then the buffer has to be refilled with new children (and if we exceed some total threshold value we also delete any children that are farthest out-of-view).

In practice that means overriding `Log._get_objects` and implement a version that handles adding and removing items as the visible items change.

ahuang11:
Needs cleanup and maybe some logic fixes with the timeout, but now it only renders the loaded_entries. 

https://github.com/holoviz/panel/assets/15331990/c58c0cae-8679-49db-a4a3-4d9b3ddce94a

ahuang11:
Will add reference gallery tomorrow.

https://github.com/holoviz/panel/assets/15331990/be9c2a91-81d4-4828-8117-8ee4ce4f8868

Strangely, it takes a long time to serve:
```python
import panel as pn

pn.extension()

pn.Log(
    *list(range(10000)), scroll=True, height=250, width=250
).servable()

```
vs one order of magnitude less so I'm not sure if it's still rendering everything first then cutting out the pieces.
```
import panel as pn

pn.extension()

pn.Log(
    *list(range(1000)), scroll=True, height=250, width=250
).servable()

```

ahuang11:
Some things I noticed, but not sure how to address:
1. it takes similar time as Column to execute on the backend, but on the browser, it's significantly faster to render and see the webpage:
image
2. The scrollbar bounces around, due to the objects being replaced I think and my algorithm to figure out the original scroll position with new objects prepended
3. If height isn't set, there's no "personal" scrollbar so it doesn't load more.

```python
import panel as pn
pn.extension()

pn.Log(*[pn.chat.ChatMessage(i) for i in range(1000)], height=500).show()
```

philippjfr:
The current implementation doesn't really seem like what we discussed, specifically I thought the idea was that we would keep track of the items that are in the viewport and load (and unload) a buffer around that but here it seems like we are just continuously loading more models and only ever loading items from the bottom. I'll have a go at changing the implementation.

ahuang11:
Right, I wanted to get the loading part right before trying to add the unload/buffer part.

ahuang11:
Strangely, it works in notebook, but not browser:
image


image

ahuang11:
Another todo(?) is to connect setting visible objects and scrolling to the visible object.

philippjfr:
Cannot reproduce your browser issues, and yes, visible_objects should be readonly.

ahuang11:
Another idea is to set a `last_visible_index` to scroll to that entry.

ahuang11:
I guess we already have a scroll_position, but perhaps we should have a scroll_index.
https://github.com/holoviz/panel/issues/6083

philippjfr:
Can you explain a little bit? I am totally lost what you are suggesting and what problem you are trying to solve.

ahuang11:
Okay the scroll button works now!
https://github.com/holoviz/panel/assets/15331990/4d519292-1581-4036-8356-133f0660cfc0

ahuang11:
I think this is ready for review.

In future PRs, I want to:

Allow scroll bar to always be shown https://github.com/holoviz/panel/issues/6237

and

Introduce scroll_index to both Column + Log https://github.com/holoviz/panel/issues/6083

Hoxbro:
I know we have a history of directly having Layout in the panel namespace, but this seems to be more of a niche layout and a bit ambiguous, as it is not straightforward to know it is a layout and not a logging tool. So maybe only have it in the Layout namespace?

philippjfr:
I would definitely agree with that, not sure about `Log` as a name. Let's brainstorm a little bit, I think something based around `Feed` or `ScrollFeed` would be best but am open to other suggestions.

panel/layout/log.py
```diff
@@ -0,0 +1,159 @@
...
+    load_buffer = param.Integer(default=50, bounds=(0, None), doc="""
+        The number of objects loaded on each side of the visible objects.
+        When scrolled halfway into the buffer, the log will automatically
+        load additional objects while unloading objects on the opposite side.""")
+
+    scroll = param.Boolean(default=True, doc="""
+        Whether to add scrollbars if the content overflows the size
+        of the container.""")
+
+    visible_indices = param.List(constant=True, doc="""
```
philippjfr, line 32: readonly: cannot be set, ever
constant: can only be set on instantiation

```suggestion
    visible_indices = param.List(readonly=True, doc="""
```

philippjfr: Maybe we should change this to `visible_range` after all (but leave the `visible_objects` bokeh property the same).

philippjfr:
@jlstevens suggested `LiveFeed` which I like but only really makes sense if we have a `follow` option to always jump to latest.

ahuang11:
> I know we have a history of directly having Layout in the panel namespace, but this seems to be more of a niche layout and a bit ambiguous, as it is not straightforward to know it is a layout and not a logging tool. So maybe only have it in the Layout namespace?

To clarify, maybe it's `pn.Logs`? I believe it should be consistent with other layouts in the panel namespace.

> @jlstevens suggested `LiveFeed` which I like but only really makes sense if we have a follow option to always jump to latest.

I think it's already built-in from inheriting `Column` with:
```
auto_scroll_limit = param.Integer(bounds=(0, None), doc="""
        Max pixel distance from the latest object in the Column to
        activate automatic scrolling upon update. Setting to 0
        disables auto-scrolling.""")
```
So it could default to some high number.

panel/layout/log.py
ahuang11: I need to set it though to handle the scroll button I think:
https://github.com/holoviz/panel/pull/6031/files#diff-9420fdc4f24f25791a1047e728c7ab27e328e6bc0fec1a56427e9d57d03d2139R151-R155

philippjfr: It's not synced with JS, so you'd simply update _synced_indices.

philippjfr:
Don't think `Logs` works as a name either, it doesn't indicate it's a layout and brings an expectation that it's text based only.

ahuang11:
`BufferedColumn`?

philippjfr:
Not bad but don't love it, think we need to give it a few days but make sure you review the thread we had discussing this on Discord.

philippjfr:
Any opinion on `FeedColumn` or `ColumnFeed`? Just trying to keep some symmetry with `ChatFeed`.

ahuang11:
ColumnFeed is fine with me.

panel/layout/log.py
ahuang11: I might be misunderstanding and missing something.

If I do have a `_synced_indices.setter`, wouldn't all this logic be gone?
```python
    @property
    def _synced_indices(self):
        n = len(self.objects)
        if self.visible_indices:
            return list(range(
                max(min(self.visible_indices) - self.load_buffer, 0),
                min(max(self.visible_indices) + self.load_buffer, n)
            ))
        elif self.view_latest:
            return list(range(max(n - self.load_buffer * 2, 0), n))
        else:
            return list(range(min(self.load_buffer, n)))
```

If I do `self._custom_synced_indices`

```python

    @property
    def _synced_indices(self):
        if self._custom_synced_indices is not None:
            return self._custom_synced_indices

        n = len(self.objects)
        if self.visible_indices:
            return list(range(
                max(min(self.visible_indices) - self.load_buffer, 0),
                min(max(self.visible_indices) + self.load_buffer, n)
            ))
        elif self.view_latest:
            return list(range(max(n - self.load_buffer * 2, 0), n))
        else:
            return list(range(min(self.load_buffer, n)))
```

```python

        self._custom_synced_indices = list(
            range(max(n - len(self.visible_indices) + 1, 0), n)
        )
        self.param.trigger("objects")
        self._custom_synced_indices = None
```

There seems to be some race condition and it unsuccessfully scrolls.

panel/layout/feed.py
```diff
@@ -0,0 +1,158 @@
...
+    from pyviz_comms import Comm
+
+    from ..viewable import Viewable
+
+
+class Feed(Column):
+
+    load_buffer = param.Integer(default=50, bounds=(0, None), doc="""
+        The number of objects loaded on each side of the visible objects.
+        When scrolled halfway into the buffer, the log will automatically
```
ahuang11, line 26:
```suggestion
        When scrolled halfway into the buffer, the feed will automatically
```

panel/layout/feed.py
```diff
@@ -0,0 +1,158 @@
...
+        load additional objects while unloading objects on the opposite side.""")
+
+    scroll = param.Boolean(default=True, doc="""
+        Whether to add scrollbars if the content overflows the size
+        of the container.""")
+
+    visible_range = param.Range(readonly=True, doc="""
+        Read-only upper and lower bounds of the currently visible log objects.
```
ahuang11, line 34:
```suggestion
        Read-only upper and lower bounds of the currently visible feed objects.
```

philippjfr (approved):

panel/tests/ui/layout/test_feed.py
```diff
@@ -0,0 +1,94 @@
...
+        '(element) => element.shadowRoot.querySelectorAll(".bk-panel-models-markup-HTML").length'
+    )
+    assert children_count == 58
+
+    # Now scroll to somewhere down
+    feed_el.evaluate('(el) => el.scrollTo({top: 100})')
+    children_count = feed_el.evaluate(
+        '(element) => element.shadowRoot.querySelectorAll(".bk-panel-models-markup-HTML").length'
+    )
+    assert children_count == 58
```
ahuang11, line 34:
```suggestion
    assert children_count == 61
```

panel/tests/ui/layout/test_feed.py
```diff
@@ -0,0 +1,94 @@
...
+    bbox = feed_el.bounding_box()
+    assert bbox["height"] == 250
+
+    expect(feed_el).to_have_class("bk-panel-models-feed-Feed scrollable-vertical")
+
+    # Assert scroll is not at 0 (view_latest)
+    assert feed_el.evaluate('(el) => el.scrollTop') > 0
+
+    last_pre_element = page.query_selector_all('pre')[-1]
+    assert last_pre_element.inner_text() == "999"
```
ahuang11, line 61:
```suggestion
    wait_until(
        lambda: int(last_pre_element.inner_text()) == 999
    )
```

panel/tests/ui/layout/test_feed.py
```diff
@@ -0,0 +1,96 @@
...
+    # assert scroll button is visible on render
+    scroll_arrow = page.locator(".scroll-button")
+    expect(scroll_arrow).to_have_class('scroll-button visible')
+    expect(scroll_arrow).to_be_visible()
+
+    # click on scroll arrow
+    scroll_arrow.click()
+
+    # Assert scroll is not at 0 (view_latest)
+    assert feed_el.evaluate('(el) => el.scrollTop') > 0
```
ahuang11, line 81:
```suggestion
    wait_until(lambda: feed_el.evaluate('(el) => el.scrollTop') > 0)
```

panel/tests/ui/layout/test_feed.py
```diff
@@ -0,0 +1,96 @@
...
+    assert bbox["height"] == 250
+
+    expect(feed_el).to_have_class("bk-panel-models-feed-Feed scrollable-vertical")
+
+    # Assert scroll is not at 0 (view_latest)
+    assert feed_el.evaluate('(el) => el.scrollTop') > 0
+
+    last_pre_element = page.query_selector_all('pre')[-1]
+    wait_until(
+        lambda: int(last_pre_element.inner_text()) == 999
```
ahuang11, line 62:
```suggestion
        lambda: int(last_pre_element.inner_text()) > 950
```

panel/tests/ui/layout/test_feed.py
```diff
@@ -0,0 +1,96 @@
...
+
+    bbox = feed_el.bounding_box()
+    assert bbox["height"] == 250
+
+    expect(feed_el).to_have_class("bk-panel-models-feed-Feed scrollable-vertical")
+
+    children_count = feed_el.evaluate(
+        '(element) => element.shadowRoot.querySelectorAll(".bk-panel-models-markup-HTML").length'
+    )
+    assert children_count == 58
```
ahuang11, line 27:
```suggestion
    assert 50 <= children_count <= 60
```

panel/tests/ui/layout/test_feed.py
```diff
@@ -0,0 +1,96 @@
...
+
+    # Now scroll to somewhere down
+    feed_el.evaluate('(el) => el.scrollTo({top: 100})')
+    children_count = feed_el.evaluate(
+        '(element) => element.shadowRoot.querySelectorAll(".bk-panel-models-markup-HTML").length'
+    )
+    assert children_count == 61
```
ahuang11, line 34:
```suggestion
    assert 50 <= children_count <= 65
```

panel/tests/ui/layout/test_feed.py
```diff
@@ -0,0 +1,96 @@
...
+
+    # Now scroll to top
+    feed_el.evaluate('(el) => el.scrollTo({top: 0})')
+    wait_until(
+        lambda: feed_el.evaluate(
+            '(element) => element.shadowRoot.querySelectorAll(".bk-panel-models-markup-HTML").length'
+        )
+        == 58
```
ahuang11, line 42:
```suggestion
        >= 50
```

ahuang11:
I had to cover some edge cases where:
1. the number of visible objects was less than the load buffer and it would never buffer again, trapping the user
2. the visible range exceeded the number of objects available which caused all objects to disappear

The trapping in action:
https://github.com/holoviz/panel/assets/15331990/50b6a502-6c66-4b51-8b15-fc483848a428

PR #6036, opened 2023-12-13
Fix pandas warning when sorting filtered tabulator

TBym:
When a Tabulator with applied filter(s) is sorted the code yields a warning from pandas "A value is trying to be set on a copy of a slice from a DataFrame".
This changes fixes the ussie by first sorting Tabulator data and then applying filter. Alternative is to create a copy of passed dataframe to _sort_df function.


TBym:
If you look to my branch I made a second commit fixing a bug in Tabulator selected_dataframe function. I tried to create two separate pull requests but it seems that I can create only single request for all my changes.

cdeil:
@TBym - do you have a minimal reproducer for the warning and issue you were getting?

Is the issue resolved in the meantime?

There have been a few tabulator fixes in the meantime, wee https://holoviz-dev.github.io/panel/about/releases.html or try with
```
pip install -U panel --pre
```

Issue #5628, opened 2023-10-12
HoloViews Holomap widgets not rendered when in a layout


maximlt:
When a HoloViews HoloMap is wrapped in a HoloViews pane that is passed to a layout, the widgets are not rendered.

```python
import holoviews as hv
import panel as pn

from bokeh.sampledata.penguins import data as df

hv.extension('bokeh')

ds = hv.Dataset(df, kdims=['bill_depth_mm', 'bill_length_mm', 'species'])
plot = ds.to(hv.Points, kdims=['bill_depth_mm', 'bill_length_mm'])
# pane = pn.pane.HoloViews(plot, sizing_mode='stretch_width')
pane = pn.pane.HoloViews(plot)

pane.servable()                            # OK: Widgets displayed
pn.Row('Test Row', pane).servable()        # Not OK: Widgets missing
pn.Column('Test Column', pane).servable()  # Not OK: Widgets missing
```

![image](https://github.com/holoviz/panel/assets/35924738/249bf675-1c7e-472f-b329-51322ed973bb)

maximlt:
No errors in the console to be found.

philippjfr:
I appreciate that this is pretty weird, but a HoloViews pane rendered on its own does not include the widgets (and never has). The simplest way to get the full layout is:

```python
pn.Row('Test Row', pane.layout).servable()
```

or alternatively:

```python
pn.Row('Test Row', pane, pane.widgets).servable()
```

Not opposed to finding a way to make this behave reasonably but suspect it might be difficult.

maximlt:
Ok. I had no clue about that. I still don't understand what's the difference between my first example (`pane.servable()                            # OK: Widgets displayed`) and the other ones?

I thought it was a Panel/Bokeh/HoloViews bug. Apparently it's not so I'll have to make changes to hvPlot since now a Dynamic map doesn't render with its widgets in the Explorer (I guess it's because now the explorer preview is now in a Tabs layout).

philippjfr:
>I still don't understand what's the difference between my first example (pane.servable() # OK: Widgets displayed) and the other ones?

Yeah, definitely quirky and hard to reason about. It's internal abstractions leaking. Internally Panel will unpack layouts if they are placed inside another layout but if they are rendered as a root then there's no layout that it could update and it ends up rendering the entire `.layout`. For most other components the `.layout` is simply a Column containing the one item that is being updated but for `HoloViews` it's the layout of the plot and the widgets.

[...]

PR #6063, opened 2023-12-16
Fix HoloViews.layout issue

MarcSkovMadsen:
Fixes `HoloViews.layout` issue described in https://github.com/holoviz/panel/issues/5628#issuecomment-1858729895.

# Todo.

- [x] Add test cases
- [ ] Get Philipp to review test cases
- [ ] Find issue
- [ ] Implement solution

### DIFF
--- a/panel/tests/pane/test_holoviews.py
+++ b/panel/tests/pane/test_holoviews.py
@@ -743,3 +743,30 @@ def test_holoviews_property_override(document, comm):
 
     assert model.styles["background"] == 'red'
     assert model.children[0].css_classes == ['test_class']
+
+widget = pn.widgets.IntInput(value=1)
+def _plot(value):
+    return hv.Curve(list(range(value)))
+plot = pn.bind(_plot, widget)
+
+@hv_available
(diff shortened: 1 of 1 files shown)


panel/tests/pane/test_holoviews.py
```diff
@@ -743,3 +743,30 @@ def test_holoviews_property_override(document, comm):
...
     assert model.styles["background"] == 'red'
     assert model.children[0].css_classes == ['test_class']
+
+widget = pn.widgets.IntInput(value=1)
+def _plot(value):
+    return hv.Curve(list(range(value)))
+plot = pn.bind(_plot, widget)
+
+@hv_available
+def test_holoviews_layout_objects(document, comm):
```
MarcSkovMadsen, line 753: Please review the below test cases. If you agree that the layout should not contain the `widget_layout` when its empty. Then we should try to find fix. If you know where the fix should be applied I would appreciate you tell me or implement the fix. Thanks.

MarcSkovMadsen:
This still needs your input @philippjfr

PR #6065, opened 2023-12-16
Improve HoloViews reference guide

MarcSkovMadsen:
Via https://github.com/holoviz/panel/issues/5628 its clear that the `HoloViews` reference notebook does not clearly explain how things are expected to work. After working with it I can also see that its not updated to reflect how the `HoloViews` pane works today. And there are also a issues (bugs).

### DIFF
--- a/examples/reference/panes/HoloViews.ipynb
+++ b/examples/reference/panes/HoloViews.ipynb
@@ -6,33 +6,62 @@
    "metadata": {},
    "outputs": [],
    "source": [
+    "import holoviews as hva\n",
     "import panel as pn\n",
-    "pn.extension('plotly')"
+    "\n",
+    "hv.extension(\"bokeh\") # DON'T KNOW WHY THIS WAS SUDDENLY NEEDED. PLEASE HELP\n",
+    "pn.extension('plotly') # We add 'plotly' to be able to use the 'plotly' plot backend"
    ]
   },
   {
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "The ``HoloViews`` pane renders HoloViews plots with one of the plotting backends supported by HoloViews. It supports the regular HoloViews widgets for exploring the key dimensions of a ``HoloMap`` or ``DynamicMap``, but is more flexible than the native HoloViews widgets since it also allows customizing widget types and their position relative to the plot.\n",
+    "[HoloViews](https://holoviews.org/) is a very powerful data visualization library supporting many data and plotting *backends*.\n",
+    "\n",
+    "[hvPlot](https://hvplot.holoviz.org/index.html) (quick viz) and [GeoViews](https://holoviz.org/assets/geoviews.png) (spatial viz) are built on top of HoloViews and produces `HoloViews` objects.\n",
+    "\n",
+    "**Panel, HoloViews, hvPlot and GeoViews are all members of the [HoloViz](https://holoviz.org) ecosystem and you can expect them to work perfectly together**.\n",
+    "\n",
+    "<img src=\"https://holoviz.org/assets/holoviews.png\" style=\"height:50px;margin-right:10px\"> <img src=\"https://holoviz.org/assets/hvplot.png\" style=\"height:50px;margin-right:10px\"> <img src=\"https://holoviz.org/assets/geoviews.png\" style=\"height:50px;margin-right:10px\"> <img src=\"https://holoviz.org/assets/panel.png\" style=\"height:50px;margin-right:10px\">\n",
+    "\n",
+    "In this reference notebook we will assume a basic level of understanding of HoloViews and optionally hvPlot or Geoviews if you want to use them with Panel.\n",
+    "\n",
+    "---\n",
+    "\n",
+    "The `HoloViews` pane renders [HoloViews](https://holoviews.org/) objects with one of the *plotting backends* supported by HoloViews. This includes objects produced by [hvPlot](https://hvplot.holoviz.org/index.html) and [GeoViews](https://holoviz.org/assets/geoviews.png).\n",
+    "\n",
+    "The `HoloViews` pane supports displaying interactive [`HoloMap`](https://holoviews.org/reference/containers/bokeh/HoloMap.html) and [`DynamicMap`](https://holoviews.org/reference/containers/bokeh/DynamicMap.html) objects containing widgets. The `HoloViews` pane even allows customizing the widget types and their position relative to the plot.\n",
     "\n",
     "#### Parameters:\n",
     "\n",
     "For details on other options for customizing the component see the [layout](../../how_to/layout/index.md) and [styling](../../how_to/styling/index.md) how-to guides.\n",
     "\n",
-    "* **``backend``** (str): Any of the supported HoloViews backends ('bokeh', 'matplotlib', or 'plotly')\n",
-    "* **``center``** (boolean, default=False): Whether to center the plot\n",
-    "* **``linked_axes``** (boolean, default=True): Whether to link axes across plots in a panel layout\n",
+    "The main argument is the `object` parameter\n",
+    "\n",
     "* **``object``** (object): The HoloViews object being displayed\n",
+    "\n",
+    "You can control the way the `object` is rendered and sharing axes with other plots via the parameters\n",
+    "\n",
+    "* **``backend``** (str): Any of the supported HoloViews backends ('bokeh', 'matplotlib', or 'plotly'). If none is specified defaults to the active holoviews renderer. This is by default 'bokeh'.\n",
+    "* **``linked_axes``** (boolean, default=True): Whether to link axes across plots in a panel layout\n",
+    "* **``renderer``**: Explicit [HoloViews Renderer](https://holoviews.org/user_guide/Plots_and_Renderers.html#renderers) instance to use for rendering the HoloViews  plot. Overrides the` backend` parameter.\n",
+    "* **``theme`` (str, Theme)**: [Bokeh theme](https://docs.bokeh.org/en/latest/docs/reference/themes.html) to apply to the HoloViews plot.\n",
+    "\n",
+    "You can access the layout and (optional) widgets via\n",
+    "\n",
+    "* **``layout``** (pn.layout.Panel): The layout containing the plot pane and (optionally) the `widget_box` layout.\n",
+    "* **``widget_box``** (ListPanel): The layout containing the widgets\n",
+    "\n",
+    "##### Layout and Widget Parameters\n",
+    "\n",
+    "The below parameters are used to control the layout and widgets using `pn.panel` or `pn.pane.HoloViews(...).layout`.\n",
+    "\n",
+    "* **``center``** (boolean, default=False): Whether to center the plot or not.\n",
+    "* **``widgets``** (dict, argument): A mapping from dimension name to a widget class, instance, or dictionary of overrides to modify the default widgets. Provided as an argument.\n",
     "* **``widget_location``** (str): Where to lay out the widget relative to the plot \n",
-    "* **``widget_layout``** (ListPanel type): The object to lay the widgets out in, one of ``Row``, ``Column`` or ``WidgetBox``\n",
+    "* **``widget_layout``** (ListPanel): The object to lay the widgets out in, one of ``Row``, ``Column`` or ``WidgetBox``\n",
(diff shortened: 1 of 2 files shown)


examples/reference/panes/HoloViews.ipynb
```diff
@@ -6,41 +6,70 @@
    "metadata": {},
    "outputs": [],
    "source": [
+    "import holoviews as hv\n",
     "import panel as pn\n",
-    "pn.extension('plotly')"
+    "\n",
+    "hv.extension(\"bokeh\")\n",
```
MarcSkovMadsen, line 12: When I changed the notebook examples suddenly they started complaining that I needed to run `hv.extension(...)`. So I did. But I don't know why its needed now and not before??

examples/reference/panes/HoloViews.ipynb
```diff
@@ -119,18 +207,205 @@
...
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "Please note that in a *server context* you might have to set the matplotlib backend like below depending on your setup.\n",
+    "\n",
+    "```python\n",
+    "import matplotlib\n",
```
MarcSkovMadsen, line 221: Is the comment below still needed?

```python
import matplotlib
matplotlib.use('agg')
```

I did never do that. And it still works.

examples/reference/panes/HoloViews.ipynb
```diff
@@ -119,18 +207,205 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "widgets = hv_panel[1]\n",
+    "plot = df.hvplot.scatter(x=\"group\", y=\"value\")\n",
+    "pn.pane.HoloViews(plot, backend='matplotlib', sizing_mode=\"stretch_width\", height=300)"
```
MarcSkovMadsen, line 211: I would like to add an example where the matplotlib plot stretches responsively. As that is what I concretely would use in applications and when I try to communicate the message that hvPlot/ HoloViews is pretty flexible because you can replace bokeh with matplotlib or plotly easily.

How do I do that?

examples/reference/panes/HoloViews.ipynb
```diff
@@ -119,18 +207,205 @@
...
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "hvplot.extension(\"plotly\")\n",
+    "\n",
+    "plot = df.hvplot.scatter(x=\"group\", y=\"value\", height=300, responsive=True)\n",
+    "pn.pane.HoloViews(plot, backend='plotly', height=300)"
```
MarcSkovMadsen, line 253: There is a **bug** here. The resulting plot is higher than the `300px` specified both with hvPlot and HoloViews.

image

examples/reference/panes/HoloViews.ipynb
```diff
@@ -119,18 +207,205 @@
...
+    "You can change the plotting backend dynamically via a widget too."
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "plot = df.hvplot.scatter(x=\"group\", y=\"value\", height=300, responsive=True, title=\"Try changing the backend\")\n",
```
MarcSkovMadsen, line 287: This example is the demonstrating the core of HoloViews to me. That I can easily output to bokeh, matplotlib and plotly.

Unfortunately it does not really work.

- Only the bokeh plot has the `title`.
- The matplotlib plot does not stretch_width like the other plots
- The plotly plot does not stretch_width. And its height is higher than the 300px specified both with hvPlot and HoloViews.


https://github.com/holoviz/panel/assets/42288570/ec3bd862-e374-476d-ae2c-ad403cc26653

examples/reference/panes/HoloViews.ipynb
```diff
@@ -119,18 +207,205 @@
...
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "plot = df.hvplot.scatter(x=\"group\", y=\"value\", height=300, responsive=True)\n",
+    "pn.pane.HoloViews(plot, height=300, theme=\"night_sky\")"
```
MarcSkovMadsen, line 335: The `theme` parameter does not work (in VS Code). This is an already reported issue.

image

examples/reference/panes/HoloViews.ipynb
```diff
@@ -119,18 +207,205 @@
...
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "plot = df.hvplot.scatter(x=\"group\", y=\"value\", height=300, width=400)\n",
     "\n",
     "pn.Column(\n",
-    "    pn.Row(*widgets),\n",
-    "    hv_panel[0])"
+    "    \"## Not Centered: `center=False`\",\n",
```
MarcSkovMadsen, line 362: Does `center` work as expected when combined with `sizing_mode`? I don't think so. But I can find a way to achieve what I need.

image

examples/reference/panes/HoloViews.ipynb
```diff
@@ -139,24 +414,91 @@
...
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "float_slider = hv_panel[1][0]\n",
+    "float_slider.name=\"Frequency (updated)\"\n",
+    "float_slider"
```
MarcSkovMadsen, line 435: When I change the value of the floatslider it raises an exception. I don't know why and need help.

image

examples/reference/panes/HoloViews.ipynb
```diff
@@ -139,24 +414,91 @@
...
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "widgets = hv_panel[1]\n",
+    "\n",
+    "pn.Row(pn.Column(*widgets), hv_panel[0])"
```
MarcSkovMadsen, line 453: When I reorganize the layout it raises an exception.

image

examples/reference/panes/HoloViews.ipynb
```diff
@@ -139,24 +414,91 @@
...
+   ]
+  },
   {
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "#### Customizing widgets\n",
+    "#### `widget_layout`\n",
+    "\n",
+    "Lets change the widget layout to a `Row`."
```
MarcSkovMadsen, line 485: When I change the `widget_layout` to a row, then the widgets are not nicely aligned.

image

If you inspect the object you can see its because some custom margin that works in a Column but not Row has been added.

```bash
Column(sizing_mode='stretch_width')
    [0] Row
        [0] FloatSlider(end=10, margin=(20, 20, 5, 20), name='frequency', start=0.1, value=0.1, width=250)
        [1] IntSlider(end=10, margin=(0, 20, 5, 20), name='amplitude', start=1, value=1, width=250)
        [2] Select(margin=(5, 20, 20, 20), name='function', options=['sin', 'cos', 'tan'], value='sin', width=250)
    [1] HoloViews(DynamicMap, center=True, widget_layout=<class 'panel.layout.base...., widget_location='top_left')
```

examples/reference/panes/HoloViews.ipynb
```diff
@@ -197,36 +540,23 @@
...
+    "widget = pn.widgets.IntSlider(value=50, start=1, end=100, name=\"Number of points\")\n",
+    "hv_pane = pn.pane.HoloViews(hv.DynamicMap(pn.bind(get_plot, widget)),)\n",
     "\n",
-    "```python\n",
-    "import matplotlib\n",
-    "matplotlib.use('agg')\n",
-    "```\n",
-    "\n",
-    "The ``backend``, like all other parameters, can be modified after the fact.  To demonstrate, we can set up a select widget to toggle between backends for the above plot:"
+    "pn.Column(widget, hv_pane)"
```
MarcSkovMadsen, line 552: I don't understand why the widgets of the `DynamicMap` are not shown here. Its quite inconsistant that the widgets are shown if the `DynamicMap` is created via HoloViews redim as shown in the top example. But not when created with Panel bind.

image

What is the *model* for understanding this?

Wrapping a bound function in a DynamicMap has been a *secret* among a few super users for a very long time. Please help me explain why this is beneficial. Thanks.

MarcSkovMadsen:
There are so many issues here. I need an expert on HoloViews (the framework) and HoloViews (the pane) to go through my comments and give me some feedback on who should do what.

I don't plan to do more before I get that feedback.

examples/reference/panes/HoloViews.ipynb
```diff
@@ -6,41 +6,70 @@
    "metadata": {},
```
MarcSkovMadsen, line 6: I'm not able to test this notebook in Jupyterlab because I started getting.

image

I've tried everything to get rid of it. But cannot.

panel==1.3.4, bokeh==3.3.2, param=2.0.1, holoviews==1.18.1, hvplot=0.9.0

MarcSkovMadsen:
Maybe I forgot to "release this". I've done that now. It is ready for review.

philippjfr:
I'll retarget for 1.4.0 because I don't want to hold up the release much more, but if I do get a chance before tomorrow I will backport it.

examples/reference/panes/HoloViews.ipynb
philippjfr: This is because you changed the name of the widget (which is not allowed). I will force the name to be constant.

examples/reference/panes/HoloViews.ipynb
philippjfr: Fixed.

examples/reference/panes/HoloViews.ipynb
philippjfr: Good question, the reason is historical. Originally HoloViews always constructed widgets for the dimensions internally (even before Panel). When we added `streams` in HoloViews it became possible to bind other arguments to the functions on a `DynamicMap` but those were generally obtained from actual `Stream` objects, such as the `RangeXY` or `Tap`. At some point it became possible to directly bind parameters using the `Param` stream and as syntactic sugar it became possible for HoloViews to accept functions annotated with `depends` or `bind`, however we didn't go the step further and inspect what the bound parameters or dependencies were and then create widgets for them. I agree that you could argue that this should be done but I'll leave that as an exercise for later and put some explanation of the current behavior into the reference docs now.

examples/reference/panes/HoloViews.ipynb
philippjfr: Can't reproduce anymore.

examples/reference/panes/HoloViews.ipynb
philippjfr: Indeed, this would be very nice. Thanks for opening this issue: https://github.com/holoviz/hvplot/issues/1236 where we will track this going forward.

examples/reference/panes/HoloViews.ipynb
philippjfr: You can now set `format='svg'` and then get responsive plots.

examples/reference/panes/HoloViews.ipynb
```diff
@@ -96,21 +146,59 @@
...
     "    xs = np.arange(200)/200*20.0\n",
     "    ys = amplitude*getattr(np, function)(frequency*xs)\n",
-    "    return pd.DataFrame(dict(y=ys), index=xs).hvplot()\n",
+    "    return pd.DataFrame(dict(y=ys), index=xs).hvplot(height=250, responsive=True)\n",
     "\n",
     "dmap = hv.DynamicMap(sine, kdims=['frequency', 'amplitude', 'function']).redim.range(\n",
     "    frequency=(0.1, 10), amplitude=(1, 10)).redim.values(function=['sin', 'cos', 'tan'])\n",
     "\n",
-    "hv_panel = pn.panel(dmap)\n",
+    "hv_panel = pn.pane.HoloViews(dmap)\n",
```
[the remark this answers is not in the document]
philippjfr: Let's address this one separately.

examples/reference/panes/HoloViews.ipynb
philippjfr: Changed it to load both extensions with `hv.extension()`

philippjfr (approved):

Issue #6064, opened 2023-12-16
Creating ChatMessages is unreasonably slow


EllenOrange:
#### ALL software version info
WSL Ubuntu 
Python 3.11.6
Panel 1.3.4
Bokeh 3.3.2

#### Description of expected behavior and the observed behavior
It seems as though creating a pn.chat.ChatMessage takes ~0.3 seconds per message (!). This means if I load a history with 100 messages, the load time goes up by nearly 30 seconds, which makes for an unreasonably slow development experience.

In the example code I create the objects manually and then set the .objects on the ChatFeed, but I've verified that this also occurs if I send the messages one by one.

It's very possible I'm missing something simple here, but I've followed the available instructions as closely as I can manage.

#### Complete, minimal, self-contained example code that reproduces the issue

```
import panel as pn
pn.extension(template="fast")
from pydantic import BaseModel
from typing import List
import time

class HistoryMessage(BaseModel):
    role: str
    content: str

number_of_messages = 100
history_messages = [HistoryMessage(role='user', content='Test string') for _ in range(number_of_messages)]

start_time = time.perf_counter()
chatmessages = [pn.chat.ChatMessage(hm.content, user=hm.role) for hm in history_messages]
end_time = time.perf_counter()

execution_time = end_time - start_time
print(f"Create {number_of_messages} messages = {execution_time} seconds")

chatfeed = pn.chat.ChatFeed()
chatfeed.objects = chatmessages

chatfeed.servable()
```

#### Screenshots or screencasts of the bug in action
(test_app) elean@DESKTOP-290LANC:~/perforce/mainline/test_app$  cd /home/elean/perforce/mainline/test_app ; /usr/bin/env /home/elean/miniforge3/envs/test_app/bin/python /home/elean/.vscode-server/extensions/ms-python.python-2023.22.1/pythonFiles/lib/python/debugpy/adapter/../../debugpy/launcher 42127 -- -m panel serve app.py --show --autoreload 
Create 100 messages = 29.158785282983445 seconds
2023-12-16 07:56:31,763 Starting Bokeh server version 3.3.2 (running on Tornado 6.4)
2023-12-16 07:56:31,764 User authentication hooks NOT provided (default user enabled)
2023-12-16 07:56:31,766 Bokeh app running at: http://localhost:5006/app
2023-12-16 07:56:31,766 Starting Bokeh server with process id: 612354

For completeness, I just ran the same case outside of debugpy and got:
$ panel serve app.py --show --autoreload
[...]

PR #6069, opened 2023-12-17
Optimize rendering of ChatMessage

philippjfr:
Attempt at improving https://github.com/holoviz/panel/issues/6064

```python
%%timeit
pn.chat.ChatMessage('Some content', user='user')
```

Before:

> 19.6 ms ± 591 µs per loop (mean ± std. dev. of 7 runs, 10 loops each

After:

> 10.9 ms ± 138 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)

The main performance gains here came from avoiding use of `ParamMethod` and instead favoring manual watch callbacks.

Note those numbers also already account for the changes in: https://github.com/holoviz/param/pull/893

Without that it goes from:

> 21.6 ms ± 974 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)

to:

> 11.7 ms ± 280 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)

### DIFF
--- a/panel/chat/message.py
+++ b/panel/chat/message.py
@@ -208,8 +208,6 @@ class ChatMessage(PaneBase):
     user = param.Parameter(default="User", doc="""
         Name of the user who sent the message.""")
 
-    _object_panel = param.Parameter(allow_refs=False, doc="The rendered object panel.")
-
     _stylesheets: ClassVar[List[str]] = [f"{CDN_DIST}css/chat_message.css"]
 
     # Declares whether Pane supports updates to the Bokeh model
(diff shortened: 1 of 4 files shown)


ahuang11:
There seems to be some weird indentation after these changes:
image

After removing width 100% - 15px, it's back to normal
image

ahuang11:
Some flickering seems to occur again:

https://github.com/holoviz/panel/assets/15331990/bea6b92b-15d3-4ff6-97bf-d3461250a015

On main:
https://github.com/holoviz/panel/assets/15331990/97ad3453-48a1-41f9-8607-b5caa13293f1

PR #6072, opened 2023-12-18
Support for Bokeh 3.4

Hoxbro:
I have chosen to hard-pin the build step. This can always be reversed.

- `selectbox` file was renamed to `select` in https://github.com/bokeh/bokeh/pull/13497.
- `InputWidgetView._render_input()` was added in https://github.com/bokeh/bokeh/pull/13604
- `URL` dataclass was added in https://github.com/bokeh/bokeh/pull/13042
- `ColorMap` was renamed to `PaletteSelect` in https://github.com/bokeh/bokeh/pull/13537


panel/io/resources.py
```diff
@@ -600,6 +602,8 @@ def adjust_paths(self, resources):
         new_resources = []
         cdn_base = f'{config.npm_cdn}/@holoviz/panel@{JS_VERSION}/dist/'
         for resource in resources:
+            if not isinstance(resource, str):  # Bokeh 3.4.0.dev4 uses bokeh.embed.bundle.URL
+                resource = resource.url
```
droumis, line 606:
```suggestion
                resource = str(resource)
```

This can come back as a PosixPath instead of a bokeh.embed.bundle.URL, in which case it does not have attr .url, so maybe just do this.

panel/models/plotly.ts
```diff
@@ -316,7 +316,7 @@ export class PlotlyPlotView extends HTMLBoxView {
   }
 
   _get_trace(index: number, update: boolean): any {
-    const trace = clone(this.model.data[index]);
+    const trace = clone(this.model.data[index]) as any;
```
Hoxbro, line 319: Else, I would give a type error.

panel/tests/io/test_reload.py
```diff
@@ -40,6 +40,7 @@ def test_watch():
     _watched_files.clear()
 
 @pytest.mark.flaky(reruns=3)
+@pytest.mark.asyncio
```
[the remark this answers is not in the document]
Hoxbro: It seems to make Windows fail, so I have commented out the line to be sure.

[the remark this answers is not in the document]
Hoxbro: It does have removed the line for now.

panel/io/resources.py
```diff
@@ -383,7 +391,7 @@ def bundle_resources(roots, resources, notebook=False, reloading=False, enable_m
     extensions = _bundle_extensions(None, js_resources)
     if reloading:
         extensions = [
-            ext for ext in extensions if not (ext.cdn_url is not None and ext.cdn_url.startswith('https://unpkg.com/@holoviz/panel@'))
+            ext for ext in extensions if not (str(ext.cdn_url) is not None and str(ext.cdn_url).startswith('https://unpkg.com/@holoviz/panel@'))
```
Hoxbro, line 394:
```suggestion
            ext for ext in extensions if not (ext.cdn_url is not None and str(ext.cdn_url).startswith('https://unpkg.com/@holoviz/panel@'))
```

setup.py
```diff
@@ -235,7 +235,7 @@ def run(self):
     'setuptools >=42',
     'requests',
     'packaging',
-    'bokeh >=3.3.0,<3.4.0',
+    "bokeh ==3.4.0.dev4",
```
Hoxbro, line 238:
```suggestion
    "bokeh ==3.4.0.dev6",
```

setup.py
```diff
@@ -235,7 +235,7 @@ def run(self):
     'setuptools >=42',
     'requests',
     'packaging',
-    'bokeh >=3.3.0,<3.4.0',
+    "bokeh ==3.4.0.dev6",
```
Hoxbro, line 238:
```suggestion
    'bokeh ==3.4.0.dev6',
```

Hoxbro:
I can see the UI tests are starting to fail between Bokeh 3.4.0.dev4 and 3.4.0.dev5, this is likely to be because of https://github.com/bokeh/bokeh/pull/13618, @mattpap do you agree? The PR is pretty big, and I can't see if it is backward compatible, with older versions of Bokeh. 

A quick glance shows the following models are failing:
- Tabulator
- Plotly and Vega plots
- ReactiveHTML

philippjfr:
Suspect this is due to the dict -> object/Map conversion. It used to be the case that most dicts would end up being converted to a JS `Map` but now you will end up with a standard JS object. Suspect those models will have to be updated.

mattpap:
> The PR is pretty big, and I can't see if it is backward compatible, with older versions of Bokeh.

It's not compatible on the implementation level. Previously if there was a `dict` in bokeh, then it was deserialized in bokehjs either as a plain object `{}`, if all keys were strings, or as `Map` in all other cases. Now `dict` is always deserialized as a `Map`. However, in all contexts, in particular all properties, that previously supported plain objects, now both plain objects and `Map` are supported. This is to keep API compatibility, while allowing preservation of object identity (no cloning) and performance (no proxying).

So what needs to happen, is that all cases where plain objects are used, instead of manipulate those objects directly, one should use APIs from `core/util/object`, like `keys()`, `values()`, `entries()`, `dict()`, `to_object()`, etc. Also don't ever use `for .. in`, but use `for (const [key, vall] of entries(plain_object_or_map))`.

This comes with some caveats. This approach breaks pseudo structs that are encoded as dictionaries, see e.g. https://github.com/bokeh/bokeh/issues/13637. That issue is a critical regression to fix to make 3.4 viable. Otherwise I will need to revert PR bokeh/bokeh#13618. If you previously passed a plain object from a property to a third-party JS API, then now you have to use `to_object()` to make sure it's still a plain object.

Hoxbro:
Thank you, @mattpap, for the explanation. 

Can you give an estimated time for when you look at the regression? (No pressure, just curious)

@philippjfr, I will wait with this PR until the regression or reversion has been made in Bokeh.

mattpap:
> Can you give an estimated time for when you look at the regression? (No pressure, just curious)

As soon as I'm done with UIs, so hopefully tomorrow.

philippjfr:
Note I don't think the regression affects Panel, unless I'm misunderstanding.

Hoxbro:
> Note I don't think the regression affects Panel, unless I'm misunderstanding.

I agree, but if the PR is reverted, there is no need to update Panel to support it.

mattpap:
@Hoxbro, I reverted dict/Map related changes, upgraded to dev8, fixed `tsconfig.json` (as mentioned in the meeting) and renamed one bokeh API call (`rebuild` -> `update_children`).

mattpap:
I will publish removed commits in an new PR when this one is merged.

panel/models/reactive_html.ts
```diff
@@ -138,7 +138,7 @@ export class ReactiveHTMLView extends HTMLBoxView {
     super.connect_signals()
     this.connect(this.model.properties.children.change, async () => {
       this.html = htmlDecode(this.model.html) || this.model.html
-      await this.rebuild()
+      await this.update_children()
```
philippjfr, line 141: Can you explain this? It's breaking things.

philippjfr: I see, `rebuild` was removed.

mattpap: I didn't realize panel was using this API, which was used in bokehjs only once and was duplicating work unnecessarily. That's why we need to test panel continuously (at least whether it builds) in bokeh's CI.

philippjfr (approved):

PR #6079, opened 2023-12-18
Remove contourpy fro pyodide again

Hoxbro:
I was too eager with the removal of contourpy here. 

We are using an old version of pyodide which did not package a newer version of contourpy.


philippjfr:
Should be able to resolve this soon by upgrading to recent PyScript.

philippjfr (approved):

PR #6081, opened 2023-12-18
Rename `unix_only` to `linux_only`

Hoxbro:
image


Hoxbro:
It only runs on Linux. :+1:

Hoxbro:
I thought `unix_only` meant both Linux and macOS, but it is only Linux.  So I will reuse the PR to rename `unix_only` to `linux_only`.

philippjfr (approved):

PR #6084, opened 2023-12-18
Auth: execute async scheduled refresh callback

maximlt:
`refresh_cb` is a partial built from the `_scheduled_refresh` async method, it cannot be called directly.


philippjfr:
Looks like 1.3.6 is in the near future :)

philippjfr (approved):

PR #6092, opened 2023-12-19
Add explicit size option to ToggleIcon

philippjfr:
`ToggleIcon` now allows either `width`/`height` or an explicit size to be defined.

### DIFF
--- a/examples/reference/widgets/ToggleIcon.ipynb
+++ b/examples/reference/widgets/ToggleIcon.ipynb
@@ -14,7 +14,7 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "The ``ToggleIcon`` widget allows toggling a single condition between ``True``/``False`` states. This widget is interchangeable with the ``Checkbox`` widget.\n",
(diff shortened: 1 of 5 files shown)


examples/reference/widgets/ToggleIcon.ipynb
```diff
@@ -93,8 +94,23 @@
...
-    "toggle = pn.widgets.ToggleIcon(icon=\"thumb-down\", active_icon=\"thumb-up\")\n",
-    "toggle"
+    "pn.widgets.ToggleIcon(icon=\"thumb-down\", active_icon=\"thumb-up\")"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "The icon will automatically adapt to the specified `width`/`height` but you may also provide an explicit `size`:"
```
ahuang11, line 104:
```suggestion
    "The icon will automatically adapt to the specified `width`/`height` but you may also provide an explicit `size` which takes precedence:"
```

ahuang11 (approved):
Tested and looks good to me! Thanks for cleaning up the notebook too.

PR #6111, opened 2023-12-21
Update bootstrap.html - Changing navbar toggle icon to white

mayonnaisecolouredbenz7:
Changing the navbar toggle icon to white as is the case in other templates.  

I attempted to create the variable in the css file that the template is referring to originally _bs-navbar-toggler-icon-bg_.  However it wasn't overriding the variable shown in the console:
image

So it was easier to explicitly change the background-image property in the navbar-toggler-icon class in html file


philippjfr:
pre-commit.ci autofix

philippjfr:
>I attempted to create the variable in the css file that the template is referring to originally bs-navbar-toggler-icon-bg.

I suspect the problem is that you have to run a command to bundle the CSS files when you want them to be reflected using `panel bundle --templates --only-local` (which is not documented). Overall I'd prefer using the variable.

mayonnaisecolouredbenz7:
Made the necessary changes

panel/template/bootstrap/bootstrap.html
```diff
@@ -60,10 +60,11 @@
       }
       #header {
 	--bs-body-bg: var(--header-background);
+
```
philippjfr, line 63:
```suggestion
```

philippjfr (approved):

philippjfr:
Thank you @mayonnaisecolouredbenz7!

PR #6119, opened 2023-12-22
Dynamic generate unit test matrix and add cache option

Hoxbro:
See https://github.com/holoviz/holoviews/pull/6043 for more details.

Built on top of https://github.com/holoviz/panel/pull/6096

image

If needed, I can also make it so that the UI test for the downstream option only runs on one OS.


Hoxbro:
To be extra clear, this was merged into #6096. If you don't want #6096 merged, this PR will not go into main

Issue #6118, opened 2023-12-22
ChatInterface Regression:  value must be an instance of Future, not Future pending


MarcSkovMadsen:
I just upgraded from Panel 1.3.4 to 1.3.6. I did not change my code. I get the exception.

![image](https://github.com/holoviz/panel/assets/42288570/b725f051-9ad8-4e4d-8dd8-687a751ad056)

```bash
Encountered ValueError("ClassSelector parameter 'ChatInterface._callback_future' value must be an instance of Future, not <Future pending cb=[_chain_future.<locals>._call_check_cancel() at /opt/conda/lib/python3.11/asyncio/futures.py:387]>."). Set callback_exception='verbose' to see the full traceback.
```

If I downgrade to Panel 1.3.4 it works again.

I will make a minimum, reproducible example when I get back to work.

philippjfr:
Guess we're having a 1.3.7, @ahuang11 any ideas?

MarcSkovMadsen:
My work app is working in the same way as https://github.com/run-llama/llama-hub/blob/e621be95150d15ae9048c1a521c5f5c9821edbb4/llama_hub/llama_packs/panel_chatbot/app.py. Just populating the vector store from a local folder instead of github.

ahuang11:
The easy fix would be changing callback future into param.Parameter since I don't see the difference between future pending and future type

ahuang11:
If we're releasing 1.3.7, would also like https://github.com/holoviz/panel/pull/6105 in.

PR #6120, opened 2023-12-22
Loosen callback future type

ahuang11:
Fixes https://github.com/holoviz/panel/issues/6118.

I print the type of `future` and it results in `<class 'asyncio.futures.Future'>`

I also do:
```
            print(isinstance(future, asyncio.Future), isinstance(future, asyncio.futures.Future))
```
which returns True, True

but I still get:
```
ValueError: ClassSelector parameter 'ChatInterface._callback_future' value must be an instance of Future, not <Future pending cb=[_chain_future.<locals>._call_check_cancel() at /Users/ahuang/miniconda3/envs/panel/lib/python3.10/asyncio/futures.py:385]>.
```

So I just set to param.Parameter and it works.

image

Also, I needed to specify the height manually in the app, or else no messages show up:
image

### DIFF
--- a/panel/chat/feed.py
+++ b/panel/chat/feed.py
@@ -205,9 +205,6 @@ class ChatFeed(ListPanel):
         The placeholder wrapped in a ChatMessage object;
         primarily to prevent recursion error in _update_placeholder.""")
 
-    _callback_future = param.Parameter(allow_None=True, doc="""
-        The current, cancellable async task being executed.""")
-
     _callback_state = param.ObjectSelector(objects=list(CallbackState), doc="""
         The current state of the callback.""")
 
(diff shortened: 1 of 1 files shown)


panel/chat/feed.py
```diff
@@ -205,7 +205,7 @@ class ChatFeed(ListPanel):
         The placeholder wrapped in a ChatMessage object;
         primarily to prevent recursion error in _update_placeholder.""")
 
-    _callback_future = param.ClassSelector(class_=asyncio.Future, allow_None=True, doc="""
```
philippjfr, line 208: I don't think it needs to be a parameter in the first place, just make it an attribute.

philippjfr (approved):

PR #6135, opened 2023-12-27
Tweak Chat Message CSS to center text

ahuang11:
Before this PR, integer objects were not centered:
```python
import panel as pn
pn.extension()

pn.Column(*[pn.chat.ChatMessage(i) for i in range(100)], height=500).show()
```

image

After:
image

Strings align properly with/without the change:
```python
import panel as pn
pn.extension()

pn.Column(*[pn.chat.ChatMessage(f"{i}") for i in range(100)], height=500).show()
```
image

### DIFF
--- a/panel/dist/css/chat_message.css
+++ b/panel/dist/css/chat_message.css
@@ -80,6 +80,9 @@
   max-width: calc(100% - 40px);
   padding: 5px;
   width: fit-content;
+  display: flex;
+  align-items: center;
+  justify-content: center;
 }
 
 .footer {


philippjfr (approved):
Makes total sense to me but can you confirm that this doesn't cause any issues with other content, markdown, images, etc.?

ahuang11:
Since `message` class only gets added to Markdown, I think it doesn't cause issues:

image

image

image

image

image

philippjfr (approved):

Issue #6038, opened 2023-12-14
Add ButtonIcon (or IconButton)?


ahuang11:
Now that ToggleIcon is available, https://github.com/holoviz/panel/pull/6034

PR #6138, opened 2023-12-28
Implement ButtonIcon and achieve parity with Button/Toggle widgets

ahuang11:
Closes https://github.com/holoviz/panel/issues/6038

Abstracts away ToggleIcon into ClickableIcon and makes both inherit from that.

Eventually used for `ChatCopyIcon`, but also many other uses, similar to mobile apps

```python
import panel as pn

pn.extension()

pn.widgets.ButtonIcon(icon="clipboard", active_icon="check", active_duration=300, size='5em').servable()
```

https://github.com/holoviz/panel/assets/15331990/e86214e7-8653-4b42-b609-364c8c8e2f4a

Needs tests + docs


panel/models/icon.ts
```diff
@@ -134,11 +186,76 @@ export class ToggleIcon extends Control {
...
+  update_cursor(): void {
+    this.icon_view.el.style.cursor = this.model.disabled ? 'default' : 'pointer';
+  }
+
+  get_icon(): string {
+    return this.value ? this.get_active_icon() : this.model.icon;
+  }
+
+  connect_signals(): void {
+    super.connect_signals();
```
ahuang11, line 230: Might be missing clicks here to trigger toggle.

panel/models/icon.ts
```diff
@@ -5,8 +5,8 @@ import type { IterViews } from '@bokehjs/core/build_views';
 import * as p from "@bokehjs/core/properties";
 import { build_view } from '@bokehjs/core/build_views';
 
-export class ToggleIconView extends ControlView {
-  model: ToggleIcon;
+export class ClickableIconView extends ControlView {
```
philippjfr, line 8: Please pull it out into separate modules.

panel/models/icon.ts
```diff
@@ -134,11 +186,76 @@ export class ToggleIcon extends Control {
...
+    if (this.model.disabled) {
+      return;
+    }
+    const updateState = (value: boolean, disabled: boolean) => {
+      this.value = value;
+      this.model.disabled = disabled;
+      this.update_icon();
+    };
+    updateState(true, true);
+    this.model.clicks += 1;
```
philippjfr, line 213: Would prefer to use a Bokeh event for this, not a property. Property changes can be silently dropped if there is a more recent property change but events will not.

panel/widgets/icon.py
```diff
@@ -42,3 +42,22 @@ def _update_icon(self):
...
+
+    _widget_type = _PnToggleIcon
+
+
+class ButtonIcon(ClickableIcon):
+
+    clicks = param.Integer(default=0, doc="""
+        The number of times the button has been clicked.""")
+
+    active_duration = param.Integer(default=75, doc="""
```
philippjfr, line 60: It also controls how long the button is disabled right?

panel/models/index.ts
```diff
@@ -2,6 +2,7 @@ export {AcePlot} from "./ace"
 export {Audio} from "./audio"
 export {BrowserInfo} from "./browser"
 export {Button} from "./button"
+export {ButtonIcon} from "./button_icon"
```
ahuang11, line 5: Do I need to export ClickableIcon if its an abstract class?

panel/widgets/icon.py
ahuang11: Yup, added mention. I'm also torn between naming it `disabled_duration` vs `active_duration` since `active_icon` and `disabled` both get affected.

ahuang11: Maybe just toggle_duration.

panel/models/icon.ts
ahuang11: I found something similar in FileDownload, but not 100% sure if this is what you mean.

```typescript
  render(): void {
    super.render();
    this._click_listener = this.increment_clicks.bind(this)
    this.icon_view.el.addEventListener("click", this._click_listener)
  }

  increment_clicks(): void {
    if (!this.model.disabled) {
      this.model.clicks = this.model.clicks + 1
    }
  }
```

panel/widgets/button.py
```diff
@@ -89,16 +89,12 @@ def _process_param_change(self, params):
         return super()._process_param_change(params)
 
 
-class _ClickButton(_ButtonBase, IconMixin):
+class _ClickButton(Widget):
```
ahuang11, line 92: I checked and I think only button.py uses _ClickButton.

panel/widgets/button.py
```diff
@@ -89,16 +89,12 @@ def _process_param_change(self, params):
...
 
     __abstract = True
 
     _event: ClassVar[str] = 'button_click'
 
-    _source_transforms: ClassVar[Mapping[str, str | None]] = {
```
ahuang11, line 98: This is no longer needed as it no longer inherits from _ButtonBase I think.

panel/widgets/button.py
```diff
@@ -156,7 +152,7 @@ def jscallback(self, args: Dict[str, Any] = {}, **callbacks: str) -> Callback:
         return Callback(self, code=callbacks, args=args)
 
 
-class Button(_ClickButton, TooltipMixin):
+class Button(_ButtonBase, _ClickButton, IconMixin, TooltipMixin):
```
ahuang11, line 155: _ClickButton used to inherit from `_ButtonBase, IconMixin`, but no longer, so we need to subclass these.

panel/models/icon.ts
ahuang11: Okay I looked at button.ts in bokeh, and I think that does what you're suggesting.

ahuang11:
Okay I think this is ready!

params demo
https://github.com/holoviz/panel/assets/15331990/d34bf3e2-b9b8-4719-9782-9ab773234cd4

js on click demo
https://github.com/holoviz/panel/assets/15331990/6cf13818-67e2-4a61-bc54-c13221a322c8

philippjfr (changes requested):
Looks great, just one more comment.

panel/models/button_icon.ts
```diff
@@ -0,0 +1,60 @@
...
+  update_cursor(): void {
+    this.icon_view.el.style.cursor = this.model.disabled ? 'default' : 'pointer';
+  }
+
+  click(): void {
+    if (this.model.disabled) {
+      return;
+    }
+    super.click();
+    this.model.clicks += 1;
```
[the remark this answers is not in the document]
ahuang11: Maybe I implemented it wrong, but if I remove `this.model.clicks += 1` it doesn't increment anymore.

https://github.com/holoviz/panel/assets/15331990/d7022f5a-6720-4c28-bc43-0aafdaa138b6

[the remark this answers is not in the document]
philippjfr: Definitely not implemented correctly then.

panel/widgets/icon.py
```diff
@@ -25,20 +31,87 @@ class ToggleIcon(Widget):
...
+            The function to run on click events. Must accept a positional `Event` argument
+
+        Returns
+        -------
+        watcher: param.Parameterized.Watcher
+          A `Watcher` that executes the callback when the MenuButton is clicked.
+        """
+        return self.param.watch(callback, 'clicks', onlychanged=False)
+
+    def _process_event(self, event: param.parameterized.Event) -> None:
```
philippjfr, line 113: This seems like your problem, you need to implement the logic:

```
        self.param.trigger('value')
        self.clicks += 1
```

philippjfr: Note that the type annotation here is also incorrect. If you could please fix this one and the one on `Button._process_event` both should use:

```python
from bokeh.events import ButtonClick
...
    def _process_event(self, event: ButtonClick) -> None:
```

ahuang11: Just wondering is there a benefit to implement clicks += 1 on the Python side vs TS?

philippjfr: Yes, as I said, property changes are throttled and potentially discarded if they are queued up, events are not.

philippjfr:
Once you've removed the clicks property and it's working this is looking really great, thanks!

ahuang11:
With the latest changes:

https://github.com/holoviz/panel/assets/15331990/41a82b71-e84d-4a84-b3ec-e1f7c40a8dda

philippjfr:
If you could remove the `test_button_icon_jslink` test I'll merge.

panel/tests/ui/widgets/test_icon.py
```diff
@@ -161,15 +186,167 @@ def test_toggle_icon_svg_to_tabler(page):
...
+    button.jslink(int_slider, clicks="value")
+    row = Row(button, int_slider)
+
+    serve_component(page, row)
+
+    page.click(".bk-TablerIcon")
+    wait_until(lambda: int_slider.value == 1, page)
+    assert int_slider.value == 1
+
+
```
ahuang11, line 332:
```suggestion
```

philippjfr (approved):

Issue #6143, opened 2024-01-02
Fix datetimepicker for most common Pangeo xarray use case


rsignell-usgs:
There are a bunch of [open issues for datetimepicker](https://github.com/holoviz/panel/issues?q=is%3Aissue+is%3Aopen+datetimepicker) and it seems like many of them could be summarized by making one of the most common Pangeo xarray use cases work -- simply using a datetimepicker instead of a slider or select when selecting n-dimensional array output.  

For example, make this simple xarray tutorial dataset work:
``` python
import hvplot.xarray  
import xarray as xr
import panel as pn
pn.extension()

ds = xr.tutorial.open_dataset('air_temperature')
viz = ds['air'].hvplot(x='lon', y='lat')
pn.panel(viz, widgets={'time': pn.widgets.DatetimePicker})    
```

amsnyder:
Our project is planning to make several panel apps for datasets with large time series, so this would be helpful for our team as well.

ahuang11:
Thanks for raising this! I made changes in #6152 to fix this and other various DatetimePicker functionality. If you have the time, I'd appreciate help testing that PR!
![image](https://github.com/holoviz/panel/assets/15331990/a18857b6-07df-4734-92b9-4405eea81b84)

rsignell:
@ahuang11 , that's awesome!   
I tried the above and it works, but how do you get the whole calendar to appear in a notebook cell?
![Screenshot 2024-01-03 164404](https://github.com/holoviz/panel/assets/125569335/e5da85a6-7ddf-4e94-82ae-339c3a9652c9)

ahuang11:
Not sure why the CSS is not updated in jupyter lab cells. Maybe when it's released, it'll be fixed? Or clearing of cache?

<img width="1162" alt="image" src="https://github.com/holoviz/panel/assets/15331990/39b1b20e-9e95-4cf7-9079-33f30d9312dc">

rsignell:
@ahuang11 I tried [following this procedure from @philippjfr](https://discourse.holoviz.org/t/blank-output-with-latest-holoviz-from-conda-forge/2276/11?u=rsignell), which usually clears up my strange holoviz display issues. 

Didn't work this time though.  When I run the cell and click on the widget, I can only just barely see the top of the calendar and can't figure out how to get the rest to display:
![Screenshot 2024-01-04 100943](https://github.com/holoviz/panel/assets/125569335/38c613ea-b449-40e3-9822-fc38b8591692)

[...]

PR #6152, opened 2024-01-03
Improve date and datetime picker functionality

ahuang11:
Some bug fixes (start/end were not applied if datetime64) and makes the following work:

Closes https://github.com/holoviz/panel/issues/6143

image

```python
import datetime as dt
import panel as pn
import numpy as np
import xarray as xr
import hvplot.xarray

pn.extension()

ds = xr.tutorial.open_dataset("air_temperature")
viz = ds["air"].hvplot(x="lon", y="lat")
pn.panel(viz, widgets={"time": pn.widgets.DatetimePicker}).show()
```

### DIFF
--- a/panel/dist/css/widgetbox.css
+++ b/panel/dist/css/widgetbox.css
@@ -4,6 +4,4 @@
   border-radius: 4px;
   -webkit-box-shadow: inset 0 1px 1px rgba(0, 0, 0, 0.05);
   box-shadow: inset 0 1px 1px rgba(0, 0, 0, 0.05);
-  overflow-x: hidden;
-  overflow-y: hidden;
 }
--- a/panel/models/datetime_picker.py
+++ b/panel/models/datetime_picker.py
@@ -10,7 +10,7 @@ class DatetimePicker(InputWidget):
 
     '''
 
-    value = Nullable(Either(String, Date, Datetime), help="""
+    value = Nullable(String, help="""
     The initial or picked date.
     """)
 
--- a/panel/models/datetime_picker.py
+++ b/panel/models/datetime_picker.py
@@ -10,7 +10,7 @@ class DatetimePicker(InputWidget):
 
     '''
 
-    value = Nullable(String, help="""
+    value = Nullable(Either(String, Date, Datetime), help="""
     The initial or picked date.
     """)
 
--- a/panel/util/__init__.py
+++ b/panel/util/__init__.py
@@ -469,6 +469,6 @@ def styler_update(styler, new_df):
 
 
 def try_datetime64_to_datetime(value):
-    if hasattr(value, "astype"):
+    if isinstance(value, np.datetime64):
         value = value.astype('datetime64[ms]').astype(datetime)
     return value
--- a/panel/util/__init__.py
+++ b/panel/util/__init__.py
@@ -466,3 +466,9 @@ def styler_update(styler, new_df):
         todo = tuple(ops)
         todos.append(todo)
     return todos
+
+
+def try_datetime64_to_datetime(value):
+    if hasattr(value, "astype"):
+        value = value.astype('datetime64[ms]').astype(datetime)
+    return value
--- a/panel/widgets/input.py
+++ b/panel/widgets/input.py
@@ -334,6 +334,21 @@ class DatePicker(Widget):
 
     _widget_type: ClassVar[Type[Model]] = _BkDatePicker
 
+    def __init__(self, **params):
+        if 'options' in params:
+            options = list(params.pop('options'))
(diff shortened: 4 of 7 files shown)


panel/dist/css/widgetbox.css
```diff
@@ -4,6 +4,4 @@
   border-radius: 4px;
   -webkit-box-shadow: inset 0 1px 1px rgba(0, 0, 0, 0.05);
   box-shadow: inset 0 1px 1px rgba(0, 0, 0, 0.05);
-  overflow-x: hidden;
-  overflow-y: hidden;
```
ahuang11, line 8: These were hiding the calendar display. I can't think of a scenario to crop the widget display, so I removed this.

panel/widgets/input.py
```diff
@@ -469,6 +469,9 @@ class _DatetimePickerBase(Widget):
     __abstract = True
 
     def __init__(self, **params):
+        if "options" in params:
+            options = list(params.pop("options"))
+            params["enabled_dates"] = options
```
ahuang11, line 474: I was initially thinking of putting this in the [holoviews.py module where the widget  conversion happens](https://github.com/holoviz/panel/blob/main/panel/pane/holoviews.py#L663), but it felt awkward to add DatetimePicker logic there, so I put it here. Also, since `options` is used for most other widgets, it makes sense to also support options for the DatetimePicker widget too.

panel/widgets/input.py
```diff
@@ -478,14 +481,18 @@ def _convert_to_datetime(v):
             return v
         elif isinstance(v, date):
             return datetime(v.year, v.month, v.day)
+        elif hasattr(v, "astype"):
+            return v.astype('datetime64[ms]').astype(datetime)
```
ahuang11, line 485: ChatGPT's explanation on why I need two astypes:
"""
When you use astype() on a numpy.datetime64 object to convert it to a Python datetime.datetime object, it might unexpectedly return an integer instead of a datetime object. This happens because astype(datetime) is actually interpreted as astype('O'), where 'O' stands for object, and it ends up converting the datetime64 to its underlying integer representation (the number of time units since a certain epoch).

To correctly convert a numpy.datetime64 object to a Python datetime.datetime object, you should explicitly import the datetime module and use datetime.datetime in the astype() method
"""

panel/widgets/input.py
```diff
@@ -469,23 +470,39 @@ class _DatetimePickerBase(Widget):
...
+            )
+
         if isinstance(v, datetime):
             return v
         elif isinstance(v, date):
             return datetime(v.year, v.month, v.day)
+        elif hasattr(v, "astype"):
+            return v.astype('datetime64[ms]').astype(datetime)
+        else:
+            raise ValueError(f"Could not convert {v} to datetime")
```
ahuang11, line 495: Just so nothing goes unnoticed.

panel/models/datetime_picker.py
```diff
@@ -10,7 +10,7 @@ class DatetimePicker(InputWidget):
 
     '''
 
-    value = Nullable(String, help="""
+    value = Nullable(Either(String, Date, Datetime), help="""
```
philippjfr, line 13: Is this really needed or correct? What does bokeh convert the Date, Datetime values to?

panel/widgets/input.py
```diff
@@ -469,23 +485,42 @@ class _DatetimePickerBase(Widget):
     __abstract = True
 
     def __init__(self, **params):
+        if 'options' in params:
```
philippjfr, line 488: Add a comment on why.

ahuang11:
Actually I think there's an issue with the implementation, and I'm not sure how to address:

image

Initially, the value is a np.datetime64

image

When a value is selected, it reverts back to `datetime.datetime`

image

Which I think is the root of a blank plot:

image

CC: @rsignell

ahuang11:
Would it be acceptable to have a as_numpy_datetime64 param?

https://github.com/holoviz/panel/assets/15331990/39589d2a-d888-43d6-b7c8-8e1a0d6b04e6

panel/util/__init__.py
```diff
@@ -466,3 +466,9 @@ def styler_update(styler, new_df):
         todo = tuple(ops)
         todos.append(todo)
     return todos
+
+
+def try_datetime64_to_datetime(value):
+    if hasattr(value, "astype"):
```
philippjfr, line 472: I guess my main question is why you can't explicitly check for a datetime64 value? This seems unsafe.

ahuang11: Oh yeah; I forget we import numpy.

philippjfr (approved):

PR #6166, opened 2024-01-06
handle title_formatter if is type dict

jz314:
handle `title_formatter` in the same way as `formatter` immediately above; otherwise if `title_formatter` is dict, it is still treated as string


philippjfr (approved):
Good catch, thank you!

Issue #6162, opened 2024-01-05
Perspective does not show data if datetime column contains NaN


MarcSkovMadsen:
I'm on Panel==1.3.6 trying to use Perspective.

```python
import pandas as pd
import pandas as pd
import numpy as np

# Create a sample DataFrame with a datetime column
data = {'Datetime': pd.to_datetime(['2022-01-01', '2022-01-02', '2022-01-03', '2022-01-04'])}
df = pd.DataFrame(data)

# Assign NaN values to specific rows or cells
df.loc[1, 'Datetime'] = np.nan
df.loc[3, 'Datetime'] = np.nan
print(df)
print(df.dtypes)

import panel as pn
pn.extension("perspective")

pn.pane.Perspective(df, sizing_mode="stretch_width").servable()
```

![image](https://github.com/holoviz/panel/assets/42288570/123754d0-0712-494c-88c3-21819f538f81)

```bash
format_cell.js:71  Uncaught (in promise) RangeError: Invalid time value
```

PR #6176, opened 2024-01-09
Handle NaT values on Perspective pane

philippjfr:
Fixes https://github.com/holoviz/panel/issues/6162

### DIFF
--- a/panel/models/perspective.ts
+++ b/panel/models/perspective.ts
@@ -219,8 +219,14 @@ export class PerspectiveView extends HTMLBoxView {
 
   get data(): any {
     const data: any = {}
-    for (const column of this.model.source.columns())
-      data[column] = this.model.source.get_array(column)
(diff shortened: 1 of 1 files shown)


panel/models/perspective.ts
```diff
@@ -219,8 +219,14 @@ export class PerspectiveView extends HTMLBoxView {
...
-    for (const column of this.model.source.columns())
-      data[column] = this.model.source.get_array(column)
+    for (const column of this.model.source.columns()) {
+      let array = this.model.source.get_array(column)
+      if (this.model.schema[column] == 'datetime' && array.includes(-9223372036854776)) {
+	array = array.map((v) => v === -9223372036854776 ? null : v)
+      }
+      data[column] = array
+    }
+    console.log(data)
```
philippjfr, line 229:
```suggestion
```

PR #6181, opened 2024-01-11
Add Textual pane

philippjfr:
textual_app

- [x] Add tests
- [x] Disconnect the app from the event loop when all views are deleted


MarcSkovMadsen:
Does this "just work" in a multiuser setting. Anything users need to be aware of?

PR #6187, opened 2024-01-11
Set ChatMessage min_width_0 and break anywhere

ahuang11:
Simon mentioned this: https://www.youtube.com/watch?v=cH8VbLM1958 
https://www.joshwcomeau.com/css/interactive-guide-to-flexbox/#the-minimum-size-gotcha-11

Awkward overflowing before:
image

Now:
image


philippjfr (approved):

MarcSkovMadsen:
Have you tested the different templates? In my experience css changes can behave different in each of these.

PR #6188, opened 2024-01-12
correct duplicate wording

fayssalelmofatiche:
the word "construct" was duplicated. NeoVim decided to do a bit of MarkDown reformatting upon as well.


philippjfr:
Thanks. That said I don't quite understand the reformatting, `_` is not preferable over `*`, they're equivalent. It's also inconsistent with the rest of the codebase, so on balance I'd prefer not to apply those changes.

fayssalelmofatiche:
I thought I should revert it as well, but it is indeed equivalent. Let me take it out to stay consistent with the rest as you mentioned.

philippjfr (approved):
Really appreciate it, thanks!

PR #6198, opened 2024-01-14
Improve the user experience for resetting Jupyterlite.

MarcSkovMadsen:
I just discovered by coincidence that the Getting_Started notebook in Panelite now contains functionality to reset. This is great.

I think the functionality can improved

- Make it more easily discoverable by putting in separate, clearly named notebook.
- Avoid having to import, run and scroll down. This is slow. This can be avoided by putting it in separate notebook. (Yes. I know you don't have to import and run. But its just a very natural thing to do).
- Send a clear signal that the storage has been cleared. Today when you click the button nothing visible happens. Thus you don't feel confident that the script finished with success.

Furthermore as a developer

- I cannot find the `Getting_Start.ipynb` notebook via my IDE because its actually ignored in the gitignore. So its simply excluded by VS code.
- I do not get the notebook cleaned when I use pre-commit because its actually skipped in the configuration.

This PR fixes these issues.

image

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -116,7 +116,7 @@ jupyter_execute
 # this dir contains the whole website and should not be checked in on main
 builtdocs/
 .doit.db
-
+lite/files/*/
 .jupyterlite.doit.db
 
 # jetbrains settings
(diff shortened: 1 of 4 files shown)


philippjfr (approved):
Good idea, thanks!

PR #6245, opened 2024-01-19
Fix ChatFeed / Interface tests and async generator placeholders

ahuang11:
Ensures that the tests actually run and raises errors by setting `callback_exception = "raise"` rather than having the errors piped to the chat logs

This resulted in finding a lot of tests failing, **but** most of them were just tests not written properly (e.g. not using `wait_until` to assert)

The only bug I found was a result of https://github.com/holoviz/panel/pull/6105, where async generator functions execution time was based on `placeholder_threshold`, i.e. `await _schedule_placeholder` had to run in completion, before the *actual callback* started executing.

This was remedied by wrapping `_serialize_response` (where the iteration of the generator was happening) alongside the actual callback so that when checking `task.done()`, it checks the execution of both of these.

### DIFF
--- a/panel/chat/feed.py
+++ b/panel/chat/feed.py
@@ -10,7 +10,10 @@ import traceback
 
 from enum import Enum
 from functools import partial
-from inspect import isasyncgen, isawaitable, isgenerator
+from inspect import (
+    isasyncgen, isasyncgenfunction, isawaitable, iscoroutinefunction,
+    isgenerator,
+)
 from io import BytesIO
 from typing import (
     TYPE_CHECKING, Any, Callable, ClassVar, Dict, List, Literal,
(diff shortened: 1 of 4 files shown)


ahuang11:
I manually tested this branch on a few examples:

https://github.com/holoviz/panel/assets/15331990/2fc554bf-20f6-46db-9b92-c2fbd1d6cf8d

https://github.com/holoviz/panel/assets/15331990/2bf28f88-358d-4e70-9e23-03ddcc625c03

https://github.com/holoviz/panel/assets/15331990/2ac9c58a-5131-44f0-9455-539cdb7970a0

panel/chat/feed.py
```diff
@@ -470,13 +473,25 @@ async def _schedule_placeholder(
             return
 
         start = asyncio.get_event_loop().time()
-        while not self._callback_state == CallbackState.IDLE and num_entries == len(self._chat_log):
+        while not task.done() and num_entries == len(self._chat_log):
             duration = asyncio.get_event_loop().time() - start
             if duration > self.placeholder_threshold:
                 self.append(self._placeholder)
                 return
             await asyncio.sleep(0.28)
```
philippjfr, line 481: Any specific reason for this timeout?

```suggestion
            await asyncio.sleep(0.28)
```

ahuang11: So it doesn't completely take over the event loop.

philippjfr: Sure, but why 0.28s? Seems oddly specific and also quite long.

ahuang11: It's one of my favorite numbers xD

No objections to changing it to like 0.1.

ahuang11, line 481:
```suggestion
            await asyncio.sleep(0.1)
```

philippjfr: Haha okay, yes something like 100ms seems more reasonable. The loop body should only take a few hundred milliseconds to execute.

philippjfr (approved):

PR #6253, opened 2024-01-22
fix typo

fohrloop:
Fixes a small typo in the docs, found on this page: https://panel.holoviz.org/explanation/dependencies/param.html#runtime-type-checking (`x`  should be `t`).


maximlt (approved):
Good catch! Thanks :)

PR #6259, opened 2024-01-22
Update to work with pandas 2.2

Hoxbro:


panel/tests/widgets/test_tables.py
```diff
@@ -27,6 +24,16 @@
                             reason="Requires latest pandas")
 
 
+def makeMixedDataFrame():
```
Hoxbro, line 27: I used a function here because it was easier to drop in than having to use pytest fixtures.

philippjfr (approved):

Issue #6246, opened 2024-01-20
Not clear from Reference Documentation how TextAreaInput works


MarcSkovMadsen:
I wanted to try out and learn about the new `TextAreaInput`, but I could not figure it out because.

## No Default Value Parameter List

![image](https://github.com/holoviz/panel/assets/42288570/30718834-7752-4c73-ba8c-aae39c233ef4)

## Controls not working

No matter which value I select for `resizable` in the `Controls I cannot change the width of the Text Area Input with the mouse.

![image](https://github.com/holoviz/panel/assets/42288570/4c17761a-adfc-4c67-a010-484f85060447)

philippjfr:
Ping @ahuang11

PR #6264, opened 2024-01-23
Improve documentation on TextAreaInput

ahuang11:
Closes https://github.com/holoviz/panel/issues/6246#issuecomment-1906209924

Added mention of default, an example, and ~enable readonly=True.~ Okay readonly doesn't initialize.

Unsure how to remove / disable resizable from the jslink side since it can't be set after it has been initialized.

I can't even find resizable on Bokeh's TextAreaInput:
https://github.com/bokeh/bokeh/blob/4b61612d14f840b990a74e455d49feb012badc12/bokehjs/src/lib/models/widgets/textarea_input.ts#L41

Only https://github.com/bokeh/bokeh/blob/4b61612d14f840b990a74e455d49feb012badc12/src/bokeh/models/layouts.py#L287-L289

But it works

https://github.com/holoviz/panel/assets/15331990/ec4780b9-6d9d-486d-98b3-35721d77fc67

### DIFF
--- a/examples/reference/widgets/TextAreaInput.ipynb
+++ b/examples/reference/widgets/TextAreaInput.ipynb
@@ -39,7 +39,7 @@
     "* **`name`** (str): The title of the widget\n",
     "* **`placeholder`** (str): A placeholder string displayed when no value is entered\n",
     "* **`rows`** (int, default=2): The number of rows in the text input field. \n",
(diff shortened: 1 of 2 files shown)


MarcSkovMadsen (approved):
Perfect. Thanks

maximlt (approved):

philippjfr (approved):

PR #6266, opened 2024-01-23
Add scroll options to permanently toggle on scrollbar

ahuang11:
Adds option to permanently turn on scroll in direction(s), like for use in `pn.Feed` to prevent the layout from bouncing around if objects are added / removed.

```python
import panel as pn

col = pn.Column(
    pn.Spacer(styles=dict(background='red'), width=200, height=200),
    pn.Spacer(styles=dict(background='green'), width=200, height=200),
    pn.Spacer(styles=dict(background='blue'), width=200, height=200),
    scroll="y", height=1000
)
col.servable()
```
image


panel/layout/base.py
```diff
@@ -785,9 +794,18 @@ class ListPanel(ListLike, Panel):
     An abstract baseclass for Panel objects with list-like children.
     """
 
-    scroll = param.Boolean(default=False, doc="""
-        Whether to add scrollbars if the content overflows the size
-        of the container.""")
+    scroll = param.ObjectSelector(
```
Hoxbro, line 797: I think `param.Selector` should be used here as `ObjectSelector` is soft-deprecated.

panel/layout/base.py
```diff
@@ -817,15 +835,16 @@ def _linked_properties(self):
         )
 
     def _process_param_change(self, params: Dict[str, Any]) -> Dict[str, Any]:
-        if 'scroll' in params:
+        if params.get('scroll'):
```
Hoxbro, line 838: With the change, you no longer run the if statement if `scroll` is False. Is that deliberate?

If it is, you could use a walrus operator and combine it with the next line.

ahuang11: > With the change, you no longer run the if statement if scroll is False. Is that deliberate?

Yes, setting `scroll=False` didn't do anything I think.

philippjfr (approved):

Issue #6265, opened 2024-01-23
Panel._compute_sizing_mode not gracefully handling None


banesullivan-kobold:
#### ALL software version info
(this library, plus any other relevant software, e.g. bokeh, python, notebook, OS, browser, etc)

```
--------------------------------------------------------------------------------
  Date: Tue Jan 23 17:55:49 2024 GMT

                OS : Linux
            CPU(s) : 12
           Machine : aarch64
      Architecture : 64bit
               RAM : 23.4 GiB
       Environment : Python

  Python 3.10.13 (main, Oct 12 2023, 02:03:44) [GCC 12.2.0]

             panel : 1.3.7
             bokeh : 3.3.1
             numpy : 1.23.5
             scipy : 1.10.1
           IPython : 8.12.0
        matplotlib : 3.7.1
            scooby : 0.8.0
--------------------------------------------------------------------------------
```

#### Description of expected behavior and the observed behavior

`Panel._compute_sizing_mode` is not gracefully handling if `margin` is set as `None` in the passed `props`.

#### Complete, minimal, self-contained example code that reproduces the issue

I am struggling to make a MRE for this as this quirk/bug surfaced after some seemingly innocuous changes to other parts of a larger Panel application I'm working on. I cannot seem to track down exactly why the margin is being set to `None` with the `pn.Row` that is causing this. 

I'm hoping that adding a check to make sure the margin is not `None` in this function is easy enough without tracking down the root cause though I also would love to get to the bottom of why the `margin` in the passed `props` to this function would ever be `None`.

Confirmed that `self.margin` has a value of `0`, but that the passed `props` has `margin` of `None` which is preferred here:

https://github.com/holoviz/panel/blob/075b5ae81265bf0710e273d42c8c44a894faeead/panel/layout/base.py#L212

#### Stack traceback and/or browser JavaScript console output

```
  File "/python/lib/python3.10/site-packages/panel/layout/base.py", line 511, in pop
    self.objects = new_objects
  File "/python/lib/python3.10/site-packages/param/parameterized.py", line 525, in _f
    instance_param.__set__(obj, val)
  File "/python/lib/python3.10/site-packages/param/parameterized.py", line 527, in _f
    return f(self, obj, val)
  File "/python/lib/python3.10/site-packages/param/parameterized.py", line 1545, in __set__
    obj.param._call_watcher(watcher, event)
  File "/python/lib/python3.10/site-packages/param/parameterized.py", line 2486, in _call_watcher
[...]

PR #6267, opened 2024-01-24
Handle margin=None in layout sizing mode computation

philippjfr:
Fixes https://github.com/holoviz/panel/issues/6265


banesullivan-kobold (approved):
Fantastic! Thank you for addressing this!

PR #6280, opened 2024-01-29
Del frame after stack level is found

Hoxbro:
It can lead to memory leaks if not deleted.

### DIFF
--- a/panel/util/warnings.py
+++ b/panel/util/warnings.py
@@ -40,13 +40,17 @@ def find_stack_level() -> int:
 
     frame = inspect.currentframe()
     stacklevel = 0
-    while frame:
-        fname = inspect.getfile(frame)
-        if fname.startswith((pkg_dir, param_dir)) and not fname.startswith(test_dir):
-            frame = frame.f_back
-            stacklevel += 1
-        else:
-            break
+    try:
+        while frame:
+            fname = inspect.getfile(frame)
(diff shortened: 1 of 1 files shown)


panel/util/warnings.py
```diff
@@ -40,13 +40,17 @@ def find_stack_level() -> int:
...
+        while frame:
+            fname = inspect.getfile(frame)
+            if fname.startswith((pkg_dir, param_dir)) and not fname.startswith(test_dir):
+                frame = frame.f_back
+                stacklevel += 1
+            else:
+                break
+    finally:
+        # See: https://docs.python.org/3/library/inspect.html#inspect.Traceback
+        del frame
```
philippjfr, line 53: Not blocking this but I'm a little confused, I would have thought at the function boundary it would have called `__del__` anyway.

Hoxbro: I'm just following what Pandas does now.

maximlt: [Polars applied the same fix](https://github.com/pola-rs/polars/pull/10630/files) and included a part of Python's docs as a comment:

```
    finally:
        # https://docs.python.org/3/library/inspect.html
        # > Though the cycle detector will catch these, destruction of the frames
        # > (and local variables) can be made deterministic by removing the cycle
        # > in a finally clause.
        del frame
```

philippjfr (approved):

PR #6283, opened 2024-01-30
Add downloads badge to table

ahuang11:
I think we should advertise it since it's pretty impressive!

image

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,8 +19,7 @@ Enjoying Panel? Show your support with a [Github star](https://github.com/holovi
 <tbody>
 <tr>
 <td>Downloads</td>
-<td><a href="https://pypistats.org/packages/panel" alt="PyPi Downloads"><img src="https://img.shields.io/pypi/dm/panel" /></a></td>
-<td><a href="https://anaconda.org/pyviz/panel" alt="Conda Downloads"><img src="https://pyviz.org/_static/cache/panel_conda_downloads_badge.svg" /></a></td>
+<td><a href="https://pypistats.org/packages/panel" alt="PyPi Downloads"><img src="https://img.shields.io/pypi/dm/panel" /></a><a href="https://anaconda.org/pyviz/panel" alt="Conda Downloads"><img src="https://pyviz.org/_static/cache/panel_conda_downloads_badge.svg" /></a></td>
 </tr>
 <tr>
 <td>Build Status</td>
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -18,6 +18,10 @@ Enjoying Panel? Show your support with a [Github star](https://github.com/holovi
 <table>
 <tbody>
 <tr>
+<td>Downloads</td>
```
MarcSkovMadsen, line 21: Can we add the conda downloads of 170k too?

MarcSkovMadsen (approved):

README.md
maximlt: Yep why not. The conda badge on pyviz.org is actually manually built when the site is built (see this script https://github.com/pyviz/pyviz.org/blob/master/tools/conda_downloads.py). It's available at this URL for Panel: https://pyviz.org/_static/cache/panel_conda_downloads_badge.svg.

ahuang11: I'm leaning towards keeping them separate, mainly because I don't know how and I don't want this small change to take up too much time

Also, I think people can do some quick mental math to combine those numbers
image

philippjfr: +1, certainly as a first cut.

Hoxbro: Just a quick look at the screenshot. It seems to lack a space between the badges, and maybe change downloads to PyPi downloads or something like that.

ahuang11: The badge label is default to Downloads; not sure how to change it.

ahuang11: image

README.md
```diff
@@ -18,6 +18,10 @@ Enjoying Panel? Show your support with a [Github star](https://github.com/holovi
...
+<td><a href="https://pypistats.org/packages/panel" alt="PyPi Downloads"><img src="https://img.shields.io/pypi/dm/panel?label=pypi" /></a><a href="https://anaconda.org/pyviz/panel" alt="Conda Downloads"><img src="https://pyviz.org/_static/cache/panel_conda_downloads_badge.svg" /></a></td>
```
Hoxbro, line 22:
```suggestion
<td><a href="https://pypistats.org/packages/panel" alt="PyPi Downloads"><img src="https://img.shields.io/pypi/dm/panel?label=pypi" /></a> <a href="https://anaconda.org/pyviz/panel" alt="Conda Downloads"><img src="https://pyviz.org/_static/cache/panel_conda_downloads_badge.svg" /></a></td>
```
I think this adds a space between the badges.

PR #6285, opened 2024-01-30
fix docs typo: pn.state.onload

fohrloop:
was pn.state.on_load

Docs on this page: https://panel.holoviz.org/how_to/callbacks/load.html


maximlt (approved):
Thanks!

PR #6289, opened 2024-01-31
Add explanation docs about APIs

philippjfr:
In this PR I plan to add two different explanation sections:

- Reactivity vs Callbacks ([Preview](https://github.com/holoviz/panel/blob/reactive_explanation/doc/explanation/api/reactivity.md))
- Functions vs Classes


doc/explanation/api/reactivity.md
```diff
@@ -0,0 +1,177 @@
+# Reactivity in Panel
+
+Panel's approach to expressing interactivity and managing state differs from many other frameworks and offers multiple APIs with different affordances. In this overview section we will try to convey the thinking that led to the different APIs in Panel, providing you with context on when to use which and why. In the process we will try to unpack a number of concepts that are central to most UI frameworks and how these concepts translate into Panel's unique approach.
+
+:::{note}
+Panel has gone through a significant evolution over time and many of the concepts discussed here really only fully crystallized when Param 2.0 was released.
```
MarcSkovMadsen, line 6:
```suggestion
Panel has gone through a significant evolution over time and many of the concepts discussed here really only fully crystallized when Param 2.0 was released late 2023.
```

doc/explanation/api/reactivity.md
```diff
@@ -0,0 +1,177 @@
...
+:::
+
+## Concepts
+
+The core concept a large part of the discussion will focus on is the difference between reactive and event driven approaches for writing UIs. Panel has placed strong emphasis on innovating to provide novel **reactive** approaches to writing applications. Popular reactive frameworks include spreadsheets like Excel and frontend UI frameworks like React, each of which has a different approach to reactivity, but two core principles are conserved:
+
+1. They are **data-driven**, i.e. the UI updates automatically in response to changes in the data model. In Excel the data model is represented by value of a cell, while React has use state hooks. In Panel the data model is defined by Parameters, i.e. components and user classes define some parameters which drive changes in the UI. In all these frameworks this is achieved by what we will call **data binding**, how precisely this works will be the primary focus of this section.
```
MarcSkovMadsen, line 13: Is the data model still defined by parameters only after the release of `param.rx`?

doc/explanation/api/reactivity.md
```diff
@@ -0,0 +1,177 @@
...
+
+2. They are **declarative**, i.e. the user, defines what should be rendered and the reactive framework figures out how to efficiently make the required updates. In Excel this is simple, it simply has to re-evaluate any formulas that depend on the changed inputs and then update the cell's output. In React this can be significantly more involved, requiring diffing of the document model and then figuring out the most efficient updates to reflect the latest changes. In Panel this works either by defining a function who's output is diffed or by binding references to a declarative component (if you don't know what that means yet, don't worry, that's what this guide is for).
```
MarcSkovMadsen, line 15: Its true that *behind the scenes* Reacts updates can be significantly more involved. But for the developer the idea is that everything should be simple because they just need to relate to a linear flow from top to bottom.

doc/explanation/api/reactivity.md
```diff
@@ -0,0 +1,177 @@
...
+
+![Diagram of callback based UI code](../../_static/images/callbacks_diagram.png)
+
+The problem we immediately run into if some action we want to take requires more than one of the inputs is how we coordinate these changes. This is why many older callback based frameworks rely on what is called the Model-View-Controller (MVC) pattern, where the **view** that the user interacts with, gathers the inputs, which are fed to a **controller** that is able to coordinate the various events and manipulate the underlying model. This requires significant state handling, which in practice is only really feasible by implement a class based controller.
+
+In the reactive approach things are much simpler, the inputs are bound to a declarative function or component that immediately reflects the required changes in the rendered view:
+
+![Diagram of reactive UI code](../../_static/images/reactive_diagram.png)
```
MarcSkovMadsen, line 25: I like the figures. They explain well that the *reactive*/ *declarative* approach is simpler to reason about.

I would add big fat figure titles to make it even clearer which approach the figure refers to.

doc/explanation/api/reactivity.md
```diff
@@ -0,0 +1,177 @@
...
+
+The alternative to the reactive approach is an event-driven approach, involving callbacks that are triggered by user interactions or other events. This approach usually relies on **imperative** rather than **declarative** programming, i.e. you define step-by-step what should happen when a particular action is triggered. This is usually significantly more cumbersome, especially as an application grows in size and complexity. Even in a relatively simple application where we gather user inputs from multiple widgets and then perform some update we quickly run into problems. Let us take a simple example of an application that has three widgets that gather user input. In a callback based approach we have to register a callback with each of the three widgets and then perform some action as a result:
```
MarcSkovMadsen, line 17:
```suggestion
The alternative to the reactive approach is an **event-driven** approach, involving callbacks that are triggered by user interactions or other events. This approach usually relies on **imperative** rather than **declarative** programming, i.e. you define step-by-step what should happen when a particular action is triggered. This is usually significantly more cumbersome, especially as an application grows in size and complexity. Even in a relatively simple application where we gather user inputs from multiple widgets and then perform some update we quickly run into problems. Let us take a simple example of an application that has three widgets that gather user input. In a callback based approach we have to register a callback with each of the three widgets and then perform some action as a result:
```

doc/explanation/api/reactivity.md
```diff
@@ -0,0 +1,177 @@
+# Reactivity in Panel
+
+Panel's approach to expressing interactivity and managing state differs from many other frameworks and offers multiple APIs with different affordances. In this overview section we will try to convey the thinking that led to the different APIs in Panel, providing you with context on when to use which and why. In the process we will try to unpack a number of concepts that are central to most UI frameworks and how these concepts translate into Panel's unique approach.
+
+:::{note}
+Panel has gone through a significant evolution over time and many of the concepts discussed here really only fully crystallized when Param 2.0 was released.
+:::
+
+## Concepts
```
MarcSkovMadsen, line 9:
```suggestion
## Concepts: Reactive vs Event-Driven
```

doc/explanation/api/reactivity.md
```diff
@@ -0,0 +1,177 @@
...
+```
+
+In this example, we have bound two inputs to the function and return an output, the `Str` pane. Internally a few things will happen to make this work:
+
+1. Panel wraps the bound function in a so called `ParamFunction` pane.
+2. The `ParamFunction` pane watches the inputs to the bound function for changes.
+3. When a slider is dragged and the input changes the `ParamFunction` will re-evaluate the function with the new inputs and then figure out the appropriate update.
+
+:::{important}
+Since Panel cannot know if you have a handle on the `Str` pane that is being returned, it cannot safely update the `Str` pane inplace. Therefore it re-renders the corresponding model every time the inputs change. For complex panes and output this can lead to undesirable flicker. To avoid this you have to tell Panel that the output can be safely updated `inplace`:
```
MarcSkovMadsen, line 60: I don't understand "Since Panel cannot know if you have a handle on the `Str` pane that is being returned, it cannot safely update the `Str` pane inplace.".

MarcSkovMadsen, line 60: What do you mean by updating `inplace`? What does this more precisely do?

doc/explanation/api/reactivity.md
```diff
@@ -0,0 +1,177 @@
...
+
+The alternative to the reactive approach is an event-driven approach, involving callbacks that are triggered by user interactions or other events. This approach usually relies on **imperative** rather than **declarative** programming, i.e. you define step-by-step what should happen when a particular action is triggered. This is usually significantly more cumbersome, especially as an application grows in size and complexity. Even in a relatively simple application where we gather user inputs from multiple widgets and then perform some update we quickly run into problems. Let us take a simple example of an application that has three widgets that gather user input. In a callback based approach we have to register a callback with each of the three widgets and then perform some action as a result:
+
+![Diagram of callback based UI code](../../_static/images/callbacks_diagram.png)
```
MarcSkovMadsen, line 19: Would it be possible to have two tabs. On the first tab this image. On the second tab e2e code that can run illustrating the picture?

doc/explanation/api/reactivity.md
MarcSkovMadsen: Would it be possible to have two tabs. On the first tab this image. On the second tab e2e code that can run illustrating the picture?

doc/explanation/api/reactivity.md
```diff
@@ -0,0 +1,177 @@
...
+pn.pane.Markdown(object=text.param.value)
+```
+
+You might have wondered: "why is this useful?". You rarely want to bind the output of a widget directly to some other output or parameter. In practice an output or parameter value will usually be the result of transforming one or more inputs. For this reason Panel (and Param, which it builds on) allows a number of different types to be used as a reference:
+
+1. Parameters: A `Parameter` object is the simplest reference.
+2. Widgets: A Panel `Widget` acts as a proxy for its `value` parameter, i.e. passing `text` and `text.param.value` are equivalent in the last example.
+3. Bound functions: A function which has other references bound to it using the `pn.bind` helper may also be used as a reference.
+4. Reactive Expressions: Param offers so called reactive expressions which act as proxies for their underlying objects and can be chained into arbitrary expressions.
+5. Async Generators: Asynchronous generators functions can be used as a reference to drive streaming outputs.
```
MarcSkovMadsen, line 87: Would it be possible to add 5 tabs with e2e runnable code. Each tab illustrating one of the 5 different reference types?

doc/explanation/api/reactivity.md
```diff
@@ -0,0 +1,177 @@
...
+
+So far we have expounded at length about the benefits of reactive approaches, so you might wonder: "why allow callbacks at all?" One pragmatic reason for having callbacks is that reactivity is built on top of the callback based `.param.watch` API in Param, its the lowest level API offered and lets users build higher-level APIs on top. However, from a practical perspective there's a few reasons one might use callbacks even in a scenario where reactivity drives the majority of your application.
+
+### Performance
+
+One major reason for using callbacks is again related to the fact that they are the lowest level available API. This means they have the least amount of overhead and provide the ability to perform very fine-grained updates, i.e. update the specific parameter value that should be updated.
+
+### Side Effects
+
+Another common use case for callbacks are side-effects, e.g. you may want to trigger some external action or log user interactions. In this case callbacks are the main
```
MarcSkovMadsen, line 123: What do you mean by "are the main"? "main" what?

doc/explanation/api/reactivity.md
```diff
@@ -0,0 +1,177 @@
...
+
+### Transient Events
+
+Certain user interactions are, by their very nature more amenable to an event-driven approach because they are transient. One very common such example are button clicks. Since they are transient, it's not immediately obvious how to structure code to handle the event.
```
MarcSkovMadsen, line 127: Could you define what you mean by *transient*?

doc/explanation/api/reactivity.md
```diff
@@ -0,0 +1,177 @@
...
+a_slider = pn.widgets.FloatSlider()
+b_slider = pn.widgets.FloatSlider()
+button = pn.widgets.Button(name='Calculate')
+
+def add(a, b, compute):
+    if not compute:
+	    raise pn.param.Skip()
+    return f'{a} + {b} = {a+b}'
+
+pn.Column(a_slider, b_slider, button, pn.bind(add, a_slider, b_slider, button))
```
MarcSkovMadsen, line 158: You invented a new way to implement a *form*. I.e. you added `Skip` exception? I think its nice.

Is `Skip` something that should go into Param?

doc/explanation/api/reactivity.md
```diff
@@ -0,0 +1,177 @@
+# Reactivity in Panel
+
+Panel's approach to expressing interactivity and managing state differs from many other frameworks and offers multiple APIs with different affordances. In this overview section we will try to convey the thinking that led to the different APIs in Panel, providing you with context on when to use which and why. In the process we will try to unpack a number of concepts that are central to most UI frameworks and how these concepts translate into Panel's unique approach.
```
MarcSkovMadsen, line 3: I think I learn some things from this guide. But I don't think I have really learned anything about when to which and why. Make I can deduce it from what is written. But its not clear to me. Can it be summarized in a decission tree or table?

philippjfr: Think I'll write the second guide on functions vs classes and then figure out how to give that guidance.

doc/explanation/api/reactivity.md
philippjfr: I've honestly been meaning to add that for a very, very long time (several years) but always forgot. Yes it probably has to move up to param so this can be supported for bound function references.

doc/explanation/api/reactivity.md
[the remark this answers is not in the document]
jbednar: (Or something like that!)

jbednar (approved):
Looks good. For expedience I pushed some direct edits to the PR, so please review that commit and make sure I didn't break anything. :-)  I think this guide helps, but I still think we'll need more info, e.g. "what are _all_ the cases where I can provide a parameter object and it will be treated as a reactive object"? Clearly some Panel objects accept them for Parameter values; is it _all_ Panel objects, for _all_ Parameters?

doc/explanation/api/reactivity.md
philippjfr: In the background `param.rx` defines a parameter so technically yes, but I suppose to a user it wouldn't appear that way.

doc/explanation/api/index.md
```diff
@@ -12,6 +12,20 @@ Let's work through each API with an example app, while pointing out the benefits
 ::::{grid} 1 2 2 3
 :gutter: 1 1 1 2
 
+:::{grid-item-card} {octicon}`infinity;2.5em;sd-mr-1 sd-animate-grow50` Reactivity in Panel
+:link: reactivity
+:link-type: doc
+
+A discussion that contrasts reactive and callback based APIs in Panel.
```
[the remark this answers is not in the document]
Coderambling: Might misunderstand the above context, but I think I ran into the above issue when using layouts. To my surprise it was quite easy (for Philip at least :-) ) to update this in the Panel code for Flexbox. See below.

A number of ListLike layouts do not support objects =  at the moment. The PR below would fix that and make behaviour more uniform.

I Made a PR for Flexbox, Philip, was able to fix it with +- 5 lines of changed code. At his request I made this PR for the other layouts: https://github.com/holoviz/panel/issues/6389

doc/explanation/api/reactivity.md
```diff
@@ -0,0 +1,309 @@
...
+
+## Concepts: Reactive vs. Event-Driven
+
```
Coderambling, line 6: Add ### Reactive : subheadding below the heading, then the text for Reactive

PR #6290, opened 2024-01-31
Don't pop ChatInterface variables

Hoxbro:
Without these changes, ChatInterface will not use the new username.

``` python
import panel as pn
pn.extension()

pn.chat.ChatInterface(callback=lambda _0, user, _2: user, user="NEW USERNAME")
```

image

### DIFF
--- a/panel/chat/feed.py
+++ b/panel/chat/feed.py
@@ -231,8 +231,8 @@ class ChatFeed(ListPanel):
 
         # forward message params to ChatMessage for convenience
         message_params = params.get("message_params", {})
-        for param_key in list(params.keys()):
-            if param_key not in ChatFeed.param and param_key in ChatMessage.param:
+        for param_key in params.copy():
+            if param_key not in self.param and param_key in ChatMessage.param:
                 message_params[param_key] = params.pop(param_key)
         params["message_params"] = message_params
 
--- a/panel/tests/chat/test_interface.py
+++ b/panel/tests/chat/test_interface.py
@@ -358,6 +358,10 @@ class TestChatInterface:
         assert chat_interface.objects[1].object == "2"
         assert chat_interface.objects[2].object == "3"
 
+    def test_manual_user(self):
+        chat_interface = ChatInterface(user="New User")
+        assert chat_interface.user == "New User"
+
 class TestChatInterfaceWidgetsSizingMode:
     def test_none(self):
         chat_interface = ChatInterface()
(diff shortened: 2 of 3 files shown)


panel/chat/feed.py
```diff
@@ -231,8 +231,8 @@ def __init__(self, *objects, **params):
 
         # forward message params to ChatMessage for convenience
         message_params = params.get("message_params", {})
-        for param_key in list(params.keys()):
-            if param_key not in ChatFeed.param and param_key in ChatMessage.param:
+        for param_key in params.copy():
```
philippjfr, line 234: I would usually just do:

```suggestion
        for param_key in list(params):
```

No strong preference.

philippjfr (commented):
@ahuang11 please review.

panel/tests/chat/test_interface.py
```diff
@@ -358,6 +358,10 @@ def post_callback(instance, event):
         assert chat_interface.objects[1].object == "2"
         assert chat_interface.objects[2].object == "3"
 
+    def test_manual_user(self):
+        chat_interface = ChatInterface(user="New User")
+        assert chat_interface.user == "New User"
```
ahuang11, line 363: I'd recommend also testing the message object
```suggestion
        assert chat_interface.user == "New User"
        chat_interface.send("Test")
        assert chat_interface.objects[0].user == "New User"
```

ahuang11 (approved):
Looks good!

panel/chat/feed.py
Hoxbro: I follow this tip: https://twitter.com/raymondh/status/1251953119556104192

image

philippjfr: `copy` is faster, so you've converted me:

Screenshot 2024-01-31 at 18 31 42

philippjfr: Should have known, lists are slow, dicts are fast.

ahuang11: What if the dict size is much larger? 100k?

philippjfr: There does appear to be a cross-over point, somewhere near 8M on my machine:

Screenshot 2024-01-31 at 18 35 19

Hoxbro:
Looking at the failing test and a bit at the code. I think there are some possibilities: 

1. We promote the parameters `user` and `avatar` from `Interface` to `Feed`.
2. Inherit `send` and `stream` (I think) in `Interface` and check if they are not set as a keyword argument, set them to the parameter if they haven't been set, and then call super. 
3. Update the if statement so we both have them as parameters and in the message kwargs. 
4. Maybe another way forward.

I think I like 2 the most.

ahuang11:
I like 2 too.

ahuang11 (approved):

PR #6296, opened 2024-02-02
Integrate feed

ahuang11:
Replaces Column with Feed; works decently after fixing some bugs in the Feed branch.

https://github.com/holoviz/panel/assets/15331990/58a6b70d-c4eb-448d-9d20-4903da5ff9f8

### DIFF
--- a/panel/chat/feed.py
+++ b/panel/chat/feed.py
@@ -23,7 +23,7 @@ import param
 
 from .._param import Margin
 from ..io.resources import CDN_DIST
-from ..layout import Column, ListPanel
+from ..layout import Feed, ListPanel
 from ..layout.card import Card
 from ..layout.spacer import VSpacer
 from ..pane.image import SVG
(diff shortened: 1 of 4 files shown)


panel/chat/feed.py
```diff
@@ -248,14 +253,17 @@ def __init__(self, *objects, **params):
...
             *objects,
+            load_buffer=self.load_buffer,
             auto_scroll_limit=self.auto_scroll_limit,
             scroll_button_threshold=self.scroll_button_threshold,
+            view_latest=self.view_latest,
             css_classes=["chat-feed-log"],
             stylesheets=self._stylesheets,
             **linked_params
         )
+        self._chat_log.height = None
```
philippjfr, line 266: Why is this needed? Should height be removed from `linked_params`?

ahuang11: `pn.Feed` sets a fixed height if height is not explicitly set since no local scrollbar appears and thus no buffering is enabled.

Maybe it will be alleviated with this: https://github.com/holoviz/panel/pull/6266

philippjfr (approved):

Issue #5901, opened 2023-11-21
Tabulator Frozen Column Positions


CTPassion:
#### Is your feature request related to a problem? Please describe.
I want to freeze two columns in my pn.Widgets.Tabulator but when I do so they are moved to be the first columns of the table.

#### Describe the solution you'd like
I want their position to be customizable or at least remain where they were. They were originally the columns the furthest right representing a TOTAL and EXPECTED value. I would like to freeze them on the right and be able to scroll across the other columns before them.

#### Describe alternatives you've considered
Having them unfrozen and scrolling to the end of the table
**If anyone can think of a workaround for now, I am happy to use that.**

#### Additional context
Frozen:
![image](https://github.com/holoviz/panel/assets/60933427/4d50aaf9-f424-4155-b78a-c4d916d70d1d)

Unfrozen:
![image](https://github.com/holoviz/panel/assets/60933427/4ae970f1-235c-45da-83bb-46ee0a90be77)

maximlt:
Hi @CTPassion, it looks like Tabulator allows to freeze the right most columns (see https://tabulator.info/docs/5.5/layout#frozen-column) so that's a feature that could be implemented in Panel. Do you want to give it a try?

CTPassion:
@maximlt  If I get time I will give it a try :)

suryako:
Ran into this too! @CTPassion did you have any luck? Can give it a try otherwise! :)

CTPassion:
> Ran into this too! @CTPassion did you have any luck? Can give it a try otherwise! :)

Yes I did do it and was easy to implement, I just need to get it in a PR. Will attempt tonight :)

PR #6309, opened 2024-02-05
Choose position to freeze tabulator columns.

CTPassion:
As a solution to my issue #5901 

Columns are ordered by the panel BE and afterwards given the "frozen": True attribute. Hence to freeze to the right of the table, those columns just needed to be put at the end of the ordered list.

The solution to freeze "left" or "right" only I think makes the most sense and is easy enough to implement - unless there is demand for more complicated ways to freeze columns.

The change only adds a new way to freeze columns, the old way will work identically.

https://github.com/holoviz/panel/assets/60933427/89978d39-3695-4f98-8a25-9ab42e445e5b

### DIFF
--- a/examples/reference/widgets/Tabulator.ipynb
+++ b/examples/reference/widgets/Tabulator.ipynb
@@ -37,7 +37,12 @@
     "* **`expanded`** (`list`): The currently expanded rows as a list of integer indexes.\n",
     "* **`filters`** (`list`): A list of client-side filter definitions that are applied to the table.\n",
     "* **`formatters`** (`dict`): A dictionary mapping from column name to a bokeh `CellFormatter` instance or *Tabulator* formatter specification.\n",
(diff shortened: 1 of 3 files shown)


suryako (approved):

philippjfr (approved):
Thanks so much @CTPassion, this is really great! When I first implemented this I did plan to support left and right but didn't follow through. I don't think Tabulator supports frozen columns that are not right or left aligned anyway and the syntax makes sense to me.

CTPassion:
> Thanks so much @CTPassion, this is really great! When I first implemented this I did plan to support left and right but didn't follow through. I don't think Tabulator supports frozen columns that are not right or left aligned anyway and the syntax makes sense to me.

No problem :) Next up would be to freeze groups left or right - but as this wasn't a requirement for me I left it for now.

PR #6311, opened 2024-02-06
Add default help user logo and help_text to ChatFeed

ahuang11:
In many of the panel-chat-examples, I have to do something like:
```python
chat_interface = pn.chat.ChatInterface()
chat_interface.send("Instructions for users", user="System")
chat_interface.serialize()[1:]
```

Which can be a tad tedious when repeatedly done; this PR adds a convenient way to include a help message, and exclude users from serialization. Additionally, it makes the distinction between a "help" message vs a "system" message (which presumably could be a system prompt)
```python
chat_interface = pn.chat.ChatInterface(help_text="Instructions for users")
chat_interface.serialize()
```

### DIFF
--- a/panel/chat/feed.py
+++ b/panel/chat/feed.py
@@ -35,58 +35,6 @@ if TYPE_CHECKING:
     from bokeh.model import Model
     from pyviz_comms import Comm
 
-USER_LOGO = "🧑"
-ASSISTANT_LOGO = "🤖"
-SYSTEM_LOGO = "⚙️"
-ERROR_LOGO = "❌"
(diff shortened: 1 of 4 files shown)


panel/chat/feed.py
```diff
@@ -35,58 +35,6 @@
     from bokeh.model import Model
     from pyviz_comms import Comm
 
-USER_LOGO = "🧑"
```
ahuang11, line 38: Was wondering why adding default avatars here didn't work; turns out this was migrated to `message.py` and this was left behind.
https://github.com/holoviz/panel/blob/main/panel/chat/message.py#L48-L101

philippjfr (approved):

Issue #6295, opened 2024-02-02
Generator function not working


MarcSkovMadsen:
I'm on panel==1.3.8. I thought generator functions worked. But it seems not.

Furthermore they just don't display anything which can make it super hard to understand for beginners.

I would suggest generator functions worked. They can be a simple way for users to create animations and progressively updating apps.

Alternative that some exception was raised.

## Minimum Reproducible Example

```python
import panel as pn
from time import sleep
pn.extension()

def generator():
    yield "hello"
    sleep(0.3)
    yield "world"

pn.pane.Str(generator).servable()
```

The app is just blank and no errors are shown in the terminal.

![image](https://github.com/holoviz/panel/assets/42288570/aaea8cdb-5e89-41ca-ad00-1f2cd19e8764)

## Additional Context

Async Generators work. But `async` is not for *basic* users.

ahuang11:
That is why I commented on https://github.com/holoviz/panel/pull/6213/files#diff-23f2a17c4d19972dadffda8107a86a3900a75f300afe3e5f518ea7e51d9552ecR45-R61

But surprisingly that works haha

MarcSkovMadsen:
> That is why I commented on https://github.com/holoviz/panel/pull/6213/files#diff-23f2a17c4d19972dadffda8107a86a3900a75f300afe3e5f518ea7e51d9552ecR45-R61
> 
> But surprisingly that works haha

For me generator functions are working in the chat interface. My guess is that I filed a bug back in the days and you fixed it?

philippjfr:
Generators really do not seem useful here, a generator function is blocking, therefore even if this did work you'd just get the final output and never see the intermediate outputs.

philippjfr:
There is clearly something wrong here though, you should see something or get an error.

MarcSkovMadsen:
> Generators really do not seem useful here, a generator function is blocking, therefore even if this did work you'd just get the final output and never see the intermediate outputs.

Unless you wrapped the generator in an async generator behind the scenes that did `await` a little bit for each `yield`. The benefit would be simplicity for *basic users*.

philippjfr:
Yeah would have to be executed on a thread but with asyncify that should be easy right?

ahuang11:
Yes this is what is done in ChatInterface: https://github.com/holoviz/panel/blob/main/panel/chat/feed.py#L498-L501

MarcSkovMadsen:
[...]

PR #6312, opened 2024-02-06
Fix `stop` button for sync

ahuang11:
As Philipp mentioned https://github.com/holoviz/panel/issues/6295#issuecomment-1924471058, "Generators really do not seem useful here, a generator function is blocking, therefore even if this did work you'd just get the final output and never see the intermediate outputs."

After, I realized the stop button didn't quite work for sync functions. For some reason, at some point in time, I thought `loop.run_in_executor` would prevent it from blocking the placeholder / stop clicks, like https://stackoverflow.com/questions/41063331/how-to-use-asyncio-with-existing-blocking-library, but I couldn't get it to work.

This PR hides the stop button if it's a sync function and immediately shows the placeholder.

https://github.com/holoviz/panel/assets/15331990/22830b32-65dc-40d3-8595-da61be4361ef

```python
from time import sleep

import panel as pn

pn.extension()


    sleep(5)
    message = ""
    for char in "Echoing User: " + contents:
        sleep(0.05)
        message += char
        yield message


chat_interface = pn.chat.ChatInterface(callback=callback, placeholder_threshold=0.5)
chat_interface.send(
    "Enter a message in the TextInput below and receive an echo!",
    user="System",
    respond=False,
)
chat_interface.servable()

```


panel/chat/feed.py
```diff
@@ -484,21 +484,20 @@ async def _schedule_placeholder(
...
         if iscoroutinefunction(self.callback):
             response = await self.callback(*callback_args)
         elif isasyncgenfunction(self.callback):
             response = self.callback(*callback_args)
         else:
-            response = await loop.run_in_executor(
-                None, partial(self.callback, *callback_args)
-            )
+            self._callback_future = None
+            response = await asyncio.to_thread(partial(self.callback, *callback_args))
```
ahuang11, line 500:
```suggestion
            response = await asyncio.to_thread(self.callback, *callback_args)
```

philippjfr (approved):

panel/chat/feed.py
```diff
@@ -488,6 +491,35 @@ async def _schedule_placeholder(
                 return
             await asyncio.sleep(0.1)
 
+    async def _to_thread(self, func, /, *args, **kwargs):
```
philippjfr, line 494: We don't support 3.8 in Panel.

panel/chat/feed.py
```diff
@@ -496,7 +528,10 @@ async def _handle_callback(self, message, loop: asyncio.BaseEventLoop):
             response = self.callback(*callback_args)
         else:
             self._callback_future = None
-            response = await asyncio.to_thread(self.callback, *callback_args)
+            if isgeneratorfunction(self.callback):
+                response = self._to_async_gen(self.callback(*callback_args))
```
philippjfr, line 532: This doesn't look quite right as is, doesn't there have to be an await somewhere? I'm actually a little confused how generators work in here. How do the yielded values become responses?

ahuang11: Actually, I just tested; there's no need to introduce complexity because serialize response handles the generator fine.

This test should also catch failures.
https://github.com/holoviz/panel/pull/6312/files#diff-071d65f945462dbf3517ddd97071c3da199d35c6f6a74174dbf3eec329826808R530-R531

image


```python
    async def _serialize_response(self, response: Any) -> ChatMessage | None:
        """
        Serializes the response by iterating over it and
        updating the message's value.
        """
        response_message = None
        try:
            if isasyncgen(response):
                self._callback_state = CallbackState.GENERATING
                async for token in response:
                    response_message = self._upsert_message(token, response_message)
                    response_message.show_activity_dot = self.show_activity_dot
            elif isgenerator(response):
                self._callback_state = CallbackState.GENERATING
                for token in response:
                    response_message = self._upsert_message(token, response_message)
                    response_message.show_activity_dot = self.show_activity_dot
            elif isawaitable(response):
                response_message = self._upsert_message(await response, response_message)
            else:
                response_message = self._upsert_message(response, response_message)
        finally:
            if response_message:
                response_message.show_activity_dot = False
        return response_message

```

philippjfr: It's still blocking though right?

ahuang11: Yes it's still blocking. Based on that, I think I should just be calling it as a normal function...? https://github.com/holoviz/panel/pull/6312/commits/7f5af91ca002863f5c83ed30580a08645688176b

philippjfr:
JupyterLite test execution is failing the UI tests, which is unrelated. Will merge.

PR #6315, opened 2024-02-06
Tutorial reorg - AN EXPERIMENT

MarcSkovMadsen:
Based on feedback I would like to experiment with a reorganisation.

1. A 50% as long skills based tutorial broken into 7 parts.
2. A set of reference *basic apps* that users with *basic skills* should be able to build. The goal is to give the users the skills needed to build these apps.
3. A set of supplementary *basic topics*.

I will try with 3 tabs. One for each of the above. I hope this will make things feel more approachable making it clear what is the *core* (*mandatory*) skills sections and what sections can be put together to fit the specific user or use case.


philippjfr:
Should this replace the other PR so they don't go out of sync?

MarcSkovMadsen:
Has been merged into original PR

PR #6348, opened 2024-02-15
Add HoloViz logos as ChatMessage avatars

ahuang11:
Makes it more convenient to just use `user="hvplot"` or `user="lumen"` without manually setting avatar.


MarcSkovMadsen (commented):
Should all the avatars not be available from cdn?

1) we are in control
2) loads faster?

ahuang11:
That sounds good to me; I don't have experience uploading to CDN / using CDN resources though. Is it documented somewhere?

panel/chat/message.py
```diff
@@ -53,6 +53,10 @@
 GPT_3_LOGO = "https://upload.wikimedia.org/wikipedia/commons/thumb/0/04/ChatGPT_logo.svg/1024px-ChatGPT_logo.svg.png?20230318122128"
 GPT_4_LOGO = "https://upload.wikimedia.org/wikipedia/commons/a/a4/GPT-4.png"
 WOLFRAM_LOGO = "https://upload.wikimedia.org/wikipedia/commons/thumb/e/eb/WolframCorporateLogo.svg/1920px-WolframCorporateLogo.svg.png"
+LUMEN_LOGO = "https://raw.githubusercontent.com/holoviz/lumen/main/doc/_static/logo.svg"
+HOLOVIEWS_LOGO = "https://raw.githubusercontent.com/holoviz/holoviews/main/doc/_static/logo.svg"
+HVPLOT_LOGO = "https://github.com/holoviz/hvplot/blob/main/doc/_static/logo.png"
+PANEL_LOGO = "https://raw.githubusercontent.com/holoviz/panel/main/doc/_static/logo.svg"
```
philippjfr, line 59: I don't much like loading any of these from random URLs. If we want to support them we should bundle them into `/dist/assets/logos/` and then load them from there.

philippjfr: While you're at it they should sized consistently and not be too large.

ahuang11: I think SVGs are ideal.

ahuang11:
image

panel/chat/message.py
philippjfr: +1

panel/chat/message.py
```diff
@@ -21,6 +21,8 @@
 
 import param
 
+from panel.io.resources import get_dist_path
```
philippjfr, line 24: Imports should be relative, we already import from `.io.resources` below.

panel/chat/message.py
```diff
@@ -45,14 +47,19 @@
 Avatar = Union[str, BytesIO, bytes, ImageBase]
 AvatarDict = Dict[str, Avatar]
 
+DIST_PATH = get_dist_path()
```
philippjfr, line 50: The dist path cannot be resolved on import, it must be resolved dynamically when rendering.

panel/chat/message.py
```diff
@@ -221,6 +231,10 @@ class ChatMessage(PaneBase):
 
     def __init__(self, object=None, **params):
         self._exit_stack = ExitStack()
+        self._dist_path = get_dist_path()
```
philippjfr, line 234: This cannot ever be cached, it always has to happen at runtime.

philippjfr: In fact it has to happen when the model is generated, not before.

panel/chat/message.py
```diff
@@ -437,7 +436,7 @@ def _avatar_lookup(self, user: str) -> Avatar:
         }
         # now lookup the avatar
         return updated_avatars.get(alpha_numeric_key, self.avatar).format(
-            dist_path=self._dist_path
```
philippjfr: Sadly even this is likely not enough, it has to be recomputed in `_get_model` somehow. I had a look but didn't see an obvious/easy way to do it.

ahuang11: I'm confused because it seems to work:

```python
import panel as pn

pn.extension()

area = pn.chat.ChatInterface(user="wolfram")
area.servable()
```

image

philippjfr: It will work in most basic cases but will break when using prefixed paths, nested routes and a few other conditions.

philippjfr (approved):

PR #6351, opened 2024-02-15
Fix the Azure authentication test

Hoxbro:
Looking at the artifact, it seems like it is stuck on Stay signed in. 

page@14c6f7e59bb4c4e36039c79e8e3f5710-1707976201 458092

This is done without any local test because I don't have the credentials.

### DIFF
--- a/panel/tests/ui/test_auth.py
+++ b/panel/tests/ui/test_auth.py
@@ -104,8 +104,7 @@ def test_azure_oauth(py_file, page):
         page.locator('input[type="submit"]').click(force=True)
 
         page.wait_for_timeout(1000)
-        page.locator('button[type="submit"]').click(force=True)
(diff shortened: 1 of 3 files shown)


panel/tests/ui/test_auth.py
```diff
@@ -104,6 +104,9 @@ def test_azure_oauth(py_file, page):
         page.locator('input[type="submit"]').click(force=True)
         page.locator('input[type="submit"]').click(force=True)
```
Hoxbro, line 105:
```suggestion
```

Hoxbro:
I can see cleanup never finished on UI  - Windows taking, so I have bumped the time to 120 minutes. Hopefully, this will finish it and make it reduce the time in the future.

PR #6373, opened 2024-02-22
Set step in DatetimeRangeSlider to 1 minute

Hoxbro:
I noticed the step was set to 1ms on DatetimeRangeSlider; this sets it to a more sensible default. 

This makes it possible to update the sliders with the arrow key and actual see a change.

https://github.com/holoviz/panel/assets/19758978/41c2e831-cf95-4a49-b9fe-42c322614f53

``` python
import panel as pn
from datetime import datetime
pn.extension()

pn.widgets.DatetimeRangeSlider(value=(datetime(2020, 1, 1), datetime(2020, 3, 1)), start=datetime(2020, 1, 1), end=datetime(2020, 3, 1))
```

### DIFF
--- a/panel/widgets/slider.py
+++ b/panel/widgets/slider.py
@@ -774,8 +774,8 @@ class DatetimeRangeSlider(DateRangeSlider):
     ... )
     """
 
-    step = param.Number(default=600_000, doc="""
-        The step size in ms. Default is 10 min.""")
+    step = param.Number(default=60_000, doc="""
+        The step size in ms. Default is 1 min.""")
 
     _property_conversion = staticmethod(value_as_datetime)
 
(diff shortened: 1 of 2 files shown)


philippjfr:
I'm fine with this but maybe 1 minute or 1 hour is a more "standard" default?

panel/widgets/slider.py
```diff
@@ -774,6 +774,9 @@ class DatetimeRangeSlider(DateRangeSlider):
     ... )
     """
 
+    step = param.Number(default=600_000, doc="""
+        The step size in ms. Default is 10 min.""")
```
Hoxbro, line 778:
```suggestion
    step = param.Number(default=60_000, doc="""
        The step size in ms. Default is 1 min.""")
```

PR #6377, opened 2024-02-22
ChatCopyIcon to inherit from ButtonIcon

ahuang11:
A couple breaking changes:
1. `value` renamed to `text` because ButtonCopyIcon already uses `value`
2. `fill` is dropped
Since it's not exposed on the component gallery, and nested on pn.chat.icons, will this affect anyone?

Hmm also doesn't seem to copy the latest chat message; I think the js_on_click is overriding one another. Is there a way to disable previous watchers?

### DIFF
--- a/panel/chat/icon.py
+++ b/panel/chat/icon.py
@@ -109,6 +109,14 @@ class ChatCopyIcon(ButtonIcon):
 
     _rename: ClassVar[param.Dict] = {"text": None, **ButtonIcon._rename}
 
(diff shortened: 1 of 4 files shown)


ahuang11:
I moved it up @philippjfr but I still can't get it to copy the streaming text strangely.

https://github.com/holoviz/panel/assets/15331990/b62786f6-cbde-43cf-a6ad-0b1b702316d0

PR #6378, opened 2024-02-22
Hide ChatMessage footer until hover

ahuang11:
I wanted to do something like this:
```css
.chat-message:hover .reaction-icons,
.chat-message:hover .footer,
.chat-message:hover .header {
  opacity: 1;
}

.reaction-icons,
.footer,
.header {
  opacity: 0;
  transition: opacity 0.2s;
}
```

To hide all elements until hover over the chat message

https://github.com/holoviz/panel/assets/15331990/3aadfede-b6ae-4031-8e24-4c898157f3fe

But shadow root only allows me to target one at a time I think:

https://github.com/holoviz/panel/assets/15331990/efdc09e1-b514-4f6a-ba29-d228fab8fee1

### DIFF
--- a/panel/chat/message.py
+++ b/panel/chat/message.py
@@ -297,19 +297,17 @@ class ChatMessage(PaneBase):
             css_classes=["timestamp"],
             visible=self.param.show_timestamp,
         )
-        footer_row = Row(
+        self._footer_row = Row(
             self._timestamp_html,
             stylesheets=self._stylesheets + self.param.stylesheets.rx(),
             sizing_mode="stretch_width",
-            css_classes=["footer"] + self.param.show_footer.rx().rx.pipe(
-                lambda x: ["show-on-hover"] if x == "hover" else []
-            ),
-            visible=self.param.show_footer.rx().rx.pipe(bool)
+            css_classes=["footer"],
         )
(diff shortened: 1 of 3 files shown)


philippjfr:
>But shadow root only allows me to target one at a time I think:

I'm quite unsure what you mean.

ahuang11:
I can't do this
```
.chat-message:hover .reaction-icons,
.chat-message:hover .footer,
.chat-message:hover .header {
  opacity: 1;
}

.reaction-icons,
.footer,
.header {
  opacity: 0;
  transition: opacity 0.2s;
}
``` because reaction icons are nested in shadow root

**Comment by philippjfr** [comment]
Wouldn't that have to be?

```
:host(.chat-message:hover) .reaction-icons,
:host(.chat-message:hover) .footer,
:host(.chat-message:hover) .header {
  opacity: 1;
}
```

ahuang11:
I think I tried that too and it didn't work

philippjfr:
I see, the problem with that the DOM is nested pretty deep and it's not quite clear to me how to write a selector that traverses nested shadow hosts.

ahuang11:
Don't think this is possible with shadow doms; closing for now.

Issue #6284, opened 2024-01-30
TextAreaInput. Make it possible to "commit" with ENTER and "add newline" with SHIFT+ENTER


MarcSkovMadsen:
The `TextAreaInput` is what I want to use for chatbots because I need to either enter or copy paste multiline texts.

But currently there is no way to "enter"/ "commit" the value when I am done AND stay in the focused `TextAreaInput` so that I can quickly enter my next question or instruction. Instead I have to use the mouse to click SEND. This confuses users and slows down the chat process.

For the `TextAreaInput` you use ENTER to create a new line.

Please support a chatGPT like workflow as described below.

## Comparison ChatGPT text input

With chatGPT the value is "committed" when you press ENTER and the chat bot responds. You use SHIFT+ENTER to create a new line.

## Comparison `TextInput`

With `TextInput` you value is "committed" when you press ENTER and the chat bot responds

PR #6379, opened 2024-02-23
Add chat area input and integrate

ahuang11:
Closes https://github.com/holoviz/panel/issues/6284
Upon clicking shift enter, triggers `value`. (Actually I should probably remove the shift to be consistent with ChatGPT).

Something seems fishy about handling both `value_input` & `value`. Need to double check logic it doesn't trigger something twice, but other than that pretty exciting!

https://github.com/holoviz/panel/assets/15331990/ddcc6d3e-e53d-4523-94db-034dcdd972d6

- [x] ensure text area is focused first
- [x] docs
- [x] tests


panel/chat/chatarea_input.py
```diff
@@ -0,0 +1,65 @@
...
+    Reference: https://panel.holoviz.org/reference/widgets/TextAreaInput.html
+    :Example:
+
+    >>> ChatAreaInput(
+    ...     name='Description', placeholder='Enter your description here...'
+    ... )
+    """
+
+    resizable = param.ObjectSelector(
+        objects=["both", "width", "height", False],
```
ahuang11, line 37: default to height only.

panel/chat/chatarea_input.py
```diff
@@ -0,0 +1,65 @@
...
+        model = super()._get_model(doc, root, parent, comm)
+        self._register_events("shift_enter_key_down", model=model, doc=doc, comm=comm)
+        return model
+
+    def _process_event(self, event: ShiftEnterKeyDown) -> None:
+        """
+        Clear value on shift enter key down.
+        """
+        with param.parameterized.batch_call_watchers(self):
+            self.value = self.value_input
```
ahuang11, line 62:
```suggestion
        self.value = self.value_input
```

philippjfr (changes requested):
My immediate feeling here is that we should reserve `ChatAreaInput` until we have a more complete and polished component with an inline send button rather than adding it simply for shift-Enter support. Also I don't think we should reset `value`/`value_input` on the Python side, I'd rather create a `ChatMessageEvent` that contains the value of the message and syncs the value with the `value` parameter on submit, while JS resets `value` and `value_input`.

ahuang11:
Discussed with Philipp; we agreed that we can enhance ChatAreaInput over time since drop files is supported in text area (I think)
https://stackoverflow.com/questions/62093212/is-possible-drag-drop-files-into-standard-html-textarea

panel/chat/__init__.py
```diff
@@ -30,6 +30,7 @@
 """
 import importlib as _importlib
 
+from .chatarea_input import ChatAreaInput  # noqa
```
philippjfr, line 33: Would rename the module to `input.py`

panel/chat/chatarea_input.py
```diff
@@ -0,0 +1,78 @@
...
+        doc="""
+        Whether the layout is interactively resizable,
+        and if so in which dimensions: `width`, `height`, or `both`.
+        Can only be set during initialization.""",
+    )
+
+    _widget_type: ClassVar[Type[Model]] = _bkChatAreaInput
+
+    _rename: ClassVar[Mapping[str, str | None]] = {
+        "value": None, **_PnTextAreaInput._rename,
```
philippjfr, line 58: If we want to be able to provide an initial message (which I'm not sure about), we'd have to add:

```python
def _get_properties(self, doc: Document) -> Dict[str, Any]:
    props = super()._get_properties(doc)
    props['value'] = self.value
    return props
```

philippjfr (approved):

Issue #444, opened 2019-05-24
Set bokeh app title and favicon in an easy way


philippjfr:

philippjfr:
Favicon support needs to be added in bokeh first https://github.com/bokeh/bokeh/issues/9086

Setting the title is already supported as the first argument to the servable method, something we should do in all existing examples.

philippjfr:
Templates allow setting the favicon and the app title can be set via the `servable` method.

PR #6380, opened 2024-02-23
Fixes ability to serve notebook with string which contains #

Hoxbro:
Noticed when running `panel convert examples/gallery/*.ipynb doc/how_to/*/examples/*.md --to pyodide-worker --out ./builtdocs/pyodide/ --pwa --index --requirements doc/pyodide_dependencies.json`

I could recreate in a notebook with a cell `dict(**{"background-color": "#444"})`  and running `panel serve notebook.ipynb`

Also did some other fixes related to deprecations and one wrong string start `"` instead of `"""`.

### DIFF
--- a/panel/io/handlers.py
+++ b/panel/io/handlers.py
@@ -151,7 +151,13 @@ def capture_code_cell(cell):
 
     # Remove code comments
     if '#' in cell_out:
-        cell_out = cell_out[:cell_out.index('#')].rstrip()
+        try:
+            # To not remove "#000000"
+            cell_tmp = cell_out[:cell_out.index('#')].rstrip()
+            ast.parse(cell_tmp)
+            cell_out = cell_tmp
+        except SyntaxError:
+            pass
 
     # Use eval mode to check whether cell ends in a statement or an
     # expression that will be rendered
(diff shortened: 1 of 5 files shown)


panel/io/handlers.py
```diff
@@ -151,7 +151,13 @@ def capture_code_cell(cell):
...
     # Remove code comments
     if '#' in cell_out:
-        cell_out = cell_out[:cell_out.index('#')].rstrip()
+        try:
+            # To not remove "#000000"
+            cell_tmp = cell_out[:cell_out.index('#')].rstrip()
+            ast.parse(cell_tmp)
+            cell_out = cell_tmp
+        except SyntaxError:
+            pass
```
philippjfr, line 160: May make another pass at this with a regex.

philippjfr (approved):

panel/io/handlers.py
Hoxbro: My first instinct was to remove the if-statement. Then I thought about doing something more advanced like regex, but I felt that would take too much time to do and could lead to more problems, such as multiline strings containing #.

philippjfr: Yup, it's not at all straightforward. Worth opening an issue maybe.

PR #6381, opened 2024-02-23
install packaging for the CDN upload

Hoxbro:
Package build here: https://github.com/holoviz/panel/actions/runs/8022775766

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -249,8 +249,8 @@ jobs:
         with:
           name: cdn_wheels
           path: |
-            ./build/panel*.whl
-            ./build/bokeh*.whl
+            ./dist/wheels/panel*.whl
+            ./dist/wheels/bokeh*.whl
           if-no-files-found: error
       - name: launch jupyter
         run: |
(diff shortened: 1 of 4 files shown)


.github/workflows/test.yaml
```diff
@@ -240,10 +240,6 @@ jobs:
         run: |
           conda activate test-environment
           pip install --pre ipywidgets_bokeh
-      - name: build pyodide wheels
-        run: |
-          conda activate test-environment
-          python scripts/build_pyodide_wheels.py
```
Hoxbro, line 246: This is run again in the build jupyterlite

scripts/build_pyodide_wheels.py
```diff
@@ -26,8 +31,24 @@
...
 )
+parser.add_argument(
+    "--verify-clean",
+    action="store_true",
+    default=False,
+    help="Check if panel folder is clean before running.",
+)
 args = parser.parse_args()
 
+if args.verify_clean:
```
Hoxbro, line 42: The test is not clean, so I have added it as an option and enforced it on the build.

philippjfr (approved):

Issue #6275, opened 2024-01-27
Custom template don't work anymore


Davide-sd:
#### ALL software version info

* Python 3.10.12
* panel 1.3.8
* param 2.0.2
* bokeh 3.3..4

#### Description of expected behavior and the observed behavior

I have a package that uses Panel to create interactive plots. My last build of the package was done at 21 September 2023. Everything worked fine. I can’t tell the version of panel used on that build because Github logs expired. Based on my local environment, it must have been panel 1.2.2.

The package implements a custom template based on `BootstrapTemplate`, with new parameters.

Yesterday I needed to update the package. However, with panel 1.3.8, the template fails to load and raises a weird error:

```text
ValueError: '/home/davide/Documents/Development/sympy_plot_backends/spb/interactive/bootstrap_spb/bootstrap.html' is not in the subpath of '/home/davide/Documents/Development/envs/plot/lib/python3.10/site-packages/panel/template' OR one path is relative and the other is absolute.
```

#### Complete, minimal, self-contained example code that reproduces the issue

At the moment, I'm unable to create a minimal working example. To see the bug in action you'd have to download and install the master branch from [SymPy Plotting Backend](https://github.com/Davide-sd/sympy-plot-backends/), or follow the stack traceback down below.

This is the structure of the submodule that implements interactivity with panel. Note that I’m using a custom template based on bootstrap.

```text
- interactive
    - bootstrap_spb: folder containing the template files
        - __init__.py: contains my custom template class, `class SymPyBootstrapTemplate(BootstrapTemplate):`,
                       which implements new parameters.
        - bootstrap.css
        - bootstrap.html: template file
    - __init__.py
    - panel.py: file containing the logic to create the interactive application,
                using the template contained into bootstrap_spb.
```

This is how I load the template (code from my `panel.py` file):

```py
from spb.interactive.bootstrap_spb import SymPyBootstrapTemplate

# kwargs contains customization options
template = SymPyBootstrapTemplate(**kwargs)

# add the plot
template.main.append(self.pane)
# add the controls
template.sidebar.append(self.layout_controls)
```

#### Stack traceback and/or browser JavaScript console output

```py
from sympy import *
from spb import *
var("a b c x y z")
plot(a*sin(b*x), params={a: (1, 0, 5), b: (1, 0, 5)}, imodule="panel", servable=True)
```

```text
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
Cell In[3], line 3
[...]

PR #6383, opened 2024-02-24
Catch a ValueError in order to load custom templates

Davide-sd:
Solves issue #6275 

When a custom template is created (for example, based on `BootstrapTemplate`), a bunch of files could be saved in some sub-folder on the disk (the `.html` template file, the `.css` file, the `__init__.py` file).

Latest version of panel is unable to load them, raising a `ValueError`.

This PR catches that `ValueError`, forcing the load to occur with `parse_template`. This makes custom templates working again.


philippjfr (approved):

philippjfr:
Really appreciate it, thanks!

Issue #6390, opened 2024-02-26
Finalize Tutorials for Panel 1.4


MarcSkovMadsen:
This issue keeps track of outstanding issues before Tutorials are ready for Panel 1.4 release.

## Todo

- [ ] Write React to User input (pn_bind.md) tutorial (Philipp)
- [ ] Handle State (state.md) tutorial (Philipp)
  - [ ] Rename to Use State
- [ ] Update Todo App Code (Philipp) and explain the code
- [ ] Update Image Classifier App Code (Philipp) and explain the code
- [ ] Get `{pyodide}` code blocks to render output when build and interactive when run (Philipp).
- [ ] Improve Streaming Dashboard (build_streaming_dashboard.md) to be consistent with other tutorials.
- [ ] Convert (tutorial) mp4 videos to iphone compatible mp4 videos

## Nice to Have

- [ ] Update Class-Based Approach (param.md)
  - [ ] List selected parts of Param user guide as *prerequisites*
- [ ] Update Interactivity (Intermediate) (interactivity)
- [ ] Add Toc Tree to Intermediate Tutorials
- [ ] Review and update the Intermediate and Experienced index.md files

PR #6392, opened 2024-02-26
Add ParamRef baseclass for ParamFunction and ParamMethod

philippjfr:
The class hierarchy for ParamFunction and ParamMethod has been quite confusing since their inception. Specifically `ParamMethod` was a subclass of `ParamFunction`, which conceptually makes no sense. Additionally, since we added generalized support and the concept of non-function references, there was no component in Panel that could take a generic reference and resolve it. Therefore we introduce the new `ParamRef` component in the class hierarchy, which forms a base class for both `ParamMethod` and `ParamFunction`. Not only is this organization now conceptually correct, but it also patches a hole in the Panel API which is to provide a component that can simply render the current value of a reference such as a parameter, widget, or reactive expression. Unlike `ParamFunction`, `ParamRef` will not have any display priority, i.e. a parameter will still be rendered using the `Param` pane, i.e. as a widget, but if e.g. we want to display the output of a widget or reactive expression the `ParamRef` pane comes in helpful.

Let's see an example:

```python
ticker = pn.widgets.Select(name='Ticker', options=tickers)
window = pn.widgets.IntSlider(name='Window Size', value=6, start=1, end=51, step=5)

plots = {
    'altair': pn.pane.Vega(pn.bind(get_altair, ticker, window)),
    'hvplot': pn.pane.HoloViews(pn.bind(get_hvplot, ticker, window)),
    'matplotlib': pn.pane.Matplotlib(pn.bind(get_mpl, ticker, window), format='svg', sizing_mode='stretch_width'),
    'plotly': pn.pane.Plotly(pn.bind(get_plotly, ticker, window))
}

backend = pn.widgets.Select(name='Backend', options=plots)

pn.Row(
    pn.Column(backend, ticker, window),
    pn.param.ParamRef(backend)
).servable()
```
Screenshot 2024-02-26 at 15 06 23

Here the `backend` widget allows selecting between a few values while the `ParamRef` pane can be used to resolve and render the currently selected component.

## Alternatives

Currently there are some alternatives to this new API but they are either not general enough or cumbersome to use.

### 1. Reference binding

One alternative is passing the reference to some other pane to resolve, e.g. this works well if the type of the object is well defined, e.g.:

```python
pn.pane.DataFrame(df_parameter)
``` 

Here the `DataFrame` pane will resolve the value and correctly and efficiently update the output but in our example above there is no pane that can render `Vega`, `HoloViews` etc. types.

### 2. Layout reference binding

In the case where the return value is a list of components we can once again use reference binding to render multiple objects, e.g.:

```python
multi_select = pn.widgets.MultiSelect(options=plots, value=[plots['altair'], plots['hvplot']])

pn.Row(multi_select, pn.FlexBox(objects=multi_select))
```

Screenshot 2024-02-26 at 15 27 42

### 3. Reactive Expression pane

Lastly we do have the option of rendering the component as a reactive expression using the existing `ReactiveExpr` pane. This works but requires a few workarounds for the simpler cases, e.g. to render the value of a select widget we have to call `.rx()` to turn our valid reference into a reactive expression and we also then have to disable rendering of the widgets:

```python
pn.Row(
    pn.Column(backend, ticker, window),
    pn.param.ReactiveExpr(backend.rx(), show_widgets=False)
).servable()
```

## Future work

As I finish writing https://github.com/holoviz/panel/pull/6289 and the remaining tutorials mentioned in https://github.com/holoviz/panel/issues/6390 I will also finally try to work out how to correctly document this important, but as of yet under-explained portion of the Panel API.

### DIFF
--- a/panel/param.py
+++ b/panel/param.py
@@ -1049,6 +1049,10 @@ class ParamFunction(ParamRef):
             return None
         return False
 
+    @classmethod
+    def eval(self, ref):
+        return eval_function_with_deps(ref)
+
 
 class ReactiveExpr(PaneBase):
     """
--- a/panel/param.py
+++ b/panel/param.py
@@ -967,8 +967,6 @@ class ParamMethod(ParamRef):
 
     priority: ClassVar[float | bool | None] = 0.6
 
-    _applies_kw: ClassVar[bool] = True
-
     @classmethod
     def applies(cls, obj: Any) -> float | bool | None:
         return inspect.ismethod(obj) and isinstance(get_method_owner(obj), param.Parameterized)
--- a/panel/param.py
+++ b/panel/param.py
@@ -783,6 +783,8 @@ class ParamRef(ReplacementPane):
         Whether to show a loading indicator while the pane is updating.
         Can be set as parameter or by setting panel.config.loading_indicator.""")
 
+    priority: ClassVar[float | bool | None] = 0
+
     def __init__(self, object=None, **params):
         if 'defer_load' not in params:
             params['defer_load'] = config.defer_load
@@ -799,10 +801,15 @@ class ParamRef(ReplacementPane):
 
     @param.depends('object', watch=True)
     def _validate_object(self):
+        if not callable(self.object):
+            return
         dependencies = getattr(self.object, '_dinfo', {})
         if not dependencies or not dependencies.get('watch'):
             return
-        fn_type = 'method' if type(self) is ParamMethod else 'function'
+        if isinstance(self, ParamMethod):
+            fn_type = 'method'
+        else:
+            fn_type = 'function'
         self.param.warning(
             f"The {fn_type} supplied for Panel to display was declared "
             f"with `watch=True`, which will cause the {fn_type} to be "
@@ -813,6 +820,10 @@ class ParamRef(ReplacementPane):
             "via side-effects, e.g. by modifying internal state of a "
             "class or global state in an application's namespace."
         )
+
+    @classmethod
+    def applies(cls, obj: Any) -> float | bool | None:
+        return bool(resolve_ref(obj))
 
     #----------------------------------------------------------------
     # Callback API
@@ -883,7 +894,7 @@ class ParamRef(ReplacementPane):
     def _update_pane(self, *events):
         callbacks = []
         for watcher in self._internal_callbacks:
-            obj = watcher.inst if watcher.inst is None else watcher.cls
+            obj = watcher.cls if watcher.inst is None else watcher.inst
             if obj is self:
                 callbacks.append(watcher)
                 continue
@@ -908,7 +919,7 @@ class ParamRef(ReplacementPane):
 
     def _link_object_params(self):
         dep_params = resolve_ref(self.object)
-        if not dep_params and not self.lazy and not self.defer_load and not iscoroutinefunction(self.object):
+        if callable(self.object) and not (dep_params or self.lazy or self.defer_load or iscoroutinefunction(self.object)):
             fn = getattr(self.object, '__bound_function__', self.object)
             fn_name = getattr(fn, '__name__', repr(self.object))
             self.param.warning(
@@ -953,6 +964,10 @@ class ParamMethod(ParamRef):
     the method using the param.depends decorator. The method may
     return any object which itself can be rendered as a Pane.
     """
+
+    priority: ClassVar[float | bool | None] = 0.6
+
+    _applies_kw: ClassVar[bool] = True
 
     @classmethod
     def applies(cls, obj: Any) -> float | bool | None:
--- a/panel/param.py
+++ b/panel/param.py
@@ -799,28 +799,6 @@ class ParamRef(ReplacementPane):
             if not self.defer_load:
                 self._replace_pane()
 
-    @param.depends('object', watch=True)
-    def _validate_object(self):
-        if not callable(self.object):
-            return
-        dependencies = getattr(self.object, '_dinfo', {})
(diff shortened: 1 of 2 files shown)


panel/param.py
```diff
@@ -761,14 +761,11 @@ def select(self, selector=None):
         return super().select(selector) + self.layout.select(selector)
 
 
-class ParamMethod(ReplacementPane):
+class ParamRef(ReplacementPane):
```
Hoxbro, line 764: Should we use `ParamRef`, or is it private or abstract?

panel/param.py
```diff
@@ -802,10 +801,15 @@ def __init__(self, object=None, **params):
...
     def _validate_object(self):
-        dependencies = getattr(self.object, '_dinfo', None)
+        if not callable(self.object):
+            return
+        dependencies = getattr(self.object, '_dinfo', {})
         if not dependencies or not dependencies.get('watch'):
             return
-        fn_type = 'method' if type(self) is ParamMethod else 'function'
+        if isinstance(self, ParamMethod):
+            fn_type = 'method'
```
Hoxbro, line 810: `fn_type` should probably be a class variable.

panel/param.py
```diff
@@ -817,13 +821,17 @@ def _validate_object(self):
...
+        return bool(resolve_ref(obj))
+
     #----------------------------------------------------------------
     # Callback API
     #----------------------------------------------------------------
 
     @classmethod
-    def eval(self, function):
-        return eval_function_with_deps(function)
+    def eval(self, ref):
```
Hoxbro, line 833:
```suggestion
    def eval(cls, ref):
```
As you are already making some changes in this area

panel/param.py
```diff
@@ -937,37 +1013,12 @@ def update_pane(*events):
...
-        return super()._get_model(doc, root, parent, comm)
-
-    #----------------------------------------------------------------
-    # Public API
-    #----------------------------------------------------------------
-
     @classmethod
-    def applies(cls, obj: Any) -> float | bool | None:
-        return inspect.ismethod(obj) and isinstance(get_method_owner(obj), param.Parameterized)
+    def eval(self, ref):
```
Hoxbro, line 1017:
```suggestion
    def eval(cls, ref):
```

panel/param.py
philippjfr: Neither private nor abstract.

panel/param.py
Hoxbro: Can `ParamRef` only accept a `function` or any `param.rx` object? If it is the latter, it could have a custom `fn_type`.

panel/param.py
```diff
@@ -817,13 +821,17 @@ def _validate_object(self):
             "class or global state in an application's namespace."
         )
 
+    @classmethod
+    def applies(cls, obj: Any) -> float | bool | None:
```
Hoxbro, line 825:
```suggestion
    def applies(cls, obj: Any) -> bool:
```

panel/param.py
```diff
@@ -895,6 +903,74 @@ def _update_pane(self, *events):
...
+    default ParamMethod will watch all parameters on the class owning
+    the method or can be restricted to certain parameters by annotating
+    the method using the param.depends decorator. The method may
+    return any object which itself can be rendered as a Pane.
+    """
+
+    priority: ClassVar[float | bool | None] = 0.5
+
+    @classmethod
+    def applies(cls, obj: Any) -> float | bool | None:
```
Hoxbro, line 971:
```suggestion
    def applies(cls, obj: Any) -> bool:
```

panel/param.py
```diff
@@ -937,37 +1013,12 @@ def update_pane(*events):
...
-@param.depends(config.param.defer_load, watch=True)
-def _update_defer_load_default(default_value):
-    ParamMethod.param.defer_load.default = default_value
 
-@param.depends(config.param.loading_indicator, watch=True)
-def _update_loading_indicator_default(default_value):
-    ParamMethod.param.loading_indicator.default = default_value
-
-class ParamFunction(ParamMethod):
+class ParamFunction(ParamRef):
```
Hoxbro, line 1021: Shouldn't this have the same eval as `ParamMethod`?

panel/param.py
Hoxbro: I'm actually not entirely sure what the correct thing to do here is. Reflect on the actual method or the possibility of changes in inherited classes.

panel/param.py
philippjfr: It can accept any valid reference (and param allows you to register custom references), `fn_type` makes no sense to me in that context.

panel/param.py
philippjfr: Same, was never quite clear to me.

philippjfr: Okay, I think covariant typing is generally what you should do so in principle it would be fine to narrow the type here. However I actually do not want to box things in for these cases. Subclasses should be allowed to override `applies` and give it a different  return type, e.g. you might make a very specific subclass of `ParamRef` that only supports a certain reference type but returns a float value in `applies` so it actually has a display precedence.

panel/param.py
philippjfr: I think your objection comes from the fact that I'm reference subclasses in this method and I agree that adding a fn_type would be a way out of that but I'm probably just going to move this validation logic instead.

philippjfr (commented):
Will merge, if anyone doesn't like the name please chime in soon so it can be renamed before an RC.

philippjfr:
There's likely some follow up work on `ReactiveExpr` but again that can happen later.

PR #6405, opened 2024-02-28
Include `stylesheets` downstream, including layouts in ChatMessage

ahuang11:
Before (no message background, even though css class is set.):
image

After:
image

```python
"""
Demonstrates how to use the `ChatInterface` to create a chatbot using
OpenAI's with async/await.
"""

import panel as pn

pn.extension()


        model="gpt-3.5-turbo",
        stream=True,
    )
    message = None
    parts = ""
    indicator = pn.indicators.Dial(name="Length")
    async for chunk in response:
        if part:
            parts += part
            indicator.value = len(parts)
            layout = pn.Row(
                pn.pane.Markdown(parts, css_classes=["message"]),
                indicator,
            )
            yield layout


aclient = AsyncOpenAI()
chat_interface.send(
)
chat_interface.servable()

```


panel/chat/message.py
```diff
@@ -498,6 +498,7 @@ def _create_panel(self, value, old=None):
         """
         if isinstance(value, Viewable):
             self._internal = False
+            self._include_stylesheets_inplace(value)
```
philippjfr, line 501: I'm always a little wary for applying stylesheets inplace on objects the user created but I guess it's okay here.

philippjfr (approved):

PR #6406, opened 2024-02-28
Preliminary cleanup of the TS codebase

mattpap:
This PR:
* updates `eslint.js` to match bokehjs'
* updates `tsconfig.json` to match bokehjs (as much as possible)
* fixes all trivial lint errors (styling related, type imports, etc.)
* fixes all errors resulting from `tsconfig.json` changes
  - adds missing or fixes misused `override` and `declare` modifiers
  - fixes a few stray type errors


ahuang11:
Any ideas if anything in this needs changing?
https://github.com/holoviz/panel/pull/6379

mattpap:
> Any ideas if anything in this needs changing?

Yes, quite a few things, but we will resolve this depending on the PR merge order, because some of the required changes may not be applicable until `tsconfig.json` is updated.

mattpap:
@philippjfr, maybe we could get this merged unless there are any concerns? I have some more gradual follow up work I would like to submit.

PR #6407, opened 2024-02-28
Add `message` into `css_classes` to ChatMessage's markup

ahuang11:
Builds upon https://github.com/holoviz/panel/pull/6405

Text wrapped in pn.Row/pn.Column without the grey background looks extremely out of place; this alleviates it by attaching "message" into css_classes, unless the user provides their own `css_classes`.

Before:
image

After:
image

Note, no need to pass `css_classes=["message"]`; simply pass `parts` (previously `pn.pane.Markdown(parts, css_classes=["message"])`)

```python
import panel as pn

pn.extension()


        model="gpt-3.5-turbo",
        stream=True,
    )
    message = None
    parts = ""
    indicator = pn.indicators.Dial(name="Length")
    async for chunk in response:
        if part:
            parts += part
            indicator.value = len(parts)
            layout = pn.Row(
                parts,
                indicator,
            )
            yield layout


aclient = AsyncOpenAI()
chat_interface.send(
)
chat_interface.servable()
```

### DIFF
--- a/panel/chat/message.py
+++ b/panel/chat/message.py
@@ -471,27 +471,33 @@ class ChatMessage(PaneBase):
         return contents, renderer
 
     def _include_stylesheets_inplace(self, obj):
+        if hasattr(obj, "objects"):
+            obj.objects[:] = [
+                self._include_stylesheets_inplace(o) for o in obj.objects
+            ]
+        else:
+            obj = _panel(obj)
         obj.stylesheets = [
             stylesheet for stylesheet in self._stylesheets + self.stylesheets
             if stylesheet not in obj.stylesheets
         ] + obj.stylesheets
-        if hasattr(obj, "objects"):
-            for o in obj.objects:
-                self._include_stylesheets_inplace(o)
+        return obj
 
     def _include_message_css_class_inplace(self, obj):
         if hasattr(obj, "objects"):
-            for o in obj.objects:
+            obj.objects[:] = [
                 self._include_message_css_class_inplace(o)
-        elif isinstance(obj, str):
-            obj = Markdown(obj)
-
+                for o in obj.objects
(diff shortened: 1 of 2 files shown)


panel/chat/message.py
```diff
@@ -479,16 +479,28 @@ def _include_stylesheets_inplace(self, obj):
             for o in obj.objects:
                 self._include_stylesheets_inplace(o)
 
+    def _include_message_css_class_inplace(self, obj):
+        if hasattr(obj, "objects"):
+            for o in obj.objects:
+                self._include_message_css_class_inplace(o)
+        elif isinstance(obj, str):
+            return self._include_message_css_class_inplace(Markdown(obj))
```
philippjfr, line 487: Don't quite understand this one, the `Markdown` object will never be used, will it?

ahuang11: It's just converting it to markdown because str doesn't have attributes like `css_classes` and `sizing_mode` etc

panel/chat/message.py
```diff
@@ -479,16 +479,28 @@ def _include_stylesheets_inplace(self, obj):
             for o in obj.objects:
                 self._include_stylesheets_inplace(o)
 
+    def _include_message_css_class_inplace(self, obj):
+        if hasattr(obj, "objects"):
+            for o in obj.objects:
+                self._include_message_css_class_inplace(o)
+        elif isinstance(obj, str):
+            obj = Markdown(obj)
```
philippjfr, line 487: Still don't understand this, my point was that you're creating this `Markdown` object but it's never used, so why even create it instead of skipping?!

philippjfr: But `obj` is created, then modified and then never used again.

Issue #6298, opened 2024-02-03
Panel with ipywidget does not display in Colab


pkumar0404:
Thanks for contacting us! Please read and follow these instructions carefully, then delete this introductory text to keep your issue easy to read. Note that the issue tracker is NOT the place for usage questions and technical assistance; post those at [Discourse](https://discourse.holoviz.org) instead. Issues without the required information below may be closed immediately.

#### ALL software version info
Google Colab version as of 2024
Python 3.10.12
Panel  1.3.8
Ipywidgets  8.1.1

#### Description of expected behavior and the observed behavior
Combination of ipywidgets and panel does not display any output in Colab.
The combination works find in Jupyter Lab.
Expected behavior on Colab is for the output to be shown.

#### Complete, minimal, self-contained example code that reproduces the issue

```
import panel as pn
import ipywidgets as w
pn.extension()
pn.panel(w.Button(description=“Help!”))

```

#### Screenshots or screencasts of the bug in action

- [ ] I may be interested in making a pull request to address this

Coderambling:
Hi. I ran into the same issue some weeks ago. In Developer tools console, your code shows error: Unable to find supported output mime type on  {output_type: 'display_data', data: {…}, metadata: {…}, displayId: 'some_id', item_id: '4'} in output.binary.js:186. Just before that line, there is a codeblock that tests if the output is one of about 12 datatypes.

```
... else if ("image/gif"in b.data)
                        il(b.data["image/gif"], d["image/gif"], "image/gif", c);
else if ("application/vnd.colab-display-data+json"in b.data)
                        a.u(b.data["application/vnd.colab-display-data+json"], c);
```

Apparently, output_type: 'display_data'does not match any of these data types?

In addition, when I run 

`pn.panel(w.FloatSlider(description='Float'))
`

the console returns the above error, plus:

IPyWidget model could not find a WidgetManager panel.min.js:124 and:

Uncaught TypeError: Cannot read properties of undefined (reading 'set_log_level') VM414:212  , perhaps because bokeh server sets log_level to info? This is according to console returns: [bokeh] setting log level to 'info' bokeh-3.3.4.min.js:184 and:

Uncaught (in promise) TypeError: Cannot read properties of undefined (reading 'set_state') panel.min.js:124

Running the below did not make a difference in any of the above errors.

```
from google.colab import output
output.enable_custom_widget_manager()
```
[...]

PR #6429, opened 2024-03-03
Add warning about use of ipywidgets with Panel in Colab, and to upgrade Panel when using Anaconda Cloud other_nb.md

Coderambling:
Add a warning that ipywidgets + Panel does not work in Colab.

I think this should be pointed out in the documentation (I have wasted considerable time trying to get it to work, which is currently not possible due to Colab constraints regarding comms). 

Maybe even add a warning in the code that detects use of ipywidgets + Colab and provides a warning in the Cell output?

See Discord message by @jbednar and surrounding messages here: https://discord.com/channels/1075331058024861767/1088157184489164831/1206732316513271818

And Github issue here: https://github.com/holoviz/panel/issues/6298

Added mention of Anaconda Cloud as a Notebook environment, and mention that users should upgrade Panel when using it. I am not sure what the correct / optimal !pip install command is. The specific command should probably be added, but at least this gives users a heads-up.

Should Panel have code that checks its own version, and produces  a warning if  the latest version is not being used?

That would seem sensible, also to alert users in local environments that are currently on the latest version, but might not be when the next version comes out. So this would provide them with a heads-up.

### DIFF
--- a/doc/how_to/notebook/other_nb.md
+++ b/doc/how_to/notebook/other_nb.md
@@ -8,6 +8,12 @@ This guide addresses how to develop apps in Google Colab, VSCode, nteract, and o
 
 In the Google Colaboratory notebook, first make sure to load the `pn.extension()`. Panel objects will then render themselves if they are the last item in a notebook cell. Please note that in Colab rendering for each notebook cell is isolated, which means that every cell must reload the Panel extension code separately. This will result in somewhat slower and larger notebook than with other notebook technologies.
 
+* Be aware: using Panel with ipywidgets in Colab does not work, due to Colab's architecture constraints.
+
+### Anaconda Cloud
+
+When using Anaconda Coud, make sure to upgrade the Panel package to the latest version using !pip install
+
 ### VSCode notebook
 
 Visual Studio Code (VSCode) versions 2020.4.74986 and later support ipywidgets, and Panel objects can be used as ipywidgets since Panel 0.10 thanks to `jupyter_bokeh`, which means that you can now use Panel components interactively in VSCode. Ensure you install `jupyter_bokeh` with `pip install jupyter_bokeh` or `conda install -c bokeh jupyter_bokeh` and then enable the extension with `pn.extension()`.
--- a/doc/how_to/notebook/other_nb.md
+++ b/doc/how_to/notebook/other_nb.md
@@ -12,7 +12,7 @@ In the Google Colaboratory notebook, first make sure to load the `pn.extension()
 
 ### Anaconda Cloud
 
-When using Anaconda Coud, make sure to upgrade the Panel package to the latest version using !pip install
+When using Anaconda Cloud, make sure to upgrade the Panel package to the latest version using !pip install
 
 ### VSCode notebook
 


doc/how_to/notebook/other_nb.md
```diff
@@ -8,6 +8,12 @@ This guide addresses how to develop apps in Google Colab, VSCode, nteract, and o
 
 In the Google Colaboratory notebook, first make sure to load the `pn.extension()`. Panel objects will then render themselves if they are the last item in a notebook cell. Please note that in Colab rendering for each notebook cell is isolated, which means that every cell must reload the Panel extension code separately. This will result in somewhat slower and larger notebook than with other notebook technologies.
 
+* Be aware: using Panel with ipywidgets in Colab does not work, due to Colab's architecture constraints.
+
+### Anaconda Cloud
+
+When using Anaconda Coud, make sure to upgrade the Panel package to the latest version using !pip install
```
philippjfr, line 15:
```suggestion
When using Anaconda Cloud, make sure to upgrade the Panel package to the latest version using !pip install
```

Coderambling:
Maybe change the text to explain that this applies not only to Colab, but to any other environments that use the ipywidgets interface to render output? What are those other environments?

doc/how_to/notebook/other_nb.md
```diff
@@ -8,6 +8,12 @@ This guide addresses how to develop apps in Google Colab, VSCode, nteract, and o
 
 In the Google Colaboratory notebook, first make sure to load the `pn.extension()`. Panel objects will then render themselves if they are the last item in a notebook cell. Please note that in Colab rendering for each notebook cell is isolated, which means that every cell must reload the Panel extension code separately. This will result in somewhat slower and larger notebook than with other notebook technologies.
 
+* Be aware: using Panel with ipywidgets in Colab does not work, due to Colab's architecture constraints.
```
MarcSkovMadsen, line 11: I would suggest turning this into a nice warning admonition and making the text more specific.

````markdown

:::{warning}
Panels [Ipywidget](../../reference/panes/IPyWidget.ipynb) pane does not work in Colab due to architecture constraints.
:::

````markdown

File: doc/how_to/notebook/other_nb.md [hunk 3]
```diff
@@ -8,6 +8,12 @@ This guide addresses how to develop apps in Google Colab, VSCode, nteract, and o
...
+
+### Anaconda Cloud
+
+When using Anaconda Cloud, make sure to upgrade the Panel package to the latest version using !pip install
```
MarcSkovMadsen, line 15: I think we should have a separate explanation on upgrading Panel to link to.

It should 1) Describe how to upgrade with pip and conda 2) recommend upgrading `param` and `bokeh` together with Panel 3) recommend optionally upgrading `hvplot` and `holoviews` together with Panel. 4)  Describe how to hard refresh the browser if no output is displayed 5) remember users to upgrade the jupyter lab python environment at the same time.

doc/how_to/notebook/other_nb.md
Coderambling: Agree but I'm not sure what the optimal text should be, in terms of warning (how are those usually formulated on the site?) and technical specifics.Does the Discord discussion I referred to help?

doc/how_to/notebook/other_nb.md
Coderambling: Agree. People that start out need to be aware of the above at earliest opportunity, otherwise they will run into silent fails / confusion.

That's why I feel strongly that, as wel as documenting it, Panel should check itself and its environment after initializing, and emit the appropriate warning in the output. That seems the surest way to alert users.

So I guess a warning that is generated when doing import panel . Should that be a separate PR?

Hoxbro: Are you suggesting we check (and warn if we are not on the latest version) for the latest version of Panel while importing it?

Coderambling: Yes absolutely! Just makes sense. And (separately(?) a warning regarding ipywidgets as indicated above.

doc/how_to/notebook/other_nb.md
Coderambling: Suggested addition below new text.

Best practice is to always

-Check for the latest version of Panel (with command ...)
-If not on the latest version, upgrade to Panel to the latest version with !pip install. See installation guide for further details.

doc/how_to/notebook/other_nb.md
philippjfr: The real problem here continues to be lack of versioning for our docs.

Coderambling: @jbednar Totally agree that it shouldn't break older apps. I am just talking about generating a warning in output / console. Could that in itself causes issues?

With "checking on import" I meant:

-import panel as pn -> Produces warning in output that "... Your Panel version is x, while latest version is y".

Would such a warning in any way cause issues, or impede an app from still working on an older version?

In that case, only show the warning on import panel as pn is the last statement in a file or Notebook cell?

Or only in console or debugger?

Totally agree apps should stay working on older version.

I did NOT mean: do import panel as pn, then anything else happens, besides showing a warning.

Do you agree there is an added value to such a warning?

Advantages:

-Informs users there is a new version available.
-Incentivizes them to look at new version and features of Panel when they revisit their old app after a while.
-Could encourage users to upgrade, and therefore reduce the installed base of older Panel versions that are more and more obsolete.

Separately, I have proposed some changes earlier to the docs to provide a heads-up on this for (new) users. 

Maybe add something in addition to the getting started / installation like: "New Panel versions are released regularly. Please be aware of this, and consider testing and upgrading your app to take advantage of new features and improvements".

One question is: do most users of Panel already know there are frequent versions changes, and that they should take that into consideration?

-People with a Python background: probably.
-Data Science background: maybe not so much?
-Users new to Panel: ?
-Other (future) target groups?

I don't suppose there is any data on the main groups of people using Panel?

Another approach: would it be easy to do Github queries that check, for example, how many old forks of Panel exist? Don't know if that would be a useful metric, but maybe another Github query might reveal insights? Or volume of !pip install vs. !pip install -U over the years? Sorry if this is silly / wrong / irrelevant :-)

doc/how_to/notebook/other_nb.md
Coderambling: Iteration: * Be aware: using ipywidgets with Panel in Colab does currentlly not work, due to Colab's architecture constraints.

doc/how_to/notebook/other_nb.md
Coderambling: My main point is that users should be alerted in any way possible to things like the Colab issue / older versions, rather then have them fail silently, or with errors that do not make clear what the root cause is.

Coderambling:
When using Anaconda Cloud, make sure to upgrade the Panel package to the latest version using !pip install

Alternative:

When using existing / hosted environments, make sure to check the version of Panel and upgrade the Panel package to the latest version using !pip install . Some of these environments, for example Anaconda Cloud, do not always have the latest version of Panel installed.

Coderambling:
Anywidget is considering a similar solution regarding detection of supported environments here:
 https://github.com/manzt/anywidget/issues/37

Coderambling:
Hi all. Is there any Github action required by me at this time regarding this PR?

philippjfr:
This has been addressed in another PR.

PR #6437, opened 2024-03-04
Bump bokeh versions

philippjfr:
Bokeh version pins are resulting in JupyterLite tests not being able to resolve the bokeh wheel.

### DIFF
--- a/panel/tests/ui/template/test_fastlisttemplate.py
+++ b/panel/tests/ui/template/test_fastlisttemplate.py
@@ -26,8 +26,8 @@ def test_fast_list_template_no_console_errors(page):
         "[bokeh] Websocket connection 0 is now open",
         "[bokeh] document idle at",
         "Bokeh items were rendered successfully",
-        "[bokeh] JS/Python version mismatch (3.4.0-rc.",
-        "[bokeh] Library versions: JS"
+        "[bokeh] JS/Python version mismatch",
+        "[bokeh] Library versions: JS (3.4.0-rc."
     ]
     assert len([
         msg for msg in msgs if not any(msg.text.startswith(known) for known in known_messages)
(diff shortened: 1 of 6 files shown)


panel/tests/ui/template/test_fastlisttemplate.py
```diff
@@ -25,7 +25,9 @@ def test_fast_list_template_no_console_errors(page):
         "[bokeh] setting log level to: 'info'",
         "[bokeh] Websocket connection 0 is now open",
         "[bokeh] document idle at",
-        "Bokeh items were rendered successfully"
+        "Bokeh items were rendered successfully",
+        "[bokeh] JS/Python version mismatch",
```
Hoxbro, line 29: Maybe the last additions here should be behind a specific version check of bokeh 3.4.0. Could in the future be a real error.

philippjfr:
```suggestion
        "[bokeh] JS/Python version mismatch (3.4.0-rc.",
```

panel/tests/ui/template/test_fastlisttemplate.py
```diff
@@ -25,7 +25,9 @@ def test_fast_list_template_no_console_errors(page):
         "[bokeh] setting log level to: 'info'",
         "[bokeh] Websocket connection 0 is now open",
         "[bokeh] document idle at",
-        "Bokeh items were rendered successfully"
+        "Bokeh items were rendered successfully",
+        "[bokeh] JS/Python version mismatch (3.4.0-rc.",
+        "[bokeh] Library versions: JS"
```
philippjfr, line 30:
```suggestion
        "[bokeh] JS/Python version mismatch",
        "[bokeh] Library versions: JS (3.4.0-rc."
```

PR #6438, opened 2024-03-05
Rewrite `models/video.ts` and support CSS modules

mattpap:
~~This will have to wait for https://github.com/bokeh/bokeh/pull/13737, because otherwise it requires two pass compilation~~.

### DIFF
--- a/panel/models/video.ts
+++ b/panel/models/video.ts
@@ -1,21 +1,21 @@
 import type * as p from "@bokehjs/core/properties"
+import type {StyleSheetLike} from "@bokehjs/core/dom"
+import {px} from "@bokehjs/core/dom"
 
 import {HTMLBox, HTMLBoxView} from "./layout"
+import video_css from "styles/models/video.css"
 
 export class VideoView extends HTMLBoxView {
   declare model: Video
 
-  protected videoEl: HTMLVideoElement
-  protected containerEl: HTMLElement
-  protected dialogEl: HTMLElement
-  private _blocked: boolean
-  private _time: any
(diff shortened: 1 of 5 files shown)


panel/models/video.ts
```diff
@@ -1,122 +1,133 @@
...
-    this.videoEl.style.height = "100%"
-    if (!this.model.sizing_mode || this.model.sizing_mode === "fixed") {
-      if (this.model.height) {
-        this.videoEl.height = this.model.height
+
+    this.video_el = document.createElement("video")
+    const container_el = document.createElement("div")
+    container_el.className = "pn-video-container"
+    container_el.style.height = "100%"
+    container_el.style.width = "100%"
```
mattpap, line 46: I feel like this wrapper element is unnecessary, but maybe there's a reason for this. If there's one, it would be good to explain it here.

panel/models/widgets.py
```diff
@@ -110,7 +110,7 @@ class Video(HTMLBox):
     throttle = Int(250, help="""
         The frequency at which the time value is updated in milliseconds.""")
 
-    value = Any(help="Encoded file data")
+    value = String(help="Encoded file data")
```
mattpap, line 113: `HTMLVideoElement.src` is of `string` type.

mattpap:
I upgraded to bokeh 3.4.0-rc.2 in this PR to fix the build, allowing full support for CSS modules.

philippjfr:
Merging and then finally resolving the Auth UI test failures in another PR.

PR #6454, opened 2024-03-07
Replace typeof with bokehjs' type queries

mattpap:
Type queries can a bit tricky in JS, so it's better to use standardized APIs for this, so e.g. instead of `typeof x == "boolean"` use `isBoolean(x)` (from `@bokehjs/core/util/types`), etc.


philippjfr (approved):

philippjfr:
UI test failures are unrelated, going to merge.

Issue #6453, opened 2024-03-07
autoreload does not work with notebook files


singharpit94:
I am using panel 1.3.8 and bokeh 3.3.4 and noticed that when using autoreload option with a notebook does not work anymore though the same works with .py files

Simple Reproducer:

```
import panel as pn

pn.extension()
pn.pane.Markdown("# Test ").servable()
```

To reproduce the issue, create a notebook with a cell having the above code and run panel serve --autoreload. Can you please check this?

PR #6459, opened 2024-03-07
Improve autoreload of external packages

philippjfr:
Currently `--autoreload` only reloads the app itself and any one module that has changed. In practice this isn't really enough because most modules will be interconnected pretty deeply, so instead of reloading the specific module we are talking about we will reload all modules that were recorded during startup.

- [x] Reload all modules only if they are in some other package not for relative imports from the app itself
- [x] Fixes https://github.com/holoviz/panel/issues/6453
- [x] Add more tests


panel/io/handlers.py
```diff
@@ -180,11 +180,182 @@ def capture_code_cell(cell):
...
+
+    try:
+        state._launching.append(doc)
+        with _monkeypatch_io(handler._loggers):
+            with patch_curdoc(doc):
+                with profile_ctx(config.profiler) as sessions:
+                    with record_modules(handler=handler):
+                        handler._runner.run(module, post_check)
+                        if post_run:
+                            post_run()
```
Hoxbro, line 231:
```suggestion
        with (
            _monkeypatch_io(handler._loggers),
            patch_curdoc(doc),
            profile_ctx(config.profiler) as sessions,
            record_modules(handler=handler),
        ):
            handler._runner.run(module, post_check)
            if post_run:
                post_run()
```

First supported in Python 3.10. I'm unsure if this will work, e.g, a context manager depends on a setting in the previous one, but I already did the code changes and didn't want to delete it. 

If we want it now we could do something like this:

``` python
with context1 as a,\
     context2 as b:
    pass
``

File: panel/io/handlers.py [hunk 2]
```diff
@@ -180,11 +180,182 @@ def capture_code_cell(cell):
...
+        Keywords:
+            filename (str) : a path to a Python source (".py") file
+
+        '''
+        with open(filename, encoding='utf-8') as f:
+            source = f.read()
+
+        super().__init__(source=source, filename=filename, argv=argv, package=package)
+
+bokeh.application.handlers.directory.NotebookHandler = ScriptHandler
```
Hoxbro, line 349:
```suggestion
bokeh.application.handlers.directory.ScriptHandler = ScriptHandler
```
It's overwritten further down.

panel/io/reload.py
```diff
@@ -101,16 +105,20 @@ async def async_file_watcher(stop_event=None):
             continue
         if path.endswith((".pyc", ".pyo")):
             path = path[:-1]
-        modules[path] = module_name
+        path = str(pathlib.Path(path).resolve())
```
Hoxbro, line 108:
```suggestion
        path = os.path.abspath(path) 
```

Alternative: `os.path.abspath(os.path.realpath(path))`

panel/io/reload.py
```diff
@@ -133,20 +141,35 @@ async def setup_autoreload_watcher(stop_event=None):
...
     """
     Add a file to the watch list.
-
-    All imported modules are watched by default.
     """
     _watched_files.add(filename)
 
+def is_subpath(subpath, path):
+    try:
+        return bool(pathlib.Path(subpath).relative_to(pathlib.Path(path)))
```
Hoxbro, line 149:
```suggestion
        return os.path.commonpath([path, subpath]) == path
```

panel/io/reload.py
```diff
@@ -133,20 +141,35 @@ async def setup_autoreload_watcher(stop_event=None):
...
+
 @contextmanager
-def record_modules():
+def record_modules(applications=None, handler=None):
     """
     Records modules which are currently imported.
     """
+    app_paths = set()
+    if hasattr(handler, '_runner'):
+        app_paths.add(pathlib.Path(handler._runner.path).parent)
```
Hoxbro, line 160:
```suggestion
        app_paths.add(os.path.dirname(handler._runner.path))
```

panel/io/reload.py
```diff
@@ -133,20 +141,35 @@ async def setup_autoreload_watcher(stop_event=None):
...
+    for app in (applications or ()):
+        if not app._handlers:
+            continue
+        for handler in app._handlers:
+            if isinstance(handler, CodeHandler):
+                break
+        else:
+            continue
+        if hasattr(handler, '_runner'):
+            app_paths.add(pathlib.Path(handler._runner.path).parent)
```
Hoxbro, line 170:
```suggestion
            app_paths.add(os.path.dirname(handler._runner.path))
```

panel/io/reload.py
```diff
@@ -166,22 +189,55 @@ def record_modules():
                 continue
 
             if not os.path.isfile(filepath): # e.g. built-in
-                continue
-            _modules.add(module_name)
+               continue
+
+            parent_path = pathlib.Path(filepath).parent
```
Hoxbro, line 194:
```suggestion
            parent_path = os.path.dirname(filepath)
```

Hoxbro (commented):
I have looked at the code. I still want to test it out, but I left some comments.

I don't think we should use Pathlib for this low-level stuff. There is a time penalty for creating each of the objects. I have tried to update some, but please check.

panel/io/reload.py
```diff
@@ -51,10 +55,10 @@ async def awatch(*files, stop_event=None):
 
 IGNORED_MODULES = [
     'bokeh_app',
-    'panel.'
+    'panel.',
```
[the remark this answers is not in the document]
philippjfr: We could probably make this works by clearing the models from `bokeh.model.Model.model_class_reverse_map` at the same time.

panel/io/state.py
```diff
@@ -644,18 +644,16 @@ def kill_all_servers(self) -> None:
...
-                pass
-        self._threads = {}
         for server_id in self._servers:
-            try:
-                self._servers[server_id][0].stop()
-            except AssertionError:  # can't stop a server twice
-                pass
-        self._servers = {}
+            if server_id in self._threads:
+                self._threads[server_id].stop()
```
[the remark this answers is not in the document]
philippjfr: Not quite following.

panel/io/handlers.py
philippjfr: Oooh, neat, but we still support 3.9 😢

philippjfr:
Thanks for the review, I'll wait for tests to pass and then apply your suggestions.

panel/io/state.py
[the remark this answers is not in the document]
Hoxbro: The lines above, was wrapped in try/except.

```
        for thread in self._threads.values():
            try:
                thread.stop()
            except Exception:
                pass
``

File: panel/io/reload.py (hunk 4 above)
Comment (Hoxbro) [reply panel/io/reload.py:149]: This could also be: `os.path.abspath(subpath).startswith(os.path.abspath(path))`

File: pyproject.toml [hunk 10]
```diff
@@ -75,3 +75,4 @@ filterwarnings = [
     "ignore:Deprecated call to `pkg_resources.+?'zope:DeprecationWarning",  # https://github.com/zopefoundation/meta/issues/194
     "ignore: pkg_resources is deprecated as an API:DeprecationWarning:streamz.plugins", # https://github.com/python-streamz/streamz/issues/460
 ]
+asyncio_mode = "auto"
```
philippjfr, line 78:
```suggestion
```

PR #6465, opened 2024-03-09
Normalize connection of signals

mattpap:
Replaces
```ts
this.connect(this.model.properties.some.change, () => ...)
```
with 
```ts
this.on_change(this.model.properties.some, () => ...)
```

### DIFF
--- a/panel/models/perspective.ts
+++ b/panel/models/perspective.ts
@@ -72,82 +72,64 @@ export class PerspectiveView extends HTMLBoxView {
     this.connect(this.model.source.properties.data.change, () => this.setData())
     this.connect(this.model.source.streaming, () => this.stream())
     this.connect(this.model.source.patching, () => this.patch())
-    this.connect(this.model.properties.schema.change, () => {
+
+    const {
+      schema, toggle_config, columns, expressions, split_by, group_by,
+      aggregates, filters, sort, plugin, selectable, editable, theme,
+    } = this.model.properties
+
+    const not_updating = (fn: () => void) => {
+      return () => {
+        if (this._updating) {
+          return
+        }
+        fn()
+      }
+    }
+
+    this.on_change(schema, () => {
       this.worker.table(this.model.schema).then((table: any) => {
         this.table = table
         this.table.update(this.data)
         this.perspective_element.load(this.table)
       })
     })
-    this.connect(this.model.properties.toggle_config.change, () => {
+    this.on_change(toggle_config, () => {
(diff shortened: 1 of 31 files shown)


panel/models/perspective.ts
```diff
@@ -72,82 +72,64 @@ export class PerspectiveView extends HTMLBoxView {
...
     this.connect(this.model.source.streaming, () => this.stream())
     this.connect(this.model.source.patching, () => this.patch())
-    this.connect(this.model.properties.schema.change, () => {
+
+    const {
+      schema, toggle_config, columns, expressions, split_by, group_by,
+      aggregates, filters, sort, plugin, selectable, editable, theme,
+    } = this.model.properties
+
+    const not_updating = (fn: () => void) => {
```
Hoxbro, line 81: I think this is too much abstraction for my likening.

panel/models/vtk/vtkjs.ts
```diff
@@ -8,7 +8,8 @@ export class VTKJSPlotView extends AbstractVTKView {
 
   override connect_signals(): void {
     super.connect_signals()
-    this.connect(this.model.properties.data.change, () => {
+    const {data} = this.model.properties
+    this.on_change(data, () => {
```
Hoxbro, line 12: Why not keep this as one line?

mattpap: Consistency with all other changes I made.

panel/models/perspective.ts
mattpap: This is no different from a simple decorator. Not only this allows to shorten the code, but more importantly, it semantically connects a collection of previously unrelated signal handlers.

Hoxbro: You are overengineering a simple if statement. Your change make the code harder to understand.

panel/models/vtk/vtkjs.ts
Hoxbro: I think you are a bit to consistent then. I don't see any problem with doing what you wrote in the original post: `this.on_change(this.model.properties.some, () => ...)`

But I don't have any strong opinions about this.

panel/models/perspective.ts
philippjfr: On balance I'd agree with keeping the simpler version just because it's such a simple if condition and barely saves any lines of code. That said I don't feel particularly strongly either way.

philippjfr (approved):

PR #6472, opened 2024-03-11
Bump bokeh requirements to 3.4.0rc2

philippjfr:
- [x] Fix JupyterLite test failures
- [ ] Fix auth test failures

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -112,7 +112,7 @@ except Exception:
 ########## dependencies ##########
 
 install_requires = [
-    'bokeh >=3.4.0rc2,<3.5,!=3.4.0.dev*',
+    'bokeh ==3.4.0rc2',
     'param >=2.0.0,<3.0',
     'pyviz_comms >=2.0.0',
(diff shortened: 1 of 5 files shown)


setup.py
```diff
@@ -248,7 +248,7 @@ def run(self):
     'setuptools >=42',
     'requests',
     'packaging',
-    'bokeh ==3.4.0rc1',
+    'bokeh >=3.4.0rc2,<3.5.0',
```
Hoxbro, line 251: I think this will pull in Bokeh 3.4.0.dev8

Hoxbro: It did 
image

philippjfr: On conda?

philippjfr: 😢

Hoxbro: Yes, because of conda.

philippjfr:
Not sure what's up with the auth errors yet but this is a big improvement, so merging.

PR #6479, opened 2024-03-11
Add stylesheet to fileinput button

### LINKED ISSUE #6475: Make FileInput Button adhere to the design
``` python
import panel as pn

pn.extension(design="bootstrap")

file_input_stylesheet = """
input[type="file"].bk-input {
  padding-left: 10px;
}
input[type="file"]::file-selector-button {
  background: var(--surface-color);
  color: var(--surface-text-color);
}
"""

pn.widgets.FileInput().servable()
pn.widgets.Button(name="Button").servable()
pn.widgets.FileInput(stylesheets=[file_input_stylesheet]).servable()
```

image

Hoxbro:
Resolves #6475

I don't know if this is the best place to put the CSS in. 

image

Code in the original issue.

### DIFF
--- a/panel/dist/css/button.css
+++ b/panel/dist/css/button.css
@@ -157,3 +157,12 @@
 :host(.bk-panel-models-widgets-FileDownload) .bk-btn .bk-TablerIcon {
   margin-left: 0.5em;
 }
+
+input[type='file'].bk-input {
+  padding-left: 10px;
+}
+
+input[type='file']::file-selector-button {
+  background: var(--surface-color);
+  color: var(--surface-text-color);
+}
--- a/panel/widgets/input.py
+++ b/panel/widgets/input.py
@@ -10,8 +10,8 @@ import json
 from base64 import b64decode
 from datetime import date, datetime
 from typing import (
-    TYPE_CHECKING, Any, ClassVar, Dict, Iterable, Mapping, Optional, Tuple,
-    Type,
+    TYPE_CHECKING, Any, ClassVar, Dict, Iterable, List, Mapping, Optional,
+    Tuple, Type,
 )
 
 import numpy as np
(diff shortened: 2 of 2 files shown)


panel/dist/css/button.css
```diff
@@ -157,3 +157,12 @@
 :host(.bk-panel-models-widgets-FileDownload) .bk-btn .bk-TablerIcon {
   margin-left: 0.5em;
 }
+
+input[type='file'].bk-input {
+  padding-left: 10px;
+}
+
+input[type='file']::file-selector-button {
```
philippjfr, line 165: How does this selector work?

Hoxbro: I found it by googling, and it seemed to do the trick :upside_down_face:  

https://developer.mozilla.org/en-US/docs/Web/CSS/::file-selector-button

philippjfr: Ah, makes sense!

philippjfr (approved):

monodera:
This change seems to affect the other templates as well and to me these does not look nice. 

Without template:
Screenshot 2024-03-28 at 14 58 32

With Material template:
Screenshot 2024-03-28 at 14 57 56

Is it possible to revert it for other templates and make the change specific to the bootstrap template (if it's not an intension)?

Hoxbro:
I'm not sure what you dislike... Is it the height? If so, I don't think this PR changed that behavior. Try commenting out the `stylesheets` line to see if that restores the behavior.

monodera:
I'm sorry that I forgot to show the one with Panel 1.3.8 (shown below). 

Screenshot 2024-03-28 at 22 36 23

What I don't like for the 1.4 style is the horizontal space from the left edge of the input box to the Browse button. For the Bootstrap theme, the space is equally distributed at the left, top, and bottom, but for (at least) Material and the default templates, the top and bottom paddings are zero. I feel some lack of consistency for the treatment of margins between templates.

Issue #5081, opened 2023-06-09
Perspective Bugs


MarcSkovMadsen:
I'm on the current `main` branch of Panel trying to improve the reference guide of the `Perspective` pane.

## Datetime issue

When I run the below example

```python
import panel as pn
import pandas as pd
import numpy as np
import random
import datetime

pn.extension('perspective')

data = {
    'int': [random.randint(-10, 10) for _ in range(10)],
    'float': [random.uniform(-10, 10) for _ in range(10)],
    'date': [(datetime.datetime.now() + datetime.timedelta(days=i)).date() for i in range(10)],
    'datetime': [(datetime.datetime.now() + datetime.timedelta(hours=i)) for i in range(10)],
    'category': ['Category A', 'Category B', 'Category C', 'Category A', 'Category B',
             'Category C', 'Category A', 'Category B', 'Category C', 'Category A']
}

pn.pane.Perspective(data, width=1000)
```

I would expect the `datetime` column to display the hours and the `category` column to display the categories. Instead it looks like

![image](https://github.com/holoviz/panel/assets/42288570/8f823f4a-ef51-49b0-8756-523a968682ba)

## String issue

It looks there is a string issue too.

```python
data = {
    'category': ['Category A', 'Category B', 'Category C', 'Category A', 'Category B',
             'Category C', 'Category A', 'Category B', 'Category C', 'Category A'],
}

pn.pane.Perspective(data, width=1000)
```

![image](https://github.com/holoviz/panel/assets/42288570/ff23e0cf-9527-4156-8445-fd686aff2768)

## No Issue when data is a dataframe

```python
import datetime
data = {
    'int': [random.randint(-10, 10) for _ in range(10)],
    'float': [random.uniform(-10, 10) for _ in range(10)],
    'date': [(datetime.datetime.now() + datetime.timedelta(days=i)).date() for i in range(10)],
    'datetime': [(datetime.datetime.now() + datetime.timedelta(hours=i)) for i in range(10)],
    'category1': ['Category A', 'Category B', 'Category C', 'Category A', 'Category B',
             'Category C', 'Category A', 'Category B', 'Category C', 'Category A'],
    'category2': ['Category A', 'Category B', 'Category C', 'Category A', 'Category B',
             'Category C', 'Category A', 'Category B', 'Category C', 'Category A'],
}

pn.pane.Perspective(pd.DataFrame(data), width=1000)
```

[...]

PR #6482, opened 2024-03-12
Add title and settings and fix datetime to Perspective

philippjfr:
Fixes remaining issues in https://github.com/holoviz/panel/issues/5081


panel/models/perspective.ts
```diff
@@ -163,6 +170,7 @@ export class PerspectiveView extends HTMLBoxView {
     set_size(container, this.model)
     this.shadow_el.appendChild(container)
 
+    console.log(this.model.schema)
```
philippjfr, line 173:
```suggestion
```

panel/models/perspective.ts
```diff
@@ -221,6 +231,7 @@ export class PerspectiveView extends HTMLBoxView {
         props[option] = value
       }
       this._updating = true
+      console.log(props)
```
philippjfr, line 234:
```suggestion
```

PR #6490, opened 2024-03-13
Fix binary serialization from JS -> Pyodide

philippjfr:
Recent versions of Bokeh made two changes that are causing issues for the pyodide bi-directional syncing:

1. JS -> Python communication can now include binary buffers
2. JS events no longer include a setter causing events to boomerang

This PR handles both issues by:

- Adding a converter for the binary data in the FFI layer of JS -> Python conversion
- Setting a locking variable that prevents events from boomeranging while the serialization is in process

### DIFF
--- a/panel/io/pyodide.py
+++ b/panel/io/pyodide.py
@@ -8,6 +8,7 @@ import json
 import os
 import pathlib
 import sys
+import uuid
 
 from typing import (
     Any, Callable, List, Tuple,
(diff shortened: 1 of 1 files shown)


panel/io/pyodide.py
```diff
@@ -253,6 +254,25 @@ def _process_document_events(doc: Document, events: List[Any]):
...
 
+_current_buffers = []
+_patching = False
+
+def _bytes_converter(value, converter, other):
+    if not hasattr(value, 'buffer'):
+        return value
+    value = dict(value.object_entries())
+    uid = uuid.uuid4().hex
+    _current_buffers.append(Buffer(id=uid, data=value['buffer'].to_bytes()))
```
philippjfr, line 265:
```suggestion
    _current_buffers.append(
        Buffer(id=uid, data=value['buffer'].to_bytes())
    )
```

PR #6491, opened 2024-03-14
Add intermediate Todo App tutorial

MarcSkovMadsen:
A Todo App using a class based approach will help users build extensible and maintainable apps with Panel.

I add a testing tutorial because we don't have that. And I believe its key for becoming an intermediate Panel developer that you can test your apps. I would test my team this.


philippjfr (approved):
Merging as is, will do one round of final reviews for the tutorial materials.

PR #6492, opened 2024-03-14
Ensure tests don't hang

philippjfr:
Tests were hanging and it seems this was due to manually using the `serve` utility rather than using the more recent `serve_and_wait` and `serve_and_request` utilities.

 ~~I'm still a little confused at the difference since internally those utilities don't do anything too different except for having logic to wait until the server is actually up and running before continuing with the test.~~

The issue was that unless you explicitly provided a server_id to `get_server` the server_id in `state._threads` would not match the server_id on `state._servers`, meaning that when it came time to shut down the threads/servers the recent changes I made to the clean up logic would never find the thread to shut down.

### DIFF
--- a/panel/io/server.py
+++ b/panel/io/server.py
@@ -811,10 +811,13 @@ def serve(
     ))
     if threaded:
         kwargs['loop'] = loop = IOLoop(make_current=False) if loop is None else loop
+        if 'server_id' not in kwargs:
+            kwargs['server_id'] = uuid.uuid4().hex
+
         server = StoppableThread(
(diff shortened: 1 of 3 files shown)


panel/io/server.py
```diff
@@ -811,10 +811,13 @@ def serve(
     ))
     if threaded:
         kwargs['loop'] = loop = IOLoop(make_current=False) if loop is None else loop
+        if 'server_id' not in kwargs:
```
philippjfr, line 814:
```suggestion
        # To ensure that we have correspondence between state._threads and state._servers
        # we must provide a server_id here
        if 'server_id' not in kwargs:
```

PR #6497, opened 2024-03-14
Redesign index pages

philippjfr:
Aligns the index page design with the error page changes in https://github.com/holoviz/panel/pull/6496.

Overall I'm aiming for a cleaner and more neutral look for the default pages. The previous version was really nice particularly with the theme toggle etc. but pulling in a whole design toolkit for a simple static page seemed like a lot and with the banner it felt quite busy.

Screenshot 2024-03-14 at 21 46 20

Compared to the old version:

Screenshot 2024-03-14 at 21 47 21

### DIFF
--- a/panel/_templates/index.html
+++ b/panel/_templates/index.html
@@ -1,276 +1,200 @@
 <!DOCTYPE html>
 <html lang="en">
-<head>
-  <!-- Basic Page Needs
-  -------------------------------------------------- -->
-  <meta charset="utf-8">
-  <title>Panel Applications</title>
+  <head>
+    <meta charset="utf-8">
+    <title>Panel Applications</title>
 
-  <!-- Mobile Specific Metas
-  -------------------------------------------------- -->
-  <meta name="viewport" content="width=device-width, initial-scale=1">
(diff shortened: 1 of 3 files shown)


panel/_templates/index.html
```diff
@@ -1,276 +1,200 @@
...
+    </script>
+  </head>
+  <body>
+    <section class="header">
+      <div class="header-content">
+        <div>
+          <img id="panel-logo" src="{{ PANEL_CDN }}images/logo_horizontal.png"/>
+        </div>
+        <div class="title-area">
+          <span id="title">Running Applications</span>
```
ahuang11, line 182: Is this needed? Maybe can remove since serving num apps is the same?

ahuang11:
Do app previews support descriptions, or just name?

philippjfr:
> Do app previews support descriptions, or just name?

To be precise they support neither, these are simply the app filenames after being transformed with `name.replace("_", " ").title()`

> Also, would it make sense to swap the location of search and Running 7 applications?

Think I prefer this.

panel/_templates/index.html
philippjfr: Agreed, will find alternative wording.

Issue #6080, opened 2023-12-18
Update Pyscript version


Hoxbro:
We are using an older version of PyScript (2023.05.1 ->  2023.12.1), which uses an old version of pyodide (0.23.2). This "old" version of pyodide does not have "contourpy >1".

When we update we should revert https://github.com/holoviz/panel/pull/6079

PR #6501, opened 2024-03-15
Remove the removal of contourpy for pyodide wheel

Hoxbro:
See https://github.com/holoviz/panel/issues/6080#issuecomment-1997848859


philippjfr:
I'm actually a little dubious that it should be included by default on Pyodide based builds but I guess in the grand scheme of things an extra 200 KB is not a big deal.

PR #6508, opened 2024-03-16
Add `disabled_enter` to ChatAreaInput

ahuang11:
Previously, while the ChatInterface was disabled (executing callback) and the user presses enter, the user's (potentially long) message disappears into the void.

This PR adds `disabled_enter` to ChatAreaInput so it prevents that from happening.

I implemented this instead of using `disabled` because once disabled, the keyboard focus is lost and users cannot continue typing, which is annoying.

I'm debating between naming it `disabled_submit` or `disabled_enter` or `disabled_input` (to mirror `value_input`


philippjfr:
Please list parameters alphabetically.

ahuang11:
I still want the disabled separately in case users do want it greyed out.

philippjfr:
Fair point.

PR #6514, opened 2024-03-17
Build and publish

Hoxbro:
Release the panel package using a three-step procedure:
1) Build all the steps
2) Approve of build
3) Publish the build

The four builds/publish needed:
- [x] conda
- [x] pip
- [x] npm
- [x] cdn

### Dry-run example
https://github.com/holoviz/panel/actions/runs/8317167537
image

# TODO:
- [x] I need to figure out if CDN should always be published or, like everything else, only on tags. For now, these lines are assuming a tag is needed.
- [x] If we should have an approved step, right now, this is commented out, see https://github.com/holoviz/holoviews/pull/6031 for how it could look 

image

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -17,7 +17,7 @@ defaults:
 
 env:
   PACKAGE: "panel"
-  PYTHON_VERSION: "3.9"
+  PYTHON_VERSION: "3.11"
   NODE_VERSION: "20"
   MPLBACKEND: "Agg"
   SETUPTOOLS_ENABLE_FEATURES: "legacy-editable"
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -10,6 +10,11 @@ on:
   workflow_dispatch:
   schedule:
     - cron: "0 19 * * SUN"
+  # TEMP ONLY FOR THIS PR
+  pull_request:
+    branches:
+    - '*'
+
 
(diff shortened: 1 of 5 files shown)


.github/workflows/build.yaml
```diff
@@ -2,88 +2,164 @@ name: packages
...
-      - name: Set output
-        id: vars
-        run: echo "tag=${GITHUB_REF#refs/*/}" >> $GITHUB_OUTPUT
       - name: conda setup
         run: |
-          conda config --set always_yes True
-          conda config --append channels pyviz/label/dev
-          conda config --append channels bokeh/label/dev
-          conda install -y conda-build anaconda-client build
+          conda install -y conda-build build -c pyviz/label/dev
```
Hoxbro, line 53:
```suggestion
          conda install -y conda-build build
```

.github/workflows/build.yaml
```diff
@@ -2,88 +2,164 @@ name: packages
...
-    - cron: '0 19 * * SUN'
+    - cron: "0 19 * * SUN"
+
+defaults:
+  run:
+    shell: bash -el {0}
 
 env:
+  PACKAGE: "panel"
   PYTHON_VERSION: "3.9"
```
Hoxbro, line 20:
```suggestion
  PYTHON_VERSION: "3.11"
```

philippjfr (approved):

Issue #6502, opened 2024-03-15
Panel performance degredation while running load tests


armaaar:
Once we tried to measure the performance of Panel apps using [K6](https://k6.io/), we noticed huge degredation in performance for only 10 users using the app simultanously, even after enabling defer loading and refactor pages into classes.

We have created a "Hello, World!" Panel page in isolation and ran our load tests against it, which shows huge performance degredation the more the panel server is hit. We suspect that a huge part of it is from how Panel handles sessions. Apparently, sessions destruction (and maybe initialization as well) is not using threads and is blocking the whole server.

Here is a repository that demostrates the performance degredation of panel app, being hit by 10 users simultanously: <https://github.com/armaaar/panel-performance>

#### ALL software version info

- Machine:
  - OS: macOS Ventura 13.3.1
  - CPU: Apple M1 Pro
  - #Cores: 10 (8 performance and 2 efficiency)
  - Memory: 16 GB
- Panel Server:
  - Python: 3.10.13
  - Panel: 1.3.8
  - Bokeh: 3.3.4
  - Param: 2.0.2
  - #Threads: 14
- K6 version: k6 v0.49.0 (go1.21.6, darwin/arm64)

#### Description of expected behavior and the observed behavior

- Sessions handling shouldn't block the server completely
- A hello world page should always load under 1 seconds for 10 different users

#### Complete, minimal, self-contained example code that reproduces the issue

Please refer to our test repository here: <https://github.com/armaaar/panel-performance>. The readme file has instalation, how to run, and results documented.

#### Screenshots or screencasts of the bug in action

Everything can be found in the repo readme: <https://github.com/armaaar/panel-performance/blob/master/README.md>

- [ ] I may be interested in making a pull request to address this

MarcSkovMadsen:
My experience is that Panel can handle the initial load of 2-3 users per second.

Its not been a problem for my use cases so far. But it also means that there are use cases I stay away from currently.

philippjfr:
One quick question after reviewing your document, was the admin page always enabled when you ran the load tests? I'm asking because the admin dashboard itself may be adding non-significant overhead. After disabling it here are the timings I get with latest Panel:

```
     browser_data_received...............................: 1.8 GB  6.4 MB/s
     browser_data_sent...................................: 5.7 MB  20 kB/s
     browser_http_req_duration...........................: avg=385.23ms min=2.03ms   med=245.18ms max=11.74s   p(90)=584.01ms p(95)=736.06ms
[...]

PR #6518, opened 2024-03-18
Avoid overeager garbage collection

philippjfr:
As part of https://github.com/holoviz/panel/issues/6502 I discovered that we still run garbage collection whenever a session is destroyed. Previously we had set up scheduled tasks to run garbage collection AFTER a session is destroyed ensuring that if multiple sessions are closed in succession we only run the task once. In this PR we do two things:

1. Replace the `_on_session_destroyed` handler on `DocumentLifecycleHandler` overriding the default handler which runs garbage collection
2. Add debouncing to the `Document.destroy` garbage collection task so it runs 5 seconds after the last session was destroyed, i.e. if multiple sessions are destroyed in quick succession we only run the garbage collection cycle 5 seconds after the last session is destroyed.

### DIFF
--- a/panel/io/application.py
+++ b/panel/io/application.py
@@ -13,6 +13,9 @@ import bokeh.command.util
 
 from bokeh.application import Application as BkApplication
 from bokeh.application.handlers.directory import DirectoryHandler
+from bokeh.application.handlers.document_lifecycle import (
+    DocumentLifecycleHandler,
+)
 
 from ..config import config
 from .document import _destroy_document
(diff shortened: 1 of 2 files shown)


panel/io/application.py
```diff
@@ -21,11 +24,26 @@
 from .state import set_curdoc, state
 
 if TYPE_CHECKING:
+    from boekh.application.application import SessionContext
```
Hoxbro, line 27:
```suggestion
    from bokeh.application.application import SessionContext
```

panel/io/document.py
```diff
@@ -208,8 +229,10 @@ def _destroy_document(self, session):
             del state_obj[self]
 
     # Schedule GC
-    at = dt.datetime.now() + dt.timedelta(seconds=5)
-    state.schedule_task('gc.collect', gc.collect, at=at)
+    global _last_cleanup
```
Hoxbro, line 232:
```suggestion
    global _panel_last_cleanup
```

Just so nothing else changes the global `_last_cleanup`



I'm curious if we need global for this one. Is it because it is inside an event loop?

philippjfr: Yep, indeed, basically sessions can be destroyed while this task is scheduled and when it runs we want to check that another session wasn't just cleaned up.

panel/io/document.py
```diff
@@ -175,6 +179,13 @@ async def _dispatch_msgs(doc, msgs):
     await asyncio.sleep(0.01)
     _dispatch_write_task(doc, _dispatch_msgs, doc, remaining)
 
+def _garbage_collect():
+    if (new_time:= time.monotonic()-_panel_last_cleanup) < GC_DEBOUNCE:
+        at = dt.datetime.now() + dt.timedelta(seconds=new_time)
+        state.schedule_task('gc.collect', _garbage_collect, at=at)
+        return
+    gc.collect()
```
pankajp, line 187: We can set `_panel_last_cleanup` here so that if a session is destroyed say every 4 seconds (maybe some automated polling script), then we don't end up *never* running cleanup!

That would also mean the first session destroyed will cause very quick gc but any immediately following session destruction will only call gc at 5s intervals.

PR #6519, opened 2024-03-18
Mention code highlighting in Markdown and Feed

ahuang11:
I'm not sure if admonitions can be used in Reference notebooks?


ahuang11:
Is `:::{note}` what you meant?`

philippjfr:
Let's go with :::{note}

ahuang11:
Okay, I think it's ready then.

Issue #6521, opened 2024-03-18
_pending_writes should be non-None when ChatInterface.send with callback is called before served


ahuang11:
```python
import panel as pn

def callback(contents, user, interface):
    for i in range(0, 12):
        message = chat_feed.stream(f"new message {i}", message=message)

chat_feed = pn.chat.ChatInterface(callback=callback)
message = pn.chat.ChatMessage(user="System")
chat_feed.send(message)
chat_feed.servable()
```

```python
Traceback (most recent call last):
  File "/Users/ahuang/miniconda3/envs/panel/lib/python3.10/site-packages/tornado/ioloop.py", line 738, in _run_callback
    ret = callback()
  File "/Users/ahuang/miniconda3/envs/panel/lib/python3.10/site-packages/tornado/ioloop.py", line 762, in _discard_future_result
    future.result()
  File "/Users/ahuang/repos/panel/panel/io/server.py", line 177, in wrapper
    state._handle_exception(e)
  File "/Users/ahuang/repos/panel/panel/io/state.py", line 445, in _handle_exception
    raise exception
  File "/Users/ahuang/repos/panel/panel/io/server.py", line 175, in wrapper
    return await func(*args, **kw)
  File "/Users/ahuang/repos/panel/panel/chat/feed.py", line 537, in _prepare_response
    await self._cleanup_response()
  File "/Users/ahuang/repos/panel/panel/chat/interface.py", line 605, in _cleanup_response
    await super()._cleanup_response()
  File "/Users/ahuang/repos/panel/panel/chat/feed.py", line 543, in _cleanup_response
    with param.parameterized.batch_call_watchers(self):
  File "/Users/ahuang/miniconda3/envs/panel/lib/python3.10/contextlib.py", line 142, in __exit__
    next(self.gen)
  File "/Users/ahuang/miniconda3/envs/panel/lib/python3.10/site-packages/param/parameterized.py", line 307, in batch_call_watchers
    parameterized.param._batch_call_watchers()
  File "/Users/ahuang/miniconda3/envs/panel/lib/python3.10/site-packages/param/parameterized.py", line 2506, in _batch_call_watchers
    self_._execute_watcher(watcher, events)
  File "/Users/ahuang/miniconda3/envs/panel/lib/python3.10/site-packages/param/parameterized.py", line 2468, in _execute_watcher
    watcher.fn(*args, **kwargs)
  File "/Users/ahuang/repos/panel/panel/reactive.py", line 675, in link_cb
[...]

PR #6528, opened 2024-03-19
Ensure no events are dispatched before the websocket is open

philippjfr:
Fixes https://github.com/holoviz/panel/issues/6521

### DIFF
--- a/panel/chat/feed.py
+++ b/panel/chat/feed.py
@@ -26,7 +26,6 @@ from ..layout import Feed, ListPanel
 from ..layout.card import Card
 from ..layout.spacer import VSpacer
 from ..pane.image import SVG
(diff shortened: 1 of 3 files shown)


panel/chat/feed.py
```diff
@@ -260,8 +261,8 @@ def __init__(self, *objects, **params):
         )
 
         # handle async callbacks using this trick
-        self._callback_trigger = Button(visible=False)
-        self._callback_trigger.on_click(self._prepare_response)
+        self.param.watch(self._prepare_response, '_callback_trigger')
+        self.param.trigger('_callback_trigger')
```
philippjfr, line 265:
```suggestion
```

PR #6375, opened 2024-02-22
Add bokeh dev conda channel to editable install

doraaki:
Tried setting up the dev environment for the first time and this command didn't work as Conda wasn't finding bokeh==3.4.0.dev8 which is in the dependency reqiurements. For me needs the dev channel added.

Should the library use the latest dev version of Bokeh?
Also, should I raise an issue first for this small pull request?

PR #6530, opened 2024-03-19
Fix documentation and gallery builds

philippjfr:


Hoxbro:
When you are at it removing `bokeh/label/dev`, can you also revert this https://github.com/holoviz/panel/pull/6375, as it is no longer needed?

PR #4452, opened 2023-02-16
Fix issues sizing components 

philippjfr:
- [x] Fixes FileSelector
- [x] Fixes CrossSelector
- [x] Fixes TextEditor 
- [x] Fixes Perspective styling issues (except for column selector scrollbar issue)
- [x] Fixes DeckGL styling/sizing
- [x] Fixes Vega sizing/styling (
   - fixes https://github.com/holoviz/panel/issues/4379
   - fixes https://github.com/holoviz/panel/issues/3362
   - fixes https://github.com/holoviz/panel/issues/3732
   - fixes https://github.com/holoviz/panel/issues/2219
   - fixes https://github.com/holoviz/panel/issues/4034
   - fixes https://github.com/holoviz/panel/issues/3572
   - fixes https://github.com/holoviz/panel/issues/3571
   - fixes https://github.com/holoviz/panel/issues/1516
- [x] Updates theme handling for Tabulator
- [x] Disables auto-embed/download of images
   - fixes https://github.com/holoviz/panel/issues/4284
- [x] Fixes Param pane sizing 
  - fixes https://github.com/holoviz/panel/issues/1230
  - fixes https://github.com/holoviz/panel/issues/2729
- [x] Fixes ECharts sizing 
   - fixes https://github.com/holoviz/panel/issues/1817
   - fixes https://github.com/holoviz/panel/issues/2236
- [x] Fixes issues with box-sizing in Jupyter
- [x] Fixes issues with clashing ids in Jupyter
- [x] Fixes layout invalidation issue on FlexBox 
   - fixes https://github.com/holoviz/panel/issues/2701
- [x] Fixes ToggleGroup overlap 
   - fixes https://github.com/holoviz/panel/issues/3114
- [x] Fixes RadioBoxGroup sizing 
   - fixes https://github.com/holoviz/panel/issues/3533
- [x] Fixes Plotly resizing 
   - fixes https://github.com/holoviz/panel/issues/3667
- [x] Fixes Progress scrollbar 
   - fixes https://github.com/holoviz/panel/issues/4112

PR #6538, opened 2024-03-21
Small improvements to the PDF model

Hoxbro:
I had problems loading PDFs larger than 2 MB. I thought I had fixed this with: https://github.com/holoviz/panel/issues/3696

I did, but in https://github.com/holoviz/panel/pull/4452 `embed` was changed from True to False in the Panel class, making the problem reappear.  I have updated the Bokeh model to reflect this. 

As I already was in the file I did some minor updates to styling.

Also, set the embed src to always have a size of 100% in width and height.  So the container is what defines the size.

### DIFF
--- a/panel/models/markup.py
+++ b/panel/models/markup.py
@@ -31,6 +31,6 @@ class JSON(Markup):
 
 class PDF(Markup):
 
-    embed = Bool(True, help="Whether to embed the file")
+    embed = Bool(False, help="Whether to embed the file")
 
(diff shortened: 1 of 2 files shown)


philippjfr (approved):

Hoxbro:
An alternative is we set `embed` as True for the PDF class.

philippjfr:
> An alternative is we set embed as True for the PDF class.

No I think this is right.

PR #6553, opened 2024-03-24
Fix releases.md formatting

cdeil:
Tidy up some Markdown formatting issues here: https://holoviz-dev.github.io/panel/about/releases.html

Screenshot 2024-03-24 at 13 04 44


cdeil:
@philippjfr - there's also some bug in `CHANGELOG.md`, the `Version 0.12.0` section appears twice:

https://github.com/holoviz/panel/blob/6e0dd6dd68fa97f26ac13c0d69791ec512efd22c/CHANGELOG.md?plain=1#L1651

https://github.com/holoviz/panel/blob/6e0dd6dd68fa97f26ac13c0d69791ec512efd22c/CHANGELOG.md?plain=1#L2506

I didn't touch it since I don't know what the workflow overall is, whether `doc/about/releases.md` and/or `CHANGELOG.md` need to be manually edited or if one or both are auto-generated somehow. Is it possible to simplify and just have one?

philippjfr:
>I didn't touch it since I don't know what the workflow overall is, whether doc/about/releases.md and/or CHANGELOG.md need to be manually edited or if one or both are auto-generated somehow. Is it possible to simplify and just have one?

Currently they really are just duplicated and manually edited but it may it'd be possible to simply include the CHANGELOG.md in the releases.md

philippjfr:
For now I've manually cleaned things up and kept the duplication but I do want to unify them.

PR #6582, opened 2024-03-26
Do not toggle busy indicator on Column scroll

philippjfr:
Certain parameter/property changes do not warrant an update to the busy indicator. This adds the ability to ignore certain property events and ignores the `Column.scroll_position`.

### DIFF
--- a/panel/layout/base.py
+++ b/panel/layout/base.py
@@ -939,11 +939,6 @@ class Column(ListPanel):
     >>> pn.Column(some_widget, some_pane, some_python_object)
     """
 
-    scroll_position = param.Integer(default=0, doc="""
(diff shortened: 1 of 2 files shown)


MarcSkovMadsen:
Do we always send events across when scrolling? 😵

Wouldn't that make it hard to scale the apps?

philippjfr:
I think we've done so for a little while already but I agree, we should disable this unless absolutely necessary. @ahuang11 Could you open an issue to look into this?

ahuang11:
> Could you open an issue to look into this?

Can you clarify "this"? Do you mean scaling?

philippjfr:
No, sorry, I mean disabling syncing of Column scroll positions.

PR #6593, opened 2024-03-27
Add enter_pressed option to TextInput

Hoxbro:
This adds a counter to the number of times the enter key has been pressed. You can then watch this parameter, and you can make TextInput behave like a search bar. 

TODO:
- [ ] Suggestion for a better name
- [x] Add UI test
- [x] Add documentation
- [x] <s> Maybe add it to `_busy__ignore`</s>, not needed because we use it as an event.


panel/widgets/input.py
```diff
@@ -57,6 +58,9 @@ class TextInput(Widget):
     description = param.String(default=None, doc="""
         An HTML string describing the function of this component.""")
 
+    enter_pressed = param.Integer(default=0, readonly=True, doc="""
```
philippjfr, line 61: Would have this as a `param.Action` and sync the event as a Bokeh Event rather than a counter. May have to do something like `ChatAreaInput` though and send the current value_input in the event since events will be processed faster than the `value_input` property change.

panel/widgets/input.py
```diff
@@ -98,7 +89,41 @@ def from_param(cls, parameter: param.Parameter, onkeyup=False, **params) -> View
...
+        parent: Optional[Model] = None, comm: Optional[Comm] = None
+    ) -> Model:
+        model = super()._get_model(doc, root, parent, comm)
+        self._register_events('enter-pressed', model=model, doc=doc, comm=comm)
+        return model
+
+    def _process_event(self, event) -> None:
+        if event.event_name == 'enter-pressed':
+            self.value = event.value_input
+            self.value_input = event.value_input
```
Hoxbro, line 122: Thinking about using `param.discard_events` for these two.

Hoxbro:
I'm not entirely sure why test_active in panel/tests/chat/test_interface.py is failing after I sub-classed `TextInput`, can you see why @ahuang11?

ahuang11:
It's because ChatAreaInput inherits from TextAreaInput, which is now `class TextAreaInput(_TextInputBase):` instead of `class TextAreaInput(TextInput)`.

However, the tests should check if it's ChatAreaInput not TextInput.

panel/widgets/input.py
philippjfr: Don't think that'll work since it'll suppress the subsequent change event as well since at that point the value won't changed. Events generally are processed before any value change and if we set it here then that rule is violated but I get that the values have to be set before the event.

philippjfr (approved):

PR #6602, opened 2024-03-28
Fix layout persistence issues in dashboard builder

philippjfr:
Fixes issues with layout persistence ensuring that the temporary `.layout` file takes precedence over the notebook metadata.

### DIFF
--- a/panel/io/handlers.py
+++ b/panel/io/handlers.py
@@ -533,7 +533,10 @@ class NotebookHandler(PanelCodeHandler):
 
         layouts, outputs, cells = {}, {}, {}
         for cell_id, out in state._cell_outputs.items():
-            spec = state._cell_layouts[self].get(cell_id, {})
+            if cell_id in self._layout.get('cells', {}):
+                spec = self._layout['cells'][cell_id]
+            else:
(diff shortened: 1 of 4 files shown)


MarcSkovMadsen:
I've 

- created a new python environment and activated it
- pip installed jupyterlab param bokeh and matplotlib in it
- installed the branch `pip install -e .`
- started jupyter lab
- created a new notebook (not named or saved it, but it has been autosaved)

image

panel/io/handlers.py
```diff
@@ -533,7 +533,10 @@ def _render_template(self, doc, path):
```
MarcSkovMadsen, line 533: The preview won't start currently.

MarcSkovMadsen (changes requested):
I should be able to use the preview correctly before merging.

philippjfr:
That's really unrelated to what's being done in this PR. In a development install is different from a package install, so you may have to run:

```
jupyter serverextension enable panel.io.jupyter_server_extension
```

philippjfr:
I'll just cut a dev release for you.

PR #6608, opened 2024-03-29
Make it easy to use dataframes with NestedSelect

### LINKED ISSUE #6604: Make it easy to use DataFrame with NestedSelect
The new `NestedSelect` will be really useful. But 100% of my use cases starts with a Pandas Dataframe. And its currently not very clear how to use that with `NestedSelect`.

I would recommend either

- Documenting how to convert categorical columns of a DataFrame to options.
- Provide one or more methods to easily create a `NestedSelect` from a dataframe.

Personally I would strongly recommend the second option. I would suggest adding methods similar to `get_options_from_dataframe` and `create_from_data` to the `NestedSelect`.

## Example Code

```python
import panel as pn
from bokeh.sampledata.autompg import autompg_clean
import pandas as pd

def _build_nested_dict(df, depth=0, max_depth=None):
    if max_depth is None:
        max_depth = len(df.columns)
    
    # Base case: if depth reaches the last column before values
    if depth == max_depth - 1:
        return df[df.columns[depth]].tolist()
    
    # Recursive case: build dictionary at current depth
    nested_dict = {}
    for value in df[df.columns[depth]].unique():
        filtered_df = df[df[df.columns[depth]] == value]
        nested_dict[value] = _build_nested_dict(filtered_df, depth + 1, max_depth)
    return nested_dict

def get_options_from_dataframe(df, cols=None):
    if not cols:
        cols = list(df.columns)

    df = df[cols].drop_duplicates().sort_values(cols).reset_index(drop=True)
    options = _build_nested_dict(df)
    return options

def test_get_options_from_dataframe():
    data = {
        'continent': ['Europe', 'Europe', 'Asia', 'Asia', 'North America'],
        'country': ['France', 'France', 'Japan', 'Japan', 'USA'],
        'manufacturer': ['Fiat', 'Peugeot', 'Toyota', 'Nissan', 'Ford'],
        'model': ['500', '208', 'Corolla', 'Sentra', 'Mustang']
    }
    df = pd.DataFrame(data)
    options = get_options_from_dataframe(df)
    print(options)

test_get_options_from_dataframe()

def create_from_dataframe(df, cols=None, **params):
    if not cols:
        cols = list(df.columns)

    options = get_options_from_dataframe(df, cols)
    params["levels"]=params.get("levels", cols)
    return pn.widgets.NestedSelect(options=options, **params)


cols = ["origin", "mfr", "name", ]
import panel as pn

pn.extension()

select=create_from_dataframe(autompg_clean, cols=cols, levels=["Origin", "Manufacturer", "Name"])
select.servable()
```

https://github.com/holoviz/panel/assets/42288570/f5559379-1e7e-40f4-8194-6aa366cf8bf2

MarcSkovMadsen:
Closes #6604

### DIFF
--- a/panel/widgets/select.py
+++ b/panel/widgets/select.py
@@ -713,6 +713,33 @@ class NestedSelect(CompositeWidget):
         params["levels"]=params.get("levels", columns)
         return NestedSelect(options=options, **params)
 
+    @classmethod
+    def filter_dataframe(cls, data: pd.DataFrame, value: Dict, columns: List|None=None, **params)->pd.DataFrame:
+        """Returns the subset of the data corresponding to the filter given by the value
+
+        Args:
+            data (pd.DataFrame): The original data to filter
+            value (Dict): A dictionary of key, value filter options
+            columns (List | None, optional): An optional list of column names to use if the keys of
+                the value do not correspond to the column names. Defaults to None.
+
+        Returns:
+            pd.DataFrame: The filtered subset
+        """
+        if data.empty or not value:
+            return data
+
+        if columns:
+            value = {new_key: value for new_key, value in zip(columns, value.values())}
+
+        import pandas as pd
+        mask = pd.Series(True, index=data.index)  # Start with all True
+        for column, filter_value in value.items():
+            mask = mask & (data[column] == filter_value)
+
+        filtered_data = data.loc[mask]
+        return filtered_data
+
 
 class ColorMap(SingleSelectBase):
     """
--- a/panel/widgets/select.py
+++ b/panel/widgets/select.py
@@ -36,6 +36,8 @@ from .button import Button, _ButtonBase
 from .input import TextAreaInput, TextInput
 
 if TYPE_CHECKING:
+    import pandas as pd
+
     from bokeh.model import Model
 
 
@@ -320,6 +322,23 @@ class Select(SingleSelectBase):
             else:
                 return list(itertools.chain(*self.groups.values()))
 
+def _build_nested_dict(data: pd.DataFrame, depth: int=0, max_depth: int|None=None)->Dict|List:
+    if data.empty:
+        return []
+
+    if max_depth is None:
+        max_depth = len(data.columns)
+
+    # Base case: if depth reaches the last column before values
+    if depth == max_depth - 1:
+        return data[data.columns[depth]].tolist()
+
+    # Recursive case: build dictionary at current depth
+    nested_dict = {}
+    for value in data[data.columns[depth]].unique():
+        filtered_df = data[data[data.columns[depth]] == value]
+        nested_dict[value] = _build_nested_dict(filtered_df, depth + 1, max_depth)
+    return nested_dict
 
 class NestedSelect(CompositeWidget):
     """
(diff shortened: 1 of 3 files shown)


panel/widgets/select.py
```diff
@@ -36,6 +36,8 @@
 from .input import TextAreaInput, TextInput
 
 if TYPE_CHECKING:
+    import pandas as pd
```
MarcSkovMadsen, line 39: Should I just import pandas outside of TYPE_CHECKING?

panel/widgets/select.py
```diff
@@ -654,6 +673,73 @@ def _update_options_programmatically(self):
...
+        Returns:
+            pd.DataFrame: The filtered subset
+        """
+        if data.empty or not value:
+            return data
+
+        if columns:
+            value = {new_key: value for new_key, value in zip(columns, value.values())}
+
+        import pandas as pd
```
MarcSkovMadsen, line 735: Should I just import pandas at top of module?

panel/widgets/select.py
```diff
@@ -654,6 +673,73 @@ def _update_options_programmatically(self):
...
+        """
+        if not columns:
+            columns = list(data.columns)
+
+        options = cls.create_options_from_dataframe(data, columns)
+        params["levels"]=params.get("levels", columns)
+        return NestedSelect(options=options, **params)
+
+    @classmethod
+    def filter_dataframe(cls, data: pd.DataFrame, value: Dict, columns: List|None=None, **params)->pd.DataFrame:
```
MarcSkovMadsen, line 717: Should we make this instance method instead? And use `self.value` instead of the provided `value`?

It would make it a bit easier to use but generally less useful.

panel/widgets/select.py
Hoxbro: Importing pandas takes time, and we don't want to do it unless you have to use it. So, what you have done here is correct.

ahuang11:
I'm confused on the difference between the two methods; I think there should only be one method: `from_dataframe` or `from_pandas` to differentiate from polars.

Also, I personally like my shorter implementation :P 
```python

def from_pandas(cls, df, **kwargs):
    cols = list(df.columns)
    grouped = df.groupby(cols[:-1])
    nested = grouped[cols[-1]].apply(lambda x: x.tolist()).to_dict()
    create_nested_defaultdict = lambda depth: defaultdict(
        lambda: create_nested_defaultdict(depth - 1)
    )
    nested_data = create_nested_defaultdict(len(cols) - 1)
    for keys, values in nested.items():
        if isinstance(keys, str):
            keys = (keys,)
        current_dict = nested_data
        for i, key in enumerate(keys):
            if i != len(keys) - 1:
                current_dict = current_dict[key]
            else:
                current_dict[key] = values
    return cls(options=nested_data, **kwargs)
```

panel/widgets/select.py
ahuang11: No I think it should be just in TYPE_CHECKING.

panel/widgets/select.py
ahuang11: I don't follow why we need this; I don't think NestedSelect should have a method for this. If this functionality is added, it should be under the class method's keyword args (e.g. `subset` kwarg)

panel/widgets/select.py
ahuang11: I think it should at least at the start of the function.

philippjfr: Here is fine, if moved to the start of the function you'd import it even if it wasn't needed.

panel/widgets/select.py
philippjfr: Agree with @ahuang11 here, I do not think `filter_dataframe` belongs here. There's at least 20 components that would have equal claim to implementing such a method. I get that it's useful, but we can't provide utilities for everything that's useful.

panel/widgets/select.py
```diff
@@ -654,6 +673,73 @@ def _update_options_programmatically(self):
                 self.value = original_values
                 raise
 
+    @classmethod
+    def create_options_from_dataframe(cls, data: pd.DataFrame, columns: List|None=None)->Dict|List:
```
philippjfr, line 677: I do not think this should be public.

panel/widgets/select.py
```diff
@@ -654,6 +673,73 @@ def _update_options_programmatically(self):
...
+        """Returns the options to use with a NestedSelect
+
+        Args:
```
philippjfr, line 680: These docstrings do not follow the style we use throughout Panel, i.e. they're Google style not NumPy style.

panel/widgets/select.py
```diff
@@ -654,6 +673,73 @@ def _update_options_programmatically(self):
...
+        """
+        if columns is None:
+            columns = list(data.columns)
+
+        data = data[columns].drop_duplicates().sort_values(columns).reset_index(drop=True)
+        options = _build_nested_dict(data)
+        return options
+
+    @classmethod
+    def create_from_dataframe(cls, data: pd.DataFrame, columns: List|None=None, **params)->'NestedSelect':
```
philippjfr, line 696: Again, if we're going to add something like this we should generalize it and add it to other widgets. Even without having a greater plan though, creating the options from columns is just one approach, creating it from an index is just as valid I would think.

philippjfr:
Overall I'm -1 on merging this as is. Before merging something like this I'd like to have an overall plan for generalizing this for widgets other than `NestedSelect` and also not be limited to operating on DataFrame columns, i.e. `from_data` should support array-like objects (ndarrays and Series), DataFrames, indexes and more and we should define an API that can be consistently supported for all widgets.

panel/widgets/select.py
MarcSkovMadsen: The benefit of making it public is that users have more options for controlling performance. With the `create_from_dataframe` option there is no caching and the call will potentially be quite costly if the dataframe is large.

I would argue it should be public.

panel/widgets/select.py
MarcSkovMadsen: Please reconsider. You have `groupby` in hvplot for a reason. This is there for the same reason. Being able to easily create nested selections from a dataset. This is really hard to do correctly and takes time to implement. But its a very common use case.

philippjfr:
I'm still largely -1 on adding the filtering logic here (although potentially open to add some mechanism for filtering). I will fold some of the logic of getting the options into https://github.com/holoviz/panel/pull/6608

PR #6610, opened 2024-03-29
Only apply styling to Bootstrap buttons

Hoxbro:
Fixes https://github.com/holoviz/panel/pull/6479#issuecomment-2026885559

For some reason,, I can't get the updated bootstrap css to work, but If I paste it into the existing one I get the desired behavior.

### DIFF
--- a/panel/theme/css/bootstrap.css
+++ b/panel/theme/css/bootstrap.css
@@ -595,6 +595,15 @@ textarea.bk-input {
   --bs-btn-disabled-border-color: #212529;
 }
 
+input[type='file'].bk-input {
+  padding-left: 10px;
+}
+
+input[type='file']::file-selector-button {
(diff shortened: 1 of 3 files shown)


panel/theme/css/bootstrap.css
```diff
@@ -595,6 +595,15 @@ textarea.bk-input {
   --bs-btn-disabled-border-color: #212529;
 }
 
+input[type='file'].bk-input {
+  padding-left: 10px;
+}
+
+input[type='file']::file-selector-button {
+  background: var(--surface-color);
+  color: var(--surface-text-color);
```
philippjfr, line 604:
```suggestion
  background: var(--bs-surface-bg);
  color: var(--bs-surface-color);
```

philippjfr (approved):

PR #6615, opened 2024-03-29
Ensure all docs pages are indexed

MarcSkovMadsen:
@atisor73 contributed docs on aws but its not been released with Panel 1.4.0 because its not included in an index.

This PR fixes this.


philippjfr (approved):
I'll add a unit test to this PR ensuring that all files are added to an index.

Issue #6598, opened 2024-03-27
ChatAccordion and ChatCard


ahuang11:
Edit: I realized I have the ChatAccordion and ChatStatusCard reversed below in the design, i.e. ChatCard should wrap ChatAccordion

The design should also mention how to integrate this functionality seamlessly in `callback`. My initial thought is yielding/returning a dict with `intermediate=True` `{"The text to stream", "intermediate": True}` creates a ChatMessage with these chat status cards. Maybe context manager too:

```
def callback(contents, user, instance):
    with ChatCard(title="Thinking...") as cc:
        yield cc
        cc.append(...)
        cc.append(...)
        cc.title = "Completed result"
```

https://github.com/holoviz/panel/assets/15331990/a60da3dc-caef-4a95-b2da-da384304b629

--

To improve user experience while LLMs are performing intermediate steps, I'd like to propose two new chat components, namely, `ChatAccordion` and `ChatStatusCard`.

Unlike streaming text, character by character, these two components will improve perception of runtime by outputting each intermediate steps as `ChatStatusCard`s.

Here's an example:
<img width="347" alt="image" src="https://github.com/holoviz/panel/assets/15331990/40fe8db0-ebd6-4ec9-9e03-63b3ec7d2893">

<img width="368" alt="image" src="https://github.com/holoviz/panel/assets/15331990/aaface16-b5a2-498b-b2a1-6ea83418f51d">

The LLM could output a step by step plan to achieve the outcome:
- as each step is determined, a new `ChatStatusCard` is appended to `ChatAccordion`
- as each step is completed, the results are nested under the `ChatStatusCard` which users may open at any time to view
- The status emojis will change from yellow to green if succeeded or red if failed, and perhaps another emoji if needs user input to continue

Once completed, it'll show the final result at the bottom.

<img width="291" alt="image" src="https://github.com/holoviz/panel/assets/15331990/0be8c08d-52a6-454f-b130-5f59387bb1fc">

The intermediate steps can also be hidden.

<img width="300" alt="image" src="https://github.com/holoviz/panel/assets/15331990/f9bec786-b36f-4cf1-9edc-cc399b83adf0">

Here's the abstract version:

<img width="321" alt="image" src="https://github.com/holoviz/panel/assets/15331990/ec168fa3-517d-4e69-a1f6-365ad60362c9">

I believe the API would look like:

```python
class ChatAccordion(Accordion):

    objects = param.List()

    status = param.Selector()

    result = param.Parameter()

[...]

PR #6617, opened 2024-03-30
Implement ChatSteps and ChatStep to show/hide intermediate steps

ahuang11:
Edit: see notebook for updated usage.

Closes https://github.com/holoviz/panel/issues/6598

The relationship:
ChatInterface > ChatFeed > Steps or Step
Steps > Step

Easily available to use by:
`instance.step(**kwargs)` -> `instance.stream(ChatStep(**kwargs))`
`instance.steps(**kwargs)` -> `instance.stream(ChatSteps(**kwargs))`
`steps.step(**kwargs)` -> `steps.append(step)`

`ChatStep`(s) methods mimic `ChatFeed` methods:
`step.stream()`
`step.stream_title()`

```python
import time
import panel as pn

pn.extension()


    with instance.steps() as steps:
        for i in range(0, 4):
            with steps.step() as step:
                message = None
                for c in contents[i:]:
                    time.sleep(0.3)
                    message = step.stream(c, message=message)
                    step.title = f"Processing: {c}"

chat_interface = pn.chat.ChatInterface(callback=callback, callback_exception="verbose")
chat_interface.show()
```

These context managers are optional and the equivalent of:
```python
import time
import panel as pn

pn.extension()


    steps = pn.chat.ChatSteps(title="Running...", status="running")
    instance.stream(steps)
    
    for i in range(0, 4):
        step = pn.chat.ChatStep(margin=0)
        step.title = "Running..."
        step.status = "running"
        steps.append(step)
        message = None
        for c in contents[i:]:
            time.sleep(0.3)
            message = step.stream(c, message=message)
            step.title = f"Processing: {c}"
        step.status = "completed"
    steps.title = step.objects[-1].object
    steps.status = "completed"

chat_interface = pn.chat.ChatInterface(callback=callback, callback_exception="verbose")
chat_interface.show()
```

https://github.com/holoviz/panel/assets/15331990/5706cfed-6103-4c32-b872-7083f94dedcc

- [ ] docs
- [ ] tests
- [ ] serialize


ahuang11:
I like the idea of a `Status` component that inherits from `Card`.

> I like the context manager a lot. But if we introduce it here would it not be a good idea to analyze which components/ flows could benefit from a context manager approach and then make a general design specification?

I'm not 100% sure that's neceessary since it's a matter of implementing `__enter__` & `__exit__` in the corresponding classes if needed--no special method names required.

MarcSkovMadsen:
The reason i think its important with a strategy for Context managers is that its much easier to use a framework if it follows some general principløs. It's much harder if suddenly one component can be used as a Context manager and the rest of the components cannot.

ahuang11:
Can you think of other components that context managers could be useful for?

I imagine indicators, boolean, Card/Accordion.

Also, maybe disabled/loading?

ahuang11:
Cleaner layout:

https://github.com/holoviz/panel/assets/15331990/b0478c81-f6ac-434f-9d4a-a820ab184735

ahuang11:
Okay this is ready for review!

image

panel/chat/feed.py
```diff
@@ -197,6 +198,7 @@ class ChatFeed(ListPanel):
     _stylesheets: ClassVar[list[str]] = [f"{CDN_DIST}css/chat_feed.css"]
 
     def __init__(self, *objects, **params):
+        self._steps = None
```
ahuang11, line 201:
```suggestion
```

panel/chat/message.py
```diff
@@ -335,70 +336,6 @@ def _build_layout(self):
         viewable_params['stylesheets'] = self._stylesheets + self.param.stylesheets.rx()
         self._composite = Row(left_col, right_col, **viewable_params)
 
-    def _get_obj_label(self, obj):
```
ahuang11, line 338: ported to utils.py

panel/chat/message.py
```diff
@@ -782,8 +656,6 @@ def serialize(
 
         Arguments
         ---------
-        obj : Any
```
ahuang11, line 785: No obj arg.

ahuang11:
Added some new methods while implementing https://github.com/holoviz/lumen/pull/560. Need to change the docs, and also add tests, which I realized I didn't commit it and accidentally removed it locally :(

https://github.com/holoviz/panel/assets/15331990/2da12fbd-8e11-40e9-bd6d-c0fcf62ba28f

ahuang11:
ready for review now

https://github.com/holoviz/panel/assets/15331990/44957965-a46c-4735-9297-1479330a1b7d

panel/chat/step.py
```diff
@@ -0,0 +1,239 @@
...
+            params["default_title"] = params["title"]
+
+        super().__init__(*objects, **params)
+        self._render_avatar()
+        self._title_pane = HTML(
+            self.param.title,
+            margin=0,
+            css_classes=["step-title"],
+        )
+        self.header = Row(
```
philippjfr, line 102: Make `header` parameter readonly and then override it with `edit_readonly`.

panel/chat/step.py
```diff
@@ -0,0 +1,239 @@
...
+    def __enter__(self):
+        self.status = "running"
+        return self
+
+    def __exit__(self, exc_type, exc_value, traceback):
+        if exc_type is not None:
+            self.status = "failed"
+            if self.failed_title is None:
+                # convert to str to wrap repr in apostrophes
+                self.failed_title = f"Failed due to: {str(exc_value)!r}"
```
philippjfr, line 117: After the first time the failed_title will be set and therefore not be updated if the context manager is invoked a second time.

panel/chat/step.py
```diff
@@ -0,0 +1,239 @@
...
+        doc="Whether the contents of the Card are collapsed.")
+
+    collapsed_on_success = param.Boolean(
+        default=True,
+        doc="Whether to collapse the card on completion.")
+
+    success_title = param.String(default=None, doc="""
+        Title to display when status is success.""")
+
+    default_avatars = param.Dict(
```
philippjfr, line 37: I associate the word avatar with a user, here it's used for the status. Not sure what I would suggest yet. Also docstring probably has to list the valid status values.

panel/chat/step.py
```diff
@@ -0,0 +1,239 @@
...
+
+    running_title = param.String(
+        default=None,
+        doc="Title to display when status is running."
+    )
+
+    status = param.Selector(
+        default="pending", objects=["pending", "running", "success", "failed"])
+
+    title = param.String(default="", constant=True, doc="""
```
philippjfr, line 68: Probably should be `readonly` rather than `constant`.

philippjfr: Nvm, with the logic on L90 this looks fine.

panel/chat/step.py
```diff
@@ -0,0 +1,239 @@
...
+
+        Arguments:
+        ---------
+        token : str
+            The token to stream.
+        replace : bool
+            Whether to replace the existing text.
+        """
+        if replace:
+            self.title = token
```
philippjfr, line 183: Isn't title constant?!

philippjfr: Also a little confused, I thought the title was always generated from the `<status>_title` parameters? Maybe you can only stream to the `running_title`?

philippjfr: Or allow setting `status`?

panel/chat/feed.py
```diff
@@ -681,6 +683,55 @@ def stream(
         self.param.trigger("_post_hook_trigger")
         return message
 
+    def create_steps(
```
philippjfr, line 686: Maybe this should simply be done by doing `.stream("", user='foo', steps=True)`? The `steps` option might need a better name.

panel/chat/feed.py
```diff
@@ -681,6 +683,55 @@ def stream(
...
+
+        Returns
+        -------
+        The ChatSteps that was created.
+        """
+        steps = ChatSteps(**steps_params)
+        self.stream(steps, user=user, avatar=avatar)
+        return steps
+
+    def attach_step(self, objects: str | list[str] | None = None, **step_params) -> ChatStep:
```
philippjfr, line 714: What happens right now if you `.stream()` when the last message contains `ChatSteps` could that not simply be used to attach a new step?

philippjfr: If this can't be removed I'd suggest renaming maybe, maybe `add_step` or `stream_step`?

philippjfr: `append_step`?

panel/chat/steps.py
```diff
@@ -0,0 +1,111 @@
...
+
+    _stylesheets = [f"{CDN_DIST}css/chat_steps.css"]
+
+    @param.depends("objects", watch=True, on_init=True)
+    def _validate_steps(self):
+        for step in self.objects:
+            if not isinstance(step, ChatStep):
+                raise ValueError(f"Expected ChatStep, got {step.__class__.__name__}")
+
+    def attach_step(self, objects: str | list[str] | None = None, **step_params):
```
philippjfr, line 39: Align with Feed.attach_step rename.

panel/chat/feed.py
ahuang11: I decided against adding the functionality to `stream` since it can get quite complex due to the kwargs `value` accepting `ChatMessage` types and also `message`.

Instead, I compressed the two functions into `append_step`, where it accepts a `steps` kwarg which can be `new`, `append`, or a `ChatSteps` type.

panel/chat/feed.py
ahuang11: See https://github.com/holoviz/panel/pull/6617#discussion_r1625104395

panel/chat/steps.py
```diff
@@ -0,0 +1,111 @@
...
+
+    active = param.Boolean(
+        default=True,
+        doc="Whether additional steps can be automatically appended to the ChatSteps.",
+    )
+
+    css_classes = param.List(
+        default=["chat-steps"],
+        doc="CSS classes to apply to the component.",
+    )
```
philippjfr, line 27: Probably want to declare `objects = param.List(item_type=ChatStep)`.

panel/chat/step.py
```diff
@@ -0,0 +1,261 @@
...
+
+DEFAULT_STATUS_BADGES = {
+    "pending": BooleanStatus(value=False, margin=0, color="primary"),
+    "running": BooleanStatus(value=True, margin=0, color="warning"),
+    "success": BooleanStatus(value=True, margin=0, color="success"),
+    "failed": BooleanStatus(value=True, margin=0, color="danger"),
+}
+
+
+class ChatStep(Card):
```
philippjfr, line 26: Missing docstring.

philippjfr:
Sorry for the delay in reviewing this. The functionality here looks great but now that I'm actually playing around with this I'm finding it pretty confusing. It's not helped by the fact that the examples in the original comment haven't been updated so I'm guessing at the proper way to use this. I'll have to spend some more time with it to form a better opinion on what the API should be but I'm fairly convinced that this isn't mergeable in the current state.

philippjfr:
I think the main thing I'm struggling with is that I don't quite understand the point of the `ChatSteps` component. It's a very light wrapper around a Column and since `ChatFeed.append_step` returns the step component, not the steps component it's not even really part of the API most people will use. So the only thing it really adds is the `active` parameter, which I also do not really understand. What I'd really have liked to see is a writeup that explains the reason for adding each of these components, as it is I'm going to work towards eliminating `ChatSteps`.

philippjfr:
Okay, I've made some changes here:

1. `ChatSteps` no longer exists
2. Renamed the `append_step` method to `ChatFeed.add_step`
3. The logic for appending steps is largely implicit now, i.e. the first time you `add_step` it adds a Column that will collect the steps and subsequent `add_step` calls will be appended as long as the steps are considered active. To replace the logic around an "active" ChatSteps we now treat it as active only as long as the steps are the last item in the feed. A new set of steps can be appended by either specifying a different user OR explicitly setting `append=False`.

I'm quite happy with this approach, but may have missed something, e.g. maybe you should be able to append to an existing set of steps even if it's not the last message. If that's something we need to support we can revisit. For now I'll merge to unblock some of the other work going on.

philippjfr (approved):

PR #6623, opened 2024-03-31
Small typing fixes

cdeil:
Fix some typing problems PyCharm points out when opening the Panel project.
Please check if the imports are correct now.


cdeil:
The type definitions in state.py might also be off. But I don't know what the correct incantation would be, so didn't try to touch it.

Screenshot 2024-03-31 at 11 06 57

cdeil:
PyCharm also infers types from docstrings and then complains that in Panel kwargs are declared as dict in the docstring e.g. here:

https://github.com/holoviz/panel/blob/9bfff01ed982f7ac03d323fc34e68b1aec12b8e7/panel/widgets/base.py#L87-L88

Screenshot 2024-03-31 at 11 18 34

I see https://peps.python.org/pep-0692/ but it seems complicated and new.

What would be best for how? Leave as is? Remove the docstring entry claiming it's dict? Add some type hint?

philippjfr:
>What would be best for how? Leave as is? Remove the docstring entry claiming it's dict? Add some type hint?

Yeah, I'd probably remove it for now.

cdeil:
OK now?

cdeil:
CI fail seems unrelated?
https://github.com/holoviz/panel/actions/runs/8510194144/job/23307330687?pr=6623

philippjfr (approved):

philippjfr:
Yes, unrelated, will merge and fix separately. I'll plan to enable strict type checking for the 1.5 release.

PR #6632, opened 2024-04-01
pyupgrade to Python 3.9

cdeil:
Modernise a bit (mostly type annotations) using https://docs.astral.sh/ruff/rules/#pyupgrade-up

Best reviewed one commit at a time.

E.g. for 4947552bb6f4875b35b7f58efc02b9874ba75601 I ran this:
```
ruff check --target-version py39 --select UP006 --unsafe-fixes
ruff check --target-version py39 --select F401
```

There should be zero real changes in the code here for Python 3.9+, this should be all just a little more consistent and modern way to write some things in Python (better f-strings, type annotations, ...)


Hoxbro:
Can you add these rules to `pyproject.toml`, so it can be caught with pre-commit?

panel/io/jupyter_server_extension.py
```diff
@@ -265,13 +265,11 @@ async def get(self, path=None):
 
         kernel_env = {**os.environ}
         kernel_id = await ensure_async(
-            (
```
Hoxbro, line 268: Maybe reindent this

Hoxbro (commented):
There have been some recent changes in Ruff around multi line f-strings, what version of Ruff are you running?

panel/io/jupyter_server_extension.py
cdeil: Done in e5c42e9116a1311341ff4a78636efbaf9f0b67f8

cdeil:
> Can you add these rules to pyproject.toml, so it can be caught with pre-commit?

Done in 82a19e7dc89de3375bde37b41e6296d9efce79b1

There's more that can be updated, especially with the Optional and other type annotation improvements. But already the diff here is very long and hard to review and I wanted to see if you want to apply such a PR at all.

> what version of Ruff are you running?

```
(panel_dev) panel $ ruff --version
ruff 0.3.4
```

pyproject.toml
```diff
@@ -30,6 +30,7 @@ exclude = [
 ]
 line-length = 165
 fix = true
+target-version = "py39"
```
Hoxbro, line 33: This is not needed, as Ruff is smart enough to use `project.requires-python`, see https://docs.astral.sh/ruff/settings/#target-version

Hoxbro:
> There's more that can be updated, especially with the Optional and other type annotation improvements. But already the diff here is very long and hard to review and I wanted to see if you want to apply such a PR at all.

I think it is fine to split it up into multiple PRs. 

Ruff 0.3.4 has the latest multiline features, just wanted to be sure that we didn't need to update this again, when upgrading Ruff in pre-commit.

pyproject.toml
philippjfr, line 33:
```suggestion
```

philippjfr:
pre-commit.ci autofix

philippjfr:
Sorry this has sat around for so long! Looks good now.

PR #6635, opened 2024-04-01
Use assert_allclose in test_svg_native_size

cdeil:
This makes test_svg_native_size robust by using assert_allclose for the float check.

When I run `doit test_ui` locally I see 4 fails:
```
FAILED panel/tests/ui/pane/test_image.py::test_svg_native_size[False] - assert 507.2109375 == 507.203125
FAILED panel/tests/ui/pane/test_image.py::test_svg_native_size[True] - assert 507.2109375 == 507.203125
FAILED panel/tests/ui/io/test_jupyterlite.py::test_jupyterlite_execution - playwright._impl._errors.TimeoutError: Timeout 20000ms exceeded.
FAILED panel/tests/ui/widgets/test_tabulator.py::test_tabulator_patch_no_height_resize - TimeoutError: wait_until timed out in 5000 milliseconds
```

The other two I can run like this:
```
pytest -v panel/tests/ui/io/test_jupyterlite.py --ui
pytest -v panel/tests/ui/widgets/test_tabulator.py --ui
```
and see that this is the problem:
```
OSError: [Errno 24] Too many open files
```
But I'm not sure why Panel opens so many files or how to fix this.

### DIFF
--- a/panel/tests/ui/pane/test_image.py
+++ b/panel/tests/ui/pane/test_image.py
@@ -1,5 +1,7 @@
 import pytest
 
+from numpy.testing import assert_allclose
+
 pytestmark = pytest.mark.ui
 
 from panel.layout import Row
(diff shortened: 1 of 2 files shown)


philippjfr:
>But I'm not sure why Panel opens so many files or how to fix this.

Unclear to me too, suspect it must be sockets not files.

philippjfr:
Will merge, failing tests are fixed separately.

PR #6637, opened 2024-04-01
Replace OrderedDict with dict

cdeil:
Modernise code base a bit by replacing OrderedDict with dict.
Should be complementary and not touch the lines from #6632 I hope.

### DIFF
--- a/panel/models/deckgl.py
+++ b/panel/models/deckgl.py
@@ -9,8 +9,6 @@ And now DeckGL provides Python bindings. See
 - [PyDeck Repo](https://github.com/uber/deck.gl/tree/master/bindings/pydeck)
 """
 
-from collections import OrderedDict
-
 from bokeh.core.properties import (
     Any, Bool, Dict, Either, Instance, Int, List, Override, String,
 )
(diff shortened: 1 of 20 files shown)


panel/models/deckgl.py
```diff
@@ -49,20 +47,20 @@ def __javascript__(cls):
...
-        ]),
+        'paths': {
+            "h3": f"{config.npm_cdn}/h3-js@3.7.2/dist/h3-js.umd",
+            "deck-gl": f"{config.npm_cdn}/deck.gl@8.6.7/dist.min",
+            "deck-json": f"{config.npm_cdn}/@deck.gl/json@8.6.7/dist.min",
+            "loader-csv": f"{config.npm_cdn}/@loaders.gl/csv@3.1.7/dist/dist.min",
+            "loader-json": f"{config.npm_cdn}/@loaders.gl/json@3.1.7/dist/dist.min",
+            "loader-tiles": f"{config.npm_cdn}/@loaders.gl/3d-tiles@3.1.7/dist/dist.min",
+            "mapbox-gl": "https://api.mapbox.com/mapbox-gl-js/v3.0.1/mapbox-gl",
+            "carto": "{config.npm_cdn}/@deck.gl/carto@^8.7.0/dist.min.js",
```
cdeil, line 62: Probably a bug here with missing f-string?

Don't want to mix it in here, better as separate PR.

panel/util/__init__.py
```diff
@@ -124,8 +124,6 @@ def abbreviated_repr(value, max_length=25, natural_breaks=(',', ' ')):
         end_char = ''
         if isinstance(value, list):
             end_char = ']'
-        elif isinstance(value, OrderedDict):
```
cdeil, line 127: Is this OK? Or is special type cast or case needed for OrderedDict still here?

panel/tests/test_util.py
```diff
@@ -87,16 +84,6 @@ def test_abbreviated_repr_list():
     assert abbreviated_repr(['some really, really long string']) == "['some really, ...]"
 
 
-def test_abbreviated_repr_ordereddict():
```
Hoxbro, line 90: I think we should still support OrderedDict, even though the code base does not use it.

panel/util/__init__.py
Hoxbro: I think it is fine to keep it.

panel/widgets/select.py
```diff
@@ -1361,12 +1359,12 @@ def _apply_selection(self, event):
         """
         selected = event.obj is self._buttons[True]
 
-        new = OrderedDict([(k, self._items[k]) for k in self._selections[not selected]])
+        new = {(k, self._items[k]) for k in self._selections[not selected]}
```
Hoxbro, line 1362: This is wrong (and the same below). You are making a set with tuples, not a dictionary.

panel/models/deckgl.py
Hoxbro: I'm fine with fixing glaring bugs like this one :+1:

philippjfr: Will fix separately.

panel/widgets/select.py
philippjfr, line 1362:
```suggestion
        new = {k: self._items[k] for k in self._selections[not selected]}
```

panel/widgets/select.py
```diff
@@ -1361,12 +1359,12 @@ def _apply_selection(self, event):
...
         old = self._lists[selected].options
         other = self._lists[not selected].options
 
-        merged = OrderedDict([(k, k) for k in list(old)+list(new)])
-        leftovers = OrderedDict([(k, k) for k in other if k not in new])
+        merged ={(k, k) for k in list(old)+list(new)}
+        leftovers = {(k, k) for k in other if k not in new}
```
philippjfr, line 1367:
```suggestion
        merged = {k: k for k in list(old)+list(new)}
        leftovers = {k: k for k in other if k not in new}
```

philippjfr (approved):

PR #6639, opened 2024-04-01
Fix card header margins

ahuang11:
The card header CSS seemed to be all over the place and imbalanced; this PR tries to make it less complex and have a balanced margin/padding.

Before (notice the left has a bit more padding vs the right):
image

After:
image

### DIFF
--- a/panel/models/card.ts
+++ b/panel/models/card.ts
@@ -5,6 +5,14 @@ import type * as p from "@bokehjs/core/properties"
 
 import card_css from "styles/models/card.css"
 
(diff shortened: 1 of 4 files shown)


philippjfr:
Merging despite the failures as I'm resolving those in https://github.com/holoviz/panel/pull/6641

philippjfr (approved):
I would have accepted one change for the chevrons and the margins but duplicating them like you have done here is a bit confusing.

philippjfr:
Really appreciate these changes btw, I've been planning to do this for a **long** time.

PR #6662, opened 2024-04-04
defined callback_user in lines 104 and 382

shwetabh-23:
I added the line : 
callback_avatar = param.ClassSelector(class_=(str, BinaryIO), doc="""
        The default avatar to use for the entry provided by the callback.
        Takes precedence over `ChatEntry.default_avatars` if set; else, if None,
        defaults to the avatar set in `ChatEntry.default_avatars` if matching key exists.
        Otherwise defaults to the first character of the `callback_user`.""")
and changed line : 
avatar = None

to : 
avatar = self.callback_avatar

### DIFF
--- a/panel/chat/feed.py
+++ b/panel/chat/feed.py
@@ -33,7 +33,7 @@ if TYPE_CHECKING:
     from bokeh.model import Model
     from pyviz_comms import Comm
 
-
+from typing import BinaryIO
 PLACEHOLDER_SVG = """
     <svg xmlns="http://www.w3.org/2000/svg" class="icon icon-tabler icon-tabler-loader-3" width="35" height="35" viewBox="0 0 24 24" stroke-width="2" stroke="currentColor" fill="none" stroke-linecap="round" stroke-linejoin="round">
         <path stroke="none" d="M0 0h24v24H0z" fill="none"></path>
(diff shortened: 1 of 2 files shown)


ahuang11:
Thanks for creating this!

Could you please run
`pre-commit install`

Also, can you add a few tests like https://github.com/holoviz/panel/commit/a1420603fb5da14d45c106e3813a5b38f7a3b1aa#diff-5408514bec6f5895a04fa7826621bcce676dd161c74eff71bce900e6c677a584L595-L600

shwetabh-23:
I ran the "pre-commit install" as suggested and added a line (384): 

user = self.callback_user
        avatar = self.callback_avatar

        assert avatar == "👨"
let me know if this is correct or if some modifications are to be made

shwetabh-23:
I cannot find both files at the mentioned directories. Please confirm from your end. in the main directory (https://github.com/holoviz/panel/tree/main), I get to panel/tests/widgets and there is no test_chat.py file and no chat.py file at panel/widgets. what is going wrong?

ahuang11:
Woops, we moved them to panel/tests/chat/

shwetabh-23:
okay, can you verify if this is correct? I've added an assertion in the test_feed.py file under the function test_default_avatars_superseded_by_callback_avatar.

ahuang11:
I'd explicitly set callback_avatar somewhere.

ahuang11:
Implemented in https://github.com/holoviz/panel/pull/6958

PR #6669, opened 2024-04-04
Update image path for `template_areas.png`

jrycw:
Currently, the image `template_areas.png` is not displayed on the [web](https://panel.holoviz.org/explanation/components/components_overview.html). This PR proposes a fix for this issue.


philippjfr (approved):
Good catch!

PR #6670, opened 2024-04-04
Documentation Fix for `Panel and Param`

jrycw:
To maintain consistency with the [provided example](https://panel.holoviz.org/explanation/dependencies/param.html#runtime-type-checking), the interval should span from 5 to 15 instead of 5 to 10.


philippjfr (approved):
Thank you!

PR #6679, opened 2024-04-05
Documentation Update for `pn.bind.md`

jrycw:
1. In [harnessing-throttling-for-performance](https://panel.holoviz.org/tutorials/basic/pn_bind.html#harnessing-throttling-for-performance), there's no object called `power`. I believe you might be referring to `calculate_power_bnd` instead.
2. From the context of [binding-to-bound-functions](https://panel.holoviz.org/tutorials/basic/pn_bind.html#binding-to-bound-functions), I believe the `calculate_power` function should be named `power_generation` instead.


philippjfr (approved):
Many thanks!

Issue #3670, opened 2022-07-04
`selected_dataframe` is linked to the `current_view`


maximlt:
The `selected_dataframe` property is computed based on the `current_view`, which can return wrong data or raise an error if some filters are applied. I think the `selected_dataframe` should not be linked to the `current_view`, it seems useful in a large table to be able to filter (and sort) and select rows at the same time.

MarcSkovMadsen:
To understand what happens you can play with the code below.

```python
import panel as pn
import pandas as pd
import random
from datetime import datetime, timedelta

pn.extension("tabulator", template="fast")

df = pd.DataFrame({
    "x": [1,3,2],
    "y": ["a", "b", "c"]
})

table = pn.widgets.Tabulator(df, sizing_mode="stretch_both", selectable='checkbox', height=300)
update = pn.widgets.Button(name="Update")

@pn.depends(update)
def result(event):
    iloc = df.iloc[table.selection,:]
    loc = df.loc[table.selection,:]

    return pn.Row(
        pn.Column("df", df), 
        pn.Column("table.current_view", table.current_view),
        pn.Column("table.selection", table.selection),
        pn.Column("selected_dataframe", table.selected_dataframe), 
        pn.Column("df.iloc", iloc), 
        pn.Column("df.loc", loc),
        pn.Column("table.current_view.iloc", table.current_view.iloc[table.selection,:]), 
        pn.Column("table.current_view.loc", table.current_view.loc[table.selection,:]),
    )

pn.Column(update, table, result).servable()
```

As long as the user has not sorted Tabulator things works.

![image](https://github.com/holoviz/panel/assets/42288570/7da8d8e6-1a69-4545-8a69-4c4234722eb0)

But if we sort (on column x) things gets out of sync.

![image](https://github.com/holoviz/panel/assets/42288570/1dad535e-c2e0-4384-87e9-ecb70321247c)

I think was it not clearly documented is that `selection` returns the selected indicies from the `index`.

![image](https://github.com/holoviz/panel/assets/42288570/4341253d-aebc-407f-a008-da91a98cc369)

At least for use case using `df.loc[table.selection]` instead of `table.selected_dataframe` will work.

Cyb3r-Monk:
[...]

PR #6685, opened 2024-04-07
selection indices seem to be on _processed and not current view

dwr-psandhu:
The selecton indices seem to return the correct value after filtering and sorting on the _processed. I don't have indepth  knowledge of this code but making this change seems to return the correct rows.

My version of panel is 1.3.8, so will need @philippjfr and @Hoxbro to take a look. 

Fixes https://github.com/holoviz/panel/issues/3670

### DIFF
--- a/panel/tests/ui/widgets/test_tabulator.py
+++ b/panel/tests/ui/widgets/test_tabulator.py
@@ -2948,7 +2948,6 @@ def test_tabulator_click_event_selection_integrations(page, sorter, python_filte
     assert widget.selected_dataframe.equals(expected_selected)
 
 
-@pytest.mark.xfail(reason='See https://github.com/holoviz/panel/issues/3664')
 def test_tabulator_selection_sorters_on_init(page, df_mixed):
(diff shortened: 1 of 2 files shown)


maximlt:
Thanks for opening this PR! However, I suspect it's a change in behavior that is better to combine with a bunch of other fixes to Tabulator's selection and that should be released in a minor release (1.5 hopefully).

philippjfr (approved):
Apologies this has sat around for so long. I do believe this change itself is correct but will have to be accompanied by further changes. I will make those fixes in followup PRs, so will merge now.

dwr-psandhu:
Thanks @philippjfr I am glad you are working on this. This change as you said doesn’t work on its own. In particular, filtering and then sorting causes further issues in case you are thinking of tests

dwr-psandhu:
[heart]         Sandhu, ***@***.*** reacted to your message:
________________________________

philippjfr:
Fix is here: https://github.com/holoviz/panel/pull/7058

PR #6710, opened 2024-04-11
Remove duplicate code in `Structure with a DataStore`

jrycw:
It seems we have duplicate code [here](https://panel.holoviz.org/tutorials/intermediate/structure_data_store.html#continuing-with-views).


philippjfr (approved):
Good catch, thanks!

PR #6721, opened 2024-04-13
Remove Perspective toggle_config

### LINKED ISSUE #6720: Perspective toggle_config not working
I've been used to using Perspectives `toggle_config` parameter. After upgrade to 1.4.0 it no longer works on instantiation.

This can also be seen in the reference guide

image

After some inspection of the code base I can see a new `settings` parameter was introduced up to 1.4.0 and this is shadowing the `toggle_config` parameter.

Both parameters are mentioned (in different ways) in the reference documentation. But they do the same thing.

image

Only the `settings` parameter should be available. The `toggle_config` parameter needs to be removed.

MarcSkovMadsen:
Closes #6720


panel/models/perspective.ts
```diff
@@ -170,10 +167,12 @@ export class PerspectiveView extends HTMLBoxView {
     })
     container.innerHTML = "<perspective-viewer style='height:100%; width:100%;'></perspective-viewer>"
     this.perspective_element = container.children[0]
-    this.perspective_element.resetThemes([...Object.values(THEMES)]).catch(() => {})
-    if (this.model.toggle_config) {
-      this.perspective_element.toggleConfig()
-    }
+
+    const themesArray = Object.values(THEMES);
```
MarcSkovMadsen, line 171: Perspective uses the first theme in the array as the default theme. If we don't put the `model.theme` first in the list we will see a *glitch* when it later changes to the `model.theme`.

panel/pane/perspective.py
```diff
@@ -382,10 +379,6 @@ def _get_properties(self, doc, source=None):
         if 'theme' in props and 'material' in props['theme']:
             props['theme'] = props['theme'].replace('material', 'pro')
         del props['object']
-        if props.get('toggle_config'):
```
MarcSkovMadsen, line 385: I'm in double if this should be replaces with 'settings' check instead of removed?

philippjfr: I don't think this is needed anymore. I think it's a leftover from before 1.0 when explicit sizing was required.

philippjfr (approved):

philippjfr:
Damn, shouldn't have merged, this breaks Perspective. pre-commit shouldn't bypass the tests.

panel/pane/perspective.py
philippjfr: Turns out this was needed 😄

PR #6727, opened 2024-04-14
Add server video stream tutorial

### LINKED ISSUE #6724: Set up Manual Threading how-to guide is highly problematic
I've been looking at the [Set up Manual Threading](https://panel.holoviz.org/how_to/concurrency/manual_threading.html) multiple times and I think it will cause more harm the good.

- It lacks the `import panel as pn` text.
- It does not explain how to stop the thread when the session is ended. Thus more and more threads will be created.
- It does not explain how to stop the thread when the app is autoreloaded.
- Even though there is some attempt at explaining why use of `c = threading.Condition()` is nescessary I'm not sure the explanation is True. My understanding the Condition enables another thread to wait until a condition. And this is not what we do here. See https://docs.python.org/3/library/threading.html#condition-objects.
- I lack some comment about safety of updating Panel components from separate threads. Is that safe? Why?

Maybe use of session life cycle hooks are solution? https://panel.holoviz.org/how_to/callbacks/session.html.

---

Furthermore the guide only explains how to use manual thread in main `app.py` file that is not shared between sessions. I would also like to see an explanation of using a single shared thread in external module. As in https://github.com/holoviz/panel/issues/6723. I think here concerns about performance comes in. Should this thread be pushing updates to all sessions that `.depends` on the threads updates? Or should each individual session be running a periodic callback to check for updates? What is most safe and performant?

---

[x] Yes. I'm willing to contribute a PR updating the guide. But I need some guidance and more understanding to be able to do it.

MarcSkovMadsen:
Adds a `ServerVideoStream` tutorial motivated by questions on discourse. For example in https://discourse.holoviz.org/t/best-practice-for-displaying-high-resolution-camera-images-captured-on-server/4285 and  https://discourse.holoviz.org/t/streaming-local-video/6929.

As this tutorial uses *manual threading* I would also like to fix #6724.

### DIFF
--- a/doc/how_to/concurrency/manual_threading.md
+++ b/doc/how_to/concurrency/manual_threading.md
@@ -5,6 +5,9 @@ Enabling threading in Panel like we saw in the [automatic threading guide](threa
 ```python
 import time
 import threading
+import panel as pn
+
+pn.extension()
 
 c = threading.Condition()
 
--- a/doc/how_to/concurrency/manual_threading.md
+++ b/doc/how_to/concurrency/manual_threading.md
@@ -1,47 +1,130 @@
-# Set up Manual Threading
+# Set Up Manual Threading
 
-Enabling threading in Panel like we saw in the [automatic threading guide](threading) is a simple way to achieve concurrency, however sometimes we need more control. Below we will demonstrate an example of a `Thread` which we start in the background to process items we put in a queue for processing. We simulate the processing with a `time.sleep` but it could be any long-running computation. The `threading.Condition` allows us to manipulate the global shared `queue`.
+Enabling threading in Panel, as demonstrated in the [automatic threading guide](threading.md), provides a simple method to achieve concurrency. However, there are situations where greater control is necessary.
+
+Below, we will demonstrate how to safely implement threads either per session or globally across multiple sessions.
+
+## Session Thread
+
+This section illustrates how to use a `Thread` to process tasks within a queue, with one thread initiated per session to handle tasks individually per session.
+
+We simulate task processing using `time.sleep`, but this could represent any long-running computation.
 
 ```python
-import time
+import datetime
 import threading
+import time
+
+from typing import Callable
+
+import param
+
 import panel as pn
 
 pn.extension()
 
-c = threading.Condition()
+class SessionTaskRunner(pn.viewable.Viewer):
+    value = param.Parameter(
+        doc="The last result or exception", label="Last Result", constant=True
+    )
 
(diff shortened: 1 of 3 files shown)


MarcSkovMadsen:
This PR can be reviewed. I can work for users.

But it would be much stronger if https://github.com/holoviz/panel/issues/6724 was first addressed and applied to this PR.

philippjfr (approved):

philippjfr:
>this PR updated accordingly.

I'm not sure what you mean by that, are you saying that this guide currently does not use best practices?

MarcSkovMadsen:
> > this PR updated accordingly.
> 
> I'm not sure what you mean by that, are you saying that this guide currently does not use best practices?

No i'm saying I don't know what best practice is. It would be great to get that established in the how-to guide and then apply here.

philippjfr:
Got you, thanks!

MarcSkovMadsen:
> Got you, thanks!

This use case is not described in the how-to guide. It explains how to setup a worker thread for each session in a served file.

This use case sets up a shared worker thread in a separate module. All sessions listen/ subscribe/ take the same frame. Currently shared via a parameter on a Parameterized class.

For example right now each time the app is autoreloaded a new worker thread is started and the old not stopped. Your comment in other issue about using `on_session_destroyed` does not apply here as that only works for sessions.

philippjfr:
>For example right now each time the app is autoreloaded a new worker thread is started and the old not stopped. Your comment in other issue about using on_session_destroyed does not apply here as that only works for sessions.

I don't follow, autoreload creates new sessions.

philippjfr:
Ah sorry, I guess I'm following you. You're saying this guide demonstrates the shared worker thread and you want each session to subscribe and unsubscribe. That still should still be done via `on_session_destroyed` handlers though, so I'm not entirely following this point "on_session_destroyed does not apply here as that only works for sessions".

MarcSkovMadsen:
But i dont want that friction. I just want to panel serve as normal such that the component can be importen and used as other component even though its shared.

philippjfr:
You can use the cache to to share the thread instance. I appreciate the pursuit of "reducing friction" but you can't reduce all friction for every conceivable complex use case.

philippjfr:
>The session is not the problem. The problem is that the external module is reimported on autoreload - thus starting another thread. 

Got it, that makes sense. I guess another reason to use the cache.

MarcSkovMadsen:
The cache is a good idea. Did not Think about that.

philippjfr (approved):

PR #6734, opened 2024-04-15
Do not include placeholder in serialize and properly replace placeholder for streams

ahuang11:
Before, the placeholder disappeared, but no message.

https://github.com/holoviz/panel/assets/15331990/fae5daff-a0f6-4004-b198-206d42cbbabc

After fixes:

https://github.com/holoviz/panel/assets/15331990/aee5935b-07f5-4af7-9016-5ce97e086424

### DIFF
--- a/panel/chat/feed.py
+++ b/panel/chat/feed.py
@@ -693,7 +693,7 @@ class ChatFeed(ListPanel):
             return []
         messages = self._chat_log.objects
         undone_entries = messages[-count:]
-        self._chat_log.objects[:] = messages[:-count]
+        self._chat_log.objects = messages[:-count]
         return undone_entries
 
     def clear(self) -> List[Any]:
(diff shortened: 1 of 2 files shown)


panel/chat/feed.py
```diff
@@ -699,7 +693,7 @@ def undo(self, count: int = 1) -> List[Any]:
             return []
         messages = self._chat_log.objects
         undone_entries = messages[-count:]
-        self[:] = messages[:-count]
+        self._chat_log.objects[:] = messages[:-count]
```
philippjfr, line 696: This will not trigger a render, is that intentional?

ahuang11: Thanks for catching, no.

ahuang11, line 696:
```suggestion
        self._chat_log.objects = messages[:-count]
```

Issue #6737, opened 2024-04-16
pn.pane.HTML fails to display bokeh plot


bcm0:
#### Complete, minimal, self-contained example code that reproduces the issue
```python
import panel as pn
pn.extension()
import numpy as np
import pandas as pd
import hvplot.pandas

np.random.seed(1)
idx = pd.date_range('1/1/2000', periods=1000)
df = pd.DataFrame(np.random.randn(1000, 4), index=idx, columns=list('ABCD')).cumsum()
plot = df.hvplot()

hvplot.save(plot, "plot.html")

with open("plot.html") as file:
    html = file.read()
pn.pane.HTML(html) # fails
#from IPython.display import HTML
#display(HTML(html)) # works
```

#### Stack traceback and/or browser JavaScript console output
```
Uncaught TypeError: document.getElementById(...) is null
```
This bug was confirmed on discord

MarcSkovMadsen:
Hi @bcm0 

The HTML pane is meant to display basic *html* not a full html document importing javascript etc.

To display a full *html* document you should wrap the html in an iframe.

```python
import html

import hvplot.pandas
import numpy as np
import pandas as pd

import panel as pn

pn.extension()

np.random.seed(1)
idx = pd.date_range("1/1/2000", periods=1000)
df = pd.DataFrame(np.random.randn(1000, 4), index=idx, columns=list("ABCD")).cumsum()
plot = df.hvplot()

hvplot.save(plot, "plot.html")

with open("plot.html") as file:
    html_content = file.read()

escaped_html = html.escape(html_content)
iframe_html = f'<iframe srcdoc="{escaped_html}" style="height:100%;width:100%" frameBorder="0"></iframe>'

pn.pane.HTML(iframe_html, height=500, sizing_mode="stretch_width").servable()
```

![image](https://github.com/holoviz/panel/assets/42288570/1958f177-8357-4586-bdc6-13b88ebfe51f)

You will have to experiment a bit to get the height, width and alignment working for your use case.

I will close the issue. Feel free to reopen if this does not work for you or if you have a more specific request for improving Panel.

PR #6740, opened 2024-04-17
Document how display full html document in iframe

MarcSkovMadsen:
Closing #6737 by adding an example of displaying a full html document in an iframe using the `HTML` pane.

### DIFF
--- a/examples/reference/panes/HTML.ipynb
+++ b/examples/reference/panes/HTML.ipynb
@@ -141,7 +141,7 @@
     "hvplot.save(plot, plot_file)\n",
     "plot_file.seek(0)  # Move to the beginning of the StringIO object\n",
     "\n",
-    "# Read the HTML content\n",
+    "# Read the HTML content and escape it\n",
     "html_content = plot_file.read()\n",
     "escaped_html = html.escape(html_content)\n",
     "\n",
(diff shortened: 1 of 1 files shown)


MarcSkovMadsen:
The tests fail because panel cannot build the panel .js. I don't believe its related.

examples/reference/panes/HTML.ipynb
```diff
@@ -97,6 +97,65 @@
    "source": [
     "html_pane.styles = dict(html_pane.styles, border='2px solid red')"
    ]
+  },
```
MarcSkovMadsen, line 100: An alternative to documenting this would be to add a `Iframe` pane. Believe I asked for this many years ago but was rejected. I see it needed from time to time and it takes time to construct the code. Its friction.

philippjfr (approved):

examples/reference/panes/HTML.ipynb
philippjfr: I don't remember rejecting the idea.

MarcSkovMadsen:
Would you like to merge too?

philippjfr:
> Would you like to merge too?

Yes, we can consider whether to add the `IFrame` pane for 1.5.

Issue #6744, opened 2024-04-17
Cannot sync DataFrame parameter


MarcSkovMadsen:
I'm on Panel==1.4.1 and would like to `.sync` a dataframe parameter. But I raises an exception

```python
import panel as pn
import param
import pandas as pd

pn.extension("tabulator")

class TestClass(param.Parameterized):
    value = param.DataFrame()

    def __init__(self, **params):
        super().__init__(**params)

        if pn.state.location:
            pn.state.location.sync(self, ["value"])

test = TestClass(value=pd.DataFrame({"x": [1]}))

def _update_value(event):
    value = test.value
    value.iloc[0,0]+=1
    test.param.trigger("value")

submit = pn.widgets.Button(name="Submit", on_click=_update_value)
pn.Column(test.param.value, submit).servable()
```

- `panel serve app.py`
- Click the submit button
- See that the url contains the url parameters: `..?value=%5B%7B"x"%3A+2%7D%5D`
- Reload the page 
- See that it raises exception

![image](https://github.com/holoviz/panel/assets/42288570/c1f2ab36-5aff-4168-9072-9275112f67c8)

```bash
ValueError: The truth value of a DataFrame is ambiguous. Use a.empty, a.bool(), a.item(), a.any() or a.all().

Traceback (most recent call last):
  File "/home/jovyan/repos/mt-quant-frontend/.venv/lib/python3.11/site-packages/bokeh/application/handlers/code_runner.py", line 229, in run
    exec(self._code, module.__dict__)
  File "/home/jovyan/repos/mt-quant-frontend/script.py", line 16, in <module>
    test = TestClass(value=pd.DataFrame({"x": [1]}))
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/jovyan/repos/mt-quant-frontend/script.py", line 14, in __init__
    pn.state.location.sync(self, ["value"])
  File "/home/jovyan/repos/mt-quant-frontend/.venv/lib/python3.11/site-packages/panel/io/location.py", line 239, in sync
    self._update_synced()
  File "/home/jovyan/repos/mt-quant-frontend/.venv/lib/python3.11/site-packages/panel/io/location.py", line 170, in _update_synced
    if not equal:
  File "/home/jovyan/repos/mt-quant-frontend/.venv/lib/python3.11/site-packages/pandas/core/generic.py", line 1527, in __nonzero__
    raise ValueError(
[...]

PR #6745, opened 2024-04-17
Enable syncing dataframe parameters

MarcSkovMadsen:
Closing #6744

### DIFF
--- a/panel/io/location.py
+++ b/panel/io/location.py
@@ -24,6 +24,24 @@ if TYPE_CHECKING:
     from bokeh.server.contexts import BokehSessionContext
     from pyviz_comms import Comm
 
+def _default_is_equal(value, other):
+    return value==other
+
+def _dataframe_is_equal(value, other):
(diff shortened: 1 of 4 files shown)


panel/io/location.py
```diff
@@ -164,9 +182,10 @@ def _update_synced(self, event: param.parameterized.Event = None) -> None:
                 except Exception:
                     pass
                 try:
-                    equal = v == getattr(p, pname)
+                    equal = _is_equal(v, getattr(p, pname), type(p.param[pname]))
```
philippjfr, line 185: This probably should just be something like:

```
v is (new:= getattr(p, pname)) or _generate_hash(v) == _generate_hash(new)
```

where `_generate_hash` is taken from `panel.io.cache`. Or rather we should add a `is_equal` that does that to `pn.io.cache` and then import it here.

philippjfr (approved):

Issue #5096, opened 2023-06-12
Broken click_data event on plotly data point since panel 1.x.x


afrimberger:
#### ALL software version info
Chrome 113
Python 3.11.4

```
panel==1.1.0
bokeh==3.1.1
plotly==5.14.1 (same issue with plotly==5.15.0)
```

#### Description of expected behavior and the observed behavior

**Observed behavior**

<img width="1304" alt="image" src="https://github.com/holoviz/panel/assets/1241470/5fc0486e-b007-4546-8126-f22e704a80f8">

Clicking on one of the data points either doesn't trigger the Python `click_data` callback handler or it triggers the callback handler with wrong customdata (e.g. url).

For example, I clicked on every data point, but the Python callback was only fired for the first one:

<img width="802" alt="image" src="https://github.com/holoviz/panel/assets/1241470/2656d72d-269b-47b3-b773-d6f8efa0e41b">

I'm experiencing this issue since panel 1.x.x

**Expected behavior**

Clicking on a data point fires the Python `click_data` event handler on every click with the associated customdata.

#### Complete, minimal, self-contained example code that reproduces the issue

```python
import pandas as pd
import panel as pn

import plotly.express as px

pn.extension("plotly")
pn.extension(sizing_mode="stretch_width")

bootstrap = pn.template.BootstrapTemplate()


def job_panel():
    data = {
        "job_name": {
            "1507302": "job1",
            "1507303": "job1",
            "1508602": "job1",
            "1443920": "job1",
            "1442532": "job1",
            "1429689": "job1",
            "1439023": "job1",
            "1431578": "job1",
            "1426812": "job1",
            "1501430": "job1",
            "1426410": "job1",
            "1425319": "job1",
            "1509387": "job1",
            "1431356": "job1",
            "1426386": "job1",
        },
        "duration": {
            "1507302": 120.0424790333,
            "1507303": 100.0424790333,
            "1508602": 76.2438600667,
            "1443920": 11.3378994167,
            "1442532": 7.7382620833,
            "1429689": 5.7965679,
            "1439023": 4.68264615,
            "1431578": 3.7771960167,
            "1426812": 3.7264409333,
            "1501430": 3.0562511,
            "1426410": 2.7512232,
            "1425319": 2.71998275,
            "1509387": 2.71287155,
            "1431356": 2.6406619333,
            "1426386": 0.1588929833,
        },
        "job_web_url": {
            "1507302": "https://example.com/1507302",
[...]

PR #6751, opened 2024-04-18
Enhance Plotly documentation

MarcSkovMadsen:
The Plotly reference guide does not explain the Plotly pane to a great extent. I believe more is needed. This PR adresses that.

In most of the sections I try to start from a working example. I've always found it hard using the reference guides across the HoloViz ecosystem when I have to compose a working example from something at the top of a notebook, something from the middle, something from the next cell and then finally what I have further down. So if you don't like me to write sections starting from a working example its time to tell me.

I would have liked to also describe how to use the callbacks. But it seems the `click_data` is not working. So it does not make sense to do now. See https://github.com/holoviz/panel/issues/5096.

### DIFF
--- a/examples/reference/panes/Plotly.ipynb
+++ b/examples/reference/panes/Plotly.ipynb
@@ -1,38 +1,51 @@
 {
  "cells": [
   {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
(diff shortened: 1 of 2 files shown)


philippjfr:
Want to follow up with updating the callbacks docs now that click data is working?

MarcSkovMadsen:
> Want to follow up with updating the callbacks docs now that click data is working?

Some day in near future after work. If you will be releasing 1.4.2 before that feel free to merge this one. Can be merged stand alone.

Issue #6743, opened 2024-04-17
Issue with updating AutoCompleteInput Options


krishanunayak:
#### Complete, minimal, self-contained example code that reproduces the issue

```python
import panel as pn
import param


class AutoComp(pn.viewable.Viewer):
    select = param.Selector(objects=["Fruits", "Animals"], default="Fruits")
    autocomp = param.Selector()

    @param.depends("select")
    def get_autocomp_options(self):
        return {
            "Fruits": ["Apple", "Banana", "Mango"],
            "Animals": ["Monkey", "Dog", "Elephant"],
        }.get(self.select)

    def __panel__(self):
        return pn.Column(
            pn.widgets.Select.from_param(self.param.select),
            pn.widgets.AutocompleteInput.from_param(
                self.param.autocomp, options=self.get_autocomp_options
            ),
        )


AutoComp().servable()

```

#### Stack traceback and/or browser JavaScript console output
TypeError: 'method' object is not iterable

Traceback (most recent call last):
  File "C:\Users\Krishanu.Nayak\OneDrive - Rio Tinto\Desktop\Predicta\.predicta-env\Lib\site-packages\panel\io\handlers.py", line 384, in run
    exec(self._code, module.__dict__)
  File "C:\Users\Krishanu.Nayak\OneDrive - Rio Tinto\Desktop\Predicta\payload_monitor\src\emulate_autocomp.py", line 25, in <module>
    AutoComp().servable()
  File "C:\Users\Krishanu.Nayak\OneDrive - Rio Tinto\Desktop\Predicta\.predicta-env\Lib\site-packages\panel\viewable.py", line 1054, in servable
    return self._create_view().servable(title, location, area, target)
           ^^^^^^^^^^^^^^^^^^^
  File "C:\Users\Krishanu.Nayak\OneDrive - Rio Tinto\Desktop\Predicta\.predicta-env\Lib\site-packages\panel\viewable.py", line 1046, in _create_view
    view = panel(self.__panel__())
                 ^^^^^^^^^^^^^^^^
  File "C:\Users\Krishanu.Nayak\OneDrive - Rio Tinto\Desktop\Predicta\payload_monitor\src\emulate_autocomp.py", line 19, in __panel__
    pn.widgets.AutocompleteInput.from_param(
  File "C:\Users\Krishanu.Nayak\OneDrive - Rio Tinto\Desktop\Predicta\.predicta-env\Lib\site-packages\panel\widgets\base.py", line 93, in from_param
[...]

PR #6762, opened 2024-04-19
Ensure reference is resolved when passing options to from_param

philippjfr:
Fixes https://github.com/holoviz/panel/issues/6743

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -113,7 +113,7 @@ except Exception:
 
 install_requires = [
     'bokeh >=3.4.0,<3.5.0',
-    'param >=2.0.0,<3.0',
+    'param >=2.1.0,<3.0',
     'pyviz_comms >=2.0.0',
     'xyzservices >=2021.09.1', # Bokeh dependency, but pyodide 23.0.0 does not always pick it up
     'markdown',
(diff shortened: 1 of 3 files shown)


setup.py
```diff
@@ -113,7 +113,7 @@ def run(self):
 
 install_requires = [
     'bokeh >=3.4.0,<3.5.0',
-    'param >=2.0.0,<3.0',
+    'param >=2.1.0,<3.0',
```
Hoxbro, line 116: Are you sure that we want to bump versions in a micro release?

philippjfr: I would argue bumping versions to resolve bugs is appropriate in a patch release, but I'm open to alternative opinions.

Hoxbro: My feeling is that if it were a bug in a 3rd party library like Pandas, we would not bump the version. We would likely do a version check, and accept the shortcomings if you are not using a version supporting the bug.

philippjfr: Depends on the severity of the bug I suppose. If it was just this bug I'd definitely postpone the bump but there's a variety of other bugs that are more far reaching. Really should have bumped in 1.4.0 already.

PR #6768, opened 2024-04-20
Speed up `panel bundle --all`

Hoxbro:
I'm just adding a threadpool executor to download all the URLs at the end. Goes from 16s to 5s.

image

I also combined the download command into a single function, which I found cleaner.

I also added a cache for the data, making it take around 2 seconds. I can remove it again on request.


Hoxbro:
I can see that we bundle (large) map files for some packages. Are those needed?

image

philippjfr:
> I can see that we bundle (large) map files for some packages. Are those needed when packaging?

Debatable, I explicitly included them at some point. The reason being that if users inspect the JS console they will otherwise see errors trying to fetch the map files.

philippjfr (approved):

PR #6796, opened 2024-04-28
Add `allow_refs` example to tutorials

MarcSkovMadsen:
Based on the question in https://discourse.holoviz.org/t/dynamic-update-of-one-class-based-on-the-other/7179, I realized that `allow_refs` was not properly introduced in the intermediate tutorials.

This PR introduces `allow_refs`.

ps. I don't think its explained properly in the Param documentation either.

### DIFF
--- a/doc/tutorials/intermediate/reusable_components.md
+++ b/doc/tutorials/intermediate/reusable_components.md
@@ -194,6 +194,55 @@ DataExplorer(data=df).servable()
 
 :::
 
+## Allow References
+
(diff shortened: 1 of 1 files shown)


philippjfr:
>ps. I don't think its explained properly in the Param documentation either.

Please open an issue if you have suggestions for improving the Param docs:

https://param.holoviz.org/user_guide/References.html#parameter-references

philippjfr (approved):

Issue #6684, opened 2024-04-06
`oauth_expiry` cookie provides incorrect information when using `refresh_token` 


nenb:
#### ALL software version info

```
Python 3.9.18
[GCC 12.2.0] on linux
>>> import panel
>>> panel.__version__
'1.3.8'
```

Probably not specific to `panel` version though.

#### Description of expected behavior and the observed behavior

I am using the `generic` OAuth handler. I am also using the [refresh token](https://panel.holoviz.org/how_to/authentication/access_tokens.html#refreshing-access-tokens) functionality.

The issue I face is that after the `access_token` has been refreshed, the `oauth_expiry` cookie appears to provide incorrect information (it still provides information relevant to the original `access_token`).

It's quite possible that I am doing something incorrect, and this is not in fact a bug. I had a quick look at the code to try and understand what was going on. 

It appears that the `oauth_expiry` cookie is set in a [single location](https://github.com/holoviz/panel/blob/5f77d4f85df7c30de4e784a11b449f7f41c06e07/panel/auth.py#L427) inside the `_on_auth` function. It also appears that this function is [not entered](https://github.com/holoviz/panel/blob/5f77d4f85df7c30de4e784a11b449f7f41c06e07/panel/auth.py#L243-L248) when handling a `refresh_token`. This suggests to me that this might actually be a bug ie the cookie is not updated. If I have misunderstood, I will try and share a MRE that should hopefully resolve the issue one way or another.

#### Complete, minimal, self-contained example code that reproduces the issue

Not included.

- [x] I may be interested in making a pull request to address this -> perhaps as simple as removing the early return?

PR #6802, opened 2024-04-30
Ensure token refresh is always scheduled

philippjfr:
There appears to have been an oversight in setting up token refreshes, in particular how/when refreshes are scheduled. In particular the `_schedule_refresh` function was only ever called when a user was either:

1. Visiting the application for the first time and going through the whole auth flow
2. Re-visiting the application for the first time after the tokens expired once
3. Returning to the application after both the access and refresh tokens expired

In the case where a user was revisiting the application after their token had already been refreshed once we did not schedule another refresh. If you revisited after a long time when your token was already expired you'd be forced through the auth flow again and everything would be fine too but if you got unlucky and visited while a refreshed token was stilled just barely valid you could end up with the token expiring without any refresh having been scheduled. Here we now ensure that we ALWAYS schedule the tokens to be refreshed.

Additionally we also now update the cookies when a user revisits the application ensuring that the access_token, oauth_expiry and refresh_token cookies reflect the latest refresh values. However if a token is refreshed while a session is running these cookies may still be out-of-date until the next time the user visits the application:

Fixes https://github.com/holoviz/panel/issues/6684

### DIFF
--- a/panel/auth.py
+++ b/panel/auth.py
@@ -1016,12 +1016,12 @@ class OAuthProvider(BasicAuthProvider):
                 refresh_cookie = handler.get_secure_cookie('refresh_token', max_age_days=config.oauth_expiry)
                 if refresh_cookie:
                     refresh_token = state._decrypt_cookie(refresh_cookie)
-                    self._schedule_refresh(access_json['exp'], user, refresh_token, handler.application, handler.request)
                 else:
                     refresh_token = None
 
-            if expiry > now_ts:
+            if expiry > now_ts and refresh_token:
                 log.debug("Fully authenticated and access_token still valid.")
+                self._schedule_refresh(expiry, user, refresh_token, handler.application, handler.request)
                 return user
 
             if refresh_token:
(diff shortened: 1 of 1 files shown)


panel/auth.py
```diff
@@ -1076,13 +1089,14 @@ def _schedule_refresh(self, expiry_ts, user, refresh_token, application, request
         if not state._active_users.get(user):
             return
         now_ts = dt.datetime.now(dt.timezone.utc).timestamp()
-        expiry_seconds = expiry_ts - now_ts - 10
-        log.debug("%s scheduling token refresh in %d seconds", type(self).__name__, expiry_seconds)
+        expiry_seconds = expiry_ts - now_ts - 60
         expiry_date = dt.datetime.now() + dt.timedelta(seconds=expiry_seconds) # schedule_task is in local TZ
         refresh_cb = partial(self._scheduled_refresh, user, refresh_token, application, request)
         if expiry_seconds <= 0:
+            log.debug("%s token expired unexpectedly, refreshing immediately.", type(self).__name__, expiry_seconds)
```
maximlt, line 1096:
```suggestion
            log.debug("%s token expired unexpectedly, refreshing immediately.", type(self).__name__)
```

PR #6805, opened 2024-05-01
Allow callbacks after append and stream

ahuang11:
Motivated by https://github.com/holoviz/lumen/pull/544

I'd like to trigger a callback after every new message, but it's hard to do right now because of the various ways to append a message.

Initially I wanted to just super `append` but it doesn't cover the case of streaming (e.g. the contents were empty for the callback trigger). Here, I wait for all the streaming to be done before triggering append callback.


philippjfr:
Hmm, should we not simply be waiting to trigger the regular callback until it's done streaming?

ahuang11:
>  to trigger the regular callback until it's done streaming?

If I understand correctly, the regular callback triggers the streaming, i.e.
1. user sends a message
2. "regular callback" runs the callback
3. append_callback runs after the regular callback

PR #6814, opened 2024-05-08
Raise on ChatReactionIcons instance in ChatFeed.message_params

ahuang11:
```python
import panel as pn

pn.extension()

chat_feed = pn.chat.ChatFeed(reaction_icons=pn.chat.ChatReactionIcons(options={"like": "thumb-up", "dislike": "thumb-down"}))
msg1 = chat_feed.send("Hello")
msg2 = chat_feed.send("Hey")

msg1.reactions = ["like"]

chat_feed
```

No longer linked

image

### DIFF
--- a/panel/chat/feed.py
+++ b/panel/chat/feed.py
@@ -26,6 +26,7 @@ from ..layout import Feed, ListPanel
 from ..layout.card import Card
 from ..layout.spacer import VSpacer
 from ..pane.image import SVG
+from .icon import ChatReactionIcons
 from .message import ChatMessage
 
 if TYPE_CHECKING:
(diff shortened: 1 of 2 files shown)


panel/chat/feed.py
```diff
@@ -350,6 +350,11 @@ def _build_message(
                 f"e.g. {{'object': 'Hello World'}}; got {value!r}"
             )
         message_params = dict(value, renderers=self.renderers, **self.message_params)
+        for param_key, param_value in message_params.items():
```
philippjfr, line 353: I don't understand this. Why are we passing in the same reaction icons instance to all messages in the first place? This approach seems weird.

ahuang11: I suppose we could do:

```
import panel as pn

pn.extension()

chat_feed = pn.chat.ChatFeed(reaction_icons={"like": "thumb-up", "dislike": "thumb-down"})
msg1 = chat_feed.send("Hello")
msg2 = chat_feed.send("Hey")

msg1.reactions = ["like"]

chat_feed
```

Should I disallow passing `ChatReactionIcon` as an instance then?

philippjfr: Yeah was just looking at that. I think we should yes. The problem is that by cloning you break some assumptions, e.g. a user might have added some watchers to the reaction icons and by cloning you silently drop those.

philippjfr (approved):

PR #6815, opened 2024-05-08
Fix dynamic chat reaction icons

ahuang11:
Follow up on https://github.com/holoviz/panel/pull/6807/files, properly implementing the fix
- inserting the reactive expression directly (which indeed means it’ll be nested inside a ReactiveExpr pane which creates a layout)
- handling updates to the reactive_icons correctly, so stylesheets, the link and show_reaction_icons are not reapplied.

### DIFF
--- a/panel/chat/message.py
+++ b/panel/chat/message.py
@@ -246,21 +246,16 @@ class ChatMessage(PaneBase):
             elif state.browser_info and state.browser_info.timezone:
                 tz = ZoneInfo(state.browser_info.timezone)
             params["timestamp"] = datetime.datetime.now(tz=tz)
+
         reaction_icons = params.get("reaction_icons", {"favorite": "heart"})
         if isinstance(reaction_icons, dict):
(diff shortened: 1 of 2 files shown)


panel/chat/message.py
```diff
@@ -630,6 +626,21 @@ def _update_avatar_pane(self, event=None):
             del params['name']
             self._left_col[0].param.update(**params)
 
+    def _render_reaction_icons(self):
+        reaction_icons = self.reaction_icons
```
philippjfr, line 630: Seems like if we don't allow a dict to be passed in dynamically we will have no way to provide reaction_icons to a ChatFeed since we can't pass an instance there.

ahuang11: I've always wondered about this:
https://discourse.holoviz.org/t/is-there-a-way-to-validate-a-parameter-in-param/7196

But I believe we don't need to dynamically pass a dict because ChatFeed creates new messages (if I'm not mistaken)

philippjfr (approved):

PR #6818, opened 2024-05-09
Ensure binary buffers are correctly extracted in CommManager

philippjfr:
Binary buffers in Bokeh are now held in a `Buffer` object which means our old check for `isPlainObject` no longer works and we have to explicitly check for it.

- [ ] Add test


philippjfr:
Fixes https://github.com/holoviz/holoviews/issues/6225

PR #6820, opened 2024-05-10
Allow external library to define additional socket handler

philippjfr:
Allows an external library (like `bokeh-fastapi`) to define a socket handler that allows for immediate dispatch of messages.


MarcSkovMadsen:
Looking forward to review the tests 😄

philippjfr:
Ideally we'd instrument such that all existing server based tests run against both Tornado and Uvicorn eventually.

PR #6833, opened 2024-05-14
Update conf.py with list of Rediraffe 301 redirects.

Coderambling:
Added first batch of redirects. Please check if this is OK? Will a syntax error / typo  in the redirects will cause issues beyond Rediraffe?

'user_guide/APIs.html': 'explanation/api/index.html#apis',
    'user_guide/Pipelines.html': 'how_to/pipeline/index.html',
    'user_guide/Templates.html': 'how_to/templates/index.html',
    '_static/images/sazure_deployment.png': '_static/images/azure_deployment.png',
    'user_guide/Server_Configuration.html': 'how_to/server/index.html',

### DIFF
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -97,6 +97,14 @@ pyodide_url = 'https://holoviz-dev.github.io/panel/pyodide' if is_dev else 'http
 rediraffe_redirects = {
     # Removal of the developer testing page
     'developer_guide/testing': 'developer_guide/index',
+    'user_guide/APIs.html': 'explanation/api/index.html#apis',
+    'user_guide/Pipelines.html': 'how_to/pipeline/index.html',
+    'user_guide/Templates.html': 'how_to/templates/index.html',
+    '_static/images/sazure_deployment.png': '_static/images/azure_deployment.png',
+    'user_guide/Server_Configuration.html': 'how_to/server/index.html',
+
+# Todo
+# https://panel.holoviz.org/user_guide/Customization.html : what page to redirect this to?
 }
 
 nbsite_analytics = {


MarcSkovMadsen:
Thx for the PR @Coderambling 

FYI. This PR will reach the PROD docs at https://panel.holoviz.org the next time a new version of Panel is released.

It will reach the DEV docs at https://holoviz-dev.github.io/panel/ the next time someone or a trigger rebuilds and deployes that site. I've triggered a rebuild and it should show up in ~1 hour.

image

Coderambling:
Does this mean that, if a broken link is introduced in a Holoviz website, it will persist until the next build of the site, which usually coincides with the next version of a Holoviz library?

Not a major issue I guess, but still. Could be mitigated by including a link checking tool in the testing of the dev build of the websites.

Coderambling:
Did an impromptu check of link issues in the dev build.

60 in total, some already known from current version, some new. See 2 examples below.

My question concerns added value (or not) of such a link issue list for dev.

Are some of them (like below) simply artefacts of being in dev, and they will be ok in prod?

Is it useful / worth it to put them in a spreadsheet, that is updated after each dev build?

In theory this could be automated to trigger and run a check after each build, and could easily be extended to a spreadsheet per Holoviz website.

In this way maintainers can have a look after a build, and pick out relevant link issues.

But would this work in practice? Are there better alternatives? 

Any thoughts on this? I am not familiar enough with the site development process to have a good idea of this.

Regrettably, actually fixing the links will mostly remain a manual task, aside from site build issues which can generate a bunch of bad links, but are easy to fix.

The only other way to minimize bad links is if the link is loaded (in a preview) immediately when someone is creating a link in the content, so there is immediate feedback if the load fails.

Some systems have that,  I am guessing Sphinx doesn't.



Examples:

1.

https://www.nginx.com/wp-content/uploads/2014/07/what-is-load-balancing-diagram-NGINX-1024x518.png



in page:

https://holoviz-dev.github.io/panel/how_to/concurrency/load_balancing.html

2.

https://github.com/plotly/dash-ag-grid/blob/dev/docs/demo_stock_portfolio.py

Should be:

https://github.com/plotly/dash-ag-grid/blob/main/docs/demo_stock_portfolio.py

in page:

https://holoviz-dev.github.io/panel/gallery/portfolio_analyzer.html

PR #6836, opened 2024-05-15
Docs: remove Google Analytics

maximlt:
GoatCounter was already in place, and we've seen that the metrics both services provide are close enough.


philippjfr (approved):

philippjfr:
Thank you!

PR #6837, opened 2024-05-15
Add jsTree based FileTreeSelector

philippjfr:
Adapts the [panel-jstree](https://github.com/madeline-scyphers/panel-jstree/tree/main) components to support selecting both local and remote filesystems.

Screenshot 2024-05-15 at 18 57 56

- [ ] Allow navigating up and down a directory tree while respecting a `max_depth` parameter setting
- [ ] Clean up `RemoteFileProvider`
- [ ] Add composite component that adds additional controls for navigating different directories
- [ ] Allow asynchronously fetching directories
- [ ] Display loading icons while fetching directories


madeline-scyphers:
This is great. It was always my hope that panel-jstree would get put into panel someday. I am gonna put some comments on some parts for the things I was trying to work on when I had time. 

The next thing I was trying to do for a side project I help with that was trying to use panel-jstree was create a composite widget, which here https://github.com/madeline-scyphers/panel-jstree/blob/1e364fcf5ff1947912ce8ba2b1f4d9753c4498d8/src/panel_jstree/widgets/jstree.py#L290 
```python
    def _set_data_from_directory(self, *event):
        self._data = [{"id": fullpath(self.directory),
                       "text": Path(self.directory).name,
                       "icon": self._folder_icon,
                       "state": {"opened": True},
                       "children": self._get_children_cb(Path(self.directory).name, self.directory,  depth=1)
                       }]
```
is a FileTree cb triggered on the directory change to swap out the entire data from scratch. I found this worked, but was a bit laggy (lag came from recreate the tree in the browser, not the directory search). Maybe the asynchronous  tree will help, but if not, I was trying to play around with the massload plugin to see if that could also help but I never finished it with trying to finish my master's https://www.jstree.com/api/#/?f=$.jstree.defaults.massload 

One thing I also wanted to mention was that I was trying really hard to make sure that there was a general tree implementation someone could use that would be independent of a FileTree if they just wanted to explore Tree data. It looks like that is still working properly, but I just want to underscore that I think that is really important, and ideally as many features can be generalized to work for a generalized Tree, not just a FileTree.

I added support for a number of jsTree's plugins, but there are a few more that might be cool to do. drag and drop was one I wanted to do next, which allows the user to just rearrange the Tree, though by default it can move a leaf to another node, so maybe not for the FileTree. Sort also is another one that might be nice. There is a search plugin as well.

That is most of most of what I had left that I wanted to do. I think the remote file provider is super cool, as well as the other features you all are adding. I would love to help with the last things to get this in.

panel/models/jstree.py
```diff
@@ -0,0 +1,69 @@
...
+    show_dots = Bool(default=False)
+    show_stripes = Bool(default=True)
+    sort = Bool(default=True)
+    _new_nodes = Nullable(List(Any))
+
+    # Callback properties
+    value = List(Any)
+    nodes = List(Any)
+
+    directory = String()
```
philippjfr, line 69: I don't think this is used.

panel/models/jstree.py
```diff
@@ -0,0 +1,69 @@
...
+    """
+    A Bokeh model that wraps around a jsTree editor and renders it inside
+    a Bokeh plot.
+    """
+
+    __css_raw__ = [
+        f"{config.npm_cdn}/jstree@3.3.16/dist/themes/default/style.min.css",
+    ]
+
+    __resources__ = [
```
philippjfr, line 36: I replaced these icons, let's see if we can get rid of this.

panel/models/jstree.py
```diff
@@ -0,0 +1,69 @@
...
+        f"{config.npm_cdn}/jstree@3.3.16/dist/themes/default/throbber.gif"
+    ]
+
+    @classproperty
+    def __css__(cls):
+        cls.__css_raw__ = cls.__css_raw__[:1]
+        return bundled_files(cls, 'css')
+
+    __javascript_raw__ = [
+        "https://cdnjs.cloudflare.com/ajax/libs/jquery/3.5.1/jquery.min.js", #JS_URLS['jQuery'],
```
philippjfr, line 47: `JS_URLS['jQuery']` currently refers to a slim version of jQuery which isn't sufficient here (it causes errors). We should simply switch the JS_URLS['jQuery'] definition to `f'{CDN_DIST}bundled/jquery/jquery.min.js'`

panel/models/jstree.py
hoxbro: Have removed it for now

hoxbro:
Example of max_depth:

image

``` python
import s3fs
import panel as pn

fs = s3fs.S3FileSystem(anon=True)
pn.widgets.FileTree(directory="s3://datasets.holoviz.org", fs=fs, max_depth=2).servable()

```

philippjfr:
A little `FileTreeSelector` demo:

filetree

panel/models/jstree.ts
```diff
@@ -0,0 +1,311 @@
...
+    this.shadow_el.appendChild(this._container)
+    this._jstree = jQuery(this._container).jstree({
+      allow_reselect: true,
+      checkbox: {
+        three_state: this.model.cascade,
+        cascade: this.model.cascade ? "undetermined" : "down+undetermined",
+        cascade_to_disabled: false,
+        tie_selection: false,
+        whole_node: false,
+      },
```
madeline-scyphers, line 115: I think we need more options for cascade. Especially for generic trees outside of just file trees. I am using my version of jstree right now for something at work to model some genome taxonomy and trying to use the panel-jstree widget I was building until this is ready to be able to cleanly subselect by any level of taxonomy (remove all shown samples with this class, phylum, domain, or species) for example. 

So when deselecting a class, everything below it should deselect, but when selecting a certain domain, it shouldn't select everything below it. That behavior isn't achievable right now with these two cascade options, and I think for generic, non file tree use cases, that would be something wanted.

panel/widgets/file_selector.py
```diff
@@ -280,20 +445,209 @@ def _select(self, event: param.parameterized.Event):
...
+            self._param_change(event)
+
+    def _process_param_change(self, params):
+        if 'directory' in params:
+            params['nodes'] = self._nodes
+            del params['directory']
+        props = super()._process_param_change(params)
+        if 'value' in props:
+            props['checked'] = props.pop('value')
+        return props
```
madeline-scyphers, line 518: I think this section could be moved to the base tree, since it I think should always happen in any concrete implementation, and the generic Tree implementation needs it too.

panel/widgets/tree.py
```diff
@@ -0,0 +1,238 @@
...
+        for subnode in node.get('children', []):
+            nodes += cls._traverse(subnode)
+        return nodes
+
+    def _reindex(self, reset=True):
+        if reset:
+            self._index = {}
+        for node in self._nodes:
+            for subnode in self._traverse(node):
+                self._index[subnode['id']] = subnode
```
madeline-scyphers, line 143: Potential problem with this, jstree doesn't require the initial node json data to include the id (https://www.jstree.com/docs/json/). If you don't include it, then jstree will provide it for you. Obviously the FileTree can assume an id because it is coded that way and it makes sense to make the id the unique file path, but a generic tree using the Tree class may not be constructed with an id and the nodes will only get an id after the first render with the jstree library. 

Since the FileTree uses this method to initialize the _index cache on on init, I wanted to mention that this won't work when work is completed to get the generic Tree class working again.

madeline-scyphers:
So for my job, I am building a dashboard for some genome taxonomy data that is the output of a model. And I am trying to use my version of panel-jstree and eventually this version to filter based on taxonomy rank (domain, phylum, etc.). While doing this, I have a little function that allows me to convert an edgelist dataframe into a format that jstree needs. 

<details>
  <summary>So this edgelist df</summary>

```bash
  source target
0      a    a-1
1      a    a-2
2      b    b-1
3      b    b-2
4      b    b-3
5    b-1  b-1-1
6    b-2  b-2-1
7    b-2  b-2-2
```

</details>

<details>
  <summary>Gets transformed into this node list</summary>

```python
>>>build_tree(edge_df, state={"opened": False, "selected": False})
[{'text': 'a',
  'children': [{'text': 'a-1',
    'children': [],
    'state': {'opened': False, 'selected': False}},
   {'text': 'a-2',
    'children': [],
    'state': {'opened': False, 'selected': False}}],
  'state': {'opened': False, 'selected': False}},
 {'text': 'b',
  'children': [{'text': 'b-1',
    'children': [{'text': 'b-1-1',
      'children': [],
      'state': {'opened': False, 'selected': False}}],
    'state': {'opened': False, 'selected': False}},
   {'text': 'b-2',
    'children': [{'text': 'b-2-1',
      'children': [],
      'state': {'opened': False, 'selected': False}},
     {'text': 'b-2-2',
      'children': [],
      'state': {'opened': False, 'selected': False}}],
    'state': {'opened': False, 'selected': False}},
   {'text': 'b-3',
    'children': [],
    'state': {'opened': False, 'selected': False}}],
  'state': {'opened': False, 'selected': False}}]

```
</details>
  
  
And I was thinking it would be nice to be able to construct the generic Tree class from a edgelist df, maybe as a class_method? `Tree.from_edgelist(df)`

I can put in a PR into this branch to add that. It isn't that complicated of a function, and I tried to do it without something like NetworkX to not add any dependencies. Though I imagine that would speed it up. So there could be an optional dependency on NetworkX to speed it up?

panel/widgets/tree.py
```diff
@@ -0,0 +1,238 @@
...
+    ):
+        if "selected" not in state:
+            state["selected"] = id_ in self.value
+        jsn = dict(id=id_, text=label, children=children or [], state=state, **kwargs)
+        if parent:
+            jsn["parent"] = parent
+        if icon:
+            jsn["icon"] = icon
+        else:
+            jsn["icon"] = "jstree-leaf"
```
madeline-scyphers, line 128: Is it possible to let people specify any icon from [tabler-icons.io](https://tabler-icons.io/) from other places? This was always something I wanted to try get working but never got around to. It would especially be useful in a generic Tree when people are using it for all sorts of stuff. Having DNA icons for taxonomy things for my genome projects I am working on, or using local custom icons for a branching history tree, or any other tree.

panel/models/jstree.ts
madeline-scyphers: I am rethinking what I want in my own situation actually 🤣 but probably the option to choose more cascade choices would still be good.

philippjfr:
Thanks for all your helpful review comments @madeline-scyphers! Will try to get back to this PR soon.

SebastianHambura:
I was looking for option to display the content of Zarr files in a panel widget, and a tree would be perfect for that. Is there any reason why this is not part of the 'official' panel? What is missing for a merge?

philippjfr:
@SebastianHambura  Honestly, this implementation is kind of a mess. I'd try using the `panel-material-ui` [Tree widget](https://panel-material-ui.holoviz.org/reference/menus/Tree.html) and see how that goes.

PR #6842, opened 2024-05-16
Testing docs build

hoxbro:
Reverts holoviz/panel#6833

Docs passes: https://github.com/holoviz/panel/actions/runs/9114403040

### DIFF
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -97,14 +97,6 @@ pyodide_url = 'https://holoviz-dev.github.io/panel/pyodide' if is_dev else 'http
 rediraffe_redirects = {
     # Removal of the developer testing page
     'developer_guide/testing': 'developer_guide/index',
-    'user_guide/APIs.html': 'explanation/api/index.html#apis',
-    'user_guide/Pipelines.html': 'how_to/pipeline/index.html',
-    'user_guide/Templates.html': 'how_to/templates/index.html',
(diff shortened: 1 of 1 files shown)


maximlt:
I think the fix is to simply remove the `.html` suffix in both keys and values. cc @Coderambling

Coderambling:
D*rn I had not realized that.

Is it still OK to do i.e.:

 'user_guide/APIs : 'explanation/api/index.html#apis' ? As that 2nd part (the value) does not end in .html? Or would that also cause issues?

hoxbro:
Have set over another run here: https://github.com/holoviz/panel/actions/runs/9130979122

doc/conf.py
```diff
@@ -97,14 +97,11 @@
 rediraffe_redirects = {
     # Removal of the developer testing page
     'developer_guide/testing': 'developer_guide/index',
-    'user_guide/APIs': 'explanation/api/index.html#apis',
+    'user_guide/APIs': 'explanation/api/index#apis',
     'user_guide/Pipelines': 'how_to/pipeline/index',
     'user_guide/Templates': 'how_to/templates/index',
     '_static/images/sazure_deployment.png': '_static/images/azure_deployment.png',
```
philippjfr, line 103: Pngs can't be redirected it seems

```suggestion
```

Issue #6787, opened 2024-04-24
Pyiodide Conversion: requirements.txt not able to deal with wheel-URLS


flxmr:
#### ALL software version info
`panel@900fb095b4dfe7c3500321ba7fe84f188b08721a`  on Python 3.12.2

#### Description of expected behavior and the observed behavior
I would like to specify a custom url for a wheel. In standard Python I can just put the url in the `requirements.txt`. As panel uses [`packaging.requirements.Requirement`](https://packaging.pypa.io/en/stable/requirements.html#packaging.requirements.Requirement) this doesn't work, because the URL is not a valid requirement. This would not be bad if one could use `package @ wheelurl`. However this fails, because panel passes this as a string to micropip, which doesn't understand this spec and tries to get `http://localhost:8000/mywheel@https://localhost:8000/myhweel.whl`.

According to micropip one should https://micropip.pyodide.org/en/stable/project/api.html#micropip.install only pass package names, http or emfs-urls.

#### Complete, minimal, self-contained example code that reproduces the issue

```
panel convert mypanel.py --compiled --to pyodide-worker --out pyodide-excel --requirements reqs.txt
```
where `reqs.txt` looks like
```
http://localhost:8000/mywheel.whl
```
results in:
```
Failed to convert mypanel.py to pyodide target: Requirements parser raised following error: Expected end or semicolon (after name and no valid version specifier)
    https://localhost:8000/mywheel.whl
```

- [x ] I may be interested in making a pull request to address this. Likely I will try to convert the `Requirements` to package specifications accepted by micropip

flxmr:
also:
https://github.com/holoviz/panel/issues/6170 (fix the requirements-heuristic)
https://github.com/holoviz/panel/issues/5700 (and this bug somewhere)

flxmr:
As pyodide does only provide one set of emscripten-wheels for every version (...) it doesn't make a lot of sense to use formal `Requirements` at all I would say. Also the "requirements"-files as read by pip don't have to have a `Requirement` in every line. This is funny and despite the polish shown by the various "you can run Python in the browser"-projects, this seems very toy-like. (also pip is the only implementation of a proper dependency resolver - and that without an API)

Issue #4164, opened 2022-11-30
Cannot load static assets when panel convert pyiodide


ahuang11:
```python
from PIL import Image

import panel as pn

def expand2square(pil_img, background_color):
    width, height = pil_img.size
    if width == height:
        return pil_img
    elif width > height:
        result = Image.new(pil_img.mode, (width, width), background_color)
        result.paste(pil_img, (0, (width - height) // 2))
        return result
    else:
        result = Image.new(pil_img.mode, (height, height), background_color)
        result.paste(pil_img, ((height - width) // 2, 0))
        return result

img_in = Image.open("nintendo-7794.png")
img_out = expand2square(img_in, background_color="black")

png = pn.pane.PNG(img_out, sizing_mode="scale_both", max_width=100)
png.servable()
```

`panel convert test.py --requirements pillow --to pyodide-worker --out test`

`FileNotFoundError: [Errno 44] No such file or directory: 'nintendo-7794.png'` even though `panel serve test.py` works.

legout:
Same here. Is there already a fix?

PR #6859, opened 2024-05-22
Pyodide Conversion: fix requirements handling

flxmr:
This fixes:  #6787 AND #4164 and adjusts the documentation

Why both? adding custom wheels as requirements requires making those available to the Pyodide environment at runtime. One could place them into the output structure as wheels to be served with the app, but this might clutter it a lot if there are many wheels (and also clash with any bundling system to be implemented). Also semantics with this solution seem clearer: you convert a panel app requiring custom wheels, you get a zip with dependencies and you distribute the app and the zip-file.

As the infrastructure for fixing #4164 was in place with this solution, I then also added a `--resources`-flag to the convert-command to include custom resources (such as template files/data files/...) required by the custom-python-portion of the panel app. Everything is packed into a single zipfile, because that's easiest to use with the `pyscript` `config` (which only allows to uncompress one zip to a specific wildcard-location); also I bumped the pyscript version, because that feature only got added with one of the small releases.

Currently BROKEN: on main, someone updated the version number in `package.json` which is used for getting the `CDN_DIST`-url. I think `PY_VERSION` in `panel/io/convert.py` (which references a python `__version__` should also be updated accordingly and possibly also be acquired from a single-source-of-truth!

Reasons for restructuring:
- installing the dependencies one-by-one with micropip for the pyodide-worker seems clever and gives some responsiveness. However this doesn't use custom packages installed in previous calls for dependency resolution (...) and thus does not work if you want to override specific dependencies with custom-builds!
- I split the code for finding the requirements into a function `collect_python_requirements` (finding the requirements or parsing the requirements-file) and some code rewriting/collecting filesystem-URLs. `script_to_html` now takes a micropip-ready list of package-specifications!

Small things:
- I would put the documentation for standalone-usage of panel in WASM in front of the conversion docs. Currently this is very weird clicking through, showing a conversion tool and then showing how to manually write your code.
- I replaced some very inconsistent usage of single and doublequotes
- as http-patch is in the pyodide-lockfile (from which micropip installs WASM-wheels) and this only ever includes one version (for a specific version of Pyodide), it doesn't make sense to version-lock this. Also removed the `MIN_VERSIONS` for this reason. Once pyodide implements proper requirements handling this should be put into the requirements-file like for every other project.

### DIFF
--- a/panel/_templates/pyodide_worker.js
+++ b/panel/_templates/pyodide_worker.js
@@ -13,29 +13,29 @@ async function startApplication() {
   self.postMessage({type: 'status', msg: 'Loading pyodide'})
   self.pyodide = await loadPyodide();
   self.pyodide.globals.set("sendPatch", sendPatch);
-  console.log("Loaded!");
+  console.log("Loaded pyodide!");
+  const data_archives = [{{ data_archives }}];
+  for (const archive of data_archives) {
+    let zipResponse = await fetch(archive);
+    let zipBinary = await zipResponse.arrayBuffer();
+    self.postMessage({type: 'status', msg: `Unpacking ${archive}`})
+    self.pyodide.unpackArchive(zipBinary, "zip");
+  }
   await self.pyodide.loadPackage("micropip");
-  const env_spec = [{{ env_spec }}]
-  for (const pkg of env_spec) {
-    let pkg_name;
-    if (pkg.endsWith('.whl')) {
-      pkg_name = pkg.split('/').slice(-1)[0].split('-')[0]
-    } else {
-      pkg_name = pkg
-    }
-    self.postMessage({type: 'status', msg: `Installing ${pkg_name}`})
-    try {
-      await self.pyodide.runPythonAsync(`
-        import micropip
-        await micropip.install('${pkg}');
-      `);
-    } catch(e) {
-      console.log(e)
-      self.postMessage({
-	type: 'status',
-	msg: `Error while installing ${pkg_name}`
-      });
-    }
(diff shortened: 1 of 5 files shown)


panel/_templates/pyodide_worker.js
```diff
@@ -13,29 +13,29 @@ async function startApplication() {
...
-      `);
-    } catch(e) {
-      console.log(e)
-      self.postMessage({
-	type: 'status',
-	msg: `Error while installing ${pkg_name}`
-      });
-    }
+  self.postMessage({type: 'status', msg: `Installing packages`});
+  // a finegrained approach installing dependencies one after another with status updates was implemented previously
```
philippjfr, line 26: Unfortunate but better to handle this correctly. Would be great if micropip supported hooks to report progress.

flxmr: ah, that was not commented properly for the review! The problem is that the dependency resolution fails for locally installed wheels (...) - probably this should be fixed in micropip. I will try to do some tests too and poke the micropip-people a bit in the meantime.

philippjfr: Sweet thanks! Am trying to test this PR and then get it merged in time for the Panel 1.5 release.

philippjfr:
I'm really sorry this has sat around for so long. It's a great contribution and should have gone in long ago. I didn't want to burden you with having to resolve the conflicts so I've restarted this effort in https://github.com/holoviz/panel/pull/8181, preserving your commits.

PR #6860, opened 2024-05-22
Update pn_panel.md: replace mention of 'dice' with ' wind turbine' in the tutorial.

Coderambling:
There is a mention of 'dice' in this tutorial, which uses wind turbines as an example. Replaced 'dice' with 'wind turbine'.


ahuang11 (approved):

ahuang11:
Thanks!

PR #6861, opened 2024-05-22
feat: add PAM authenticator

pgierz:
Hi Panel Team,

I built a very simple PAM authenticator, and it seems to work provided you have given the correct environmental variables. Here is an example, assuming you have an app named `app.py`. For demonstration, I just use the md5 sum of the program code:

```console
$ export PANEL_OAUTH_KEY=$(md5 -q app.py)
$ export PANEL_OAUTH_SECRET=$(panel secret)
$ export PANEL_OAUTH_PROVIDER=pam
$ export PANEL_OAUTH_ENCRYPTION=$(panel oauth-secret)
$ export PANEL_COOKIE_SECRET=$(panel oauth-secret)  # Twice, gives unique secret each call
$ panel serve app.py
```

Or in one line:
```
$ PANEL_OAUTH_KEY=$(md5 -q app.py) PANEL_OAUTH_SECRET=$(panel secret) PANEL_OAUTH_PROVIDER=pam PANEL_OAUTH_ENCRYPTION=$(panel oauth-secret) PANEL_COOKIE_SECRET=$(panel oauth-secret) panel serve app.py
```

Maybe someone else finds this useful.


philippjfr (approved):
Thanks, looks good, two main comments:

1. We shouldn't add `pamela` as a core dependency, so the import should be inlined.
2. We'll need some documentation on how to use it.

pgierz:
I moved the the Pamela import to be a optional dependency and changed the inclusion in the pyproject.toml file.

I haven't contributed to the documentation before, is there a best practices guideline for that?

MarcSkovMadsen:
Regarding best practice for documentation. We are guided by the Diataxis framework for technical documentation https://diataxis.fr/. You might see something non-consistent in practice - either for historical reasons or because we don't yet understand well enough to write according to Diataxis.

What I would recommend doing is looking at the documentation for other OAuth Providers and then making a similar contribution.

Coderambling:
Hi @pgierz . Cool feature! OAuth docs are here:

https://panel.holoviz.org/how_to/authentication/providers.html

philippjfr:
Thanks again @pgierz, I finally got around to this and added some docs. Will merge.

philippjfr (approved):

pyproject.toml
```diff
@@ -75,6 +75,7 @@ recommended = [
     'matplotlib',
     'pillow',
     'plotly',
+    'pamela',
```
philippjfr, line 78:
```suggestion
```

PR #6878, opened 2024-05-30
Remove optional dependencies from meta.yaml

hoxbro:
We could see in previous versions of our conda package we didn't package with recommendations as we do in HoloViews. 

image

This also removes the need for the standard optional dependencies group.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -75,8 +75,6 @@ recommended = [
     'matplotlib',
     'pillow',
     'plotly',
-]
-standard = [
     'watchfiles',
 ]
 tests = [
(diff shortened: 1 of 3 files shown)


pyproject.toml
```diff
@@ -75,8 +75,6 @@ recommended = [
     'matplotlib',
     'pillow',
     'plotly',
-]
-standard = [
     'watchfiles',
```
philippjfr, line 78: I do like the idea of watchfiles as a little more as recommended. Either that or the `recommended` group should be trimmed down to basically just watchfiles.

hoxbro: I have reverted those changes.

philippjfr (approved):

Issue #6673, opened 2024-04-04
Tabulator: error when editing a table that contains a column with dtype `datetime64[ns, UTC]`


maximlt:
When a DataFrame contains a column with dtype `datetime64[ns, UTC]`, editing a value in the DataFrame through the Tabulator widget raises an Exception.

```python
import pandas as pd
import numpy as np
import panel as pn
pn.extension('tabulator')

data = {'datetime_utc': pd.date_range(start='2024-01-01', end='2024-01-05', freq='D')}
df = pd.DataFrame(data)
# Convert the dtype of the datetime column to datetime64[ns, UTC]
df['datetime_utc'] = df['datetime_utc'].dt.tz_localize('UTC')
df["test"] = ""
table = pn.widgets.Tabulator(df)
table.servable()
```

Traceback:
```
Traceback (most recent call last):
  File "/home/liquetm/miniconda3/envs/dev/lib/python3.11/site-packages/pyviz_comms/__init__.py", line 340, in _handle_msg
    self._on_msg(msg)
  File "/home/liquetm/miniconda3/envs/dev/lib/python3.11/site-packages/panel/viewable.py", line 479, in _on_msg
    doc.unhold()
  File "/home/liquetm/miniconda3/envs/dev/lib/python3.11/site-packages/bokeh/document/document.py", line 776, in unhold
    self.callbacks.unhold()
  File "/home/liquetm/miniconda3/envs/dev/lib/python3.11/site-packages/bokeh/document/callbacks.py", line 431, in unhold
    self.trigger_on_change(event)
  File "/home/liquetm/miniconda3/envs/dev/lib/python3.11/site-packages/bokeh/document/callbacks.py", line 408, in trigger_on_change
    invoke_with_curdoc(doc, event.callback_invoker)
  File "/home/liquetm/miniconda3/envs/dev/lib/python3.11/site-packages/bokeh/document/callbacks.py", line 443, in invoke_with_curdoc
    return f()
           ^^^
  File "/home/liquetm/miniconda3/envs/dev/lib/python3.11/site-packages/bokeh/util/callback_manager.py", line 185, in invoke
    callback(attr, old, new)
  File "/home/liquetm/miniconda3/envs/dev/lib/python3.11/site-packages/panel/reactive.py", line 470, in _comm_change
    state._handle_exception(e)
  File "/home/liquetm/miniconda3/envs/dev/lib/python3.11/site-packages/panel/io/state.py", line 458, in _handle_exception
    raise exception
[...]

PR #6879, opened 2024-05-30
Tabulator: Fix editing a table that contains a timezone-aware datetime column

maximlt:
Fixes https://github.com/holoviz/panel/issues/6673

### DIFF
--- a/panel/reactive.py
+++ b/panel/reactive.py
@@ -1223,21 +1223,25 @@ class ReactiveData(SyncableData):
         if dtype.kind == 'M':
             if values.dtype.kind in 'if':
                 if getattr(dtype, 'tz', None):
-                    import pandas as pd
+                    # dtype has a timezone
+                    if dtype.tz == dt.timezone.utc:
+                        # Milliseconds to nanoseconds, to datetime64.
+                        converted = (values * 1e6).astype('datetime64[ns]')
+                    else:
+                        import pandas as pd
 
-                    # dtype has a timezone, using pandas to convert
-                    # from milliseconds timezone aware to utc nanoseconds.
-                    converted = (
-                        pd.Series(pd.to_datetime(values, unit="ms"))
-                        .dt.tz_localize(dtype.tz)
-                        .dt.tz_convert('utc')
-                        .astype(np.int64)
-                    )
-                    converted = converted.astype('datetime64[ns]')
+                        # Using pandas to convert from milliseconds
+                        # timezone-aware, to UTC nanoseconds, to datetime64.
+                        converted = (
+                            pd.Series(pd.to_datetime(values, unit="ms"))
+                            .dt.tz_localize(dtype.tz)
+                            .dt.tz_convert('utc')
+                            .dt.tz_localize(None)
+                        )
                 else:
-                    # timestamps converted from milliseconds to nanoseconds,
-                    # and to datetime.
-                    converted = (values * 10e5).astype(dtype)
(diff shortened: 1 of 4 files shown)


panel/reactive.py
```diff
@@ -1221,9 +1221,23 @@ def _convert_column(
...
+
+                    # dtype has a timezone, using pandas to convert
+                    # from milliseconds timezone aware to utc nanoseconds.
+                    converted = (
+                        pd.Series(pd.to_datetime(values, unit="ms"))
+                        .dt.tz_localize(dtype.tz)
+                        .dt.tz_convert('utc')
+                        .astype(np.int64)
+                    )
+                    converted = converted.astype('datetime64[ns]')
```
hoxbro, line 1236:
```suggestion
                        .dt.localize(None)
                    )
```

Maybe check if the timezone is UTC and skip some of the steps.

panel/tests/widgets/test_tables.py
```diff
@@ -1973,6 +1972,39 @@ def test_server_edit_event():
...
     assert events[0].old == 1
 
+
+def test_edit_with_datetime_aware_column():
+    # https://github.com/holoviz/panel/issues/6673
+
+    # The order of these columns matter, 'B' should be first as it's in fact
+    # processed first when 'A' is edited.
+    data = {
+        "B": pd.date_range(start='2024-01-01', end='2024-01-03', freq='D', tz='utc'),
```
hoxbro, line 1982: Maybe also have another `tz` than 'utc'?

``` python
import pytz
pytz.timezone('US/Eastern')
```

panel/reactive.py
```diff
@@ -1221,9 +1221,27 @@ def _convert_column(
         dtype = old_values.dtype
         converted: list | np.ndarray | None = None
         if dtype.kind == 'M':
-            if values.dtype.kind in 'if':# TODO: need to doublecheck if this is still needed
-                NATs = np.isnan(values)
-                converted = np.where(NATs, np.nan, values * 10e5).astype(dtype)
```
philippjfr, line 1226: So the NaT conversion isn't needed anymore?

maximlt: Indeed, the original code was as follows:

```python
                NATs = values == BOKEH_JS_NAT
                converted = np.where(NATs, np.nan, values * 10e5).astype(dtype)
```

Now Bokeh no longer uses a special value to mark NaT, it's just `np.nan`, so we no longer need to replace them with ... `np.nan` :)

panel/reactive.py
```diff
@@ -1221,9 +1221,27 @@ def _convert_column(
...
+                    else:
+                        import pandas as pd
+
+                        # Using pandas to convert from milliseconds
+                        # timezone-aware, to UTC nanoseconds, to datetime64.
+                        converted = (
+                            pd.Series(pd.to_datetime(values, unit="ms"))
+                            .dt.tz_localize(dtype.tz)
+                            .dt.tz_convert('utc')
+                            .dt.tz_localize(None)
```
philippjfr, line 1239: What is the intention behind these last two lines? This will soon cause pandas to error and also I'm quite confused. This localizes in the original provided timezone, then shifts it to UTC, which seems wrong.

philippjfr: I've made a change here https://github.com/holoviz/panel/pull/6889/commits/8d1e9f8dc92696cf98c65408b1233d9c1e20cb00

PR #6883, opened 2024-05-31
Test dev release of Bokeh 3.5

hoxbro:
Bokeh 3.5 has dropped Python 3.9.

### DIFF
--- a/panel/models/ace.ts
+++ b/panel/models/ace.ts
@@ -51,7 +51,7 @@ export class AcePlotView extends HTMLBoxView {
         height: "100%",
         zIndex: 0,
       },
-    })
+    } as any)
     this.shadow_el.append(this._container)
     this._container.textContent = this.model.code
     this._editor = ace.edit(this._container)
--- a/panel/models/ace.ts
+++ b/panel/models/ace.ts
@@ -49,7 +49,7 @@ export class AcePlotView extends HTMLBoxView {
       style: {
         width: "100%",
         height: "100%",
-        zIndex: 0,
+        zIndex: "0",
       },
     } as any)
     this.shadow_el.append(this._container)
--- a/panel/models/ace.ts
+++ b/panel/models/ace.ts
@@ -51,7 +51,7 @@ export class AcePlotView extends HTMLBoxView {
         height: "100%",
         zIndex: "0",
       },
-    } as any)
+    })
     this.shadow_el.append(this._container)
     this._container.textContent = this.model.code
     this._editor = ace.edit(this._container)
--- a/panel/models/column.ts
+++ b/panel/models/column.ts
@@ -1,5 +1,5 @@
 import {Column as BkColumn, ColumnView as BkColumnView} from "@bokehjs/models/layouts/column"
-import * as DOM from "@bokehjs/core/dom"
+import {div} from "@bokehjs/core/dom"
 import type * as p from "@bokehjs/core/properties"
 
 export class ColumnView extends BkColumnView {
@@ -62,7 +62,7 @@ export class ColumnView extends BkColumnView {
 
   override render(): void {
     super.render()
-    this.scroll_down_button_el = DOM.create_element("div", {class: "scroll-button"})
+    this.scroll_down_button_el = div({class: "scroll-button"})
     this.shadow_el.appendChild(this.scroll_down_button_el)
     this.el.addEventListener("scroll", () => {
       this.record_scroll_position()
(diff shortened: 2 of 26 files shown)


scripts/build_pyodide_wheels.py
```diff
@@ -107,14 +107,13 @@
     ):
         continue
     elif filename.startswith("bokeh-") and filename.endswith("METADATA"):
-        # remove tornado dependency
+        # Replace tornado dependency with bokeh-sampledata
```
hoxbro, line 110: This is a bit of a hack...

panel/models/ace.ts
```diff
@@ -51,7 +51,7 @@ export class AcePlotView extends HTMLBoxView {
         height: "100%",
         zIndex: 0,
       },
-    })
+    } as any)
```
hoxbro, line 54: Should be removed when https://github.com/bokeh/bokeh/issues/13919 is resolved

panel/models/singleselect.ts
```diff
@@ -37,7 +37,7 @@ export class SingleSelectView extends InputWidgetView {
 
       const disabled = this.model.disabled_options.includes(value)
 
-      return option({value, disabled}, _label)
+      return option({value, disabled} as any, _label)
```
hoxbro, line 40: Should be removed when https://github.com/bokeh/bokeh/issues/13919 is resolved

panel/models/perspective.ts
```diff
@@ -166,7 +166,7 @@ export class PerspectiveView extends HTMLBoxView {
     const container = div({
       class: "pnx-perspective-viewer",
       style: {
-        zIndex: 0,
+        zIndex: "0",
```
mattpap, line 169:
```ts
    panel/models/perspective.ts:166:27 - error TS2345: Argument of type '{ class: string; style: { zIndex: number; }; }' is not assignable to parameter of type 'HTMLChild | HTMLAttrs<"div", {}>'.
      Types of property 'style' are incompatible.
        Type '{ zIndex: number; }' is not assignable to type 'CSSStyles'.
          Type '{ zIndex: number; }' is not assignable to type 'CSSStylesCamel'.
            Types of property 'zIndex' are incompatible.
```
The correct type is `string`, but some CSS APIs allow `any` and convert to `string`. I allow only strings to make sure the conversion is intentional and not a bug (`any` will gladly accept anything without second thoughts and CSS parser/validator doesn't issue any errors or even warnings). There aren't many CSS properties that allow plain numbers (e.g. `width: 100;` is invalid), but I can make an exception for `z-index` and perhaps some more.

panel/models/column.ts
```diff
@@ -62,7 +62,7 @@ export class ColumnView extends BkColumnView {
 
   override render(): void {
     super.render()
-    this.scroll_down_button_el = DOM.createElement("div", {class: "scroll-button"})
+    this.scroll_down_button_el = DOM.create_element("div", {class: "scroll-button"})
```
mattpap, line 65: `createElement()` wasn't meant to be exposed. I would suggest using `div(...)` for non variadic cases.

panel/models/quill.ts
```diff
@@ -57,7 +57,7 @@ export class QuillInputView extends HTMLBoxView {
 
   override render(): void {
     super.render()
-    this.container = div({style: "visibility: hidden;"})
+    this.container = div({style: {visibility: "hidden"}})
```
mattpap, line 60: I will restore `string` support in `style` attribute, because this is perfectly valid DOM.

panel/models/perspective.ts
hoxbro: There's no need for that; I like the strictness. 

I just didn't have time to look at it Friday, and just pasted in the build errors in the first post.

pixi.toml
```diff
@@ -1,14 +1,13 @@
 [project]
 name = "panel"
-channels = ["pyviz/label/dev", "bokeh", "conda-forge"]
+channels = ["pyviz/label/dev", "bokeh/label/dev", "conda-forge"]
 platforms = ["linux-64", "osx-arm64", "osx-64", "win-64"]
 
 [tasks]
-download-data = 'bokeh sampledata'
+# download-data = 'bokeh sampledata'
```
hoxbro, line 7:
```suggestion
```

philippjfr (approved):

scripts/build_pyodide_wheels.py
philippjfr: I'm okay with it.

philippjfr (approved):

PR #6889, opened 2024-06-03
Update warnings to raise errors in test

hoxbro:
Now we raise warnings as exceptions.


pyproject.toml
```diff
@@ -197,7 +197,12 @@ xfail_strict = true
...
 log_cli_level = "INFO"
 filterwarnings = [
+    "error",
     # 2023-11: `pkg_resources` is deprecated
     "ignore:Deprecated call to `pkg_resources.+?'zope:DeprecationWarning",  # https://github.com/zopefoundation/meta/issues/194
     "ignore: pkg_resources is deprecated as an API:DeprecationWarning:streamz.plugins", # https://github.com/python-streamz/streamz/issues/460
+    # 2024-06: Adding error to the filterwarnings
+    "ignore:Jupyter is migrating its paths to use standard platformdirs:DeprecationWarning", # OK
+    "ignore:distutils Version classes are deprecated:DeprecationWarning:ipywidgets_bokeh.kernel", # OK
+    "ignore:unclosed file <_io.TextIOWrapper name='/dev/null' mode='w':ResourceWarning", # OK
```
hoxbro, line 207:
```suggestion
    "ignore:unclosed file <_io.TextIOWrapper name='/dev/null' mode='w':ResourceWarning", # OK (Unix)
    "ignore:unclosed file <_io.TextIOWrapper name='nul' mode='w':ResourceWarning", # OK (Windows)
```

hoxbro:
The two big things I referred to are seen in the UI test.

panel/widgets/tables.py
```diff
@@ -712,7 +712,10 @@ def stream(self, stream_value, rollover=None, reset_index=True):
             if reset_index:
                 stream_value = stream_value.reset_index(drop=True)
                 stream_value.index += value_index_start
-            combined = pd.concat([self.value, stream_value])
+            if self.value.empty:
+                combined = stream_value
```
hoxbro, line 716: Could be a problem if the stream_value does not contain all columns. I don't know if this is possible here.

panel/io/ipywidget.py
```diff
@@ -81,8 +82,10 @@ def _on_widget_constructed(widget, doc=None):
         'metadata': {
             'version': __protocol_version__
         },
-        'kernel': kernel
+        'kernel': kernel,
     }
+    if Version(ipykernel.__version__) >= Version('6.22.0'):
+        args['show_warning'] = False
```
hoxbro, line 88: Isn't this just hiding the problem?

philippjfr: No, this is the mechanism they use internally as well.

hoxbro: `ipykernel.comm.Comm` is deprecated for `comm.create_comm`, so they will remove it at some point, and things will start to fail.  If I recall correctly, they had some problems with this transition a year ago, which is why this class still exists. 

Just pasted the link to the warning message so I don't have to find it again: https://github.com/ipython/ipykernel/blob/a7d66ae2197e0d7471ba160542cf5ff7713084b5/ipykernel/comm/comm.py#L82-L83

philippjfr: Calling the Comm class directly is deprecated, the class itself isn't going away.

philippjfr: Actually maybe using the ipykernel BaseComm is sufficient.

pixi.toml
```diff
@@ -73,7 +73,7 @@ ipywidgets = "*"
 ipywidgets_bokeh = "*"
 jupyter_bokeh = ">=3.0.7"
 networkx = ">=2.5"
-pydeck = "*"
+pydeck = ">=0.7.1"
```
hoxbro, line 76: First release after Python 3.10

hoxbro: Needs this pin because if numpy 2.0 is used pydeck v0.2.1 will be used.

panel/widgets/tables.py
```diff
@@ -371,6 +371,7 @@ def _sort_df(self, df: pd.DataFrame) -> pd.DataFrame:
 
         # Temporarily add _index_ column because Tabulator uses internal _index
         # as additional sorter to break ties
+        df = df.copy()
```
hoxbro, line 374: I want to flag this line. Before this change, we changed the column dtype of the original dataframe. @maximlt I know you have spent some time sorting with tabulator/python.

maximlt: Could you precise what warning this is fixing and which column dtype is changed?

hoxbro: The following warnings, in the UI test ([example](https://github.com/holoviz/panel/actions/runs/9704627234/job/26785216150#step:6:150))
``` python-traceback
panel/tests/ui/widgets/test_tabulator.py: 27 warnings
  D:\a\panel\panel\panel\widgets\tables.py:374: SettingWithCopyWarning: 
  A value is trying to be set on a copy of a slice from a DataFrame.
  Try using .loc[row_indexer,col_indexer] = value instead
  
  See the caveats in the documentation: https://pandas.pydata.org/pandas-docs/stable/user_guide/indexing.html#returning-a-view-versus-a-copy
    df['_index_'] = np.arange(len(df)).astype(str)

panel/tests/ui/widgets/test_tabulator.py: 27 warnings
  D:\a\panel\panel\panel\widgets\tables.py:403: SettingWithCopyWarning: 
  A value is trying to be set on a copy of a slice from a DataFrame
  
  See the caveats in the documentation: https://pandas.pydata.org/pandas-docs/stable/user_guide/indexing.html#returning-a-view-versus-a-copy
    df.drop(columns=['_index_'], inplace=True)
```

And it change the dtype of the column itself (`df.columns.dtype`), see `test_tabulator_sorters_unnamed_index` and `test_tabulator_sorters_int_name_column`

maximlt: Ok thanks. I think the warning is caused by the fact that `_sort_df` is called after `_filter_dataframe` which may return a view of `self.value`. If you think that's the reason, adding a comment wouldn't hurt (and there's maybe a better place where to call `.copy()`).

panel/reactive.py
```diff
@@ -1284,21 +1281,14 @@ def _convert_column(
...
-
-                        # Using pandas to convert from milliseconds
-                        # timezone-aware, to UTC nanoseconds, to datetime64.
-                        converted = (
-                            pd.Series(pd.to_datetime(values, unit="ms"))
-                            .dt.tz_localize(dtype.tz)
-                            .dt.tz_convert('utc')
-                            .dt.tz_localize(None)
-                        )
+                    import pandas as pd
```
maximlt, line 1284: @philippjfr commented on https://github.com/holoviz/panel/pull/6879 about the original code:

```
converted = (
    pd.Series(pd.to_datetime(values, unit="ms"))
    .dt.tz_localize(dtype.tz)
    .dt.tz_convert('utc')
    .dt.tz_localize(None)
)
````

> What is the intention behind these last two lines? This will soon cause pandas to error and also I'm quite confused. This localizes in the original provided timezone, then shifts it to UTC, which seems wrong.

I think that was me just being confused with timezone handling and Bokeh serialization to Unix timestamps. It seems to work fine with these changes!

philippjfr: Perfect, thanks for confirming!

philippjfr (approved):

PR #6890, opened 2024-06-03
Tabulator: clean up old Bokeh NaT usage

maximlt:
Tiny PR that updates the custom timestamp sorter that was relying on the old Bokeh NaT sentinel (-9223372036854776), Bokeh uses now `NaN` instead.


maximlt:
I noticed part of the issue described in https://github.com/holoviz/panel/pull/3839 is still valid, i.e. that there are differences between how Tabulator sorts and `df.sort_values` with where NaNs end up, I'll address those in another PR.

philippjfr (approved):

PR #6903, opened 2024-06-07
Update conf.py to dynamically show current release number in announcement text at top of Panel site.

Coderambling:
Announcement at top of site now says: "Panel 1.4 has just been released. ..."

Introducing a variable that retrieves current version number , so it will say "Panel 1.4.x has just been released. ..."

This way the announcement will always show the actual latest release number to new and returning users. More accurate and more informative.

Also, it will more accurately reflect the fact that Panel has frequent releases. Now it says 1.4 just released for a very long period,.

The announcement is on every page of the website, so every site visitor sees this...

Thanks to @maximlt for pointing out line 77 of conf py .

### DIFF
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -44,7 +44,8 @@ html_theme = "pydata_sphinx_theme"
 html_favicon = "_static/icons/favicon.ico"
 
 current_release = panel.__version__  # Current release version variable
-announcement_text = f"Panel {current_release} has just been released! Checkout the <a href='https://panel.holoviz.org/about/releases.html#version-1-4-0'>release notes</a> and support Panel by giving it a 🌟 on <a href='https://github.com/holoviz/panel'>Github</a>."
+
+announcement_text = f"Panel {current_release} has just been released! Check out the <a href='https://panel.holoviz.org/about/releases.html#version-1-4-0'>release notes</a> and support Panel by giving it a 🌟 on <a href='https://github.com/holoviz/panel'>Github</a>."
 
 
 html_theme_options = {
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -43,6 +43,10 @@ html_css_files += [
 html_theme = "pydata_sphinx_theme"
 html_favicon = "_static/icons/favicon.ico"
 
+current_release = panel.__version__  # Current release version variable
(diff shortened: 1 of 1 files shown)


Coderambling:
Checks have passed successfully.  Does that make sure the version number will display on the site as intended?

Or can that only be confirmed after a new build of the site?

philippjfr:
> Or can that only be confirmed after a new build of the site?

Correct but the change looks safe.

philippjfr (approved):

doc/conf.py
```diff
@@ -43,6 +43,11 @@
 html_theme = "pydata_sphinx_theme"
 html_favicon = "_static/icons/favicon.ico"
 
+current_release = panel.__version__  # Current release version variable
+
+announcement_text = f"Panel {current_release} has just been released! Check out the <a href='https://panel.holoviz.org/about/releases.html#version-1-4-0'>release notes</a> and support Panel by giving it a 🌟 on <a href='https://github.com/holoviz/panel'>Github</a>."
```
maximlt, line 48: Arf unfortunately this is not quite correct, `annoucement_text` contains a link to `releases.html#version-1-4-0` ... @Coderambling can you please open a new PR to remove this fragment? No need to build it dynamically I think, since we always want users to see the top of the release notes page.

Coderambling: Yes wel spotted, I will replace that link with the one below ok? That will show the top of the release page.

https://panel.holoviz.org/about/releases.html

maximlt: Ok.

Coderambling: Also, as the next PR after that, the Github API could be used to retrieve the release date,  so instead the text would then be

Panel `current_release` was released on `release_date`.

Would that be an improvement?

Example code to retrieve release date:


`import requests


url = "https://api.github.com/repos/holoviz/panel/releases/latest"

response = requests.get(url)
data = response.json()

print("The latest release date is:", data['published_at'])
`

It would mean importing the requests library in conf.py though. Would that be an issue?

Advantage is that it could then also be used to retrieve the number of Github Panel downloads and forks for display on the site

This would be a good way to communicate to (first time) site visitors that Panel is in active development and is a popular framework.

Coderambling (commented):
See comment.

doc/conf.py
Coderambling: Updated the hyperlink:

https://github.com/holoviz/panel/pull/7046

Coderambling:
Hm it seems this wasn't applied. 

Top of webpages still shows the old text. Is this because it is scheduled to be applied in version 1.5?

Or is there another cause?

PR #6908, opened 2024-06-10
Auth: guard against user being removed

maximlt:
A user reported sometimes getting a `KeyError` at line 1008 in `auth.py`.

### DIFF
--- a/panel/auth.py
+++ b/panel/auth.py
@@ -1108,8 +1108,8 @@ class OAuthProvider(BasicAuthProvider):
         state._active_users[user] -= 1
         if not state._active_users[user]:
             del state._active_users[user]
-            # Don't remove the user when it's set to an empty dict in the
-            # overrides, as it means it is being refreshed.
+            # Don't remove the user override when it is set to None or
+            # is missing, as this means it is being refreshed.
             if state._oauth_user_overrides.get(user) is not None:
                 del state._oauth_user_overrides[user]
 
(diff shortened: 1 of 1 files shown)


panel/auth.py
```diff
@@ -1005,7 +1005,7 @@ async def get_user(handler):
             now_ts = dt.datetime.now(dt.timezone.utc).timestamp()
             expiry = None
             if user in state._oauth_user_overrides:
-                while not state._oauth_user_overrides[user]:
+                while not state._oauth_user_overrides.get(user):
```
philippjfr, line 1008: This change isn't safe. If the user override is deleted then the next line after the while loop will error out.

philippjfr: Sorry, as you rightly pointed out the problem here is that this change can cause this while loop to run forever.

philippjfr: The correct fix is to ensure that we do not delete the user override while it is refreshing:

```python
if state._oauth_user_overrides.get(user) is not None:
    del state._oauth_user_overrides[user]
```

maximlt: Alright it makes sense to me! I reverted the fix changes and applied this one.

panel/auth.py
```diff
@@ -1108,7 +1108,9 @@ def _remove_user(self, session_context):
         state._active_users[user] -= 1
         if not state._active_users[user]:
             del state._active_users[user]
-            if user in state._oauth_user_overrides:
+            # Don't remove the user when it's set to an empty dict in the
+            # overrides, as it means it is being refreshed.
```
philippjfr, line 1112:
```suggestion
            # Don't remove the user override when it is set to None or
            # is missing, as this means it is being refreshed.
```

Issue #6902, opened 2024-06-07
test_multi_select_double_click error


cdeil:
@hoxbro - I see this UI test error:

```
(base) panel $ pixi run test-ui
✨ Pixi task (_install-ui in test-ui): playwright install chromium

✨ Pixi task (test-ui in test-ui): pytest panel/tests/ui --ui --browser chromium -n logical --dist loadgroup --reruns 3 --reruns-delay 10
=================================================================================== test session starts ====================================================================================
platform darwin -- Python 3.12.3, pytest-7.4.4, pluggy-1.5.0
rootdir: /Users/cdeil/code/oss/panel
configfile: pyproject.toml
plugins: cov-5.0.0, github-actions-annotate-failures-0.2.0, playwright-0.5.0, rerunfailures-14.0, anyio-4.4.0, xdist-3.5.0, base-url-2.1.0, asyncio-0.23.6
asyncio: mode=Mode.AUTO
10 workers [529 items]  kipped                                                                                                                                                             
sssssssss.............s...........................................................................................................................................R......R.......... [ 33%]
......................R.........................................................................................................................R.x.R.xR.xx....R.................... [ 67%]
.......................x.....................R.xx..R.......x.....R.x.RR..........................xxx.xxRxxFxxxxxx..xxxxxx.xxxxx..........................................R.....RR..F [ 99%]
F..RF.                                                                                                                                                                               [100%]
========================================================================================= FAILURES =========================================================================================
______________________________________________________________________________ test_multi_select_double_click ______________________________________________________________________________
[gw5] darwin -- Python 3.12.3 /Users/cdeil/code/oss/panel/.pixi/envs/test-ui/bin/python3.12

page = <Page url='http://localhost:55237/'>

    def test_multi_select_double_click(page):
        clicks = []
        select = MultiSelect(options=['A', 'B', 'C'], on_double_click=clicks.append)
    
        serve_component(page, select)
    
        page.locator('option').nth(1).dblclick()
    
>       wait_until(lambda: clicks and clicks[0].option == 'B')
E       ValueError: `wait_until` callback must return None, True or False, returned []

panel/tests/ui/widgets/test_select.py:33: ValueError
```

Some others time out:
```
FAILED panel/tests/ui/widgets/test_select.py::test_multi_select_double_click - ValueError: `wait_until` callback must return None, True or False, returned []
[...]

PR #6909, opened 2024-06-10
Fix test-CI

hoxbro:
Conda-forge now has `playwright`, though they patched it to remove the download option, and it only works on the UNIX system for now. For now, I will keep using the one from the Microsoft channel. It does seem like the conda-forge equivalent is called `python-playwright`. 

Also, fixing the error shown in https://github.com/holoviz/panel/issues/6902

I think the failing test in `test_param_generator` is because we don't have an event loop, which takes time to start on CI, therefore missing the `wait_until` time.

### DIFF
--- a/panel/pane/holoviews.py
+++ b/panel/pane/holoviews.py
@@ -37,6 +37,12 @@ if TYPE_CHECKING:
     from pyviz_comms import Comm
 
 
+def check_holoviews(version):
+    import holoviews as hv
+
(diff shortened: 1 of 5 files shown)


hoxbro:
The failing holoviews is because of changes made in https://github.com/holoviz/holoviews/pull/6256

philippjfr (approved):

PR #6892, opened 2024-06-05
Improve Ipywidgets reference guide

MarcSkovMadsen:
Working on https://github.com/developmentseed/lonboard/pull/539 and reading the questions in https://github.com/developmentseed/lonboard/issues/262 I released that the Ipywidgets reference guide does not explain fully how Panel supports ipywidgets.

This PR adds the missing explanations.

I've also found inspiration in the Shiny docs https://shiny.posit.co/py/docs/jupyter-widgets.html. Thanks to Shiny.

Issue #6911, opened 2024-06-10
Document how to use .from_param to create a widget from a reactive parameter


MarcSkovMadsen:
I can't figure out how to create a widget from a reactive parameter.

More specifically

```python
import panel as pn

pn.extension()

zoom = pn.rx(2)

slider = pn.widgets.IntSlider.from_param(zoom, start=1, end=10)
```

I get

```bash
AttributeError: 'rx' object has no attribute 'name'

Traceback (most recent call last):
  File "/home/jovyan/repos/private/panel-geospatial/.venv/lib/python3.11/site-packages/panel/io/handlers.py", line 389, in run
    exec(self._code, module.__dict__)
  File "/home/jovyan/repos/private/panel-geospatial/pages/03_mapbox.py", line 11, in <module>
    pn.widgets.IntSlider.from_param(zoom, start=1, end=10)
  File "/home/jovyan/repos/private/panel-geospatial/.venv/lib/python3.11/site-packages/panel/widgets/base.py", line 93, in from_param
    parameter, widgets={parameter.name: dict(type=cls, **params)},
                        ^^^^^^^^^^^^^^
  File "/home/jovyan/repos/private/panel-geospatial/.venv/lib/python3.11/site-packages/param/reactive.py", line 1032, in __getattribute__
    return super().__getattribute__(name)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AttributeError: 'rx' object has no attribute 'name'
```

I cannot find this documented anywhere.

MarcSkovMadsen:
A workaround would be to define the widget first and the from it the reactive parameter. But the problem is that then I cannot update the value of the reactive parameter

```bash
AttributeError: Setting the value of a reactive expression is only supported if it wraps a concrete value. A reactive expression wrapping a Parameter or another dynamic reference cannot be updated.
```

I believe it should be possible to update it.

```python
import panel as pn

pn.extension("ipywidgets")

zoom_input = pn.widgets.IntSlider(value=2, start=1, end=10, name="Zoom")
zoom = zoom_input.rx()
zoom.rx.value=3
```

MarcSkovMadsen:
This work work either

```python
import panel as pn

pn.extension()

zoom = pn.rx(2)
zoom_input = pn.widgets.FloatSlider(value=zoom, start=1, end=10, step=1.0, name="Zoom")
zoom_input.value=3
print(zoom.rx.value)
```

[...]

Issue #3687, opened 2022-07-12
Bind Api: Make it easy to create forms with .bind like api.


MarcSkovMadsen:
# Request

Make it simple to create forms via a like `pn.bind` api.

## Motivation

I'm trying to create an example for lighting.ai. One of the pages is a form. 

![image](https://user-images.githubusercontent.com/42288570/178566767-ab812d12-f605-43d2-8ea7-4c7f0e4225bd.png)

I would like to 

- use the `pn.bind` api as that is the recommended one.
- not nest functions inside functions as that makes the logic hard to reason about and test.
- write simple and readable code

But I found it takes thinking and *wrapper* functions to support this very common workflow with the `pn.bind` api.

## Form code: As Is

```python
import panel as pn

def execute_business_logic(input1, input2):
    print(input1, input2)

# Now I want to make a form to execute this 

input1 = pn.widgets.TextInput(value="1")
input2 = pn.widgets.TextInput(value="2")
submit_button = pn.widgets.Button(name="Submit")

def submit(_):
    # Lots of users don't know _
    # I takes mental bandwidth to figure out you need a wrapper function
    execute_business_logic(input1.value, input2.value)

pn.bind(submit, submit_button, watch=True)

# Create the form
pn.Column(
    input1, input2, submit_button
).servable()
```

I would like to avoid the `submit` wrapper function as this is such a normal pattern.

## Form Code: To Be

We could introduce `bind_as_form`

```python
import panel as pn

def _to_value(value):
    if hasattr(value, "value"):
        return value.value
    return value

def bind_as_form(function, *args, submit, watch=False, **kwargs):
    """Extends pn.bind to support "Forms" like binding. I.e. triggering only when a Submit button is clicked,
    but using the dynamic values of widgets or Parameters as inputs.
    
    Args:
        function (_type_): The function to execute
        submit (_type_): The Submit widget or parameter to depend on
        watch (bool, optional): Defaults to False.

    Returns:
        _type_: A Reactive Function
    """
    if not args:
        args = []
    if not kwargs:
        kwargs = {}

    def function_wrapper(_, args=args, kwargs=kwargs):
        args=[_to_value[value] for value in args]
        kwargs={key: _to_value(value) for key, value in kwargs.items()}
        return function(*args, **kwargs)
    return pn.bind(function_wrapper, submit, watch=watch)
```

This would make the api much simpler

```python
def execute_business_logic(input1, input2):
    print(input1, input2)

# Now I want to make a form to execute this 

[...]

PR #6912, opened 2024-06-11
Make it easy to use ipywidgets using familiar apis

MarcSkovMadsen:
Continuation of #6892.

Also motivated by me trying to demonstrate that you can just as well use Panel for geospatial applications as Solara by creating apps similar to https://github.com/opengeos/solara-geospatial/tree/main/pages. But currently Panel is harder to use because it requires adding more code for using `observer` pattern.

## Todo

- [x] Create design proposal
- [ ] Get design feedback
- [ ] Update design proposal
- [ ] Implement design
- [ ] Document
- [ ] Review and update

### DIFF
--- a/panel/ipywidget.py
+++ b/panel/ipywidget.py
@@ -1,0 +1,48 @@
+"""Functionality to enable easily interacting with ipywidgets via familiar APIs like pn.bind,
+@pn.depends and pn.rx"""
+
+from typing import TYPE_CHECKING, Iterable
+
+import param
+
+from .viewable import Viewer
+
+if TYPE_CHECKING:
+    try:
+        from traitlets import HasTraits
+    except ModuleNotFoundError:
+
+        class HasTraits:  # type: ignore
+            """Mock class"""
+
+else:
+
+    class HasTraits:  # type: ignore
+        """Mock class"""
+
+
+def to_viewer(widget: HasTraits, parameters: Iterable | None = None) -> Viewer:
+    """Returns a Viewer object with parameters synced to the ipywidget widget parameters
+
+    Args:
+        widget (HasTraits): The ipywidget to create the Viewer from.
+        parameters (Iterable | None): The parameters to add to the Viewer and to sync.
+            If no parameters are specified all parameters on the widget will be added
+            and synced.
+
+    """
+    raise NotImplementedError()
+
+
+def to_rx(
+    widget: HasTraits, parameters: Iterable | None = None
+) -> tuple[param.rx, ...]:
+    """Returns a tuple of `rx` parameters. Each one synced to a parameter of the ipywidget widget.
+
+    Args:
+        widget (HasTraits): The ipywidget to create the `rx` parameters from.
+        parameters (Iterable): The parameters to create `rx` parameters from and to sync.
+            If no parameters are specified all parameters on the widget will be added
+            and synced.
+    """
+    raise NotImplementedError()
--- a/panel/tests/test_ipywidget.py
+++ b/panel/tests/test_ipywidget.py
@@ -1,0 +1,130 @@
+"""We test that we can work with Ipywidgets and AnyWidgets via Parameters or Reactive values
+instead of Traitlets observer pattern.
+
+The purpose is to enable Panel users to easily use Ipywidgets using familiar APIs.
+"""
+
+import param
+
+from traitlets import (
+    Float, HasTraits, Int, Unicode,
+)
+
+import panel as pn
+
+from panel.ipywidget import to_rx, to_viewer
+
+
+class ExampleTraitlets(HasTraits):
+    name = Unicode("Default Name").tag(description="A string trait")
+    age = Int(0).tag(description="An integer trait")
+    height = Float(0.0).tag(description="A float trait")
+
+
+def test_to_viewer():
+    widget = ExampleTraitlets(name="A", age=1, height=1.1)
+
+    viewer = to_viewer(widget)
+
+    assert isinstance(viewer, pn.viewable.Viewer)
+    assert {"name", "age", "height"} == set(viewer.param.params)
+
+    assert viewer.name == widget.name
+    assert viewer.age == widget.age
+    assert viewer.height == widget.height
+
+    # widget synced to widget
+    widget.name = "B"
+    widget.age = 2
+    widget.height = 2.2
+    assert viewer.name == widget.name
+    assert viewer.age == widget.age
+    assert viewer.height == widget.height
+
+    # widget synced to viewer
+    viewer.name = "C"
+    viewer.age = 3
+    viewer.height = 3.3
+    assert viewer.name == widget.name
+    assert viewer.age == widget.age
+    assert viewer.height == widget.height
+
+    component = viewer.__panel__()
+    assert isinstance(component, pn.pane.IPyWidget)
+    assert component.object == widget
+
+
+def test_to_viewer_parameter_list():
+    widget = ExampleTraitlets(name="A", age=1, height=1.1)
+
+    viewer = to_viewer(widget, parameters=["name", "age"])
+
+    assert isinstance(viewer, pn.viewable.Viewer)
+    assert {"name", "age"} == set(viewer.param.params)
+
+    assert viewer.name == widget.name
+    assert viewer.age == widget.age
+
+    # widget synced to widget
+    widget.name = "B"
+    widget.age = 2
+    assert viewer.name == widget.name
+    assert viewer.age == widget.age
+
+    # widget synced to viewer
+    viewer.name = "C"
+    viewer.age = 3
+    assert viewer.name == widget.name
+    assert viewer.age == widget.age
+
+
+def test_to_rx():
+    widget = ExampleTraitlets(name="A", age=1, height=1.1)
+
+    name, age, height = to_rx(widget)
+    assert isinstance(name, param.rx)
+    assert isinstance(age, param.rx)
+    assert isinstance(height, param.rx)
+
+    assert name.rx.value == widget.name
+    assert age.rx.value == widget.age
+    assert height.rx.value == widget.height
+
+    # widget synced to reactive
+    widget.name = "B"
+    widget.age = 2
+    widget.height = 2.2
+    assert name.rx.value == widget.name
+    assert age.rx.value == widget.age
+    assert height.rx.value == widget.height
+
+    # reactive synced to viewer
+    name.rx.value = "C"
+    age.rx.value = 3
+    height.rx.value = 3.3
+    assert name.rx.value == widget.name
+    assert age.rx.value == widget.age
+    assert height.rx.value == widget.height
+
+
+def test_to_rx_parameter_list():
+    widget = ExampleTraitlets(name="A", age=1, height=1.1)
+
+    name, age = to_rx(widget, parameters=["name", "age"])
+    assert isinstance(name, param.rx)
+    assert isinstance(age, param.rx)
+
+    assert name.rx.value == widget.name
+    assert age.rx.value == widget.age
+
+    # widget synced to reactive
+    widget.name = "B"
+    widget.age = 2
+    assert name.rx.value == widget.name
+    assert age.rx.value == widget.age
+
+    # reactive synced to viewer
+    name.rx.value = "C"
+    age.rx.value = 3
+    assert name.rx.value == widget.name
+    assert age.rx.value == widget.age


MarcSkovMadsen:
Hi @philippjfr . Would you take a look at the design spec, i.e. the current files? Thanks.

panel/ipywidget.py
```diff
@@ -0,0 +1,49 @@
...
+        class HasTraits:  # type: ignore
+            """Mock class"""
+
+else:
+
+    class HasTraits:  # type: ignore
+        """Mock class"""
+
+
+def to_viewer(widget: HasTraits, parameters: Iterable | None = None, **kwargs) -> Viewer:
```
philippjfr, line 24: A full traitlets -> param converter is quite a heavy thing to maintain.

panel/ipywidget.py
```diff
@@ -0,0 +1,49 @@
...
+        parameters (Iterable | None): The parameters to add to the Viewer and to sync.
+            If no parameters are specified all parameters on the widget will be added
+            and synced.
+        kwargs: Any kwargs provided will be forwarded to the IPyWidget pane used to display the
+            widget.
+    """
+    raise NotImplementedError()
+
+
+def to_rx(
```
philippjfr, line 38: See `pn.pane.ipywidgets._ipywidget_transform` for a traitlet -> param converter (currently only handling the `value`).

MarcSkovMadsen: Thx. I've used it.

panel/ipywidget.py
MarcSkovMadsen: The current implementation contains roughly what I would expect to contribute. I hope you would give it a second quick look @philippjfr  and help direct me. You can see what I expect to do more in the todo list in the first post. Thx.

examples/reference/panes/examples/ipywidgets_rx.py
```diff
@@ -0,0 +1,51 @@
+# Original Source: https://github.com/opengeos/solara-geospatial/blob/main/pages/01_leafmap.py
```
MarcSkovMadsen, line 1: This example should be updated and moved to a reference notebook or documentation.

examples/reference/panes/examples/ipywidgets_viewer.py
```diff
@@ -0,0 +1,56 @@
+# Original Source: https://github.com/opengeos/solara-geospatial/blob/main/pages/01_leafmap.py
```
MarcSkovMadsen, line 1: This example should be updated and moved to a reference notebook or documentation.

panel/ipywidget.py
```diff
@@ -0,0 +1,217 @@
...
+    bases: None | Parameterized | Iterable[Parameterized],
+) -> tuple[Parameterized]:
+    if not bases:
+        bases = ()
+    if isclass(bases) and issubclass(bases, Parameterized):
+        bases = (bases,)
+    return tuple(item for item in bases)
+
+
+def to_parameterized(
```
MarcSkovMadsen, line 104: functionality up to and including this function could be moved to Param one day? Should we do that?

examples/reference/panes/examples/ipywidgets_rx.py
```diff
@@ -0,0 +1,48 @@
...
+import leafmap
```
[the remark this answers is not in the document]
MarcSkovMadsen: I've added `pytest.importorskip("ipywidgets")`. Will se if it solves the problem.

panel/wrappers/ipywidget.py
```diff
@@ -0,0 +1,303 @@
+"""Functionality to enable easy interaction with Traitlets models and ipywidgets via familiar APIs from Param like watch, bind, depends, and rx.
```
MarcSkovMadsen, line 1: I would not put the ipywidget wrapping functionality in an `ipywidget` file in the root folder because panels root `__init__` already imports `ipywidget` from `panel.io`.

panel/observers/ipywidget.py
```diff
@@ -0,0 +1,429 @@
...
+    ```
+    """
+    sync_with_parameterized(model, widget, names={name: "value"})
+
+class ModelParameterized(Parameterized):
+    """An abstract Parameterized base class for wrapping a traitlets HasTraits class or instance."""
+
+    model: HasTraits = param.Parameter(allow_None=False, constant=True)
+    # Todo: consider renaming to names or model_names because it will be used publicly
+    _names: Iterable[str] | dict[str, str] = ()
```
MarcSkovMadsen, line 200: Should we rename from `_names` to `names` or `model_names`. It is used publicly. I used the name `_names` to not interfere with a names parameter on an ipywidget.

panel/observers/ipywidget.py
```diff
@@ -0,0 +1,429 @@
+"""Functionality to enable easy interaction with Traitlets models and ipywidgets via familiar APIs from Param like watch, bind, depends, and rx.
```
MarcSkovMadsen, line 1: Should we add the same functionality for Pydantic?

It would 1) mature the api further from the feedback of implementing Pydantic 2) be tremendously helpful as it would make working with Pydantic models so much easier.

panel/__init__.py
```diff
@@ -49,6 +49,7 @@
 
 from . import layout  # noqa
 from . import links  # noqa
+from . import observers  # noqa
```
MarcSkovMadsen, line 52: Should we rename from `panel.observers` to `panel.observer`?

The first is inline with `widgets`, `panes` etc. The second reads better.

doc/how_to/observers/ipywidget.md
```diff
@@ -0,0 +1,186 @@
+# Observe ipywidgets
```
MarcSkovMadsen, line 1: Should we move the Traitlets functionality to Param @philippjfr ? I would like to avoid the effort. So it should be important.

doc/how_to/observers/ipywidget.md
```diff
@@ -0,0 +1,186 @@
...
+viewer = MapViewer(sizing_mode="stretch_both")
+
+pn.Row(pn.Column(viewer.param, scroll=True), viewer, height=400).servable()
+```
+
+The `_names` attribute is an optional iterable or dictionary. It specifies which traits to synchronize to which parameters.
+
+## How to Create a Reactive Value from the Trait of an ipywidget
+
+Use `create_rx` to create a reactive value from the trait of an ipywidget.
```
MarcSkovMadsen, line 163: Its quite constraining that its not possible to create bidirectionally synced widgets from reactive values. Do you know how to solve https://github.com/holoviz/panel/issues/6911 @philippjfr ?

panel/observers/ipywidget.py
```diff
@@ -0,0 +1,429 @@
...
+        model = params["model"]
+        names = params.pop("_names", self._names)
+        if not names:
+            names = _get_public_and_relevant_trait_names(model)
+        names = _ensure_dict(names)
+        self._names = names
+
+        for parameter in names.values():
+            if parameter not in self.param:
+                self.param.add_parameter(parameter, param.Parameter())
```
MarcSkovMadsen, line 215: Should we implement more advanced functionality to get a more specific Parameter than `param.Parameter`? Just support basic `String`, `Int`, `Number` and `Boolean` would be really, really helpful and speed everything up.

doc/how_to/observers/ipywidget.md
MarcSkovMadsen, line 1: The tests fail because this file is not an `index.md` file. But I can't see how that can make sense because later if/ when more observers are added the link will break. I need help.

image

jbednar: Sounds intriguing.

jbednar:
Can `pn.panel` support auto-detecting an ipywidget so that `pn.panel(leaflet_map)` would work, as it's much easier to remember than `pn.wrappers.ipywidget.WidgetViewer(model=leaflet_map)`?

philippjfr:
Would be a breaking change but really shouldn't break much so probably worth doing.

MarcSkovMadsen:
The reason why its deeply nested is that there is already a panel.ipywidget module which is really panel.io.ipywidget.

MarcSkovMadsen:
pn.panel(some_ipywidget) already works as it returns the panel.

MarcSkovMadsen:
> Can `pn.panel` support auto-detecting an ipywidget so that `pn.panel(leaflet_map)` would work, as it's much easier to remember than `pn.wrappers.ipywidget.WidgetViewer(model=leaflet_map)`?

The functionality is now exposed via `panel.dataclass`. The idea is that the functionality/ API can accomodate dataclasses, pydantic etc. in the future.

```python
pn.dataclass.ModelViewer(model=leaflet_map)
```

or

```python
pn.dataclass.ModelParameterized(model=some_pydantic_class)
```

or

```python
pn.dataclass.create_rx(model=some_attrs_class, name="some_attr")
```

panel/observers/ipywidget.py
MarcSkovMadsen: renamed to `_model_names`. In public use, you will be using `names`.

panel/__init__.py
MarcSkovMadsen: obsolete

doc/how_to/observers/ipywidget.md
MarcSkovMadsen: Apparently its not possible

MarcSkovMadsen:
There is not so much more I can do right now. The final thing would be to review `dev` docs. But its broken: `TypeError: Cannot read properties of undefined (reading 'loader')`.

image

I will be waiting for `dev` docs fix and reviews/ feedback.

MarcSkovMadsen:
Ok. I did one more thing. Add support for Pydantic.

I did because

- Its extremely useful
- It helps generalize the code to enable us to avoid breaking changes in the future.

doc/how_to/dataclass/pydantic.md
```diff
@@ -0,0 +1,258 @@
...
+def view_model(*args):
+    return json.loads(model.json())
+
+pn.Column(
+    widget,
+    pn.pane.JSON(pn.bind(view_model, widget)),
+).servable()
+```
+
+Todo: Fix order of updates. The JSON pane updates before the model is updated.
```
MarcSkovMadsen, line 61: For some unknown reason the json pane updates before the model. A workaround is to make `view_model` async and wait 250ms.

This should be understood and fixed.

panel/tests/dataclasses/test_dataclass.py
```diff
@@ -0,0 +1,505 @@
...
+
+
+def test_dont_sync_non_shared_trait(model):
+    class ExampleWidgetViewer(param.Parameterized):
+        wealth = param.Integer(3, bounds=(0, 10))
+
+    parameterized = ExampleWidgetViewer()
+    sync_with_parameterized(model=model, parameterized=parameterized, names=["wealth"])
+
+def test_model_parameterized_parameters_added_to_instance_not_class(model):
```
MarcSkovMadsen, line 496: This test fails. I discovered that `self.param.add_parameter(parameter, param.Parameter())` adds to the class (NOT EXPECTED BY ME) and not the instance.

I need help to understand and fix.

panel/dataclass.py
```diff
@@ -0,0 +1,357 @@
...
+            params["model"] = self._model_class()
+
+        model = params["model"]
+        names = params.pop("names", self._model_names)
+        utils = _get_utils(model)
+        if not names:
+            names = utils.get_public_and_relevant_field_names(model)
+        names = utils.ensure_dict(names)
+        self._model_names = names
+        for parameter in names.values():
```
MarcSkovMadsen, line 149: I discovered that `self.param.add_parameter(parameter, param.Parameter())` adds to the class (NOT EXPECTED BY ME) and not the instance. C.f. test failure.

I need this to only add parameters to the instance. How do I do that?

philippjfr: Not currently possible afaik. Why not dynamically create a class?

panel/dataclass.py
```diff
@@ -0,0 +1,357 @@
...
+    return TraitletsUtils
+
+def sync_with_parameterized(
+    model: DataClassLike,
+    parameterized: Parameterized,
+    names: Iterable[str] | dict[str, str] = (),
+):
+    """Syncs the fields of the model with the parameters of the Parameterized object.
+
+    Args:
```
philippjfr, line 66: NumPy style, not Google style docstrings please.

panel/dataclass.py
philippjfr: Yeah I'll change this to create the class dynamically.

philippjfr: Hmm, I'm exceptionally confused by this. Shouldn't `create_parameterized` be in charge of creating a subclass that creates the Parameters?

panel/dataclass.py
```diff
@@ -0,0 +1,357 @@
...
+    name = type(model).__name__
+    key = (name, tuple(names.items()), bases)
+    if name in _cache_of_model_classes:
+        parameterized = _cache_of_model_classes[key]
+    else:
+        existing_params = ()
+        for base in bases:
+            existing_params += tuple(base.param)
+        params = {
+            name: param.Parameterized()
```
philippjfr, line 198: I think the main mistake was that this should be `param.Parameter`.

philippjfr (commented):
I have solved the instance add_parameter problem by adding a `__new__` method that creates a class dynamically given the supplied names and model. Let me know if you're happy with that approach.

panel/observers/ipywidget.py
philippjfr: Yes, let's start mapping parameter types, initially we can just define the basic ones you describe.

panel/_dataclasses/pydantic.py
```diff
@@ -0,0 +1,62 @@
...
+    @classmethod
+    def observe_field(
+        cls,
+        model,
+        field: str,
+        handle_change,
+    ):
+        # We don't know if this is possible
+        # Maybe solution can be found in
+        # https://github.com/pydantic/pydantic/discussions/7127 or
```
philippjfr, line 40: Pretty certain it isn't without significantly modifying the original class.

doc/how_to/dataclass/pydantic.md
philippjfr: Fixed

philippjfr:
I would like to suggest the following renames:

- `create_parameterized` -> `to_parameterized`
- `create_viewer` -> `to_panel`
- `create_rx` -> `to_rx`

Also I'm still trying to iterate on the sync method API, I would really love it if we didn't need 3 different methods there but I get why you added them.

panel/observers/ipywidget.py
MarcSkovMadsen: Done

MarcSkovMadsen:
> I would like to suggest the following renames:
> 
> * `create_parameterized` -> `to_parameterized`
> * `create_viewer` -> `to_panel`
> * `create_rx` -> `to_rx`
> 
> Also I'm still trying to iterate on the sync method API, I would really love it if we didn't need 3 different methods there but I get why you added them.

You can just used the `sync_xyz` method to create a method like the pseudo method below

```python
def sync(model, parameterized_or_rx, names):
    if isinstance(parameterized_or_rx, param.rx) and isinstance(names, Iterable):
        sync_to_rx(model, parameterized_or_rx, *names)
    elif isinstance(parameterized_or_rx, pn.widgets.Widget) and isinstance(names, str):
        sync_to_widget(model, parameterized_or_rx, names)
    elif isinstance(parameterized_or_rx, param.Parameterized):
        sync_to_parameterized(model, parameterized_or_rx, names)
    raise ValueError()
```

philippjfr:
>You can just use the sync_xyz method to create a method like the pseudo method below

Right, just wondering whether that's too overloaded.

MarcSkovMadsen:
I have starting adding support for specific parameter types. Its easy. But i need to refactor the Way initial values are assigned.

philippjfr:
>I have starting adding support for specific parameter types. Its easy. But i need to refactor the Way initial values are assigned.

Ooops, missed this. I just pushed it.

panel/_dataclasses/pydantic.py
```diff
@@ -43,20 +41,15 @@ def observe_field(
...
     def get_layout(cls, model, self, layout_params):
-        # Todo: Find proper solution to update the layout after the model is updated
-        # Hack: For some reason the model is updated after the UI is rendered
-        # We might setup a callback to update the layout below after the model is updated
-        async def view_model(*args, model=model):
-            if not cls._is_testing:
-                await sleep(0.250)
-            return json.loads(model.json())
+        exclude = list(layout_params)
+        def view_model(*args):
```
MarcSkovMadsen, line 45: Why exclude attributes of a Pydantic model?

MarcSkovMadsen:
> > I have starting adding support for specific parameter types. Its easy. But i need to refactor the Way initial values are assigned.
> 
> Ooops, missed this. I just pushed it.

No worries. I'm happy.

panel/_dataclasses/base.py
```diff
@@ -140,18 +147,19 @@ def sync_with_parameterized(
             if parameter not in mapping:
                 mapping[parameter] = []
             mapping[parameter].append(field)
-            model_field = getattr(model, field)
-            parameterized_value = getattr(parameterized, parameter)
             parameters.append(parameter)
-            if parameter != "name" and parameterized_value is not None:
+
+            field_value = getattr(model, field)
```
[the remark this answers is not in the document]
philippjfr: I wasn't really sure about the correct behavior here so I focused on getting the existing tests passing.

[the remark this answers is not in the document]
philippjfr: If you want to adjust please make changes as you see fit.

panel/_dataclasses/pydantic.py
philippjfr: Don't think there was a good reason.

philippjfr:
Not sure I love `panel.dataclass` naming. While yes, they're dataclass-like things they're not actually dataclasses and I'd rather focus on the functionality this provides, i.e. the fact that this provides a compatibility layer or bridge between the libraries.

MarcSkovMadsen:
> Not sure I love `panel.dataclass` naming. While yes, they're dataclass-like things they're not actually dataclasses and I'd rather focus on the functionality this provides, i.e. the fact that this provides a compatibility layer or bridge between the libraries.

### 1. `panel.compat`
**Explanation:** This name focuses on the core functionality of the module, which is to provide compatibility between different frameworks. It conveys the idea that this module helps integrate various dataclass-like libraries with HoloViz Param.

### 2. `panel.bridge`
**Explanation:** The term "bridge" clearly indicates that the module serves as a connector between disparate systems. It emphasizes the role of the module in linking different frameworks, highlighting its purpose without misrepresenting the nature of the classes involved.

### 3. `panel.integration`
**Explanation:** This name underscores the module's role in facilitating integration between HoloViz Param and other frameworks like traitlets, ipywidgets, and pydantic. It suggests a broader scope of bringing together various components, focusing on the unifying function of the module.

Each of these names highlights the module's purpose of enhancing interoperability and integration between HoloViz Param and other dataclass-like frameworks, without suggesting that the module contains actual dataclasses.

philippjfr:
Docs build now working.

panel/_dataclasses/pydantic.py
MarcSkovMadsen: psygnal is a separate project. But it adds observer pattern/ events to Pydantic.

MarcSkovMadsen:
When I click the *run with pyodide button* I see

image

It seems we need to add sqlite as a dependency.

MarcSkovMadsen:
I can see that the ipywidgets `create_parameter` is not working for the example

```python
import panel as pn
import ipyleaflet as ipyl

pn.extension("ipywidgets")

leaflet_map = ipyl.Map(zoom=4)

viewer = pn.dataclass.ModelViewer(model=leaflet_map, sizing_mode="stretch_both")
pn.Row(pn.Column(viewer.param, scroll=True), viewer, height=400).servable()
```

The problem is that the `trait` is an instance and not a type

image

MarcSkovMadsen:
`Pydantic` seems to missing as a dependency for the docs

image

MarcSkovMadsen:
Ahh. The dataclass functionality for Pydantic does not add specific parameter types.

MarcSkovMadsen:
Our code instantiates pydantic models. Often then don't have default values. Instead initial values are required.

This makes it a bit hard to use our features. Especially for creating *forms* with validation which a popular use case (c.f. pydantic-panel, streamlit-pydantic, dash-pydantic-form).

For example the below test will currently raise an exception

```python
def test_to_parameterized_no_defaults():
    from pydantic import BaseModel
    class ExampleModel(BaseModel):
        some_text: str
        some_number: int
    
    class ExampleModelParameterized(ModelParameterized):
        _model_class = ExampleModel

    ExampleModelParameterized()
```

panel/dataclass.py
```diff
@@ -0,0 +1,507 @@
...
+    for name in names:
+        rx = param.rx()
+        sync_with_rx(model, name, rx)
+        rx_values.append(rx)
+    if len(rx_values) == 1:
+        return rx_values[0]
+    return tuple(rx_values)
+
+
+class ModelForm(Viewer):
```
MarcSkovMadsen, line 407: I added the `ModelForm` to support the *form* use case. This is the use case that pydantic-panel, streamlit-pydantic and dash-pydantic-form all support.

Its really a more general request as in https://github.com/holoviz/panel/issues/3687. We can solve this is several ways:

- not solve :-)
- document how to solve
- make general Panel form functionality
- add this dataclass specific form functionality.

What do you recommend @philippjfr ?

MarcSkovMadsen: Its worth noting that I achieve the *form* functionality a bit differently than the packages mentioned above. They translate pydantic fields to widgets while I translate pydantic fields to Parameters which Panel can translate to widgets. I.e. in my version we instantiate a Parameterized and then use Param. In the packages they can avoid instantiating a model before the user has filled out the form including required values.

Is my version the right way to go @philippjfr ?

MarcSkovMadsen:
I've tried to achieve almost feature parity with pydantic-panel.

What is missing is BaseModel attributes and pandas intervals. 

- BaseModels will be added later
- I did not understand the pandas intervals support.

image

```python
import pydantic
import panel as pn
from typing import List
from pydantic_panel.dispatchers import infer_widget
import numpy as np
import pandas as pd

pn.extension("tabulator")

class ChildModel(pydantic.BaseModel):
    name: str = "child"

class SomeModel(pydantic.BaseModel):
    name: str = "some model"
    
    child_field: ChildModel = ChildModel()
    date_field: date = date(2024,1,2)
    dateframe: pd.DataFrame = pd.DataFrame({"x": [1], "y": ["a"]})
    datetime_field: datetime = datetime(2024,1,1)
    dict_field: dict = {"a": 1}
    float_field: float = 42
    int_field: int = pydantic.Field(default=2, lt=10, gt=0, multiple_of=2)
    list_field: list = [1, "two"]
    nparray_field: np.ndarray = np.array([1, 2, 3])
    str_field: str = pydantic.Field(default = "to", min_length=2, max_length=10)
    tuple_field: tuple = ("a", 1)

    class Config:
        arbitrary_types_allowed = True # to allow np.array

model = SomeModel()

pydantic_panel_editor = pn.panel(model, sizing_mode="fixed") # Pydantic(model).layout[0]
print(type(pydantic_panel_editor))
panel_editor = pn.Param(pn.dataclass.to_parameterized(model))

pn.Row(
    pydantic_panel_editor,
    panel_editor,
).servable()
```

MarcSkovMadsen:
What else should be done here?

philippjfr:
I'm still fully on board with the aims of this PR but it's simply too large a PR to make it into 1.5.0 at this point.

panel/dataclass.py
Coderambling: Great work! And the form use case should definitely be supported.

doc/how_to/dataclass/ipywidget.md
```diff
@@ -0,0 +1,185 @@
...
+    zoom = param.Number(4, bounds=(0, 24), step=1)
+
+viewer = MapViewer(sizing_mode="stretch_both")
+
+pn.Row(pn.Column(viewer.param, scroll=True), viewer, height=400).servable()
+```
+
+The `_model_names` attribute is an optional iterable or dictionary. It specifies which traits to synchronize to which parameters.
+
+## Create Reactive Values from the Traits of an ipywidget
```
maximlt, line 160: I think we use *reactive expressions* in the documentation elsewhere (panel, param).

pixi.toml
```diff
@@ -155,6 +155,7 @@ polars = "*"
 reacton = "*"
 scipy = "*"
 textual = "*"
+pydantic = "*"
```
maximlt, line 158: It's best to add a minimum pin, as the code is likely not to be compatible with all the versions of pydantic. And yes, most of the dependencies there have no pins, it doesn't mean that's the right way :)

maximlt (commented):
Amusingly, given the PR title and module name, I was expecting support for Python's stdlib `dataclass` :)

grepler:
Peanut gallery over here, I'm keen to see interoperability between Pydantic, but note that this hasn't made it's way onto newer roadmaps - is this still a possibility?

jbednar:
We're all still very interested, but haven't had time to work on it!

philippjfr:
>Peanut gallery over here, I'm keen to see interoperability between Pydantic

It is very much on the roadmap, but it will likely start with a pydantic interop in Param, which we can then build on in Panel.

Issue #6799, opened 2024-04-29
Make Panel Chat Interface callback not require user and instance args


ahuang11:
I find myself typing it excessively, but not using it:

```
def callback(contents, user, instance):
    return contents
```

I'd like to make the latter optional by perhaps inspecting the signature to see if it contains it.

```
def callback(contents):
    return contents
```

That way, it also conforms a bit more to `pn.bind(callback, contents=input.param.value)`.

PR #6917, opened 2024-06-13
Simplify ChatFeed/ChatInterface callback signature

ahuang11:
Closes https://github.com/holoviz/panel/issues/6799

I noticed many times, I don't use `user` and `instance` much, and based on feedback I received, it can be confusing (especially user), not to mention it can be tedious to include all three if it's not used.

This PR allows users to use one of these signatures:
- contents
- contents, user
- contents, user, instance

It's also easier to get started in my opinion and also conforms a bit more to `pn.bind(callback, contents=input.param.value)`.

e.g.

Before:

```python
import panel as pn
pn.extension()

def callback(contents, user, instance):
    return f"Echo: {contents}"

pn.chat.ChatInterface(callback=callback)
```

After:

```python
import panel as pn
pn.extension()

def callback(contents):
    return f"Echo: {contents}"

pn.chat.ChatInterface(callback=callback)
```


philippjfr (commented):
So should these be valid?

```python
def callback(*args):
    ...

def callback(contents, *args):
    ...
```

The first one will error, the second will have empty `args`. Not disagreeing just trying to understand what you would expect to happen here.

ahuang11:
Right those should be valid; I added support and tests for those as well.

philippjfr (approved):

Issue #5486, opened 2023-09-08
CodeEditor changes acts like `value_input` instead of `value`


ahuang11:
```
import panel as pn
pn.extension("codeeditor")

def print_code(contents):
    print(contents)

code = pn.widgets.CodeEditor()
pn.bind(print_code, contents=code.param.value, watch=True)
code
```

On every edit, the contents get printed, which is not ideal for writing code.

I was hoping for the same behavior as `pn.widgets.TextInput.param.value` not `param.value_input`

PR #6919, opened 2024-06-13
Add on_keyup and value_input for code editor

ahuang11:
Closes https://github.com/holoviz/panel/issues/5486

Although this is a breaking change, I think it's worth it because if you try to bind a callback that runs the code from value, it triggers on every keystroke... and as you might imagine, writing valid code often requires more than one character change.

```python
import panel as pn

pn.extension()


def callback(value):
    exec(value)

code = pn.widgets.CodeEditor(value="import this")
pn.bind(callback, code.param.value, watch=True)
code.show()
```

Previously, removing one character would trigger this error:

image

The workaround to this before is having a separate `pn.widgets.Button` to submit the code, but now, users can simply watch `value`, or if they want to watch on every keystroke `value_input`


MarcSkovMadsen:
Why not have a value_throttled instead?

ahuang11:
`value_input` is what other Bokeh widgets use:
https://panel.holoviz.org/reference/widgets/TextInput.html
https://panel.holoviz.org/reference/widgets/TextAreaInput.html

panel/models/ace.ts
```diff
@@ -68,7 +68,8 @@ export class AcePlotView extends HTMLBoxView {
     this._update_language()
     this._editor.setReadOnly(this.model.readonly)
     this._editor.setShowPrintMargin(this.model.print_margin)
-    this._editor.on("change", () => this._update_code_from_editor())
+    this._editor.on("blur", () => this._update_code_from_editor())
```
philippjfr, line 71: This seems wrong, I don't think Ace intentionally triggers a blur event, e.g. see https://github.com/ajaxorg/ace/issues/3323

philippjfr: I guess maybe it's fine but at minimum there should be a test checking that the event is triggered on unfocus and when pressing enter.

ahuang11: From my understanding, a blur event occurs when an HTML element loses focus. I have added a test on onfocus and removed on enter.

panel/tests/ui/widgets/test_codeeditor.py
```diff
@@ -0,0 +1,33 @@
...
+    ace_input = page.locator(".ace_content")
+    expect(ace_input).to_have_count(1)
+    ace_input.click()
+
+    page.keyboard.press("Enter")
+    page.keyboard.type('print("Hello Panel!")')
+    wait_until(
+        lambda: page.locator(".ace_content").inner_text()
+        == "print('Hello World!')\nprint(\"Hello Panel!\")"
+    )
```
philippjfr, line 26: expect(page.locator(".ace_content")).to_have_text("print('Hello World!')\nprint(\"Hello Panel!\")")

panel/models/ace.ts
philippjfr: Yeah I was primarily confused because `blur` on enter didn't make sense to me. I think there should be a way to sync the `value` without having to unfocus the editor but not sure what that should be. Suppose Ctrl+S would be somewhat standard.

philippjfr:
This does end up being a fairly major change in behavior. I agree with it but still.

panel/models/ace.ts
ahuang11: Yeah, I was basing it off TextInput/TextAreaInput behavior. I think Ctrl+S works or Enter like ChatAreaInput.

philippjfr: I'd be fine with enter actually, triggering an update on a newline seems reasonable.

panel/tests/ui/widgets/test_codeeditor.py
philippjfr: Should have warned you, I didn't actually test this but you should always use the `expect` when checking page elements. `wait_until` is for testing if the Panel component has updated.

panel/models/ace.ts
ahuang11: I'd be against Enter because it's code.

e.g.
```
def test():  # press enter here
```

This would trigger a syntax error. I'd be fine with like Ctrl+Enter though, but I do like Ctrl+S most.

ahuang11: Okay, Ctrl+S is actually used by the browser to save webpage, so I went with Ctrl/Cmd+Enter.

hoxbro:
Is there a way to have a flag that reverts back to the old version to smooth the transition?

ahuang11:
Or maybe it'd make sense to keep the current behavior with a flag that toggles value_input on (instead of off).

ahuang11:
> Is there a way to have a flag that reverts back to the old version to smooth the transition?

I think this is a good suggestion to prevent breaking changes so I have implemented a flag that de-syncs `value` from `value_input`, i.e. update value only on loss of focus / command hotkey.

ahuang11:
Test failure seems unrelated
> timeout waiting for merge ref

philippjfr (approved):

PR #6925, opened 2024-06-16
Move env variables to pixi.toml and simplify conda build script

hoxbro:


scripts/conda/build.sh
```diff
@@ -4,16 +4,12 @@ set -euxo pipefail
...
-    if [ -e "$file" ]; then
-        echo "dist folder already contains $(basename "$file"). Please delete it before running this script."
-        exit 1
-    fi
-done
-
 git diff --exit-code
 python -m build -w .
 
+VERSION=$(python -c "import $PACKAGE; print($PACKAGE._version.__version__)")
```
maximlt, line 10: Curious, why importing from `._version`?

hoxbro: Then I'm sure we get the version created by `setuptools_scm` / `hatch-vcs` from building the wheel.

philippjfr: Wondering if we should just make the `package.json` the source of truth here.

hoxbro: I don't know; it feels a bit weird having a mainly Python package being dictated by a Javascript file. 

We could set up a post-commit hook, which updates the `package.json` + `package-lock.json` when a tag is made and re-tag that commit.

philippjfr: I'm not super concerned with weird, I do think it's probably simplest and having a single source of truth is a good thing. That said, it's a larger change so I'm fine with this approach for now.

philippjfr (approved):

PR #6931, opened 2024-06-20
Detect WebGL support on BrowserInfo

philippjfr:
Detects webgl support, e.g. so we can toggle webgl rendering automatically in HoloViews.


panel/models/browser.ts
```diff
@@ -24,6 +24,15 @@ export class BrowserInfoView extends View {
...
       this.model.timezone_offset = timezone_offset
     }
+    try {
+      var canvas = document.createElement('canvas');
+      this.model.webgl = !!(
+        window.WebGLRenderingContext &&
+          (canvas.getContext('webgl') || canvas.getContext('experimental-webgl'))
+      );
+    } catch (e) {
+      this.model.webgl
```
philippjfr, line 34:
```suggestion
      this.model.webgl = false
```

panel/models/browser.ts
```diff
@@ -24,6 +24,15 @@ export class BrowserInfoView extends View {
     if (timezone_offset != null) {
       this.model.timezone_offset = timezone_offset
     }
+    try {
+      var canvas = document.createElement('canvas');
+      this.model.webgl = !!(
+        window.WebGLRenderingContext &&
+          (canvas.getContext('webgl') || canvas.getContext('experimental-webgl'))
```
philippjfr, line 31:
```suggestion
      var canvas = document.createElement("canvas");
      this.model.webgl = !!(
        window.WebGLRenderingContext &&
          (canvas.getContext("webgl") || canvas.getContext("experimental-webgl"))
```

panel/pane/holoviews.py
```diff
@@ -505,6 +505,8 @@ def _render(self, doc, comm, root):
                 Store.current_backend = loaded_backend
             backend = self.backend or Store.current_backend
             renderer = Store.renderers[backend]
+        if state.browser_info and 'webgl' in renderer.param:
+            renderer.webgl = state.browser_info.webgl
```
hoxbro, line 509: I don't think it is Panel job to do this.

hoxbro (commented):
An electron bug caused the issue with Jupyter Desktop. See the comments at https://github.com/bokeh/bokeh/issues/13665. It was fixed by deleting a specific folder. 

I don't know if this would solve the problem more than hide it. I could be wrong.

panel/pane/holoviews.py
philippjfr: Can see arguments both ways. I'd say this is the best place to do it though.

hoxbro, line 509:
```suggestion
```

hoxbro (approved):

Issue #6954, opened 2024-07-04
ChatFeed documentation wrong in regards of ChatFeed/ChatInterface 'callback_avatar' and ChatMessage 'value'


awesomebytes:
First, let me thank you for your wonderful work. I am testing your whole ecosystem for the first time and I am quite pleased and impressed with it. As I was trying to build a toy project with your Chat interfaces I found discrepancies between your latest available docs and the code.

#### ALL software version info
I cloned https://github.com/holoviz-topics/panel-chat-examples and used devcontainer environment. So all versions as specified in:
https://github.com/holoviz-topics/panel-chat-examples/blob/main/pyproject.toml

panel's version is: panel==1.5.0a8

My output of `pip freeze` in the devcontainer is:
<details>

aiohttp==3.9.5
aiosignal==1.3.1
annotated-types==0.7.0
anyio==4.4.0
attrs==23.2.0
beautifulsoup4==4.12.3
bleach==6.1.0
bokeh==3.5.0rc1
certifi==2024.7.4
charset-normalizer==3.3.2
click==8.1.7
contourpy==1.2.1
dataclasses-json==0.6.7
Deprecated==1.2.14
dirtyjson==1.0.8
diskcache==5.6.3
distro==1.9.0
frozenlist==1.4.1
fsspec==2024.6.1
gitdb==4.0.11
GitPython==3.1.41
greenlet==3.0.3
h11==0.14.0
httpcore==1.0.5
httpx==0.27.0
idna==3.7
Jinja2==3.1.4
joblib==1.4.2
jsonpatch==1.33
jsonpointer==3.0.0
langchain==0.2.6
langchain-community==0.2.6
langchain-core==0.2.11
langchain-text-splitters==0.2.2
langsmith==0.1.83
linkify-it-py==2.0.3
llama-cloud==0.0.6
llama-index==0.10.52
llama-index-agent-openai==0.2.7
llama-index-cli==0.1.12
llama-index-core==0.10.52.post1
llama-index-embeddings-openai==0.1.10
llama-index-indices-managed-llama-cloud==0.2.3
llama-index-legacy==0.9.48
llama-index-llms-openai==0.1.25
llama-index-multi-modal-llms-openai==0.1.6
llama-index-program-openai==0.1.6
llama-index-question-gen-openai==0.1.3
llama-index-readers-file==0.1.27
llama-index-readers-llama-parse==0.1.6
llama-parse==0.4.5
llama_cpp_python==0.2.81
[...]

PR #6959, opened 2024-07-08
Support `value` key for ChatFeed.callback generators

ahuang11:
Closes https://github.com/holoviz/panel/issues/6954

We have handling for `value` in `_build_message`, but not for generators.
```

    def _build_message(
        self,
        value: dict,
        user: str | None = None,
        avatar: str | bytes | BytesIO | None = None,
        **input_message_params
    ) -> ChatMessage | None:
        """
        Builds a ChatMessage from the value.
        """
        if "value" in value and "object" in value:
            raise ValueError(f"Cannot pass both 'value' and 'object' together; got {value!r}")
        elif "value" in value:
            value["object"] = value.pop("value")
        elif "object" not in value:
            raise ValueError(
                f"If 'value' is a dict, it must contain an 'object' key, "
                f"e.g. {{'object': 'Hello World'}}; got {value!r}"
            )
```

### DIFF
--- a/panel/chat/message.py
+++ b/panel/chat/message.py
@@ -632,6 +632,11 @@ class ChatMessage(Pane):
                 updates["user"] = user
             if avatar:
                 updates["avatar"] = avatar
+            if "value" in updates and "object" in updates:
+                raise ValueError(
+                    "Cannot set both 'value' and 'object' in the update dict."
+                )
+            updates["object"] = updates.pop("value", updates.get("object"))
         elif isinstance(value, ChatMessage):
             if user is not None or avatar is not None:
                 raise ValueError(
--- a/panel/tests/chat/test_feed.py
+++ b/panel/tests/chat/test_feed.py
@@ -699,6 +699,20 @@ class TestChatFeedCallback:
         assert chat_feed.objects[1].object == "Message"
         assert not chat_feed.objects[-1].show_activity_dot
 
+    @pytest.mark.parametrize("key", ["value", "object"])
+    def test_generator_dict(self, chat_feed, key):
+        def echo(contents, user, instance):
+            message = ""
+            for char in contents:
+                message += char
+                yield {key: message}
+
+        chat_feed.callback = echo
+        chat_feed.send("Message", respond=True)
+        wait_until(lambda: len(chat_feed.objects) == 2)
+        assert chat_feed.objects[1].object == "Message"
+        assert chat_feed.objects[-1].object == "Message"
+
     @pytest.mark.asyncio
     async def test_async_generator(self, chat_feed):
         async def async_gen(contents):
@@ -717,6 +731,25 @@ class TestChatFeedCallback:
         await async_wait_until(lambda: len(chat_feed.objects) == 2)
         assert chat_feed.objects[1].object == "Message"
         assert not chat_feed.objects[-1].show_activity_dot
+
+    @pytest.mark.asyncio
+    @pytest.mark.parametrize("key", ["value", "object"])
+    async def test_async_generator_dict(self, chat_feed, key):
+        async def async_gen(contents):
+            for char in contents:
+                yield char
+
+        async def echo(contents, user, instance):
+            message = ""
+            async for char in async_gen(contents):
+                message += char
+                yield {key: message}
+
+        chat_feed.callback = echo
+        chat_feed.send("Message", respond=True)
+        await async_wait_until(lambda: len(chat_feed.objects) == 2)
+        assert chat_feed.objects[1].object == "Message"
+        assert chat_feed.objects[-1].object == "Message"
 
     def test_placeholder_text_params(self, chat_feed):
         def echo(contents, user, instance):


awesomebytes:
Hello, as a user I find a bit unintuitive the name `value` for the message.

After reading the definition of ChatMessage, and learning that it is originally called `object` because it accepts anything that can be rendered by panel (which sounds great, btw)... maybe  just change the docs to reflect that you must pass `object` instead of the name `value`?

I understand this would be an API change, and that naming is one of the hardest things, but if this would be a good moment to consider changing the name to something that reflects a bit better what that field is, I would propose to change it now or start a deprecation process now (whichever way this project works).

And thanks for looking into it!

ahuang11:
Thanks for the feedback! A bit of a backstory; when we started work on this, all of these Chat components were widgets, thus was part of the reason why `value` was kept. `ChatMessage` was eventually refactored into a `pane`, thus the arg is named `object` now.

A more convincing argument though is that `ChatInterface` includes a `ChatAreaInput` widget to allow users to input `value`s and serialize into `ChatMessage`.

Based on that, I believe we should keep `value` because the programmatic APIs of `ChatAreaInput` widget (`ChatInterface.send` `ChatInterface.stream`) both accepts `value`, and `return` and `yield` from functions/generators should reflect `send` and `stream` signature in my opinion.

awesomebytes:
That does make a lot of sense!


Let me hijack this thread for something that will most probably be classified as offtopic, but oh well, I don't know anyone else using the library to ask:

Would this be a good place to also bring up that a, probably typical, use case for the `ChatAreaInput` would be to accept voice/speech input to inject text into the field?
I've been playing around with doing this, and, man I've had issues. I wanted to use the SpeechToText widget from panel (so the browser would do the ugly backend stuff, at least for PoC) and it hasn't been easy!

I ended up doing some monkeypatching:
```python
def value_getter(self):
    return getattr(self, "_value", "")

def value_setter(self, new_value):
    self._value = new_value

pn.Row.value = property(value_getter, value_setter)
```

This is because the ChatInterface tries to get `.value` from the input widget.

To patch the Row class that is currently containing my TextInput + SpeechToText:
```python
class SpeechTextInput(param.Parameterized):
    value = param.String(default="", doc="The current value of the text input")
    placeholder = param.String(
        default="Enter text or click to speak", doc="Placeholder text for the input"
    )
    speech_button_width = param.Integer(default=50, doc="Width of the speech button")
    speech_button_type = param.String(
        default="default", doc="Button type for the speech button"
    )
    # Add common Panel widget parameters
    name = param.String(default="")
    sizing_mode = param.ObjectSelector(
        default=None,
        objects=[
            None,
            "fixed",
            "stretch_width",
            "stretch_height",
            "stretch_both",
            "scale_width",
            "scale_height",
            "scale_both",
        ],
    )
    width = param.Integer(default=None)
    height = param.Integer(default=None)
    margin = param.Parameter(default=(5, 5, 5, 5))
    css_classes = param.List(default=[])

    def __init__(self, **params):
        super().__init__(**params)
        self._text_input = pn.widgets.TextInput(
            value=self.value,
            placeholder=self.placeholder,
            sizing_mode=self.sizing_mode,
            width=self.width,
            height=self.height,
            margin=self.margin,
            css_classes=self.css_classes,
        )
        self._speech_to_text = pn.widgets.SpeechToText(
            button_type=self.speech_button_type,
            width=self.speech_button_width,
            height=self.height,
        )
        self._text_input.param.watch(self._update_value, "value")
        self._speech_to_text.param.watch(self._update_from_speech, "value")
        self.param.watch(self._update_layout_value, "value")

    def _update_value(self, event):
        self.value = event.new

    def _update_from_speech(self, event):
        if event.new:
            self.value = event.new
            self._text_input.value = event.new
            self._speech_to_text.value = ""  # Reset the speech input

    def _update_layout_value(self, event):
        if hasattr(self, "_layout"):
            self._layout.value = event.new

    @param.depends(
        "value",
        "placeholder",
        "speech_button_width",
        "speech_button_type",
        "sizing_mode",
        "width",
        "height",
        "margin",
        "css_classes",
    )
    def view(self):
        self._layout = pn.Row(
            self._text_input,
            self._speech_to_text,
            sizing_mode=self.sizing_mode,
            width=self.width,
            height=self.height,
            margin=self.margin,
            css_classes=self.css_classes,
        )
        self._layout.value = self.value
        return self._layout

    def __panel__(self):
        return self.view()
```
There's probably a fair bit of boilerplate that's useless there, but I built it as I was experimenting and finding errors (like sizing_mode being needed).


I'd like to note that:
```python
# subclassing anything like TextInput didn't work
# subclassing Row didn't work
# Trying to use ChatTextAreaInput as an example of an inherited class from TextInput didn't work
```

With my current solution, emptying the text field after clicking send does not work... oh well. But at least I can press the speech icon and get the recognized text into the textinput (and then I need to hit send button, and then delete by hand from the field).

I must say I've fallen into this rabbit hole without knowing very well what I was doing. Well, what I am doing.

P.S.: If you made it this far, I really really thank you.

ahuang11:
I think SpeechToText needs to be revamped; it hasn't been working for me.

I think your solution can be simplified!

https://github.com/user-attachments/assets/84d9f858-e21c-4a7e-9bbc-f815975a924b

> Let me hijack this thread for something that will most probably be classified as offtopic, but oh well, I don't know anyone else using the library to ask:
We have a community discourse channel so let's move this off-topic discussion there after this final message. https://discourse.holoviz.org/t/how-to-record-and-transcribe-text-and-send-to-chatinterface/7478

```python
import panel as pn

from marvin.ai.audio import transcribe
from marvin.audio import record_phrase
pn.extension()

def recognize_speech(instance, event):
    with instance.active_widget.param.update(loading=True):
        audio = record_phrase()
        transcription = transcribe(audio)
        try:
            instance.active_widget.value = transcription
        except sr.RequestError as e:
            instance.stream("Could not request results; {0}".format(e), user="System")
        except sr.UnknownValueError:
            instance.stream("Unknown error occurred", user="System")

    messages = instance.serialize()
        model="gpt-3.5-turbo",
        messages=messages,
        stream=True,
    )
    message = ""
    async for chunk in response:
        if part is not None:
            message += part
            yield message

aclient = AsyncOpenAI()
chat = pn.chat.ChatInterface(
    callback=callback,
    button_properties={"speak": {"callback": recognize_speech, "icon": "microphone"}},
    show_rerun=False,
    show_undo=False,
    show_clear=False,
)
chat.servable()

```

philippjfr (approved):

PR #6979, opened 2024-07-12
Allow providing DataFrame as part of vega(-lite) spec

philippjfr:
Useful when generating vega(-lite) specs directly, e.g. in Lumen.ai.

### DIFF
--- a/panel/tests/pane/test_vega.py
+++ b/panel/tests/pane/test_vega.py
@@ -11,6 +11,7 @@ except Exception:
 altair_available = pytest.mark.skipif(alt is None, reason="requires altair")
 
 import numpy as np
+import pandas as pd
 
 import panel as pn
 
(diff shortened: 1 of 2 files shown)


ahuang11 (approved):

panel/tests/pane/test_vega.py
```diff
@@ -194,8 +207,9 @@ def test_get_vega_pane_type_from_dict():
     assert PaneBase.get_pane_type(vega_example) is Vega
 
 
-def test_vega_pane(document, comm):
-    pane = pn.panel(vega_example)
+@pytest.mark.parametrize('example', [vega_example, vega_df_example])
+def test_vega_pane(document, comm, example):
```
philippjfr, line 211: Ideally it would confirm that the dataframe is correctly converted to a `ColumnDataSource`.

philippjfr: Also, thanks!

philippjfr: Nvm, just realized that's already what this test is doing.

PR #6982, opened 2024-07-14
Add missing validation on slider

### LINKED ISSUE #6963: Can set FloatSlider values beyond start/end
```python
import panel as pn

pn.extension(throttled=True)
mag_slider = pn.widgets.FloatSlider(name='Minimum Magnitude', start=0, end=9, value=6)
mag_slider
mag_slider.value
mag_slider.value = 100  # should error?
```

MarcSkovMadsen:
Closes #6963

### DIFF
--- a/panel/tests/widgets/test_slider.py
+++ b/panel/tests/widgets/test_slider.py
@@ -774,3 +774,19 @@ def test_editable_fixed_nosoftbounds_fixed_end(editableslider):
     slider = editableslider(fixed_end=fixed_end, end=end, step=step)
     assert slider.start == end - step
     assert slider.end == end
+
+@pytest.mark.parametrize(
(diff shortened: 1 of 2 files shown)


MarcSkovMadsen:
I've added the fix for #6963 including tests. But as you can see this makes a lot of other tests fail.

The reason is that its actually a feature of the editable sliders that you can set a value outside of the start and end!

image

I would need some guidance on what behavior we would like. I see different options:

- Not wanting to validate values outside bounds.
- Implementing different validations for non-editable and editable sliders
- Changing the api of editable sliders such that values outside start and end are not supported.

philippjfr:
Superseded by https://github.com/holoviz/panel/pull/6982

PR #6994, opened 2024-07-16
Bump bokeh version to 3.5.0

philippjfr:
Now that it's released we can bump the pin.


pixi.toml
```diff
@@ -37,7 +38,7 @@ requests = "*"
 tqdm = ">=4.48.0"
 typing_extensions = "*"
 # Recommended
-holoviews = ">=1.16.0"
+holoviews = ">=1.18.0"
```
philippjfr, line 41:
```suggestion
holoviews = ">=1.16.0"
```

Issue #6995, opened 2024-07-16
PyScript reference example no longer working


MarcSkovMadsen:
## Does not work with historical versions of Pyscript because they no longer exist

If I try the PyScript reference example at https://panel.holoviz.org/how_to/wasm/standalone.html#pyscript, It no longer works because the pyscript .js and .css files no longer exist.

![image](https://github.com/user-attachments/assets/bf264a2c-9d03-44cc-910c-2183c522f031)

For example [https://pyscript.net/releases/2024.2.1/pyscript.js](https://pyscript.net/releases/2024.2.1/pyscript.js) no longer exists.

Might be related to https://github.com/holoviz/panel/issues/6991 and https://github.com/pyscript/pyscript/issues/2118

MarcSkovMadsen:
If I try to use the latest versions from https://docs.pyscript.net/2024.7.1/beginning-pyscript/ nothing is shown.

![image](https://github.com/user-attachments/assets/9c44a44f-d731-4e0b-ab85-50f8043f8c68)

```html
<!DOCTYPE html>
<html>
  <head>
    <script type="text/javascript" src="https://cdn.bokeh.org/bokeh/release/bokeh-3.4.1.js"></script>
    <script type="text/javascript" src="https://cdn.bokeh.org/bokeh/release/bokeh-widgets-3.4.1.min.js"></script>
    <script type="text/javascript" src="https://cdn.bokeh.org/bokeh/release/bokeh-tables-3.4.1.min.js"></script>
    <script type="text/javascript" src="https://cdn.jsdelivr.net/npm/@holoviz/panel@1.4.4/dist/panel.min.js"></script>

    <link rel="stylesheet" href="https://pyscript.net/releases/2024.7.1/core.css">  
    <script type="module" src="https://pyscript.net/releases/2024.7.1/core.js"></script>
  </head>
  <body>
    <py-config>
       packages = [
          "https://cdn.holoviz.org/panel/1.4.4/dist/wheels/bokeh-3.4.1-py3-none-any.whl",
          "https://cdn.holoviz.org/panel/1.4.4/dist/wheels/panel-1.4.4-py3-none-any.whl"
       ]
    </py-config>
    <div id="simple_app"></div>
    <py-script>
      import panel as pn

      pn.extension(sizing_mode="stretch_width")

      slider = pn.widgets.FloatSlider(start=0, end=10, name='Amplitude')

      def callback(new):
         return f'Amplitude is: {new}'

[...]

PR #6998, opened 2024-07-17
Fix issue detecting PyScript worker

philippjfr:
This fixes an issue that @antocuni and @WebReflection heroically tracked down. Specifically the issue is that Panel makes various assumptions on which JS APIs it can use in a pyodide context based on whether `js.document` is defined. These assumptions do not hold when we are in a PyScript worker because there [`coincident`](https://github.com/WebReflection/coincident) is used to automatically proxy various JS APIs without having to rely on the custom JS <-> JS Worker message-passing implementation that Panel ships.  The logic was incorrectly setting `IN_WORKER=True` when we were in a PyScript worker, which in turn causes Panel to make incorrect assumptions on how to pass messages and access DOM APIs.

Fixes https://github.com/holoviz/panel/issues/6995

### DIFF
--- a/panel/io/convert.py
+++ b/panel/io/convert.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import base64
 import concurrent.futures
 import dataclasses
 import json
@@ -42,8 +43,9 @@ BOKEH_VERSION = base_version(bokeh.__version__)
 PY_VERSION = base_version(__version__)
 PYODIDE_VERSION = 'v0.25.0'
 PYSCRIPT_VERSION = '2024.2.1'
-PANEL_LOCAL_WHL = DIST_DIR / 'wheels' / f'panel-{__version__.replace("-dirty", "")}-py3-none-any.whl'
-BOKEH_LOCAL_WHL = DIST_DIR / 'wheels' / f'bokeh-{BOKEH_VERSION}-py3-none-any.whl'
+WHL_PATH = DIST_DIR / 'wheels'
+PANEL_LOCAL_WHL = WHL_PATH / f'panel-{__version__.replace("-dirty", "")}-py3-none-any.whl'
+BOKEH_LOCAL_WHL = WHL_PATH / f'bokeh-{BOKEH_VERSION}-py3-none-any.whl'
 PANEL_CDN_WHL = f'{CDN_DIST}wheels/panel-{PY_VERSION}-py3-none-any.whl'
 BOKEH_CDN_WHL = f'{CDN_ROOT}wheels/bokeh-{BOKEH_VERSION}-py3-none-any.whl'
 PYODIDE_URL = f'https://cdn.jsdelivr.net/pyodide/{PYODIDE_VERSION}/full/pyodide.js'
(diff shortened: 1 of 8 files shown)


panel/io/convert.py
```diff
@@ -342,7 +344,14 @@ def script_to_html(
...
         if inline:
-            loading_base = (DIST_DIR / "css" / "loading.css").read_text(encoding='utf-8')
+            svg_name = f'{config.loading_spinner}_spinner.svg'
+            svg_b64 = base64.b64encode((DIST_DIR / 'assets' / svg_name).read_bytes()).decode('utf-8')
+            loading_base = (
+                DIST_DIR / "css" / "loading.css"
+            ).read_text(encoding='utf-8').replace(
+                f'../assets/{svg_name}', f'data:image/svg+xml;base64,{svg_b64}'
+            )
+            print(loading_base)
```
philippjfr, line 354:
```suggestion
```

WebReflection:
gotta be honest ... [this line](https://github.com/holoviz/panel/blob/1ec22c234c2e0bcac9bc88a2c0f78e482f2e4494/panel/io/pyodide.py#L64) doesn't fully reason well to me ... if you are targeting very old version of PyScript Next you are patching your env the same ... by adding both `js.document` and `js.window` to your Worker global context.

Accordingly, if it works for the previous `try` attempt where you then set `_IN_WORKER = False` if no issues are found, why wouldn't that be the case also for ancient versions of PyScript next?

Or to rephrase ... why isn't the first `try` enough? If people are stuck in very old PyScript next they are also stuck in very old version of Pyodide thta might not necessarily bring in that very old version of *panel* neither ... do we need to support people stuck with outdated or deprecated PyScript versions there? Honest question though, maybe you had use cases before, I just think we should expect people being on 6 months old project or later at this point, and that `js.document` patch comes from very long time ago.

Thanks in advance for eventually expanding on that.

philippjfr:
>gotta be honest ... this line doesn't fully reason well to me ... if you are targeting very old version of PyScript Next you are patching your env the same ... by adding both js.document and js.window to your Worker global context.

Agreed, that was definitely not correct either.

> If people are stuck in very old PyScript next they are also stuck in very old version of Pyodide thta might not necessarily bring in that very old version of panel neither 

We strongly recommend people install Panel with the custom wheels we ship in our CDN (they are optimized for size), so the assumption that a particular old version of Pyodide will only load old versions of Panel doesn't necessarily hold.

I generally agree that we can reasonably expect people to be on more recent versions of Pyodide/PyScript but with the fixed logic I don't think there's any harm in keeping this.

WebReflection:
> with the fixed logic I don't think there's any harm in keeping this.

fair enough ... as long as you are testing/proving that logic still works ... which would surprise me, as the previous `try` fixes things in a different way and I expect that outer exception to produce broken logic you already fixed for the first try.

Those lines require more testing, more code, more everything, but it's your project, I am just saying "*we encourage our users to always use our latest offer as the project is still relatively young and while robust/stable enough these days ... it's indeed in these days that we reached such stability*" ... if that makes sense.

philippjfr:
Part of the problem here was that recent changes to our build infrastructure caused all the PyScript tests to be skipped. I'm working on re-enabling the tests now and will have them error out if they can't be run.

WebReflection:
All I am saying is that if you could just target latest PyScript and be sure stuff works we'll be all good, because we won't back-port fixes to previous releases and all we care about is that current stable channel remains stable.

That should be less work for your CI too ... and to e honest, we have people with problems from the 2yo PyScript version, those that jumped into Next already, are usually fine and well understands they might just update the version and have fixes all over the place out of the box.

PR #7013, opened 2024-07-25
Add TimePicker from Bokeh

ahuang11:
Adds TimePicker

image

The times mismatch due to timezone:
https://github.com/bokeh/bokeh/issues/13768


ahuang11:
I tried correcting the timezone by adding back the difference, but I keep getting assertion error and I have no idea why. I thought it was the type initially, but whatever I change it to, number/string, it still crashes.

image

panel/models/time_picker.ts
```diff
@@ -0,0 +1,52 @@
...
+
+        console.log(date);
+
+        // get type of date
+        console.log(this.model.value)
+        console.log("Type of Date:", typeof this.model.value);
+
+        console.log("Converted Date:", date.toISOString());
+        console.log("Converted Time:", date.getTime());
+        this.model.value = 72000000
```
ahuang11, line 23: This breaks things for some reason.

panel/models/time_picker.ts
```diff
@@ -0,0 +1,52 @@
...
+
+
+export class TimePickerView extends BkTimePickerView {
+    declare model: TimePicker
+
+    // Override the render method or any method where time rendering is done
+    override render(): void {
+        // Fix time zone difference; add back 8 hours
+        const date = this.model.value ? new Date(this.model.value) : new Date();
+        date.setHours(date.getHours() + 8);
```
hoxbro, line 13: This is likely only fixing the case for people in your time zone (if you weren't aware) 

If you find a fix, I think we should upstream this to Bokeh instead of having the model in Panel.

mattpap: > If you find a fix, I think we should upstream this to Bokeh instead of having the model in Panel.

That would be the optimal thing to do.

panel/models/time_picker.ts
mattpap: I would start with using development resources, so that we can get a proper stack trace. Either use `BOKEH_DEV=true` on the command line or panel's equivalent. However, this seems to be an assertion originating from third party code (`flatpickr`).

panel/models/time_picker.ts
ahuang11: Yes I am (time-zone) aware :D 

Was just debugging, and I think the issue was that putting this in render, the flatpickr widget wasn't created yet, so needed to put in after_render.

However, better option is adjust defaultDate
https://github.com/bokeh/bokeh/blob/61b87ac7e47df2d920635b4f13a89e00062f07f1/bokehjs/src/lib/models/widgets/time_picker.ts#L63-L65

Yes, the plan is to first implement in Panel since faster release cycle, then upstream it Bokeh.

ahuang11:
Okay I think it's ready

https://github.com/user-attachments/assets/67fef6a7-9942-4c9f-ad23-0d276e40bd8e

philippjfr:
Maybe worth adding a test for a different timezone: https://github.com/microsoft/playwright/issues/26996

philippjfr (approved):

Issue #7015, opened 2024-07-26
pyodide-worker raising error with pyodide 0.26.1


MarcSkovMadsen:
I was trying to upgrade the pyodide and wasm code + documentation to latest version. I found that `--pyodide-worker` will not work with latest version of pyodide 0.26.1 and panel 1.5.0b2

## Reproduce

- `pip install panel==1.5.0-b.2`
- Follow the instructions in the reference example https://panel.holoviz.org/how_to/wasm/convert.html
  -  after the conversion manually update 0.25.0 to 0.26.1 in the script.js file.

See the error in the browser console

![image](https://github.com/user-attachments/assets/d7a99e4a-8767-42ed-aeed-02219ad68e9a)

```bash
PythonError: Traceback (most recent call last):
  File "/lib/python312.zip/_pyodide/_base.py", line 574, in eval_code_async
    await CodeRunner(
  File "/lib/python312.zip/_pyodide/_base.py", line 394, in run_async
    coroutine = eval(self.code, globals, locals)
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/lib/python312.zip/contextlib.py", line 158, in __exit__
    self.gen.throw(value)
  File "/lib/python3.12/site-packages/panel/util/parameters.py", line 51, in edit_readonly
    yield
  File "<exec>", line 8, in <module>
  File "/lib/python3.12/site-packages/param/parameterized.py", line 2319, in update
    restore = dict(self_._update(arg, **kwargs))
                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/lib/python3.12/site-packages/param/parameterized.py", line 2352, in _update
    self_._batch_call_watchers()
  File "/lib/python312.zip/contextlib.py", line 158, in __exit__
    self.gen.throw(value)
  File "/lib/python3.12/site-packages/param/parameterized.py", line 284, in _batch_call_watchers
    yield
  File "/lib/python3.12/site-packages/param/parameterized.py", line 2546, in _batch_call_watchers
    self_._execute_watcher(watcher, events)
  File "/lib/python3.12/site-packages/param/parameterized.py", line 2506, in _execute_watcher
    watcher.fn(*args, **kwargs)
  File "/lib/python3.12/site-packages/panel/reactive.py", line 412, in _param_change
    self._apply_update(named_events, properties, model, ref)
  File "/lib/python312.zip/contextlib.py", line 158, in __exit__
    self.gen.throw(value)
[...]

PR #7016, opened 2024-07-26
Update Pyscript

MarcSkovMadsen:
I would like to bring Pyscript up to date.

I have tested manually that the updates works.

PR #7017, opened 2024-07-26
Describe usage of pyscript editor

MarcSkovMadsen:
This updates to pyodide 0.26.1 and adds a description of the usage of the pyscript editor. Starts from #7016.

I took the opportunity to make the guides a bit more `how-to` like to align with Diataxis.

# Todo:

- [x] Philipp? Fix https://github.com/holoviz/panel/issues/7015.


Coderambling:
Looks great! 2 suggestions:

-Maybe introduce some paragraph spacing ( empty lines) in text blocks.

For example, the text block at the top is quite long. Without spacing the text becomes monolithic and harder to follow.

This is true when viewed in desktop version, but especially in mobile.

-Regarding the below: maybe provide a short explanation and / or a link to a a way / tools to check if libraries have been compiled for Pyodide, or are available as pure-Python wheels from PyPI?

...that the libraries have been compiled for Pyodide or are available as pure-Python wheels from PyPI.

So something like:

Ro check for a compiled version, go to [the Pyodide packages list](https://pyodide.org/en/stable/usage/packages-in-pyodide.html)

Or run command x

- To check if a wheel is available, look for a .whl file in the files section of the lbrary on pypi.org, or run the following command:

`pip download --no-deps --only-binary :all: <package-name>`

philippjfr (approved):

PR #7022, opened 2024-07-27
Allow using Card as default layout for ChatStep

ahuang11:
https://github.com/user-attachments/assets/7393bb36-a00d-4a01-896d-24e1446406c9

Collapsed:
image

Expanded
image

### DIFF
--- a/panel/chat/feed.py
+++ b/panel/chat/feed.py
@@ -26,7 +26,7 @@ from ..layout import Column, Feed, ListPanel
 from ..layout.card import Card
 from ..layout.spacer import VSpacer
 from ..pane.image import SVG, ImageBase
-from ..pane.markup import Markdown
+from ..pane.markup import HTML, Markdown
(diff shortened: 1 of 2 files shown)


panel/chat/feed.py
```diff
@@ -724,9 +726,15 @@ def add_step(
         avatar : str | bytes | BytesIO | None
             The avatar to use; overrides the message's avatar if provided.
             Will default to the avatar parameter. Only applicable if steps is "new".
-        steps_column : Column | None
+        steps_layout : Column | None
             An existing Column of steps to stream to, if None is provided
```
MarcSkovMadsen, line 730: As its now a layout, I believe the text explanation needs to change from Column to something more general.

ahuang11, line 730:
```suggestion
            An existing layout of steps to stream to, if None is provided
```

philippjfr (approved):

PR #7023, opened 2024-07-27
Enhancement/document vega dataframe support

MarcSkovMadsen:
Documents #6979 such that users will know about it.


MarcSkovMadsen:
Ups. Included files that should not be included

PR #7029, opened 2024-07-27
Document esm panes

MarcSkovMadsen:
For consistency we should also document how to create Panes with ESM.

## Todo

- [x] Convert reactive_html_panes docs to JSComponents.
- [x] Add React versions
- [x] Add AnyWidget versions
- [x] Philipp? Fix issues.


doc/how_to/custom_components/esm/panes.md
```diff
@@ -0,0 +1,178 @@
...
+    if (chart) {
+        chart.destroy();
+    }
+    chart = new Chart(canvasEl.getContext('2d'), model.object);
+}
+
+export function render({ model }) {
+    const canvasEl = document.createElement('canvas');
+    // The chart will not render before after the layout has been created
+    requestAnimationFrame(() => createChart(canvasEl, model));
```
MarcSkovMadsen, line 33: The `ReactiveHTML`  creates the chart in `after_layout`. There is nothing similar here. Postponing the creating via requestAnimationFrame seems to do the trick. Without it Chart.js won't render.

What is the correct way to do this?

doc/how_to/custom_components/esm/panes.md
```diff
@@ -0,0 +1,178 @@
...
+    object = param.Dict()
+
+    _esm = """
+import { Chart } from "https://esm.sh/chart.js/auto"
+
+let chart = null;
+
+function createChart(canvasEl, model) {
+    if (chart) {
+        chart.destroy();
```
MarcSkovMadsen, line 25: What is the correct and recommended way to clean up before re-rendering?

doc/how_to/custom_components/esm/panes.md
```diff
@@ -0,0 +1,178 @@
...
+    });
+    cy.style().resetToDefault().append(model.style).update()
+
+    model.on('object', () => {cy.json({elements: model.object});cy.resize().fit()})
+    model.on('layout', () => {cy.layout({name: model.layout}).run()})
+    model.on('zoom', () => {cy.zoom(model.zoom)})
+    model.on('pan', () => {cy.pan(model.pan)})
+    model.on('style', () => {cy.style().resetToDefault().append(model.style).update()})
+
+    window.addEventListener('resize', function(event){
```
MarcSkovMadsen, line 138: What is the correct and recommended way to trigger responsive resizing?

doc/how_to/custom_components/esm/panes.md
```diff
@@ -0,0 +1,178 @@
...
+
+    _esm = """
+import { default as cytoscape} from "https://esm.sh/cytoscape"
+let cy = null;
+
+export function render({ model, el }) {
+    if (cy) { cy.destroy() }
+
+    cy = cytoscape({
+        container: el,
```
MarcSkovMadsen, line 120: Which element would we recommend rendering cytospace into. The `el` element or a new element? Why?

doc/how_to/custom_components/esm/panes.md
```diff
@@ -0,0 +1,178 @@
...
+            ],
+        },
+        "options": {
+            "responsive": True,
+            "maintainAspectRatio": False,
+        },
+    }
+
+
+chart_type = pn.widgets.RadioBoxGroup(
```
MarcSkovMadsen, line 64: When I click the radiobutton an exception is logged in the browser console. I can't see its related to my code.

Is it a Panel esm bug @philippjfr ?

image

doc/how_to/custom_components/esm/panes.md
```diff
@@ -0,0 +1,178 @@
...
+    _esm = """
+import { default as cytoscape} from "https://esm.sh/cytoscape"
+let cy = null;
+
+export function render({ model, el }) {
+    if (cy) { cy.destroy() }
+
+    cy = cytoscape({
+        container: el,
+        layout: {name: model.layout},
```
MarcSkovMadsen, line 121: We should get the warning `A Cytoscape container has style position:static and so can not use UI extensions properly`  fixed

image

doc/how_to/custom_components/esm/panes.md
philippjfr: `model.on('resize', ...)` or more explicitly `model.on('lifecycle:resize', ...)`

doc/how_to/custom_components/esm/panes.md
philippjfr: The `el`, there shouldn't be a need to nest but one reason you might want to nest it if you want to set specific layout options on the `el`.

doc/how_to/custom_components/esm/panes.md
philippjfr: `model.on('after_render', ...)` or more explicitly `model.on('lifecycle:after_render', ...)`

MarcSkovMadsen: works in the example

doc/how_to/custom_components/esm/panes.md
MarcSkovMadsen: The answer must be to use `model.on('remove', removeChart);`

doc/how_to/custom_components/esm/panes.md
MarcSkovMadsen: technically `model.on('after_resize', function(event){...` works. But the redraw is not very smooth

https://github.com/user-attachments/assets/ab286890-3613-4dd4-aae4-6837c27257b7

Instead `window.addEventListener('resize', function(event){...` works much smoother

https://github.com/user-attachments/assets/d90daa44-1896-4143-8a13-f4fa8a8fc1d0

I would recommend using the last option. Is there any way to get the first option working properly @philippjfr ?

doc/how_to/custom_components/esm/panes.md
MarcSkovMadsen: Fixed. by creating a div with position relative. But then the div needs to be created `after_render` which makes the whole example a bit more convoluted.

examples/reference/custom_components/JSComponent.md
```diff
@@ -73,13 +73,16 @@ The following signatures are valid when listening to change events:
...
-- `.on('change:<parameter>', callback)`: The `change:` prefix allows disambiguating change events from lifecycle hooks should a parameter name and lifecycle hook overlap.
+
+The `change:` prefix allows disambiguating change events from lifecycle hooks should a parameter name and lifecycle hook overlap.
 
 #### Lifecycle Hooks
 
-- `after_render`: Called once after the component has been fully rendered.
-- `after_resize`: Called after the component has been resized.
-- `remove`: Called when the component view is being removed from the DOM.
+- .on(`after_render`, callback): Called once after the component has been fully rendered.
```
MarcSkovMadsen, line 81: Are you sure you want to call it `after_render` and not `after_layout`? I mixed them up because its `after_layout` when working directly with bokeh models. I.e. it might be mental overhead at least for some users with different names.

doc/how_to/custom_components/esm/panes.md
```diff
@@ -0,0 +1,311 @@
...
+    data = param.List(doc="Use to send node's data/attributes to Cytoscape")
+
+    selected_nodes = param.List()
+    selected_edges = param.List()
+
+    _esm = """
+import CytoscapeComponent from 'https://esm.sh/react-cytoscapejs';
+
+export function render({ model }) {
+    function configure(cy){
```
MarcSkovMadsen, line 242: I'm in doubt whether this is the correct way to do this or not.

I would have expected the React render function to be rerun whenever a model property changed. I.e. I would not have to configure `model.on`. I would have have expected to be able to set event handlers for `select` and `unselect` events directly on the react `CytoscapeComponent`. But don't know how to do it.

We can leave it as is. But maybe there is a better way that we should be showing of instead.

MarcSkovMadsen: I would have have expected to be able to set event handlers for `select` and `unselect` events directly on the react `CytoscapeComponent`. But don't know how to do it. Maybe its called `tapNode`, `tapNodeData`, `tapEdge`, `tapEdgeData`. See https://github.com/plotly/dash-cytoscape/blob/main/dash_cytoscape/Cytoscape.py.

After searching around I can see others are doing it similar to what we are. For example in https://codesandbox.io/s/react-cytoscape-cl1dxl?file=/src/App.js:5891-5907. Search for `setSelectedNodes` in the example.

doc/how_to/custom_components/esm/panes.md
```diff
@@ -0,0 +1,426 @@
...
+::: `AnyWidgetComponent`
+
+```{pyodide}
+import param
+import panel as pn
+
+from panel.custom import AnyWidgetComponent
+
+
+class CytoscapeAnyWidget(AnyWidgetComponent):
```
MarcSkovMadsen, line 320: After converting the Cytoscape example from `JSComponent` to `AnyWidgetComponent` I do understand why you would like to have the `JSComponent` API. The `AnyWidget` api is just much more verbose.

One thing I got hit by is the missing support for `_css` attribute. It took me some time to remember its not supported on `AnyWidgetComponent` and I have to use `stylesheets` attribute.

doc/how_to/custom_components/esm/panes.md
```diff
@@ -0,0 +1,426 @@
...
+
+This example will show you the basics of creating a [ChartJS](https://www.chartjs.org/docs/latest/) pane.
+
+```{pyodide}
+import panel as pn
+import param
+from panel.custom import JSComponent
+
+
+class ChartJSComponent(JSComponent):
```
MarcSkovMadsen, line 15: When trying to convert to `ReactComponent` there is an unexpected error in the console. Its unexpected because I don't use `useRef`. **Is it a bug in Panels implementation?**

```python
import panel as pn
import param
from panel.custom import ReactComponent


class ChartJSComponent(ReactComponent):
    object = param.Dict()

    _esm = """
import {
  BarController,
  LinearScale,
  BarElement,
  TimeScale,
  Tooltip,
} from 'https://esm.sh/chart.js';
import 'https://esm.sh/chartjs-adapter-date-fns';
import { enUS } from 'https://esm.sh/date-fns/locale';
import { ReactChart } from "https://esm.sh/chartjs-react"

ReactChart.register(BarController, LinearScale, BarElement, TimeScale, Tooltip);
export function render({ model }) {
    return (
        <ReactChart type={ model.object.type } data={ model.object.data } />
    );
};
"""


def data(chart_type="line"):
    return {
        "type": chart_type,
        "data": {
            "labels": ["January", "February", "March", "April", "May", "June", "July"],
            "datasets": [
                {
                    "label": "Data",
                    "backgroundColor": "rgb(255, 99, 132)",
                    "borderColor": "rgb(255, 99, 132)",
                    "data": [0, 10, 5, 2, 20, 30, 45],
                }
            ],
        },
        "options": {
            "responsive": True,
            "maintainAspectRatio": False,
        },
    }


chart_type = pn.widgets.RadioBoxGroup(
    name="Chart Type", options=["bar", "line"], inline=True
)
chart = ChartJSComponent(
    object=pn.bind(data, chart_type), height=400, sizing_mode="stretch_width"
)
pn.Column(chart_type, chart).servable()
```

image

doc/how_to/custom_components/esm/panes.md
MarcSkovMadsen: I see the same exception with the `ReactComponent` example.

MarcSkovMadsen: I see the same exception with the `AnyWidgetComponent` example.

doc/how_to/custom_components/esm/panes.md
philippjfr: In my current testing `after_resize` works just the same. I think we should rename it to `model.on('resize', ...)` though.

doc/how_to/custom_components/esm/panes.md
philippjfr: No, you just had to make sure to specify compatible versions of Chart.js and the React library.

doc/how_to/custom_components/esm/panes.md
philippjfr: Fixed.

doc/how_to/custom_components/esm/panes.md
philippjfr: Don't see this anymore.

doc/how_to/custom_components/esm/panes.md
MarcSkovMadsen: How would I know what those are?

philippjfr: By reading the docs issues for Chart.js and the React wrapper, really nothing at all to do with Panel.

MarcSkovMadsen: Ok. Its probably a bit more general than Chart.js. I believe I've seen it when using other libraries too.

philippjfr: Yes, the JS world is finicky particularly in regard to React version compatibility but that's not something we can help them with beyond highlighting it in the documentation (which we should do).

Issue #6071, opened 2023-12-18
Instantiate widgets from data


ahuang11:
> Yes, I still really want to see the .from_data (naming to be decided) method implemented on widgets where it makes sense. I find myself writing code to find the min/max or the unique values in a column or index constantly.

From https://github.com/holoviz/panel/pull/6011#issuecomment-1858784980

Issue #7003, opened 2024-07-19
Expand `pn.widget` to accept Pandas columns


jbednar:
I recently made a presentation trying to convey as simple a flow as possible for employing reactive expressions in a notebook. The pitch was meant to be "if you have a notebook that works, modify it only a tiny bit to get it to work with a widget controlling it". For instance, if you have `df[df['mag'] > 5].head()` now, how easily can we change "5" to a slider?

The notebook I have so far is at https://anaconda.cloud/share/notebooks/972482fb-5c0e-479e-b2f1-fe9a60a96d0d/preview , and it all works well, but I feel that the step of getting a widget instantiated is more complicated than it needs to be:

```python
import panel as pn
pn.extension(throttled=False, template='material')

slider = pn.widgets.FloatSlider(name='Minimum Magnitude', start=0, end=9, value=6)
```

That's a lot of syntax to remember or guess, plus specific numeric values to decide on and/or look up. Instead, what if we extended the currently fairly hidden `pn.widget` function to accept a Pandas Series, from which it can infer the type and configuration of a widget? It seems fairly well defined to implement `pn.widget(df.mag)` to return the equivalent of:

```python
slider = pn.widgets.FloatSlider(name='Minimum Magnitude', start=df.mag.min(), end=df.mag.max(), value=(df.mag.max()-df.mag.min())/2)
```

Similarly, `pn.widget(df.index)` would create the equivalent of:

```python
date = pn.widgets.DatetimeRangeSlider(name='Date', start=df.index[0], end=df.index[-1])
```

while a column (`pd.Series`) with categorical data would create a Selector, and so on. I think this approach would make working with DataFrames (and presumably similarly for Xarray dimensions or coordinates) much cleaner, clearer, easier to learn, easier to use, and less error prone.

Which widget to pick isn't always clearly defined, of course. E.g. integers are sometimes full ordinal values for which an IntSlider is appropriate, but also sometimes they just represent categories that happen to be denoted with numeric values. Might need to have a rule of thumb for integer columns to use a Selector when the number of unique values is < 50, and otherwise a slider.

If we use this approach, the example I was working on would reduce to the following code:

```python
import numpy as np, pandas as pd, panel as pn
pn.extension(throttled=False, template='material') #

df = pd.read_parquet('../data/earthquakes-projected.parq')
df.index = df.index.tz_localize(None)
df = df[['mag', 'depth', 'place', 'type']][df['northing'] < 20037508]
df = pn.rx(df) #

[...]

PR #7033, opened 2024-07-28
Implement Widget.from_values

philippjfr:
Implements https://github.com/holoviz/panel/issues/6071

- [x] Add docs
- [x] Update examples
- [x] Add tests

### DIFF
--- a/doc/how_to/components/index.md
+++ b/doc/how_to/components/index.md
@@ -26,6 +26,13 @@ How to access Pane Type.
 How to access and set widget values.
 :::
 
+:::{grid-item-card} {octicon}`telescope;2.5em;sd-mr-1 sd-animate-grow50` Construct Widget from Data
+:link: widget_values
+:link-type: doc
+
+How to automatically infer widget parameters from data values.
+:::
+
 :::{grid-item-card} {octicon}`file-diff;2.5em;sd-mr-1 sd-animate-grow50` Add or Remove Components
 :link: add_remove
 :link-type: doc
@@ -42,5 +49,6 @@ How to add or remove components from a panel.
 construct_panes
 pane_type
 widget_values
+widget_from_values
 add_remove
 ```
--- a/doc/how_to/components/index.md
+++ b/doc/how_to/components/index.md
@@ -27,7 +27,7 @@ How to access and set widget values.
 :::
 
 :::{grid-item-card} {octicon}`telescope;2.5em;sd-mr-1 sd-animate-grow50` Construct Widget from Data
-:link: widget_values
+:link: widget_from_values
 :link-type: doc
 
 How to automatically infer widget parameters from data values.
--- a/doc/how_to/components/widget_from_values.md
+++ b/doc/how_to/components/widget_from_values.md
@@ -1,0 +1,35 @@
+# Construct Widgets from Data
+
+This guide discusses how to automatically generate widget from data.
+
+---
+
+When working with data, be it in the form of lists, arrays or DataFrames it is common to want to filter that data. Manually computing the `start` and `end` values of a slider or the unique values of a dropdown can be an annoyance so widgets have a `classmethod` called `from_values` to help with this.
+
+```{pyodide}
+import pandas as pd
+import panel as pn
+pn.extension() # for notebook
+
+df = pd.read_csv("https://datasets.holoviz.org/penguins/v1/penguins.csv")
+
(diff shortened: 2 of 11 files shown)


jbednar:
This is a good start towards https://github.com/holoviz/panel/issues/7003; thanks!  A few questions:
1. What starting `value` is used by default?  Seems like you'd want a midpoint value by default, not just the start value?
2. For a Pandas Series, it should be cheap to calculate the min, avg, and max in a single pass (e.g.  `df['column_name'].agg(['min', 'average', 'max'])`) to get the start, value, and end; is that a useful special case for a Series?
3. Range widgets seem like they'd need different behavior, since a range should normally default to the start and end of the range. Will that already happen automatically?

philippjfr:
>What starting value is used by default? Seems like you'd want a midpoint value by default, not just the start value?

Honestly, while that's the logic for something like `interact` I'm not convinced that this is actually a better default.

>For a Pandas Series, it should be cheap to calculate the min, avg, and max in a single pass (e.g. df['column_name'].agg(['min', 'average', 'max'])) to get the start, value, and end; is that a useful special case for a Series?

Afaik, that isn't the case. It's no cheaper than doing the individual queries.

>Range widgets seem like they'd need different behavior, since a range should normally default to the start and end of the range. Will that already happen automatically?

Will take a look.

jbednar:
>> What starting value is used by default? Seems like you'd want a midpoint value by default, not just the start value?
> Honestly, while that's the logic for something like interact I'm not convinced that this is actually a better default.

Hmm. I think I'd typically rather have a middle value mainly because choosing the lower or the upper seems arbitrary, and is also by definition an extreme value.  Still, I guess it's not worth a lot of extra work. Plus using a midpoint could be problematic in some cases, e.g. when only a finite enumerable list of values is allowed, without values in between. 

>> For a Pandas Series, it should be cheap to calculate the min, avg, and max in a single pass (e.g. df['column_name'].agg(['min', 'average', 'max'])) to get the start, value, and end; is that a useful special case for a Series?
> Afaik, that isn't the case. It's no cheaper than doing the individual queries.

Drat. I saw some docs that implied they were performed together, but it didn't actually state that explicitly, and briefly clicking around in the source code suggests that it's decomposed into a separate call per agg function. Alas! You do need to run `min` and `max` to get the range bounds, but I guess running `average` would be an additional cost.

philippjfr:
>Hmm. I think I'd typically rather have a middle value mainly because choosing the lower or the upper seems arbitrary, and is also by definition an extreme value

I can think of quite a few common scenarios the middle value is arbitrary and the lower bound isn't. One relatively common case is when filtering where you treat the single value slider as a `>=` filter, in which case the lower bound will include everything and a middle value will arbitrarily exclude half the data.

jbednar:
> I can think of quite a few common scenarios the middle value is arbitrary and the lower bound isn't. One relatively common case is when filtering where you treat the single value slider as a >= filter, in which case the lower bound will include everything and a middle value will arbitrarily exclude half the data.

I think filtering should normally be done with a range widget (where there is no notion of a middle value), as discussed in https://github.com/holoviz/panel/issues/7003#issuecomment-2244287108 and https://github.com/holoviz/panel/pull/7033#issuecomment-2289926082 . In fact I suspect that filtering is the most typical and common use for this functionality, so I'd argue that returning a range widget by default is appropriate. 

In that case, a user will need to be explicit about requesting a single-value widget instead, and once they do that they might well be expected to specify explicitly that they want a midpoint default anyway.

jbednar:
In a side channel, @philippjfr pointed out that this particular code is only called once a concrete widget type has been selected, i.e. it's `Widget.from_values` (for a _particular_ Widget subclass in any given call), while `pn.widget()` is the one where classes are inferred. Good point, and so please ignore anything I said about choosing a range widget type, since in this code the type is already known. But that said, once we know we aren't instantiating a range widget, I still hold that a midpoint value is more useful as a default than an arbitrarily extreme value like the minimum value. 

Still, I defer to Philipp here, since (a) computing the midpoint isn't free, (b) computing the midpoint can be ambiguous (e.g. should it be the actual midpoint (which might not be a valid value or one that exists in this dataset), or should it be the nearest value actually existing in this dataset, and (c) Philipp is the one implementing it and using more than I am. :-)

philippjfr:
Thanks for summarizing that here so it's recorded.

Issue #6510, opened 2024-03-16
FastListTemplate PasswordInput text is hidden


ahuang11:
```python
import panel as pn

api_key_input = pn.widgets.PasswordInput(
    placeholder="sk-... uses OPENAI_API_KEY if not set",
    sizing_mode="stretch_width",
    styles={"color": "black"}
)
chat_interface = pn.chat.ChatInterface(
    callback=callback,
    callback_user="ChatGPT",
    help_text="Send a message to get a reply from ChatGPT!",
)
template = pn.template.FastListTemplate(
    title="OpenAI GPT-3.5",
    header_background="#212121",
    main=[chat_interface],
    header=[api_key_input],
)
template.servable()
```

<img width="1309" alt="image" src="https://github.com/holoviz/panel/assets/15331990/f548b1a4-17ee-4f00-84d3-d92a13c570b1">
<img width="830" alt="image" src="https://github.com/holoviz/panel/assets/15331990/38947725-8812-4f25-a996-4210336a2e16">

<img width="346" alt="image" src="https://github.com/holoviz/panel/assets/15331990/3aa8febe-81ba-4365-b7d3-b59533ad36b8">

PR #7040, opened 2024-07-29
fix header input color

ahuang11:
Closes https://github.com/holoviz/panel/issues/6510

I couldn't actually get it to point to the updated stylesheet even with `panel build panel` and cache clear, but I'm fairly certain this is the change needed

image

Because if I change it in the browser

image

The CSS var is based off the theme toggle

image


philippjfr (approved):

philippjfr:
Please revert this and make sure you test CSS changes like this in a variety of scenarios.

Screenshot 2024-08-27 at 18 52 25

PR #7042, opened 2024-07-29
Add placeholder while loading to ChatFeed

ahuang11:
To indicate activity while some callbacks are blocking, I've been wanting to have some control over the loading spinner, and have it persist.

https://github.com/user-attachments/assets/7bdd4855-4cc3-4a17-8379-c1ce7806a32f

```python
import asyncio
import panel as pn

pn.extension()

async def callback(message):
    await asyncio.sleep(2)
    return message


chat = pn.chat.ChatInterface(callback=callback, callback_exception="verbose", loading=True)
chat.servable()
```

### DIFF
--- a/panel/chat/feed.py
+++ b/panel/chat/feed.py
@@ -337,6 +337,13 @@ class ChatFeed(ListPanel):
             **self.placeholder_params
         )
 
+    @param.depends("loading", watch=True, on_init=True)
+    def _show_placeholder(self):
+        if self.loading:
+            self.append(self._placeholder)
+        else:
+            self._replace_placeholder(None)
+
     def _replace_placeholder(self, message: ChatMessage | None = None) -> None:
         """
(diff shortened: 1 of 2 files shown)


panel/chat/feed.py
```diff
@@ -337,6 +337,13 @@ def _update_placeholder(self):
             **self.placeholder_params
         )
 
+    @param.depends("loading", watch=True, on_init=True)
```
philippjfr, line 340: How does this interact with the usual loading spinner?

ahuang11: Do you mean something like `pn.Column("Hello", loading=True)`?

image

I don't think anything gets overlaid on ChatInterface besides the individual buttons
`pn.chat.ChatInterface("Hello", loading=True).show()`

image

philippjfr (approved):

Issue #7030, opened 2024-07-28
Enable me to create Viewer widgets


MarcSkovMadsen:
I've many, many times been wanting to create custom widgets using the `Viewer`. But I've experienced just as many times its not possible.

For example now to help P720 in https://discourse.holoviz.org/t/panel-chat-interface/7505 I would like to create a custom widget and use that for the `ChatInterface`. But it cannot work as the `Viewer` cannot be made to work like a `Widget`.

## Minimum, Reproducible Example

```python
import panel as pn
import param

pn.extension()

class PreferenceInput(pn.viewable.Layoutable, pn.widgets.WidgetBase, pn.viewable.Viewer):
    value = param.Parameter()

    def __init__(self, **params):
        super().__init__(**params)
        
        self._preference_widget = pn.widgets.RadioBoxGroup(options=["blue", "red", "green"], value="blue", name="preference")
        pn.bind(self._update_value, self._preference_widget)
        self._layout = pn.Column(
            "What is your preference?",
            self._preference_widget,
        )
    
    def _update_value(self, preference):
        self.value = "My preference is {preference}."

    def __panel__(self):
        return self._layout

preference_input = PreferenceInput()

def even_or_odd(contents, user, instance):
    if len(contents) % 2 == 0:
        return "Even number of characters."
    return "Odd number of characters."

pn.chat.ChatInterface(callback=even_or_odd, widgets=[preference_input]).servable()
```

![image](https://github.com/user-attachments/assets/9442956f-2df8-4f94-a19a-a0b6668cbb18)

It works fine until I click send. Then I get

```bash
  File "/home/jovyan/repos/private/panel/panel/chat/interface.py", line 405, in _click_send
    value = active_widget.value
```

In this case the problem is that when a `Viewer` is put in a row its replaced with its `__panel__()`. Thus then the assumptions about how to retrieve the `ChatInterface.active_widget` property breaks down. In this case you could probable "fix" the active_widget implementation. But in my experience its a problem all over the place that its not possible to get a `Viewer` to behave like a widget.

![image](https://github.com/user-attachments/assets/21fbb27a-51fd-4882-8402-8df442cf2c53)

As a bonus its also hard and not documented how to get a `Viewer` to be `Layoutable`. That is always another friction. The ChatInterface also assumes that about the input widgets. It assumes there is a `sizing_mode` on the widget.

ahuang11:
[...]

PR #7051, opened 2024-08-01
Implement PyComponent

MarcSkovMadsen:
I've been pointing out multiple times that its very hard to create real Panel widgets from the `Viewer`. Andrew then told me in https://github.com/holoviz/panel/issues/7030 that I should be using the `CompositeWidget`. I had not used that because the `Viewer` was originally added to enable me and others to create custom Panel components in general, so I thought that was the one I should use.


examples/reference/custom_components/CompositeWidget.md
```diff
@@ -0,0 +1,129 @@
...
+`CompositeWidget` simplifies the creation of custom Panel widgets using only Python and Panel components.
+
+```{pyodide}
+import param
+import panel as pn
+from panel.widgets import CompositeWidget
+
+pn.extension()
+
+class ButtonGroup(CompositeWidget):
```
MarcSkovMadsen, line 12: This component is really missing from Panel. It works in a way similar to https://www.gradio.app/docs/gradio/examples.

Its different from a `RadioButtonGroup` because 1) it does not show which option you selected last 2) instead you can reselect the same option and the `value` event will re-retrigger. This is very useful if the selection you made might be changed in other ways as seen in for example the second example.

MarcSkovMadsen, line 12: It would have been really nice of the `Viewer` and the `CompositeWidget` had similar names and worked in similar ways.

I believe it would have made things simpler if 

- they had had the names `CustomComponent` and `CustomWidget` respectively.
- they had worked in a similar way either via a `_composite` parameter or a `__panel__` method.
- that the `Viewer` was a `Layoutable` and also had its `Layoutable` parameters synced with the component shown. Right now you have to write a lot of boilerplate to get that working.

I don't know if now is the time to fix this friction?

philippjfr: Yes, all solid points. `CompositeWidget` was really developed for internal usage and never advertised while `Viewer` was actually advertised as an official mechanism for extending Panel.

examples/reference/custom_components/AnyWidget.md
```diff
@@ -37,7 +37,7 @@ CounterButton().servable()
 :::{note}
 Panel's `AnyWidgetComponent` supports using the [`AnyWidget`](https://anywidget.dev/) API on the JavaScript side and the [`param`](https://param.holoviz.org/) parameters API on the Python side.
 
-If you are looking to create custom components using Python and Panel component only, check out [`Viewer`](Viewer.md).
+If you are looking to create custom components using Python and Panel component only, check out [`Viewer`](Viewer.md) and [`CompositeWidget`](CompositeWidget.md).
```
ahuang11, line 40: As a new Panel user (and experienced) one, I think this should mention how one should go about choosing between Viewer and CompositeWidget, i.e. what scenarios should I use one over the other?

examples/reference/custom_components/CompositeWidget.md
ahuang11: Would it be possible to refactor Viewer internals to inherit from CompositeWidget?

examples/reference/custom_components/AnyWidget.md
MarcSkovMadsen: The CompositeWidget is for widgets. The Viewer for everything Else including Panes and layouts.

ahuang11: Should probably mention this since the names aren't intuitive.

MarcSkovMadsen: I tried to make that clear. But as you point out its to implicit and should be made more clear.

examples/reference/custom_components/CompositeWidget.md
MarcSkovMadsen: The problem is that it will then be a breaking change because the `CompositeWidget` and `Viewer` have different APIs.

But I agree we should simplify the story and chose one of the APIs for both widgets and non-widgets.

MarcSkovMadsen: Hmm. I confused about which class a `PyComponent` class should inherit from? Panel mixes in a lot in the `Widget` and `Pane` base classes.

MarcSkovMadsen:
The content here is ready for review.

I believe though that it should be refactor using a new `PyComponent` class together with the `BaseWidget` class. The `PyComponent` should address the missing features of the `Viewer` and enable to create widget by combining with `BaseWidget`.

examples/reference/custom_components/CompositeWidget.md
MarcSkovMadsen: This component would also have been useful for @ahuang11 chat example in https://discourse.holoviz.org/t/panel-chat-interface/7505/5?u=marc. The current version does not have a great UX because the `RadioButton` used does not support an unselected state.

philippjfr:
Just read your last comment, glad we agree 😄

philippjfr:
Just had a look at the tutorial and imo it absolutely does not read like a tutorial. There's no real steps for the user to follow along except for pasting the entire thing into a file and then running it. It's really much closer to a how-to guide.

MarcSkovMadsen:
> Just had a look at the tutorial and imo it absolutely does not read like a tutorial. There's no real steps for the user to follow along except for pasting the entire thing into a file and then running it. It's really much closer to a how-to guide.

Fair Feedback.

I would still like to achieve the goal of enabling me and other Panel users to create custom widgets using only Python. As it requires a lot of work I need some guidance.

- [ ] Is it ok to continue towards providing a `PyComponent` that 1) inherits from `Viewer` and `Layoutable` thus solving the problem that the `Viewer` has of not having a clear and easy path to support `sizing_mode` and other parameters from `Layoutable` 2) Aligns with the ESM components naming conventions etc.
- [ ] Add a `PyComponent` reference guide similar to `JSComponent` reference guide
- [ ] Update the tutorial to use `PyComponent` and adress your feedback.

MarcSkovMadsen:
> Just had a look at the tutorial and imo it absolutely does not read like a tutorial. There's no real steps for the user to follow along except for pasting the entire thing into a file and then running it. It's really much closer to a how-to guide.

Fair point. I will try to improve it. Its just hard for me to see how I can break it further down as it consists of a single class definition and a single function.

MarcSkovMadsen:
It would like to help finalize this, but its very unclear for me how to proceed.

It seems to me that the `PyComponent` is in a very preliminary state?

When I run

```python
import panel as pn
import param

from panel.custom import PyComponent


class CounterButton(PyComponent):

    value = param.Integer()

    def __init__(self, **params):
        super().__init__()
        self._layout = pn.widgets.Button(
            name=self._button_name, on_click=self._on_click, **params
        )

    def _on_click(self, event):
        self.value += 1

    @param.depends("value")
    def _button_name(self):
        return f"count is {self.value}"

    def __panel__(self):
        return self._layout

CounterButton(width=300)
```

I get 

```python
TypeError: PyComponent._sync__views() takes 1 positional argument but 2 were given
```

Do you plan on finalizing the implementation of the `PyComponent` @philippjfr ? What would the next step be?

panel/custom.py
```diff
@@ -46,6 +46,43 @@
...
+    `PyComponent` will forward layout parameters to the underlying
+    object.
+    """
+
+    def __init__(self, **params):
+        super().__init__(**params)
+        self.__views = []
+        self.param.watch(self._sync__views, [p for p in Layoutable.param if p != 'name'])
+
+    def _sync__views(self, **events):
```
philippjfr, line 64:
```suggestion
    def _sync__views(self, *events):
```

philippjfr:
>It seems to me that the PyComponent is in a very preliminary state?

I don't plan on changing anything else, there was just a small typo. I'm unclear whether it's useful or adds anything on top of what `Viewer` does.

panel/custom.py
```diff
@@ -46,6 +46,43 @@
...
+    """
+
+    def __init__(self, **params):
+        super().__init__(**params)
+        self.__views = []
+        self.param.watch(self._sync__views, [p for p in Layoutable.param if p != 'name'])
+
+    def _sync__views(self, *events):
+        params = {e.name: e.new for e in events}
+        for view in self._views:
```
philippjfr, line 66:
```suggestion
        for view in self.__views:
```

philippjfr:
Nevermind, it does need a lot more work.

philippjfr:
The docs can still be improved here but the component now behaves as I'd like it to and I want it to get into an RC release.

philippjfr (approved):

PR #7053, opened 2024-08-01
Update Tabulator.ipynb to show correct version number of Tabulator

Coderambling:
Doc says Tabulator 5.5 is used, but Tabulator.py shows 6.2.1 is used.

Changed version number in doc to 6.2.1 .

https://github.com/holoviz/panel/blob/main/panel%2Fmodels%2Ftabulator.py#L19

### DIFF
--- a/examples/reference/widgets/Tabulator.ipynb
+++ b/examples/reference/widgets/Tabulator.ipynb
@@ -19,7 +19,7 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "The `Tabulator` widget allows displaying and editing a pandas DataFrame. The `Tabulator` is a largely backward compatible replacement for the [`DataFrame`](./DataFrame.ipynb) widget and will eventually replace it. It is built on the **version 5.5** of the [Tabulator](http://tabulator.info/) library, which provides for a wide range of features.\n",
+    "The `Tabulator` widget allows displaying and editing a pandas DataFrame. The `Tabulator` is a largely backward compatible replacement for the [`DataFrame`](./DataFrame.ipynb) widget and will eventually replace it. It is built on the **version 6.2.1** of the [Tabulator](http://tabulator.info/) library, which provides for a wide range of features.\n",
     "\n",
     "Discover more on using widgets to add interactivity to your applications in the [how-to guides on interactivity](../how_to/interactivity/index.md). Alternatively, learn [how to set up callbacks and (JS-)links between parameters](../../how_to/links/index.md) or [how to use them as part of declarative UIs with Param](../../how_to/param/index.html).\n",
     "\n",
--- a/examples/reference/widgets/Tabulator.ipynb
+++ b/examples/reference/widgets/Tabulator.ipynb
@@ -19,7 +19,7 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "The `Tabulator` widget allows displaying and editing a pandas DataFrame. The `Tabulator` is a largely backward compatible replacement for the [`DataFrame`](./DataFrame.ipynb) widget and will eventually replace it. It is built on the **version 6.2.1** of the [Tabulator](http://tabulator.info/) library, which provides for a wide range of features.\n",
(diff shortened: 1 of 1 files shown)


hoxbro:
Let's try automating this like done in https://github.com/holoviz/panel/pull/5243

Coderambling:
Yes that would be great! Was going through my head when I made the PR. Didn't know that was possible.

Should this PR still go through in the mean time? I would say yes, as it corrects the info and creates  a window to apply automation.

I don't think I will be able to apply the PR you mention to automate it, as I am not familiar with that code/ Myst.

examples/reference/widgets/Tabulator.ipynb
```diff
@@ -19,7 +19,7 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "The `Tabulator` widget allows displaying and editing a pandas DataFrame. The `Tabulator` is a largely backward compatible replacement for the [`DataFrame`](./DataFrame.ipynb) widget and will eventually replace it. It is built on the **version 5.5** of the [Tabulator](http://tabulator.info/) library, which provides for a wide range of features.\n",
+    "The `Tabulator` widget allows displaying and editing a pandas DataFrame. The `Tabulator` is a largely backward compatible replacement for the [`DataFrame`](./DataFrame.ipynb) widget and will eventually replace it. It is built on the **version 6.2.1** of the [Tabulator](http://tabulator.info/) library, which provides for a wide range of features.\n",
```
hoxbro, line 22:
```suggestion
    "The `Tabulator` widget allows displaying and editing a pandas DataFrame. The `Tabulator` is a largely backward compatible replacement for the [`DataFrame`](./DataFrame.ipynb) widget and will eventually replace it. It is built on the **version {{TABULATOR_VERSION}}** of the [Tabulator](http://tabulator.info/) library, which provides for a wide range of features.\n",
```

hoxbro (commented):
> I don't think I will be able to apply the PR you mention to automate it, as I am not familiar with that code.

It is not hard to do! You just need to change the placeholder and update the conf.py file with the following: 

```    python
	version_replace = {
		...  # Other versions
		"{{TABULATOR_VERSION}}": "6.2.1",
    }
```

examples/reference/widgets/Tabulator.ipynb
Coderambling: Ok by doing that the .ipynb will take the version number from conf.py? But the correct version number will then still have to be entered in conf.py manually? Or can it be automated there as well?

hoxbro: Yes, you should be able to get the version automatically with `from panel.models.tabulator import TABULATOR_VERSION`

Coderambling:
Ok added it to conf.py

https://github.com/holoviz/panel/commit/2d9d5e5859a297dd25c57476f1b178cf6a4aa413

Coderambling:
Hm the line below causes the tests to fail with exit code 1.

When I comment it out, tests are successful.

@hoxbro , should it be:

from panel.io.models.tabulator

instead?



from panel.models.tabulator import TABULATOR_VERSION

hoxbro:
Please combine this and https://github.com/holoviz/panel/pull/7062, and let us take it from there.

You can confirm the import is correct with `python -c "from panel.models.tabulator import TABULATOR_VERSION; print(TABULATOR_VERSION)"` in an environment with panel installed.

philippjfr:
Guess I'll just merge both.

philippjfr (approved):

Coderambling:
Note: Is conf.py udated automatically / manually with the latest version?

Should we do the same for other docs like for Perspective etc, so that correct library versions are always automatically shown in the various docs?

MarcSkovMadsen:
+1 on showing version number in docs for components where it is relevant to study documents of js library to be able to use.

Coderambling:
Which other docs would this apply to?

Perspective
VTK
VTKJS
...?

MarcSkovMadsen:
+ Deck.gl/ PyDeck, Echarts/ PyEcharts, Katex/ MathJax, Plotly, Altair & Vega, Vizzu,

Issue #7045, opened 2024-07-31
How to stop event propagation from a TextInput to a Card header?


giladpn:
#### ALL software version info
panel 1.4.4
bokeh 3.4.2
uvicorn 0.30.1
panel server running using pn.serve() inside a uvicorn server
Linux server on Ubuntu 22.04
Workstation Windows 11
On windows, browser Microsoft Edge Version 127.0.2651.74

#### Description of expected behavior and the observed behavior
Hi, and thanks as always for the help.

My app needs an TextInput in the header of a Card, so users can type a title for that Card. However, I have a couple of issues with event propagation from that TextInput.

When a user clicks inside the TextInput (because the user wants to type in it), the click will also cause the entire Card to collapse (or de-collapse).
When a user types inside the TextInput, and happens to type a space, the entire Card will collapse (or de-collapse).
I do realise this is because the event is propagating from the TextInput to the parent Card. How do I fix this behavior?

Code below:

#### Complete, minimal, self-contained example code that reproduces the issue

```
import panel as pn

pn.extension()

header_text = pn.widgets.TextInput(
    placeholder='describe your analysis here',
    disabled=False,
    max_length=255,
    width_policy='max',
    sizing_mode='stretch_width',
)

header_row = pn.Row(
    header_text,
    width_policy='max',
    sizing_mode='stretch_width',
)

card = pn.Card(
    "A line with some text",
    header=header_row,
    collapsed=False,
    collapsible=True,
    width_policy='max',
    sizing_mode='stretch_width',
)

card.servable()
```

PR #7057, opened 2024-08-02
Do not propagate clicks on input elements in Card header

philippjfr:
Fixes https://github.com/holoviz/panel/issues/7045

### DIFF
--- a/panel/models/card.ts
+++ b/panel/models/card.ts
@@ -121,7 +121,7 @@ export class CardView extends ColumnView {
 
   _toggle_button(e: MouseEvent): void {
     for (const path of e.composedPath()) {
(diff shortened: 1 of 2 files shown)


panel/models/card.ts
```diff
@@ -120,7 +119,12 @@ export class CardView extends ColumnView {
     this.invalidate_layout()
   }
 
-  _toggle_button(): void {
+  _toggle_button(e: MouseEvent): void {
+    for (const path of e.composedPath()) {
+      if (path instanceof HTMLInputElement ) {
```
philippjfr, line 124:
```suggestion
      if (path instanceof HTMLInputElement) {
```

Issue #5207, opened 2023-07-04
Add speed values to Player widget


droumis:
#### Is your feature request related to a problem? Please describe.
When the `-` or `+` buttons to slow down or speed up the player are pressed, there's no display of what the speed increment has been changed to.

#### Describe the solution you'd like
Include speed values like 0.5, 1.5, 2.0, etc to the Player

#### Describe alternatives you've considered
1. Seeing how current values change might be an alternative (https://github.com/holoviz/panel/issues/229)
2. Use a discrete slider for speed instead of `-`, `+` buttons

PR #7064, opened 2024-08-02
Show interval value on click

ahuang11:
Adds CSS classes to the buttons + display value on speed increase

https://github.com/user-attachments/assets/121147f1-cf4a-4b9a-83c0-ef7a0929542e

Closes https://github.com/holoviz/panel/issues/5207


philippjfr:
Just checking, but would showing `fps` rather than `ms` be clearer?

ahuang11:
I like that idea of using fps.

panel/models/player.ts
```diff
@@ -320,8 +333,17 @@ export class PlayerView extends WidgetView {
     this.set_frame(this.model.end)
   }
 
+  updateSpeedButton(button: HTMLButtonElement, interval: number, originalSVG: string): void {
+    const fps = 1000 / interval
+    button.innerHTML = `${fps.toFixed(1)}<br>fps`
+    setTimeout(() => {
+      button.innerHTML = originalSVG
+    }, 1500) // Show for 1.5 seconds
```
hoxbro, line 341: I suggest having a parameter for this so it can be set in Python.

panel/models/player.ts
```diff
@@ -320,8 +333,17 @@ export class PlayerView extends WidgetView {
     this.set_frame(this.model.end)
   }
 
+  updateSpeedButton(button: HTMLButtonElement, interval: number, originalSVG: string): void {
+    const fps = 1000 / interval
+    button.innerHTML = `${fps.toFixed(1)}<br>fps`
+    setTimeout(() => {
+      button.innerHTML = originalSVG
+    }, this.model.preview_duration) // Show for 1.5 seconds
```
hoxbro, line 341:
```suggestion
    }, this.model.preview_duration)
```

How does this behave if `preview_duration` is equal to zero.

panel/widgets/player.py
```diff
@@ -31,6 +33,10 @@ class PlayerBase(Widget):
         default='once', objects=['once', 'loop', 'reflect'], doc="""
         Policy used when player hits last frame""")
 
+    preview_duration = param.Integer(default=1500, doc="""
```
hoxbro, line 36:
```suggestion
    preview_duration = param.Integer(default=1500, bounds=(0, None), doc="""
```

philippjfr (approved):

Issue #7078, opened 2024-08-05
Incomplete/missing type annotations


gandhis1:
A number of elements of the public interface are untyped and thus can get flagged in user code that uses strict typing. Here are just a few. I'm flagging these as it's not clear whether this is intentional - large parts of `panel` are annotated and these seem to just have been missed more so than a deliberate choice to not type them.

```
error: Untyped decorator makes function X untyped  [misc] (this was for @pn.cache)
error: Call to untyped function "Button" in typed context  [no-untyped-call] (pn.widgets.Button)
error: Call to untyped function X in typed context  [no-untyped-call] (inherits from pn.widgets.Tabulator)
error: Call to untyped function "StaticText" in typed context  [no-untyped-call] (pn.widgets.StaticText)
error: Call to untyped function "Card" in typed context  [no-untyped-call] (pn.Card)
error: Call to untyped function "HSpacer" in typed context  [no-untyped-call] (pn.layout.HSpacer)
error: Call to untyped function "Number" in typed context  [no-untyped-call] (pn.indicators.Number)
error: Call to untyped function X in typed context  [no-untyped-call] (inherits from pn.viewable.Viewer)
```

In particular, annotations appear to be missing from a bunch of `__init__` - even an `Any` would be better than nothing.

I've annotated tons of code in the past, including on several major OSS code bases, so happy to help out if there is interest.

MarcSkovMadsen:
Help would be awesome. Type annotations makes Panel much easier to use for users.

I would try to make a smaller contribution as a start. See how its received and if anything needs to be aligned/ discussed. Then do one or more larger PRs.

I would like to be a reviewer.

PR #7079, opened 2024-08-05
Annotate pn.io.cache decorator

gandhis1:
Part of https://github.com/holoviz/panel/issues/7078

In the docs I see that [minimum Python version is 3.9](https://panel.holoviz.org/developer_guide/index.html#create-a-development-environment), however in the `pyproject.toml` it states `requires-python = ">=3.10"`. Importing `ParamSpec` from `typing` instead of `typing_extensions` requires 3.10. This PR assumes the latter is correct and that `typing_extensions` is not required.

@MarcSkovMadsen

### DIFF
--- a/panel/io/cache.py
+++ b/panel/io/cache.py
@@ -320,7 +320,7 @@ def cache(
     policy: Literal['FIFO', 'LRU', 'LFU'] = ...,
     ttl: float | None = ...,
     to_disk: bool = ...,
-    cache_path: str = ...,
+    cache_path: str | os.PathLike = ...,
     per_session: bool = ...,
 ) -> Callable[[Callable[P, R]], Callable[P, R]]:
     ...
@@ -333,7 +333,7 @@ def cache(
     policy: Literal['FIFO', 'LRU', 'LFU'] = ...,
     ttl: float | None = ...,
     to_disk: bool = ...,
-    cache_path: str = ...,
+    cache_path: str | os.PathLike = ...,
     per_session: bool = ...,
 ) -> Callable[P, R]:
     ...
@@ -345,7 +345,7 @@ def cache(
     policy: Literal['FIFO', 'LRU', 'LFU'] = 'LRU',
     ttl: float | None = None,
     to_disk: bool = False,
-    cache_path: str = './cache',
+    cache_path: str | os.PathLike = './cache',
     per_session: bool = False
 ) -> Callable[P, R] | Callable[[Callable[P, R]], Callable[P, R]]:
     """
--- a/panel/io/cache.py
+++ b/panel/io/cache.py
@@ -18,21 +18,8 @@ import weakref
 
 from contextlib import contextmanager
 from typing import (
-    TYPE_CHECKING, Any, Callable, Hashable, Literal, ParamSpec, Protocol,
-    TypeVar, overload,
+    Any, Callable, Hashable, Literal, ParamSpec, Protocol, TypeVar, overload,
 )
-
-if TYPE_CHECKING:
-    P = ParamSpec("P")
-    R = TypeVar("R")
-    T = TypeVar("T")
-    CallableT = TypeVar("CallableT", bound=Callable)
-
-    class CachedFunc(Protocol[CallableT]):
-        def clear(self, func_hashes: list[str | None]) -> None:
-            pass
-
-        __call__: CallableT
 
 import param
 
@@ -43,6 +30,16 @@ from .state import state
 #---------------------------------------------------------------------
 # Private API
 #---------------------------------------------------------------------
+
+_P = ParamSpec("_P")
+_R = TypeVar("_R")
+_CallableT = TypeVar("_CallableT", bound=Callable)
+
+class _CachedFunc(Protocol[_CallableT]):
+    def clear(self, func_hashes: list[str | None]) -> None:
+        pass
+
+    __call__: _CallableT
 
 _CYCLE_PLACEHOLDER = b"panel-93KZ39Q-floatingdangeroushomechose-CYCLE"
 
@@ -331,12 +328,12 @@ def cache(
     to_disk: bool = ...,
     cache_path: str | os.PathLike = ...,
     per_session: bool = ...,
-) -> Callable[[Callable[P, R]], CachedFunc[Callable[P, R]]]:
+) -> Callable[[Callable[_P, _R]], _CachedFunc[Callable[_P, _R]]]:
     ...
 
 @overload
 def cache(
-    func: Callable[P, R],
+    func: Callable[_P, _R],
     hash_funcs: dict[type[Any], Callable[[Any], bytes]] | None = ...,
     max_items: int | None = ...,
     policy: Literal['FIFO', 'LRU', 'LFU'] = ...,
@@ -344,11 +341,11 @@ def cache(
     to_disk: bool = ...,
     cache_path: str | os.PathLike = ...,
     per_session: bool = ...,
-) -> CachedFunc[Callable[P, R]]:
+) -> _CachedFunc[Callable[_P, _R]]:
     ...
 
 def cache(
-    func: Callable[P, R] | None = None,
+    func: Callable[_P, _R] | None = None,
     hash_funcs: dict[type[Any], Callable[[Any], bytes]] | None = None,
     max_items: int | None = None,
     policy: Literal['FIFO', 'LRU', 'LFU'] = 'LRU',
@@ -356,7 +353,7 @@ def cache(
     to_disk: bool = False,
     cache_path: str | os.PathLike = './cache',
     per_session: bool = False
-) -> CachedFunc[Callable[P, R]] | Callable[[Callable[P, R]], CachedFunc[Callable[P, R]]]:
+) -> _CachedFunc[Callable[_P, _R]] | Callable[[Callable[_P, _R]], _CachedFunc[Callable[_P, _R]]]:
     """
     Memoizes functions for a user session. Can be used as function annotation or just directly.
 
@@ -395,7 +392,7 @@ def cache(
 
     hash_funcs = hash_funcs or {}
     if func is None:
-        def decorator(func: Callable[P, R]) -> CachedFunc[Callable[P, R]]:
+        def decorator(func: Callable[_P, _R]) -> _CachedFunc[Callable[_P, _R]]:
             return cache(
                 func=func,
                 hash_funcs=hash_funcs,
@@ -463,7 +460,7 @@ def cache(
 
     if iscoroutinefunction(func):
         @functools.wraps(func)
-        async def wrapped_func(*args: P.args, **kwargs: P.kwargs) -> R:
+        async def wrapped_func(*args: _P.args, **kwargs: _P.kwargs) -> _R:
             func_cache, hash_value, time = hash_func(*args, **kwargs)
             if hash_value in func_cache:
                 with lock:
@@ -476,7 +473,7 @@ def cache(
             return ret
     else:
         @functools.wraps(func)
-        def wrapped_func(*args: P.args, **kwargs: P.kwargs) -> R:
+        def wrapped_func(*args: _P.args, **kwargs: _P.kwargs) -> _R:
             func_cache, hash_value, time = hash_func(*args, **kwargs)
             if hash_value in func_cache:
                 with lock:
--- a/panel/io/cache.py
+++ b/panel/io/cache.py
@@ -1,6 +1,8 @@
 """
 Implements memoization for functions with arbitrary arguments
 """
+from __future__ import annotations
+
 import datetime as dt
 import functools
 import hashlib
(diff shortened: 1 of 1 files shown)


gandhis1:
It looks like the `main` build was failing until it was fixed a few hours ago, so I will rebase this...

```
=========================== short test summary info ============================
FAILED panel/tests/ui/widgets/test_tabulator.py::test_tabulator_header_filter_no_horizontal_rescroll[remote] - AssertionError: assert {'height': 20...: 264, 'y': 9} == {'height': 20...: 714, 'y': 9}
  Omitting 3 identical items, use -vv to show
  Differing items:
  {'x': 264} != {'x': 714}
  Full diff:
  - {'height': 20, 'width': 91, 'x': 714, 'y': 9}
  ?                                  ^^
  + {'height': 20, 'width': 91, 'x': 264, 'y': 9}
  ?                                  ^^
== 1 failed, 562 passed, 10 skipped, 28 xfailed, 4 rerun in 770.47s (0:12:50) ==
```

MarcSkovMadsen:
I'm wondering whats the best way to check this via `mypy`.

I tried `mypy --follow-imports=skip panel/io/cache.py` but it errors.

```bash
$ mypy --follow-imports=skip panel/io/cache.py
  File "/home/jovyan/repos/private/panel/.venv/bin/mypy", line 8, in <module>
    sys.exit(console_entry())
             ^^^^^^^^^^^^^^^
  File "/home/jovyan/repos/private/panel/.venv/lib/python3.11/site-packages/mypy/__main__.py", line 15, in console_entry
    main()
  File "mypy/main.py", line 103, in main
  File "mypy/main.py", line 187, in run_build
  File "mypy/build.py", line 193, in build
  File "mypy/build.py", line 268, in _build
  File "mypy/build.py", line 2950, in dispatch
  File "mypy/build.py", line 3348, in process_graph
  File "mypy/build.py", line 3475, in process_stale_scc
  File "mypy/build.py", line 2507, in write_cache
  File "mypy/build.py", line 1568, in write_cache
  File "mypy/nodes.py", line 390, in serialize
  File "mypy/nodes.py", line 4012, in serialize
  File "mypy/nodes.py", line 3949, in serialize
  File "mypy/nodes.py", line 3374, in serialize
  File "mypy/types.py", line 671, in serialize
  File "mypy/types.py", line 2428, in serialize
  File "mypy/types.py", line 1451, in serialize
  File "mypy/types.py", line 671, in serialize
  File "mypy/types.py", line 3093, in serialize
AssertionError: Internal error: unresolved placeholder type None
```

Then I tried `mypy panel/io/cache.py` which seems to work better.

```bash
$ mypy panel/io/cache.py
panel/io/cache.py:24: error: Skipping analyzing "param": module is installed, but missing library stubs or py.typed marker  [import-untyped]
panel/io/cache.py:26: error: Skipping analyzing "param.parameterized": module is installed, but missing library stubs or py.typed marker  [import-untyped]
panel/io/cache.py:26: note: See https://mypy.readthedocs.io/en/stable/running_mypy.html#missing-imports
panel/io/cache.py:38: error: Need type annotation for "_HASH_MAP" (hint: "_HASH_MAP: dict[<type>, <type>] = ...")  [var-annotated]
panel/io/cache.py:40: error: Need type annotation for "_HASH_STACKS"  [var-annotated]
panel/io/cache.py:110: error: Library stubs not installed for "pandas"  [import-untyped]
panel/io/cache.py:110: note: Hint: "python3 -m pip install pandas-stubs"
panel/io/cache.py:110: note: (or run "mypy --install-types" to install all missing stub packages)
panel/io/cache.py:177: error: Incompatible types in assignment (expression has type "bytes", target has type "Callable[[Any], Any]")  [assignment]
panel/io/cache.py:389: error: Incompatible return value type (got "Callable[[Any], Callable[[Callable[P, R]], Callable[P, R]]]", expected "Callable[P, R@cache] | Callable[[Callable[P, R@cache]], Callable[P, R@cache]]")  [return-value]
panel/io/cache.py:436: error: Cannot find implementation or library stub for module named "diskcache"  [import-not-found]
panel/io/cache.py:462: error: Incompatible types in "await" (actual type "R", expected type "Awaitable[Any]")  [misc]
panel/io/cache.py:493: error: "_Wrapped[P, R, P, Coroutine[Any, Any, R]]" has no attribute "clear"  [attr-defined]
panel/io/cache.py:505: error: Incompatible return value type (got "_Wrapped[P, R, P, Coroutine[Any, Any, R]]", expected "Callable[P, R] | Callable[[Callable[P, R]], Callable[P, R]]")  [return-value]
Found 11 errors in 1 file (checked 1 source file)
```

MarcSkovMadsen:
I've tried to configure `mypy` in https://github.com/holoviz/panel/pull/7081. Could you do a review of that one @gandhis1? Thanks.

MarcSkovMadsen:
I can see that @hoxbro updated to Python 3.10 in #6883. This PR has not yet been released. I.e. the upgrade to Python 3.10 for users will happen when 1.5.0 is released.

panel/io/cache.py
```diff
@@ -305,11 +308,46 @@ def compute_hash(func, hash_funcs, args, kwargs):
         _HASH_MAP[key] = hash_value
     return hash_value
 
+P = ParamSpec("P")
+R = TypeVar("R")
+T = TypeVar("T")
+
+@overload
```
MarcSkovMadsen, line 315: After merging this PR with 7081 the errors I see for the `panel/io/cache.py` file are

```bash
panel/io/cache.py:38: error: Need type annotation for "_HASH_MAP" (hint: "_HASH_MAP: dict[<type>, <type>] = ...")  [var-annotated]
panel/io/cache.py:40: error: Need type annotation for "_HASH_STACKS"  [var-annotated]
panel/io/cache.py:177: error: Incompatible types in assignment (expression has type "bytes", target has type "Callable[[Any], Any]")  [assignment]
panel/io/cache.py:389: error: Incompatible return value type (got "Callable[[Any], Callable[[Callable[P, R]], Callable[P, R]]]", expected "Callable[P, R@cache] | Callable[[Callable[P, R@cache]], Callable[P, R@cache]]")  [return-value]
panel/io/cache.py:462: error: Incompatible types in "await" (actual type "R", expected type "Awaitable[Any]")  [misc]
panel/io/cache.py:493: error: "_Wrapped[P, R, P, Coroutine[Any, Any, R]]" has no attribute "clear"  [attr-defined]
panel/io/cache.py:505: error: Incompatible return value type (got "_Wrapped[P, R, P, Coroutine[Any, Any, R]]", expected "Callable[P, R] | Callable[[Callable[P, R]], Callable[P, R]]")  [return-value]
```

Thus I see some issue in line 389 after the cache type annoation is introduced. I don't see issues introduced in other files.

panel/io/cache.py
```diff
@@ -305,11 +308,46 @@ def compute_hash(func, hash_funcs, args, kwargs):
...
+def cache(
+    func: Literal[None] = ...,
+    hash_funcs: dict[type[Any], Callable[[Any], bytes]] | None = ...,
+    max_items: int | None = ...,
+    policy: Literal['FIFO', 'LRU', 'LFU'] = ...,
+    ttl: float | None = ...,
+    to_disk: bool = ...,
+    cache_path: str = ...,
+    per_session: bool = ...,
+) -> Callable[[Callable[P, R]], Callable[P, R]]:
```
MarcSkovMadsen, line 325: I'm not an expert on type check, so I would like to learn why this is needed. Can you please explain?

panel/io/cache.py
```diff
@@ -305,11 +308,46 @@ def compute_hash(func, hash_funcs, args, kwargs):
...
-    func=None, hash_funcs=None, max_items=None, policy='LRU',
-    ttl=None, to_disk=False, cache_path='./cache', per_session=False
-):
+    func: Callable[P, R] | None = None,
+    hash_funcs: dict[type[Any], Callable[[Any], bytes]] | None = None,
+    max_items: int | None = None,
+    policy: Literal['FIFO', 'LRU', 'LFU'] = 'LRU',
+    ttl: float | None = None,
+    to_disk: bool = False,
+    cache_path: str = './cache',
```
MarcSkovMadsen, line 348: I see `cache_path` is documented as `cache_dir` in the docstring. Please consider fixing. Thanks.

MarcSkovMadsen, line 348: As far as I can see `cache_path` is used as input to `os.path.join`. Thus I believe that it could also be a `pathlib.Path` type and that the type annotation should be updated to this?

panel/io/cache.py
```diff
@@ -414,7 +452,7 @@ def hash_func(*args, **kwargs):
 
     if iscoroutinefunction(func):
         @functools.wraps(func)
-        async def wrapped_func(*args, **kwargs):
+        async def wrapped_func(*args: P.args, **kwargs: P.kwargs) -> R:
```
MarcSkovMadsen, line 455: For me to learn. Can you explain what goes on here? Thanks.

MarcSkovMadsen:
Note to my self. I should consider contributing an example to the docstring and maybe a link to reference docs to align with other docstrings.

panel/io/cache.py
hoxbro: I _think_ `str | os.PathLike` is right.

panel/io/cache.py
```diff
@@ -305,11 +308,46 @@ def compute_hash(func, hash_funcs, args, kwargs):
         _HASH_MAP[key] = hash_value
     return hash_value
 
+P = ParamSpec("P")
+R = TypeVar("R")
+T = TypeVar("T")
```
hoxbro, line 313: Maybe add these below  the imports behind `TYPE_CHECKING`

hoxbro:
> I can see that @hoxbro updated to Python 3.10 in https://github.com/holoviz/panel/pull/6883.

Yep, Bokeh 3.5 dropped Python 3.9 support. So we dropped it, too, as we are hard pinned to 3.5.

panel/io/cache.py
hoxbro: Maybe a relevant youtube video:  https://www.youtube.com/watch?v=_QXlbwRmqgI

panel/io/cache.py
gandhis1: `cache` has situational behavior depending on the arguments, it can either be a first-order decorator or a second-order decorator.

If you use it as `@pn.cache(ttl=60)`, then you are effectively using as a function that returns a decorator (`pn.cache(ttl=60)(my_func)`) or a second-order decorator. That is one of the overloads.

If you use it as `@st.cache`. you are using it as a first-order decorator (`pn.cache(my_func)`). That is the second overload.

The overloads are vitally important because for user code the most important thing is that, regardless of how `cache` is used, it needs to always return a final function with a type signature that looks the same as the user's original function. This should achieve that.

panel/io/cache.py
gandhis1: `ParamSpec` `P` represents the type signatures of the `func` that was passed in. Since `func` was annotated as `Callable[P, R]`, this is just saying that `wrapped_func` has the same signature as `func`. It doesn't really make a practical difference here, but for instance, something you could do is add an extra positional or keyword argument to `wrapped_func`, and that would get validated in subsequent calls.

gandhis1:
I will make the requested changes and address the Mypy errors later today.

panel/io/cache.py
gandhis1: The line 389 issue I fixed by converting the `lambda` into a regular `def`, which supports annotations.

I took an initial look at fixing the others as well:

- Added an annotation for `_HASH_MAP`
- `_HASH_STACKS` does not even appear to be used anywhere, so I couldn't figure out what it was supposed to be annotated as. `_STACK` is likewise not used anywhere. Can these be deleted?
- The line 177 error sure looks like a bug? How can a hash function be a bytes constant (see snippet below)? Those are not callable and `hash_func` is used as a callable.
```
for name in _FFI_TYPE_NAMES:
    _hash_funcs[name] = b'0'
```
- The line 493 error I ignored for now, as it is assigned an arbitrary attribute `clear` on a function. We might be able to define a more precise return type that annotates the `clear` as well but that is probably overkill.
- The remaining `async` / `await` errors I partially fixed by adding an `Awaitable` variant, however I had to add a `type: ignore` for one of them because `iscoroutinefunction` does not implement a `TypeGuard`, which it would need to to avoid the ignore. (note that `inspect.iscoroutinefunction` does implement this in `typeshed`).

MarcSkovMadsen (approved):
With the (limited) knowledge of typing I have, I can approve.

I believe the remaining questions can should be answered by Philipp.

hoxbro:
Having variables in `TYPE_CHECKING` makes them private. I prefer having them there, as it also makes it easier to read, but I'm not sure what the "standard" is.

gandhis1:
> Having variables in `TYPE_CHECKING` makes them private. I prefer having them there, as it also makes it easier to read, but I'm not sure what the "standard" is.

I restored this - the errors I previously encountered when doing this was because I was missing `from __future__ import annotations`.

gandhis1:
The tests keep on failing in places I am unfamiliar with. Do these test failures actually have anything to do with this PR?

philippjfr:
Some tests seem to be a little flakey will look at those again soon, but you're good to go here!

PR #7081, opened 2024-08-06
Enable using mypy manually

MarcSkovMadsen:
In #7079, @gandhis1 contributes a typing improvement. In order to be able to review this PR I would like to run `mypy`.

This PR configures the project in such a way that you can

- run `pip install -e .[mypy]` to install the `mypy` dependencies
- run `mypy path/to/file.py` to check a specific file.

It removes the errors raised about missing stubs and dependencies that do not support typing.


hoxbro:
I pushed some changes so we can see the result of type checking in the CI. I have set the CI to never fail, but this should be removed at some point.

pyproject.toml
```diff
@@ -207,3 +217,52 @@ filterwarnings = [
     "ignore:unclosed file <_io.TextIOWrapper name='(/dev/null|nul)' mode='w':ResourceWarning", # OK
     "ignore:Deprecated in traitlets 4.1, use the instance .metadata dictionary directly", # OK (ipywidgets internal)
 ]
+
+[tool.mypy]
+python_version = "3.10"
```
hoxbro, line 222: I think this is automatically inferred by `requires-python`

pyproject.toml
```diff
@@ -207,3 +217,52 @@ filterwarnings = [
...
+    "reacton.*",
+    "rpy2.*",
+    "scipy.*",
+    "setuptools_scm.*",
+    "snakeviz.*",
+    "streamz.*",
+    "textual.*",
+    "tranquilizer.*",
+    "vtk.*",
+
```
hoxbro, line 266:
```suggestion

```

.github/workflows/test.yaml
```diff
@@ -257,3 +257,22 @@ jobs:
       - name: Test Unit
         run: |
           pixi run -e ${{ matrix.environment }} test-unit
+
+  type_test_suite:
+    name: lint:${{ matrix.environment }}:${{ matrix.os }}
```
hoxbro, line 262:
```suggestion
    name: type:${{ matrix.environment }}:${{ matrix.os }}
```

.github/workflows/test.yaml
```diff
@@ -257,3 +257,22 @@ jobs:
...
+        os: ["ubuntu-latest"]
+        environment: ["test-type"]
+    timeout-minutes: 120
+    steps:
+      - uses: holoviz-dev/holoviz_tasks/pixi_install@pixi
+        with:
+          environments: ${{ matrix.environment }}
+      - name: Test Type
+        run: |
+          pixi run -e ${{ matrix.environment }} type || echo "Failed"
```
hoxbro, line 278:
```suggestion
          pixi run -e ${{ matrix.environment }} test-type || echo "Failed"
```

pixi.toml
```diff
@@ -144,6 +145,22 @@ _install-ui = 'playwright install chromium'
...
+pandas-stubs = "*"
+types-bleach = "*"
+types-croniter = "*"
+types-Markdown = "*"
+types-psutil = "*"
+types-requests = "*"
+types-tqdm = "*"
+
+[feature.type.tasks]
+type = 'mypy panel'
```
hoxbro, line 162:
```suggestion
test-type = 'mypy panel'
```

pixi.toml
```diff
@@ -144,6 +145,22 @@ _install-ui = 'playwright install chromium'
 cmd = 'pytest panel/tests/ui --ui --browser chromium -n logical --dist loadgroup --reruns 3 --reruns-delay 10'
 depends_on = ["_install-ui"]
 
+# =============================================
+# ================== TYPES ====================
+# =============================================
+[feature.type.dependencies]
+mypy = "*"
+pandas-stubs = "*"
```
gandhis1, line 153: `typing-extensions` usually useful to have for features that may have been backported from a newer version of Python

hoxbro: Agreed, just copied the dependencies over from pyproject.

MarcSkovMadsen: I've added typing-extensions.

pyproject.toml
MarcSkovMadsen: I've removed the line.

pyproject.toml
```diff
@@ -207,3 +218,50 @@ filterwarnings = [
     "ignore:unclosed file <_io.TextIOWrapper name='(/dev/null|nul)' mode='w':ResourceWarning", # OK
     "ignore:Deprecated in traitlets 4.1, use the instance .metadata dictionary directly", # OK (ipywidgets internal)
 ]
+
+[tool.mypy]
+namespace_packages = true
+explicit_package_bases = true
+mypy_path = ""
+exclude = []
```
gandhis1, line 226: I ran Mypy in the same way you were @MarcSkovMadsen and I got a bunch of additional errors on unrelated files. At least for now, maybe it's worth turning off the unrelated file errors?

```suggestion
exclude = []
follow_imports = "silent"
```

From https://mypy.readthedocs.io/en/stable/running_mypy.html#following-imports:

> If you are starting a new codebase and plan on using type hints from the start, we recommend you use either [--follow-imports=normal](https://mypy.readthedocs.io/en/stable/command_line.html#cmdoption-mypy-follow-imports) (the default) or [--follow-imports=error](https://mypy.readthedocs.io/en/stable/command_line.html#cmdoption-mypy-follow-imports). Either option will help make sure you are not skipping checking any part of your codebase by accident.

> If you are planning on adding type hints to a large, existing code base, we recommend you start by trying to make your entire codebase (including files that do not use type hints) pass under [--follow-imports=normal](https://mypy.readthedocs.io/en/stable/command_line.html#cmdoption-mypy-follow-imports). This is usually not too difficult to do: mypy is designed to report as few error messages as possible when it is looking at unannotated code.

> **Only if doing this is intractable, we recommend passing mypy just the files you want to type check and use [--follow-imports=silent](https://mypy.readthedocs.io/en/stable/command_line.html#cmdoption-mypy-follow-imports).** Even if mypy is unable to perfectly type check a file, it can still glean some useful information by parsing it (for example, understanding what methods a given object has). See [Using mypy with an existing codebase](https://mypy.readthedocs.io/en/stable/existing_code.html#existing-code) for more recommendations.

panel/io/cache.py
```diff
@@ -427,7 +465,7 @@ async def wrapped_func(*args, **kwargs):
             return ret
     else:
         @functools.wraps(func)
-        def wrapped_func(*args, **kwargs):
+        def wrapped_func(*args: P.args, **kwargs: P.kwargs) -> R:
```
gandhis1, line 468: I think you may have inadvertently included this file?

pyproject.toml
MarcSkovMadsen: I don't think we should change it with the argument that what we want to check in the pipeline is everything as that is the end goal for the Panel project to have no issues whether internal facing or external facing.

I believe for "the current effort" which is focused on outward facing api we should add the relevant settings via flags manually when running mypy manually. Makes sense?

panel/io/cache.py
MarcSkovMadsen: Thx. I believe I've "undone" it by running `git checkout main -- panel/io/cache.py`.

pixi.toml
```diff
@@ -144,6 +145,22 @@ _install-ui = 'playwright install chromium'
...
+types-bleach = "*"
+types-croniter = "*"
+types-Markdown = "*"
+types-psutil = "*"
+types-requests = "*"
+types-tqdm = "*"
```
hoxbro, line 159:
```suggestion
types-tqdm = "*"
typing-extensions = "*"
```

gandhis1 (approved):
LGTM

hoxbro (approved):

philippjfr (approved):

PR #7083, opened 2024-08-06
Add a standalone Modal layout

### LINKED ISSUE #4700: Add a standalone Modal layout
I know the Templates have a modal. But its also powerful to be able to use and test it in notebooks. And its also much easier to search and learn about it as a separate component.

I've had such a great success with https://github.com/awesome-panel/panel-modal for my own Apps at work. Its just so easy to use and it always works.

There are some small changes in the api I would make and some more features from the underlying js library I would support if contributed.

I would like to contribute it to Panel if you would be willing to accept it?

hoxbro:
Fixes #4700 

Heavily inspired by https://github.com/awesome-panel/panel-modal. Thank you @MarcSkovMadsen :heart: 

``` python
import panel as pn
import holoviews as hv
from panel.layout.modal import Modal

pn.extension("modal")
hv.extension("bokeh")

reviews = "test"
close_method = pn.widgets.Button(name="Close (method)")
open_method = pn.widgets.Button(name="Open (method)").servable()
close_parameter = pn.widgets.Button(name="Close (parameter)")
open_parameter = pn.widgets.Button(name="Open (parameter)").servable()
modal = Modal(reviews, close_method, close_parameter, hv.Curve([]), width=1000, height=1500).servable()

close_method.on_click(lambda event: modal.hide())
open_method.on_click(lambda event: modal.show())
close_parameter.on_click(lambda event: setattr(modal, "open", False))
open_parameter.on_click(lambda event: setattr(modal, "open", True))

close_method.on_click(lambda event: print(modal.open))
open_method.on_click(lambda event: print(modal.open))
close_parameter.on_click(lambda event: print(modal.open))
open_parameter.on_click(lambda event: print(modal.open))

```

https://github.com/user-attachments/assets/3749cb8b-1674-42d0-83ca-d72c3d48381f

TO DO:
- [x] Add tests
- [x] Add documentation
- [x] Look into objects showing up before the modal
- [x] Improve height and width
- [x] Maybe look into `window as any`.


panel/layout/modal.py
```diff
@@ -0,0 +1,42 @@
...
+if TYPE_CHECKING:
+    from bokeh.model import Model
+
+
+class Modal(ListPanel):
+    height = param.Integer(default=None, bounds=(0, None))
+
+    width = param.Integer(default=None, bounds=(0, None))
+
+    is_open = param.Boolean(default=False, readonly=True, doc="Whether the modal is open.")
```
MarcSkovMadsen, line 21: Love this PR.

I've experienced that for some use cases I would like the `open` and `close` methods. For some it was better if I could set `is_open=True` or `is_open=False`. Please consider the last feature :-)

panel/layout/modal.py
```diff
@@ -0,0 +1,42 @@
+from __future__ import annotations
```
MarcSkovMadsen, line 1: When adding documentation please consider if request https://github.com/awesome-panel/panel-modal/issues/3 should be part of the docs. Thx.

edoerpani (approved):

contang0:
This is an excellent addition. I'm a heavy user of panel_modal, and always wished that it shipped together with panel. Modal windows allow me to save a lot of space in the app.

One thing I couldn't get panel_modal to do is to update graphs once they have been created (Tabulator meanwhile updates fine), I hope this development will address it.

panel/layout/modal.py
contang0: @MarcSkovMadsen I also missed these two methods in panel_modal. I ended up using `modal._show()`.

ahuang11:
Anything I can help with to get this PR merged?

hoxbro:
> Anything I can help with to get this PR merged?

No, but thank you for asking. I will soon start working on this again.

panel/layout/modal.py
```diff
@@ -0,0 +1,62 @@
...
+
```
MarcSkovMadsen, line 2: One thing to consider is if the `Modal` could be made easier to use? Could the api of the `Modal` be generalized to a concept of *actions*/ *action button*.

## Could the api be simplified?

The api of `panel-modal` that I made is:

```python
import panel as pn

from panel_modal import Modal

pn.extension("modal")

modal = Modal(pn.panel("Hi. I am the Panel Modal!", width=200))

pn.Column(modal.param.open, modal).servable()
```

I requires you to include both the button (`modal.param.open`) and the modal `modal`. If you want to customize the button it requires something like `pn.widgets.Button.from_param(modal.param.open`.

Could the api be simplified? Could it be enough to just include the modal?

```python
import panel as pn

from panel_modal import Modal

pn.extension("modal")

pn.Column(
    Modal(pn.panel("Hi. I am the Panel Modal!", width=200))
).servable()
```

That should show the button. And when clicked the modal should open/ close. That would make it much simpler to drop in as you don't need to define a variable holding the `Modal`.

If you want to customize the button you should be able to provide a button or a dictionary of button kwargs.

```python
import panel as pn

from panel_modal import Modal

pn.extension("modal")

pn.Column(
    Modal(
        pn.panel("Hi. I am the Panel Modal!", width=200),
        button=pn.widgets.Button(...., button_style="outline") # could also be = {..., button_style="outline"}
)
).servable()
```

## Could the API be generalized?

A Modal is actually a part of a general concept for *action buttons*. You want to be able to trigger an action via a Button. You want to be able to easily set the content (`object`) and be able to customize the button.

```python
modal_button = ModalButton(object=pn.panel(...), button=...)
copy_button= CopyPasteButton(object="...my code", button=...)
paste_button = PasteButton(object=pn.widgets.Tabulator, button=...., ) # Here the object is the target. Not the source.
link_button = LinkButton(object="https://panel.holoviz.org, target="_blank", button=...)

maximize_content = pn.Column(...)
maximize_button = MaximizeButton(object=maximize_content, button=...) # Maximizes the object to full window size
...
```

## Alternative Api

Add a method - for example `.layout` to include then `object` if relevant.

```python
import panel as pn

from panel_modal import Modal

pn.extension("modal")

pn.Column(
    Modal(
        pn.panel("Hi. I am the Panel Modal!", width=200),
        button=pn.widgets.Button(...., button_style="outline") # could also be = {..., button_style="outline"}
).layout()
).servable()
```

@philippjfr. What do you think?

panel/dist/css/models/modal.css
```diff
@@ -0,0 +1,86 @@
...
+  font-size: 1.5em;
+  width: 1.5em;
+  height: 1.5em;
+  text-align: center;
+  cursor: pointer;
+  transition: 0.15s;
+  border-radius: 50%;
+  z-index: 100003;
+}
+fast-design-system-provider .pnx-dialog-close {
```
philippjfr, line 75: Should move into `panel/theme/css/fast.css` I think.

philippjfr: Same with the other fast related CSS.

panel/layout/__init__.py
```diff
@@ -38,6 +38,7 @@
 from .float import FloatPanel  # noqa
 from .grid import GridBox, GridSpec  # noqa
 from .gridstack import GridStack  # noqa
+from .modal import Modal
```
hoxbro, line 41: Should we also add it to root, so we can `pn.Modal`?

panel/layout/modal.py
```diff
@@ -0,0 +1,74 @@
...
+    def hide(self):
+        self.open = False
+
+    def toggle(self):
+        self.open = not self.open
+
+    @param.depends("open", watch=True)
+    def _open(self):
+        self._send_event(ModalDialogEvent, open=self.open)
```
hoxbro, line 57: Maybe we should have a warning here if we are in a server setting (`panel serve`) and the Modal has not been `servable()`ed. 

Is there a way to detect this?

philippjfr: `self._models is None` would be simplest but technically you'd have to check if any of the `self._models` values are in the `pn.state.curdoc`.

panel/layout/__init__.py
philippjfr: Good question, we do it for other layouts, so for consistency the answer should be yes I suppose.

philippjfr:
The `panel/dist/css/models/modal.css` should be moved into `panel/styles/models/modal.less` and then be included in the model stylesheets like the other models that use less files (e.g. plotly or tabulator).

panel/layout/modal.py
philippjfr: Checking `if not self._models: warn` seems sufficient actually.

panel/layout/modal.py
philippjfr: Sorry I never responded, especially as we're working towards merging this. My feeling is that while I can see how it might be useful it kind of breaks a few mental models people have built. A few more lines of code seem like an okay price to pay for keeping a relatively consistent mental model for how things worked. For a `Modal` to render as a button and something like `pn.widgets.Button.from_param(Modal())` to work we'd also internally have to break a lot of models on how things work, so I'm roughly -1 on this, though it's something we could think about adding later.

The one thing we could consider doing to make this potential option backward compatible would be to only allow the modal to be rendered as a root, i.e. `pn.Row(modal).servable()` would be disallowed and `modal.servable()` would be the only valid way to add a modal. What do you think @hoxbro?

hoxbro: I added `Modal.create_button` after this comment to make it easier to use the Modal. 

I'm not sure about the suggested API. I'm not hard against it. If you still want this, maybe open a new issue to discuss it.

philippjfr (approved):
Looks good!

bytecodesygnul (approved):

Issue #6738, opened 2024-04-17
CodeEditor widget is unusable when automatic theading is enabled


chris-branch:
#### ALL software version info
python 3.11.9
panel 1.4.0
bokeh 3.4.0
Chrome 123.0.6312.107 on Windows 11

#### Description of expected behavior and the observed behavior
The expected behavior is that I can run the application code below, type inside the CodeEditor widget, and the editor works as you'd expect for a typical text input field (e.g., holding down the 'x' key will continuously append additional 'x' characters while the key is held down).

Actual behavior is that holding down the 'x' key will add some 'x' characters, then delete the entire contents of the field so that it's empty, then start over appending 'x' characters.  This repeats over and over, though the interval for text deletion is somewhat random.  It also also happens intermittently with normal text input (not just holding down a key), but holding down a single key is the easiest/fastest way to reproduce.

If the `nthreads=0` config option is removed or commented out, then CodeEditor works as expected (no issues). The presence of nthreads with any value (including 1) will cause all existing text to be deleted periodically as new keystrokes occur (i.e., it acts as if the user performed Select All followed by Backspace).

Based on the documentation, enabling automatic theading via the `nthreads` option means that the order in which events are processed becomes non-deterministic. It appears that CodeEditor will send a `ModelChanged` event for every keystroke (to update the model on the backend), and my theory is that the problem occurs if/when these `ModelChanged` events are processed out-of-order. 

https://panel.holoviz.org/how_to/concurrency/threading.html

#### Complete, minimal, self-contained example code that reproduces the issue
```
import panel as pn
pn.extension(
    'codeeditor',
    nthreads=0,
)

template = pn.template.FastListTemplate()
template.main.append(
    pn.Row(pn.widgets.CodeEditor(language='yaml', sizing_mode='stretch_both')),
)

pn.serve(
    {
        'bug': template,
    },
    port=8080,
    show=True,
    start=True,
    websocket_origin=['localhost:8080', '127.0.0.1:8080'],
)
```
Save as `bug.py`, and execute as `python -m bug.py`

Potentially related to this issue: https://github.com/holoviz/panel/issues/6414

chris-branch:
Are there any plans to address this?

Issue #6414, opened 2024-02-29
TextInput widget loses user input text when typing quickly with a thread pool


ronan-delta:
#### ALL software version info
panel 1.3.8
bokeh 3.3.4

#### Description of expected behavior and the observed behavior
When using the pn.widgets.TextInput widget either alone, as the widget for a param.String, or in the new chat interface I run into an issue where the some of the characters of the input text are lost when the user is typing quickly. 

You can see the text very briefly then it'll look like it's rolled back to a split second before the latest character or two is entered. For example, you will be typing something like "hello world" and all characters will appear then a split second later it will show "hello worl" with the cursor now after the "l". 

#### Complete, minimal, self-contained example code that reproduces the issue
Below is a simple example that appears to replicate the issue on my end, either served locally or on a hosted server. It maybe related to the nthreads parameter as it appears to at least be far less common when that parameter is None/no thread pool is created. 
```
import panel as pn

pn.extension(
    nthreads=0,
)

def dashboard_page():
    return pn.widgets.TextInput(name="Enter text", placeholder="Type here...")

ROUTES = {
    "Dashboard": dashboard_page,
}
pn.serve(ROUTES, port=8501, admin=True, static_dirs={"thumbnails": "./static/thumbnails"})
```

#### Stack traceback and/or browser JavaScript console output
N/A
#### Screenshots or screencasts of the bug in action
Difficult to capture in action, but seems to happen frequently enough when typing fast. 

- [X] I may be interested in making a pull request to address this, but unsure where to start on this one.

ahuang11:
I can reproduce this.

chris-branch:
Same or similar issue with CodeEditor:

https://discourse.holoviz.org/t/codeeditor-widget-is-unusable-when-automatic-theading-is-enabled/7111

duarteocarmo:
Hey @philippjfr  any idea when this will be available? (I know 1.4.5, just wondering if there is a release date for that, or an option to release it sooner?) 

Thanks!

PR #7093, opened 2024-08-07
Avoid events boomeranging from frontend

philippjfr:
This adds a long overdue change which ensures that events arriving from the frontend do not boomerang. There was a previous attempt at doing this but instead of skipping events being applied that originated on the frontend it cancelled the processing of new events after a property change was made. These approaches have potentially similar effects but in fact are quite different, since the old change effectively would potentially cause a property to revert to an older value.

Fixes https://github.com/holoviz/panel/issues/7096
Fixes https://github.com/holoviz/panel/issues/6738
Fixes https://github.com/holoviz/panel/issues/6414

### DIFF
--- a/panel/reactive.py
+++ b/panel/reactive.py
@@ -330,13 +330,17 @@ class Syncable(Renderable):
         ref = root.ref['id']
         self._changing[ref] = attrs = []
         for attr, value in dict(msg).items():
-            # Do not apply model change that is in flight
             if attr in self._in_process__events and value is self._in_process__events[attr]:
+                # Do not apply change that originated directly from
+                # the frontend since this may cause boomerang if a
+                # new property value is already in-flight
                 del self._in_process__events[attr]
                 del msg[attr]
                 continue
             elif attr in self._events:
-                del msg[attr]
+                # Do not override a property value that was just changed
+                # on the frontend
+                del self._events[attr]
                 continue
 
             # Bokeh raises UnsetValueError if the value is Undefined.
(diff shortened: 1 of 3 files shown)


MarcSkovMadsen:
Will this solve issue with ChatInterface? I.e. the problem that if you don't type slowly then text will disappear?

philippjfr:
>Will this solve issue with ChatInterface? I.e. the problem that if you don't type slowly then text will disappear?

I have never encountered that problem, the problem it is meant to solve is that the chat feed scroll position isn't stable.

MarcSkovMadsen:
I will make a separate issue then.

philippjfr:
A reproducer would be very helpful. I definitely can't reproduce this with a bare chat interface.

MarcSkovMadsen:
Hmmm. The problem is only with the deployed app. When developing on jupyter hub with proxy server there is no problem. That is strange as it should be the same. But I also experience my deployed apps being slower even though they have the same resources available.

philippjfr:
Yeah literally never seen it and we have many chat apps in production with different clients.

duarteocarmo:
Hey @philippjfr any idea when this will be available? (I know 1.4.5, just wondering if there is a release date for that, or an option to release it sooner?)

Thanks!

duarteocarmo:
We are observing the same issue as @MarcSkovMadsen in #7096

philippjfr:
I was planning to cut 1.5.0rc1 today and then make the proper release some time next week.

duarteocarmo:
Awesome - appreciate it. 👍🏻

PR #7094, opened 2024-08-07
Hide typing constructs behind TYPE_CHECKING guard

gandhis1:
@philippjfr Clean up to adopt @hoxbro feedback on the other PR

I deliberately left `ProfilingEngine` public as I didn't want to preclude the possibility of someone actually using this type themselves in their own code.

### DIFF
--- a/panel/io/profile.py
+++ b/panel/io/profile.py
@@ -1,3 +1,5 @@
+from __future__ import annotations
+
 import io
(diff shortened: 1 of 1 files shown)


philippjfr (approved):

philippjfr:
Much appreciated, thanks for following up!

PR #7105, opened 2024-08-08
[docs] Remove extraneous cell from Param Component Gallery example

sjdemartini:
The cell with `Select` and `RadioButtonGroup` seems to be there by mistake. Its variables etc aren't used elsewhere, and it doesn't relate to the rest of the Athlete/Power Curve example. (Noticed this looking here https://panel.holoviz.org/reference/panes/Param.html)


philippjfr (approved):

philippjfr:
Good catch, thank you!

PR #7088, opened 2024-08-07
hello world

MarcSkovMadsen:

PR #7109, opened 2024-08-09
Optimize rendering of children in JSComponent

philippjfr:
Avoids a full re-render of JSComponent children if they are added to the same node and in sequence.

### DIFF
--- a/panel/models/reactive_esm.ts
+++ b/panel/models/reactive_esm.ts
@@ -154,6 +154,7 @@ export class ReactiveESMView extends HTMLBoxView {
     ["after_resize", []],
     ["remove", []],
   ])
+  _parent_map: Map<string, ParentNode> = new Map()
   _rendered: boolean = false
 
   override initialize(): void {
@@ -291,7 +292,8 @@ Promise.resolve(output).then((out) => {
       cb()
     }
     this.render_children()
-    this.model_proxy.on(this.accessed_children, () => this.render_esm())
+    const rerender = this.accessed_children.filter((child) => !this._parent_map.has(child))
+    this.model_proxy.on(rerender, () => this.render_esm())
     this._rendered = true
   }
 
@@ -300,6 +302,7 @@ Promise.resolve(output).then((out) => {
       return
     }
     this.accessed_properties = []
+    this._parent_map.clear()
     for (const lf of this._lifecycle_handlers.keys()) {
       (this._lifecycle_handlers.get(lf) || []).splice(0)
     }
@@ -316,15 +319,30 @@ Promise.resolve(output).then((out) => {
     for (const child of this.model.children) {
       const child_model = this.model.data[child]
       const children = isArray(child_model) ? child_model : [child_model]
+      const nodes = []
       for (const subchild of children) {
         const view = this._child_views.get(subchild)
         if (!view) {
           continue
         }
         const parent = view.el.parentNode
+        nodes.push(parent)
         if (parent) {
           view.render()
           view.after_render()
+        }
+      }
+      if ((new Set(nodes)).size === 1 && nodes[0] != null) {
+        const parent = nodes[0]
+        let in_sequence = true
+        for (let i=0; i<nodes.length; i++) {
+          if (parent.children[i] !== this._child_views.get(children[i])?.el) {
+            in_sequence = false
+            break
+          }
+        }
+        if (in_sequence && parent) {
+          this._parent_map.set(child, parent)
         }
       }
     }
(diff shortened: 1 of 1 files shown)


MarcSkovMadsen:
How can we document What users should do. I dont understand the requirements for efficiency.

philippjfr:
They don't have to do anything.

MarcSkovMadsen:
But you write "if they are added to the same node and in sequence". That must mean I have to do something specific to achieve effiency. Or that I can do inefficient things?

philippjfr:
Yes sorry, was drafting another reply and then got pulled away. That wasn't very helpful and it's a good question. This would be optimized:

```javascript
const div = document.createElement('div') 
div.append(...model.get_child('children'))
return div
```

this wouldn't be because we can't automatically handle the wrapping elements:

```javascript
const div = document.createElement('div') 
const children = model.get_child('children')
for (const child of children) {
    const child_div = document.createElement('div')
    div.append(child_div)
}
return div
```

so we have to re-run the render function. I'm not yet sure how best to document that but will think about it.

philippjfr:
Will merge for now and think about ways to make this clear in the docs.

MarcSkovMadsen:
How should the below example be done *efficiently*?

```python
import panel as pn
import param
from panel.custom import JSComponent
from panel.layout.base import NamedListLike

pn.extension()


class ListLikeLayout(NamedListLike, JSComponent):
    objects = param.List()

    _esm = """
    export function render({ model }) {
      const div = document.createElement('div')
      let objects = model.get_child("objects")

      objects.forEach((object, index) => {
        div.appendChild(object);

        // If it's not the last object, add a divider
        if (index < objects.length - 1) {
            const divider = document.createElement("div");
            divider.className = "divider";
            div.appendChild(divider);
        }
        });
      return div
    }"""

    _stylesheets = [
        """
.divider {border-top: 3px dotted #bbb};
"""
    ]


ListLikeLayout(
    "I love beat boxing",
    "https://upload.wikimedia.org/wikipedia/commons/d/d3/Beatboxset1_pepouni.ogg",
    "Yes I do!",
    styles={"border": "2px solid lightgray"},
).servable()
```

MarcSkovMadsen:
There is another example with two `Child` parameters that seem inefficient too:

https://github.com/holoviz/panel/pull/7088#discussion_r1712543585

MarcSkovMadsen:
And this is a similar example. How would you optimize that?

```python
import panel as pn
import param
from panel.custom import JSComponent
from panel.layout.base import NamedListLike

CSS = """
.gutter {
    background-color: #eee;
    background-repeat: no-repeat;
    background-position: 50%;
}
.gutter.gutter-vertical {
    background-image: url('data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAAB4AAAAFAQMAAABo7865AAAABlBMVEVHcEzMzMzyAv2sAAAAAXRSTlMAQObYZgAAABBJREFUeF5jOAMEEAIEEFwAn3kMwcB6I2AAAAAASUVORK5CYII=');
    cursor: row-resize;
}
"""

# https://github.com/nathancahill/split/issues/799

class ListLikeLayout(NamedListLike, JSComponent):
    objects = param.List()

    _esm = """
    import Split from 'https://esm.sh/split.js@1.6.5'

    export function render({ model, el }) {
      const objects = model.get_child("objects")
      
      const splitDiv = document.createElement('div');
      splitDiv.className = 'split';
      splitDiv.style.height = 'calc( 100% - 2*10px )';
      el.append(splitDiv)
      
      let splits = [];
      
      objects.forEach((object, index) => {
        const split = document.createElement('div');
        splits.push(split)
        
        splitDiv.appendChild(split);
        split.appendChild(object);
      })
      
      Split(splits, {direction: 'vertical'})    
    }"""

    _stylesheets = [CSS]

pn.extension("codeeditor")

layout = ListLikeLayout(
    pn.widgets.CodeEditor(value="I love beat boxing\n"*10, theme="monokai", sizing_mode="stretch_both"),
    pn.panel("https://upload.wikimedia.org/wikipedia/commons/d/d3/Beatboxset1_pepouni.ogg", sizing_mode="stretch_both"),
    pn.widgets.CodeEditor(value="Yes I do!\n"*10, theme="monokai", sizing_mode="stretch_both"),
    styles={"border": "2px solid lightgray"},
    height=800,
    width=500,
    sizing_mode="fixed",
).servable()
```

MarcSkovMadsen:
My guess is that `div.append(...model.get_child('children'))` is not really what use cases need. We can allready append children in a `Column`, `Row` or `FlexBox` layout. Custom layouts will be looking to do something else.

MarcSkovMadsen:
The React version of the grid cannot be appended to or popped. https://github.com/holoviz/panel/pull/7088#discussion_r1712562021

PR #7112, opened 2024-08-09
Fix 'dictionary chaged size during update' in panel/base

jeffrey-hicks:
I've been using Panel to make and dashboard and have gotten periodic crashes.  This seems to be the exact same type of error in https://github.com/holoviz/panel/issues/2631 so I just copied its solution for the panel/base/_update_pane.


philippjfr (approved):
Thanks for catching this one!

PR #7124, opened 2024-08-10
Simplify and generalize JSComponent child re-rendering

philippjfr:
Reverts https://github.com/holoviz/panel/pull/7109 and finds a general solution to avoid unnecessary re-rendering of children.


philippjfr:
Textual test failures are unrelated.

Issue #6603, opened 2024-03-28
Sneaky Race condition using asyncio create_task (reliably reproduced)


tomascsantos:
#### Context
Hey team, I'm back with I think perhaps the same bug as: https://github.com/holoviz/panel/issues/6019
I've managed to significantly simplify the minimum reproducible code, and I suspect that the last issue was really caused by this bug as my code was not fixed in the patch that was released to address the previous ticket. (Thank you for trying!)

I've been stuck on panel==1.3.4 because I think this race condition becomes worse starting from 1.3.5 onwards. I saw the awesome changes in 1.4.0 and would love to jump to the latest version.

#### ALL software version info
panel==1.4.0 (error is much less likely but still possible to reproduce with 1.3.4)
bokeh==3.4.0
Python==3.11.5
param==2.1.0

#### Description of expected behavior and the observed behavior
The final visible "state" should be "SUCCESS", instead it gets stuck in "LOADING". In the code below, I add some comments indicating how to increase or decrease the likelihood of the race condition appearing. It's tuned right now to fail every time. 

#### Complete, minimal, self-contained example code that reproduces the issue

```python
"""A minimum reproducible example of a timing bug"""
import asyncio
import panel as pn
import param
from structlog import get_logger
log = get_logger(__name__)

TERMINAL_STATES = ['ERROR', 'SUCCESS', 'WARNING']

class StateMachine(param.Parameterized):
    """A simple state machine for fetching data"""
    state = param.String(default='SUCCESS')

    async def fetch_data_simple(self):
        """Would normally return a dataframe, not included for MVP"""
        self.state = 'STARTING'
        await asyncio.sleep(.01)  # the smaller this number the more likely, increasing to 1 fixes.
        self.state = 'SUCCESS'


class QueryViewer(pn.viewable.Viewer):
    """A viewer which transitions views based on the state machine"""

    box: StateMachine = param.ClassSelector(class_=StateMachine)

    def __init__(self, box, **params) -> None:
        super().__init__(box=box, **params)

        self.container = pn.Column()
        self.loading = pn.Column('LOADING', styles={'background-color': 'aqua'})
        self.green = pn.Column('SUCCESS', styles={'background-color': 'green'})

        self._update_state()
        self.box.param.watch(self._update_state, 'state')

    def _update_state(self, event=None):
        """Update the viewer state based on the state machine"""
        log.info('_update_state', state=self.box.state)
        if self.box.state == 'SUCCESS':
[...]

PR #7125, opened 2024-08-10
Allow streaming chunks to HTML panes

philippjfr:
Often, especially for the chat interfaces you are slowly streaming chunks to the frontend. Currently however, we send the entire text every time any of it is updated. This PR proposes to add a `stream` parameter to HTML based components, e.g. like `HTML` itself or `Markdown`. When enabled the pane will find the length of overlap and then simply sends the non-overlapping chunk to to the frontend using a server-side event, massively reducing the amount of text sent to the frontend. Since this can lead to super rapid updates we currently batch the incoming messages and apply the update at most every 10 ms.

Unfortunately for large messages this quickly runs into this issue: https://github.com/holoviz/panel/issues/6603

### DIFF
--- a/examples/reference/panes/HTML.ipynb
+++ b/examples/reference/panes/HTML.ipynb
@@ -22,6 +22,7 @@
     "For details on other options for customizing the component see the [layout](../../how_to/layout/index.md) and [styling](../../how_to/styling/index.md) how-to guides.\n",
     "\n",
     "* **`disable_math`** (boolean, `default=True`): Whether to disable MathJax math rendering for strings escaped with `$$` delimiters.\n",
+    "* **`enable_streaming`** (boolean, `default=False`): Whether to enable streaming of text snippets. This will diff the `object` when it is updated and only send the trailing chunk that was added.\n",
     "* **`object`** (str or object): The string or object with ``_repr_html_`` method to display\n",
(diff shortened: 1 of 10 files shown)


MarcSkovMadsen:
Is there not a more general way to diff a string based on position?

Then instead of s new stream parameter we could keep using the value parameter.

philippjfr:
I think you're misunderstanding the PR, the `stream` parameter is a boolean toggle which determines whether to send diffs or not.

philippjfr:
Okay, sorry reading back your reply I misread it. Generally I agree and I'd be okay with a `mode` or `diff_mode`, that said streaming is really the main scenario where you send tons of small individual updates. If you're editing a large document you might want to send smaller patches but finding the patches becomes computationally expensive and in many cases may not be worth it.

MarcSkovMadsen:
Makes sense. 

If `stream` is the only additional *mode* we will ever support then calling it `stream` is fine.

philippjfr:
>I think my context is coming from streamlit. My understanding is that they have a virtual document on the python side and only send diff patches using protobuf.

Do you know of a good writeup of this? This sounds more analogous to the bokeh document than diffing of the contents of an individual update but I'd love to know more.

PR #7135, opened 2024-08-12
Don't install mkl in Windows

hoxbro:
Notice we install mkl in Windows environments, which is likely why download is significantly slower on Windows.

### DIFF
--- a/pixi.toml
+++ b/pixi.toml
@@ -22,18 +22,20 @@ build = ["py311", "build"]
 lint = ["py311", "lint"]
 
 [dependencies]
+nodejs = ">=18"
+nomkl = "*"
+pip = "*"
+# Required
 bleach = "*"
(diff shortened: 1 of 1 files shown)


pixi.toml
```diff
@@ -22,18 +22,20 @@ build = ["py311", "build"]
...
+pip = "*"
+# Required
 bleach = "*"
 bokeh = ">=3.5.0,<3.6.0"
 linkify-it-py = "*"
 markdown = "*"
 markdown-it-py = "*"
 mdit-py-plugins = "*"
-nodejs = ">=18"
 numpy = "<2.0" # Temporary pin until panel release with support for bokeh 3.5.0 is available
```
hoxbro, line 35: Also wants to try and remove this.

philippjfr (approved):

Issue #7116, opened 2024-08-10
pixi run test-example picks up old Panel code version


cdeil:
I'm on 9404b4348a80a190b3dcda7f270ba3e5b3c10210 and followed instructions here:
https://holoviz-dev.github.io/panel/developer_guide/index.html#example-tests

I get this error:
```
(panel) (panel) panel $ pixi run test-example
✨ Pixi task (test-example in test-312): pytest -n logical --dist loadscope --nbval-lax examples
=============================================================================================== test session starts ================================================================================================
platform darwin -- Python 3.12.5, pytest-8.3.2, pluggy-1.5.0
rootdir: /Users/cdeil/code/oss/panel
configfile: pyproject.toml
plugins: asyncio-0.23.8, cov-5.0.0, github-actions-annotate-failures-0.2.0, rerunfailures-14.0, anyio-4.4.0, nbval-0.11.0, xdist-3.6.1
asyncio: mode=Mode.AUTO
10 workers [987 items]    
............................................................................................................................................................................................................ [ 20%]
............................................................................................................................................................................................................ [ 41%]
............................................................................................................................................................................................................ [ 62%]
.....................................................................................................................FFFF.............FFFF.................................................................. [ 82%]
...........................................................................................................................................................................                                  [100%]
===================================================================================================== FAILURES =====================================================================================================
_____________________________________________________________________________ examples/reference/widgets/DiscretePlayer.ipynb::Cell 1 ______________________________________________________________________________
[gw4] darwin -- Python 3.12.5 /Users/cdeil/code/oss/panel/.pixi/envs/test-312/bin/python3.12
Notebook cell execution failed
Cell 1: Cell execution caused an exception

Input:
discrete_player = pn.widgets.DiscretePlayer(name='Discrete Player', options=[2, 4, 8, 16, 32, 64, 128],
                                            value=8, loop_policy='loop', show_value=True, value_align='start')

discrete_player

Traceback:

---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
Cell In[1], line 1
----> 1 discrete_player = pn.widgets.DiscretePlayer(name='Discrete Player', options=[2, 4, 8, 16, 32, 64, 128],
      2                                             value=8, loop_policy='loop', show_value=True, value_align='start')
      4 discrete_player

File ~/code/oss/panel/.pixi/envs/test-312/lib/python3.12/site-packages/panel/widgets/player.py:55, in PlayerBase.__init__(self, **params)
[...]

PR #7147, opened 2024-08-14
Remove show in ChatFeed.ipynb

cdeil:
See https://github.com/holoviz/panel/issues/7116#issuecomment-2282201500


ahuang11 (approved):

ahuang11:
Thanks for catching that!

PR #7149, opened 2024-08-15
Add client-side rendering option for Markdown pane based on MyST

philippjfr:
This PR adds a new client-side rendering option to the `Markdown` pane. Specifically you can now set `renderer='mystjs'` and the Markdown will be transformed into HTML clientside, which has number of benefits:

- The `Markdown` value can be jslinked with e.g. an editor, without requiring a roundtrip to Python
- Markdown generally is less verbose than HTML so it should be more efficient
- Together with the recent support for streaming it should now be even more efficient


philippjfr:
Okay, I only just realized mystjs uses an ancient version of myst 😢

MarcSkovMadsen:
Its a great beginning. A lot of Markdown is probably pretty simple and supported by the ancient version.

We can add a comment or request for an update at the mystjs github.

MarcSkovMadsen:
Also remember to think about if plugins are a thing for mystjs and whether we support it.

philippjfr:
Okay, figured it out, this now uses latest myst-parser and myst-to-html + highlight.js for syntax highlighting. Works great, still some alignment on code styling needed but overall I'm quite happy.

panel/config.py
```diff
@@ -689,6 +690,7 @@ class panel_extension(_pyviz_extension):
         'gridstack': ['GridStack'],
         'katex': ['katex'],
         'mathjax': ['MathJax'],
+        'myst': ['mystjs'],
```
philippjfr, line 693:
```suggestion
        'myst': ['mystparser'],
```

philippjfr:
Tested with the streaming support and as expected it does zero backtracking now, i.e. it only ever sends the newly added snippets rather than having to go back and modify the previous text.

philippjfr:
> Also remember to think about if plugins are a thing for mystjs and whether we support it.

Yes, but not really. It's a pretty opinionated Markdown implementation with lots of features supported out-of-the-box.

MarcSkovMadsen:
Have you tested in different templates whether it looks ok?

PR #7151, opened 2024-08-15
Enable test of types in CI

hoxbro:
Currently, we just ran `mypy`, this will make it so it fails if type testing fails. I have, set it up so you manually have to enable modules. 

Copied settings from: https://github.com/bokeh/bokeh/blob/bba83ab0db6986f5b8e82ab01e2ad2d09b877cea/pyproject.toml

Commented out strict for now, as it have many failing tests. Currently, I see two failing tests.  

Currently two errors, the changes in the codebase is to fix other errors. 

```
panel/io/cache.py: note: In function "cache":
panel/io/cache.py:472:23: error: Incompatible types in "await" (actual type "_R", expected type "Awaitable[Any]")  [misc]
                    ret = await func(*args, **kwargs)
                          ^~~~~~~~~~~~~~~~~~~~~~~~~~~
panel/io/cache.py:515:12: error: Incompatible return value type (got "_Wrapped[_P, _R, _P, Coroutine[Any, Any, _R]]", expected "_CachedFunc[Callable[_P, _R]] | Callable[[Callable[_P, _R]], _CachedFunc[Callable[_P, _R]]]")  [return-value]
        return wrapped_func
               ^~~~~~~~~~~~
Found 2 errors in 1 file (checked 341 source files)
```

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -270,4 +270,4 @@ jobs:
           environments: ${{ matrix.environment }}
       - name: Test Type
         run: |
-          pixi run -e ${{ matrix.environment }} test-type || echo "Failed"
+          pixi run -e ${{ matrix.environment }} test-type
--- a/panel/io/cache.py
+++ b/panel/io/cache.py
@@ -15,7 +15,6 @@ import sys
 import threading
 import time
 import unittest.mock
-import weakref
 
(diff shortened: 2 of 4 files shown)


philippjfr:
Superseded by https://github.com/holoviz/panel/pull/7497

PR #7154, opened 2024-08-15
ChatFeed.add_step default to assistant

ahuang11:
I think it makes more sense that the steps would be used more with an Assistant LLM vs User.

### DIFF
--- a/panel/chat/feed.py
+++ b/panel/chat/feed.py
@@ -801,7 +801,7 @@ class ChatFeed(ListPanel):
             if layout_params:
                 input_layout_params.update(layout_params)
             steps_layout = layout(step, **input_layout_params)
-            self.stream(steps_layout, user=user, avatar=avatar)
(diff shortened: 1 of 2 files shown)


panel/chat/feed.py
```diff
@@ -801,7 +801,7 @@ def add_step(
             if layout_params:
                 input_layout_params.update(layout_params)
             steps_layout = layout(step, **input_layout_params)
-            self.stream(steps_layout, user=user, avatar=avatar)
+            self.stream(steps_layout, user=user or "Assistant", avatar=avatar)
```
philippjfr, line 804: Shouldn't this be:

```suggestion
            self.stream(steps_layout, user=user or self.callback_user, avatar=avatar)
```

ahuang11: Good point!

philippjfr (approved):

PR #7177, opened 2024-08-21
Fix CodeEditor UI tests

philippjfr:


panel/tests/ui/widgets/test_codeeditor.py
```diff
@@ -81,4 +82,5 @@ def test_code_editor_not_on_keyup(page):
         page.keyboard.down("Meta")
```
hoxbro, line 82: This if statement is wrong.

Issue #7119, opened 2024-08-10
pixi run docs-build missing Webdriver


cdeil:
I followed https://holoviz-dev.github.io/panel/developer_guide/index.html#documentation to run
```
panel $ pixi run docs-build 
```
which ran
```
✨ Pixi task (_docs-generate in docs): nbsite build --what=html --output=builtdocs --org holoviz --project-name panel
```
which gave this RuntimeError:
```
getting thumbnail code for /Users/cdeil/code/oss/panel/examples/reference/widgets/FileDropper.ipynb
Path exists True
Traceback (most recent call last):
  File "/var/folders/6v/0_6nt0pj07x9xjhd8qzkyy700000gn/T/tmp6jv8j0iz", line 67, in <module>
    from nbsite.gallery.thumbnailer import thumbnail;thumbnail(file_dropper, '/Users/cdeil/code/oss/panel/doc/reference/widgets/thumbnails/FileDropper')
                                                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/cdeil/code/oss/panel/.pixi/envs/docs/lib/python3.11/site-packages/nbsite/gallery/thumbnailer.py", line 133, in thumbnail
    obj.save(basename+'.png')
  File "/Users/cdeil/code/oss/panel/panel/viewable.py", line 964, in save
    return save(
           ^^^^^
  File "/Users/cdeil/code/oss/panel/panel/io/save.py", line 270, in save
    return save_png(
           ^^^^^^^^^
  File "/Users/cdeil/code/oss/panel/panel/io/save.py", line 85, in save_png
    state.webdriver = webdriver_control.create()
                      ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/cdeil/code/oss/panel/.pixi/envs/docs/lib/python3.11/site-packages/bokeh/io/webdriver.py", line 180, in create
    driver = self._create(kind, scale_factor=scale_factor)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/cdeil/code/oss/panel/.pixi/envs/docs/lib/python3.11/site-packages/bokeh/io/webdriver.py", line 198, in _create
[...]

PR #7182, opened 2024-08-24
Fix bad markdown file

hoxbro:
Fixes the problem seen here: https://github.com/holoviz/panel/issues/7119#issuecomment-2307673850

I will run the docs build to verify. https://github.com/holoviz/panel/actions/runs/10538265033

### DIFF
--- a/doc/how_to/custom_components/esm/build.md
+++ b/doc/how_to/custom_components/esm/build.md
@@ -100,7 +100,8 @@ To recap here are the three files that we need:
 
 ::::{tab-set}
 
-:::tab-item package.json
+:::{tab-item} package.json
+
 ```json
 {
   "name": "confetti-button",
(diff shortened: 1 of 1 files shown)


doc/how_to/custom_components/esm/build.md
```diff
@@ -100,7 +100,8 @@ To recap here are the three files that we need:
 
 ::::{tab-set}
 
-:::tab-item package.json
+:::{tab-item} package.json
```
hoxbro, line 103: This is the main problem

hoxbro:
We should have a way to test markdown files so an error like this does not happen again.

philippjfr (approved):

philippjfr:
Good catch, thanks!

philippjfr:
>We should have a way to test markdown files so an error like this does not happen again.

Agreed, guess we could just run MyST parser on all markdown files.

philippjfr:
Will merge this for now.

PR #7184, opened 2024-08-26
Ensure proper cleanup of watchfiles threads on server shutdown

philippjfr:
Makes sure that the `watchfiles` watcher threads are cleaned up by calling cleanup on all models before shutdown.


panel/io/server.py
```diff
@@ -104,6 +104,8 @@
 INDEX_HTML = os.path.join(os.path.dirname(__file__), '..', '_templates', "index.html")
 DEFAULT_TITLE = "Panel Application"
 
+_running_tasks = set()
```
hoxbro, line 107: Is this needed?

philippjfr: No, not anymore.

panel/io/server.py
```diff
@@ -104,6 +104,8 @@
...
+
```
philippjfr, line 108:
```suggestion
```

panel/io/server.py
```diff
@@ -1173,14 +1176,11 @@ def show_callback():
...
 
-    def sig_exit(*args, **kwargs):
-        server.io_loop.add_callback_from_signal(do_stop)
-
     def do_stop(*args, **kwargs):
         server.io_loop.stop()
 
     try:
-        signal.signal(signal.SIGINT, sig_exit)
+        server.io_loop.asyncio_loop.add_signal_handler(signal.SIGINT, do_stop)
```
philippjfr, line 1183:
```suggestion
        signal.signal(signal.SIGINT, sig_exit)
```

PR #7194, opened 2024-08-27
Correctly detect max depth of NestedSelect if level is empty

philippjfr:
Empty levels were not treated correctly, resulting in errors when a leaf was empty and incorrectly treating a two-level `NestedSelect` as a single level

### DIFF
--- a/panel/tests/widgets/test_select.py
+++ b/panel/tests/widgets/test_select.py
@@ -300,6 +300,15 @@ def test_nested_select_init_empty(document, comm):
     assert select.options is None
     assert select.levels == []
 
+def test_nested_select_max_depth_empty_first_sublevel(document, comm):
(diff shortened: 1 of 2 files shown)


panel/tests/widgets/test_select.py
```diff
@@ -500,7 +509,7 @@ def test_nested_select_partial_options_set(document, comm):
     select.options = {"Ben": []}
     assert select._widgets[0].value == 'Ben'
     assert select._widgets[0].visible
-    assert select.value == {0: 'Ben'}
+    assert select.value == {0: 'Ben', 1: None}
```
ahuang11, line 512: I'm wondering if this is correct?

philippjfr: Seems correct to me, there are two levels after all, even if one of them is empty.

PR #7204, opened 2024-08-29
Add panel compile command to bundle ESM components

philippjfr:
This PR makes it possible to compile and bundle dependencies of an ESM component using a simple command. Specifically the user can run:

```
panel compile my_module_containing_esm_components.py
```

which will perform the following steps:

1. Create a temporary module for the provided file(s)
2. Find all `ReactiveESM` components in those file(s)
3. For each `ReactiveESM` component extract the required dependencies from the `importmap` and the code itself
4. Write the JS code and the generated nodeJS `package.json` containing the dependencies to a temporary directory
5. Invoke `npm install` to fetch the dependencies
6. Invoke `esbuild index.js --bundled --format=esm --minify`
7. Write the generated file to the directory where the original file is located

The compiled JS bundle will then be used the next time the component is used, as long as `--autoreload` is disabled.

### DIFF
--- a/panel/custom.py
+++ b/panel/custom.py
@@ -99,6 +99,8 @@ class ReactiveESM(ReactiveCustomBase, metaclass=ReactiveESMMetaclass):
 
     _esm: ClassVar[str | os.PathLike] = ""
 
+    _exports: ClassVar[str] = ""
+
     _importmap: ClassVar[dict[Literal['imports', 'scopes'], str]] = {}
 
     __abstract = True
@@ -148,6 +150,7 @@ class ReactiveESM(ReactiveCustomBase, metaclass=ReactiveESMMetaclass):
         else:
             esm = cls._esm
         esm = textwrap.dedent(esm)
+        esm += textwrap.dedent(cls._exports)
         return esm
 
     def _cleanup(self, root: Model | None) -> None:
@@ -394,6 +397,11 @@ class ReactComponent(ReactiveESM):
     _bokeh_model = _BkReactComponent
 
     _react_version = '18.3.1'
+
+    _exports = """
+    import * as React from "react"
+    import { createRoot } from "react-dom/client"
+    export default {render, React, createRoot}"""
 
     @classmethod
     def _process_importmap(cls):
--- a/panel/custom.py
+++ b/panel/custom.py
@@ -1,9 +1,11 @@
 from __future__ import annotations
 
 import asyncio
+import hashlib
 import inspect
 import os
 import pathlib
+import sys
 import textwrap
 
 from collections import defaultdict
@@ -119,7 +121,7 @@ class ReactiveESM(ReactiveCustomBase, metaclass=ReactiveESMMetaclass):
         self._event_callbacks = defaultdict(list)
 
     @classproperty
-    def _compiled_path(cls) -> os.PathLike | None:
+    def _bundle_path(cls) -> os.PathLike | None:
         if config.autoreload:
             return
         try:
@@ -143,14 +145,24 @@ class ReactiveESM(ReactiveCustomBase, metaclass=ReactiveESMMetaclass):
                     'path or a Path object to a .js file extension.'
                 )
         path = mod_path / f'{cls.__name__}.bundle.js'
-        return path if path.is_file() else None
+        if path.is_file():
+            return path
+        module = cls.__module__
+        path = mod_path / f'{module}.bundle.js'
+        if path.is_file():
+            return path
+        elif module in sys.modules:
+            module = os.path.basename(sys.modules[module].__file__).replace('.py', '')
+            path = mod_path / f'{module}.bundle.js'
+            return path if path.is_file() else None
+        return None
 
     @classmethod
     def _esm_path(cls, compiled: bool = True) -> os.PathLike | None:
         if compiled:
-            compiled_path = cls._compiled_path
-            if compiled_path:
-                return compiled_path
+            bundle_path = cls._bundle_path
+            if bundle_path:
+                return bundle_path
         esm = cls._esm
         if isinstance(esm, pathlib.PurePath):
             return esm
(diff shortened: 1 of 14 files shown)


MarcSkovMadsen:
In the end I believe we should have one external `panel-xyz` package that demonstrates all of this and makes it easy for others to do.

philippjfr:
> In the end I believe we should have one external panel-xyz package that demonstrates all of this and makes it easy for others to do.

Yes, ideally a cookie-cutter with CI setup.

panel/io/compile.py
```diff
@@ -0,0 +1,177 @@
...
+            key = key[:-1]
+
+        if match:
+            dependencies[key] = f"^{match.group(1)}"
+        else:
+            dependencies[key] = "latest"
+
+    esm = component._render_esm(compiled=False)
+    if issubclass(component, ReactComponent):
+        esm += '\nimport * as React from "react"\nimport { createRoot } from "react-dom/client"\nexport default {render, React, createRoot}'
```
hoxbro, line 111:
```suggestion
        esm += '\nimport * as React from "react"\nimport { createRoot } from "react-dom/client"\nexport default {render, React, createRoot}'
```

Maybe use `os.linesep` here?

panel/custom.py
```diff
@@ -108,24 +108,45 @@ def __init__(self, **params):
...
+            compiled_path = cls._compiled_path
+            if compiled_path:
+                return compiled_path
+        esm = cls._esm
         if isinstance(esm, pathlib.PurePath):
             return esm
         try:
-            esm_path = pathlib.Path(inspect.getfile(type(self))).parent / esm
+            if hasattr(cls, '__path__'):
+                mod_path = cls.__path__
```
hoxbro, line 134:
```suggestion
                mod_path = Path(cls.__path__[0])
```

image

panel/custom.py
```diff
@@ -182,17 +204,24 @@ def _init_params(self) -> dict[str, Any]:
...
+            'importmap': importmap,
+            'precompiled': precompiled
         })
         return params
 
-    def _process_importmap(self):
-        return self._importmap
+    @classmethod
+    def _process_importmap(cls):
+        return cls._importmap
```
hoxbro, line 224: This could now share instances between different objects is that ok?

panel/custom.py
```diff
@@ -182,17 +204,24 @@ def _init_params(self) -> dict[str, Any]:
                 params.pop(k)
             data_params[k] = v
         data_props = self._process_param_change(data_params)
+        precompiled = self._compiled_path is not None
+        if precompiled:
+            importmap = {}
+        else:
+            importmap = self._process_importmap()
```
hoxbro, line 211:
```suggestion
        precompiled = self._compiled_path is not None
        importmap = {} if precompiled else self._process_importmap()
```
nit

panel/custom.py
```diff
@@ -219,7 +248,7 @@ def _setup_autoreload(self):
         if not ((config.autoreload or getattr(self, '_debug', False)) and import_available('watchfiles')):
             return
         super()._setup_autoreload()
-        if (self._esm_path and not self._watching_esm):
+        if (self._esm_path(False) and not self._watching_esm):
```
hoxbro, line 251:
```suggestion
        if (self._esm_path(compile=False) and not self._watching_esm):
```

Nit, for readability.

panel/io/compile.py
```diff
@@ -0,0 +1,177 @@
...
+    if not importmap:
+        raise ValueError(
+            f'{component.__name__} did not declare an importmap '
+            'and cannot be compiled.'
+        )
+
+    # Extract dependencies and versions
+    dependencies = {}
+    for key, url in importmap["imports"].items():
+        match = re.search(r'@(\d+\.\d+\.\d+)', url)
```
hoxbro, line 100: Move regex compile outside of the for loop.

panel/io/compile.py
```diff
@@ -0,0 +1,177 @@
...
+def compile_component(component: type[ReactiveESM], minify: bool = True):
+    """
+    Compiles and bundles the ESM code of a ReactiveESM component and
+    writes the compiled JS file to a file adjacent to the module
+    the component was defined in.
+    """
+    name = camel_to_kebab(component.__name__)
+    ext = 'jsx' if issubclass(component, ReactComponent) else 'js'
+    if hasattr(component, '__path__'):
+        out_path = component.__path__
```
hoxbro, line 133:
```suggestion
        out_path = Path(component.__path__[0]) 
```
Same as above.

panel/io/compile.py
```diff
@@ -0,0 +1,177 @@
...
+    else:
+        out_path = pathlib.Path(inspect.getfile(component)).parent
+    out = (pathlib.Path(out_path) / f'{name}.compiled.js').absolute()
```
hoxbro, line 136:
```suggestion
    out = (out_path / f'{name}.compiled.js').absolute()
```

Done in the if statement

panel/io/compile.py
```diff
@@ -0,0 +1,177 @@
...
+    with tempfile.TemporaryDirectory() as temp_dir:
+        package_json_path = os.path.join(temp_dir, 'package.json')
+        with open(package_json_path, 'w') as package_json_file:
+            json.dump(package_json, package_json_file, indent=2)
+
+        index_js_path = os.path.join(temp_dir, f'index.{ext}')
+        with open(index_js_path, 'w') as index_js_file:
+            index_js_file.write(code)
+
+        os.chdir(temp_dir)
```
hoxbro, line 149: No need to change back?

panel/io/compile.py
```diff
@@ -0,0 +1,177 @@
...
+
+        install_cmd = ['npm', 'install']
```
hoxbro, line 151: Do a check here. Maybe you can reuse what is already in `bokeh build`.

panel/io/compile.py
```diff
@@ -0,0 +1,177 @@
...
+        minify_arg = ['--minify'] if minify else []
+        build_cmd = ['esbuild', index_js_path, '--bundle', '--format=esm', f'--outfile={out}'] + minify_arg
+        try:
+            print(f"Running command: {' '.join(build_cmd)}")  # noqa
+            result = subprocess.run(build_cmd, check=True, capture_output=True, text=True)
+            print("esbuild output:", result.stdout)  # noqa
+            if result.stderr:
+                print("esbuild errors:", result.stderr)  # noqa
+        except subprocess.CalledProcessError as e:
+            print("An error occurred while running esbuild:")  # noqa
```
hoxbro, line 171: If you want to spice it up with colors, you can take a look at `hatch_build.py`

panel/custom.py
philippjfr: Yes, `_importmap` is defined as a class variable. Could make a copy here but it's a private API.

panel/custom.py
philippjfr: `__path__` is something I patch onto the classes in the dynamic module at compile time. Probably should have chosen some better name.

panel/io/compile.py
philippjfr: Done

panel/io/compile.py
philippjfr: Bokeh has a far more complex build chain and asks you to run `npm install` yourself.

PR #7205, opened 2024-08-29
Add FastAPI based serve functions

philippjfr:
Builds on top of [bokeh-fastapi](https://github.com/bokeh/bokeh-fastapi/) and adds two main user facing APIs:

- `panel.io.fastapi.serve`: Mirrors the `panel.serve` API but launches a FastAPI backed uvicorn server (optionally on a separate thread).
- `panel.io.fastapi.add_applications`: Adds Panel applications to an existing FastAPI application.
- `panel.io.fastapi.add_application`: Decorator to add a function returning Panel objects to the FastAPI server.

@pmeier, thoughts and review appreciated. I'd like to get an initial cut of this in pretty soon but then we'll have about 1 - 1.5 weeks to add docs and more tests.

### DIFF
--- a/panel/io/application.py
+++ b/panel/io/application.py
@@ -6,9 +6,13 @@ from __future__ import annotations
 import json
 import logging
 import os
+import pathlib
 
 from functools import partial
-from typing import TYPE_CHECKING, Any
+from types import FunctionType, MethodType
+from typing import (
+    TYPE_CHECKING, Any, Callable, Mapping,
+)
 
 import bokeh.command.util
 
@@ -17,19 +21,63 @@ from bokeh.application.handlers.directory import DirectoryHandler
 from bokeh.application.handlers.document_lifecycle import (
     DocumentLifecycleHandler,
 )
+from bokeh.application.handlers.function import FunctionHandler
+from bokeh.models import CustomJS
 
 from ..config import config
 from .document import _destroy_document
 from .handlers import MarkdownHandler, NotebookHandler, ScriptHandler
+from .loading import LOADING_INDICATOR_CSS_CLASS
 from .logging import LOG_SESSION_DESTROYED, LOG_SESSION_LAUNCHING
 from .state import set_curdoc, state
 
 if TYPE_CHECKING:
     from bokeh.application.application import SessionContext
     from bokeh.application.handlers.handler import Handler
+    from bokeh.document.document import Document
 
-log = logging.getLogger('panel.io.application')
+    from ..template.base import BaseTemplate
+    from ..viewable import Viewable, Viewer
+    from .location import Location
 
+    TViewable = Viewable | Viewer | BaseTemplate
+    TViewableFuncOrPath = TViewable | Callable[[], TViewable] | os.PathLike | str
+
+
+logger = logging.getLogger('panel.io.application')
+
+
+def _eval_panel(
+    panel: TViewableFuncOrPath, server_id: str, title: str,
+    location: bool | Location, admin: bool, doc: Document
+):
+    from ..pane import panel as as_panel
+    from ..template import BaseTemplate
+
+    if config.global_loading_spinner:
+        doc.js_on_event(
+            'document_ready', CustomJS(code=f"""
+            const body = document.getElementsByTagName('body')[0]
+            body.classList.remove({LOADING_INDICATOR_CSS_CLASS!r}, {config.loading_spinner!r})
+            """)
+        )
+
+    doc.on_event('document_ready', partial(state._schedule_on_load, doc))
+
+    # Set up instrumentation for logging sessions
+    logger.info(LOG_SESSION_LAUNCHING, id(doc))
+    def _log_session_destroyed(session_context):
+        logger.info(LOG_SESSION_DESTROYED, id(doc))
+    doc.on_session_destroyed(_log_session_destroyed)
+
+    with set_curdoc(doc):
+        if isinstance(panel, (FunctionType, MethodType)):
+            panel = panel()
+        if isinstance(panel, BaseTemplate):
+            doc = panel._modify_doc(server_id, title, doc, location)
+        else:
+            doc = as_panel(panel)._modify_doc(server_id, title, doc, location)
+        return doc
 
 def _on_session_destroyed(session_context: SessionContext) -> None:
     """
@@ -41,8 +89,10 @@ def _on_session_destroyed(session_context: SessionContext) -> None:
         try:
             callback(session_context)
         except Exception as e:
-            log.warning("DocumentLifecycleHandler on_session_destroyed "
-                        f"callback {callback} failed with following error: {e}")
+            logger.warning(
+                "DocumentLifecycleHandler on_session_destroyed "
+                f"callback {callback} failed with following error: {e}"
+            )
 
 
 class Application(BkApplication):
@@ -73,14 +123,14 @@ class Application(BkApplication):
         super().add(handler)
 
     def initialize_document(self, doc):
-        log.info(LOG_SESSION_LAUNCHING, id(doc))
+        logger.info(LOG_SESSION_LAUNCHING, id(doc))
         super().initialize_document(doc)
         if doc in state._templates and doc not in state._templates[doc]._documents:
             template = state._templates[doc]
             with set_curdoc(doc):
                 template.server_doc(title=template.title, location=True, doc=doc)
         def _log_session_destroyed(session_context):
-            log.info(LOG_SESSION_DESTROYED, id(doc))
+            logger.info(LOG_SESSION_DESTROYED, id(doc))
         doc.destroy = partial(_destroy_document, doc) # type: ignore
         doc.on_event('document_ready', partial(state._schedule_on_load, doc))
         doc.on_session_destroyed(_log_session_destroyed)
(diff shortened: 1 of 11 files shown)


panel/io/application.py
```diff
@@ -17,19 +21,63 @@
...
 
+    if config.global_loading_spinner:
+        doc.js_on_event(
+            'document_ready', CustomJS(code=f"""
+            const body = document.getElementsByTagName('body')[0]
+            body.classList.remove({LOADING_INDICATOR_CSS_CLASS!r}, {config.loading_spinner!r})
+            """)
+        )
+
+    doc.on_event('document_ready', partial(state._schedule_on_load, doc))
```
hoxbro, line 65: Can this cause a memory leak. In doc never being cleaned up, or am I being paranoid?

panel/io/application.py
```diff
@@ -144,3 +194,75 @@ def build_single_handler_application(path, argv=None):
...
+
+
+def build_applications(
+    panel: TViewableFuncOrPath | Mapping[str, TViewableFuncOrPath],
+    title: str | dict[str, str] = None,
+    location: bool | Location = True,
+    admin: bool = False,
+    server_id: str | None = None,
+    custom_handlers: list = None
+) -> dict[str, Application]:
```
hoxbro, line 206:
```suggestion
def build_applications(
    panel: TViewableFuncOrPath | Mapping[str, TViewableFuncOrPath],
    title: str | dict[str, str] | None = None,
    location: bool | Location = True,
    admin: bool = False,
    server_id: str | None = None,
    custom_handlers: list | None = None
) -> dict[str, Application]:
```

panel/io/application.py
```diff
@@ -144,3 +194,75 @@ def build_single_handler_application(path, argv=None):
...
+            if app is not None:
+                break
+        else:
+            new_app = app
+        if new_app is not None:
+            app = new_app
+            if app is True:
+                continue
+
+        if isinstance(app, pathlib.Path):
```
hoxbro, line 257:
```suggestion
        if isinstance(app, os.PathLike):
```

panel/io/fastapi.py
```diff
@@ -0,0 +1,229 @@
...
+    return [conn._socket.send_message(msg)]
+
+extra_socket_handlers[WSHandler] = dispatch_fastapi
+
+
+def add_applications(
+    panel: TViewableFuncOrPath | Mapping[str, TViewableFuncOrPath],
+    server: FastAPI | None = None,
+    title: str | dict[str, str] = None,
+    location: bool | Location = True,
```
hoxbro, line 39:
```suggestion
    title: str | dict[str, str] | None = None,
    location: bool | Location | None = True,
```

panel/io/fastapi.py
```diff
@@ -0,0 +1,229 @@
...
+        return FileResponse(resolved_path)
+
+    return application
+
+
+def get_server(
+    panel: TViewableFuncOrPath | Mapping[str, TViewableFuncOrPath],
+    port: int | None = 0,
+    start: bool = False,
+    title: str | dict[str, str] = None,
```
hoxbro, line 85:
```suggestion
    title: str | dict[str, str] | None = None,
```

panel/io/server.py
```diff
@@ -816,7 +776,6 @@ def serve(
     """
     # Empty layout are valid and the Bokeh warning is silenced as usually
     # not relevant to Panel users.
```
hoxbro, line 778:
```suggestion
    """
```

panel/io/server.py
```diff
@@ -888,7 +847,7 @@ def get_server(
     loop: Optional[IOLoop] = None,
     show: bool = False,
     start: bool = False,
-    title: bool = None,
+    title: str | dict[str, str] = None,
```
hoxbro, line 850:
```suggestion
    title: str | dict[str, str] | None = None,
```

panel/io/server.py
```diff
@@ -1015,57 +974,29 @@ def get_server(
...
-            panel = {'/': app}
-        elif (isinstance(panel, str) and panel.endswith(('.py', '.ipynb', '.md'))
-            and os.path.isfile(panel)):
-            apps = {'/': build_single_handler_application(panel)}
-        else:
-            handler = FunctionHandler(partial(_eval_panel, panel, server_id, title, location, admin))
-            apps = {'/': Application(handler, admin=admin)}
+    kwargs['extra_patterns'] = extra_patterns = list(kwargs.get('extra_patterns', []))
+
+    def flask_handler(slug, app):
```
hoxbro, line 981: Maybe expose this outside the method so people can hack and change it.

panel/io/application.py
```diff
@@ -144,3 +194,75 @@ def build_single_handler_application(path, argv=None):
...
+    title : str or {str: str} (optional, default=None)
+        An HTML title for the application or a dictionary mapping
+        from the URL slug to a customized title.
+    location : boolean or panel.io.location.Location
+        Whether to create a Location component to observe and
+        set the URL location.
+    admin: boolean (default=False)
+        Whether to enable the admin panel
+    server_id: str
+        ID of the server running the application(s)
```
ahuang11, line 224: Missing `custom_handlers` doc

panel/io/application.py
```diff
@@ -144,3 +194,75 @@ def build_single_handler_application(path, argv=None):
...
+    """
+    if not isinstance(panel, dict):
+        panel = {'/': panel}
+
+    apps = {}
+    for slug, app in panel.items():
+        if slug.endswith('/') and not slug == '/':
```
ahuang11, line 231:
```suggestion
        if slug.endswith('/') and slug != '/':
```

panel/io/fastapi.py
ahuang11: `location` shouldn't be None I think

hoxbro: Thanks

panel/io/application.py
philippjfr: The document gets cleaned up when the session is destroyed, including all handlers.

panel/io/fastapi.py
```diff
@@ -0,0 +1,229 @@
...
+    location : boolean or panel.io.location.Location
+        Whether to create a Location component to observe and
+        set the URL location.
+    admin: boolean (default=False)
+        Whether to enable the admin panel
+    **kwargs:
+        Additional keyword arguments to pass to the BokehFastAPI application
+    """
+    apps = build_applications(panel)
+    application = BokehFastAPI(apps, server=server, **kwargs)
```
pmeier, line 65:
```suggestion
    application = BokehFastAPI(apps, app=app, **kwargs)
```

MarcSkovMadsen:
If this make it very easy to add FAST API (REST) endpoints in general it would be of great value.

panel/io/fastapi.py
```diff
@@ -0,0 +1,229 @@
+from __future__ import annotations
```
MarcSkovMadsen, line 1: Have you considered whether there would be a more FastAPI like way of integrating Panel?

Something that would be very understandable to users of FastAPI? Where Panel is more seen as one component of many in a larger FASTAPI server?

MarcSkovMadsen: Something that would make sense in the head of Sebastian Ramirez and align with how other components integrate?

MarcSkovMadsen: Communication wise its much more valuable to say Panel now provides a FastAPI plugin or component than trying to explain Panel runs on FastAPI but in its own unique way.

philippjfr: I am honestly not quite following what your suggestion is. You can add Panel apps to an existing FastAPI app, so not sure what else it is you're asking for.

MarcSkovMadsen: Ok. I guess to need the documentation for how this is used.

philippjfr: Hope the docs here clarified things.

panel/io/server.py
philippjfr: It's a closure and I think it should be deprecated eventually.

panel/io/fastapi.py
```diff
@@ -0,0 +1,319 @@
...
+
+import asyncio
+import uuid
+
+from functools import wraps
+from typing import TYPE_CHECKING, Mapping, cast
+
+from bokeh_fastapi import BokehFastAPI
+from bokeh_fastapi.handler import WSHandler
```
pmeier, line 10: Do you want to add a proper error message here if the import fails given that this is not a required dependency?

panel/tests/conftest.py
```diff
@@ -563,3 +566,16 @@ def df_strings():
     code = [f'{i:02d}' for i in range(len(descr))]
 
     return pd.DataFrame(dict(code=code, descr=descr))
+
+@pytest.fixture
+def server_implementation(request):
```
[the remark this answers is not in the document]
philippjfr: I was looking to eventually run all server based tests on both servers. This was moving in that direction.

panel/io/fastapi.py
hoxbro: I have added an error message, and I'm open to suggestions. 

Chose `bokeh_fastapi` with an underscore because this is what other bokeh conda projects have [[ref](https://anaconda.org/bokeh/)], and conda will not install `bokeh-fastapi` out of the box.  Pip is smarter and will install no matter if you install it as `bokeh_fastapi` or `bokeh-fastapi`.

panel/tests/test_server.py
```diff
@@ -23,12 +24,28 @@
...
+@pytest.fixture(params=["tornado", "fastapi"])
+def server_implementation(request):
+    try:
+        importlib.import_module(request.param)
+    except Exception:
+        pytest.skip(f'{request.param!r} is not installed')
+    old = serve_and_wait.server_implementation
+    serve_and_wait.server_implementation = request.param
+    try:
+        yield
```
hoxbro, line 44:
```suggestion
        yield request.param
```

Mostly for convenience.

philippjfr:
This pull request has been mentioned on **HoloViz Discourse**. There might be relevant details there:

https://discourse.holoviz.org/t/how-to-integrate-a-panel-app-with-fastapi-on-the-same-port/7601/6

Issue #6083, opened 2023-12-18
Make Column scroll position usefull


hyamanieu:
#### Is your feature request related to a problem? Please describe.
I find the `scroll_position` parameter to be only useful for setting it to 0, but not for jumping to a specific area. I naively thought the integer represented the index of the listed objects, but it corresponds to the [`scrollTop`](https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollTop) property.

I want a method to programmatically jump to an element of the list.

In general, I believe it can be usefull for other subclass of `ListPanel`, like `ChatFeed`, as long as a scroll bar may appear.

#### Describe the solution you'd like
I can see two very different approaches with pros and cons:

1. Implement a `focus` method on python side which does the same as [`HTMLElement.focus({ preventScroll: false })`](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/focus#preventscroll). Then it's not about modifying `ListPanel` but about adding this method to all Viewable objects.

  - pros: it responds directly to the problem
  - cons: 
    - it may be more complex to implement.
    - it changes the focus (of course), which may be an unwanted side-effect unrelated to the original issue (not my  case)

2. Modify column.ts so that ScrollHeight is a model property as well, helping with calculating a percentage of scrolling. (pros & cons are swapped)

#### Describe alternatives you've considered

1. Regarding my other issues #2834, closed PR #2842, and now #6073, I was able to find a specific object by setting a name or tags, then calling the[ `.focus()` ](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/focus#preventscroll) method. However it doesn't work anymore in Panel 1.
2. I can still find the column object by calling its specific custom class, then getting its `scrollHeight`. Then just estimating where I should be at. It's not accurate.
3. like 2, but calling for `.shadowRoot.children[X].focus()`, however I have no idea how to get child `X`.
4. like 2, but setting as many custom CSS class as there are children 🙄 

#### Additional context

I'm willing to work on a PR if given a direction either among my two approaches or a third one I have missed.

ahuang11:
Thanks for sharing this!

I think `scroll_position` is already a public parameter so if we change it, it'd be a breaking change, but we should definitely update the docs to mention it corresponds to the [scrollTop](https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollTop) property.

So, to add to your proposal, I imagine:
1. create a new parameter called scroll_index which scrolls to the index of the listed objects
[...]

PR #7206, opened 2024-08-29
Add scroll_index to Column

ahuang11:
Closes https://github.com/holoviz/panel/issues/6083

```python
import panel as pn

pn.extension()

def scroll_to_top(event):
    column.scroll_index = 100


column = pn.Column(
    *list(range(1000)),
    height=300,
    sizing_mode="fixed",
    scroll=True,
)
btn = pn.widgets.Button(name="Scroll to top", width=200, button_type="primary", on_click=scroll_to_top)
pn.Row(column, btn).servable()
```

(poorly named btn; should be scroll to index 100)

https://github.com/user-attachments/assets/0d03557d-0f27-485f-ae93-feead2a50e50


panel/layout/base.py
```diff
@@ -939,6 +939,12 @@ class Column(ListPanel):
         will update the scroll position of the Column. Setting to
         0 will scroll to the top.""")
 
+    scroll_index = param.Integer(default=None, bounds=(0, None), doc="""
```
philippjfr, line 942: I think this should be implemented as a `scroll_to_index` method which sends an event to the frontend (e.g. like the HTML diff approach).

philippjfr (approved):

PR #7207, opened 2024-08-30
Speed up import time

hoxbro:


philippjfr (approved):

panel/param.py
```diff
@@ -214,35 +213,41 @@ class Param(Pane):
...
-        param.ObjectSelector:    Select,
-        param.Parameter:         LiteralInputTyped,
-        param.Range:             RangeSlider,
-        param.Selector:          Select,
-        param.String:            TextInput,
-    }
-
-    if hasattr(param, 'Event'):
-        mapping[param.Event] = Button
+    @classproperty
```
philippjfr, line 216: Users could override the mappings in here previously.

hoxbro: Should be possible again with: [8286744](https://github.com/holoviz/panel/pull/7207/commits/82867442a672c42352805b38cb0c667ad64066dd)

(The dict merge was added in Python 3.9)

panel/param.py
```diff
@@ -53,37 +53,34 @@ class Skip(RuntimeError):
...
 from .reactive import Reactive
 from .util import (
-    abbreviated_repr, flatten, full_groupby, fullpath, is_parameterized,
-    param_name, recursive_parameterized, to_async_gen,
+    abbreviated_repr, classproperty, flatten, full_groupby, fullpath,
+    is_parameterized, param_name, recursive_parameterized, to_async_gen,
 )
 from .util.checks import is_dataframe, is_mpl_axes, is_series
 from .viewable import Layoutable, Viewable
-from .widgets import (
```
philippjfr, line 61: I am fairly confused by inlining this stuff. The `panel.widgets` module is imported by default.

philippjfr (changes requested):
Happy with most of this but I don't quite understand the inlining of `panel.widgets` and `panel.util` imports. These end up getting imported anyway. Therefore I'd suggest we postpone those changes until those imports are actually lazy (as I don't want to do that in this release).

panel/param.py
hoxbro: This is mostly preparation for the lazy import PR to avoid circular imports.

I can move these changes into that PR or open a new PR with only the necessary changes, whatever is easiest.

philippjfr: Would prefer that as I want to do that stuff after this release and carefully consider the approach.

PR #7208, opened 2024-08-30
 Make panel and panel.io lazy imports 

hoxbro:
Honestly, this should not be added to the 1.5 release, as it is too late.

But would not want to delete the changes either.

### DIFF
--- a/panel/__init__.py
+++ b/panel/__init__.py
@@ -45,34 +45,88 @@ https://blog.holoviz.org/panel_0.12.0.html#JupyterLab-previews
 To learn more about Panel check out
 https://panel.holoviz.org/getting_started/index.html
 """
-from param import rx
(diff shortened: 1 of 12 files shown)


panel/__init__.py
```diff
@@ -45,34 +45,87 @@
...
+    "config": "panel.config:config",
+    "custom": "panel.custom",
+    "indicators": "panel.widgets:indicators",
+    "interact": "panel._interact:interact",
+    "ipywidget": "panel.io.notebook:ipywidget",
+    "layout": "panel.layout",
+    "links": "panel.links",
+    "pane": "panel.pane",
+    "panel": "panel.pane:panel",
+    "param": "panel.param",
```
hoxbro, line 108:
```suggestion
```

PR #7209, opened 2024-08-30
Tweak ChatMessage layout

ahuang11:
Ensures the icon divider is invisible if reaction_icons is empty or if copy icon is invisible.

```python
import panel as pn

pn.extension()

cm = pn.chat.ChatMessage("ABC")
cm.servable()
```
image

`cm = pn.chat.ChatMessage("ABC", show_copy_icon=False)`

image

`cm = pn.chat.ChatMessage("ABC", reaction_icons={})`

image

### DIFF
--- a/panel/chat/icon.py
+++ b/panel/chat/icon.py
@@ -72,7 +72,12 @@ class ChatReactionIcons(CompositeWidget):
             icon._reaction = option
             icon.param.watch(self._update_value, "value")
             self._rendered_icons[option] = icon
-        self._composite[:] = [self.default_layout(*list(self._rendered_icons.values()))]
+        self._composite[:] = [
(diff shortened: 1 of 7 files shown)


MarcSkovMadsen:
The green activity indicator is now in the right spot. I still see the divider icon though.

image

MarcSkovMadsen (dismissed):
As described its almost perfect. The vertical bar is still there though.

philippjfr:
The line is not reproducible on my end, I think you forgot to pull.

philippjfr:
Also we will have to revisit this, e.g. serializing a whole model simply to render the separator icon is a little bit insane.

philippjfr (approved):

PR #7224, opened 2024-09-02
Make --dev an alias for --autoreload

philippjfr:
The `--dev` argument to the `panel serve` command sets up the Tornado dev tooling, e.g. to restart the server when a file changes. This does not actually work very well, while our internal `--autoreload` handling works quite well. Since `--autoreload` is kind of long and we may broaden the scope of what it does I'd like to make `--dev` an alias for `--autoreload`.


hoxbro (approved):

MarcSkovMadsen:
Like it

PR #7246, opened 2024-09-10
Various documentation fixes before 1.5.0 RC2

philippjfr:


hoxbro:
https://holoviz-dev.github.io/panel/explanation/comms.html

doc/conf.py
```diff
@@ -82,15 +84,27 @@
     "announcement": announcement_text,
 }
 
-extensions += [
-    'sphinx.ext.napoleon',
-    'nbsite.gallery',
+
+extensions = [
```
hoxbro, line 88: Wouldn't it be easy to remove the one you don't like?

pixi.toml
```diff
@@ -177,12 +177,15 @@ test-type = 'mypy panel'
 [feature.doc.activation.env]
 MOZ_HEADLESS = "1"
 PANEL_IPYWIDGET = "1"
+OBJC_DISABLE_INITIALIZE_FORK_SAFETY = "YES"
```
hoxbro, line 180: Still needed?

philippjfr: Yes

doc/conf.py
philippjfr: If I was just removing one of them yes, but with two I find this nicer/more explicit.

PR #7252, opened 2024-09-11
add fastapi extra dependencies

pmeier:
With the new functionality added to serve panel through FastAPI, this adds an extra dependency so you can `pip install panel[fastapi]` and get everything you need.

### DIFF
--- a/doc/how_to/integrations/FastAPI.md
+++ b/doc/how_to/integrations/FastAPI.md
@@ -19,8 +19,14 @@ conda install fastapi
 :::
 
 :::{tab-item} `pip`
-```pip
-conda install fastapi
+```bash
+pip install fastapi
+```
+:::
+
+:::{tab-item} `panel`
+```bash
+pip install panel[fastapi]
 ```
 :::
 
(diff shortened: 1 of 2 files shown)


doc/how_to/integrations/FastAPI.md
```diff
@@ -19,8 +19,14 @@ conda install fastapi
 :::
 
 :::{tab-item} `pip`
-```pip
-conda install fastapi
+```bash
+pip install fastapi
```
pmeier, line 23: @philippjfr This was probably a copy-paste bug before.

pyproject.toml
```diff
@@ -75,6 +75,10 @@ recommended = [
     'pillow',
     'plotly',
 ]
+fastapi = [
+    'bokeh-fastapi == 0.1.0',
```
pmeier, line 79: This will only resolve after the initial release, but that should happen before the next panel release anyway. I opted for a hard-pin here since there are no BC guarantees yet for `bokeh-fastapi`.

doc/how_to/integrations/FastAPI.md
philippjfr: Good catch thanks!

philippjfr (approved):

PR #7255, opened 2024-09-11
Tighten up ChatMessage layout

philippjfr:
Was just looking at the new `ChatMessage` and while it's an improvement the alignment of the icons and the spacing bothered me a little bit. So this PR tightens things up a bit:

## Before

Screenshot 2024-09-11 at 14 50 38

## After

Screenshot 2024-09-11 at 14 45 28

### DIFF
--- a/panel/chat/icon.py
+++ b/panel/chat/icon.py
@@ -6,6 +6,7 @@ from typing import ClassVar
 
 import param
 
+from .._param import Margin
 from ..io.resources import CDN_DIST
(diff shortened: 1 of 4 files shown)


philippjfr:
Seems a tiny bit unbalanced still, the divider should maybe have an extra pixel on the right.

ahuang11:
Seems very tight on my end
image

philippjfr:
Indeed, that looks to cramped.

Issue #6156, opened 2024-01-04
Shadow DOM prevents anchor links to other elements from working


maximlt:
Anchor links to elements withing the served page don't work. I'm demonstrating this with this Markdown example but have also been able to reproduce it with HTML code served in an `HTML` pane. Serve the app and go to `http://localhost:5006/app#tag3`, nothing happens while the page should scroll down to the `tag3` heading.

I believe this is due to the content being displayed in the shadow DOM. Here's someone explaining how they circumvented that with some custom JS: https://advancedfrontends.com/how-to-use-named-anchors-with-shadow-dom/.

```python
# app.py
import panel as pn

md = ''
for tag in ['tag1', 'tag2', 'tag3']:
    md += f'# {tag}\n\n'
    md += f'{tag} content\n' * 50

pn.pane.Markdown(md).servable()
```

- Found this bug in https://discourse.holoviz.org/t/anchor-link-not-working/6651
- xref: https://github.com/holoviz/panel/issues/1567, https://github.com/holoviz/panel/issues/4149

TianchengTang:
Hello, I ran into the same issue while developing web app using Panel. This is a critical feature to enable long dashboard, it allows user to quickly navigate through the page with menu on side panel. Any update on this bug fix? Thanks.

hmijail:
Bump. Would be great to be able to use the internal links that are already generated by the Markdown panes! 

The solution in the advancedfrontends.com link looks nice, but also looks like Panel doesn't allow giving names/ids to the panes, so IIUC it's not directly applicable. I'm trying to pare down the solution in the discourse link, but that's much more heavy-weight / overkill.

hmijail:
By the way, until this is fixed I’d suggest to mention in the docs that internal links are not supposed to work! 

In my case, anchor tags /internal links looked basic enough that I thought I could use them with my limited HTML knowledge, only to spend hours trying to debug, from browser and HTML versions to add-ons to permissions for local html files.

bryanwhiting:
bumping this, as i spent three hours yesterday building/testing/debugigng my links and didn't know this is a bug...

bryanwhiting:
@hmijail i found a solution. Perhaps this helps.

I'm sure this is a variation on the blog @maximlt posted, but what I found I got working was I had to search for all divs that had a shadowroot, and then within each shadowroot search for the id I was looking for.

[...]

PR #7258, opened 2024-09-12
Support anchor links in HTML model

philippjfr:
Searches rendered HTML contents for anchor links and then does two things:

- If document is still initializing will scroll the anchor linked in the URL into view
- Attach click event listeners on all anchor links to ensure that they work across the shadow DOM 

Fixes https://github.com/holoviz/panel/issues/6156

### DIFF
--- a/panel/models/html.ts
+++ b/panel/models/html.ts
@@ -6,6 +6,22 @@ import {Markup} from "@bokehjs/models/widgets/markup"
 import {PanelMarkupView} from "./layout"
 import {serializeEvent} from "./event-to-object"
 
+function searchAllDOMs(node: Element | ShadowRoot, selector: string): (Element | ShadowRoot)[] {
+  let found: (Element | ShadowRoot)[] = []
+
+  if (node instanceof Element && node.matches(selector)) {
+    found.push(node)
+  }
+  node.children && Array.from(node.children).forEach(child => {
+    found = found.concat(searchAllDOMs(child, selector))
+  })
+
+  if (node instanceof Element && node.shadowRoot) {
+    found = found.concat(searchAllDOMs(node.shadowRoot, selector));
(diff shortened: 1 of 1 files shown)


panel/models/html.ts
```diff
@@ -127,12 +141,27 @@ export class HTMLView extends PanelMarkupView {
...
+      const link = anchor.getAttribute("href")
+      if (link && link.startsWith("#")) {
+        anchor.addEventListener("click", () => {
+          const found = searchAllDOMs(document.body, link)
+          if ((found.length > 0) && found[0] instanceof Element) {
+            found[0].scrollIntoView()
+          }
+        })
+        if (!this.root.has_finished() && this.model.document && window.location.hash === link) {
+          this.model.document.on_event("document_ready", () => anchor.scrollIntoView())
```
hoxbro, line 162:
```suggestion
          this.model.document.on_event("document_ready", () => {
            anchor.scrollIntoView()
            setTimeout(() => anchor.scrollIntoView(), 5)
          })
```

I need to wait for the document to load; without this change, the original issue looks like this. (I'm running it twice so we are closer to the anchor for the second run.)

image

hoxbro: May be have a test for this too.

hoxbro (approved):
One small comment

MarcSkovMadsen:
Will this also work when updating page dynamically?

philippjfr:
> Will this also work when updating page dynamically?

Yes

panel/models/html.ts
philippjfr: Will try with `requestAnimationFrame` instead.

philippjfr: Wasn't enough, switched to the timeout.

PR #7266, opened 2024-09-12
Remove divider fix layout

ahuang11:
Supersedes https://github.com/holoviz/panel/pull/7255

Decided that there was way too much complexity to have a divider and it made the alignment all feel off so I removed that, and also fixed the alignment after cache refresh.

image

image


philippjfr (approved):

philippjfr:
Fully in favor, looks good!

Issue #7264, opened 2024-09-12
Bokeh: BokehJS was loaded multiple times but one version failed to initialize.


tomascsantos:
Hi team, thanks for your hard work. If possible, can we put a high priority on this fix? It's quite damaging to user experience.

#### ALL software version info
(this library, plus any other relevant software, e.g. bokeh, python, notebook, OS, browser, etc should be added within the dropdown below.)

<details>
    <summary>Software Version Info</summary>

```plaintext
acryl-datahub==0.10.5.5
aiohappyeyeballs==2.4.0
aiohttp==3.10.5
aiosignal==1.3.1
alembic==1.13.2
ansi2html==1.9.2
anyio==4.4.0
argon2-cffi==23.1.0
argon2-cffi-bindings==21.2.0
arrow==1.3.0
asttokens==2.4.1
async-generator==1.10
async-lru==2.0.4
attrs==24.2.0
autograd==1.7.0
autograd-gamma==0.5.0
avro==1.10.2
avro-gen3==0.7.10
awscli==1.33.27
babel==2.16.0
backports.tarfile==1.2.0
beautifulsoup4==4.12.3
black==24.8.0
bleach==6.1.0
blinker==1.8.2
bokeh==3.4.2
bokehtools==0.46.2
boto3==1.34.76
botocore==1.34.145
bouncer-client==0.4.1
cached-property==1.5.2
certifi==2024.7.4
certipy==0.1.3
cffi==1.17.0
charset-normalizer==3.3.2
click==8.1.7
click-default-group==1.2.4
click-spinner==0.1.10
cloudpickle==3.0.0
colorama==0.4.6
colorcet==3.0.1
comm==0.2.2
contourpy==1.3.0
cryptography==43.0.0
cycler==0.12.1
dash==2.17.1
dash-core-components==2.0.0
dash-html-components==2.0.0
dash-table==5.0.0
dask==2024.8.1
datashader==0.16.3
datatank-client==2.1.10.post12049
dataworks-common==2.1.10.post12049
debugpy==1.8.5
decorator==5.1.1
defusedxml==0.7.1
Deprecated==1.2.14
directives-client==0.4.4
docker==7.1.0
docutils==0.16
entrypoints==0.4
executing==2.0.1
expandvars==0.12.0
fastjsonschema==2.20.0
Flask==3.0.3
fonttools==4.53.1
formulaic==1.0.2
fqdn==1.5.1
frozenlist==1.4.1
fsspec==2024.6.1
[...]

PR #7269, opened 2024-09-13
Fix detection of old Bokeh in notebook load script

philippjfr:
The notebook load script was designed to detect old versions of Bokeh (and Panel) and potentially run multiple copies of the libraries side-by-side. It did this by comparing the global `Bokeh` object version, if already defined, to the new version being loaded. However it seems like scoping rules meant that in some cases it was comparing a local `Bokeh` variable rather than the global `window.Bokeh` object, which meant that it would incorrectly detect that Bokeh should be replaced, but then later correctly detect that it doesn't actually have to reload Bokeh. This meant that it would unload Bokeh and then not **re**load it.

Additionally I've found that due to escaping rules meant that special characters (such as `^`) in URLs would be escaped so that some of the checks that tried to ensure libraries weren't loaded twice were not detecting this correctly because they compared escaped with non-escaped URIs. We now escape them using the `escapeURI` function before comparing.

Fixes https://github.com/holoviz/panel/issues/7264

### DIFF
--- a/panel/_templates/autoload_panel_js.js
+++ b/panel/_templates/autoload_panel_js.js
@@ -61,9 +61,6 @@ calls it with the rendered model.
       run_callbacks();
       return null;
     }
-    if (!reloading) {
-      console.debug("Bokeh: BokehJS not loaded, scheduling load and callback at", now());
-    }
 
     function on_load() {
       root._bokeh_is_loading--;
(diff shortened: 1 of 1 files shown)


hoxbro:
I'm _very_ happy with this PR.

hoxbro (approved):

Issue #7087, opened 2024-08-06
Many relative links are broken in Panel documentation site


dennisjlee:
#### ALL software version info
Mac OSX 14.5
Chrome 127

#### Description of expected behavior and the observed behavior

Many relative links inside the Panel documentation are broken - they are being interpreted with a `#` in front of the relative path.

E.g. on [this page](https://panel.holoviz.org/reference/layouts/Accordion.html), there is a sentence near the top:

```
For further layout and styling-related parameters see the [customization user guide](https://panel.holoviz.org/reference/layouts/Accordion.html#../../user_guide/Customization.html).
```

I'm not sure where the `#` is coming from - in the source code for this documentation, the links do not have `#`: https://github.com/holoviz/panel/blob/ad94551cfa5fc15c637d459748aeb8da340974e2/examples/reference/layouts/Accordion.ipynb?short_path=d69f44e#L31

#### Complete, minimal, self-contained example code that reproduces the issue

On the documentation site panel.holoviz.org

#### Screenshots or screencasts of the bug in action

- [ ] I may be interested in making a pull request to address this

MarcSkovMadsen:
I can confirm its also broken in current dev docs https://holoviz-dev.github.io/panel/reference/layouts/Accordion.html

Coderambling:
# 7086

Coderambling:
#6418

MarcSkovMadsen:
All broken links can probably be found in the docs build logs.

This is an example https://github.com/holoviz/panel/actions/runs/10329645761/job/28597689026

![image](https://github.com/user-attachments/assets/418ca256-a8a6-4b4d-a1a2-71ab9de0ae24)

dennisjlee:
@MarcSkovMadsen thanks for fixing!

PR #7279, opened 2024-09-15
Fix broken docs links

MarcSkovMadsen:
Fixes the broken links that can be fixed manually.

Its out of scope for this PR to solve the general issue https://github.com/holoviz/panel/issues/7280 with broken links between reference notebooks.

Its also out of scope to clean up false positives like complaints about internal links `#...` not found.

Closes https://github.com/holoviz/panel/issues/7087


doc/conf.py
```diff
@@ -95,6 +95,7 @@
     'sphinx.ext.mathjax',
     'sphinx.ext.ifconfig',
     'sphinx.ext.linkcode',
+    'sphinx.ext.inheritance_diagram',
```
MarcSkovMadsen, line 98: An attempt to fix 358 errors like `/home/jovyan/repos/private/panel/doc/api/panel.__main__.rst:4: ERROR: Unknown directive type "inheritance-diagram".`

MarcSkovMadsen:
I've done a first iteration. Now I need to rebuild docs to see if problems have been fixed without introducing other issues.

doc/conf.py
```diff
@@ -86,6 +86,7 @@
 
 
 extensions = [
+    'bokeh.sphinxext.bokeh_plot',
```
MarcSkovMadsen, line 89: This is an attempt to fix 4 errors like `docstring of panel.models.widgets.CheckboxButtonGroup.button_type:4: ERROR: Unknown directive type "bokeh-plot".`.

doc/about/releases.md
MarcSkovMadsen, line 1: The webllm.md file should be added to some toctree similarly to rest of gallery examples. But I could not figure out where. Can you fix @philippjfr ?

/home/jovyan/repos/private/panel/doc/gallery/webllm.md: WARNING: document isn't included in any toctree

MarcSkovMadsen: Maybe its related to /home/jovyan/repos/private/panel/doc/gallery/index.md:186: WARNING: 'myst' cross-reference target not found: 'WebLLM' [myst.xref_missing]

MarcSkovMadsen:
I believe this fixes all or most broken links that should be fixed manually.

doc/conf.py
philippjfr: Hmm, that should have been disabled, I actually don't like the inheritance diagrams, but I'll re-enable for now.

philippjfr (approved):

PR #7285, opened 2024-09-16
Collection of minor fixes for Panel 1.5.1

MarcSkovMadsen:
- [x] fix fastapi example
- [x] replace iframe embed link in pycafe guide.
- [x] replace `--autoreload` with `--dev` in docs.
- [x] Add link to AnyWidget Front End Module (AFM) specification.


panel/command/serve.py
```diff
@@ -238,7 +238,7 @@ class Serve(_BkServe):
         )),
         ('--autoreload', dict(
             action  = 'store_true',
-            help    = "Whether to autoreload source when script changes."
+            help    = "Whether to autoreload source when script changes. We recommend using --dev instead."
```
hoxbro, line 241: This is correct for now, but there could be a future where `--dev` does more than `--autoreload`

MarcSkovMadsen: Is it just for info? Or should I remove the recommendation?

MarcSkovMadsen:
This PR is growing. So I would like it reviewed and merged while its still possible to review.

philippjfr (approved):

philippjfr:
Happy with this, thanks! I think recommending `--dev` over `--autoreload` is fine even if we expand the definition of `--dev` later.

PR #6913, opened 2024-06-11
Make `--setup`/`--autoreload`/`--warm` work with `--num-procs`

hoxbro:
Resolves #6891 
Resolves #6816

The following `generate_session(app, initialize=True)` gave file errors on Python versions below 3.12, so I have set initialize to False for those versions. I'm not entirely sure why it works fine on Python 3.12, but it could be related to sub-interpreters. 

Postpone the setup script execution after the server starts. This means the setup script runs for each process; for example, it will run twice for the following example. 

``` python 
# app.py
import panel as pn

pn.pane.Markdown("Hello World").servable()
```

``` python
# background_job.py
from datetime import datetime
import os
import panel as pn

current_time = datetime.now()


def task() -> None:
    """Refresh the data in the database."""
    print(f"Ran task at {datetime.now()} (PID {os.getpid()})")


pn.state.schedule_task(name="load_data", callback=task, at=current_time, period="5s")
```

Command: `panel serve app.py --setup background_job.py --num-procs 2`

https://github.com/holoviz/panel/assets/19758978/035e6f2d-269d-49a2-a46a-38669f793a92

Also tested:

`panel serve app.py --num-procs 2 --warm`
`panel serve app.py --num-procs 2 --autoreload`

PR #7289, opened 2024-09-17
Only postpone setup script if --num-procs is greater than 1

hoxbro:
See https://github.com/holoviz/panel/pull/6913#issuecomment-2353702106

There is no need to postpone the setup script if we do not run with multiprocessing.

### DIFF
--- a/panel/command/serve.py
+++ b/panel/command/serve.py
@@ -249,7 +249,7 @@ class Serve(_BkServe):
         ('--setup', dict(
             action  = 'store',
             type    = str,
-            help    = "Path to a setup script to run before server starts.",
+            help    = "Path to a setup script to run before server starts. If --num-procs is enabled it will be run in each process after the server has started.",
             default = None
         )),
         ('--liveness', dict(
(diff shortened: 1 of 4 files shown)


panel/command/serve.py
```diff
@@ -370,6 +370,21 @@ def customize_kwargs(self, args, server_kwargs):
...
 
+            def setup_file():
+                setup_path = state._setup_module.__dict__['__file__']
+                with open(setup_path) as f:
+                    setup_source = f.read()
+                nodes = ast.parse(setup_source, os.fspath(setup_path))
+                code = compile(nodes, filename=setup_path, mode='exec', dont_inherit=True)
+                exec(code, state._setup_module.__dict__)
+
+            if args.num_procs > 1:
```
hoxbro, line 381: I need to double-check if this is also a number on Windows.

hoxbro: It is.

https://github.com/bokeh/bokeh/blob/b1e7d94e13f11727de1e114e708740eeb19abe41/src/bokeh/command/subcommands/serve.py#L730-L737

philippjfr (approved):

PR #7290, opened 2024-09-17
Clarified wording in best practices docs

jbednar:
Added additional explanation where I thought it would be helpful.

### DIFF
--- a/doc/how_to/best_practices/dev_experience.md
+++ b/doc/how_to/best_practices/dev_experience.md
@@ -22,7 +22,7 @@ The best practices described on this page serve as a checklist of items to keep 
 
 ### Good
 
-Be sure to bind on `obj.param.{parameter}`, not just `{parameter}`.
+Be sure to bind `obj.param.{parameter}` (the Parameter object), not just `{parameter}` (the current Parameter value).
 
 ```{pyodide}
 def show_clicks(clicks):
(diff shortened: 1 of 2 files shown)


Coderambling:
Very useful! Especially the more verbose emphasis on the distinction between parameter as an object, and the value of a parameter.

Would be good to state the distinction that explicitly in all the docs where applicable.

ahuang11 (approved):
Thanks for the additional clarifications!

jbednar:
Waiting on https://github.com/holoviz/panel/pull/7286 to fix the CI.

Coderambling:
Very useful additions to the best practices document.

One remark: the paragraph below inthe doc ends with:

> because `Viewer` allows direct invocation of the class, resembling a native Panel object.

With "invocation of the class", do you mean invocation of the `param.Parameterized` class? Or if not, what class? Would be good to make this more explicit in the sentence.

> Instead of inheriting from `param.Parameterized`, using `pn.viewable.Viewer` allows direct invocation of the class, resembling a native Panel object.
`param.Parameterized` is a very general class that can be used separately from Panel for working with Parameters. But if you want a Parameterized class to use with Panel, it is usually appropriate to inherit from the Panel-specific class `pn.viewable.Viewer` instead, because `Viewer` allows direct invocation of the class, resembling a native Panel object.

jbednar:
> With "invocation of the class", do you mean invocation of the param.Parameterized class? Or if not, what class? Would be good to make this more explicit in the sentence.

I think that's a question for @ahuang11 , who can hopefully clarify that bit of the text!

Issue #7296, opened 2024-09-18
panel.chat.langchain Import recursion error


postelrich:
The new version, 1.5.0, introduces a recursive import caused by this change https://github.com/holoviz/panel/commit/d0744c50c66866396272e056282c3df38f33ecb9#diff-3e786c7bf61e242b5ec4094db9bb91fab89546aadfa12ed11b56ca0c2a142f71R56. `import_module` protected against this.

#### ALL software version info
(this library, plus any other relevant software, e.g. bokeh, python, notebook, OS, browser, etc should be added within the dropdown below.)

<details>
    <summary>Software Version Info</summary>

```plaintext
1.5.0    
```    
</details>

#### Complete, minimal, self-contained example code that reproduces the issue

```python
import panel
panel.chat.langchain
RecursionError: maximum recursion depth exceeded
```

PR #7297, opened 2024-09-18
Fix langchain import

ahuang11:
Closes https://github.com/holoviz/panel/issues/7296

Wondering if we should deprecate this module too and move it to the langchain-community?

### DIFF
--- a/panel/chat/__init__.py
+++ b/panel/chat/__init__.py
@@ -52,10 +52,9 @@ def __getattr__(name):
     """
     Lazily import langchain module when accessed.
     """
-    import importlib as _importlib
-
     if name == "langchain":
-        return _importlib.import_module("panel.chat.langchain")
+        import importlib
+        return importlib.import_module("panel.chat.langchain")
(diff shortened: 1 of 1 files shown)


panel/chat/__init__.py
```diff
@@ -52,9 +52,10 @@ def __getattr__(name):
     """
     Lazily import langchain module when accessed.
     """
+    import importlib as _importlib
+
     if name == "langchain":
-        from . import langchain
-        return langchain
+        return _importlib.import_module("panel.chat.langchain")
```
hoxbro, line 58:
```suggestion
        import importlib
        return importlib.import_module("panel.chat.langchain")
```

And delete the import outside of the if statement.

panel/chat/__init__.py
```diff
@@ -52,9 +52,10 @@ def __getattr__(name):
     """
     Lazily import langchain module when accessed.
     """
+    import importlib as _importlib
+
```
hoxbro, line 56:
```suggestion
```

hoxbro (approved):

Issue #7295, opened 2024-09-18
Tabulator list editor inside a template / jupyter lab will not stay fixed to the cell


hoxbro:
Tabulator list editor inside a template / jupyter lab will not stay fixed to the cell. 

#### ALL software version info
(this library, plus any other relevant software, e.g. bokeh, python, notebook, OS, browser, etc should be added within the dropdown below.)

<details>
    <summary>Software Version Info</summary>

```plaintext
panel 1.5.0      
```    
</details>

#### Complete, minimal, self-contained example code that reproduces the issue

```python
import pandas as pd
import panel as pn

pn.extension("tabulator")  # works
# pn.extension("tabulator", template="bootstrap")  # does not work

tabulator_editors = {"0": {"type": "list", "values": list(range(100))}}
df = pd.DataFrame(range(30))
pn.widgets.Tabulator(df, editors=tabulator_editors).servable()
```

#### Screenshots or screencasts of the bug in action

https://github.com/user-attachments/assets/b14af974-9152-4a21-a45e-db8282ec2c02

hoxbro:
An example without a template:

``` python
import pandas as pd
import panel as pn

pn.extension("tabulator")  # works

tabulator_editors = {"0": {"type": "list", "values": list(range(100))}}
df = pd.DataFrame(range(30))
tab = pn.widgets.Tabulator(df, editors=tabulator_editors)

pn.Column(tab, max_height=400, styles={"overflow-y": "auto"}).servable()
```

hoxbro:
I can recreate it this without Panel:

``` HTML
<!doctype html>
<html lang="en">
  <head>
    <meta charset="UTF-8" />
    <meta name="viewport" content="width=device-width, initial-scale=1.0" />
    <title>Tabulator with List Editor</title>

    <link
      href="https://unpkg.com/tabulator-tables@6.2.1/dist/css/tabulator.min.css"
      rel="stylesheet"
    />
    <script src="https://unpkg.com/tabulator-tables@6.2.1/dist/js/tabulator.min.js"></script>
  </head>
  <body>
    <div id="example-container" style="height: 200px; overflow-y: auto">
      <div id="example-table" style="width: 200px"></div>
    </div>

    <script>
      let tableData = []
      for (let i = 0; i < 30; i++) {
        tableData.push({ no: i })
      }

      let values = []
      for (let i = 0; i < 30; i++) {
        values.push(i)
      }

      const table = new Tabulator("#example-table", {
        data: tableData,
[...]

PR #7299, opened 2024-09-19
Set Tabulator to render into explicit popup container

philippjfr:
Not entirely happy with this as the editor isn't aligned quite right but I played with all the options and couldn't find one that produced the correct result.

Fixes https://github.com/holoviz/panel/issues/7295

### DIFF
--- a/panel/models/tabulator.ts
+++ b/panel/models/tabulator.ts
@@ -824,7 +824,7 @@ export class DataTabulatorView extends HTMLBoxView {
       paginationMode: this.model.pagination,
       paginationSize: this.model.page_size || 20,
       paginationInitialPage: 1,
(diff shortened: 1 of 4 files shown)


philippjfr:
Yep, seeing that too. Not sure what to do about it.

hoxbro:
What about making this setting an option for now?

philippjfr:
So I'm leaning towards merging this as is now. It turns out the weird offset issue occurs only for the last column and in the meantime this PR also fixes:

https://github.com/holoviz/panel/issues/8096

and the coloring of the popup.

philippjfr:
Okay, fixed the offset issue by inserting an empty last column.

PR #7301, opened 2024-09-19
Tweak best practices

ahuang11:
Clarifies some snippets further and adds a section.

### DIFF
--- a/doc/how_to/best_practices/dev_experience.md
+++ b/doc/how_to/best_practices/dev_experience.md
@@ -46,11 +46,13 @@ clicks = pn.bind(show_clicks, button.clicks)  # not button.clicks!
 pn.Row(button, clicks)
 ```
 
-## Inherit from `pn.viewer.Viewer`
+## Inherit from `pn.viewer.Viewer` or `pn.custom.PyComponent`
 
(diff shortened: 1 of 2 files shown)


Coderambling:
Looks great. The Good, Okay, Bad, Wrong narrative really makes sense!

doc/how_to/best_practices/dev_experience.md
```diff
@@ -68,6 +70,17 @@ class ExampleApp(pn.viewable.Viewer):
 ExampleApp().servable();
 ```
 
+`Viewer` is ideal if you're building a class with some fairly specific business logic, *but* if you are simply building a reusable component assembled from other Panel components use [`PyComponent`](https://panel.holoviz.org/reference/custom_components/PyComponent.html) instead.
```
philippjfr, line 73:
```suggestion
`Viewer` is ideal if you're building a class with some fairly specific business logic, *but* if you are building a reusable component assembled from other Panel components use [`PyComponent`](https://panel.holoviz.org/reference/custom_components/PyComponent.html) instead.
```

philippjfr (approved):

PR #7304, opened 2024-09-20
Ensure Tabulator selection consistency when pagination='local'

philippjfr:
When pagination was set to 'local', selections were not being computed correctly when `header_filters` were being applied. Specifically the problem was that the `header_filters` were still being applied on the server, which would cause the `selection` going out of sync with the data that was visible to the frontend. We fix this by not applying `header_filters` on the server in this scenario, and only when the user accesses the `current_view` do we apply the additional filters so that the view the user sees still matches what is visible on the frontend.

### DIFF
--- a/panel/models/tabulator.ts
+++ b/panel/models/tabulator.ts
@@ -918,7 +918,6 @@ export class DataTabulatorView extends HTMLBoxView {
     if (this.model.configuration.dataTree) {
       data = group_data(data, this.model.columns, this.model.indexes, this.model.aggregators)
     }
-    console.trace(data)
     return data
   }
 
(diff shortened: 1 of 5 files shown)


panel/models/tabulator.ts
```diff
@@ -897,10 +890,17 @@ export class DataTabulatorView extends HTMLBoxView {
...
+    let data: any[]
+    if (cds === null || (cds.columns().length === 0)) {
+      data = []
+    } else {
+      data = transform_cds_to_records(cds, true)
+    }
     if (this.model.configuration.dataTree) {
       data = group_data(data, this.model.columns, this.model.indexes, this.model.aggregators)
     }
+    console.trace(data)
```
philippjfr, line 903:
```suggestion
```

PR #7308, opened 2024-09-20
Timepicker: stricter parameters and minor doc changes

maximlt:
Funnily, I've noticed the default `clock` value is `12h` in Panel and `24h` in Bokeh. I bet the Bokeh dev is based in Europe and the Panel dev in the US 🙃


maximlt:
(I would have had a small preference for allowing only a single type for `value`, `start` and `end`, as `datetime.time`, instead of allowing that and `str` even if that's what Bokeh does).

panel/widgets/input.py
```diff
@@ -830,7 +830,7 @@ class _TimeCommon(Widget):
     selectable, and AM/PM depending on the `clock` option.
     """)
 
-    clock = param.String(default='12h', doc="""
```
philippjfr, line 833: Honestly should have aligned this with `DatetimePicker`:

```
military_time = param.Boolean(default=True, doc="""
      Whether to display time in 24 hour format.""")
```

philippjfr (approved):

PR #7310, opened 2024-09-20
Tweak ChatMessage margins

ahuang11:
Without the divider, the reaction icons were too far left.
image

Then I noticed that the chat copy icon and reaction icons were misaligned vertically.
image

I also noticed that the chat copy icon and reaction icons inline spacing were somewhat different (more space between each reaction icon vs reaction icons + chat copy icon)
image

So I tweaked it all:

Before:
image

After:
image


philippjfr (approved):
I really don't think making single pixel adjustments is the correct way to fix this, but for now it's an improvement and I'll merge.

PR #7318, opened 2024-09-24
Move session prefixes from handlers to Application

philippjfr:
Moving the handling of relative paths and the base urls from the tornado handlers to the `Application` ensures that all server implementations inherit this handling rather than merely being supported by the Tornado server implementation.

Fixes https://github.com/bokeh/bokeh-fastapi/issues/24

### DIFF
--- a/panel/io/application.py
+++ b/panel/io/application.py
@@ -12,6 +12,7 @@ from types import FunctionType, MethodType
 from typing import (
     TYPE_CHECKING, Any, Callable, Mapping,
 )
+from urllib.parse import urljoin
 
 import bokeh.command.util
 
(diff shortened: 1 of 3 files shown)


panel/io/application.py
```diff
@@ -121,8 +122,30 @@ def add(self, handler: Handler) -> None:
...
+        rel_path = '/'.join(['..'] * app_context._url.strip('/').count('/'))
+
+        with set_curdoc(doc):
+            # Handle autoload.js absolute paths
+            abs_url = request.arguments.get('bokeh-absolute-url')
+            if abs_url:
+                rel_path = abs_url[0].decode('utf-8').replace(app_context._url, '')
+
+            state.base_url = base_url
+            state.rel_path = rel_path
```
philippjfr, line 144:
```suggestion
        # Handle autoload.js absolute paths
        abs_url = request.arguments.get('bokeh-absolute-url')
        if abs_url:
            rel_path = abs_url[0].decode('utf-8').replace(app_context._url, '')

        with set_curdoc(doc):
            state.base_url = base_url
            state.rel_path = rel_path
```

Issue #7309, opened 2024-09-20
Date not showing properly in a Hovertemplate for a Plotly map. 


AndreeVela:
#### ALL software version info
<details>
    <summary>Software Version Info</summary>

```plaintext
Apple M2 Sonoma 14.6.1 (23G93)
Google Chrome
Python 3.12
Pip 24.2

IPython          : 8.26.0
ipykernel        : 6.29.5
ipywidgets       : 8.1.2
jupyter_client   : 8.6.2
jupyter_core     : 5.7.2
jupyter_server   : 2.14.1
jupyterlab       : 4.2.3
nbclient         : 0.10.0
nbconvert        : 7.16.4
nbformat         : 5.9.2
notebook         : not installed
qtconsole        : not installed
traitlets        : 5.14.3

folium                    0.17.0
hvplot                    0.10.0
kaleido                   0.2.1
matplotlib                3.9.2
matplotlib-inline         0.1.6
numpy                     2.0.1
pandas                    2.2.2
panel                     1.4.4
plotly                    5.22.0
scipy                     1.14.0
seaborn                   0.13.2
```    
</details>

#### Description of expected behavior and the observed behavior
 When plotting a Map using panels and Plotly extension, and the `Scattermapbox` function, dates data passed in the `customdata` attribute, and later used  in the `hovertemplate` are not displayed correctly. Instead they are displayed as the seconds from 1970 representation. Plotting the Fig directly without wrapping it in a Plotly pane shows the expected behavior.

#### Complete, minimal, self-contained example code that reproduces the issue

```python
import pandas as pd
import numpy as np
import plotly.graph_objects as go
import panel as pn

# Enable Panel extensions
pn.extension('plotly')

# Create sample time series data with latitude and longitude
np.random.seed(42)
n_points = 100
data = pd.DataFrame({
    'timestamp': pd.date_range(start='2023-01-01', periods=n_points, freq='T'),
    'latitude': np.cumsum(np.random.randn(n_points) * 0.01) + 52.3702,  # Centered around Amsterdam
    'longitude': np.cumsum(np.random.randn(n_points) * 0.01) + 4.8952,  # Centered around Amsterdam
})

# Function to create a map using plotly.graph_objects and Scattermapbox
def create_map(data):
    fig = go.Figure()

    # Add the route using Scattermapbox
    fig.add_trace(go.Scattermapbox(
        lat=data['latitude'],
        lon=data['longitude'],
        mode='markers+lines',
[...]

Issue #5252, opened 2023-07-14
Plotly pane ignored timezone information when adding shapes to a plot with a time axes and used local timezone instead


owenlamont:
#### ALL software version info
OS: Windows 11 Pro 10.0.22621
Browser: Vivaldi 6.1.3035.111 (Stable channel) (64-bit)
Panel 1.2.0
Plotly 5.15.0
Bokeh 3.1.1
Jupyterlab 4.0.2

#### Description of expected behavior and the observed behavior

When adding a Plotly shape (in this case a vline) to a time axes plot - I noticed the timezone of the shape position is ignored and my local timezone (in this case +9:30 UTC) was used instead. Plotting with Plotly directly works fine, this only happens when passing a plot to a Plotly pane.

#### Complete, minimal, self-contained example code that reproduces the issue

Here's a code example from Jupyter running in a conda environment (I have reproduced the same issue in PyScript too).

```python
from datetime import datetime, UTC
import panel as pn
import plotly.express as px
import pandas as pd
```

```python
pn.extension("plotly")
```

```python
start = pd.Timestamp('2022-05-11 0:00:00', tz=UTC)
date_range = pd.date_range(start=start, periods=20, freq='h')

df = pd.DataFrame({
    "x": date_range,
    "y": list(range(len(date_range))),
})

plotly_figure = px.scatter(df,
                           x="x",
                           y="y",
                           width=640,
                           height=480,
                          )
plotly_figure.add_vline(
    x=pd.Timestamp('2022-05-11 9:00:00', tz=UTC),
    line_dash="dash",
    line_color="green",
)
plotly_figure
```

```python
pn.pane.Plotly(plotly_figure)
```

#### Screenshots or screencasts of the bug in action

![image](https://github.com/holoviz/panel/assets/12672027/76e4cfd3-0135-4d01-8202-25c0f2461d31)

owenlamont:
Any chance this can be triaged @MarcSkovMadsen / @philippjfr so I can have some idea of when it might get actioned? I know it's a pretty niche edge case but it is impacting my co-workers that use a dashboard I made with Panel and I thought distorting data on a plot was a relatively serious bug.

I've got a lot going on right now but I will see if I can dive into the weeds a bit deeper in a week or two to try and narrow down where this might be going wrong.

owenlamont:
[...]

PR #7319, opened 2024-09-24
Convert Plotly datetimes

philippjfr:
Fix https://github.com/holoviz/panel/issues/7309 and fix https://github.com/holoviz/panel/issues/5252


hoxbro:
Could this potentially also fix https://github.com/holoviz/panel/issues/5252?

philippjfr:
No, to fix that we'd have to traverse the entire JSON struct and cast all datetime types to strings.

philippjfr:
Which I guess we should do :/

PR #7336, opened 2024-09-28
Fix regression handling selection on multi-indexed Tabulator data

philippjfr:
Fixes regression introduced in #7304

### DIFF
--- a/panel/widgets/tables.py
+++ b/panel/widgets/tables.py
@@ -1449,6 +1449,8 @@ class Tabulator(BaseTable):
         if self.value is None or self.style is None or self.value.empty:
             return {}
         df = self._processed
+        if len(self.indexes) > 1:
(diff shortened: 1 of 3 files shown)


panel/widgets/tables.py
```diff
@@ -1735,7 +1738,8 @@ def _map_indexes(self, indexes, existing=[], add=True):
         else:
             start = 0
         ilocs = list(existing)
-        index = self._processed.iloc[[start+ind for ind in indexes]].index
+        df = self._processed
+        index = df.iloc[[start+ind for ind in indexes]].index
```
philippjfr, line 1742:
```suggestion
        index = self._processed.iloc[[start+ind for ind in indexes]].index
```

hoxbro (approved):

PR #7339, opened 2024-09-29
Fix typo for FastListTemplate.ipynb

jrycw:
It seems that `FastGridTemplate` should be updated to reference `FastListTemplate` instead.


philippjfr (approved):

philippjfr:
Thank you!

PR #7341, opened 2024-09-29
Fix Fast checked-indicator color

MarcSkovMadsen:
## Before

image

## After

image

image

### DIFF
--- a/panel/template/fast/fast.css
+++ b/panel/template/fast/fast.css
@@ -2,6 +2,10 @@
   --track-width: 2;
 }
 
+#header-design-provider fast-switch::part(checked-indicator) {
+  --neutral-foreground-rest: #181818;
+}
+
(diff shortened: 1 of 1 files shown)


panel/template/fast/fast.css
```diff
@@ -2,6 +2,10 @@
   --track-width: 2;
 }
 
+#header-design-provider fast-switch::part(checked-indicator) {
+  --neutral-foreground-rest: #181818;
```
philippjfr, line 6: Doesn't this have to be set depending on the theme?

MarcSkovMadsen: I don't believe so. The "circle" in the swatch should always be black. It will always be shown on a white background. See pictures above.

philippjfr (approved):

PR #7344, opened 2024-09-29
docs: Ignore warnings in documentation

hoxbro:
I'm not sure if we ignore warnings with the current setup. If we don't currently ignore warnings, we could add this environment variable so warnings are not shown in documentation. If we do ignore them, this PR can be closed. 

xref: https://github.com/holoviz-dev/nbsite/issues/312


philippjfr (approved):
Maybe ignore them for the prod build but leave them for the dev build?

PR #7346, opened 2024-09-29
Improve plotly style guide

MarcSkovMadsen:
Motivated by learning from https://py.cafe/panel-org/panel-iris-analysis.

### DIFF
--- a/panel/io/state.py
+++ b/panel/io/state.py
@@ -484,7 +484,11 @@ class _state(param.Parameterized):
 
         Note: Keyword arguments must be hashable.
 
-        Example:
+        Pandas Example:
+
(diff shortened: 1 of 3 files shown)


panel/io/state.py
```diff
@@ -484,7 +484,11 @@ def as_cached(self, key: str, fn: Callable[[], T], ttl: int = None, **kwargs) ->
 
         Note: Keyword arguments must be hashable.
 
-        Example:
+        Pandas Example:
+
+        >>> data = pn.state.as_cached('data', pd.read_csv, filepath_or_buffer=DATA_URL)
```
MarcSkovMadsen, line 489: I believe this example is great to add because `pn.read_xyz` is probably the most used function to cache.

philippjfr (approved):

Issue #6953, opened 2024-07-03
Feature Request: CalendarMonthWidget


ahuang11:
I want something like this 

<img width="682" alt="image" src="https://github.com/holoviz/panel/assets/15331990/9a563d0c-da1c-4d22-a67e-965b1c9ffb04">

I imagine the `value` can be a dict:

```
{
  '2024-02-01': "Some event",
  '2024-02-02 12:00': pn.widgets.TextInput(),  # maybe?
}
```

MarcSkovMadsen:
You can find an evaluation of js calendar libraries in https://medium.com/flatlogic/top-javascript-calendar-plugins-aaac8176a759.

The most popular option seems to be https://github.com/fullcalendar/fullcalendar. There is an open source and free version (MIT license) as well as a premium version. You seem to be able to come a long way with the MIT version.

The best alternative from my point of view looks like https://github.com/jquense/react-big-calendar which is fully MIT licenses and inspired by the above calendar.

ahuang11:
This is now on https://panel-extensions.github.io/panel-full-calendar/

PR #7356, opened 2024-10-03
Start adding full calendar

ahuang11:
Closes https://github.com/holoviz/panel/issues/6953

Focusing only on the view options for now

Not sure how to handle interactions
https://fullcalendar.io/docs/date-clicking-selecting

- [x] docs
- [x] tests
- [x] fix re-rendering

```python
import panel as pn

pn.extension()

cal = pn.widgets.Calendar(sizing_mode="stretch_both")

cal.add_event("2024-10-01 12:00")
cal.show()
```

image


ahuang11:
Having trouble re-rendering the calendar on option update.

panel/widgets/calendar.py
```diff
@@ -0,0 +1,244 @@
...
+
+    def _handle_msg(self, msg):
+        if "current_date" in msg:
+            self.current_date = datetime.datetime.strptime(
+                msg["current_date"], "%Y-%m-%dT%H:%M:%S.%fZ"
+            )
+        else:
+            raise NotImplementedError(f"Unhandled message: {msg}")
+
+    def add_event(
```
ahuang11, line 218: Should add remove event too

ahuang11: Should also add a way to change view

panel/models/fullcalendar.js
```diff
@@ -0,0 +1,104 @@
+import {Calendar} from "@fullcalendar/core"
```
MarcSkovMadsen, line 1: To be honest I don't believe this component is "core" Panel. Optimally this would be in a separate package like `panel-mermaid`.

It increases the complexity of the Panel project. And the download size which is important in for example Pyodide.

ahuang11: > If loading plugins dynamically means loading modules dynamically then

I meant part of the compilation.

Is 136 KBs significant?
https://bundlephobia.com/package/@fullcalendar/core@6.1.15

MarcSkovMadsen: Its not the js I'm "worried" about. Its the panel python package.

philippjfr: I'm also leaning towards splitting this out. That said:

>And the download size which is important in for example Pyodide

Pyodide bundles do not and will not ship with the JS bundles.

philippjfr: I guess many people will get Panel directly from PyPI. We probably should finally figure out how to split that package, so `panel/dist` is shipped separately. For backward compatibility we'd probably have to create a new package called `panel-core` and then make `panel` a metapackage that depends on `panel-core` and additionally ships `panel/dist`.

ahuang11:
This is ready for review, besides the compilation part, which I don't really know what to do. 

> Optimally this would be in a separate package like panel-mermaid.

My gut is saying that this should be part of core Panel since "it's just a calendar" (can't really articulate it very well). To me, it's like Tabulator, and I honestly think panes like VTK & Vizzu should be split out. If not, I suppose this could be a good test case for using https://github.com/philippjfr/panel-copier-template and preparing the community template pages

Coderambling:
> This is ready for review, besides the compilation part, which I don't really know what to do.
> 
> > Optimally this would be in a separate package like panel-mermaid.
> 
> My gut is saying that this should be part of core Panel since "it's just a calendar" (can't really articulate it very well). To me, it's like Tabulator, and I honestly think panes like VTK & Vizzu should be split out. If not, I suppose this could be a good test case for using https://github.com/philippjfr/panel-copier-template and preparing the community template pages

Agree that calendar functionality should be in core. In terms ofwidgets, a calendar is standard in many libraries, and it seems likely that this will be used often.

ahuang11:
This is now on https://panel-extensions.github.io/panel-full-calendar/

Issue #7359, opened 2024-10-05
Enable Perspective to support large tables


MarcSkovMadsen:
I work with lots of tabular datasets of the size 50-500MB. For exploratory data analysis the Perspective Viewer is really powerful and unique. Unfortunately sending the datasets to the client is slow and often breaks of a websocket max limitation imposed by Panel, JupyterHub or Kubernetes. You can increase these limits but only to some extend and also this can be outside the control or capability of a data scientist.

I'm increasingly seeing this problem and I'm not the only one seeing this problem ([Discourse #6804](https://discourse.holoviz.org/t/websocket-max-message-size-is-not-respected/6804)). Its actually a problem that is very common in Finance and Trading where I work. Currently Excel support larger tables than we do with Perspective. I believe we should enable users to work with larger files than Excel can in Perspective.

Actually Perspective was built to support large tabular data via virtualization. See [`regular-table`](https://github.com/finos/regular-table) and [Perspective](https://perspective.finos.org/). But our implementation only use the `perspective-viewer` web component. Not the advanced client-server virtualization architecture supported.

A Panel user actually showcased how to use the client-server virtualization in [Discourse #6430](https://discourse.holoviz.org/t/panel-perspective-on-the-server/6430). But its only a complicated to use proof of concept.

Please note that the client-server virtualization architecture seems similar to Mosaic - Mosaic is just built on DuckDB. There is a request to add Mosaic in [FR #7358](https://github.com/holoviz/panel/issues/7358).

## Discussion

The Tabulator Pane provides a kind of virtualization via the `pagination` parameter ("local" or "remote"). We could support a similar parameter with Perspective making it really easy for users. On the other hand there is power in exposing more of the underlying Perspective api like the PerspectiveManager and hosting tables once but using across sessions and users. I think also Panel Perspective pane would be more useful if it implemented the Jupyter Perspective Widget api and capabilities. See [JupyterCon 2020](https://www.youtube.com/watch?v=IO-HJsGdleE), [Perspective Basics](https://github.com/sc1f/streaming-perspective-jupytercon-2020/blob/master/perspective_basics.ipynb) and [PerspectiveWidget Implementation](https://github.com/finos/perspective/blob/master/rust/perspective-python/perspective/widget/__init__.py) for inspiration.

Today Panel can be running on both Tornado and FastAPI servers. The solution should work in both environments. Personally I want to migrate to FastAPI deployments if that is possible.

## Cannot use JupyterWidget

[...]

PR #7368, opened 2024-10-07
Support Perspective Backend

MarcSkovMadsen:
Closing #7359

Currently this is exploration and analysis

### DIFF
--- a/analysis/README.md
+++ b/analysis/README.md
@@ -1,0 +1,99 @@
+# Perspective Analysis
+
+## The Problem
+
+Panel has its own [Panel Perspective Widget](https://panel.holoviz.org/reference/index.html).
+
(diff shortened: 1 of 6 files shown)


philippjfr:
Also see https://github.com/holoviz/panel/pull/2752

MarcSkovMadsen:
> Also see #2752

Having analyzed the problem I think a great solution would be to enable users to specific a *named* table and a Perspective *backend* web socket. Then make it really easy to add a Perspective *backend* web socket to Panel. The question is whether is should be almost hidden or specific via https://github.com/holoviz/panel/pull/2752.

Coderambling:
Support for Perspective as a backend should be included in the new roadmap. Do you agree @philippjfr, @MarcSkovMadsen, @jbednar ?

The use cases are laid out in the issue by Marc, and I could see it entice Perspective users to Panel and the Holoviz ecosystem.

The roadmap currently includes implementing support for Narwhals.

Not sure if direct complications would arise from doing both Narwhals, and implementing Perspective as a backend. Would there be?

#8339 

#7359

Issue #1972, opened 2021-02-10
DatetimeSlider and other missing widgets


jbednar:
Panel widgets are tied to a specific Python datatype (int, float, etc.) and various modes of interaction (slider, text box, etc.), but not all combinations are covered by existing widgets (with missing ones indicated with a `?`):

|                   | **slider**        | **input**           | **player**            | **rangeslider**       | **rangeinput**      |
| ----------------- | ----------------- | ------------------- | --------------------- | --------------------- | ------------------  |
| **int**           | IntSlider         | IntInput            | Player                | Intrangeslider        | IntrangeInput?      |
| **float**         | FloatSlider       | FloatInput          | Options for Player?   | Rangeslider           | FloatrangeInput?    |
| **date**          | Dateslider        | DateInput?          | Options for Player?   | Daterangeslider       | DaterangeInput?     |
| **datetime**      | Datetimeslider?   | Datetimeinput       | Options for Player?   | Datetimerangeslider?  | Daterangetimeinput  |
| **discrete**      | DiscreteSlider    | (Autocompleteinput) | DiscretePlayer        | (Multiselect)         | N/A?                |

My immediate use case is for a Datetimeslider, so that I can scroll through a point in time the same way I can scroll through a floating-point value. Any workaround I can think of seems very painful: Floatslider with a formatter to show as a date (awkward to work with, imprecise), DatePicker plus a time-of-day slider (hard to lay out, requires resetting time slider each midnight), DateSlider plus a time-of-day slider (requires resetting time slider each midnight). 

Other missing widgets also seem useful, though the workarounds seem better in those cases: you can use two IntInputs or FloatInputs as an IntrangeInput or FloatrangeInput (though without inherent validation for ordering), you can use a MultiSelect to select a range of discrete values (though without enforcing it to be a contiguous range necessarily), etc. 

Would it be possible to simply dispense with having all these combinations and have a type argument for the widget instead? E.g. single Slider, Input, Player, RangeSlider, and RangeInput widget classes that can be configured as int, float, date, or datetime? Even better, they could deduce their type from the initial value, if not specified, which would leave things quite a bit cleaner in user code. Seems like this approach would make it easier to find stuff in the reference guides (easily landing on the page about sliders if I want a slider).

[Not sure if Discrete fits into this categorization, really; arguably range sliders and inputs are already covered for the discrete case by MultiSelect and similar widgets and by Autocompleteinput. But I figured I'd include that row in the table here just in case it resonated somehow.]

[...]

PR #7374, opened 2024-10-08
Add DatetimeSlider widget

thuydotm:
This MR adds DatetimeSlider widget, which is mentioned in https://github.com/holoviz/panel/issues/1972. The code is mainly based on Bokeh DateSlider model and the existing Panel DateSlider widget. Huge thanks to @hoxbro for your support!

TODO:
- [ ] add tests
- [ ] add reference notebook


thuydotm:
I’m think about the unit for `step` param? Looking at the existing datetime select widget, DatetimeRangeSlider, the unit is in ms. But we only display the information up to seconds with format `"%d %b %Y %H:%M:%S"`, should it be second instead of ms?

thuydotm:
[pre-commit.ci](http://pre-commit.ci/) autofix

thuydotm:
pre-commit.ci autofix

panel/models/datetime_slider.ts
```diff
@@ -0,0 +1,51 @@
...
+import {isString} from "@bokehjs/core/util/types"
+
+export class DatetimeSliderView extends NumericalSliderView {
+  declare model: DatetimeSlider
+
+  override behaviour = "tap" as const
+  override connected = [true, false]
+
+  protected _formatter(value: number, format: string | TickFormatter): string {
+    console.log("DatetimeSlider")
```
hoxbro, line 19:
```suggestion
```

panel/models/datetime_slider.py
```diff
@@ -0,0 +1,14 @@
+from bokeh.core.properties import Override
+from bokeh.models.widgets.sliders import DateSlider
+
+
+class DatetimeSlider(DateSlider):
+    """ Slider-based datetime selection widget. """
+
+    # explicit __init__ to support Init signatures
+    def __init__(self, *args, **kwargs) -> None:
+        super().__init__(*args, **kwargs)
```
hoxbro, line 10:
```suggestion
```

This is only really needed in Bokeh.

panel/tests/widgets/test_slider.py
```diff
@@ -156,6 +156,68 @@ def test_date_slider(document, comm):
...
+    widget = datetime_slider.get_root(document, comm=comm)
+
+    assert isinstance(widget, datetime_slider._widget_type)
+    assert widget.title == 'DatetimeSlider'
+    assert widget.value == 1536019200000
+    assert widget.start == 1535760000000.0
+    assert widget.end == 1536537600000.0
+
+    epoch = datetime(1970, 1, 1)
+    widget.value = (datetime(2018, 9, 3)-epoch).total_seconds()*1000
```
hoxbro, line 182:
```suggestion
    epoch = datetime(1970, 1, 1)
    epoch_time = lambda x: (x-epoch).total_seconds()*1000
    widget.value = epoch_time(datetime(2018, 9, 3))
```

Also would need changes other places

panel/widgets/slider.py
```diff
@@ -25,6 +25,8 @@
 from bokeh.models.widgets.sliders import NumericalSlider as _BkNumericalSlider
 from param.parameterized import resolve_value
 
+from panel.models.datetime_slider import DatetimeSlider as _PnDatetimeSlider
```
hoxbro, line 28: Use relative import here. 

The convention also seems to still use `_Bk` as a prefix

https://github.com/holoviz/panel/blob/c65c390c7357e1a8c3c3346feadef33a3c41eba0/panel/widgets/select.py#L30

panel/widgets/slider.py
```diff
@@ -319,6 +321,36 @@ def _process_property_change(self, msg):
         return msg
 
 
+class DatetimeSlider(DateSlider):
+    """
+    The DatetimeSlider widget allows selecting a value within a set of
+    bounds using a slider.  Supports datetime.datetime
+    and np.datetime64 values. The step size is fixed at 1 day.
```
hoxbro, line 328:
```suggestion
    and np.datetime64 values. The step size is fixed at 60 seconds.
```

panel/widgets/slider.py
```diff
@@ -319,6 +321,36 @@ def _process_property_change(self, msg):
...
+    ...     start=dt.datetime(2025, 1, 1),
+    ...     end=dt.datetime(2025, 1, 7),
+    ...     name="A datetime value"
+    ... )
+    """
+
+    as_datetime = param.Boolean(default=True, readonly=True, doc="""
+        Whether to store the date as a datetime.""")
+
+    step = param.Number(default=60, doc="""
```
hoxbro, line 346:
```suggestion
    step = param.Number(default=60, bounds=(1, None), doc="""
```

I can see `DatetimeRangeSlider` uses step size in ms, though I agree it makes more sense with a second for a slider.

panel/models/datetime_slider.ts
```diff
@@ -0,0 +1,56 @@
...
+  constructor(attrs?: Partial<DatetimeSlider.Attrs>) {
+    super(attrs)
+  }
+  static override __module__ = "panel.models.datetime_slider"
+
+  static {
+    this.prototype.default_view = DatetimeSliderView
+
+    this.override<DatetimeSlider.Props>({
+      format: "%d %b %Y %H:%M:%S",
```
hoxbro, line 53: Should step not also be defined here?

panel/tests/widgets/test_slider.py
```diff
@@ -156,6 +156,68 @@ def test_date_slider(document, comm):
         assert widget.value == 1620777600000
 
 
+@pytest.mark.parametrize("start", [date(2018, 9, 1), datetime(2018, 9, 1)])
```
hoxbro, line 159: We could also add an option for `np.datetime64`.

hoxbro (commented):
I have left some small comments, but this PR seems solid overall.

thuydotm:
@hoxbro thanks for reviewing! I've pushed the changes as suggested.

panel/models/datetime_slider.py
thuydotm: Thanks! It's good to know.

panel/models/datetime_slider.ts
thuydotm: Yes. Good catch!

panel/tests/widgets/test_slider.py
thuydotm: Just added a new test for np.datetime64.

panel/widgets/slider.py
hoxbro: You forgot to change it to an relative import.

thuydotm: Ah yeah I forgot. Updated it now.

panel/models/timezone.d.ts
```diff
@@ -0,0 +1,3 @@
+declare module "timezone" {
```
hoxbro, line 1: Could this be moved to `declaration.d.ts` file?

panel/widgets/slider.py
```diff
@@ -319,6 +320,36 @@ def _process_property_change(self, msg):
         return msg
 
 
+class DatetimeSlider(DateSlider):
+    """
+    The DatetimeSlider widget allows selecting a value within a set of
+    bounds using a slider.  Supports datetime.datetime
```
hoxbro, line 326:
```suggestion
    bounds using a slider. Supports datetime.datetime
```

Very nit.

hoxbro (approved):

panel/models/datetime_slider.ts
```diff
@@ -0,0 +1,57 @@
+// adapted from bokeh
+// https://github.com/bokeh/bokeh/blob/branch-3.7/bokehjs/src/lib/models/widgets/sliders/date_slider.ts
+
+import tz from "timezone"
```
mattpap, line 4: External libraries like this one are not guaranteed to be importable by this name or a name altogether. I see that our date slider still uses the legacy approach. The right approach is to use `DEFAULT_FORMATTERS.datetime(value, format)` from `core/util/templating` instead. This way you also avoid adding type declarations for `timezone` library.

panel/widgets/slider.py
thuydotm: @hoxbro no, that's a good one. I also re-added `datetime.date` to list of supported dtypes to the docs.

panel/models/datetime_slider.ts
thuydotm: Thanks @mattpap! That's good to know. Applied the changes as suggested.

philippjfr (approved):

philippjfr:
Thanks @thuydotm!!

PR #7382, opened 2024-10-09
Sync dark mode on browser_info object

philippjfr:
Eventually useful when we want to dynamically toggle between dark and light themes.

### DIFF
--- a/panel/models/browser.ts
+++ b/panel/models/browser.ts
@@ -4,7 +4,6 @@ import {Model} from "@bokehjs/model"
 
 export class BrowserInfoView extends View {
   declare model: BrowserInfo
-  _dark_mode_listener: 
 
   override initialize(): void {
     super.initialize()
@@ -13,7 +12,7 @@ export class BrowserInfoView extends View {
       const darkModeMediaQuery = window.matchMedia('(prefers-color-scheme: dark)');
       darkModeMediaQuery.addEventListener('change', (e) => {
 	this.model.dark_mode = e.matches
-      });
+      })
       this.model.dark_mode = darkModeMediaQuery.matches
     }
     this.model.device_pixel_ratio = window.devicePixelRatio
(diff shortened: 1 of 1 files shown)


panel/models/browser.ts
```diff
@@ -4,12 +4,17 @@ import {Model} from "@bokehjs/model"
...
 
   override initialize(): void {
     super.initialize()
 
     if (window.matchMedia != null) {
-      this.model.dark_mode = window.matchMedia("(prefers-color-scheme: dark)").matches
+      const darkModeMediaQuery = window.matchMedia('(prefers-color-scheme: dark)');
+      darkModeMediaQuery.addEventListener('change', (e) => {
+	this.model.dark_mode = e.matches
+      });
```
philippjfr, line 16:
```suggestion
      })
```

panel/models/browser.ts
```diff
@@ -4,12 +4,17 @@ import {Model} from "@bokehjs/model"
 
 export class BrowserInfoView extends View {
   declare model: BrowserInfo
+  _dark_mode_listener:
```
philippjfr, line 7:
```suggestion
```

panel/models/browser.ts
```diff
@@ -9,7 +9,11 @@ export class BrowserInfoView extends View {
     super.initialize()
 
     if (window.matchMedia != null) {
-      this.model.dark_mode = window.matchMedia("(prefers-color-scheme: dark)").matches
+      const darkModeMediaQuery = window.matchMedia('(prefers-color-scheme: dark)');
+      darkModeMediaQuery.addEventListener('change', (e) => {
```
philippjfr, line 13:
```suggestion
      darkModeMediaQuery.addEventListener("change", (e) => {
```

MarcSkovMadsen:
Would this enable that if the user has not set `?theme=...` as query args then `pn.config.theme` is set to his/ hers browser preferences? Today its just set to `default`.

That would be a big improvement.

MarcSkovMadsen:
> That's already possible in principle.

How?

philippjfr:
> How?

```
pn.config.theme = 'default' if state.browser_info.dark_mode else 'default'
```

> And could it (please) be the default behaviour?

Maybe in future, but not in a patch release.

philippjfr:
pre-commit.ci autofix

Issue #7401, opened 2024-10-15
Tabulator may overlap with other components if max_height is set


ahuang11:
<img width="361" alt="image" src="https://github.com/user-attachments/assets/3c5fd5a4-55ca-44c0-9e06-80825b47b798">

```python
import panel as pn
import pandas as pd
import numpy as np

pn.extension("tabulator")

df = pd.DataFrame(
    np.random.randn(10000, 5),
)
pn.Column(pn.widgets.Tabulator(df, max_height=300), "Hello!").show()
```

Replacing max_height with min_height/height will fix it

PR #7403, opened 2024-10-15
Prevent tabulator from overlapping when max_height is set

ahuang11:
Closes https://github.com/holoviz/panel/issues/7401

Before:
image

After, properly truncates

```python
import panel as pn
import pandas as pd
import numpy as np

pn.extension("tabulator")


df = pd.DataFrame(
    np.random.randn(100, 5),
)
pn.Column(pn.widgets.Tabulator(df, max_height=300), "Hello!").servable()
```
image

```python
import panel as pn
import pandas as pd
import numpy as np

pn.extension("tabulator")


df = pd.DataFrame(
    np.random.randn(1, 5),
)
pn.Column(pn.widgets.Tabulator(df, max_height=1000), "Hello!").servable()
```

Does not use up all the space as max_height
image

### DIFF
--- a/panel/tests/ui/widgets/test_tabulator.py
+++ b/panel/tests/ui/widgets/test_tabulator.py
@@ -1159,6 +1159,17 @@ def test_tabulator_patch_no_height_resize(page):
     wait_until(lambda: page.locator('.pnx-tabulator').evaluate(at_bottom_script), page)
 
 
+def test_tabulator_max_height_not_overlap(page):
+    df = pd.DataFrame({'col': np.random.random(100)})
+    widget = Tabulator(df, max_height=200)
+
+    serve_component(page, widget)
+
+    table = page.locator('.pnx-tabulator')
+    expect(table).to_have_css('max-height', '200px')
+    assert table.bounding_box()['height'] <= 200
+
+
 @pytest.mark.parametrize(
     'pagination', ('local', 'remote', None)
 )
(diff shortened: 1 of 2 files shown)


philippjfr:
Also happy with this PR, but that might be simpler and clearer.

ahuang11:
Couldn't just dump it in config without a null check:

image

panel/tests/ui/widgets/test_tabulator.py
```diff
@@ -1159,6 +1159,32 @@ def test_tabulator_patch_no_height_resize(page):
...
+    If max_height is not set, Tabulator should not set it to null;
+    else there's some recursion issues in the console and lag
+    """
+    df = pd.DataFrame({'col': np.random.random(100)})
+    widget = Tabulator(df)
+
+    serve_component(page, widget)
+
+    table = page.locator('.pnx-tabulator')
+    expect(table).not_to_have_css('max-height', '200px')
```
philippjfr, line 1184: Seems like an odd test, why would I have max-height set to 200px in the first place? Seems like it should check that max-height is unset instead.

ahuang11: Okay updated--seems counterintuitive to use `to_have_css` to check for null, but I guess that's the only way!

philippjfr (approved):

Issue #7397, opened 2024-10-14
ExamplePicker widget


MarcSkovMadsen:
Its currently hard to make example apps like Gradio because you have to implement functionality to pick example input values your self.

Please add an ExamplePicker widget to make it easy to pick example values for input to your input widgets.

## Requirements

- Can pick any value/ media (text, numbers, images, audio, video, dataframes/ tables, etc.)
- Can re-pick last picked value (Cannot do that with existing radio widgets)
- Widget layout, look and feel adapts to number of options and type of options
  - Few options then as buttons layed out in FlexBox or similar.
  - Many options then as dropdown or scrollable select.
  - Maybe even search functionality if really a lot of examples 
  - Can use or set background image to make it visible what you are selecting

## Examples

- `ExamplePicker` in https://py.cafe/awesome.panel.org/panel-mermaid-diagram-editor.
- Image Classifier tutorial https://panel.holoviz.org/tutorials/basic/build_image_classifier.html
- Gradio Examples https://www.gradio.app/docs/gradio/examples

PR #7411, opened 2024-10-17
ExamplePicker

MarcSkovMadsen:
Closing #7397

## Illustration

https://github.com/user-attachments/assets/0b0d2443-9f4f-4449-bd68-9639836dfb07

## Todo

- [ ] Support Buttons background thumbnails
- [ ] Support Button icons
- [ ] Support `button_kwargs`
- [ ] Support `layout_type` (FlexBox of Buttons or Column/ Table of Rows)
- [ ] Try many examples including Gradio examples
- [ ] Docstrings
- [ ] Tests
- [ ] Reference Guide

### DIFF
--- a/examples/reference/widgets/example_picker.md
+++ b/examples/reference/widgets/example_picker.md
@@ -1,0 +1,51 @@
+```python
+import panel as pn
+
+pn.extension()
+```
+
+## Basic Usage
+
+You can create an `ExamplePicker` to pick examples for a single target:
+
+```python
+widget = pn.widgets.TextInput(name="Selection")
+example_picker = pn.widgets.ExamplePicker("Hello", "World", targets=widget)
+pn.FlexBox(example_picker, widget).servable()
+```
+
+## Multiple Targets
+
+You can create an `ExamplePicker` to pick examples for multiple targets:
+
+```python
+text_input = pn.widgets.TextInput(name="Text Input")
+audio_output = pn.pane.Audio(name="Audio Output")
+def update_name(name, text_input=text_input):
+    text_input.name=name
+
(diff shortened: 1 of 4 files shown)


panel/widgets/examples.py
```diff
@@ -0,0 +1,295 @@
+import os
```
MarcSkovMadsen, line 1: - The Gradio [`Examples`](https://www.gradio.app/docs/gradio/examples) has another api and many more (advanced) features like *batch inference* and *example caching*. Should we align closer with that api?

panel/widgets/examples.py
```diff
@@ -0,0 +1,295 @@
...
+import re
```
MarcSkovMadsen, line 2: Gradio also displayes the examples in a table. I display using Panes and in a Column. Should we display using Tabulator instead. The pro of not using Tabulator is that we can display everything. The cons is that we might need to instantiate a large number of Bokeh models which is slow. And I don't know if we can make it look pretty.

## Gradio

image


```python
import gradio as gr
from PIL import Image
import pandas as pd

# Example Data
image_example = Image.new('RGB', (100, 100), color = 'red')  # Red square image

examples = [
    ["Hello, World!", 42, 3.14, image_example, pd.DataFrame({"x": [1,2], "y": [1,2]})],
    ["Another text", 7, 0.001, image_example, pd.DataFrame({"x": [3,4], "y": [1,2]},)]
]

with gr.Blocks() as demo:
    gr.Examples(
        examples=examples,
        inputs=[gr.Textbox(), gr.Number(), gr.Number(), gr.Image(), gr.DataFrame()],
    )

demo.launch()
```

Issue #7414, opened 2024-10-17
Setting `objects` on a layout mutates the elements within the new list


maximlt:
Panel 1.5 introduced a behavior change described in this snippet, by which setting `.objects` on a layout mutates the list, replacing the original Viewer instances with the output from their `__panel__` method. I think that the recommended API isn't to set `objects` but to use the `layout[:] = new_elements` annotation which still works as expected. @philippjfr is `layout.objects = new_elements`  a supported API? `git bisect` points to https://github.com/holoviz/panel/pull/5593.

```python
import panel as pn


class SomeView(pn.viewable.Viewer):

    def __panel__(self):
        return pn.pane.Markdown("boom")


my_list = [SomeView(name="a"), SomeView(name="b")]
col = pn.Column()
col.objects = my_list  # bad
# col[:] = my_list     # ok
# col.extend(my_list)  # ok

print(my_list)
# Panel 1.5
# [Markdown(str), Markdown(str)]
# Panel 1.4.5
# [SomeView(name='a'), SomeView(name='b')]
```

philippjfr:
Setting `objects` like this is valid and indeed should not mutate.

PR #7417, opened 2024-10-18
Do not mutate Children inplace

philippjfr:
Fixes https://github.com/holoviz/panel/issues/7414

### DIFF
--- a/panel/viewable.py
+++ b/panel/viewable.py
@@ -1152,11 +1152,9 @@ class Children(param.List):
             new = []
             mutated = False
             for v in val:
-                if not isinstance(v, Viewable):
-                    n = panel(v)
-                    mutated |= v is not n
-                    v = n
-                new.append(v)
+                n = panel(v)
+                mutated |= v is not n
+                new.append(n)
             if mutated:
                 val = new
         return val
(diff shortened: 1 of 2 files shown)


panel/viewable.py
```diff
@@ -1149,10 +1149,16 @@ def __init__(
...
-            val[:] = [
-                v if isinstance(v, Viewable) else panel(v)
-                for v in val
-            ]
+            new = []
+            mutated = False
+            for v in val:
+                if not isinstance(v, Viewable):
+                    n = panel(v)
+                    mutated |= v is not n
```
maximlt, line 1157: Out of curiosity, what kind of object that is not a `Viewable` would calling `panel(obj)` on return itself?

maximlt (approved):
LGTM.

Should `ChildDict._transform_value` also be adapted not to update in place the dict? (I don't think I've ever written `grid.objects = objects` but I don't use grids much).

panel/viewable.py
philippjfr: Good point, it shouldn't return anything else. Will leave this just in case.

philippjfr:
>Should ChildDict._transform_value also be adapted not to update in place the dict? (I don't think I've ever written grid.objects = objects but I don't use grids much).

Yes it should.

Issue #2302, opened 2021-05-09
panel serve logs stops working when importing datashader


hoxbro:
#### ALL software version info
Windows 10  20H2
holoviews 1.14.3
panel 0.11.3
datashader 0.12.1

#### Description of expected behavior and the observed behavior
I'm not sure if this a `panel`, `datashader`, or a `holoviews` issue. 
The problem is when importing `holoviews.operation.datashader` in my file, the logging in the command window stops sending logs. When running`panel serve tmp.py --autoreload` with  `holoviews.operation.datashader` imported no logs show up at all and I have to guess when ready. 

#### Complete, minimal, self-contained example code that reproduces the issue
``` python
import panel as pn
import holoviews.operation.datashader

pn.panel("# Test").servable()

```
And run the command `panel serve` or `panel serve --autoreload`

#### Screenshots or screencasts of the bug in action

|  | No `datashade` import | `datashade` import |
|--|--|--|
| Without `--autoreload` |  ![2021-05-09 17_02_50](https://user-images.githubusercontent.com/19758978/117577459-24225e00-b0ea-11eb-9e2b-e77b45932b3a.png)  |  ![2021-05-09 17_05_44](https://user-images.githubusercontent.com/19758978/117577513-5e8bfb00-b0ea-11eb-92da-f1d3a4c1882a.png) |
| With `--autoreload`    | ![2021-05-09 17_04_59](https://user-images.githubusercontent.com/19758978/117577490-3e5c3c00-b0ea-11eb-8f8b-6b1416677816.png)  |  ![2021-05-09 17_06_36](https://user-images.githubusercontent.com/19758978/117577521-677ccc80-b0ea-11eb-8a64-157f0c25e5b6.png)  |

Hoxbro:
I don't know if it is related to this problem, but I'm have been seeing error message for memory leaks when having `datashader` imported. 

```
bokeh.document.document - ERROR - Module <module 'bokeh_app_2c6de2e007084afdb776cc94fdd6bb90' from 'C:\\Users\\shh\\Desktop\\tmp.py'> has extra unexpected referrers! This could indicate a serious memory leak. Extra referrers: [<cell at 0x000001A0C462DD38: module object at 0x000001A0C466B958>]
```
I can somewhat force this error, but I don't know why.

https://user-images.githubusercontent.com/19758978/117578527-13281b80-b0ef-11eb-9870-bc0441d05b14.mp4

Hoxbro:
[...]

PR #7426, opened 2024-10-21
enh: Set Bokeh log level to info when importing dask.distributed

hoxbro:
Fixes https://github.com/holoviz/panel/issues/2302

This should be non-intrusive (IMO). 

It is pretty annoying not getting any logging when working with holoviews + datashader.

### DIFF
--- a/panel/command/serve.py
+++ b/panel/command/serve.py
@@ -672,6 +672,7 @@ class Serve(_BkServe):
         silence(EMPTY_LAYOUT, True)
         # dask.distributed changes the logging level of Bokeh, we will overwrite it
         # if the environment variable is not set to the default Bokeh level
+        # See https://github.com/holoviz/panel/issues/2302
         if "DASK_DISTRIBUTED__LOGGING__BOKEH" not in os.environ:
             os.environ["DASK_DISTRIBUTED__LOGGING__BOKEH"] = "info"
         args.dev = None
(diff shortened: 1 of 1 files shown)


panel/command/serve.py
```diff
@@ -670,5 +670,9 @@ def invoke(self, args: argparse.Namespace):
         # Empty layout are valid and the Bokeh warning is silenced as usually
         # not relevant to Panel users.
         silence(EMPTY_LAYOUT, True)
+        # dask.distributed changes the logging level of Bokeh, we will overwrite it
```
maximlt, line 673: It'd be nice to link to one of the issues related to this or to this PR https://github.com/dask/distributed/pull/8634.

maximlt (changes requested):

panel/command/serve.py
```diff
@@ -670,5 +670,9 @@ def invoke(self, args: argparse.Namespace):
...
+        # if the environment variable is not set to the default Bokeh level
```
hoxbro, line 674:
```suggestion
        # if the environment variable is not set to the default Bokeh level
        # See https://github.com/holoviz/panel/issues/2302
```

panel/command/serve.py
hoxbro: Done in https://github.com/holoviz/panel/pull/7426/commits/97353bbab0dd34e1fe6d81c5ab6a524423a64de3

maximlt (approved):

philippjfr (approved):

Issue #7433, opened 2024-10-23
Panel Preview not working: KeyError


MarcSkovMadsen:
I'm working on upgrading our jupyterhub and Panel. In our existing setup the Jupyter Panel Preview is working. But not in the new setup.

## Steps

- Log into Jupyter
- Open Notebook
- Select `base` environment
- Create notebook with widget and make it .servable(). (With just display  like `pn.panel('hello').servable()` there is no problem.)
- Click the Panel Preview button
- See that the preview does not show anything and that there is an error in the browser console.

![image](https://github.com/user-attachments/assets/8c56e28d-5d37-4742-a603-606745839551)

Check the server logs

```bash
2024-10-23 11:41:32,007 - ServerApp - ERROR - Uncaught exception GET /mt-fumo/user/masma/panel-preview/render/test.ipynb/ws (10.86.100.104)
HTTPServerRequest(protocol='https', host='mnr-jupyterhub-dev.de-prod.dk', method='GET', uri='/mt-fumo/user/masma/panel-preview/render/test.ipynb/ws', version='HTTP/1.1', remote_ip='10.86.100.104')
Traceback (most recent call last):
  File "/opt/conda/lib/python3.11/site-packages/tornado/websocket.py", line 940, in _accept_connection
    await open_result
  File "/opt/conda/lib/python3.11/site-packages/panel/io/jupyter_server_extension.py", line 392, in open
    kernel_info = state._kernels[self.session_id]
                  ~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^
KeyError: 'XKRUSRxzKSS076zhAKRFPD2dUhH9z29fFg8TLcjZNv9r'
```

philippjfr:
Reading the code I really struggle to see how this is possible UNLESS the websocket connection takes >30 seconds to open. Can you confirm how quickly this happens?

MarcSkovMadsen:
Thanks so much for responding. The preview opens and closes the web socket almost instantanously.


https://github.com/user-attachments/assets/8f481b01-1404-4160-9527-4fc3eeab915f

MarcSkovMadsen:
As I see it `self.close()` must have run for some unknown reason:

![image](https://github.com/user-attachments/assets/2baca037-1559-4a50-b0e9-f19b5cf86630)

MarcSkovMadsen:
[...]

PR #7434, opened 2024-10-23
Improve preview error handling

MarcSkovMadsen:
This is related to https://github.com/holoviz/panel/issues/7433.

It will not fix the issue, but I believe it will handle the issue more correctly.

### DIFF
--- a/panel/io/jupyter_server_extension.py
+++ b/panel/io/jupyter_server_extension.py
@@ -390,7 +390,7 @@ class PanelWSProxy(WSHandler, JupyterHandler):
         if self.session_id not in state._kernels:
             self.close()
             msg = f"Session ID '{self.session_id}' does not correspond to any active kernel."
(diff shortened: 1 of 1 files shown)


panel/io/jupyter_server_extension.py
```diff
@@ -389,6 +389,9 @@ async def open(self, path, *args, **kwargs) -> None:
         self.session_id = get_session_id(token)
         if self.session_id not in state._kernels:
             self.close()
+            msg = f"Session ID '{self.session_id}' does not correspond to any active kernel."
+            raise ProtocolError(msg)
```
philippjfr, line 393: Yes, indeed, this should be logged. I would raise a simple `RuntimeError` here though, it's not ProtocolError since all the protocol validation has been executed correctly by this point.

philippjfr (approved):

PR #7447, opened 2024-10-27
Update conf.py to retrieve version numbers for various libraries, so they can be referred to dynamucally in various docs

Coderambling:
See below PR where this was implemented for Tabulator doc.

https://github.com/holoviz/panel/pull/7053

References need to be added / updated in the   various docs for the respective libraries to dynamically show the library version numbers.

### DIFF
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -273,6 +273,37 @@ def update_versions(app, docname, source):
         "{{VIZZU_VERSION}}" : VIZZU_VERSION,
     }
 
+def update_versions(app, docname, source):
+    from panel.models.deckgl import DECKGL_VERSION
+    from panel.models.echarts import ECHARTS_VERSION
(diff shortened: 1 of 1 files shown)


doc/conf.py
```diff
@@ -273,6 +273,37 @@ def update_versions(app, docname, source):
         "{{VIZZU_VERSION}}" : VIZZU_VERSION,
     }
 
+def update_versions(app, docname, source):
```
philippjfr, line 276: Seems to duplicate the code from just above.

Coderambling: Fixed the duplication @philippjfr .

philippjfr (commented):
Thanks for the fix. Most of these version variables don't exist yet and would have to be defined. We then also need to use these versions consistently across the docs.

Issue #6209, opened 2024-01-16
Make it easy to copy-paste from Markdown code blocks


MarcSkovMadsen:
It would really, really be valuable if Panel code markdown blocks supported copy-paste. All competing frameworks support this.

A small example.

````python
import panel as pn

pn.extension()

pn.Column("""
An example

```python
import panel as pn

pn.extension()
```

```bash
https://panel.holoviz.org
```
""").servable()
````

![image](https://github.com/holoviz/panel/assets/42288570/478b45d1-7553-4f44-8adb-d2c038a9eef8)

I would like something similar to what Quarto provides (below). Its only shown when the upper, right corner is hovered upon.

![image](https://github.com/holoviz/panel/assets/42288570/d26bd34d-f549-4006-bdef-1a2981b32db0)

We also have the copy-paste functionality in our docs. It could also be similar to that. I'm just not a fan of the green check mark that shows when the code has been copied as the green does not go well with the colors used in my (work) apps.

## Workaround

Make a separate copy button. Its not as nice a UX though.

```python
import panel as pn

DEFAULT_KWARGS = {
    "button_type": "default",
    "description": "Copy to the clipboard",
    "icon": "clipboard",
    "sizing_mode": "fixed",
    "width": 30,
    "margin": 0,
}


def create_copy_to_clipboard_button(text, **kwargs):
    for key, value in DEFAULT_KWARGS.items():
        if not key in kwargs:
            kwargs[key] = value

    button = pn.widgets.Button(**kwargs)

    copy_code = f"navigator.clipboard.writeText(`{text}`);"
    button.js_on_click(code=copy_code)

    return button
```

## Additional Context

- Blog Post showing how to add Github like copy button: https://junges.dev/2-how-to-add-github-copy-to-clipboard-button-on-your-docsblog

MarcSkovMadsen:
Solved by #7451

PR #7451, opened 2024-10-30
Add copy button to Markdown codeblocks

philippjfr:
Implements https://github.com/holoviz/panel/issues/6209

copy_button

### DIFF
--- a/panel/styles/models/html.less
+++ b/panel/styles/models/html.less
@@ -1,8 +1,8 @@
 .copybtn {
-  position: absolute;
+  position: sticky;
   display: flex;
-  top: .3em;
-  right: .3em;
+  top: 0;
+  left: 100%;
   width: 1.7em;
   height: 1.7em;
   opacity: 0;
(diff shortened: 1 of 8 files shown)


MarcSkovMadsen:
Would be so great for chat and many other use cases.

ahuang11:
One feature that I really like about ChatGPT is how the copy icon follows the scroll. Is it easy to implement that feature?

https://github.com/user-attachments/assets/86d6096a-2572-447f-b567-a5409a5d292b

panel/styles/models/html.less
```diff
@@ -0,0 +1,34 @@
+.copybtn {
+  position: absolute;
```
MarcSkovMadsen, line 2: I would expect the copy button to have a color aligned with the (dark) theme. But it has not.

https://github.com/user-attachments/assets/f5de2143-3aa5-4e2a-b159-228d33c97529

```python
import panel as pn
from panel.chat import ChatInterface

pn.extension("perspective")

code = """
Some code

```python
a=1
b=a+1
print(a, b)
```
"""

pn.template.FastListTemplate(
    title="Copy Button",
    sidebar=[code],
    main=[code],
).servable()
```

Coderambling:
> One feature that I really like about ChatGPT is how the "Copy Code" follows the scroll position, up to the end of the code, so users don't have to scroll all the way up to copy. Is it easy to implement that feature?
> 
>  Screen.Recording.2024-10-30.at.9.24.55.AM.mov

If the "follow" functionality is tricky to implement, maybe simply put the copy button at the end of the code block, as well as the beginning? Would also be a little less busy than a "following" button.

panel/styles/models/html.less
MarcSkovMadsen: There could be inspiration here:

- https://docs.streamlit.io/develop/api-reference/text/st.code

MarcSkovMadsen: This is how chatgpt looks for me. Could also be inspiration.

https://github.com/user-attachments/assets/3df18770-5a9a-4300-89f5-03e02771c3b7

philippjfr:
Will skip the following behavior for now. Adding two buttons is definitely too busy.

PR #7454, opened 2024-11-01
Fix bug in is_viewable_param

MarcSkovMadsen:
I was trying to use `param.ClassSelector(class_=(list,tuple)` with the `GraphicWalker` react component and I got 

```python
TypeError: issubclass() arg 1 must be a class

panel/viewable.py:1215: TypeError
```

This PR should add a test for this and fix the issue.

### DIFF
--- a/panel/viewable.py
+++ b/panel/viewable.py
@@ -1193,34 +1193,42 @@ class ChildDict(param.Dict):
 
 
 
+def _is_viewable_class_selector(class_selector: param.ClassSelector) -> bool:
+    if not class_selector.class_:
+        return False
+    if isinstance(class_selector.class_, tuple):
+        return all(issubclass(cls, Viewable) for cls in class_selector.class_)
+    return issubclass(class_selector.class_, Viewable)
+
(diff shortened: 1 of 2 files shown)


panel/viewable.py
```diff
@@ -1193,34 +1193,42 @@ def __set__(self, obj, val):
...
+    bool
+        True if the parameter specifies a Viewable type, False otherwise.
     """
-    p = parameter
-    if (
-        isinstance(p, (Child, Children)) or
-        (isinstance(p, param.ClassSelector) and p.class_ and (
-            (isinstance(p.class_, tuple) and
-             all(issubclass(cls, Viewable) for cls in p.class_)) or
-            issubclass(p.class_, Viewable)
```
MarcSkovMadsen, line 1215: The problem was that line 1215 was evaluated when it should not be.

philippjfr: Yes, good catch! I was about to submit the same fix 😄

philippjfr (approved):

PR #7461, opened 2024-11-05
doc: fix lint install pixi command

maximlt:


hoxbro:
I don't think this is needed. Do you want it because it is more verbose?

maximlt:
Somehow it was failing when I tried to run it earlier, but it's now running fine, not sure what's going on so I'm going to close the issue.

```
❯ pixi run lint-install
lint-install: command not found

Available tasks:
        install
        test-docs
        test-example
        test-subprocess
        test-unit
```

PR #7462, opened 2024-11-05
Various improvements for writing ESM components

philippjfr:
This PR provides a number of fixes and improvements for ESM components that I encountered as I was building a library with many components. This includes:

- **Ensure bundle path is correctly resolved on superclasses**: This allows compiling a single bundle for an entire package of different ESM components
- **Refactor ESM child resolution**: Makes it easier to modify the creation of child models (e.g. to allow lazily creating the models)
- **Allow defining constants on ESM model**: Allows providing a `_constants` dict on a component to modify the behavior of a component without having to define (public) parameters
- **Allow providing URL to _stylesheets**: Useful for loading external CSS resources
- **Load @emotion/styled for Material UI components**: Required for rendering certain Material UI components
- **Allow initializing NamedListPanel title dict as parameter**: Required for implementing ESM components implementing the `NamedListPanel` interfaces.

### DIFF
--- a/panel/custom.py
+++ b/panel/custom.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 
 import asyncio
 import hashlib
+import importlib
 import inspect
 import os
 import pathlib
@@ -263,21 +264,34 @@ class ReactiveESM(ReactiveCustomBase, metaclass=ReactiveESMMetaclass):
                 bundle_path = mod_path / bundle
                 if bundle_path.is_file():
                     return bundle_path
-                return
-            else:
-                raise ValueError(
-                    'Could not resolve {cls.__name__}._bundle. Ensure '
-                    'you provide either a string with a relative or absolute '
-                    'path or a Path object to a .js file extension.'
-                )
+            raise ValueError(
+                'Could not resolve {cls.__name__}._bundle: {cls._bundle}. Ensure '
+                'you provide either a string with a relative or absolute '
+                'path or a Path object to a .js file extension.'
+            )
+
+        # Attempt resolving bundle for this component specifically
         path = mod_path / f'{cls.__name__}.bundle.js'
         if path.is_file():
             return path
+
+        # Attempt to resolve bundle in current module and parent modules
         module = cls.__module__
-        path = mod_path / f'{module}.bundle.js'
-        if path.is_file():
-            return path
-        elif module in sys.modules:
+        modules = module.split('.')
+        for i in reversed(range(len(modules))):
+            submodule = '.'.join(modules[:i+1])
+            try:
+                mod = importlib.import_module(submodule)
+            except (ModuleNotFoundError, ImportError):
+                continue
+            if not hasattr(mod, '__file__'):
+                continue
+            submodule_path = pathlib.Path(mod.__file__).parent
+            path = submodule_path / f'{submodule}.bundle.js'
+            if path.is_file():
+                return path
+
+        if module in sys.modules:
             module = os.path.basename(sys.modules[module].__file__).replace('.py', '')
             path = mod_path / f'{module}.bundle.js'
             return path if path.is_file() else None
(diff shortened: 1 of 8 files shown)


griffinmilsap:
@philippjfr 

ESM Components are AWESOME! I've been thrilled for this capability for a while now!!

While you're in here making ESM improvements, I noticed a few `panel compile` issues that we could address in this PR too:

1. `panel compile` appears to not be actually compatible with "dotted module names" as indicated in the [documentation](https://panel.holoviz.org/how_to/custom_components/esm/build.html).  I believe the issue comes down to this line in [compile.py](https://github.com/holoviz/panel/blob/c06f779ecf0bae756c43229ee036110253b3cf79/panel/command/compile.py#L54) where dotted modules are treated as file extensions by `os.path.splitext`. I don't really know a good way to separate filenames from dotted module names as the last submodule looks like an extension to `os.path.splitext`.  It may be that another flag is necessary to explicitly inform the command that we're compiling a dotted module name.  Easiest thing to do for now would be to adjust docs to inform users that `panel compile` only takes filenames for now.

2. Copying dependencies into the temporary build directory can result in files that have a `<file>..js` filename instead of `<file>.js` which breaks the build.  I believe the issue is in [these lines of `panel/io/compile.py`](https://github.com/holoviz/panel/blob/c06f779ecf0bae756c43229ee036110253b3cf79/panel/io/compile.py#L299)

__Comments added pointing out issues__
``` python
    for component in components:
        name = component.__name__
        esm_path = component._esm_path(compiled=False)
        if esm_path:
            ext = esm_path.suffix # ISSUE: this returns `.js` for a standard esm js file
        else:
            ext = 'jsx' if issubclass(component, ReactComponent) else 'js' # ISSUE: this seems to assume ext won't have a dot in the suffix
        code, component_deps = extract_dependencies(component)
        # Detect default export in component code and handle import accordingly
        if _EXPORT_DEFAULT_RE.search(code):
            index += f'import {name} from "./{name}"\n'
        else:
            index += f'import * as {name} from "./{name}"\n'

        with open(path / f'{name}.{ext}', 'w') as component_file: # ISSUE: when ext is '.js', the output file is `<name>..js` which breaks build
            component_file.write(code)
```

I think the fix is just to adjust as such:

``` python
    for component in components:
        name = component.__name__
        esm_path = component._esm_path(compiled=False)
        if esm_path:
            ext = esm_path.suffix
        else:
            ext = '.jsx' if issubclass(component, ReactComponent) else '.js'
        code, component_deps = extract_dependencies(component)
        # Detect default export in component code and handle import accordingly
        if _EXPORT_DEFAULT_RE.search(code):
            index += f'import {name} from "./{name}"\n'
        else:
            index += f'import * as {name} from "./{name}"\n'

        with open(path / f'{name}{ext}', 'w') as component_file:
            component_file.write(code)
```

philippjfr:
Thanks, @griffinmilsap. I believe 2. was already fixed in https://github.com/holoviz/panel/pull/7446 and I'm about to push up fixes for 1.

panel/io/compile.py
```diff
@@ -71,6 +73,65 @@ def check_cli_tool(tool_name):
...
+    Arguments
+    ---------
+    module_spec: str
+         Module specification either as a dotted module or a path to a module.
+
+    Returns
+    -------
+    Dictionary containing the bundle paths and list of components to bundle.
+    """
+    if re.match(r'^[A-Za-z]:\\', module_spec):
```
hoxbro, line 89:
```suggestion
    if module_spec[1:3] == r":\\":
```

Can't something like this not be used? Please double-check.

panel/custom.py
```diff
@@ -235,37 +238,62 @@ def __init__(self, **params):
...
                     return bundle_path
-                return
-            else:
-                raise ValueError(
-                    'Could not resolve {cls.__name__}._bundle. Ensure '
-                    'you provide either a string with a relative or absolute '
-                    'path or a Path object to a .js file extension.'
-                )
+            raise ValueError(
+                'Could not resolve {cls.__name__}._bundle: {cls._bundle}. Ensure '
```
philippjfr, line 270:
```suggestion
                f'Could not resolve {cls.__name__}._bundle: {cls._bundle}. Ensure '
```

PR #7463, opened 2024-11-05
fix: Log authorization callback errors

hoxbro:
So a warning is emitted if something goes wrong in the authentication callback.

Also, set the status code if no auth fails to 403.

### DIFF
--- a/panel/io/server.py
+++ b/panel/io/server.py
@@ -426,6 +426,7 @@ class DocHandler(LoginUrlMixin, BkDocHandler):
                 auth_error = None
         except Exception:
             auth_error = f'Authorization callback errored. Could not validate user {state.user}.'
+            logger.warning(auth_error)
         return authorized, auth_error
 
     def _render_auth_error(self, auth_error):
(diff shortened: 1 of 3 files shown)


panel/io/server.py
```diff
@@ -426,6 +426,7 @@ def _authorize(self, session=False):
                 auth_error = None
         except Exception:
             auth_error = f'Authorization callback errored. Could not validate user {state.user}.'
+            logger.warning(auth_error)
```
hoxbro, line 429: I didn't want to print the exception as it could have sensitive information.

panel/io/server.py
```diff
@@ -455,6 +456,7 @@ async def get(self, *args, **kwargs):
             if authorized is None:
                 return
             elif not authorized:
+                self.set_status(403)
```
hoxbro, line 459: I don't know if we should set it to 200 outside this if statement.

philippjfr: Don't think so, that should be handled by the response.

philippjfr (approved):

PR #7474, opened 2024-11-07
Improve and document hold utility

philippjfr:
The `hold` utility is likely criminally underused by users simply because we haven't made it very visible. In this PR I did a few things:

- Improve the way events are batched and dispatched by `hold`
- Add how-to guide on using `hold`
- Move it from `panel.io.model` to `panel.io.document`

### DIFF
--- a/panel/io/model.py
+++ b/panel/io/model.py
@@ -22,6 +22,7 @@ from bokeh.model import DataModel
 from bokeh.models import ColumnDataSource, FlexBox, Model
 from bokeh.protocol.messages.patch_doc import patch_doc
 
+from ..util.warnings import deprecated
 from .state import state
 
 if TYPE_CHECKING:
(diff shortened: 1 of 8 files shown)


panel/io/model.py
```diff
@@ -192,22 +193,6 @@ def hold(doc: Document, policy: HoldPolicyType = 'combine', comm: Comm | None =
...
-        yield
-    finally:
-        if held:
-            doc.callbacks._hold = held
-        else:
-            if comm is not None:
-                from .notebook import push
-                push(doc, comm)
-            doc.unhold()
+    deprecated('1.7.0', 'panel.io.model.hold', 'panel.io.document.hold')
```
hoxbro, line 196: Currently, this is set up always to warn. Maybe we should change that as I feel this deprecation will be pretty verbose.

philippjfr: I'd hope most people imported it as `from panel.io import hold`.

philippjfr: Postponed the warning until 1.6.0

panel/util/warnings.py
```diff
@@ -60,13 +60,20 @@ def deprecated(
     old: str,
     new: str | None = None,
     extra: str | None = None,
```
hoxbro, line 62:
```suggestion
    *
    extra: str | None = None,
```

So that we don't make a simple input error at some point.

Issue #7468, opened 2024-11-07
Support Polars as parameter of ReactiveESM component


MarcSkovMadsen:
I would like to add support for Polars DataFrame to `GraphicWalker.object`. Its unclear to me whether this is supported and how. I guess special care is taken about pandas dataframe serialization?

```python
import panel as pn
import param

from panel.custom import JSComponent
import polars as pl
import pandas as pd
pn.extension()

class GraphicWalker(JSComponent):

    object = param.ClassSelector(class_=(pd.DataFrame, pl.DataFrame))

    _esm = '''
    export function render({ model, el }) {
      console.log(model.object)
      el.innerHTML = `<h1>Hello World</h1>`
    }
    '''

GraphicWalker(object=pl.DataFrame({"x": [1,2,3]})).servable()
```

``` bash
TypeError: the truth value of a DataFrame is ambiguous

Hint: to check if a DataFrame contains any values, use is_empty().

Traceback (most recent call last):
  File "/home/jovyan/repos/private/panel-graphic-walker/.venv/lib/python3.11/site-packages/panel/io/handlers.py", line 405, in run
    exec(self._code, module.__dict__)
  File "/home/jovyan/repos/private/panel-graphic-walker/script.py", line 20, in <module>
    GraphicWalker(object=pl.DataFrame({"x": [1,2,3]})).servable()
  File "/home/jovyan/repos/private/panel-graphic-walker/.venv/lib/python3.11/site-packages/panel/viewable.py", line 399, in servable
    self.server_doc(title=title, location=location) # type: ignore
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/jovyan/repos/private/panel-graphic-walker/.venv/lib/python3.11/site-packages/panel/viewable.py", line 1006, in server_doc
    model = self.get_root(doc)
            ^^^^^^^^^^^^^^^^^^
  File "/home/jovyan/repos/private/panel-graphic-walker/.venv/lib/python3.11/site-packages/panel/viewable.py", line 678, in get_root
    root = self._get_model(doc, comm=comm)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/jovyan/repos/private/panel-graphic-walker/.venv/lib/python3.11/site-packages/panel/custom.py", line 436, in _get_model
    model = self._bokeh_model(**self._get_properties(doc))
[...]

PR #7475, opened 2024-11-08
Add support for serializing polars DataFrame

philippjfr:
Partial fix for https://github.com/holoviz/panel/issues/7468

We additionally need to allow `param.DataFrame` to accept non-pandas DataFrames, either by default or as an option.


panel/io/datamodel.py
```diff
@@ -87,6 +109,13 @@ def bytes_param(p, kwargs):
     kwargs['default'] = None
     return bp.Nullable(bp.Bytes, **kwargs)
 
+def df_to_dict(df):
+    if 'polars' in sys.modules:
+        import polars as pl
+        if isinstance(df, (pl.LazyFrame, pl.DataFrame)):
```
hoxbro, line 115: Don't `LazyFrame` need a collect?

panel/io/datamodel.py
```diff
@@ -87,6 +109,13 @@ def bytes_param(p, kwargs):
...
+            df = df.to_pandas()
```
philippjfr, line 116:
```suggestion
        if isinstance(df, pl.LazyFrame):
            df = df.collect()
        if isinstance(df, pl.DataFrame):
            df = df.to_pandas()
```

MarcSkovMadsen:
> Partial fix for #7468
> 
> We additionally need to allow `param.DataFrame` to accept non-pandas DataFrames, either by default or as an option.

My vision for the `DataFrame` parameter is that there is a parameter that supports any kind of tabular data source including the ones supported by Narwhals. See https://github.com/narwhals-dev/narwhals/issues/1289.

philippjfr:
pre-commit.ci autofix

panel/io/datamodel.py
```diff
@@ -1,6 +1,10 @@
 from __future__ import annotations
 
+import sys
+import weakref
```
philippjfr, line 4:
```suggestion
```

philippjfr:
This change won't do anything yet but it makes sense to merge in anticipation to the coming changes in param.

PR #7512, opened 2024-11-22
Fix mypy issue

MarcSkovMadsen:
I'm currently updating and rebuilding a lot of projects. One issue I see across is that type of `Tabulator.selection` is not understood by mypy

```bash
error: Incompatible types in assignment (expression has type "list[int]", variable has type "_ListValidateWithCallable")  [assignment]
```

This PR fixes the issue.


hoxbro:
The change looks benign. I don't understand why it is needed here and not for other parameters. 

Would you happen to know why?

hoxbro (approved):

hoxbro:
> Would you happen to know why?

I think it is related to the param not being typed and, therefore, ignored. There is no reason not to merge this.

PR #7523, opened 2024-11-26
compat: altair 5.5.0

hoxbro:
In my opinion, it isn't worth keeping a compatibility layer with try/except for docstring, so I updated it to the newest API.


MarcSkovMadsen:
There is also a couple of places in the docs to fix:

image

Issue #7532, opened 2024-12-04
Ability to set disabled for NestedSelect widget


mayonnaisecolouredbenz7:
#### ALL software version info
<details>
    <summary>Software Version Info</summary>

```plaintext
panel==1.5.4
```
</details>

#### Description of expected behavior and the observed behavior
**Disabled** argument to work with NestedSelect

#### Complete, minimal, self-contained example code that reproduces the issue

```python
select = pn.widgets.NestedSelect(
    options={
        "GFS": {
            "0.25 deg": ["00Z", "06Z", "12Z", "18Z"],
            "0.5 deg": ["00Z", "12Z"],
            "1 deg": ["00Z", "12Z"],
        },
        "NAME": {
            "12 km": ["00Z", "12Z"],
            "3 km": ["00Z", "12Z"],
        },
    },
    levels=[{'name':"Model","type": pn.widgets.RadioBoxGroup}, "Resolution", "Initialization"],
    disabled = True,
)

```

- [x] I have made a PR for this

PR #7533, opened 2024-12-04
Fix issue #7532: Disabled for NestedSelect

mayonnaisecolouredbenz7:
### Description

This pull request fixes issue #7532 .

### Changes Made

- Fixed the `disabled` toggle for `NestedSelect` 
- Not sure if this is it what is required.  But is working as expected
- Documentation doesn't show disabled as a parameter, however the original code did have references to it so looks like it is intended and is inline with other widgets
- Ran tests locally with pixi before opening this PR

### DIFF
--- a/panel/widgets/select.py
+++ b/panel/widgets/select.py
@@ -545,7 +545,7 @@ class NestedSelect(CompositeWidget):
         value = self._lookup_value(i, options, self.value, error=False)
         widget_kwargs["options"] = options
         widget_kwargs["value"] = value
-        widget_kwargs["disabled"] = self.disabled
+        widget_kwargs["disabled"] = self.param.disabled
         if "visible" not in widget_kwargs:
             # first select widget always visible
             widget_kwargs["visible"] = i == 0 or callable(options) or len(options) > 0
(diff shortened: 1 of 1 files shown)


ahuang11:
Thanks for adding this! Can you add it to the reference notebook https://github.com/holoviz/panel/blob/main/examples/reference/widgets/NestedSelect.ipynb and maybe add a test too https://github.com/holoviz/panel/blob/main/panel/tests/widgets/test_select.py#L260 by checking the sub-widgets are disabled too.

Also, maybe you sync loading too.

panel/widgets/select.py
```diff
@@ -545,6 +545,7 @@ def _init_widget(self, i, options):
         value = self._lookup_value(i, options, self.value, error=False)
         widget_kwargs["options"] = options
         widget_kwargs["value"] = value
+        widget_kwargs["disabled"] = self.disabled
```
philippjfr, line 548:
```suggestion
        widget_kwargs["disabled"] = self.paramdisabled
```

philippjfr: This makes the two dynamically linked.

Issue #7505, opened 2024-11-19
Tabulator: wrong selection displayed on the front-end with `add_filter`


maximlt:
The wrong Tabulator row is selected in the front-end when the data is filtered with `add_filter`. The front-end seems to use the index of the selected row from the whole DataFrame (I'm selecting the row 2 in the whole df) which is incorrect when the data is filtered. It doesn't seem to affect `header_filter`.

```python
import panel as pn
import pandas as pd

df = pd.DataFrame({"idx": ["idx0", "idx1", "idx2", "idx3", "idx4"], "col": ["a", "aa", "b", "bb", "bbb"]})
w_pat = pn.widgets.Select(value="b", options=["a", "b"])
w = pn.widgets.Tabulator(df, selectable=1)

def f(df, pattern):
    return df[df['col'].str.contains(pattern)]

w.add_filter(pn.bind(f, pattern=w_pat))
pn.Column(w_pat, w).servable()
```

![Image](https://github.com/user-attachments/assets/ec015691-5f51-4a8d-b303-40d28deadec6)

contang0:
There's another add_filter bug when you use selectable="checkbox".  

w = pn.widgets.Tabulator(df, selectable="checkbox")

1. It doesn't actually let you check more than 1 checkbox
2. You have to click it 2-3 times for the checkbox to stay checked

contang0:
Update: seems like enabling header_filters=True fixes the above issue, but it's still a bug

PR #7534, opened 2024-12-04
Tabulator: fix selected row in the UI with add_filter

maximlt:
Just a test for now, it will address https://github.com/holoviz/panel/issues/7505 :)


philippjfr:
Should be fixed by https://github.com/holoviz/panel/pull/7712

Issue #7504, opened 2024-11-19
Panel Notifications now show the last notification multiple times for each displayed notification, instead of provided notifications


jnareb:
#### ALL software version info

<details>
    <summary>Software Version Info</summary>

```plaintext
panel==1.5.4
Python 3.10.4
Windows 10 Home 22H2 (64-bit)

bokeh==3.6.0
Chrome 130.0.6723.117 (64-bit)
```
</details>

#### Description of expected behavior and the observed behavior

When sending multiple notifications within the duration time of previous notification, I expect to see all notifications, one below another, vanishing one after the other.

What I get with Panel 1.5.4 is seeing multiple copies of the last notification (type and the message).

This is a **regression**, because I remember not having this problem before updating Panel. I thing that Panel 1.5.3 did not have this issue.

#### Complete, minimal, self-contained example code that reproduces the issue

```python
import panel as pn

pn.extension(
    notifications=True,
)
pn.state.notifications.position = 'top-center'


def onload_callback() -> None:
    pn.state.notifications.warning("Warning", duration=0)
    pn.state.notifications.info("Info", duration=0)


pn.state.onload(onload_callback)
pn.pane.Markdown("# Hello world").servable()
```

#### Screenshots or screencasts of the bug in action

![Screenshot of Panel app, with two notifications reading "Info"](https://github.com/user-attachments/assets/f05259e4-ae3f-4469-943b-790afea047a8)

I would expect to see warning notification with the "Warning" message above info notification with "Info" message, instead of seeing info notification with "Info" message twice.

- [x] I may be interested in making a pull request to address this

hoxbro:
It looks like it happens somewhere between 1.3.8 and 1.4.0.

Issue #4964, opened 2023-05-29
Notifications not properly cleaned up when more than 1 exists


tbronson:
Whenever more than 1 Notification is issued, only the last issued one will be properly cleaned and removed from the notification list along with its associated watcher in the _notification_watchers dict, due to timeout or user interaction closing it.

Calling clear() does not cleanup the watchers.

tbronson:
Debugging on the javascript side, it seems that the syncing of the notifications array is causing a copy operation on the JS-side, which I'm guessing is creating new Notification objects for any already existing in the array (you can see the all the ids change each time you issue a new notification), while all the watchers and timeout callbacks are tied to the old orphaned Notifications.

I'm not savvy enough yet with Bokeh/Panel internals or coding in javascript to know the best way to fix this issue.

PR #7560, opened 2024-12-18
Ensure data models are reused ensuring that notification state persists

philippjfr:
Previously data models would be re-created whenever the property changed which was pointless and meant that any state set on them may not persist.

Fixes https://github.com/holoviz/panel/issues/7504
Fixes https://github.com/holoviz/panel/issues/4964

### DIFF
--- a/panel/io/notifications.py
+++ b/panel/io/notifications.py
@@ -196,7 +196,6 @@ class NotificationArea(NotificationAreaBase):
         })
       """,
       "notifications": """
-        let to_render = []
         for (notification of data.notifications) {
           if (notification._destroyed || notification._rendered) {
             return
--- a/panel/io/notifications.py
+++ b/panel/io/notifications.py
@@ -34,6 +34,8 @@ class Notification(param.Parameterized):
 
     notification_type = param.String(default=None, constant=True, label='type')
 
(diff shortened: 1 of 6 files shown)


panel/io/notifications.py
```diff
@@ -194,46 +196,48 @@ def __css__(cls):
...
+            config.background = notification.background;
+          }
+          if (notification.icon != null) {
+            config.icon = notification.icon;
+          }
+          let toast = state.toaster.open(config);
+          function destroy() {
+            if (state.current !== notification) {
+              notification._destroyed = true;
+            }
```
philippjfr, line 219:
```suggestion
            notification._destroyed = true;
```

panel/io/notifications.py
```diff
@@ -194,46 +196,46 @@ def __css__(cls):
...
-        if (notification.background != null) {
-          config.background = notification.background;
-        }
-        if (notification.icon != null) {
-          config.icon = notification.icon;
-        }
-        var toast = state.toaster.open(config);
-        function destroy() {
-          if (state.current !== notification) {
+        let to_render = []
```
philippjfr, line 199:
```suggestion
```

PR #7568, opened 2024-12-23
Add form tutorial

MarcSkovMadsen:
Other frameworks like Streamlit and Reflex have components and tutorials for creating forms.

In Gradio the input is almost always a form.

We don't have anything explaining how to create a Form with Panel. This PR addresses this issue.


doc/tutorials/intermediate/index.md
```diff
@@ -67,6 +67,7 @@ Embark on a deeper exploration of supplementary topics to further hone your Pane
 Now that you've mastered the more advanced concepts of Panel, it's time to put your skills to the test:
 
 - **Create an Interactive Report:** Elevate the interactivity of your reports through embedding.
+- **[Create a Form](create_form.md):** Create a form to sign up for a news letter.
```
MarcSkovMadsen, line 70: With Reflex a Form is the first thing they teach/ show their users. But they don't have to manually add validation. Its built into the components.

doc/tutorials/intermediate/create_form.md
```diff
@@ -0,0 +1,261 @@
...
+```
+
+### 3. Create a Parameterized Form State Class
+
+The `FormState` class handles form inputs, validation, and state management.
+
+```python
+class FormState(param.Parameterized):
+    name = param.String(default="", doc="The name of the user.")
+    email = param.String(default="", doc="The email of the user.")
```
MarcSkovMadsen, line 56: It would be nice if we could add a regex for an `email`. I tried but the exceptions where raised in the terminal. It was not clear to me how to easily integrate them with the components and UI.

It would also be nice to be able to add a custom `validate` function to any `param.Parameter`. This seems not supported either. Its a standard thing for all other Model frameworks I know.

doc/tutorials/intermediate/create_form.md
```diff
@@ -0,0 +1,261 @@
...
+```python
+def create_form():
+    form_state = FormState()
+
+    header = pn.Row(
+        pn.pane.SVG(FORM_ICON, margin=0, height=80, sizing_mode="fixed"),
+        FORM_TEXT,
+    )
+
+    error_pane = pn.pane.Alert(
```
MarcSkovMadsen, line 106: It would be really nice to be able to display the errors together with the widgets. For example via red borders and tooltips. The red borders can be done with a lot of work. In theory tooltips could be added. But they will be in the tooltips icon instead of over the `input` element.

doc/tutorials/intermediate/create_form.md
```diff
@@ -0,0 +1,261 @@
...
+        object=form_state.param.validation_message,
+        visible=form_state.param.is_not_valid,
+        alert_type="danger",
+        stylesheets=["p {margin-bottom: 0}"]
+    )
+
+    name_input = pn.widgets.TextInput.from_param(form_state.param.name, name="Name*", placeholder="User Name")
```
MarcSkovMadsen, line 239: In the Reflex example a separate red `*` is manually added for *required* elements. We could also do that. I would prefer if that was integrated in the param.Parameters and aligned with the widgets.

doc/tutorials/intermediate/create_form.md
```diff
@@ -0,0 +1,261 @@
...
+    header = pn.Row(
+        pn.pane.SVG(FORM_ICON, margin=0, height=80, sizing_mode="fixed"),
+        FORM_TEXT,
+    )
+
+    error_pane = pn.pane.Alert(
+        object=form_state.param.validation_message,
+        visible=form_state.param.is_not_valid,
+        alert_type="danger",
+        stylesheets=["p {margin-bottom: 0}"]
```
MarcSkovMadsen, line 110: I don't know if its a general problem that `margin-bottom` of the Alert paragraph is set to `1rem`. Maybe it should be set to 0 in general?

doc/tutorials/intermediate/create_form.md
ahuang11: I think regex works as long as you don't use `name` as your param name because the name property overrides the param.

image


```
import param
class FormState(param.Parameterized):
    n = param.String(default="", doc="The name of the user.", regex=r"^[A-Za-z ]*$")


form_state = FormState(n="ABC")
form_state.name
```

MarcSkovMadsen: Thanks. The problem is not the `regex` not working. The problem is that the `regex` errors are not propogated to the UI. It stays on the server side.

Coderambling (commented):
Great tutorial! Would it make sense to add 2 examples of conditional logic?

So 1 case where an input field is shown / hidden based on, for example, a yes / no input value in a previous field?

And one where the options in a field are determined by the input in a previous field?

I.e. Field 1 -> Choose continent, field 2 -> Show subselection of countries that belong to that continent.

As both are very common use cases for forms, and param seems a natural fit for the type of logic.

In a broader context, would a panel form (builder) component make sense?

Is there (a lot of) latent demand for this?

doc/tutorials/intermediate/create_form.md
philippjfr: +1, let's revive your required parameter PR in param soon.

doc/tutorials/intermediate/index.md
philippjfr: Makes sense.

philippjfr:
My feeling is that this is something we definitely need to work on. This tutorial sets a baseline, warts and all, and it'll be our job to improve on the status quo from here. I'll merge as is for now.

@Coderambling Yes, I think those are all good points and we should add those later.

philippjfr (approved):

Issue #7535, opened 2024-12-04
Markdown pane does not collapse line breaks by default


maximlt:
By default, the Markdown Pane does not collapse line breaks. This is when you write on multiple lines:

```
My name
is Bond,
James Bond
```

And it is displayed on a single line:

```
My name is Bond, James Bond
```

This is the default behavior on VSCode or in JupyterLab.

![Image](https://github.com/user-attachments/assets/1e8dd841-83d3-4ba0-97c9-fcd17a9f54b5)

It doesn't seem to be the default everywhere, at least not on Github.

![Image](https://github.com/user-attachments/assets/b0c5596b-b714-461a-bb2d-ba5be9bd2140)

It looks like this used to be the default behavior in Panel but was changed in https://github.com/holoviz/panel/pull/5376/. I'm opening this issue as I was updating the Exoplanets example which has this multiline string displayed in a Markdown pane. The string is written in a way to make it easy to read in a notebook, this would also be valid for code written in a `.py` file. Unfortunately, the line breaks are all preserved and the end result doesn't look great. Removing these line breaks while keeping those desired isn't trivial, it requires either refactoring the multiline string or processing it (e.g. `txt.replace('\n\n', 'XXXXX').replace('\n', ' ').replace('XXXXX', '\n\n')`.

```python
background_text = """
For the past 25+ years, NASA has used ground- and space-based methods
to [identify exoplanets](https://exoplanets.nasa.gov/exep/about/missions-instruments) (planets outside of our solar system).

In the past ten years in particular, campaigns like Kepler, K2, and TESS have
produced an explosion of results.
To date, approximately 4,400 exoplanets have been identified, and over 3,000 potential exoplanet candidates have 
been discovered.

This dashboard uses [Holoviews](https://holoviews.org/) and [Panel](https://panel.holoviz.org) 
to visualize the discovery of confirmed and candidate exoplanets over the years.
Also included is a scatterplot that reveals details about the relationship among mass, radius, and temperature of exoplanets, 
as well as controls to filter the data based on whether the planets could support life, and if so, 
whether chemical rockets could be used to escape the planet.

See [examples.holoviz.org](https://examples.holoviz.org/exoplanets) for details on the data used here and how to interpret it.
"""
```

![Image](https://github.com/user-attachments/assets/79ff5be1-c835-4f85-a75c-239dd518ea92)

[...]

PR #7582, opened 2025-01-02
Markdown pane does not render new lines as hard line breaks, configurable with hard_line_break

maximlt:
Sort of addresses https://github.com/holoviz/panel/issues/7535

This PR adds a new `break_as_new` parameter (better name welcome!) to the `Markdown` pane to make it easier for users to declare whether they want simple line breaks to be rendered with a new line or not. Unfortunately, I haven't been able to find a simple way to configure MyST-Parser to toggle this behavior.

Importantly, note that I am -1 on the current default in Panel which is `break_as_new=True` with `markdown-it`. Even if I understand the motivation to make it easier to render Markdown generated by LLMs, it makes it way more difficult to display in a nice way some Markdown formatted text (`pn.Column(md_intro, <objs>)` vs `pn.Column(pn.pane.Markdown(md_intro, break_as_new=False), <objs>)`. @ahuang11 prefers the current default, @philippjfr I'll let you chime in with your opinion (and let's wait for Andrew to be back too).

Once this is settled for good, I'll add tests to this PR.


ahuang11:
I'm okay with having `break_as_new=False`. The only thing is setting it to `True` in `ChatMessage`.

Also, I don't think `break_as_new` is intuitive, maybe `breaks_as_newline`?

MarcSkovMadsen:
I think the name `line_breaks` is easier to remember and understand.

I also think this is actually a regression. Originally coming from Streamlit this was a major issue when moving to Panel that new lines in multiline `"""` strings got broken into new lines. I do believe it was fixed in Panel but at some stage it reverted to breaking lines which is quite annoying.

maximlt:
Ok so I renamed it to `hard_line_break` that defaults to `False`, re-establishing the older behavior.

> The only thing is setting it to `True` in `ChatMessage`.

@ahuang11 in https://github.com/holoviz/panel/pull/7582/commits/728b1fa2ee2ec97fc91945b19bc17aa420fc6d24 I've made a change to enable it in `ChatMessage`. But I wanted to ask you, do you know if all strings passed to `ChatMessage` are turned into `Markdown` panes? If yes, it could skip calling `panel(...)` and directly call `Markdown(..., hard_line_break)` for values of str type, probably making things a bit faster.

ahuang11:
> But I wanted to ask you, do you know if all strings passed to ChatMessage are turned into Markdown panes

If it's a string, it should be a Markdown, but only if isinstance(message, str).

maximlt:
> > But I wanted to ask you, do you know if all strings passed to ChatMessage are turned into Markdown panes
> 
> If it's a string, it should be a Markdown, but only if isinstance(message, str).

I shortly discussed this with Philipp this morning, who told me other components can be instantiated from a string (image pointing to a URL), and that one could see ChatMessage as a sort of opinionated repr for `panel()`. So always instantiating a Markdown pane for a string isn't appropriate in this case.

philippjfr (approved):

PR #7585, opened 2025-01-04
Update functions_vs_classes.md text so it's not empty any more and provides useful information and links

Coderambling:
Gives users a starting point, instead of no content at all.

Added some basic content. Further additions welcome.

Also, should there be / is there a table that shows how the Function vs. clasess based approach relates to the Declarative / Reactive / Callback-based API's?


philippjfr:
Superseded by https://github.com/holoviz/panel/pull/8581

Issue #7586, opened 2025-01-05
AttributeError: 'NoneType' object has no attribute 'get'


MarcSkovMadsen:
panel==1.5.5

I'm trying to configure azure oauth. Its hard for me to understand how to configure the `azure`  side.

But I do currently see a bug in Panel

```bash
Traceback (most recent call last):
  File "/home/jovyan/repos/mt-ai-assistant/.venv/lib/python3.11/site-packages/panel/auth.py", line 225, in _fetch_access_token
    response = await http.fetch(req)
               ^^^^^^^^^^^^^^^^^^^^^
tornado.httpclient.HTTPClientError: HTTP 401: Unauthorized

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/home/jovyan/repos/mt-ai-assistant/.venv/lib/python3.11/site-packages/tornado/web.py", line 1790, in _execute
    result = await result
             ^^^^^^^^^^^^
  File "/home/jovyan/repos/mt-ai-assistant/.venv/lib/python3.11/site-packages/panel/auth.py", line 391, in get
    user = await self.get_authenticated_user(**params)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/jovyan/repos/mt-ai-assistant/.venv/lib/python3.11/site-packages/panel/auth.py", line 141, in get_authenticated_user
    user, _, _, _ = await self._fetch_access_token(
                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/jovyan/repos/mt-ai-assistant/.venv/lib/python3.11/site-packages/panel/auth.py", line 228, in _fetch_access_token
    self._raise_error(e.response, status=401)
  File "/home/jovyan/repos/mt-ai-assistant/.venv/lib/python3.11/site-packages/panel/auth.py", line 455, in _raise_error
    body.get('error_description', str(body)),
    ^^^^^^^^
AttributeError: 'NoneType' object has no attribute 'get'
```

I do believe the Panel code does not handle the 401 error and `body==None` well:

![Image](https://github.com/user-attachments/assets/1043a7ee-6747-4dab-8712-1def73087b76)

philippjfr:
Fixed in #7587

PR #7587, opened 2025-01-05
Fix azure oauth issues

MarcSkovMadsen:
Closing #7586 and more issues identified while trying to get azure oauth working.

### DIFF
--- a/panel/auth.py
+++ b/panel/auth.py
@@ -54,6 +54,7 @@ def decode_response_body(response):
         body = codecs.decode(response.body, 'ascii')
     except Exception:
         body = codecs.decode(response.body, 'utf-8')
+    body = re.sub("\'", '\\"', body)
     body = re.sub('"', '\"', body)
     body = re.sub("'", '"', body)
     body = json.loads(body)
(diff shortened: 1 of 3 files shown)


panel/auth.py
```diff
@@ -54,6 +54,7 @@ def decode_response_body(response):
         body = codecs.decode(response.body, 'ascii')
     except Exception:
         body = codecs.decode(response.body, 'utf-8')
+    body = re.sub("\'", '\\"', body)
```
MarcSkovMadsen, line 57: This is a workaround as just below you replace both " and '. It works. But its hard to do right as there are no existing test cases.

panel/auth.py
```diff
@@ -450,10 +451,17 @@ def _raise_error(self, response, body=None, status=400):
             log.warning(f"{provider} OAuth provider failed to fully "
                         f"authenticate returning the following response:"
                         f"{body}.")
+        if hasattr(body, "get"):
```
MarcSkovMadsen, line 454: In my case `body` was `None` due to the decoding error triggering an `AttributeError`.

panel/command/__init__.py
```diff
@@ -91,7 +91,7 @@ def main(args: list[str] | None = None):
             try:
                 ret = parsed_args.invoke(parsed_args)
             except Exception as e:
-                if config.dev:
+                if hasattr(config, "dev") and config.dev:
```
MarcSkovMadsen, line 94: I got an error here too. I have seen this before that `config` has not attribute `dev `.

philippjfr:
```suggestion
                if config.autoreload:
```

PR #7589, opened 2025-01-05
fix: Use f-string to show filename in error message correctly

kdheepak:
This fix makes the string an f-string. Without this change, the error message doesn't show which file is actually missing:

image

### DIFF
--- a/panel/command/convert.py
+++ b/panel/command/convert.py
@@ -114,7 +114,7 @@ class Convert(Subcommand):
         for f in args.files:
             p = pathlib.Path(f).absolute()
             if not p.is_file():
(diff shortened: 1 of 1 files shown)


kdheepak:
I also think this should be a `FileNotFoundError` exception instead. Let me know if you'd like me to submit a separate PR for that.

philippjfr:
>I also think this should be a FileNotFoundError exception instead. Let me know if you'd like me to submit a separate PR for that.

Yes, I think so. Also investigating what's up with the test suite, must have regressed somehow over the holidays.

philippjfr:
Thank you!

philippjfr (approved):

PR #7608, opened 2025-01-09
Scroll when streaming to ChatFeed and ChatStep

philippjfr:
It's a nicer experience to follow along while it's streaming though we should see if there's a mechanism to disable this behavior.

### DIFF
--- a/panel/chat/feed.py
+++ b/panel/chat/feed.py
@@ -704,7 +704,7 @@ class ChatFeed(ListPanel):
 
             if message_params:
                 message.param.update(**message_params)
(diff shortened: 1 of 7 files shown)


ahuang11:
Okay now it works like auto_scroll_limit:
1. if scroll_limit = 0, then scroll_to_latest is disabled
2. if scroll_limit (in TS) is null, then scroll_to_latest will trigger
3. if scroll limit is set to a number, it'll only scroll within the limit

https://github.com/user-attachments/assets/a4b23634-875d-417b-9bac-4a594e96b9d9

ahuang11 (approved):

PR #7610, opened 2025-01-10
Add callback for onChangeText to jsoneditor.ts

### LINKED ISSUE #6583: JSONEditor with mode="text" fails to update the value parameter when changes are made in the browser
Panel 1.3.8
Bokeh 3.3.4
Python 3.11.6
Ubuntu 22.04
Chrome 123.0.6312.58

I am trying to use the *panel.widgets.JSONEditor* with *mode="text"*. I expect the *value* parameter of the JSONEditor object to be updated when the content of the editor is changed in the browser.

It works as expected with *mode="tree"*

```
import panel as pn


async def update_text(event):
    print(event.new)


json_editor = pn.widgets.JSONEditor(value={}, mode="text")
json_editor.param.watch(update_text, "value")

pn.serve(pn.Column(json_editor))
```

I believe the error is related to the following warning which is issued in the browser:

```
jsoneditor.min.js:30 Option "onChangeJSON" is not applicable to modes "text" and "code". Use "onChangeText" or "onChange" instead.
C @ jsoneditor.min.js:30
```

jonatantreijs:
Fixes #6583 

Adds a callback for onChangeText in the jsoneditor typescript that converts the text value to json and updates the model accordingly.

### DIFF
--- a/panel/models/jsoneditor.ts
+++ b/panel/models/jsoneditor.ts
@@ -69,6 +69,14 @@ export class JSONEditorView extends HTMLBoxView {
       onChangeJSON: (json: any) => {
         this.model.data = json
       },
+      onChangeText: (text: any) => {
+        try {
+          this.model.data = JSON.parse(text)
+        }
+        catch (e) {
+          console.warn(e)
+        }
+      },
       onSelectionChange: (start: any, end: any) => {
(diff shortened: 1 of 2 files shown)


philippjfr (approved):
Nice, and thanks for the test!

PR #7617, opened 2025-01-15
Add css_id parameter to Viewable components

philippjfr:
While we probably want to eventually expose this at the Bokeh level, adding an ID on the DOM nodes that correspond to Panel components is a useful way of referencing individual components for testing and when trying to look up a component. This PR implements this via `tags`, i.e. when an `css_id` is set this is added as an item in the `tags` property of the Model. On Document initialization on the frontend the `css_id` is then applied.

- [x] Add tests ensuring the id is added to all component types
- [x] Add tests ensuring the id is set on components added after initialization
- [ ] Add documentation

### DIFF
--- a/panel/io/document.py
+++ b/panel/io/document.py
@@ -333,13 +333,24 @@ def init_doc(doc: Document | None) -> Document:
     if thread_id:
         state._thread_id_[curdoc] = thread_id
 
+    ready_callbacks = [CustomJS(code="""
+        for (const v of cb_context.index.all_views()) {
+            for (const t of v.model.tags) {
+              if (typeof t === 'object' && t !== null && 'css_id' in t) {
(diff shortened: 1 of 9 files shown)


MarcSkovMadsen:
Or element_id?

Coderambling:
Is it the case that CSS framework refer to this as "id"? In that case css_element_id might be the most appropriate.

I wonder if element_id might be confusing, or if, in future, some_other_thing_element_id might be a useful name, in which case element_id could become confusing.

Am I correcting in assuming the css element (with a unique id) would be available to all components (but not necessarily applied to all components)?

Would it be useful if the documentation contained code that shows how to list what components are using the CSS element? And what CSS elements a particular component uses?

PR #7620, opened 2025-01-16
test: Stabilize flaky UI test

hoxbro:


panel/tests/ui/layout/test_feed.py
```diff
@@ -129,9 +135,12 @@ def test_feed_scroll_to_latest_within_limit(page):
 
     feed.scroll_to_latest(scroll_limit=100)
 
-    wait_until(lambda: feed_el.evaluate(
-        '(el) => el.scrollHeight - el.scrollTop - el.clientHeight'
-    ) == 0, page)
+    def case():
```
philippjfr, line 138: This seems exactly equivalent.

hoxbro: It is functionality-wise, but pytest does its magic, and you get the value of the assert output, instead of a TimeoutError. 

image

Adding the new feed was to try to slow things down, I don't think it is ideal.

philippjfr: Ah okay, I've tried and failed to make pytest understand wait_until properly. Would be nice to find a general solution.

hoxbro: I think this is the best solution, this will rewrite the function inside the test.

philippjfr (approved):

hoxbro:
The screenshot of the image not loading is:

test-failed-1

panel/tests/ui/test_custom.py
```diff
@@ -20,6 +20,13 @@
 pytestmark = pytest.mark.ui
 
 
+@pytest.fixture(scope="module", autouse=True)
+def set_expect_timeout():
+    expect.set_options(timeout=10_000)
+    yield
+    expect.set_options(timeout=5_000)
```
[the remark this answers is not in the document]
hoxbro: Have done this in https://github.com/holoviz/panel/pull/7620/commits/541cb1255232b10531d2f2c96a0b7c2f4b7d9e9b

panel/tests/ui/io/test_convert.py
```diff
@@ -27,7 +27,7 @@
 
 
 if os.name == "wt":
```
philippjfr, line 29: Isn't windows "nt"?

philippjfr:
```suggestion
if os.name == "nt":
```

Issue #7598, opened 2025-01-07
panel compile not working on windows


xavArtley:
panel version 1.5.1
python 3.12
windows 10

On windows 10 the command to check if "esbuild" is intalled failed

the following script:

```python

import subprocess as sbp


print(sbp.run(["esbuild", "--version"], capture_output=True, shell=True)) # OK
print(sbp.run(["esbuild.cmd", "--version"], capture_output=True)) # OK

print(sbp.run(["esbuild", "--version"], capture_output=True)) #Failed

```

produce
```
CompletedProcess(args=['esbuild', '--version'], returncode=0, stdout=b'0.24.2\n', stderr=b'')
CompletedProcess(args=['esbuild.cmd', '--version'], returncode=0, stdout=b'0.24.2\n', stderr=b'')
Traceback (most recent call last):
  File "c:\CIVA_RP\CIVA_ARQANE\src_python\pyciva\pyciva\confetti.py", line 24, in <module>
    print(sbp.run(["esbuild", "--version"], capture_output=True))
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\CIVA_RP\CIVA_ARQANE\ProjetCIVA\ceaComponents\dev\python\Lib\subprocess.py", line 548, in run
    with Popen(*popenargs, **kwargs) as process:
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\CIVA_RP\CIVA_ARQANE\ProjetCIVA\ceaComponents\dev\python\Lib\subprocess.py", line 1026, in __init__
    self._execute_child(args, executable, preexec_fn, close_fds,
  File "C:\CIVA_RP\CIVA_ARQANE\ProjetCIVA\ceaComponents\dev\python\Lib\subprocess.py", line 1538, in _execute_child
    hp, ht, pid, tid = _winapi.CreateProcess(executable, args,
                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
FileNotFoundError: [WinError 2] Le fichier spécifié est introuvable
```

PR #7625, opened 2025-01-17
Ensure esbuild can be detected on windows

philippjfr:
Fixes https://github.com/holoviz/panel/issues/7598

### DIFF
--- a/panel/io/compile.py
+++ b/panel/io/compile.py
@@ -65,7 +65,7 @@ def setup_build_dir(build_dir: str | os.PathLike | None = None):
 
 def check_cli_tool(tool_name):
     try:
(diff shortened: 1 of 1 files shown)


panel/io/compile.py
```diff
@@ -63,15 +63,18 @@ def setup_build_dir(build_dir: str | os.PathLike | None = None):
             shutil.rmtree(temp_dir)
 
 
-def check_cli_tool(tool_name):
+def check_cli_tool(tool_name: str) -> bool:
```
hoxbro, line 66: Maybe just use `shutil.which` to detect instead of launching a subprocess.

philippjfr: Done, thanks!

philippjfr:
Test failure unrelated.

Issue #7580, opened 2025-01-02
panel oauth-secret not a valid command


MarcSkovMadsen:
The section https://panel.holoviz.org/how_to/authentication/configuration.html#encryption mentions you can run `panel oauth-secret`. 

![Image](https://github.com/user-attachments/assets/dab02024-1876-474f-949c-47d9c298e1f1)

But you cannot with panel 1.5.5. Do you mean `panel secret`?

![Image](https://github.com/user-attachments/assets/7a903e68-3040-4349-af35-72ee6184c15a)

PR #7628, opened 2025-01-17
Fix oauth-secret command

philippjfr:
Fixes https://github.com/holoviz/panel/issues/7580

### DIFF
--- a/pixi.toml
+++ b/pixi.toml
@@ -111,6 +111,7 @@ xgboost = "*"
 # =================== TESTS ===================
 # =============================================
 [feature.test-core.dependencies]
+cryptography = "*"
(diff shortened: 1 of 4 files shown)


pixi.toml
```diff
@@ -111,6 +111,7 @@ xgboost = "*"
 # =================== TESTS ===================
 # =============================================
 [feature.test-core.dependencies]
+cryptography = "*"
```
hoxbro, line 114: Can you move this down to `[feature.test.dependencies]` and skip it for the core test?

philippjfr: Maybe

hoxbro: image

Issue #7647, opened 2025-01-21
NestedSelect widgets' options do not change on update


ahuang11:
```
import panel as pn

pn.extension()
n = pn.widgets.NestedSelect(options={"a": {"b": ["c", "d"]}})
display(n)

n.options = {"c": {"d": ["e", "f"]}}

```

<img width="460" alt="Image" src="https://github.com/user-attachments/assets/48f67d20-c7aa-4dd3-98cd-c71b975c68b5" />

PR #7649, opened 2025-01-22
Fix NestedSelect update options

ahuang11:
Closes https://github.com/holoviz/panel/issues/7647

### DIFF
--- a/panel/tests/ui/widgets/test_select.py
+++ b/panel/tests/ui/widgets/test_select.py
@@ -1,5 +1,8 @@
 import pytest
 
+
+pytest.importorskip('playwright')
+
 from playwright.sync_api import expect
 
 from panel.tests.util import serve_component, wait_until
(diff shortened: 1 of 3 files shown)


philippjfr (approved):

panel/tests/ui/widgets/test_select.py
```diff
@@ -1,7 +1,9 @@
 import pytest
 
+from playwright.sync_api import expect
```
philippjfr, line 3:
```suggestion

pytest.importorskip('playwright')

from playwright.sync_api import expect
```

panel/tests/ui/widgets/test_select.py
```diff
@@ -1,7 +1,12 @@
 import pytest
 
+
```
philippjfr, line 3:
```suggestion
```

PR #7650, opened 2025-01-22
ButtonIcon default active_icon to check mark

ahuang11:
```typescript
  get_active_icon(): string {
    return this.model.active_icon !== "" ? this.model.active_icon : `${this.model.icon}-filled`
  }
```
I noticed if I used an icon without `filled`, like `reset`, the button icon disappears completely for the toggle duration.

I'm not sure how to check if a "filled" version of the tabler icon exists, so perhaps for the button icon, default to `check`?

### DIFF
--- a/panel/widgets/icon.py
+++ b/panel/widgets/icon.py
@@ -126,6 +126,8 @@ class ButtonIcon(_ClickableIcon, _ClickButton, TooltipMixin):
     }
 
     def __init__(self, **params):
+        if "active_icon" not in params:
+            params["active_icon"] = "check"
         click_handler = params.pop('on_click', None)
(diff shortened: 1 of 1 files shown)


panel/widgets/icon.py
```diff
@@ -126,6 +126,8 @@ class ButtonIcon(_ClickableIcon, _ClickButton, TooltipMixin):
     }
 
     def __init__(self, **params):
+        if "active_icon" not in params:
+            params["active_icon"] = "check"
```
philippjfr, line 130: Actually, why not define a default?

```python
    active_icon = param.String(default='check')
```

PR #7673, opened 2025-01-29
Fix visible range on objects set in Feed

ahuang11:
Previously, the items would not show if the length of the objects was set drastically different than the original length.

### DIFF
--- a/panel/layout/feed.py
+++ b/panel/layout/feed.py
@@ -78,6 +78,11 @@ class Feed(Column):
         self._last_synced = None
         self.param.watch(self._trigger_view_latest, 'objects')
 
+    @param.depends("objects", watch=True)
+    def _reset_visible_range(self):
+        with edit_readonly(self):
+            self.visible_range = None
+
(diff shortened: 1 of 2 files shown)


panel/layout/feed.py
```diff
@@ -78,6 +78,11 @@ def __init__(self, *objects, **params):
         self._last_synced = None
         self.param.watch(self._trigger_view_latest, 'objects')
 
+    @param.depends("objects", watch=True)
+    def _reset_visible_range(self):
```
philippjfr, line 82: This is **probably** overkill and maybe too late (not sure about the order of operations).

We likely need to:

1. Detect which items were added/removed, i.e. are all currently visible objects still present, likely the best thing would be to simply recompute the visible range based on the children that are still present.
2. We may need to do this inside `_update_model` so that we can adjust the visible_range before `_get_objects` recomputes the objects to send.

philippjfr (approved):

Issue #7676, opened 2025-01-30
Invalid azure oauth redirect when using prefix


MarcSkovMadsen:
panel==1.6.0

In trying to get azure oauth working in combination with a `--prefix`. I'm running on my laptop.

I have the app

```python
import panel as pn
import logging
import sys

pn.extension()

FORMAT = "%(asctime)s | %(levelname)s | %(name)s | %(message)s"

@pn.cache
def get_logger(name, format_=FORMAT, level=logging.DEBUG):
    logger = logging.getLogger(name)

    logger.handlers.clear()

    handler = logging.StreamHandler()
    handler.setStream(sys.stdout)
    formatter = logging.Formatter(format_)
    handler.setFormatter(formatter)
    logger.addHandler(handler)
    logger.propagate = False

    logger.setLevel(level)
    logger.info("Logger successfully configured")
    return logger

get_logger(name="panel.auth")

user = pn.state.user or "Guest User"

pn.panel(user).servable()
```

To serve the app I run

```bash
panel serve script.py --index script --oauth-provider=azure --oauth-key='*' --oauth-secret='*' --cookie-secret='*' --oauth-encryption-key='*' --oauth-extra-params "{'tenant': '*'}" --prefix=mt-devops/panel --index=script --oauth-redirect-uri=http://localhost:5006/mt-devops/panel/
```

- If I use the `/mt-devops/panel/logout` url and click the `login` button I end up at `/mt-devops/panel` (not `/mt-devops/panel/`).
- If I use the `/mt-devops/panel/login` url I end up at `/mt-devops/panel` (not `/mt-devops/panel/`).
- If I use the `/mt-devops/panel/` url I end up at `/mt-devops/panel/`. THIS WORKS.

MarcSkovMadsen:
This video shows the issues

https://github.com/user-attachments/assets/063bc4e2-7983-407d-a7ec-8963b30f9409

MarcSkovMadsen:
I've also tried with `--prefix=mt-devops/panel/`. Does not change the behaviour.

philippjfr:
Let's prioritize this fix along with the `--index` problem and get a new release out asap.

philippjfr:
Can you easily test, what happens when you change it to?

```python
root_url = self.request.uri.replace(self._login_endpoint, '').split('?')[0]
if not root_url.endswith('/'):
    root_url += '/'
```

MarcSkovMadsen:
[...]

PR #7680, opened 2025-01-31
Ensure root_url is correctly determined during auth

philippjfr:
- Fixes https://github.com/holoviz/panel/issues/7676
- [x] Add tests (relies on #7679)


panel/tests/ui/test_auth.py
```diff
@@ -17,8 +17,6 @@
 )
 
 pytestmark = pytest.mark.ui
-auth_check = pytest.mark.skipif('PANEL_TEST_AUTH' not in os.environ, reason='PANEL_TEST_AUTH environment variable is required to run this test')
```
hoxbro, line 20: I would still keep this. To not accidentally run tests on personal systems that have the more generic environment variable set.

Issue #7314, opened 2024-09-23
Redirects to /app instead of ./app


MarcSkovMadsen:
I'm developing on a jupyterhub with [code server](https://anaconda.org/conda-forge/code-server), [jupyter-server-proxy](https://jupyter-server-proxy.readthedocs.io/en/latest/) and [jupyter-code-proxy installed](https://pypi.org/project/jupyter-vscode-proxy/) installed.

This enables me to start up panel app in a vscode terminal and click the url. This correctly opens https://SOME-DOMAIN/SOME-PREFIX/vscode/proxy/5006/. But unfortunately Panel the redirects to `/app` instead of `./app` which ends up redirecting to https://SOME-DOMAIN/proxy/5006/app which does not exist.

I can workaround this issue by appending the `--index app` flag. But no one else in the world would probably be able to figure out this workaround.

## Reproduce on Binder

I've tried to setup binder to reproduce in https://github.com/MarcSkovMadsen/panel-fastapi-experiments. You can open https://mybinder.org/v2/gh/MarcSkovMadsen/panel-fastapi-experiments.git/HEAD to reproduce.

Please note that depending on the binder server this might work or not work. I've experienced some binder servers catching the "proxy" endpoints and displaying their own web page instead.

MarcSkovMadsen:
Its really overhead that you have to add `--dev --index name_of_file --show` everytime you start and restart. With other frameworks you don't have to write all this.

The main challenge is that `name_of_file` is variable so you have to type it in. And you cannot configure the VS Code (Code Server) run button to start the Panel server correctly either because of the variable `name_of_file`.

PR #7704, opened 2025-02-12
Ensure index redirects are relative

philippjfr:
Fixes https://github.com/holoviz/panel/issues/7314


MarcSkovMadsen:
That will be such a time saver for people like me on a JupyterHub. Thx.

Issue #7690, opened 2025-02-06
pyright giving Expected 0 positional arguments for pn.extension() calls


chrisblodgett:
#### ALL software version info

<details>
    <summary>Panel 6.1,Pyright 1.1393 and Python 3.12</summary>
</details>

#### Description of expected behavior and the observed behavior
pyright is giving me an error -

Expected 0 positional arguments

The code runs fine and I don't get any other errors of this type with anything else I'm doing with panel in the rest of the program.

#### Complete, minimal, self-contained example code that reproduces the issue

```python
# code goes here between backticks
import panel as pn

pn.extension('tabulator')
pn.extension('jsoneditor')


```

#### Screenshots or screencasts of the bug in action
![Image](https://github.com/user-attachments/assets/a21f13a2-49c2-45f8-b0ff-2b57a9d0bd59)

- [ ] I may be interested in making a pull request to address this

Coderambling:
Discord thread regarding this issue: https://discord.com/channels/1075331058024861767/1337163041216925706

PR #7709, opened 2025-02-13
Ensure type checkers can correctly infer extension argument types

philippjfr:
Seems like type checkers get confused about `ParameterizedFunction` signatures so we help them a little bit with a `typing.overload`.

Fixes https://github.com/holoviz/panel/issues/7690

### DIFF
--- a/panel/config.py
+++ b/panel/config.py
@@ -723,7 +723,7 @@ class panel_extension(_pyviz_extension):
 
         from .reactive import ReactiveHTML, ReactiveHTMLMetaclass
 
-        _in_ipython = hasattr(builtins, '___IPYTHON__')
+        _in_ipython = hasattr(builtins, '__IPYTHON__')
         reactive_exts = {
             v._extension_name: v for k, v in param.concrete_descendents(ReactiveHTML).items()
         }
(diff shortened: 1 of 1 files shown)


panel/config.py
```diff
@@ -818,10 +823,10 @@ def __call__(self, *args, **params):
             self._load_entry_points()
 
         # Abort if IPython not found
-        try:
-            ip = params.pop('ip', None) or get_ipython() # noqa (get_ipython)
-        except Exception:
+        ipy_fn = globals().get('get_ipython', None)
```
hoxbro, line 826: You could also use builtins.__IPYTHON__, if I recall correctly.

panel/config.py
```diff
@@ -707,12 +708,22 @@ class panel_extension(_pyviz_extension):
...
+        ...
+
+    def __call__(self, *args: str, **params: Any):
         from bokeh.core.has_props import _default_resolver
         from bokeh.model import Model
         from bokeh.settings import settings as bk_settings
 
         from .reactive import ReactiveHTML, ReactiveHTMLMetaclass
+
+        _in_ipython = hasattr(builtins, '___IPYTHON__')
```
hoxbro, line 726:
```suggestion
        _in_ipython = hasattr(builtins, '__IPYTHON__')
```

PR #7715, opened 2025-02-13
fix: Recursion error in `check_cli_tool`

hoxbro:
From holoview CI

image

### DIFF
--- a/panel/io/compile.py
+++ b/panel/io/compile.py
@@ -76,8 +76,7 @@ def check_cli_tool(tool_name: str) -> bool:
     if cmd:
         return True
     if sys.platform == 'win32':
-        tool_name = f'{tool_name}.cmd'
-        return check_cli_tool(tool_name)
+        return shutil.which(f'{tool_name}.cmd')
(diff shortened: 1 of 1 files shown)


philippjfr (approved):

panel/io/compile.py
```diff
@@ -76,8 +76,7 @@ def check_cli_tool(tool_name: str) -> bool:
     if cmd:
         return True
     if sys.platform == 'win32':
-        tool_name = f'{tool_name}.cmd'
-        return check_cli_tool(tool_name)
+        return shutil.which(f'{tool_name}.cmd')
```
hoxbro, line 79: We could simplify this to

``` python
def check_cli_tool(tool_name: str) -> bool:
    if sys.platform == "win32":
        return shutil.which(tool_name) or shutil.which(f"{tool_name}.cmd")
    return shutil.which(tool_name)

philippjfr: Sure, though I'll merge now.

PR #7722, opened 2025-02-19
Refactor ChatFeed post hook

ahuang11:
As I was building out logs for Lumen, I noticed that depending on `message = self._chat_log.objects[-1]` was insufficient due to race conditions (i.e. needed a direct ref to the corresponding message for accuracy).

Also, I noticed that the post hook was prematurely being called before ChatStep finished streaming. Now, it depends on its status changing to success or failed, before triggering the post hook.

Semi-breaking change is `stream` no longer triggers post hook by default.


panel/chat/feed.py
```diff
@@ -968,6 +989,13 @@ async def _prepare_prompt(*args) -> None:
 
         param.parameterized.async_executor(_prepare_prompt)
 
+    def trigger_post_hook(self):
```
philippjfr, line 992: Is this meant to be public and if so why?

ahuang11: I think it's intentionally public, but I forget the reason why I made it public. Vaguely remember having it parallel with respond(), but now that I'm thinking about it, it might not make sense. I'll have to see if I call this in Lumen.

ahuang11: Ah looking at the tests, having it public was intentional so that users can trigger after streaming is done.

```python

    async def test_stream(self, chat_feed):
        def callback(contents, user, instance):
            message = instance.stream("Echo: ", trigger_post_hook=True)
            for char in contents:
                message = instance.stream(char, message=message)
            instance.trigger_post_hook()

        def append_callback(message, instance):
            logs.append(message.object)

        logs = []
        chat_feed.callback = callback
        chat_feed.post_hook = append_callback
        chat_feed.send("AB")
        await async_wait_until(lambda: chat_feed.objects[-1].object == "Echo: AB")
        await async_wait_until(lambda: logs == ["AB", "Echo: ", "Echo: AB"])
```

philippjfr: How would anyone know that they should call it after streaming is done?

ahuang11: Right, I can add docs to this so others know, and also remind myself :)

philippjfr (approved):

PR #7724, opened 2025-02-21
compat: Bokeh 3.7

hoxbro:
Bokeh 3.7 RC is soon gonna be released. 

This tests the latest dev release to see how much work we need. 

EDIT: It is very backward compatible.

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -267,6 +267,7 @@ jobs:
       - uses: holoviz-dev/holoviz_tasks/pixi_install@v0
         with:
           environments: ${{ matrix.environment }}
+          install: false
       - name: Test Type
         run: |
           pixi run -e ${{ matrix.environment }} test-type
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -170,7 +170,7 @@ jobs:
       matrix:
         os: ["ubuntu-latest", "macos-latest", "windows-latest"]
         environment: ["test-ui"]
-    timeout-minutes: 30
+    timeout-minutes: 60
     env:
       PANEL_LOG_LEVEL: info
       OAUTH_COOKIE_SECRET: ${{ secrets.OAUTH_COOKIE_SECRET }}
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -242,7 +242,7 @@ jobs:
       fail-fast: false
       matrix:
         os: ["ubuntu-latest"]
-        environment: ["test-core"]
+        environment: ["test-core", "test-bokeh35", "test-bokeh36"]
     timeout-minutes: 120
     steps:
       - uses: holoviz-dev/holoviz_tasks/pixi_install@v0
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -242,7 +242,7 @@ jobs:
       fail-fast: false
       matrix:
         os: ["ubuntu-latest"]
-        environment: ["test-core", "test-minimum"]
+        environment: ["test-core"]
     timeout-minutes: 120
     steps:
       - uses: holoviz-dev/holoviz_tasks/pixi_install@v0
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -170,7 +170,7 @@ jobs:
       matrix:
         os: ["ubuntu-latest", "macos-latest", "windows-latest"]
         environment: ["test-ui"]
-    timeout-minutes: 120
+    timeout-minutes: 30
     env:
       PANEL_LOG_LEVEL: info
       OAUTH_COOKIE_SECRET: ${{ secrets.OAUTH_COOKIE_SECRET }}
@@ -243,7 +243,7 @@ jobs:
       matrix:
         os: ["ubuntu-latest"]
         environment: ["test-core", "test-bokeh35", "test-bokeh36"]
-    timeout-minutes: 120
+    timeout-minutes: 30
     steps:
       - uses: holoviz-dev/holoviz_tasks/pixi_install@v0
         with:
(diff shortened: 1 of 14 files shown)


panel/models/html.ts
```diff
@@ -123,7 +123,7 @@ export class HTMLView extends PanelMarkupView {
     return [...super.stylesheets(), html_css]
   }
 
-  protected rerender() {
```
philippjfr, line 126: This one scares me and is probably not backward compatible. `rerender` used to be a panel only thing, seems like there's now an equivalent method in Bokeh that probably does something slightly different:

https://github.com/bokeh/bokeh/blob/fde15008802cf1fbb156b9f2af6ce434184bb934/bokehjs/src/lib/core/dom_view.ts#L58

hoxbro: Yeah, was not sure about this one. If we still want to be backward compatible, can't we just rename the function?

panel/links.py
```diff
@@ -490,7 +493,7 @@ def _init_callback(
         else:
             code = f"try {{ {code} }} catch(err) {{ console.log(err) }}"
 
-        src_cb = CustomJS(args=references, code=code, tags=[link_id])
+        src_cb = CustomJS(args=references, code=code, tags=[link_id])  # type: ignore [call-arg]
```
hoxbro, line 496: @mattpap, errored with `error: Unexpected keyword argument "tags" for "CustomJS"  [call-arg]`, which I don't think is correct.

panel/pane/base.py
```diff
@@ -386,6 +386,7 @@ def _update_object(
                         parent.children[node] = new_models  # type: ignore
                         break
             elif isinstance(parent, _BkTabs):
+                parent.tabs = cast(list[_BkTabPanel], parent.tabs)
```
hoxbro, line 389: @mattpap, This shouldn't be needed, in my opinion. This conversion always happens on the Bokeh site of things:

https://github.com/bokeh/bokeh/blob/4693080b6763222cf37680f62518944855ef4540/src/bokeh/models/layouts.py#L586-L589 

Maybe [default factory functions](https://docs.python.org/3/library/dataclasses.html#default-factory-functions) could be a way to reduce this overhead.

panel/widgets/tables.py
```diff
@@ -234,7 +234,7 @@ def _get_column_definitions(self, col_names: list[str], df: pd.DataFrame) -> lis
             if isinstance(data, pd.DataFrame):
                 raise ValueError("DataFrame contains duplicate column names.")
 
-            col_kwargs = {}
+            col_kwargs: dict[str, Any] = {}
```
hoxbro, line 237: In an ideal world this should be replaced with a `TypedDict`

pixi.toml
```diff
@@ -1,6 +1,6 @@
 [project]
 name = "panel"
-channels = ["pyviz/label/dev", "bokeh/label/rc", "conda-forge"]
+channels = ["pyviz/label/dev", "bokeh/label/dev", "conda-forge"]
```
hoxbro, line 3: Reminder to revert before merging.

panel/pane/deckgl.py
```diff
@@ -24,6 +26,7 @@
 if TYPE_CHECKING:
     from bokeh.document import Document
     from bokeh.model import Model
+    from bokeh.models import DataDict
```
hoxbro, line 29: @mattpap, at the top of your head do you know if this has been available since Bokeh 3.5?

panel/pane/deckgl.py
```diff
@@ -162,7 +162,8 @@ def _update_sources(cls, json_data, sources: list[ColumnDataSource]):
         # Create index of sources by columns
         source_columns = defaultdict(list)
         for i, source in enumerate(sources):
-            key = tuple(sorted(source.data.keys()))
+            # source.data = cast("DataDict", source.data)
```
hoxbro, line 165: This didn't work, failed with the error: `ValueError: 
ColumnDataSource.data properties may only be set from plain Python dicts,
not other ColumnDataSource.data values.`

Maybe `cast` could be supported in Bokeh?

panel/links.py
mattpap: Fixed in https://github.com/bokeh/bokeh/pull/14390.

panel/pane/base.py
mattpap: Currently for simplicity `.accepts()` clauses where merged into property types in `*.pyi` files. In the future, relevant declarations will have be split into setters and getters with asymmetric types, if possible at all.

panel/pane/deckgl.py
mattpap: This predates bokeh 3.0.

panel/widgets/tables.py
mattpap: In the ideal world you could use `TableColumn` as-is declared in bokeh. But, for many reasons, we had to declare models as dataclasses and you can't get `TypedDict` from a dataclass. Same problem with `Unpack[]` which was supposed to help with `**kwargs`, but we can't use it.

panel/pane/deckgl.py
hoxbro: An idea would be to make `source.data = source.data` (which is what cast is) a no-op by doing an identity check. I haven't looked at the Bokeh code to see if this is feasible.

panel/widgets/tables.py
hoxbro: This comment was more meant on Panels site of things.

.github/workflows/test.yaml
```diff
@@ -170,7 +170,7 @@ jobs:
       matrix:
         os: ["ubuntu-latest", "macos-latest", "windows-latest"]
         environment: ["test-ui"]
-    timeout-minutes: 120
+    timeout-minutes: 30
```
hoxbro, line 173:
```suggestion
    timeout-minutes: 60
```

Issue #7737, opened 2025-02-26
Improve Checkbox, Toggle and Switch docs to clearly indicate they are all interchangeable and add links


Coderambling:
Textual improvements to mor clearly clarify Checbox, Toggle and Switch are interchangeable. Right now the in the docs of each component it is mentioned for 1 of the other components, instead of both.

Adds links in the doc to the 2 other component docs.

Not entirely sure abou the correct link syntax that should be followed in an .ipnyb doc, so doing this one first. If that works out ok, I will update the other 2 docs accordingly.

I have found this PR related to that. https://github.com/holoviz-dev/nbsite/pull/315 . It introduces an URL parser and some checks.

Is ther a guide to this? It is not clear to me what the correct syntax is for entering links in doc Notebooks to other doc Notebooks, so they show up correctly after they have been converted to website pages.

Is there any guide? If Noteboo02 is in the same dir as Notebook01, the link in Notebook01 to Notebook02 should be [Some text](Notebook02.ipynb}? Or should it be prepended with ./ or / ?

If, for example Notebook03 is 2 dirs up, the link should be (../../Notebook03.ipynb) ?

Also, what about referring to a specific section of a Notebook? Will this work? (../../Notebook03.ipynb#a_section) ?

This is issue is also a partial follow-up on remarks in this PR: https://github.com/holoviz/panel/pull/7729

PR #7738, opened 2025-02-26
Update Checkbox.ipynb to indicate both Toggle and Switch are interchangeable with Checkbox

Coderambling:
…ngeable with Checbox.

Textual improvements to clarify all 3 widgets are interchangeable, and adds links to the 2 docs for the other widgets that are interchangeable. See #7737

No entirely sure if the added links are structured correctly, so that they will be converted to correct links when the webpage is generated from the Notebook. The links work when viewing the new version of the file, insofar as clicking on the link leads to the correct Notebook in Github.

See issue https://github.com/holoviz/panel/issues/7737 for request for further explanation on this.

### DIFF
--- a/examples/reference/widgets/Checkbox.ipynb
+++ b/examples/reference/widgets/Checkbox.ipynb
@@ -14,7 +14,7 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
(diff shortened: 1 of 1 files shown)


philippjfr (approved):

philippjfr:
Thanks @Coderambling. Test failures are spurious.

Coderambling:
Hi @philippjfr . Thanks for making the point about the test failure. I did wonder if that was an issue with the updated doc. Will hopefully update the other (Toggle and Switch) docs with the same text, so the info is uniform across all 3 docs.

Issue #7524, opened 2024-11-26
TextEditor: nested lists not semantically returned


maximlt:
Nested lists entered in the TextEditor aren't returned semantically nested on the `value` parameter. Instead, they're returned flattened and with some custom classes (`ql-indent-<level>`) that Quill, the Javascript library used to implement the editor, uses in the editor to visually indent the items.

```python

import panel as pn

pn.extension('texteditor')

nested_bullets = """
<h2>A Nested List</h2>
<p>Lists can be nested (list inside list):</p>

<ul>
  <li>Coffee</li>
  <li>Tea
    <ul>
      <li>Black tea</li>
      <li>Green tea</li>
    </ul>
  </li>
  <li>Milk</li>
</ul>
"""

editor = pn.widgets.TextEditor(value=nested_bullets)
html = pn.pane.HTML(w.param.value)

pn.Row(editor, html).servable()
```

![Image](https://github.com/user-attachments/assets/dd2185d4-a43c-41fa-96db-603f0f4c60c6)
![Image](https://github.com/user-attachments/assets/d098e188-45e4-496d-91f7-5bc619a4b11c)

This is apparently a pretty old issue in the Quill (https://github.com/slab/quill/issues/979) . In Panel, we get the content of the editor in the front-end using `innerHTML` (https://github.com/holoviz/panel/blob/1f821d651d8ad2b874fb90cbc2f29277f6817464/panel/models/quill.ts#L166) ), so we get the “incorrect” HTML that library uses. Potential solutions:

1. Update the TextEditor to use `getSemanticHTML` instead of `innerHTML` that was added in Quill 2.0 (https://quilljs.com/docs/api#getsemantichtml):
   - We are is still using Quill 1 (intention was to update Quill when it would officially support being rendered in the Virtual DOM which isn’t yet the case), so that needs to be updated first
   - Check whether `getSemanticHTML` is really a good fit (e.g. search related issues) and doesn't introduce other discrepancies.
2. Look for other alternatives like [Fix for Quill.js issue with list indent (use of css class instead of pure HTML)](https://gist.github.com/Daenero/3442213dc5093dc10f30711edb529729)  or https://github.com/nozer/quill-delta-to-html
3. Implement the conversion of nested lists in Python

Preference for option 1.

Issue #7460, opened 2024-11-05
The texteditor widget does not work properly: the toolbar can not format the text as designed.


yejustme:
#### ALL software version info
(this library, plus any other relevant software, e.g. bokeh, python, notebook, OS, browser, etc should be added within the dropdown below.)

<details>
    <summary>Software Version Info</summary>

```plaintext
os: windows 10
python: 3.12.1
panel: 1.5.3
bokeh: 3.4.3
browser: edge 130.0.2849.68
```
</details>

#### The texteditor widget is not working properly. If I enter any text inside and then click the buttons in the toolbar to change the format, such as bold, underline, or change color, it won't work.
However, if these buttons are clicked first, they will become selected, and then the text format can be set by entering it. But this does not conform to our general process of editing text. If every input of text needs to be pre-set to indicate whether it needs to be bolded or italicized, the thought process will be interrupted.
So, it seems like there may be a bug present.https://github.com/user-attachments/assets/b48bed9d-80be-40a6-8fd9-89f13dce0603

#### Complete, minimal, self-contained example code that reproduces the issue

```python

import panel as pn
import param

pn.extension('texteditor')
pn.config.sizing_mode = 'stretch_width'

# Full configuration
full_config = pn.widgets.TextEditor(toolbar=[
  ['bold', 'italic', 'underline', 'strike'],        # toggled buttons
  ['blockquote', 'code-block'],
  ['link', 'image'],
  [{ 'header': 1 }, { 'header': 2 }],               # custom button values
  [{ 'list': 'ordered'}, { 'list': 'bullet' }],
  [{ 'script': 'sub'}, { 'script': 'super' }],      # superscript/subscript
  [{ 'indent': '-1'}, { 'indent': '+1' }],          # outdent/indent
  [{ 'direction': 'rtl' }],                         # text direction

  [{ 'size': ['small', False, 'large', 'huge'] }],  # custom dropdown
  [{ 'header': [1, 2, 3, 4, 5, 6, False] }],

  [{ 'color': ['red','blue','green'] }, { 'background': [] }],          # dropdown with defaults from theme
  [{ 'font': [] }],
  [{ 'align': [] }],

])

pn.Column(full_config).servable()

```

malemburg:
I am seeing the same problem.

The version 1.3.7 of Quill used in Panel is from 2019. 

[...]

PR #7739, opened 2025-02-26
Upgrade Quill to version 2.0.2 and fix toolbar editing and nested lists

maximlt:
Fixes https://github.com/holoviz/panel/issues/7524
Fixes https://github.com/holoviz/panel/issues/7460

By upgrading Quill to version 2.0.2. Note the latest is 2.0.3 but has a weird bug that replaces all white spaces with `nbsp;`. The test `test_texteditor_nested_lists` failed before this PR and now passes. I've also run the tests locally on Firefox and they all pass, one needed some adjustment.

Keeping it a bit in Draft mode as I'd like to test it on a real app (I'm moderately confident on Quill's robustness and v2 still doesn't support the virtual DOM).

### DIFF
--- a/panel/models/quill.py
+++ b/panel/models/quill.py
@@ -13,12 +13,12 @@ class QuillInput(HTMLBox):
     """
 
     __css_raw__ = [
-        'https://cdn.jsdelivr.net/npm/quill@2.0.3/dist/quill.bubble.css',
-        'https://cdn.jsdelivr.net/npm/quill@2.0.3/dist/quill.snow.css',
+        'https://cdn.jsdelivr.net/npm/quill@2.0.2/dist/quill.bubble.css',
+        'https://cdn.jsdelivr.net/npm/quill@2.0.2/dist/quill.snow.css',
     ]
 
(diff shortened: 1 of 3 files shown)


hoxbro:
pre-commit.ci autofix

maximlt:
Failing UI test is unrelated.

I've been trying to run the example notebook locally but without luck due to another bug in how resources are handled in a local development setting.

We discussed whether merging a PR like this one should be part of a patch release (incoming 1.6.2) or a minor release (1.7). I think it's okay if it is released in 1.6.2 as after all this is just affecting one specific widget, which has a blast radius much smaller than the new compatibility to Bokeh 3.7. What is annoying with this widget is that it is already broken in various ways (I just found out the bubble version doesn't seem to work, at least on the website) but it does have a few UI tests that can catch potential regressions, and from my own testing I saw no new bug introduced by this PR.

image

(I guess it wouldn't be a bad idea for me to start working on a new panel extension that leverages another JS rich text editor to see if we can get something better than this one)

PR #7764, opened 2025-03-07
perf: Avoid redundant list operations in `ListLike`

s-alexey:
- Removed `list()` casts and concatenations in `__add__` and `__radd__` methods of `ListLike` and `NamedListLike` containers.


hoxbro:
Thank you

PR #7766, opened 2025-03-07
enh: Also raise error if log_level is above default

hoxbro:
It can be hard to debug why something is failing with `die`. This makes it so that when you have a higher log level than the default, it will raise the exception.  

Default is WARNING:

https://github.com/holoviz/panel/blob/321d116f6f2c808a9ee90e8c544885f37dd7880c/panel/config.py#L261-L263

### DIFF
--- a/panel/command/__init__.py
+++ b/panel/command/__init__.py
@@ -91,7 +91,7 @@ def main(args: list[str] | None = None):
             try:
                 ret = parsed_args.invoke(parsed_args)
             except Exception as e:
-                if config.autoreload or config.log_level in ("ERROR", "CRITICAL"):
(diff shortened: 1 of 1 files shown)


panel/command/__init__.py
```diff
@@ -91,7 +91,7 @@ def main(args: list[str] | None = None):
             try:
                 ret = parsed_args.invoke(parsed_args)
             except Exception as e:
-                if config.autoreload:
+                if config.autoreload or config.log_level in ("ERROR", "CRITICAL"):
```
philippjfr, line 94: That seems the exact opposite of what we want, i.e. we want to show the traceback for more verbose log levels:

```suggestion
                if config.autoreload or config.log_level in ("DEBUG", "INFO"):
```

PR #7769, opened 2025-03-11
Fix cloning of NamedListLike objects

s-alexey:
Fixed an error where cloning with only the `object` keyword argument caused an exception.


s-alexey:
> Looks ok to me. Can you add a test for this?

I've extended existing ones, hope that's ok

philippjfr (approved):

philippjfr:
Thanks for the fix!

PR #7778, opened 2025-03-13
Optimize ipywidget reference transform

philippjfr:
The IPyWidget reference transform attempts to check if an object is an IPyWidget. Since this applies to every reference it can get called A LOT. In most cases this isn't a problem but for certain objects attribute access can actually be slow, e.g. for Bokeh models. We explicitly exclude models here to avoid the problem.

### DIFF
--- a/panel/pane/ipywidget.py
+++ b/panel/pane/ipywidget.py
@@ -6,6 +6,7 @@ from typing import TYPE_CHECKING, Any, ClassVar
 
 import param
 
+from bokeh.model import Model
 from param.parameterized import register_reference_transform
 from pyviz_comms import JupyterComm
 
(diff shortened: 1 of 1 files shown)


panel/pane/ipywidget.py
```diff
@@ -130,7 +130,7 @@ def _ipywidget_transform(obj):
     Transforms an ipywidget into a Parameter that listens updates
     when the ipywidget updates.
     """
-    if not (IPyWidget.applies(obj) and hasattr(obj, 'value')):
+    if isinstance(obj, Model) or not (IPyWidget.applies(obj) and hasattr(obj, 'value')):
```
hoxbro, line 133: Wouldn't it be faster to run hasattr first and then applies?

Issue #6675, opened 2024-04-04
Tabulator background_gradient with frozen index


cdeil:
This is a follow-up to #6570 which was fixed in Panel 1.4. Below is a simple reproducer with the remaining issue we ran into with our app. (Turns out it had nothing to do with our viewer classes and where we create the Tabulator object.)

#### ALL software version info

```
pandas   2.2.1
panel     1.4.0
bokeh     3.4.0
```

#### Description of expected behavior and the observed behavior

When I create a DataFrame with named index "a" and then a Tabulator with `frozen_column=["a"]` the `background_gradient` styling doesn't work properly any more, the first real column "b" is not styled and when using `subset` it is off by one.

My expectation is that it either just works OK, or an error is raised when using `frozen_columns` incorrectly applying it to the index.

#### Complete, minimal, self-contained example code that reproduces the issue

```python
import panel as pn
import pandas as pd

df = pd.DataFrame([[1, 2, 3, 4, 5]], columns=list("abcde")).set_index("a")
tab = pn.widgets.Tabulator(df, frozen_columns=["a"])
tab.style.background_gradient(vmin=0, vmax=1, subset=["c"])
tab.servable()
```

#### Stack traceback and/or browser JavaScript console output

no error or warning

#### Screenshots or screencasts of the bug in action

The column "d" is colored instead of "c" as requested subset.

<img width="303" alt="Screenshot 2024-04-04 at 22 17 06" src="https://github.com/holoviz/panel/assets/852409/a27e9bd6-3313-468a-953f-d9ff5f101b7d">

More generally any subset choice or if coloring all columns it will be off by one, i.e. "b" is never colored.

holovizbot:
This issue has been mentioned on **HoloViz Discourse**. There might be relevant details there:

https://discourse.holoviz.org/t/css-styles-misalign-with-frozen-columns-in-panel-s-tabulator/8768/1

PR #7784, opened 2025-03-14
Update tables.py - Tabulator style with frozen index

cganuza:
#6675 When applying style with frozen index the style was applied with an offset.


JurgenKuyper:
Hi, I am working on the same project as cganuza, and with the proposed correct fix (#7792) the offset is still wrong. we are running on panel version 1.6.3, maybe we are using a very specific usecase? not sure.

PR #7787, opened 2025-03-15
Allow defining root node for ReactComponent

philippjfr:
Allows rendering `ReactComponent` into a defined root node, i.e. not inline into the shadow DOM component. This makes it possible to define globally rendered components.


MarcSkovMadsen:
Hey is that useful? What is the usecase?

philippjfr:
The ability to attach a component to some other component and/or rendering into the `body`, e.g. for a snackbar/notifications.

PR #7794, opened 2025-03-18
Improve ListLike and NamedListLike to better match Python list behavior

s-alexey:
- Allow slice assignment with different-length sequences
- Set default index in pop() to -1
- Fix NamedListLike.remove()
- Minor typing and naming improvements 

Built-in list behavior:
```python
x = ['a', 'b', 'c', 'd']
x[2:] = ['e']
assert x == ['a', 'b', 'e']
```

Previously, similar operations with `Feed` would raise an error:

```python
x = Feed(['a', 'b', 'c', 'd'])
x[2:] = ['e']
```

All positive use-cases are preserved, so this should not introduce breaking changes.

### DIFF
--- a/panel/layout/base.py
+++ b/panel/layout/base.py
@@ -8,7 +8,7 @@ from collections import defaultdict, namedtuple
 from collections.abc import (
     Generator, Iterable, Iterator, Mapping,
 )
-from typing import TYPE_CHECKING, Any, ClassVar
+from typing import TYPE_CHECKING, Any, ClassVar, overload
 
 import param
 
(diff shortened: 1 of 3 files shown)


s-alexey:
I also considered using [abc.MutableSequence](https://docs.python.org/3/library/collections.abc.html#collections.abc.MutableSequence) to enforce list-like behavior, as it would significantly reduce the number of reimplemented methods, but I wasn't sure about the performance implications.

philippjfr (commented):
Fully in favor of this but it seems like the tests don't quite like the changes yet.

s-alexey:
> Fully in favor of this but it seems like the tests don't quite like the changes yet.

There was a typo in one of the test cases. They're still failing, but it's likely unrelated to this PR.

philippjfr (approved):

philippjfr:
Greatly appreciated, thanks again!

PR #7796, opened 2025-03-19
ci: Update pins in pixi.toml

hoxbro:
Remove the libsqlite pin, as upstream has fixed this issue.

I also noticed the UI test failed because of a new release of Playwright.

Also, fix a downstream dependency missing.


hoxbro:
There is nothing controversial with these changes, so I will merge them.

PR #7797, opened 2025-03-19
Add glow to running status

ahuang11:
Add a subtle glow to the ChatStep running status

https://github.com/user-attachments/assets/c880b7bc-9ae1-4a78-8aaf-4ef30112751c


ahuang11:
Kind of want to replace ChatStep's BooleanIndicator with a Button(?) or at least improve its animation later https://discourse.holoviz.org/t/statusbadge-implementation/8653

PR #7819, opened 2025-03-29
Correct ClassSelector keyword in build_todo.md

etihwo:
Correct the keyword since ClassSelector requires class_ and not item_type


philippjfr (approved):

philippjfr:
Much appreciated @etihwo!

Issue #6161, opened 2024-01-05
Any reason to not allow defining custom cdn?


gr-mitulb:
https://github.com/holoviz/panel/blob/46e06398a4f0eafe17be6821b8c19695e0c7e56d/panel/config.py#L263

Hello, I would like to define an internal cdn that we use (we want to run panel applications in offline mode). The above link means that we are limited to 'https://unpkg.com' or 'https://cdn.jsdelivr.net/npm'.

Is there any particular reason for this? Can we just turn this into a generic "String"?

Happy to make the change if you agree with it in princple.

thanks

philippjfr:
To be clear, the `pn.config.cdn_url` is really only used during bundling (i.e. when we build a package and bundle all the CDN based resources into the package that is published to PyPI and conda) in most cases and most resources are loaded from our CDN, specifically `https://cdn.holoviz.org/panel/`. However in fact CDN resources are not the default anyway, the default for the server is that resources are loaded from the server `/static` endpoint. I'm not opposed to making the actual CDN endpoint configurable but my question for you is:

1. How would you then obtain the expected contents of the CDN and keep it updated for each Panel version?
2. Why not simply serve the resources from the `/static` endpoint?

gr-mitulb:
So the specific issue is that when trying to use perspective it seems to want to reach the outside world. https://github.com/holoviz/panel/blob/46e06398a4f0eafe17be6821b8c19695e0c7e56d/panel/models/perspective.py#L79

I couldnt find an obvious way to try and redirect that to a bundled version and hence wanted to redirect it to an internal cdn. Its entirely possible im just missing something obvious and there is a flag to just redirect to the bundled resource?

philippjfr:
Thanks for the clarification, that is a good point and one case I forgot about. There's a small number of components that are currently not properly bundled, i.e. it still fetches these from an external CDN. The real fix is to ensure that all our components are bundled correctly. We should make a list of components that are affected and then figure out if they can easily be bundled or not.

SpeedyH30:
is there any movement on getting these to work in an offline mode?

philippjfr:
>is there any movement on getting these to work in an offline mode?

What is these?

philippjfr:
If you're talking about Perspective, then the answer is yes. 1.4.0 will allow you to use it offline.

SpeedyH30:
sorry, long day, yeah the panel packages without the need to pull from the cdn.

nojaf:
[...]

PR #7826, opened 2025-04-01
Refactored some cdn urls

nojaf:
Related to #6161


nojaf:
Hi @philippjfr, any thoughts on this?

philippjfr:
Will have to test and review in detail but have you run `panel bundle panel` and then tried these components again?

nojaf:
Well, I must admit I'm seriously out of my game here.

It's been months since I used Panel, so a lot of things are less trivial to me right now.

I don't have `panel` available after running the [editable install](https://panel.holoviz.org/developer_guide/index.html#editable-install).

When running `pixi run panel bundle --all` (note that `panel bundle panel` isn't a command), something happened, although I'm not entirely sure it's what I'm after.

When I try out:

```python
import panel as pn

pn.config.npm_cdn = "https://unpkg.com"
pn.extension('katex')
latex = pn.pane.LaTeX('${a}^{{{b}}} = {d}$')
latex.servable()
```

using:

```
pixi run panel serve sample.py --dev --show
```

I do see:

`http://localhost:5006/static/extensions/panel/bundled/katex/katex@0.6.0/dist/katex.min.js`

in my network tab, so I'm guessing it didn't work?

Am I at least trying the right things here?

philippjfr:
Yep, that all sounds correct.

panel/models/vizzu.py
```diff
@@ -34,6 +35,10 @@ class VizzuChart(LayoutDOM):
         f"{config.npm_cdn}/vizzu@{VIZZU_VERSION}/dist/vizzu.min.js"
     ]
 
+    @classproperty
+    def __javascript__(cls):
```
philippjfr, line 39: Don't think this is correct. Let's skip Vizzu for now. We'll likely move it into an extension anyway.

Issue #7830, opened 2025-04-03
CodeEditor should allow setting indent


ahuang11:
For editing YAML specs, it's two spaces, but pressing <Tab> results in 4.
https://panel.holoviz.org/reference/widgets/CodeEditor.html

philippjfr:
Pressing what? Tab?

philippjfr:
Also YAML has no standard for indentation, though most people probably do use 2.

ahuang11:
Apparently if you type `<Tab>` without code backticks, it doesn't show!

Yes, but regardless of language, it would be nice to adjust indentation.

PR #7831, opened 2025-04-03
Add indent and soft_tabs options to CodeEditor

philippjfr:
Fixes https://github.com/holoviz/panel/issues/7830


ahuang11:
image

Works well!

ahuang11 (approved):
I personally prefer soft_tabs to default to True, but that's likely a breaking change

Issue #4074, opened 2022-10-31
Panel, Nginx Load Balancing and Custom Extensions breaks


govinda18:
#### ALL software version info
panel 0.14.0, nginx 1.22.0

#### Description of expected behavior and the observed behavior
I am using the architecture discussed in the bokeh docs for load balancing with panel server - [Load balancing with Nginx
](https://docs.bokeh.org/en/test/docs/user_guide/server.html#load-balancing-with-nginx). This does not work well with custom bokeh extensions such as ipywidgets_bokeh or any other extension for that matter.

The error here is that the javascript extensions are not loaded with a 404 error on the server. Here is a sample error:
```
Failed to repull session Error: Model 'my_custom_ext.models.chart.chart_model.Chart' does not exist. This could be due to a widget or a custom model not being registered before first usage.
```
and in the networks tab the js file at `/static/extensions/my_custom_ext/my_custom_ext.js?v=6b13789e43e5485634533de16a65d8ba9d34c4c9758588b665805435f80eb115` throws 404.

As I debugged a bit, the issue here is that nginx with least connection strategy sends each request to a different server and hence a different server is asked to load the extensions and a different one is requested for the javascript file of the extension. I wondered how is panel loading and it seems like at [this line](https://github.com/holoviz/panel/blob/master/panel/io/resources.py#L98), hardcoding the value to override extension_dirs.

For now, a workaround for me is to use something like `extension_dirs['my_custom_ext'] = str(Path(__file__).parent / "dist")`. Not sure what should be the idea fix here though - feels like it should be bokeh and not panel but logging it here as a starting point.

#### Complete, minimal, self-contained example code that reproduces the issue

Follow https://docs.bokeh.org/en/test/docs/user_guide/server.html#load-balancing-with-nginx to setup and use a simple code like below:

```python
import panel as pn
import ipywidgets as w
pn.Row(w.HTML("asd")).servable()
```

ndmlny-qs:
I've tried to reproduce this error, and I just cannot seem to get it give me the same errors you have gotten. My process is a bit terse, but this is the method I used to try and reproduce your error.

### Create a docker/podman container

This is my `Containerfile` as I am using podman, but I think the same will work with docker.

```dockerfile
FROM ubuntu:latest
RUN apt-get update && \
    apt-get install -y build-essential && \
    apt-get install -y wget && \
    apt-get install -y nginx && \
    apt-get clean && \
    rm -rf /var/lib/apt/lists/*
[...]

PR #7832, opened 2025-04-04
doc: Add section in load balancing about Bokeh extensions 

hoxbro:
Fixes https://github.com/holoviz/panel/issues/4074 by adding an example of how to use it with custom extensions.

### DIFF
--- a/doc/how_to/concurrency/load_balancing.md
+++ b/doc/how_to/concurrency/load_balancing.md
@@ -1,12 +1,25 @@
 # Load balancing
 
-Setting up load balancing is a huge topic dependent on the precise system you are using so we won't go into any specific implementation here. In most cases you set up a reverse proxy (like NGINX) to distribute the load across multiple application servers. If you are using a system like Kubernetes it will also handle spinning up the servers for you and can even do so dynamically depending on the amount of concurrent users to ensure that you are not wasting resources when there are fewer users.
+Setting up load balancing is a huge topic dependent on the precise system you
+are using so we won't go into any specific implementation here. In most cases
+you set up a reverse proxy (like NGINX) to distribute the load across multiple
+application servers. If you are using a system like Kubernetes it will also
+handle spinning up the servers for you and can even do so dynamically depending
+on the amount of concurrent users to ensure that you are not wasting resources
+when there are fewer users.
 
 <figure>
-<img src="https://www.nginx.com/wp-content/uploads/2014/07/what-is-load-balancing-diagram-NGINX-1024x518.png" width="768"></img>
+<img src="https://assets.holoviz.org/panel/how_to/concurrency/what-is-load-balancing-diagram-NGINX.png" width="768"></img>
 <figcaption>Diagram showing concept of load balancing (NGINX)</figcaption>
 </figure>
 
-Load balancing is the most complex approach to set up but is guaranteed to improve concurrent usage of your application since different users are not contending for access to the same process or even necessarily the same physical compute and memory resources. At the same time it is more wasteful of resources since it potentially occupies multiple machines and since each process is isolated there is no sharing of cached data or global state.
+Load balancing is the most complex approach to set up but is guaranteed to
+improve concurrent usage of your application since different users are not
+contending for access to the same process or even necessarily the same physical
+compute and memory resources. At the same time it is more wasteful of resources
+since it potentially occupies multiple machines and since each process is
+isolated there is no sharing of cached data or global state.
 
-To get started configuring a load balancer take a look at the [Bokeh documentation](https://docs.bokeh.org/en/latest/docs/user_guide/server/deploy.html#load-balancing) which provides example configurations for Apache and NGINX.
+To get started configuring a load balancer take a look at the [Bokeh
+documentation](https://docs.bokeh.org/en/latest/docs/user_guide/server/deploy.html#load-balancing)
+which provides example configurations for Apache and NGINX.
--- a/doc/how_to/concurrency/load_balancing.md
+++ b/doc/how_to/concurrency/load_balancing.md
@@ -1,7 +1,7 @@
 # Load balancing
 
 Setting up load balancing is a huge topic dependent on the precise system you
-are using so we won't go into any specific implementation here. In most cases
+are using so the example below will be very generic. In most cases
 you set up a reverse proxy (like NGINX) to distribute the load across multiple
 application servers. If you are using a system like Kubernetes it will also
 handle spinning up the servers for you and can even do so dynamically depending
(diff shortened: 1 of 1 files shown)


doc/how_to/concurrency/load_balancing.md
```diff
@@ -1,12 +1,150 @@
 # Load balancing
 
-Setting up load balancing is a huge topic dependent on the precise system you are using so we won't go into any specific implementation here. In most cases you set up a reverse proxy (like NGINX) to distribute the load across multiple application servers. If you are using a system like Kubernetes it will also handle spinning up the servers for you and can even do so dynamically depending on the amount of concurrent users to ensure that you are not wasting resources when there are fewer users.
+Setting up load balancing is a huge topic dependent on the precise system you
```
Coderambling, line 3: Setting up load balancing is a huge topic, dependent on the precise system you

doc/how_to/concurrency/load_balancing.md
```diff
@@ -1,12 +1,150 @@
...
+are using so the example below will be very generic. In most cases
```
Coderambling, line 4: are using, so the example below will be very generic. In most cases

doc/how_to/concurrency/load_balancing.md
```diff
@@ -1,12 +1,150 @@
...
+you set up a reverse proxy (like NGINX) to distribute the load across multiple
+application servers. If you are using a system like Kubernetes it will also
```
Coderambling, line 6: application servers. If you are using a system like Kubernetes, it will also

doc/how_to/concurrency/load_balancing.md
```diff
@@ -1,12 +1,150 @@
...
+handle spinning up the servers for you and can even do so dynamically depending
```
Coderambling, line 7: handle spinning up / shutting down the servers for you, and can even do so dynamically, depending

doc/how_to/concurrency/load_balancing.md
```diff
@@ -1,12 +1,150 @@
...
+on the amount of concurrent users to ensure that you are not wasting resources
+when there are fewer users.
```
Coderambling, line 9: when there are fewer users, and / or when there is less computing demand.

doc/how_to/concurrency/load_balancing.md
```diff
@@ -1,12 +1,150 @@
...
-<img src="https://www.nginx.com/wp-content/uploads/2014/07/what-is-load-balancing-diagram-NGINX-1024x518.png" width="768"></img>
+<img src="https://assets.holoviz.org/panel/how_to/concurrency/what-is-load-balancing-diagram-NGINX.png" width="768"></img>
 <figcaption>Diagram showing concept of load balancing (NGINX)</figcaption>
 </figure>
 
-Load balancing is the most complex approach to set up but is guaranteed to improve concurrent usage of your application since different users are not contending for access to the same process or even necessarily the same physical compute and memory resources. At the same time it is more wasteful of resources since it potentially occupies multiple machines and since each process is isolated there is no sharing of cached data or global state.
+Load balancing is the most complex approach to set up but is guaranteed to
+improve concurrent usage of your application since different users are not
+contending for access to the same process or even necessarily the same physical
+compute and memory resources. At the same time it is more wasteful of resources
```
Coderambling, line 19: At the same time it can be more resource-intensive. Since each process is isolated, there is no sharing of cached data or global state between processes.

Note: are there potential solutions to isolated cache / global state? Like a cache at a higher level that is accessible by each process?

If so, provide some high-level mention or hint to these solutions?

doc/how_to/concurrency/load_balancing.md
```diff
@@ -1,12 +1,150 @@
...
+since it potentially occupies multiple machines and since each process is
+isolated there is no sharing of cached data or global state.
```
Coderambling, line 22: Should there be a mention that Tornado is used by default, and that Flask and FastAPI are also options? I am not sure if and how using one of these three (or more)? options relates to concurrency / load balancing, but should this at least be mentioned, with links to the Panel docs for use of Flask and FastAPI?

Coderambling (changes requested):
Nice doc and very useful image. What software was used to create it?

Left some suggestions / comments inline, mainly adding some comma's and other interpunction, and suggestions for further clarification.

hoxbro:
Can you do the suggestions as code review ±?

The image itself is the image from nginx. Have just changed it to our server, because the original url no longer exists.

Coderambling:
Yes I went for the comment option by mistake. Can do it as code review but not before this weekend.

If a release is imminent then I will do it during the next release.

Coderambling:
Would be good to add a section on / mention of https://panel.holoviz.org/how_to/profiling/profile.html , so users know there are ways to profile an application for memory use etc. So they can use this to plan / monitor when implementing liad balancing.

doc/how_to/concurrency/load_balancing.md
hoxbro:
```suggestion
Setting up load balancing is a huge topic, dependent on the precise system you
```

doc/how_to/concurrency/load_balancing.md
hoxbro: Removing multiple machines as part of the sentence is incorrect to me, mainly because a way to share resources if you were on the same machine would be to save them to disk. 

I don't think there is a high-level solution. It will depend on the use case. One acceptable solution would be to accept it is isolated; another would be to have a server used to communicate with it or set your system up to have file access. 

People who need to use load balancing will likely already know how to handle it; if not, there will be better resources for this than this page.

doc/how_to/concurrency/load_balancing.md
hoxbro: I don't think, as it has nothing to do with load balancing.

doc/how_to/concurrency/load_balancing.md
```diff
@@ -1,12 +1,156 @@
...
-<img src="https://www.nginx.com/wp-content/uploads/2014/07/what-is-load-balancing-diagram-NGINX-1024x518.png" width="768"></img>
+<img src="https://assets.holoviz.org/panel/how_to/concurrency/what-is-load-balancing-diagram-NGINX.png" width="768"></img>
 <figcaption>Diagram showing concept of load balancing (NGINX)</figcaption>
 </figure>
 
-Load balancing is the most complex approach to set up but is guaranteed to improve concurrent usage of your application since different users are not contending for access to the same process or even necessarily the same physical compute and memory resources. At the same time it is more wasteful of resources since it potentially occupies multiple machines and since each process is isolated there is no sharing of cached data or global state.
+Load balancing is the most complex approach to set up but is guaranteed to
+improve concurrent usage of your application since different users are not
+contending for access to the same process or even necessarily the same physical
+compute and memory resources. At the same time it is more wasteful of resources
```
hoxbro, line 19:
```suggestion
compute and memory resources. At the same time it is more resource-intensive
```

doc/how_to/concurrency/load_balancing.md
```diff
@@ -1,12 +1,156 @@
...
+isolated there is no sharing of cached data or global state.
 
-To get started configuring a load balancer take a look at the [Bokeh documentation](https://docs.bokeh.org/en/latest/docs/user_guide/server/deploy.html#load-balancing) which provides example configurations for Apache and NGINX.
+To get started configuring a load balancer take a look at the [Bokeh
+documentation](https://docs.bokeh.org/en/latest/docs/user_guide/server/deploy.html#load-balancing).
+
+## Use NGINX and Containers with Panel along with other Bokeh extensions
+
+Panel is built on top of Bokeh and uses the Bokeh server to serve applications. To serve Panel-specific resources, Panel is defined as a Bokeh extension.
+Panel's resources are available on a CDN, but if you use other Bokeh Extensions, serving them directly from the Bokeh server can be beneficial.
```
maximlt, line 29: Maybe expand a little on why it can be beneficial.

maximlt (commented):
I ran the files locally pretty successfully with docker, the only change I needed was to remove `localhost/` from `docker run -dt -p 8000:80 localhost/my-container`. The app was then available at http://localhost:8000/proxy/app.

However, I don't manage to understand what this is really fixing. Is it that, in a load-balancing context, serving additional Bokeh extensions like `ipywidgets_bokeh` can only be done reliably with `BOKEH_RESOURCES=server`? Or are there other specific settings make in the Docker or NGINX files that are worth highlighting?

hoxbro:
> However, I don't manage to understand what this is really fixing. Is it that, in a load-balancing context, serving additional Bokeh extensions like ipywidgets_bokeh can only be done reliably with BOKEH_RESOURCES=server?

If you want to serve a custom Bokeh model, in the example it uses `ipywidgets_bokeh`, but it could be a custom Bokeh model developed for internal use only, and needed to be deployed behind a reverse proxy.

doc/how_to/concurrency/load_balancing.md
hoxbro:
```suggestion
Panel's resources are available on a CDN, but if you use other Bokeh Extensions, serving them directly from the Bokeh server can be the only way to access them. 
```

hoxbro: > Is it used as an example of a custom local extension that wouldn't be available on a CDN

Yes. Or something where you can't change the resources to get pulled from a CDN. 

> I'm trying to understand the consequences of all of this.

 If you haven't done that already, I suggest reading the original issue.

maximlt: Perfect thanks!

doc/how_to/concurrency/load_balancing.md
```diff
@@ -1,12 +1,158 @@
...
+Load balancing is the most complex approach to set up but is guaranteed to
+improve concurrent usage of your application since different users are not
+contending for access to the same process or even necessarily the same physical
+compute and memory resources. At the same time it is more wasteful of resources
+since it potentially occupies multiple machines and since each process is
+isolated there is no sharing of cached data or global state.
 
-To get started configuring a load balancer take a look at the [Bokeh documentation](https://docs.bokeh.org/en/latest/docs/user_guide/server/deploy.html#load-balancing) which provides example configurations for Apache and NGINX.
+To get started configuring a load balancer take a look at the [Bokeh
+documentation](https://docs.bokeh.org/en/latest/docs/user_guide/server/deploy.html#load-balancing).
```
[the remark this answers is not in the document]
hoxbro: There is still good information on that page, albeit the static section is not ideal.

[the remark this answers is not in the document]
philippjfr: Okay, I'm fine with it too. I guess we don't point to that section.

PR #7846, opened 2025-04-15
Implement smarter Column child rendering

philippjfr:
The `LayoutDOM` implementation in Bokeh removes child DOM nodes and then re-adds them. This is problematic for certain `React` based scenarios where detaching a DOM node and re-attaching it can make React lose track of certain internal state. Therefore we re-implement the `update_layout` implementation here to avoid removing elements and instead insert the new children in the appropriate locations.

### DIFF
--- a/panel/models/column.ts
+++ b/panel/models/column.ts
@@ -142,50 +142,6 @@ export class ColumnView extends BkColumnView {
   }
 
   override async update_children(): Promise<void> {
-    let current_views = [...this.child_views]
-    const created = await this.build_child_views()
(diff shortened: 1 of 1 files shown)


philippjfr:
@mattpap question for you, do you have any major concerns with this and would you be opposed to upstreaming something like this in Bokeh?

mattpap:
As long as order is maintained, this is fine. In fact, I would prefer to have a non-removing approach in bokeh.

philippjfr:
Okay, will upstream shortly.

Issue #7840, opened 2025-04-07
ChatFeed no longer displays messages in notebooks (since 1.6.2)


s-alexey:
#### ALL software version info
Jupyter lab and Colab

<details>
    <summary>Software Version Info</summary>

```plaintext
panel==1.6.2
```
</details>

#### Complete, minimal, self-contained example code that reproduces the issue
[notebook](https://colab.research.google.com/drive/1gtD6X4Q9TEyK-EskFu3kS10SJqHBxh1G?usp=sharing)

```python
import panel as pn

pn.extension()

feed = pn.chat.ChatFeed(objects=[pn.chat.ChatMessage('Hello World!')], header='# Chat')
feed
```

#### Screenshots or screencasts of the bug in action

<img width="853" alt="Image" src="https://github.com/user-attachments/assets/88ab6fe9-48f3-4c38-96d9-64701e600d32" />
<img width="796" alt="Image" src="https://github.com/user-attachments/assets/f04f9be8-cba9-462a-a2a8-36a1161bb131" />

- [x] I may be interested in making a pull request to address this

This issue was introduced by #7744. Since then, the ChatFeed no longer displays messages in notebook environments like Colab. It appears to be caused by this line: https://github.com/holoviz/panel/blob/60d506e7002562bacc660eef3fdb4285c6c38028/panel/chat/feed.py#L280
Other values like "auto" and None work correctly, but it's unclear what the correct value should be.

Issue #7849, opened 2025-04-15
pn.chat.ChatFeed height cannot be controlled


julioasotodv:
#### ALL software version info
Panel 1.6.1 and 1.6.2 (+ any compatible Bokeh version)

#### Description of expected behavior and the observed behavior

I am trying to make a `pn.chat.ChatFeed` to not fill all available vertical space, but no instantiation argument seems to work: `styles` seems to be ignored, `sizing_mode="stretch_width"` does not work either. So it looks like there is no way to make the element only fill the amount of space that its inner components require.

#### Complete, minimal, self-contained example code that reproduces the issue

```python
# This example includes a pn.chat.ChatFeed as well
# as some other component below (markdown in this case)
# to show that the ChatFeed fills all vertical space:

import panel as pn

pn.extension()

chatfeed = pn.chat.ChatFeed(sizing_mode="stretch_width") # Tried many options to avoid having the ChatFeed
                              # filling all the vertical space

chatfeed.send(pn.chat.ChatMessage("Hello!"))

some_markdown = pn.pane.Markdown("Below chat")

pn.Column(chatfeed, some_markdown).servable()

```

#### Stack traceback and/or browser JavaScript console output

Nothing noteworthy

#### Screenshots or screencasts of the bug in action

<img width="797" alt="Image" src="https://github.com/user-attachments/assets/0b02ec0c-5e16-4710-9da4-481fe81d4116" />

Thank you!

ahuang11:
Does setting `height` satisfy your need?

```
import panel as pn
pn.extension()

cf = pn.chat.ChatFeed(sizing_mode="stretch_width", height=300)
cf.send("HEY")
cf.send("HEY")
cf.send("HEY")
cf.send("HEY")
cf.send("HEY")
cf.send("HEY")
cf.show()
```

<img width="1190" alt="Image" src="https://github.com/user-attachments/assets/392011b6-f4ae-4bda-a032-050454994231" />

PR #7861, opened 2025-04-18
Ensure ChatFeed has inherent height

philippjfr:
Fixes https://github.com/holoviz/panel/issues/7840, https://github.com/holoviz/panel/issues/7849

### DIFF
--- a/panel/chat/feed.py
+++ b/panel/chat/feed.py
@@ -31,7 +31,7 @@ from ..layout.card import Card
 from ..pane.image import SVG, ImageBase
 from ..pane.markup import HTML, Markdown
 from ..util import to_async_gen
-from ..viewable import Children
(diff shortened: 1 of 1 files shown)


panel/chat/feed.py
```diff
@@ -66,6 +66,12 @@ class StopCallback(Exception):
     pass
 
 
+def _stretches_height(obj: Viewable) -> param.rx:
+    return obj.param.sizing_mode.rx().rx.pipe(
+        lambda sm: 'max' if sm is not None and 'height' in sm or 'both' in sm else 'auto'
```
hoxbro, line 71:
```suggestion
        lambda sm: 'max' if sm is not None and ('height' in sm or 'both' in sm) else 'auto'
```

PR #7888, opened 2025-05-06
fix: Jupyterlite not having narwhals as depends

hoxbro:
Bumping pyodide version and adding a workaround until micropip 0.9.0 is released because of https://github.com/pyodide/micropip/issues/171

### DIFF
--- a/scripts/jupyterlite/extra_packages.json
+++ b/scripts/jupyterlite/extra_packages.json
@@ -1,4 +1,3 @@
 [
-  "https://cdn.holoviz.org/panel/wheels/panel-1.5.5-py3-none-any.whl",
-  "pyodide-http"
+  "https://cdn.holoviz.org/panel/wheels/panel-1.5.5-py3-none-any.whl"
 ]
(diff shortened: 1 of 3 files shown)


scripts/jupyterlite/extra_packages.json
```diff
@@ -1,4 +1,3 @@
 [
-  "https://cdn.holoviz.org/panel/wheels/panel-1.5.5-py3-none-any.whl",
-  "pyodide-http"
+  "https://cdn.holoviz.org/panel/wheels/panel-1.5.5-py3-none-any.whl"
```
philippjfr, line 2: Remind me, do we automatically bump this somehow?

hoxbro: No, we want a version that is recent enough so that it can solve it. The panel and bokeh install inside the jupyterlite will be the latest of our pyodide builds. 

 When trying to bump Panel to the latest, I hit the micropip problem.

Issue #7910, opened 2025-05-11
Enable me to configure cookie path


MarcSkovMadsen:
I'm trying to add **Azure Entra Oauth** to one of my applications. The challenge is that **the cookie `path` is set to `/` and thus applies to all my individual panel deployments at my domain** example.com.

![Image](https://github.com/user-attachments/assets/4510dec6-133f-4bc8-8d56-d31600fa49a8)

As a consequence I have to configure oauth using the same configuration for all the Panel apps even though they are created and managed by different teams and don't all need oauth. Right now users of the application that has oauth are not working/ displaying anything because they are not deployed with a cookie secret.

When deploying this specific app I use the `--prefix=https://example.com/sub/path/app`.

Please add a command line flag `--cookie-path` / environment variable `PANEL_COOKIE_PATH` to enable me to set the cookie path enabling me to configure oauth individually for my Panel apps.

The `path` should apply to all the cookies

![Image](https://github.com/user-attachments/assets/29ae01b1-bf53-4f86-bd33-4c15e62c065c)

## Additional Context

Using the *prefix* as *cookie path* would not work. First it would be a breaking change. But for some use cases you might want a shared oauth configuration for an entire domain or a subset of a domain based on many individually deployed Panel apps.

PR #7911, opened 2025-05-11
Support custom cookie path. Improve oauth docs

MarcSkovMadsen:
Closing #7910 and adding oauth documentation that would have made my life much easier.

## Todo

- [x] Create POC
- [x] Systematically go through code and implement feature similarly to how `cookie_secret` is implemented
- [x] Manual Tests
- [x] Documentation

## POC

When setting `PANEL_COOKIE_PATH=apps/debug-panel-oauth ` and opening `/apps/debug-panel-oauth/script` at my domain I see

image

When I open `/` at my domain I see 

image

Which would solve my problem and let individually deployed Panel apps at my shared domain control oauth individually.


philippjfr (changes requested):
Thanks, happy with this, though I'd probably pull the Azure part out into it's own page now that it is quite detailed and do the same for the other providers eventually.

MarcSkovMadsen (commented):
I believe this is ready for review.

I believe this works based on analysis of the code and manual tests. But I don't fully know what to change or how to test more systematically.

philippjfr (approved):

Issue #5128, opened 2023-06-16
Document how to customize the index page


MarcSkovMadsen:
I took a look at the *index* (`/`) page of one of our multi page apps and in the log I can see its actually asking for thumbnails.

```bash
2023-06-16 13:53:26,324 404 GET /thumbnails/app1.png (::1) 0.62ms
2023-06-16 13:53:26,383 404 GET /thumbnails/app2.png (::1) 0.71ms
2023-06-16 13:53:26,384 404 GET /thumbnails/app2.png (::1) 0.76ms
```

I cannot find any documentation on the Panel web site about this. I think it should be documented if you can actually serve thumbnails.

If there are other options that can be configured like the accent color, header logo, header text or header background please describe how. Thanks.

## Additional Context

I still believe it would be very, very valuable and easy (as I have the code) to add a `FastGalleryTemplate` that enables more customization including branding, description, links to the source code, documentation, owner and developer of each app. It would make it an *app catalogue* similar to a *data catalogue*.

MarcSkovMadsen:
## How to get it working

- Put the thumbnails in a folder `./assets/thumbnails`
- Serve the via `--static-dirs thumbnails=./assets/thumbnails`.

There is some css issue though as the picture are not well positioned

![image](https://github.com/holoviz/panel/assets/42288570/a37ef4ba-c338-474d-80be-2a8985810787)

ledwithQ:
@MarcSkovMadsen thanks for this post.
While I agree with your motivations. Can I offer an alternative as someone who came to panel looking to avoid JS CSS etc,?  

I think it would nice to last the index page be a panel page, that the user can extend or modifies with the --index pointing to a .py or .ipynb file?

PR #7916, opened 2025-05-13
Add `--titles` flag to `panel serve` to allow customizing names in the index page cards for multi-page apps and update documentation

### LINKED ISSUE #7863: Allow customizing names in the index page cards when serving a multi-page app
Can it be made possible to customize the names displayed in the cards on the index page when serving multiple apps using `panel serve app1 app2`?

If a directory is named “XYZ Demo”, the corresponding card on the index page displays “Xyz Demo”. I would like to be able to pass the exact string to be displayed.

dalthviz:
This adds a new `--titles` flag to the `panel serve` command to pass a custom list of titles for multi-page apps index card items. This also changes the base `index.html` template to be able to use the values passed as titles via the new flag.

A preview:

* Without custom titles:
```
panel server untitled0.py untitled1.py
```

* With custom titles:
```
panel server untitled0.py untitled1.py --titles titled0 titled1
```

Image

Note:

Currently, the changes to the base index template done in this PR cause the `items` variable passed to become a list of tuples. Not totally sure if the template variables definitions should try to ensure some backwards compatibility. If that is the case, using a list of tuple could be changed to for example pass the titles as a new variable.

Fixes #7863 
Related to #5128

### DIFF
--- a/panel/config.py
+++ b/panel/config.py
@@ -224,6 +224,9 @@ class _config(_base_config):
     throttled = param.Boolean(default=False, doc="""
         If sliders and inputs should be throttled until release of mouse.""")
 
+    titles = param.List(default=[], doc="""
+        Custom titles to use for Multi Page Apps.""")
+
     _admin = param.Boolean(default=False, doc="Whether the admin panel is enabled.")
 
     _admin_endpoint = param.String(default=None, doc="Name to use for the admin endpoint.")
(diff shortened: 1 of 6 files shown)


panel/config.py
```diff
@@ -224,6 +224,9 @@ class _config(_base_config):
     throttled = param.Boolean(default=False, doc="""
         If sliders and inputs should be throttled until release of mouse.""")
 
+    titles = param.List(default=[], doc="""
+        Custom titles to use for Multi Page Apps.""")
```
philippjfr, line 228: Could you also add `_titles` to the list of `_globals` below on this class?

philippjfr (approved):
Let's go 🚀 

Really appreciate all your contributions recently @dalthviz.

Issue #7488, opened 2024-11-13
Chat Widget: Align user messages to right


tomscoorch:
I wanted to kindly suggest a small enhancement to the chat widget that I believe could improve user experience.

Currently, all messages in the chat are aligned to the left by default. In most chat interfaces (such as ChatGPT, WhatsApp, etc.), user messages are typically aligned to the right. This makes it much easier to distinguish between the user's messages and the bot's responses, and overall helps to reduce confusion.

Would it be possible to implement an option where user messages and the associated avatar/logo are aligned to the right, either as a default or on a per-user basis? I believe this small adjustment would make the chat interface much clearer and more user-friendly.

Thank you very much for considering this request. I appreciate your time and look forward to hearing your thoughts on this.

ahuang11:
Sure, I'd be okay with adding an `user_messages_align` param. Would you be interested in implementing it?

Potentially, it's just adding an if statement and `align="end"` here:
https://github.com/holoviz/panel/blob/main/panel/chat/feed.py#L398

tomscoorch:
Dear Amdrew, thanks for the quick response. I am an unprofessional hobby coder with no/few experience in GitHub, so I am afraid I can not implement it myself. Sorry. Hope to get better in coding and be able to contribute some time. Panel is really great. 

 

Von: Andrew ***@***.***> 
Gesendet: Mittwoch, 13. November 2024 09:40
An: holoviz/panel ***@***.***>
Cc: tomscoorch ***@***.***>; Author ***@***.***>
Betreff: Re: [holoviz/panel] Chat Widget: Align user messages to right (Issue #7488)

 

Sure, I'd be okay with adding an user_messages_align param. Would you be interested in implementing it?

Potentially, it's just adding an if statement and align="end" here:
https://github.com/holoviz/panel/blob/main/panel/chat/feed.py#L398

holovizbot:
This issue has been mentioned on **HoloViz Discourse**. There might be relevant details there:

https://discourse.holoviz.org/t/customize-chatinterface/6989/27

isc-patrick:
I'm surprised this has not been mentioned before as the standard for chat interfaces is to use alternating sides for the conversation as it provides the simplest and clearest visual delineation of who said what.

tomscoorch:
@ahuang11 Could you kindly add the feature that user chat messages are aligned to the right. It would make the chat much more easier to read but I am not sure how to implenent this feature. 
Thanks so much in advance.

[...]

PR #7939, opened 2025-05-22
Add per-user styling to ChatFeed (ChatInterface)

CTPassion:
There is a common requirement to only apply certain styles to users within a chat feed. For me, this was obvious when I wanted to align user and agent messages to the left and right, for a classical conversation UI.

It has been mentioned here [holoviz discourse](https://discourse.holoviz.org/t/customize-chatinterface/6989/27) and here #7488 

Rather than just providing an alignment kwarg, I believe it more useful to expose the user to the full suite of styling options they may want to apply on a per-user basis. To align right for example, just apply:

```python
ChatInterface(
    ..., 
    user_messages_styles={"user": {"align-self": "flex-end"}}
)
```

To get it to look correct, I also had to apply some CSS to the .message class:

```python
CHAT_CSS = """
    .message {
        min-height: 15px;
        min-width: fit-content;
        overflow-wrap: normal;
        word-break: normal;
    }
"""
ChatInterface(
    ..., 
    user_messages_styles={"user": {"align-self": "flex-end"}},
    stylesheets=[CHAT_CSS]
)
```

This way users can not just align messages but also change the colours, fonts, sizes etc. as before, but on a per-user basis rather than globally.

### DIFF
--- a/panel/chat/feed.py
+++ b/panel/chat/feed.py
@@ -215,6 +215,16 @@ class ChatFeed(ListPanel):
         Whether to scroll to the latest object on init. If not
         enabled the view will be on the first object.""")
 
+    user_messages_styles = param.Dict(default={}, doc="""
+        A dictionary mapping user names to a dict of CSS style properties.
+        E.g. {"User": {"background": "lightblue"}, "Assistant": {"background": "lightgrey"}}
+    """)
+
+    user_messages_stylesheets = param.Dict(default={}, doc="""
+        A dictionary mapping user names to a list of CSS stylesheets to apply to messages by that user only.
+        E.g. {"User": [".message.user"], "Assistant": [".message {background: lightgrey;}"]}
+    """)
+
     _placeholder = param.ClassSelector(class_=ChatMessage, allow_refs=False, doc="""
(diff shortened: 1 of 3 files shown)


CTPassion:
The unit tests are failing because of an unclosed socket connection, which is unrelated to my code changes (I think). Perhaps someone needs to take a look at this?

panel/chat/feed.py
```diff
@@ -215,6 +215,16 @@ class ChatFeed(ListPanel):
...
 
+    user_messages_styles = param.Dict(default={}, doc="""
+        A dictionary mapping user names to a dict of CSS style properties.
+        E.g. {"User": {"background": "lightblue"}, "Assistant": {"background": "lightgrey"}}
+    """)
+
+    user_messages_stylesheets = param.Dict(default={}, doc="""
+        A dictionary mapping user names to a list of CSS stylesheets to apply to messages by that user only.
+        E.g. {"User": [".message.user"], "Assistant": [".message {background: lightgrey;}"]}
+    """)
```
philippjfr, line 226: I'd probably combine these under `user_message_kwargs` or similar.

Issue #7936, opened 2025-05-22
Overlapping end ticks on full circle panel gauges


killcoder:
Panel full circle gauges have overlapping end ticks as evident on the following discourse [topic](https://discourse.holoviz.org/t/overlapping-end-ticks-on-full-circle-panel-gauges/7390).

Echarts example [clock](https://echarts.apache.org/examples/en/editor.html?c=gauge-clock) has a function to remove the final tick. But I’m not sure how to implement this formatter function with custom-opts.

I would like to be able to remove the overlapping end tick from full circle gauges.

holovizbot:
This issue has been mentioned on **HoloViz Discourse**. There might be relevant details there:

https://discourse.holoviz.org/t/overlapping-end-ticks-on-full-circle-panel-gauges/7390/4

PR #7940, opened 2025-05-22
enh(echarts): Support custom ticklabel formatter

hoxbro:
Resolves https://github.com/holoviz/panel/issues/7936

This could be generalized to a custom model. Currently, this is hardcoded to this one instance. 

image

<details>
<summary>Details</summary>

``` python
import panel as pn
pn.extension("echarts")


code_js = """
(value) => {
   if (value === -180) {
     return '';
   }
   return value + '';
}
"""

w1 = pn.indicators.Gauge(
    name="Roll",
    value=0,
    start_angle=-90,
    end_angle=270,
    bounds=(-180, 180),
    format="{value} Degs",
    colors=[(0.25, "red"), (0.75, "green"), (1, "red")],
    custom_opts={
        "splitNumber": 12,
        "axisLabel": {"formatter": code_js},
    },
)

code_js = """
(value) => {
   if (value === 360) {
     return '';
   }
   return value + '';
}
"""
w2 = pn.indicators.Gauge(
    name="Heading",
    value=0,
    start_angle=90,
    end_angle=-270,
    bounds=(0, 360),
    format="{value} Degs",
    custom_opts={
        "splitNumber": 12,
        "labelLayout": {"hideOverlap": True},
        "axisLabel": {"formatter": code_js},
    },
)


pn.Row(w1, w2).servable()
```

</details>

### DIFF
--- a/panel/models/echarts.ts
+++ b/panel/models/echarts.ts
@@ -89,6 +89,11 @@ export class EChartsView extends HTMLBoxView {
     if ((window as any).echarts == null) {
       return
     }
+    const series = this.model.data.series[0]
+    if (typeof series.axisLabel.formatter === "string") {
+      const format = new Function("return (" + series.axisLabel.formatter + ")")();
+      series.axisLabel.formatter = format
+    }
     this._chart.setOption(this.model.data, this.model.options)
   }
 


panel/models/echarts.ts
```diff
@@ -89,6 +89,11 @@ export class EChartsView extends HTMLBoxView {
     if ((window as any).echarts == null) {
       return
     }
+    const series = this.model.data.series[0]
+    if (typeof series.axisLabel.formatter === "string") {
```
MarcSkovMadsen, line 93: Its a general problem with our echarts implementation that we cannot use a js function. Often you find some example in the documentation. But it uses one or more js functions and we cannot use that.

MarcSkovMadsen: See pyecharts might even support/ work with python functions.

https://github.com/pyecharts/pyecharts-javascripthon

MarcSkovMadsen: PyEcharts provides `JSCode` utility class for supporting js functions.

https://github.com/pyecharts/pyecharts/blob/5228492426e7f7bd6d701af8ae1b130d5c14145c/pyecharts/commons/utils.py#L6

hoxbro: > Its a general problem with our echarts implementation that we cannot use a js function

I will go even further, as each of these escape hatches with static configurations has this problem, another example is [tabulator](https://panel.holoviz.org/reference/widgets/Tabulator.html#static-configuration). 

So I agree we should have a general solution for this. I need to look into how this can be done, as it needs to be done much earlier than here, likely with a custom model. 

This was mainly a POC I pushed (and still a draft) to check if I could do it for this requested instance.

philippjfr: The obvious solution would be to build on bokeh's `CustomJS` support.

philippjfr:
Superseded by #8162

PR #7952, opened 2025-05-30
Add websocket comms how-to guide

MarcSkovMadsen:
In https://github.com/panel-extensions/panel-material-ui/pull/278 I update the `FileInput` to enable uploading large files via chunked file transfer. Thus describing the WebSocket limitations is no longer central there. I also remember mentioning that this should be carved out into a separate guide due to its general nature.


doc/how_to/server/websockets.md
```diff
@@ -0,0 +1,122 @@
...
+| Buffer Size | 100 MB | Tornado default setting |
+| Network/Infrastructure | Varies | Kubernetes, load balancers, proxies may impose additional limits |
+
+## Configuration Guidelines
+
+### Panel/Bokeh Server Deployments
+
+For production applications using **`panel serve`**, configure larger limits at application startup:
+
+WHAT SHOULD BE DONE HERE?
```
MarcSkovMadsen, line 37: Yes. What should be done here? The problem is that there are no `pn.config` parameters you can set. You can set `--websocket-max-message-size`. I see no way of setting `http_server_kwargs`.

I need help from someone to understand how to set these parameters.

doc/how_to/server/websockets.md
```diff
@@ -0,0 +1,122 @@
...
+| Component | Default Limit | Description |
+|-----------|---------------|-------------|
+| Browser | Several GB | Rarely the bottleneck |
+| WebSocket Message Size | 20 MB | Tornado/Bokeh default setting |
+| Buffer Size | 100 MB | Tornado default setting |
+| Network/Infrastructure | Varies | Kubernetes, load balancers, proxies may impose additional limits |
+
+## Configuration Guidelines
+
+### Panel/Bokeh Server Deployments
```
MarcSkovMadsen, line 33: Why is the Bokeh server not just sending chunked messages in general @mattpap?

Why do we have to implement this for specific widgets manually like file upload/ download widgets, audio and video?

These days 10MB is not a lot and audio/ video is everywhere.

Thanks.

MarcSkovMadsen: Could we at least increase the default Tornado settings to 100MB or 250MB?

doc/how_to/server/websockets.md
```diff
@@ -0,0 +1,122 @@
...
+
+**⚠️ What happens when limits are exceeded:**
+
+- Browser tabs may crash or become unresponsive
+- WebSocket connections may close with error messages like `[bokeh] Lost websocket 0 connection, 1009 (message too big)`
+- Applications require page refresh to restore functionality
+
+**💡 Alternative approaches for large transfers:**
+
+- Use file input components with chunked messaging like the [`FileDropper`](https://panel.holoviz.org/reference/widgets/FileDropper.html) to enable uploading large files.
```
hoxbro, line 110:
```suggestion
- Use file input components with chunked messaging like the [`FileDropper`](https://panel.holoviz.org/reference/widgets/FileDropper.html) to enable uploading large files
```

Nit, but your other points do not have it

hoxbro (commented):
Just had a related problem with FileInput this week. And as far as I can see, there is no way to increase the socket size with a CLI command for `panel serve`; I would love to see it, though.

examples/reference/widgets/FileInput.ipynb
```diff
@@ -152,83 +152,7 @@
...
-    "* When a file whose size is larger than the limits is selected by a user, their browser/tab may just crash. Alternatively the web socket connection can close (sometimes with an error message printed in the browser console such as `[bokeh] Lost websocket 0 connection, 1009 (message too big)`) which means the application will become unresponsive and needs to be refreshed. "
+    "While the `FileInput` widget doesn't set any limit on the size of a file that can be selected by a user, the infrastructure onto which Panel relies (web browsers, Bokeh, Tornado, notebooks, kubernetes etc.) limits significantly what is actually possible. By default the `FileInput` widget allows to upload data that is in the order of 10 MB. To increase these limits see the [Websocket Configuration Guide](../../how_to/server/websockets.html)."
```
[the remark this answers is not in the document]
MarcSkovMadsen: Agree. But still as a user I need to understand that more than Panel basic components can limit the size. How can I convey that message?

doc/how_to/server/websockets.md
```diff
@@ -0,0 +1,122 @@
+# Configuring WebSocket Communication Settings
+
+This guide provides comprehensive instructions for configuring bidirectional WebSocket communication in Panel/Bokeh/Tornado/Jupyter/Kubernetes environments to handle larger data transfers effectively.
+
+Out of the box, you can reliably handle communication up to approximately 10 MB. This covers common use cases including visualizing tabular data and plots and uploading smaller files. But for example for audio and video use cases you may quickly run into issues.
```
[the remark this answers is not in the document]
MarcSkovMadsen: Thanks. Did not notice the difference. 

In my experience the limit is approximately 10 MB. Not 15MB, 20MB or 100 MB though. So I think my formulation is better. What do you think?

doc/how_to/server/websockets.md
```diff
@@ -0,0 +1,122 @@
...
+4. **Backend processing** - Your Python application receives and unpacks the JSON Messages.
+
+## Technical Constraints and Default Limits
+
+Understanding the various layers that impose size limitations helps you configure your environment appropriately:
+
+| Component | Default Limit | Description |
+|-----------|---------------|-------------|
+| Browser | Several GB | Rarely the bottleneck |
+| WebSocket Message Size | 20 MB | Tornado/Bokeh default setting |
```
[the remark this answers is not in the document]
MarcSkovMadsen: But for example the `FileInput` base64 encodes increasing the message size significantly. My experience is around 10 MB.

doc/how_to/server/websockets.md
philippjfr: Most servers limit message size for security and performance reasons. Allowing large messages can make it easier to run a DoS attack against a server and even in normal operation can block the server.

PR #7953, opened 2025-05-30
fix: Use `importlib.reload` for autoreload

hoxbro:
I have had a problem when auto-reloading an app as it would error out after the first time. This will reload the module instead of deleting it and having the app import it again. 

Cannot recreate with a MRE, but this is the error I see:

image


philippjfr:
I've tried this before with bad results. Closing for now.

PR #7966, opened 2025-06-04
Ensure references are up-to-date after pyodide syncing

philippjfr:
Since references are now lazily synced we only mark models as non-new if the serializer has synced them.


pyproject.toml
```diff
@@ -243,6 +243,11 @@ disable_warnings = [
     "no-data-collected",
 ]
 
+[tool.coverage.run]
+omit = [
+    "panel/io/__/_templates/*.py"
+]
+
```
philippjfr:
```suggestion
```

PR #7988, opened 2025-06-17
enh: Add header tooltip to Tabulator

hoxbro:
I don't see any reason why this shouldn't be enabled all the time. 

~We likely need to update the CSS so the font type matches the tabulator font.~ Does not seem to be easy [ref](https://stackoverflow.com/questions/60850754/how-to-apply-css-to-tooltips-in-tabulator). 

image

``` python
import pandas as pd
import panel as pn

pn.extension('tabulator')

data = {
    "This is a very long column name that should be truncated": [1, 2, 3],
    "Another extremely long header that will not fit normally": [4, 5, 6],
    "Short": [7, 8, 9]
}
pn.widgets.Tabulator(pd.DataFrame(data), widths=dict.fromkeys(data, 150)).servable()
```

### DIFF
--- a/panel/models/tabulator.ts
+++ b/panel/models/tabulator.ts
@@ -1039,6 +1039,9 @@ export class DataTabulatorView extends HTMLBoxView {
       if (tab_column.title == null) {
         tab_column.title = column.title
       }
+      if (tab_column.headerTooltip === undefined) {
+        tab_column.headerTooltip = true
+      }
       if (tab_column.width == null && column.width != null && column.width != 0) {
         tab_column.width = column.width
       }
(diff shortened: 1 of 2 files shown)


philippjfr:
Does it also show the tooltip for short column names?

hoxbro:
Yes

philippjfr:
As long as there is some reasonable delay on it I'm +1 on enabling it by default.

hoxbro:
There is a delay.

hoxbro:
https://github.com/user-attachments/assets/63c0214b-3425-4763-8dc2-39c2d1d1d7aa

PR #7991, opened 2025-06-18
Allow nested ReactComponents to ignore the shadow DOM

philippjfr:
Currently Bokeh components always create a shadow root that ensures they are isolated from the rest of the page. This isolation is very useful to ensure styling can be well scoped but also has a significant performance penalty. When using custom `ReactComponent`s this performance penalty is compounded by the fact that the shadow DOM boundary layer means that if a `ReactComponent` has children, who are themselves `ReactComponent`s, we still have to create distinct React roots for each child. So now we have nested shadow roots, containing nested React roots.

This PR provides an option to allow children of ReactComponents, that are themselves ReactComponents to be rendered directly into their parents React roots, bypassing the regular Bokeh `View.render` pipeline and instead embedding the child component directly as a React node. This has the major benefit that it is a whole lot faster, but also has the potential to have CSS leak across components in this setup. However, it is not uncommon for a bunch of components or even an entire to be styled very similarly, which means this is a perfectly acceptable approach.

To ensure that CSS and styling still applies:

- We inject the stylesheets on the shadow root of the parent (or parent's parent, ...) 
- We replace `:host` selectors with explicit IDs
- We patch the `View.el` and `View.container` to ensure that HTML attributes are applied to the correct node

Since this has to be enabled explicitly this is fully backwards compatible and I'd like to test this in the wild a bit before documenting the option in more detail.

### DIFF
--- a/panel/custom.py
+++ b/panel/custom.py
@@ -447,7 +447,8 @@ class ReactiveESM(ReactiveCustomBase, metaclass=ReactiveESMMetaclass):
         ignored = [
             p for p in Reactive.param
             if not issubclass(cls.param[p].owner, ReactiveESM) or
-            (p in Viewable.param and p != 'name' and type(Reactive.param[p]) is type(cls.param[p]))
+            (p in Viewable.param and p not in ('name', 'use_shadow_root')
+             and type(Reactive.param[p]) is type(cls.param[p]))
         ]
         for k, v in self.param.values().items():
             p = self.param[k]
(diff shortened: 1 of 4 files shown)


panel/custom.py
```diff
@@ -790,6 +791,13 @@ class CounterButton(pn.custom.ReactComponent):
         CounterButton().servable()
     '''
 
+    use_shadow_root = param.Boolean(default=False, doc="""
```
```suggestion
    use_shadow_root = param.Boolean(default=True, doc="""
```

philippjfr: Will revert the default, but wanted to run the test suite against this default.

hoxbro: Perhaps consider parameterizing this, so we always test both cases.

philippjfr: Guess the problem is that the UI test selectors aren't written to support both.

philippjfr: None of the existing tests exercise this path anyway since they don't nest react components. Added a few new tests to cover that case.

PR #7996, opened 2025-06-23
Fix initial rendering of ReactComponent

philippjfr:
Fixes regression introduced when trying to optimize ReactComponent rendering.


philippjfr:
Jupyterlite issues unrelated.

Issue #8011, opened 2025-07-01
panel compile not working at least on windows


xavArtley:
Windows 10
Panel version: 1.7.2
esbuild --version: 0.25.5
npm --version: 11.0.0

trying the doc example https://panel.holoviz.org/how_to/custom_components/esm/build.html#example-usage result in:

![Image](https://github.com/user-attachments/assets/71e12ab4-0329-4042-8634-56b0158c714f)

hoxbro:
Cannot recreate.  The error message you are seeing is translated "The specified file could not be found", which could indicate it is a Windows only problem. How have you installed nodejs and esbuild?

![Image](https://github.com/user-attachments/assets/9950b178-f86d-4d1c-b116-95d8d72bdabd)

Can you try running with `--verbose`:

![Image](https://github.com/user-attachments/assets/dc8a9bb8-d128-465e-b2c0-3f7afd748f08)

xavArtley:
I think it's specific to windows.
I endup to receate my self every step manually and it worked
May be it's a problem with the tempfiles handle by windows

xavArtley:
actually line 110 of io/compile.py should be replace by:
```python
esbuild_cmd = 'esbuild.cmd' if sys.platform == 'win32' else 'esbuild'
build_cmd = [esbuild_cmd, 'index.js', '--bundle', '--format=esm'] + extra_args
```

xavArtley:
Another error but could require an other issue is:

![Image](https://github.com/user-attachments/assets/220173f8-bc5e-41d2-8b3d-b0080d76f4bf)

![Image](https://github.com/user-attachments/assets/5ed98828-a4c7-41c3-b6ed-648c97e1a81f)


Syntax working is the following one:

![Image](https://github.com/user-attachments/assets/94304bc3-862f-48be-8bd7-6755423d58fc)

and 

![Image](https://github.com/user-attachments/assets/5d5949bd-3fe0-42f2-b0e6-65ed02c778aa)

hoxbro:
> actually line 110 of io/compile.py should be replace by:

We should check for both. I don't have `esbuild.cmd` available on my install. 

Opened #8012, where we can collect all your feedback.

> Another error but could require an other issue is:

Can you clarify what you mean after "syntax woring"?

xavArtley:
add the `.py` to `confetti` to make it work
ommitting the `.py` does not seems to work on windows

hoxbro:
[...]

PR #8012, opened 2025-07-01
enh: Improve check_cli_tool handling

hoxbro:
Improvements from https://github.com/holoviz/panel/issues/8011

### DIFF
--- a/panel/io/compile.py
+++ b/panel/io/compile.py
@@ -68,20 +68,15 @@ def setup_build_dir(build_dir: str | os.PathLike | None = None):
             shutil.rmtree(temp_dir)
 
 
-def check_cli_tool(tool_name: str) -> bool:
-    try:
-        cmd = shutil.which(tool_name)
-    except Exception:
-        cmd = None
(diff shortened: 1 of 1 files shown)


panel/io/compile.py
```diff
@@ -68,20 +68,15 @@ def setup_build_dir(build_dir: str | os.PathLike | None = None):
...
-    except Exception:
-        cmd = None
-    if cmd:
-        return True
+def check_cli_tool(tool_name: str) -> str | None:
     if sys.platform == 'win32':
-        return shutil.which(f'{tool_name}.cmd')
-    return False
+        return shutil.which(tool_name) or shutil.which(f'{tool_name}.cmd')
+    else:
```
xavArtley, line 74: only `shutil.which(tool_name)` seems sufficient even for windows
image

hoxbro: I think this can differ based on the environment variable `PATHEXT`. I believe this is still correct and will not cause any harm. 

image

PR #8021, opened 2025-07-07
Plotly improvements

philippjfr:
- Do not render Plotly if container is undefined
- Do not boomerang Plotly relayout events


philippjfr:
Test failures are related to a transient issue with esm.sh. Will merge.

PR #8026, opened 2025-07-07
Update base.py

bsdz:
stylesheets can be Path objects too.


philippjfr:
Does that resolve your issue?

philippjfr:
In any case I'll merge.

PR #8036, opened 2025-07-09
Add OAuth provider (Auth0) logout url to error template

### LINKED ISSUE #8017: Support showing OAuth logout link in the oauth error template (Auth0 provider)

Hi, so when doing a social login (i.e Google) using Auth0 as OAuth provider, if you choose the wrong account you will be shown an authentication error page:

Image

Since the error page doesn't have [the Auth0 logout link](https://auth0.com/docs/api/authentication/logout/auth-0-logout) to logout of the app, you end up needing to manually type and go to the AuthO logout url (something like `https://<auth0 subdomain>.auth0.com/v2/logout?client_id=<auth0 key>&returnTo=<app logout endpoint/>`) otherwise any subsequent page request will always take you to the error page.

It would be nice for Panel to support showing the OAuth logout link if available for the OAuth provider being used (for example Auth0) in the error page. This probably could be achived by constructing the logout url for the OAuth provider as is done for other OAuth urls like `_OAUTH_ACCESS_TOKEN_URL ` and passing this value when rendering the error template. So at:

https://github.com/holoviz/panel/blob/f44dd4dd78f1093c4a8a766ec67667ae9716b065/panel/auth.py#L474-L496

Passing a new `__OAUTH_LOGOUT_URL`, so something like:

```python
 def write_error(self, status_code, **kwargs): 
     _, e, _ = kwargs['exc_info'] 
     self.clear_all_cookies() 
     self.set_header("Content-Type", 'text/html') 
     if isinstance(e, HTTPError): 
         error, error_msg = e.reason, e.log_message 
     else: 
         provider = self.__class__.__name__.replace('LoginHandler', '') 
         log.error( 
             f'{provider} OAuth provider encountered unexpected ' 
             f'error: {e}' 
         ) 
         error, error_msg = ( 
             '500: Internal Server Error', 
             'Server encountered unexpected problem.' 
         ) 
     self.write(self._error_template.render( 
         npm_cdn=config.npm_cdn, 
         title='Panel: Authentication Error', 
         error_type='Authentication Error', 
         error=error, 
         error_msg=error_msg,
         oauth_logout_url=self.__OAUTH_LOGOUT_URL
     )) 
```

As an alternative, passing to the error template a variable/config as `oauth_extra_params`  could allow users to provide there the OAuth provider logout URL and then they would be able to use this variable in their own custom error template to show the OAuth provider specific logout link. So this only would require a change like:

```python
 def write_error(self, status_code, **kwargs): 
     _, e, _ = kwargs['exc_info'] 
     self.clear_all_cookies() 
     self.set_header("Content-Type", 'text/html') 
     if isinstance(e, HTTPError): 
         error, error_msg = e.reason, e.log_message 
     else: 
         provider = self.__class__.__name__.replace('LoginHandler', '') 
         log.error( 
             f'{provider} OAuth provider encountered unexpected ' 
             f'error: {e}' 
         ) 
         error, error_msg = ( 
             '500: Internal Server Error', 
             'Server encountered unexpected problem.' 
         ) 
     self.write(self._error_template.render( 
         npm_cdn=config.npm_cdn, 
         title='Panel: Authentication Error', 
         error_type='Authentication Error', 
         error=error, 
         error_msg=error_msg,
         oauth_extra_params=Config.oauth_extra_params,
     )) 
```

Currently, you could provide an error template with a hardcoded logout url but then, if you change the params needed for the logout url for example in the case of Auth0  `client_id`, `subdomain` and `returnTo` url, you will end up needing to change the error template itself.

#### Additional context

If this a valid issue, happy to help doing any sort of implementation (from the ideas above or any other that could get pointed out) :)

dalthviz:
Hi, so this is an initial implementation following the first solution described at https://github.com/holoviz/panel/issues/8017#issue-3203289180. With the changes here you will be able to see a `Logout` link in the error page if you are using Auth0 as OAuth provider:

imagen

## Notes:

From a quick check of the supported OAuth providers, I think the only one that also supports a `logout` URL is Okta: https://developer.okta.com/docs/api/openapi/okta-oauth/oauth/tag/OrgAS/#tag/OrgAS/operation/logout, should support for it added here too? Not completely sure from where the value of `id_token_hint` could be retrieved but if worthy let me know and I will check.

Fixes #8017

### DIFF
--- a/panel/_templates/error.html
+++ b/panel/_templates/error.html
@@ -63,6 +63,7 @@
       .error-message {
         max-width: 80%;
         padding-bottom: 10%;
+        text-align: center;
       }
       {% block extra_style %}{% endblock %}
     </style>
@@ -83,6 +84,9 @@
       <section class="error-message">
       {% block content %}
       <h3>{{ error_msg }}</h3>
+      {% if oauth_logout_link %}
+      <a href="{{ oauth_logout_link }}">Logout</a>
+      {% endif %}
       {% endblock %}
       </section>
     </div>
--- a/panel/auth.py
+++ b/panel/auth.py
@@ -492,7 +492,8 @@ class OAuthLoginHandler(tornado.web.RequestHandler, OAuth2Mixin):
             title='Panel: Authentication Error',
             error_type='Authentication Error',
             error=error,
-            error_msg=error_msg
+            error_msg=error_msg,
+            oauth_logout_link=self._OAUTH_LOGOUT_URL,
         ))
 
 
@@ -515,6 +516,10 @@ class GenericLoginHandler(OAuthLoginHandler):
     @property
     def _OAUTH_USER_URL(self):
         return config.oauth_extra_params.get('USER_URL', os.environ.get('PANEL_OAUTH_USER_URL'))
+
+    @property
+    def _OAUTH_LOGOUT_URL(self):
+        return config.oauth_extra_params.get('LOGOUT_URL', os.environ.get('PANEL_OAUTH_LOGOUT_URL'))
 
     @property
     def _USER_KEY(self):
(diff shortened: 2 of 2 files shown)


philippjfr:
Much appreciated! Let's leave it as is for now and revisit Okta if needed.

Issue #7889, opened 2025-05-07
ECharts raising error for mouse events


MarcSkovMadsen:
panel==1.6.3

Motivated by https://www.linkedin.com/posts/andfanilo_i-had-fun-building-this-small-python-echarts-activity-7325769079401709569-YTBB?utm_source=share&utm_medium=member_desktop&rcm=ACoAAADzkDYBcof8grlCeqbPxN3Zc8gYEUZ8EoU I would like to demo how easy an ECharts parent-child chart can be made.

But it seems only `click` events works. For example for `mouseover`  exceptions are raised in the console when hovering over the parent points.

```bash
4panel.min.js?v=f267ab0a5247499879a39c4b98d7d2ac4c72119b3345699d39ccb168dfa154f1:142  Uncaught TypeError: this.mouse is not a function
    at Object.pointer (panel.min.js?v=f267ab0a5247499879a39c4b98d7d2ac4c72119b3345699d39ccb168dfa154f1:142:1294)
    at a.serializeEvent (panel.min.js?v=f267ab0a5247499879a39c4b98d7d2ac4c72119b3345699d39ccb168dfa154f1:142:144)
    at e.t (panel.min.js?v=f267ab0a5247499879a39c4b98d7d2ac4c72119b3345699d39ccb168dfa154f1:207:2397)
    at t.trigger (echarts.min.js:35:13709)
    at e (echarts.min.js:45:203376)
    at t.trigger (echarts.min.js:35:13709)
    at e.dispatchToElement (echarts.min.js:35:26345)
    at e.mousemove (echarts.min.js:35:25266)
    at t.trigger (echarts.min.js:35:13709)
    at e.mousemove (echarts.min.js:35:62582)
```

```python
import panel as pn
import param

_parent_config = {
    "title": {"text": "Parent: Bubble Chart"},
    "tooltip": {"trigger": "item"},
    "xAxis": {"type": "value"},
    "yAxis": {"type": "value"},
    "series": [{
        "type": "scatter",
        "data": [
            {"name": "Category A", "value": [10, 20, 40], "symbolSize": 40},
            {"name": "Category B", "value": [20, 30, 60], "symbolSize": 60},
            {"name": "Category C", "value": [30, 10, 80], "symbolSize": 80},
        ],
        "encode": {"tooltip": [0, 1, 2]}
    }]
}

_child_config = {
    "Category A": {
        "title": {"text": "Child: Details for Category A"},
        "xAxis": {
            "type": "category",
[...]

PR #8041, opened 2025-07-11
fix echarts this.mouse

MarcSkovMadsen:
Closing #7889

I don't fully understand the fix from GPT 4.1. But its simple and the Echarts events works for me in both edge and firefox after the change.

https://github.com/user-attachments/assets/918a140b-6546-429a-a3e1-d7c9ca835967

## GPT 4.1 explanation

The error described in #fetch (TypeError: this.mouse is not a function at Object.pointer) occurs when handling mouse events like mouseover in ECharts, specifically in Edge (but not Firefox).

Looking at echarts.ts, mouse events are handled by the EChartsView._subscribe() method. For each event in event_config, it attaches a callback that serializes the event using serializeEvent:

The error trace points to a problem inside serializeEvent, specifically when it tries to call this.mouse. In event-to-object.ts, there is a class EventTransformCategories with a mouse method, but the context (this) may be lost if the method is called incorrectly (e.g., as a plain function instead of a method).

Root cause:
If EventTransformCategories.mouse is called without binding to an instance, this is undefined, so this.mouse fails. This can happen if the code does something like elementTransformCategories[category](event)category instead of elementTransformCategories.mouse.call(instance, event).

Why Edge but not Firefox?
Edge may be stricter about this binding in certain contexts, or the event object structure may differ, triggering the bug only in Edge.

How to fix:
Ensure that all methods in EventTransformCategories are called with the correct this context, e.g., by using .call(instance, event) or converting them to static methods if they don't rely on instance state.

Summary:
The bug is caused by incorrect method binding in the event serialization logic, leading to this.mouse is not a function when handling mouse events in Edge. Fix the binding in serializeEvent to resolve the issue.

EChartsView
serializeEvent
EventTransformCategories
Similar code found with 1 license type - V

### DIFF
--- a/panel/models/event-to-object.ts
+++ b/panel/models/event-to-object.ts
@@ -272,6 +272,7 @@ const eventTransformCategories: any = new EventTransformCategories()
 
 Object.keys(eventTypeCategories).forEach((category) => {
   eventTypeCategories[category].forEach((type: any) => {
-    eventTransforms[type] = eventTransformCategories[category]
+    // Bind the method to the eventTransformCategories instance to ensure correct 'this'
+    eventTransforms[type] = eventTransformCategories[category].bind(eventTransformCategories)
   })
 })


philippjfr (approved):
Not sure why `this` is needed tbh but if it works, it works.

Issue #8016, opened 2025-07-03
Azure authentication has XSS vulnerability


harmvanderheijden:
#### ALL software version info
<details>
    <summary>Software Version Info</summary>

```plaintext
Seen in panel 1.7.1 and 1.7.2
```
</details>

#### Description of expected behavior and the observed behavior
Configure a panel app for Azure access, running on localhost port 5006, then go to this URL:
http://localhost:5006/login?next=%FF%FE%3C%73%63%72%69%70%74%3E%61%6C%65%72%74%28%32%30%33%29%3C%2F%73%63%72%69%70%74%3E
It should give an error (page not found or something). Instead, it executes the script encoded in the URL (namely: "<script>alert(203)</script>"). The Nessus pen-testing tool (tenable.com) reports this as a medium level cross-site scripting vulnerability.

I have observed this only in the Azure auth route. It is not present using basic-oauth (with a credentials.json file) or when not using auth at all.

#### Complete, minimal, self-contained example code that reproduces the issue

```python
import panel as pn

if pn.state.user and pn.state.user_info:
    user_data = [f"{key} : {value}" for key, value in pn.state.user_info.items()]
    message = pn.pane.Markdown(f"# Hello {pn.state.user}, how do you do?\n\nThis is what I know about you:\n\n{'\n\n'.join(user_data)}")
else:
    message = pn.pane.Markdown(f"# No user info available")

pn.serve(message, port=5006, oauth_provider="azure", oauth_key="irrelevant for this test", oauth_secret="irrelevant for this test", cookie_secret="secretive")

```

Then run the code, and go to this URL:
http://localhost:5006/login?next=%FF%FE%3C%73%63%72%69%70%74%3E%61%6C%65%72%74%28%32%30%33%29%3C%2F%73%63%72%69%70%74%3E

You will see that the script code "alert(203)" is executed.

#### Stack traceback and/or browser JavaScript console output
Console output:

Launching server at http://localhost:5006
400 GET /login?next=%FF%FE%3C%73%63%72%69%70%74%3E%61%6C%65%72%74%28%32%30%33%29%3C%2F%73%63%72%69%70%74%3E (::1): Invalid unicode in next: b'\xff\xfe<script>alert(203)</script>'
400 GET /login?next=%FF%FE%3C%73%63%72%69%70%74%3E%61%6C%65%72%74%28%32%30%33%29%3C%2F%73%63%72%69%70%74%3E (::1) 2.57ms

[...]

PR #8049, opened 2025-07-13
Fix XSS vulnerabilities in authentication templates

harmvanderheijden:
- Add |escape filter to error_msg in error.html template
- Add |escape filter to errormessage in basic_login.html template

These variables can contain user-controlled input that was being rendered without HTML escaping, allowing XSS attacks through malformed URLs like: http://localhost:5006/login?next=%FF%FE%3C%73%63%72%69%70%74%3E%61%6C%65%72%74%28%32%30%33%29%3C%2F%73%63%72%69%70%74%3E

Fixes: holoviz/panel#8016

### DIFF
--- a/panel/_templates/basic_login.html
+++ b/panel/_templates/basic_login.html
@@ -84,7 +84,7 @@
         <p> Login to access your application</p>
       </div>
       <div id="error-message" class="form-group">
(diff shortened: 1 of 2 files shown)


philippjfr (approved):

philippjfr:
Greatly appreciated, thanks.

PR #8055, opened 2025-07-17
Fix the gallery deployment

maximlt:
I tried to lock the project locally, which failed with the usual cryptic conflict kind of message. Removing some pins helped and allowed locking. I then bumped all the pins to reflect the versions used today.

I've manually uploaded the project in AE5 and deployed it, all without problem. Let's see if it all goes well again next time a release is made.


philippjfr:
Thank you!

Issue #8000, opened 2025-06-24
Autocomplete with `restrict=false` unexpected behavior


CmpCtrl:
#### ALL software version info
(this library, plus any other relevant software, e.g. bokeh, python, notebook, OS, browser, etc should be added within the dropdown below.)

<details>
    <summary>Software Version Info</summary>

```plaintext
panel                      1.7.1
param                      2.2.1

python    3.13.0
```
</details>

#### Description of expected behavior and the observed behavior

If you're using `restrict=False` and the widget has a `value`, when you change the `options` it will clear the `value` if the `value` is not in the new `options`. That would make sense in the `restrict=True` case, but with it set `False` i expect it to leave the `value` intact.

#### Complete, minimal, self-contained example code that reproduces the issue

Run this with `panel serve ...` and enter or choose a value, then either clear or change the options. It will clear the `value` if its not in the new `options`.

```python
import panel as pn

options = ["Option 1", "Option 2", "Option 3"]

autocomplete = pn.widgets.AutocompleteInput(
    name="Autocomplete Input",
    options=options,
    restrict=False,
    min_characters=0,
)

clear_button = pn.widgets.Button(
    name="Clear Options",
)

change_options_button = pn.widgets.Button(
    name="Change Options",
)


def clear_options(event):
    """Clear the options of the select widget."""
    autocomplete.options = []


def change_options(event):
    """Change the options of the select widget."""
    autocomplete.options = ["New Option 1", "New Option 2", "New Option 3"]


_ = clear_button.param.watch(clear_options, "clicks")
_ = change_options_button.param.watch(change_options, "clicks")

layout = pn.Column(autocomplete, clear_button, change_options_button)
layout.servable(title="Autocomplete Bug Example")

```

#### Screenshots or screencasts of the bug in action

- [ ] I may be interested in making a pull request to address this

PR #8056, opened 2025-07-17
Do not reset AutocompleteInput when restrict=False

philippjfr:
https://github.com/holoviz/panel/issues/8000


panel/widgets/select.py
```diff
@@ -84,6 +84,7 @@ class SingleSelectBase(SelectBase):
     _allows_none: ClassVar[bool] = False
 
     _supports_embed: bool = True
+    _restrict: bool = False
```
philippjfr:
```suggestion
    _restrict: bool = True
```

PR #8066, opened 2025-07-18
Disable Tabulator internal auto-resize

philippjfr:
`Tabulator.js` by default enables various resize observers that re-render when the container or the page is resized. These potentially trigger multiple times in an resize event which can be very slow for large tables. In Bokeh/Panel we already handle resize events via the `after_resize` method on each model, so Tabulator is resized 2-3x on a window resize event because both our resize handling and the automatic resize handling that Tabulator installs is enabled.

#### Before

tabulator_resize_before

#### After

table_resize_after

### DIFF
--- a/panel/models/tabulator.ts
+++ b/panel/models/tabulator.ts
@@ -393,7 +393,7 @@ export class DataTabulatorView extends HTMLBoxView {
   override connect_signals(): void {
     super.connect_signals()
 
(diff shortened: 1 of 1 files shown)


Coderambling:
Nice optimization! Was this issue surfaced via observation or from performance / unit testing?

Coderambling:
Hi @philippjfr . Why was this closed unmerged?

PR #8086, opened 2025-07-21
Ensure JSONEditor data changes are propagated when the keys of a dict are reordered

maximlt:
Noticed that re-ordering the keys of a dict in the JSONEditor was not triggering a `value` change. Turns out this is because in the front-end a Python dict is a JS `Object`, and that when two `Object`s are compared by Bokeh the order of the keys does not matter. I saw in the code that there was an unused `JSONEditEvent`. Not sure why it was added in the first place, maybe to fix that issue 🙃 . Anyway, that's my starting point for now.

### DIFF
--- a/panel/models/jsoneditor.ts
+++ b/panel/models/jsoneditor.ts
@@ -67,11 +67,11 @@ export class JSONEditorView extends HTMLBoxView {
       menu: this.model.menu,
       mode,
       onChangeJSON: (json: any) => {
-        this.model.data = json
+        this.model.trigger_event(new JSONEditEvent(json))
       },
       onChangeText: (text: any) => {
         try {
-          this.model.data = JSON.parse(text)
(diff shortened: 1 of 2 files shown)


philippjfr:
Totally fine with this.

maximlt:
Yes, I think that was ready to be merged. Just a couple of notes (to self mostly):
- That PR helped receive an event on the Python side when on Object is re-ordered, but that still wouldn't trigger a `value` event as for Param the order of keys in a mapping doesn't matter (same as in Python as `dict(a=0, b=1) == dict(b=1, a=0)` evaluates to True). So the trick will be to use `onlychanged=False`.
- I am not sure why both `onChangeJSON` and `onChangeText` are used. I think this ends up sending two events on a single data change. Might come back to that if that causes some issues in the app where we use this widget.

PR #8089, opened 2025-07-23
Do not schedule model updates if document is on hold

philippjfr:
The `hold` decorator and context manager is an important tool for batching updates, however until now it was less effective than it could have been. Internally in Panel whenever a model update is to be applied we check whether we can safely apply it, i.e. if the socket is unblocked and the event can actually be pushed to the frontend. However, when the document is on hold this check is superfluous because we are not dispatching anything until the document is no longer on hold anyway.

This all has the effect that before despite the fact that we asked for events to be batched before they would actually be scheduled on the async loop and be dispatched separately. This small changes fixes this ensuring that when you ask for an event to be held, it is actually batched together. This should have significant performance implications where `hold` was used.

### DIFF
--- a/panel/io/document.py
+++ b/panel/io/document.py
@@ -551,7 +551,7 @@ def hold(doc: Document | None = None, policy: HoldPolicyType = 'combine', comm: 
         The Comm to dispatch events on when the context manager exits.
     """
     doc = doc or state.curdoc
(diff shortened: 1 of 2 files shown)


philippjfr:
Jupyterlite test failures are unrelated. Will fix separately.

Issue #8112, opened 2025-08-05
Batched update on multiple objects when using multithreaded apps


jp-gr:
#### Is your feature request related to a problem? Please describe.
I was recently looking at the `hold` functionality which is described in the "Improve Performance" section of the docs: https://panel.holoviz.org/how_to/performance/hold.html

When attempting to use this in a multi-threaded environment (as suggested in the "Improve Scalability" section of the docs: https://panel.holoviz.org/how_to/concurrency/threading.html) I ran into the issue described here: https://github.com/holoviz/panel/issues/7590

I commented on that issue with a work around to use `with param.parameterized.batch_call_watchers(app_params):`, this works fine if I just have one `param.Parameterized` object holding my application state, but I have a situation where I want to batch updates to more than one `param.Parameterized` object - I resorted to changing the model of my application to work around this restriction.

Ideally I want my application to conform to both performance and scalability best practices with elegant syntax!

#### Describe the solution you'd like
Make `@hold()` work in a multi-threaded environment, this would be my preferred since the syntax is more elegant

#### Describe alternatives you've considered
Make a version of `param.parameterized.batch_call_watchers()` which accepts more than one parameter (i.e. change function signature to `*args`)

The syntax is more clunky but workable

PR #8113, opened 2025-08-05
Add support for using hold from a thread

philippjfr:
Attempt at fixing https://github.com/holoviz/panel/issues/8112

@jp-gr could you give this a try?


philippjfr:
I will merge since this is a no-op in the non-threaded case and in a threaded scenario things were broken before this.

PR #8116, opened 2025-08-07
compat: Bokeh 3.8

hoxbro:
Just to see if there is any compatibility problems. Should not be merged before the RC has been released.


hoxbro:
Failure of `panel/tests/ui/io/test_notifications.py::test_ready_notification` are reported here: https://github.com/bokeh/bokeh/issues/14597

PR #8118, opened 2025-08-09
Allow use of name for FileInput and Switch

etihwo:
Bokeh model for FileInput accepts 'title' as argument in the latest version while in panel this argument is not passed. In addition when title is not set in the bokeh model, the 'description' is not displayed.

if this PR is accepted, I think the PR #1312 should be reverted to document the keyword.

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -242,7 +242,7 @@ jobs:
       fail-fast: false
       matrix:
         os: ["ubuntu-latest"]
-        environment: ["test-core", "test-bokeh35", "test-bokeh36"]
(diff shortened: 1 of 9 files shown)


hoxbro:
This PR should be merged after https://github.com/holoviz/panel/pull/8116, which drops support for Bokeh 3.5 and Bokeh 3.6.

philippjfr:
Note `name` in general is slated for deprecation. I'm okay with aligning with `name` though and then cutting over to `label` for all widgets.

philippjfr:
I'll merge and start working on the `label` migration.

PR #8120, opened 2025-08-12
Implement server re-connect via Notifications

philippjfr:
Builds on recent work in Bokeh to allow re-connecting to a server.

### DIFF
--- a/doc/how_to/server/reconnect.md
+++ b/doc/how_to/server/reconnect.md
@@ -6,8 +6,7 @@ This guide walks you through configuring **automatic re-connection to Panel serv
 
 ## Why re-connection matters
 
-Whenever a user visits an endpoint, a new server session is created.
-The session stays alive as long as the user remains connected.
+Whenever a user visits an endpoint, a new server session is created. The session stays alive as long as the user remains connected.
 
 However, in some scenarios—such as **poor internet connectivity** or **proxies interrupting WebSocket connections**—the connection between the server and the frontend may close.
 
@@ -36,12 +35,23 @@ When enabled, the following behavior is triggered if the WebSocket connection is
 1. An **exponential backoff loop** attempts to re-establish the connection (after 1, 2, 4, 8, 16, and 32 seconds).
    Users will see notifications during these attempts.
 
-2. If all automatic attempts fail, a **final notification** is shown, offering the user a button to start another re-connection attempt.
+2. If all automatic attempts fail, a **final notification** is shown, offering the user a notification with the option to start another re-connection attempt.
 
 :::{tip}
-You can customize the messages shown to the user—see [Configuring Notifications](#configuring-notifications).
+You can customize the messages shown to the user, see [Configuring Notifications](#configuring-notifications).
 :::
 
+If you don't want to automatically trigger re-connect attempts you can also set:
+
+```python
+pn.config.reconnect = "prompt"
+pn.config.notifications = True
+
+# or equivalently
+pn.extension(notifications=True, reconnect="prompt")
+```
+
+This will only trigger 2., allowing the user to trigger a re-connection attempt by clicking on the notification.
 
 ## Session management
 
--- a/doc/how_to/server/reconnect.md
+++ b/doc/how_to/server/reconnect.md
@@ -19,7 +19,7 @@ This feature requires Panel **≥1.8** and Bokeh **≥3.8**. If you are using an
 
 ## Enabling re-connect
 
-To enable automatic server re-connection, set the corresponding configuration option:
+To enable automatic server re-connection, set the corresponding configuration options:
 
 ```python
 import panel as pn
--- a/doc/how_to/server/reconnect.md
+++ b/doc/how_to/server/reconnect.md
@@ -1,0 +1,106 @@
+# Re-connecting to a Session
+
+This guide walks you through configuring **automatic re-connection to Panel server sessions** when the connection is disrupted.
+
+---
+
+## Why re-connection matters
+
+Whenever a user visits an endpoint, a new server session is created.
+The session stays alive as long as the user remains connected.
+
+However, in some scenarios—such as **poor internet connectivity** or **proxies interrupting WebSocket connections**—the connection between the server and the frontend may close.
+
+By default, this **permanently disconnects** the application.
+Recent versions of Panel introduce the ability to **automatically re-connect** to the server instead of dropping the session.
+
+:::{note}
+This feature requires Panel **≥1.8**.
+If you are using an older version, you will need to upgrade.
+:::
+
+## Enabling re-connect
+
+To enable automatic server re-connection, set the corresponding configuration option:
+
+```python
+import panel as pn
+
+pn.config.reconnect = True
+
+# or equivalently
+pn.extension(reconnect=True)
+```
+
+When enabled, the following behavior is triggered if the WebSocket connection is lost:
+
+1. An **exponential backoff loop** attempts to re-establish the connection (after 1, 2, 4, 8, 16, and 32 seconds).
+   Users will see notifications during these attempts.
+
+2. If all automatic attempts fail, a **final notification** is shown, offering the user a button to manually retry.
+
+:::{tip}
+You can customize the messages shown to the user—see [Configuring Notifications](#configuring-notifications).
+:::
+
+
+## Session management
+
+Sessions are automatically managed by the server and periodically cleaned up if no active connection exists.
+
+You can control session cleanup behavior via CLI flags or programmatically:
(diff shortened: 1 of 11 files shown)


panel/tests/ui/io/test_notifications.py
```diff
@@ -90,6 +87,37 @@ def app():
...
 
+    expect(page.locator('.notyf__message')).to_have_text('Disconnected!')
+
+def test_reconnect_notification(page):
+    doc = None
+
+    def app():
+        nonlocal doc
+        doc = state.curdoc
+        config.disconnect_notification = 'Disconnected!'
```
philippjfr:
```suggestion
        doc = state.curdoc
        config.reconnect = True
        config.disconnect_notification = 'Disconnected!'
```

doc/how_to/server/reconnect.md
```diff
@@ -0,0 +1,106 @@
...
+Whenever a user visits an endpoint, a new server session is created.
+The session stays alive as long as the user remains connected.
+
+However, in some scenarios—such as **poor internet connectivity** or **proxies interrupting WebSocket connections**—the connection between the server and the frontend may close.
+
+By default, this **permanently disconnects** the application.
+Recent versions of Panel introduce the ability to **automatically re-connect** to the server instead of dropping the session.
+
+:::{note}
+This feature requires Panel **≥1.8**.
```
hoxbro:
```suggestion
This feature requires Panel **≥1.8** and Bokeh **≥3.8**.
```

doc/how_to/server/reconnect.md
```diff
@@ -0,0 +1,106 @@
...
+# or equivalently
+pn.extension(reconnect=True)
+```
+
+When enabled, the following behavior is triggered if the WebSocket connection is lost:
+
+1. An **exponential backoff loop** attempts to re-establish the connection (after 1, 2, 4, 8, 16, and 32 seconds).
+   Users will see notifications during these attempts.
+
+2. If all automatic attempts fail, a **final notification** is shown, offering the user a button to manually retry.
```
hoxbro:
```suggestion
2. If all automatic attempts fail, a **final notification** is shown, offering the user a button to manually retry.
```

What does a manual retry mean? Reloading the page or just try the same thing again?

doc/how_to/server/reconnect.md
```diff
@@ -0,0 +1,106 @@
...
+---
+
+## Why re-connection matters
+
+Whenever a user visits an endpoint, a new server session is created.
+The session stays alive as long as the user remains connected.
+
+However, in some scenarios—such as **poor internet connectivity** or **proxies interrupting WebSocket connections**—the connection between the server and the frontend may close.
+
+By default, this **permanently disconnects** the application.
```
hoxbro:
```suggestion
By default, this **permanently disconnects** the application.
```

Perhaps write about what permanently disconnects it, e.g., you need to reload the page, and your state is lost.

doc/how_to/server/reconnect.md
```diff
@@ -0,0 +1,106 @@
...
+
+* `--check-unused-sessions` / `check_unused_sessions_milliseconds`
+  How often to check for unused sessions (default: **17s**).
+
+### Blocking session expiration
+
+Sometimes you may want to prevent a session from expiring—for example, if a **long-running computation** is in progress and you want the user to see the results even after re-connecting.
+
+```python
+pn.state.block_expiration()
```
hoxbro: Is there a context manager for this?

doc/how_to/server/reconnect.md
```diff
@@ -0,0 +1,106 @@
...
+## Configuring notifications
+
+Panel provides options to customize the notifications shown when disconnection occurs:
+
+* **`pn.config.disconnect_notification`**
+  Message displayed when the server disconnects, along with information about automatic re-connect attempts.
+
+:::{tip}
+Use a short, user-friendly message like:
+`pn.config.disconnect_notification = "Lost connection to server. Trying to reconnect..."`
```
hoxbro, line 104: Should we have a way to set the "final" notification when a reconnect is not possible?

panel/io/server.py
```diff
@@ -277,6 +277,9 @@ def server_html_page_for_session(
         dist_url = CDN_DIST
 
     doc = session.document
+    if BOKEH_GE_3_8:
+        with set_curdoc(doc):
+            doc.config.reconnect_session = config.reconnect
```
hoxbro:
```suggestion
            doc.config.reconnect_session = config.reconnect
            doc.config.notifications = None
            doc.config.notify_connection_status = False
```

I assume we don't want the bokeh notifications.

hoxbro: Even with these enabled I still see these with a hard refresh

Screenshot From 2025-08-27 18-07-01

doc/how_to/server/reconnect.md
```diff
@@ -0,0 +1,106 @@
...
+
+To enable automatic server re-connection, set the corresponding configuration option:
+
+```python
+import panel as pn
+
+pn.config.reconnect = True
+
+# or equivalently
+pn.extension(reconnect=True)
```
hoxbro: Do you also need to enable notifications for this to work? Like this `pn.extension(reconnect=True, notifications=True)`

hoxbro (commented):
EDIT: I just have to kill the server and restart it. 

Can you give me an example of how to view the notification? I don't seem to be able to do so...

What I have done: `panel serve example.py --unused-session-lifetime=100  --check-unused-sessions=500`

``` python
import panel as pn

pn.extension(reconnect=True, notifications=True)

# Select from alphabet
import string
alphabet = list(string.ascii_uppercase)
cross_select = pn.widgets.CrossSelector(options=alphabet, value=['A'], size=5)

cross_select.servable()
```

doc/how_to/server/reconnect.md
philippjfr: There technically is no final message because the user will be prompted to attempt a manual reconnect.

doc/how_to/server/reconnect.md
philippjfr: There is now.

panel/io/state.py
```diff
@@ -598,6 +598,33 @@ def add_periodic_callback(
...
+                session.unblock_expiration()
+
+    def unblock_expiration(self):
+        """
+        Unblock the expiration of the current session.
+        """
+        if self.curdoc is None or self.curdoc.session_context is None:
+            return
+        for session in self.curdoc.session_context.server_context.sessions:
+            session.unblock_expiration()
```
hoxbro, line 631:
```suggestion
```

Is this needed anymore?

doc/how_to/server/reconnect.md
```diff
@@ -0,0 +1,99 @@
...
+
+By default, this **permanently disconnects** the application, i.e. any interaction with the server is lost and can only be restored by reloading the page. Recent versions of Panel introduce the ability to **automatically re-connect** to the server instead of dropping the session.
+
+:::{note}
+This feature requires Panel **≥1.8** and Bokeh **≥3.8**. If you are using an older version, you will need to upgrade.
+:::
+
+## Enabling re-connect
+
+To enable automatic server re-connection, set the corresponding configuration option:
```
hoxbro:
```suggestion
To enable automatic server re-connection, set the corresponding configuration options:

doc/how_to/server/reconnect.md
hoxbro: Yes, but there could be a case for adding more information here. Not a blocker tbh.

panel/io/server.py
hoxbro: Do you also see the notification above? Isn't this a Bokeh notification?

panel/io/state.py
philippjfr: Yes, you may not want to use block_expiration as a context manager.

panel/io/server.py
philippjfr: Testing now.

philippjfr: I do not see it.

hoxbro: Can you try in firefox?

panel/io/state.py
hoxbro: Am I missing something? I don't think you can call `block_expiration` directly, or it won't do anything. 

image

philippjfr: No you are correct. Will fix.

panel/io/server.py
philippjfr: Same effect, it does not show up.

hoxbro: @mattpap, FYI

Issue #8130, opened 2025-08-17
Lots of exceptions raised when reload app across multiple taps


MarcSkovMadsen:
panel==1.7.5

I've created an Panel application which basically works as described below.

We've defined a global instance of

```python
class DataStore(param.Parameterized):
     value = param.DataFrame()

global_data_store = DataStore()
```

The `global_data_store.value` is updated every 3 seconds from a loop running in a separate thread. Every application then displays the DataFrame in a table `Tabulator(value=global_data_store.param.value, ....)`. The Tabulator table is inside a layout in the `main` area of a `FastListTemplate`. 

Ok so actually the global_data_store has multiple different parameters that we display in different Tabulator tables. But the concept is the same.

I `panel serve ...` the application.

Initially the application loads nicely. But when I start stress testing by opening multiple tabs and reloading some of them too I start seeing some exceptions raised in the terminal:

<details><summary>RuntimeError: Models must be owned by only a single document, ImportedStyleSheet(id='p1008', ...) is already in a doc</summary>

```bash
2025-08-17 18:16:10,665 | ERROR | tornado.application | Exception in callback functools.partial(<bound method IOLoop._discard_future_result of <tornado.platform.asyncio.AsyncIOMainLoop object at 0x7f9b431e3950>>, <Task finished name='Task-33876' coro=<ServerSession.with_document_locked() done, defined at /home/jovyan/repos/revenue-price-gui/.venv/lib/python3.11/site-packages/bokeh/server/session.py:77> exception=RuntimeError("Models must be owned by only a single document, ImportedStyleSheet(id='p1008', ...) is already in a doc")>)
Traceback (most recent call last):
  File "/home/jovyan/repos/revenue-price-gui/.venv/lib/python3.11/site-packages/tornado/ioloop.py", line 758, in _run_callback
    ret = callback()
          ^^^^^^^^^^
  File "/home/jovyan/repos/revenue-price-gui/.venv/lib/python3.11/site-packages/tornado/ioloop.py", line 782, in _discard_future_result
    future.result()
  File "/home/jovyan/repos/revenue-price-gui/.venv/lib/python3.11/site-packages/bokeh/server/session.py", line 94, in _needs_document_lock_wrapper
    result = func(self, *args, **kwargs)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #8137, opened 2025-08-18
Improve handling of curdoc during initialization and scheduled events

philippjfr:
This PR resolves two major sources of issues where the current document was not correctly tracked.

Specifically, we were relying on two sources of truth for the current document, the bokeh `curdoc()` utility and our internal `state.curdoc` setup. It turns out that, at least for our purposes, the bokeh `curdoc()` is somewhat unreliable in high load situation, i.e. when new sessions are created while other sessions are processing events there can be leakage between the two resulting in the wrong document being provided.

The second issue was that when scheduling an event using `document.add_next_tick_callback` the internal tracking sometimes caused the `state.curdoc` to be confused, which in turn would cause the wrong stylesheet cache to be used, resulting in models leaking across sessions. It may be the case that we need to further tighten this up and consistently use `state.execute(..., schedule=True)` throughout which in turn uses `set_curdoc` before the provided scheduled callback is executed.

Fixes https://github.com/holoviz/panel/issues/8130

### DIFF
--- a/panel/io/application.py
+++ b/panel/io/application.py
@@ -72,7 +72,7 @@ def _eval_panel(
     with set_curdoc(doc):
         if isinstance(panel, (FunctionType, MethodType)):
             panel = panel()
-        if panel is None and hasattr(state, "template"):
+        if panel is None and config.template:
             panel = state.template
         if isinstance(panel, BaseTemplate):
             doc = panel._modify_doc(server_id, title, doc, location)


panel/io/application.py
```diff
@@ -72,9 +72,11 @@ def _log_session_destroyed(session_context):
     with set_curdoc(doc):
         if isinstance(panel, (FunctionType, MethodType)):
             panel = panel()
+        if panel is None and state.template:
```
hoxbro:
```suggestion
        if panel is None and hasattr(state, "template"):
```

hoxbro: Just to see if it would make test happier, you can always revert it.

philippjfr: Should have been config.template, fixed.

MarcSkovMadsen:
Thanks for fixing this. I was wondering whether it would be possible to add automated tests? To ensure Panel always works great for "streaming" applications 1) no issues 2) performs well?

PR #8158, opened 2025-08-29
Fix issue #8157

### LINKED ISSUE #8157: UnboundLocalError thrown by reactive.py

<details>

```plaintext
panel v1.7.5
python 3.12
nixos
```
</details>

The following exception is thrown:
```
UnboundLocalError: cannot access local variable 'vs' where it is not associated with a value
```
by https://github.com/holoviz/panel/blob/2a7b50c78773947d785ee58a3f6f4764ddc5e025/panel/reactive.py#L2197

I fixed it (I think) by remplacing `vs` by `v`.

I don't know what minimal configuration can reproduce it, but I believe that the error is easy enough to understand.

c-meier:
Fix #8157: UnboundLocalError


philippjfr:
Much appreciated!

Issue #2634, opened 2021-08-14
Support JsCode in ECharts when using dicts or pyecharts objects


MarcSkovMadsen:
A user is reporting on discourse that there is a json decoding error in his when using pyecharts https://discourse.holoviz.org/t/echarts-with-custom-js/2653.

In fact I think it would be nice to have a general convention/ framework for supplying js in dicts to charts like ECharts. I support this in Panel Highcharts. The user can just supply strings starting with `function(`. They will be interpreted as js functions.

```python
import panel as pn
pn.extension('echarts')

from pyecharts import options as opts
from pyecharts.charts import Bar
from pyecharts.commons.utils import JsCode
from pyecharts.globals import ThemeType

list2 = [
    {"value": 12, "percent": 12 / (12 + 3)},
    {"value": 23, "percent": 23 / (23 + 21)},
    {"value": 33, "percent": 33 / (33 + 5)},
    {"value": 3, "percent": 3 / (3 + 52)},
    {"value": 33, "percent": 33 / (33 + 43)},
]

list3 = [
    {"value": 3, "percent": 3 / (12 + 3)},
    {"value": 21, "percent": 21 / (23 + 21)},
    {"value": 5, "percent": 5 / (33 + 5)},
    {"value": 52, "percent": 52 / (3 + 52)},
    {"value": 43, "percent": 43 / (33 + 43)},
]

c = (
    Bar(init_opts=opts.InitOpts(theme=ThemeType.LIGHT))
    .add_xaxis([1, 2, 3, 4, 5])
    .add_yaxis("product1", list2, stack="stack1", category_gap="50%")
    .add_yaxis("product2", list3, stack="stack1", category_gap="50%")
    .set_series_opts(
        label_opts=opts.LabelOpts(
            position="right",
            formatter=JsCode(
                "function(x){return Number(x.data.percent * 100).toFixed() + '%';}"
            ),
        )
    ))

c.render("stack_bar_percent.html")   # This works fine without panel

pn.pane.ECharts(c).servable() # This runs into a JSON decode error
```

The below is the value of `object.dump_options()` line 76 in `echarts.py`.

```bash
{
    "animation": true,
    "animationThreshold": 2000,
    "animationDuration": 1000,
    "animationEasing": "cubicOut",
    "animationDelay": 0,
    "animationDurationUpdate": 300,
    "animationEasingUpdate": "cubicOut",
    "animationDelayUpdate": 0,
    "series": [
        {
            "type": "bar",
[...]

PR #8162, opened 2025-09-01
Implement support for JSCode on ECharts and Tabulator 

philippjfr:
Implements https://github.com/holoviz/panel/issues/2634
Fixes https://github.com/holoviz/panel/issues/7936
Supersedes https://github.com/holoviz/panel/pull/7940


hoxbro:
I would also love to see this kind of escape hatch for the tabulator.

MarcSkovMadsen:
Yes. I would love a general escape hatch.

philippjfr:
Still has to be supported on each component explicitly but Tabulator now supports it.

PR #8163, opened 2025-09-01
test: Bump test-ui Python version

hoxbro:
Seems like a newer versions is needed to get newer versions of playwright.


hoxbro:
The test `panel/tests/ui/widgets/test_tabulator.py::test_tabulator_editors_nested` consistently fails for me both locally and on Mac CI.

PR #8172, opened 2025-09-03
test: Fix failing ui test

hoxbro:
Follow up to https://github.com/holoviz/panel/pull/8163#issuecomment-3244944384


philippjfr:
Just pushed my fix. I don't much like timeouts in tests but my change replaces the parametrization, which seems wasn't needed. I thought the problem was that selecting the current option didn't close the popup menu correctly, but that's not accurate it seems.

philippjfr (approved):

hoxbro:
Tried a lot of things to avoid the timeouts. I think it is too quick to click on the next cell when updating, causing a raised condition somewhere. Couldn't see the problem with slowmo 200.

PR #8175, opened 2025-09-04
Allow registering extension CDN

philippjfr:
Allows extensions to register a CDN alias for its dist directory, which Panel will use to resolve a resource if in CDN mode. Important for extension paths that are otherwise only available via the custom component path handler.

### DIFF
--- a/panel/io/resources.py
+++ b/panel/io/resources.py
@@ -118,6 +118,9 @@ COMPONENT_PATH = "components/"
 
 BK_PREFIX_RE = re.compile(r'\.bk\.')
 
+# Maps between extension dist directories and CDN
+EXTENSION_CDN = {}
+
 RESOURCE_URLS = {
     'font-awesome': {
         'zip': 'https://use.fontawesome.com/releases/v5.15.4/fontawesome-free-5.15.4-web.zip',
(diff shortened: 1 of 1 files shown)


panel/io/resources.py
```diff
@@ -305,6 +325,30 @@ def patch_stylesheet(stylesheet, dist_url):
 def _is_file_path(stylesheet: str)->bool:
     return stylesheet.lower().endswith(".css")
 
+def _is_subpath(path: str | Path, parent: str | Path) -> bool:
+    """
+    Return True if `path` is located inside `parent` (or equal to it).
+    Both paths are resolved (absolute, symlinks resolved).
+    """
+    path = Path(os.path.normpath(Path(path).absolute()))
+    parent = Path(os.path.normpath(Path(parent).absolute()))
```
hoxbro: Are you sure that there isn't a Pathlike function for os.path.normpath?

philippjfr: There is not, in fact the fact that I thought that `.resolve()` was equivalent caused major regressions in previous versions.

hoxbro: I would then suggest only doing one Path creation.

`Path(os.path.normpath(os.path.abspath(parent)))`

PR #8199, opened 2025-09-16
Add UTC timezone to default time for croniter

### LINKED ISSUE #8197: schedule_task default `at` should be utc datetime, not local timezone
<details>

```plaintext
panel 1.8.0
```
</details>

using `pn.state.schedule_task` with the `cron` parameter and no `at` parameter isn't firing properly (potentially just for systems with a negative time offset)  when I run the code below I would expect the task to fire each minute, but the task is firing repeatedly (my computer is in UTC-6)

If I supply the `at` parameter where the datetime.now is created as UTC the issue goes away and the task is fired each min as expected.  

I believe panel should use the UTC [time in this line](https://github.com/holoviz/panel/blob/21aa52154bbf16d7db72c4ec32e1d25d54f5a98c/panel/io/state.py#L975) to fix the issue.  I'd be happy to provide a PR to add it!

croniter_app.py: (not really needed, since all the real code that gets executed is in the setup)
```python
import panel as pn
pn.extension()

croniter_app = pn.Column(
    pn.widgets.StaticText(
        value="hello!"
    )
).servable()

```

app_setup.py: 
```python
import datetime as dt
import panel as pn

async def task():
    print(f'Task executed at: {dt.datetime.now()}')

pn.state.schedule_task(
    'task', 
    task, 
    cron="*/1 * * * *", # should fire every min
    #at=dt.datetime.now(dt.timezone.utc), # uncommenting this makes it work as expected
 )

```

panel serve croniter_app.py --setup app_setup.py

2025-09-15 18:56:08,034 Starting Bokeh server version 3.7.3 (running on Tornado 6.5.2)
2025-09-15 18:56:08,035 User authentication hooks NOT provided (default user enabled)
2025-09-15 18:56:08,036 Bokeh app running at: http://localhost:5006/croniter_app
2025-09-15 18:56:08,036 Starting Bokeh server with process id: 27700
Task executed at: 2025-09-15 18:56:08.037209
Task executed at: 2025-09-15 18:56:08.037454
Task executed at: 2025-09-15 18:56:08.037600
Task executed at: 2025-09-15 18:56:08.037698
Task executed at: 2025-09-15 18:56:08.037780
Task executed at: 2025-09-15 18:56:08.037865
Task executed at: 2025-09-15 18:56:08.037939
Task executed at: 2025-09-15 18:56:08.038015
Task executed at: 2025-09-15 18:56:08.038087
Task executed at: 2025-09-15 18:56:08.038158
Task executed at: 2025-09-15 18:56:08.038232
Task executed at: 2025-09-15 18:56:08.038306
Task executed at: 2025-09-15 18:56:08.038380
Task executed at: 2025-09-15 18:56:08.038451
Task executed at: 2025-09-15 18:56:08.038521
Task executed at: 2025-09-15 18:56:08.038592

- [ x] I may be interested in making a pull request to address this

ATL2001:
fixes #8197 

long time user, first time contributor! happy to have finally helped out ( in a very tiny way) 😊

### DIFF
--- a/panel/io/state.py
+++ b/panel/io/state.py
@@ -972,7 +972,7 @@ class _state(param.Parameterized):
             )
         else:
             from croniter import croniter
-            base = dt.datetime.now() if at is None else at
+            base = dt.datetime.now(dt.timezone.utc) if at is None else at
             diter = croniter(cron, base)
         now = dt.datetime.now().timestamp()
         try:


hoxbro:
Thank you for the quick fix. Your issue was excellent :100:

PR #8206, opened 2025-09-21
Minor CSS tweaks to the icons of the editable template

maximlt:
The few times I've tried to use the EditableTemplate it always took me a while until I could find the icons to control the grid items (my first reaction was always "hmm there's something wrong"). With these changes these icons are always visible. I've also reduced them a bit, and set the cursor.

image


maximlt:
Yes, I haven't changed that part.

philippjfr (approved):

PR #8209, opened 2025-09-21
Set cursor as pointer in the connection lost toast

maximlt:
Tiny change to set the cursor as pointer on the "Click here" link in the connection lost toast.
image


maximlt:
Pretty cool test `test_markdown_indexed` 👍 !

Issue #3555, opened 2022-05-25
OAuth module ignores URL path


TheoMathurin:
Tested with Panel 0.13.0 and some older versions, Bokeh 2.4.2.

I've been using the OAuth authentication feature provided by Panel with a custom provider for some time. It's working very well and I appreciate how easy it is to set up.

Now, I'm currently working on a new production environment where the app is behind a reverse proxy (see [recent topic](https://discourse.holoviz.org/c/panel/5) on discourse). The server is launched with something like:
```bash
panel serve myapp/ --oauth-provider provider --oauth-key key --oauth-secret secret --cookie-secret cookie-secret --oauth-encryption-key encryption-key --oauth-redirect-uri http://host/myapp --allow-websocket-origin host
```

With a random main.py in myapp/:
```python
import panel as pn

content = pn.Column('Hey')
content.servable()
```

While trying to make it work, I realized that the auth.py module has some limitations regarding the management of URLs and redirection. Namely, the process fails if the app is available at a specific path. This is typically the case in a reverse proxy setup, but  also if you use a prefix.

First, if you look at the [latest version](https://github.com/holoviz/panel/blob/978f2915e5517ac693d356b42cff9abb6700521b/panel/auth.py) of the file, [line 752](https://github.com/holoviz/panel/blob/978f2915e5517ac693d356b42cff9abb6700521b/panel/auth.py#L752) there is a return statement that builds the wrong URL. Trying to access http://host/myapp, the URL will change to http://host/login?next=%2Fmyapp, _i.e._ the path is ignored. This yields a 404 Error.

I've found that this can be fixed with the following change to the `login_url` method:
```python
def login_url(self):
    if config.oauth_redirect_uri is None:
        return '/login'
    else:
        return urlparse(config.oauth_redirect_uri).path + '/login'
```

This will get the user to the provider page. Then, the path is also ignored in the redirection (`self.redirect` method call [line 269](https://github.com/holoviz/panel/blob/978f2915e5517ac693d356b42cff9abb6700521b/panel/auth.py#L269)) so that after the user has been authorized, Tornado redirects to http://host/ instead of http://host/myapp. Passing `urlparse(redirect_uri).path` instead of `"/"` solves the issue.

Note that the same issue arises without reverse proxying if you use a prefix, although the fix I put in place does not seem to work for some reason. In other words, you won't be able to set up authentication in this configuration:
```bash
[...]

PR #8211, opened 2025-09-22
Fixes for OAuth via proxy

TheoMathurin:
The goal is to address issues mentioned in #3555

For starters, I believe the `test_auth0_oauth_via_proxy` test function does not actually test the proxy mechanism because the page is accessed via the wrong port, bypassing the reverse proxy.

Let's see if the test still passes (I don't want to create an Auth0 account to get a key so I can't check it locally).

### DIFF
--- a/panel/tests/ui/test_auth.py
+++ b/panel/tests/ui/test_auth.py
@@ -196,8 +196,8 @@ def test_auth0_oauth_via_proxy(py_file, page):
             py_file.name
         ]
         with run_panel_serve(cmd) as p:
-            port = wait_for_port(p.stdout)
-            page.goto(f"http://localhost:{port}")
+            wait_for_port(p.stdout)
+            page.goto(f"http://localhost:{proxy}/proxy/{app_name}")
 
             page.locator('input[name="username"]').fill(auth0_user)
             page.locator('input[name="password"]').fill(auth0_password)
--- a/panel/tests/ui/test_auth.py
+++ b/panel/tests/ui/test_auth.py
@@ -197,7 +197,7 @@ def test_auth0_oauth_via_proxy(py_file, page):
         ]
         with run_panel_serve(cmd) as p:
             wait_for_port(p.stdout)
-            page.goto(f"http://localhost:{proxy}/proxy/{app_name}")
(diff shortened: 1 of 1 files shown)


panel/tests/ui/test_auth.py
```diff
@@ -196,8 +196,8 @@ def test_auth0_oauth_via_proxy(py_file, page):
             py_file.name
         ]
         with run_panel_serve(cmd) as p:
-            port = wait_for_port(p.stdout)
-            page.goto(f"http://localhost:{port}")
+            wait_for_port(p.stdout)
+            page.goto(f"http://localhost:{proxy}/proxy/{app_name}")
```
philippjfr:
```suggestion
            page.goto(f"http://localhost:{proxy}/proxy/")
```

TheoMathurin:
So the test passes, which based on my own real-world tests just means that the test is flawed. It would be great if someone could reproduce the issue independently though.

NB: The test failure is unrelated.

philippjfr:
It would be good to figure out in what way the test is flawed. My suspicion is that the test correctly configures the proxy in a way that's a) not documented and b) isn't the case in your setup. What I suspect is making it work is this:

```
            {"handler": "rewrite", "strip_path_prefix": "/proxy"},
```

which strips the `/proxy` prefix when it forwards the request.

TheoMathurin:
Indeed with this reverse proxy config, the issue cannot occur since the application is accessed at a path that corresponds to the application name (/proxy/app_name rewritten as /app_name).

The configuration could be modified to make the test more general. I will experiment and come up with a configuration proposal against which basic auth and OAuth modules should be tested.

Edit: it may not be straightforward, because ideally, the reverse proxy rule should know of the application name, whereas currently it is set up before it could be known based on the temporary file name.

TheoMathurin:
I believe that [here](https://github.com/holoviz/panel/blob/b4ddd08c045a6c4c7eb0881038907604b0983f7d/panel/auth.py#L124) we should replace the base path of the app (i.e. `/prefix/app_name`) with the path used in the reverse proxy. This is how I fix it for my app anyway.

This could be controlled by adding a new environment variable, something like `PANEL_REVERSE_PROXY_PATH`.

Issue #7429, opened 2024-10-22
Off-line / air-gapped Panel: implement testing


Coderambling:
Off-line / air-gapped use cases and related requests pop up regularly.

It would make sense to implement testing for this (using reference apps?), to ensure this works well, and keeps working as new features are added.

As a start, is it posdible to run the current tests, but with internet access disabled? Can this be done using Github Actions?

Coderambling:
Related: https://github.com/holoviz/panel/issues/4257

https://discord.com/channels/1075331058024861767/1113741778496008232/1256570697174745110

General offline implementations: https://offlinefirst.org/

Coderambling:
Related? https://github.com/holoviz/panel/pull/7826

Coderambling:
Related: https://github.com/holoviz/panel/issues/4077

https://github.com/holoviz/panel/issues/7965#issuecomment-2942794662

Discord thread: https://discord.com/channels/1075331058024861767/1390256657497587756

mikeymckay:
You can test offline mode in chrome by opening the developer console, clicking on Network, Throttling and selecting offline. This blocks all outgoing requests for the page. There is a whole community dedicated to offline apps, which are related to progressive web apps (PWAs): https://offlinefirst.org/

mikeymckay:
And note that there may be a workaround/solution found in the comments of: https://github.com/holoviz/panel/issues/4077

Coderambling:
Right. Interesting link!

What I was thinking went I put the sentence below in this issue was:

The Panel library has automated tests on Github, that tests components etc. to catch issues for new releases etc.

So:

1. Copy the tests (or create a new repo for this)?
2. Make it so the tests have access to local resources, but prevent them from going outside the repo to CDN's etc.
3. Look at what the tests report.

So basically simulate offline use of Panel in this way, leveraging the exising tests. That could clarify what is needed to make offline use possible.

I don't know how viable / involved this would be, but that coulld be an approach to identify gaps / issues.


> As a start, is it possible to run the current tests, but with internet access disabled? Can this be done using Github Actions?

PR #8215, opened 2025-09-22
Improve Vega Lite to support validation

### LINKED ISSUE #8201: Confusing Vegalite error without $schema part of the spec.
```python
import panel as pn

vegalite = {
  "data": {"url": "https://raw.githubusercontent.com/vega/vega/master/docs/data/barley.json"},
  "mark": "bar",
  "encoding": {
    "x": {"aggregate": "sum", "field": "yield", "type": "quantitative"},
    "y": {"field": "variety", "type": "nominal"},
    "color": {"field": "site", "type": "nominal"}
  }
}
vgl_pane = pn.pane.Vega(vegalite, height=240)
vgl_pane
```

`ValueError: Vega pane does not support objects of type 'dict'.`

ahuang11:
Migrate Vega validation from Lumen https://github.com/holoviz/lumen/blob/main/lumen/ai/views.py#L245-L262 into Panel.

Also handles archaic error:
Closes https://github.com/holoviz/panel/issues/8201


panel/pane/vega.py
```diff
@@ -167,18 +198,95 @@ class Vega(ModelPane):
...
     def __init__(self, object=None, **params):
+        if isinstance(object, dict) and "$schema" not in object:
+            self.param.warning(f"No $schema found; using {SCHEMA_URL} by default. Specify the schema explicitly to avoid this warning.")
+            object["$schema"] = SCHEMA_URL
         super().__init__(object, **params)
         self.param.watch(self._update_selections, ['object'])
         self._update_selections()
 
+    @cache
+    def _download_vega_lite_schema(self, schema_url: str | None = None):
```
MarcSkovMadsen: What about airgapped usage of panel? Are we not trying to support that?

ahuang11: `_download_vega_lite_schema` is optional--only for validation.

```python
            try:
                schema = self._download_vega_lite_schema(schema)
            except Exception as e:
                self.param.warning(f"Skipping validation because could not load Vega schema at {schema}: {e}")
                return
```

And users have full flexibility on using a local schema, meaning it can do something like: `requests.get("localhost:8000/schema")`?

However, potentially I can do `open("schema").read()`

ahuang11: Nevermind, `open("schema").read()` isn't feasible because I don't think the JS can access it, but if user exposes it as an endpoint, it's likely doable.

```
import json
import requests

with open("schema.json", "w") as f:
    json_file = requests.get("https://vega.github.io/schema/vega-lite/v5.json").json()
    json.dump(json_file, f)

import panel as pn
pn.extension('vega')

vegalite = {
  "$schema": "schema.json",
  "data": {"url": "https://raw.githubusercontent.com/vega/vega/master/docs/data/barley.json"},
  "mark": "bar",
  "encoding": {
    "x": {"aggregate": "sum", "field": "yield", "type": "quantitative"},
    "y": {"field": "variety", "type": "nominal"},
    "color": {"field": "site", "type": "nominal"}
  }
}
vgl_pane = pn.pane.Vega(vegalite, height=240)
vgl_pane.show()
```

panel/pane/vega.py
```diff
@@ -167,18 +197,102 @@ class Vega(ModelPane):
...
+        if isinstance(object, dict) and "$schema" not in object:
+            self.param.warning(f"No $schema found; using {SCHEMA_URL} by default. Specify the schema explicitly to avoid this warning.")
+            object["$schema"] = SCHEMA_URL
         super().__init__(object, **params)
         self.param.watch(self._update_selections, ['object'])
         self._update_selections()
 
+    @cache
+    def _download_vega_lite_schema(self, schema_url: str | bytes) -> dict:
+        return requests.get(schema_url).json()
```
philippjfr: Fetching the schema can be slow. My machine sometimes even has a hangup where `requests` doesn't work. At absolute minimum we need a timeout here.

panel/pane/vega.py
```diff
@@ -167,18 +197,103 @@ class Vega(ModelPane):
...
+        if isinstance(object, dict) and "$schema" not in object:
+            self.param.warning(f"No $schema found; using {SCHEMA_URL} by default. Specify the schema explicitly to avoid this warning.")
+            object["$schema"] = SCHEMA_URL
         super().__init__(object, **params)
         self.param.watch(self._update_selections, ['object'])
         self._update_selections()
 
+    @cache
+    def _download_vega_lite_schema(self, schema_url: str | bytes) -> dict:
+        response = requests.get(schema_url, timeout=5)
```
hoxbro, line 281: Perhaps consider caching the response to a file, so you don't need to download it each time. I will likely use `platformdirs.cache_dirs`

panel/pane/vega.py
philippjfr: At minimum this needs a timeout.

Coderambling: > Nevermind, `open("schema").read()` isn't feasible because I don't think the JS can access it, but if user exposes it as an endpoint, it's likely doable.
>

Should a remark / example regarding air-gapped use be added to the Vega doc after this update?

Looks like that info is currently not part of the doc.

https://panel.holoviz.org/reference/panes/Vega.html

Coderambling: Related regarding air-gapped use: https://github.com/holoviz/panel/issues/7429

panel/pane/vega.py
```diff
@@ -167,18 +198,103 @@ class Vega(ModelPane):
...
 
     def __init__(self, object=None, **params):
+        if isinstance(object, dict) and "$schema" not in object:
+            self.param.warning(f"No $schema found; using {SCHEMA_URL} by default. Specify the schema explicitly to avoid this warning.")
+            object["$schema"] = SCHEMA_URL
         super().__init__(object, **params)
         self.param.watch(self._update_selections, ['object'])
         self._update_selections()
 
+    @cache(to_disk=True, cache_path=platformdirs.user_cache_dir("vega", "panel"))
```
philippjfr:
```suggestion
    @cache(to_disk=True, cache_path=platformdirs.user_cache_dir("panel"))
```

panel/pane/vega.py
```diff
@@ -167,18 +198,103 @@ class Vega(ModelPane):
...
+        process_error(error)
+        return "\n".join(errors.values())
+
+    @param.depends("object", watch=True, on_init=True)
+    def _validate_object(self):
+        try:
+            from jsonschema import (  # type: ignore[import-untyped]
+                Draft7Validator, ValidationError,
+            )
+        except ImportError:
```
philippjfr: This should probably warn.

ahuang11:
I think using https://github.com/vega/vl-convert to validate is a better solution, like https://github.com/holoviz/lumen/blob/main/lumen/ai/editors.py#L343-L344

PR #8218, opened 2025-09-25
Avoid panel helper function warnings

philippjfr:
~~In panel 1.8.0 we started warning when the `panel` helper function was used on an object that is already a `Viewable` but you supplied kwargs~~. Unfortunately we were a little indiscriminate here and the warnings could be generated by Panel itself. This PR restricts precisely when this warning is issued and skips calling `panel` when not necessary.


philippjfr:
Turns out I hadn't added this warning and it was actually just sitting in my stash, and what I discovered is that it raises too many false alarms. Closing.

PR #8222, opened 2025-09-27
Serve index.html when serving static directory

etihwo:
From the example described in [Serving static files](https://panel.holoviz.org/how_to/server/static_files.html), when user go to `http://localhost:5006/assets/` it will serve the index.html file inside the folder. Otherwise only the address `http://localhost:5006/assets/index.html` is valid


philippjfr:
Happy with this! Adding a quick test would be good, probably just in `panel/tests/test_server.py`.

PR #8236, opened 2025-10-06
compat: Remove direct dependency on pandas 

hoxbro:
Really early. This is to test-run our testsuite without having Pandas as a direct dependency.


philippjfr:
Great start, though we need to work on removing the dependency in bokeh first.

panel/tests/io/test_location.py
```diff
@@ -178,12 +181,14 @@ def test_location_sync_from_dataframe(location, dataframe):
     assert location.search == "?dataframe=%5B%7B%22x%22%3A+1%7D%5D"
 
 def test_location_sync_to_dataframe(location, dataframe):
+    pd = pytest.importorskip("pandas")
```
philippjfr: Can we convert these to decorators, i.e.:

```
pd_available = pytest.mark.skipif(pd is None, reason="requires pandas")
```

hoxbro: Sure, will do this before a merge. For now this is easier to do.

PR #8240, opened 2025-10-08
Minor enhancements to the gallery deployment

maximlt:
- Only attempt to deploy on non-schedule events
- Add a step the prepare the project, to ensure it solves before attempting to deploy it
- Bump the pins


examples/gallery/anaconda-project.yml
```diff
@@ -22,31 +22,30 @@ variables:
 
 packages: &pkgs
   - python >=3.12
```
hoxbro, line 24:
```suggestion
  - python >=3.13,<3.14
```
Maybe just pin to a Python version to reduce solve space and causing slow/big solve. 

Not sure about the right syntax

hoxbro (commented):
Have one comment otherwise LGTM

examples/gallery/anaconda-project.yml
maximlt: The solve I did locally used Python 3.12, that's way I didn't bump Python.

hoxbro:
```suggestion
  - python >=3.12,<3.13
```

Would also be ok. I just wanted to reduce the solve going down a suboptimal branch by limiting the Python version, but you can merge if you don't think it will be a problem.

maximlt: I'd like Python 3.13 to be picked up automatically when the solver allows for it. I will 100% forget to bump that pin to check that, so I prefer to keep it that way.

PR #8241, opened 2025-10-08
compat: Python 3.14

### LINKED ISSUE #8239: Python 3.14 support
Just making this issue to keep track of it. But the following lines will now emit a warning. 

https://github.com/holoviz/panel/blob/6e8479f502fa9240c0aa0f0188fa66bcf24308ab/panel/io/server.py#L681-L689

Image

The if statement likely just need to be for lower than 3.14 versions:

Image

hoxbro:
Fixes #8239


pixi.toml
```diff
@@ -153,6 +204,34 @@ reacton = "*"
...
+# Optional dependencies - recommended
+holoviews = ">=1.16.0"
+jupyterlab = "*"
+matplotlib-base = "*"
+pillow = "*"
+plotly = ">=6.0"
+# Optional dependencies - dev
+watchfiles = "*"
+# Test dependencies
+# altair = "*"
```
hoxbro, line 218: Needs a new release:

image

panel/io/server.py
```diff
@@ -678,8 +678,11 @@ def create_static_handler(prefix, key, app):
 # Bokeh 2.4.x patches the asyncio event loop policy but Tornado 6.1
 # support the WindowsProactorEventLoopPolicy so we restore it,
 # unless we detect we are running on jupyter_server.
+# get_event_loop_policy, WindowsProactorEventLoopPolicy, WindowsProactorEventLoopPolicy
+# is deprecated in Python 3.14.
 if (
     sys.platform == 'win32' and
+    sys.version_info < (3, 14, 0) and
```
mattpap, line 685: > class asyncio.DefaultEventLoopPolicy
> The default asyncio policy. Uses [SelectorEventLoop](https://docs.python.org/3.14/library/asyncio-eventloop.html#asyncio.SelectorEventLoop) on Unix and [ProactorEventLoop](https://docs.python.org/3.14/library/asyncio-eventloop.html#asyncio.ProactorEventLoop) on Windows.

The loop policy stays the same, so this doesn't fix the problem. I also checked that Tornado 6.x didn't get support for `ProactorEventLoopPolicy`. In bokeh for now I ignored the deprecation altogether, pending further discussion how to resolve this.

mattpap: Or maybe I'm misinterpreting this and policies are irrelevant starting with 3.14?

pixi.toml
hoxbro: Greater than 5.5.0, the main branch has a fix for this as far as I can see.

pixi.toml
```diff
@@ -110,6 +116,52 @@ vtk = "*"
...
+pygraphviz = "*"
+pyinstrument = ">=4.0"
+python-graphviz = "*"
+# pyvista = "*"  # https://github.com/conda-forge/pyvista-feedstock/issues/115
+reacton = "*"
+s3fs = "*"
+scikit-image = "*"
+scikit-learn = "*"
+seaborn = "*"
+# streamz = "*"  # raises AttributeError: 'OrderedSet' object has no attribute 'copy'
```
hoxbro, line 157: image

panel/io/server.py
hoxbro: Mateusz and I discussed this yesterday. Bokeh no longer makes any changes to the event loop policy, as they bumped the tornado dependency to 6.2 some time ago ([ref](https://github.com/bokeh/bokeh/pull/13635)).

https://github.com/bokeh/bokeh/blob/56d8cbbbd08aacbda51da0816371194820374418/src/bokeh/util/tornado.py#L61-L71

Perhaps we should remove this if statement altogether?

pixi.toml
hoxbro: And getting a RecursionError when we exit the repl

image

pyproject.toml
```diff
@@ -37,7 +38,6 @@ classifiers = [
     "Intended Audience :: Information Technology",
     "Intended Audience :: Legal Industry",
     "Intended Audience :: Other Audience",
-    "Intended Audience :: Science/Research",
```
hoxbro, line 40: Duplicated of line 35

pixi.toml
```diff
@@ -110,6 +116,52 @@ vtk = "*"
...
+ipywidgets = "*"
+ipywidgets_bokeh = "*"
+jupyter_bokeh = ">=3.0.7"
+networkx = ">=2.5"
+# pydeck = ">=0.8"  # Add back when it support ipywidgets 8.0
+pyecharts = "*"
+pygraphviz = "*"
+pyinstrument = ">=4.0"
+python-graphviz = "*"
+# pyvista = "*"  # https://github.com/conda-forge/pyvista-feedstock/issues/115
```
hoxbro:
```suggestion
pyvista = "*"
```

pixi.toml
```diff
@@ -110,6 +116,52 @@ vtk = "*"
...
+ipywidgets = "*"
+ipywidgets_bokeh = "*"
+jupyter_bokeh = ">=3.0.7"
+networkx = ">=2.5"
+# pydeck = ">=0.8"  # Add back when it support ipywidgets 8.0
+pyecharts = "*"
+pygraphviz = "*"
+pyinstrument = ">=4.0"
+python-graphviz = "*"
+pyvista = "*"
```
hoxbro:
```suggestion
# pyvista = "*"  https://github.com/conda-forge/pyvista-feedstock/issues/116
```

pixi.toml
philippjfr: Can we instead skip the altair tests for 3.14 for now.

philippjfr: Oh nvm, this is 3.14 specific.

PR #8249, opened 2025-10-14
Subclass FunctionHandler to ensure it behaves like other handlers

philippjfr:
We use custom bokeh handlers to allow us add our custom handling such as setting the curdoc, profiling, configuring session re-connects, track launching sessions etc. Until now function based apps bypassed all this meaning that they did not behave like other apps.


philippjfr:
Test failures are unrelated. I'll open a new PR to fix that.

PR #8274, opened 2025-10-29
Update card collapsed state handling to change child view model visibility attribute

### LINKED ISSUE #8198: Card callback not called when collapsed by default
I am using panel 1.7.5, bokeh 3.7.3 and ipywidgets-bokeh 1.6.0 and notice a weird issue with cards collapsed by default when using with ipywidgets.

When I have the card collapsed by default, the widgets are not updated correctly see gif 1 but when I expand the card explicitly after the application load, widgets are updated as expected see gif 2.

GIF1

Image

GIF2

Image

Related code for reproducing the issue

```
import pandas as pd
import panel as pn
import param
import ipywidgets as widgets

# Configuration
TITLE = "Simple Panel Dashboard"

pn.extension()

# Sample data
data = pd.DataFrame({
    "Name": ["Alice", "Bob", "Cathy"],
    "City": ["NY", "LA", "NY"]
})

# Column names
NAME_COL = "Name"
CITY_COL = "City"
class TestDash(param.Parameterized):

    city = param.ObjectSelector()

    def __init__(self, df):
        super().__init__()
        self.df = df
        
        # Initialize cards as None - will be created in build_widgets
        self.card_1 = None
        self.card_2 = None
        
        # Status indicator for card collapsed state
        self.card_status_indicator = None

        self.build_widgets()


    def build_widgets(self):
        """Builds all panel widgets."""
        cities = ["All"] + list(self.df[CITY_COL].unique())
        self.filter_options_dict = {CITY_COL: cities}

        self.city_filter = pn.widgets.Select.from_param(
            self.param.city, name="City", options=cities, value="All",
            sizing_mode="stretch_width"
        )

        self.selectors_dict = {CITY_COL: self.city_filter}
        self.selectors_dependents_dict = {CITY_COL: []}
        
    #_________________________________________________________________________#
    #SECTION - Data Filtering
    def filter_data(self, filter_on_dict):

        df = self.df.copy()
        for col, widget in filter_on_dict.items():
            if widget.value != "All":
                # only filter if widget is not set to "everything"
                df = df[df[col] == widget.value]

        return df


    def filter_display_data(self) -> pd.DataFrame:
        # Filter using all selectors in the dashboard
        df = self.filter_data(self.selectors_dict)

        # Save the result as an instance variable for use elsewhere
        self.filtered_df = df.reset_index(drop=True).copy()

        return df

    #!SECTION - END Data Filtering
    # Object Builders

    def contain_objects(self):

        self.build_widgets_content()
        
        # Create cards if they don't exist yet
        if self.card_1 is None:
            self.card_1 = pn.Card(
                self.widget_card_1,
                styles={"background": "WhiteSmoke"},
                width=400,
                title="Card 1 - Selector Values",
                collapsed=True
            )
            # Watch the collapsed property of the first card
            self.card_1.param.watch(self._on_card_collapsed, ['collapsed'])
        else:
            # Update the content of existing card
            self.card_1.objects = [self.widget_card_1]
            
        if self.card_2 is None:
            self.card_2 = pn.Card(
                self.widget_card_2,
                styles={"background": "LightBlue"},
                width=400,
                title="Card 2 - Filtered Data",
                collapsed=False
            )
        else:
            # Update the content of existing card
            self.card_2.objects = [self.widget_card_2]
            
        # Create status indicator if it doesn't exist
        if self.card_status_indicator is None:
            self.card_status_indicator = pn.pane.HTML(
                self._get_card_status_html(),
                styles={"background": "#f0f0f0", "padding": "10px", "border-radius": "5px"}
            )

        return [
            # Row with city filter
            pn.Row(
                pn.pane.HTML("<h3>Filter:</h3>"),
                self.city_filter,
                sizing_mode="stretch_width"
            ),
            # Status indicator showing card state
            pn.Row(
                pn.pane.HTML("<h4>Card Status:</h4>"),
                self.card_status_indicator,
                sizing_mode="stretch_width"
            ),
            # Row for cards
            pn.Row(
                self.card_1,
                self.card_2,
                sizing_mode="stretch_width"
            )
        ]

    def build_widgets_content(self):
        """Build ipywidgets content for the cards."""
        
        # Card 1: Display selector values
        selector_values = {}
        for key, widget in self.selectors_dict.items():
            selector_values[key] = widget.value
        
        display_text = "<h4>Current Filter Settings:</h4><ul>"
        for key, value in selector_values.items():
            display_text += f"<li><strong>{key}:</strong> {value}</li>"
        display_text += "</ul>"
        
        self.widget_card_1 = widgets.HTML(value=display_text)
        
        data_text = ''
        
        if hasattr(self, 'filtered_df') and not self.filtered_df.empty:

            data_text += self.filtered_df.to_html(index=False, classes='table table-striped')
        
        self.widget_card_2 = widgets.HTML(value=data_text)

    def _get_card_status_html(self):
        """Generate HTML for the card status indicator."""
        if self.card_1 is None:
            status = "Card not initialized"
            color = "#999"
            icon = "⚪"
        elif self.card_1.collapsed:
            status = "Card 1 is COLLAPSED 📦"
            color = "#ff6b6b"
            icon = "🔒"
        else:
            status = "Card 1 is EXPANDED 📂"
            color = "#51cf66"
            icon = "🔓"
            
        return f"""
        <div style="display: flex; align-items: center; gap: 10px;">
            <span style="font-size: 20px;">{icon}</span>
            <span style="color: {color}; font-weight: bold; font-size: 16px;">{status}</span>
        </div>
        """
    
    def _on_card_collapsed(self, event):
        """Callback function that gets triggered when card collapsed state changes."""
        print(f"Card collapsed state changed! New value: {event.new}")
        
        # Update the status indicator
        if self.card_status_indicator is not None:
            self.card_status_indicator.object = self._get_card_status_html()
        
        # You can add any other logic here based on the collapsed state
        if event.new:  # Card is now collapsed
            print("Card 1 was collapsed - you can trigger any action here!")
            # Example: Change card 2 background when card 1 is collapsed
            if self.card_2 is not None:
                self.card_2.styles = {"background": "#ffcccc"}
        else:  # Card is now expanded
            print("Card 1 was expanded - you can trigger any action here!")
            # Example: Restore card 2 background when card 1 is expanded
            if self.card_2 is not None:
                self.card_2.styles = {"background": "LightBlue"}

    # Callbacks
    @param.depends(
        "city",
        watch = True
    )
    def update_dashboard(self, *events):

        if not self.dashboard.objects and not events:
            #NOTE - This needs to stay, and stops the flickering of objects when
            #       widgets were cleared.
            return

        self.dashboard.loading=True

        # Update Widgets before filtering data
        self.filter_display_data()

        self.dashboard.clear()

        self.dashboard.extend(self.contain_objects())

        self._updating=False
        self.dashboard.loading=False


    #SECTION - Building Containers
    #ANCHOR - Main Container
    def __panel__(self,):

        if hasattr(self,'dashboard'):
            return self.dashboard

        self.dashboard = pn.Column(
            pn.pane.HTML(f"<h1>{TITLE}</h1>"),
            sizing_mode="stretch_width"
        )

        self.filter_display_data()

        objects = self.contain_objects()
        self.dashboard.extend(objects)

        return self.dashboard

# Create and serve the app
def create_app():
    """Create the simple Panel application."""
    dashboard = TestDash(data)
    return dashboard.__panel__()

# Run the app
    create_app().servable()
elif __name__ == "__main__":
    # For local development
    app = create_app()
    app.show(port=5007
```)

Can you please suggest a fix here or fix in upstream if needed?

### PR DESCRIPTION
**Pull request by dalthviz**
For an initial description of the approach followed for the proposed change you can check https://github.com/holoviz/panel/issues/8198#issuecomment-3437967199

Fixes #8198

### CODE REVIEW

File: panel/models/card.ts [hunk 1]
```diff
@@ -106,14 +106,11 @@ export class CardView extends ColumnView {
...
-    if (this.model.collapsed) {
-      return
-    }
-
     for (const child_view of this.child_views.slice(1)) {
-      this.shadow_el.appendChild(child_view.el)
       child_view.render()
       child_view.r_after_render()
+      this.shadow_el.appendChild(child_view.el)
+      child_view.model.visible = !this.model.collapsed
```
philippjfr: Many components require the element to be attached to the DOM before rendering:

```suggestion
      child_view.model.visible = child_view.model.visible && !this.model.collapsed
      this.shadow_el.appendChild(child_view.el)
      child_view.render()
      child_view.r_after_render()
```

panel/models/card.ts
```diff
@@ -107,10 +107,10 @@ export class CardView extends ColumnView {
     }
 
     for (const child_view of this.child_views.slice(1)) {
+      child_view.model.visible = child_view.model.visible && !this.model.collapsed
```
philippjfr: Hmm, I guess this is still problematic, because once visible=False it won't be restored to true subsequently.

dalthviz: Maybe setting the visibility here to `false` should only be done once using some flag variable like `initially_colapsed`? So initializing at some point:

```ts
this.initially_collapsed = this.model.collapsed
```

and then changing these lines to something like:

```ts
     if (this.initially_collapsed) {
        child_view.model.visible = false
        this.initially_collapsed = false
      }
```

With that seems to me that the only logic that would change the model visible attribute after that initial collapsed setup is inside the `_collapse` function that gets triggered when the model `collapsed` property changes

dalthviz:
Note: Closing and reopening to check if a CI rerun helps with the failing `test_server_thread_pool_change_event[tornado]` test

philippjfr:
That one is flaky and unrelated.

dalthviz:
Yep, in my last commit I added a `flaky` decorator to that test and CI now is green. Let me know if there is something else that should be done here or if that last change I did should be reverted/moved to a different PR!

panel/models/card.ts
```diff
@@ -106,11 +108,11 @@ export class CardView extends ColumnView {
       header.r_after_render()
     }
 
-    if (this.model.collapsed) {
-      return
-    }
-
     for (const child_view of this.child_views.slice(1)) {
+      if (this.initially_collapsed) {
```
philippjfr: Not sure I entirely follow this. Won't this just set `visible=False` on the first child?

Also I forget how things work, but do we run `render()` every time the children are updated?

dalthviz: > Won't this just set visible=False on the first child?

Ah yes, I think you are right! So then something else should be done 🤔 

> do we run render() every time the children are updated?

Yep, the cards widget seems like works like that (at least the `update_children` definition override is calling `render` from what I can see):

https://github.com/holoviz/panel/blob/42b09569a517a428f34d6ce0b8f01143246c8f84/panel/models/card.ts#L120-L124

dalthviz: Could it be that what is needed inside the `render` function is just to keep in sync things inside the `for` with something like:

```typescript
    child_view.model.visible = !this.model.collapsed
```

?

dalthviz: Latest commit makes the change mentioned above. Seems like things are passing with it but if you think of a different approach or something else is missing let me know!

dalthviz:
Note: Closing and reopening to retrigger CI

panel/tests/ui/layout/test_card.py
```diff
@@ -201,3 +201,22 @@ def test_card_widget_not_collapsed(page, card_components):
...
+    card = Card(w1, header=Row(w2), collapsed=True)
+
+    serve_component(page, card)
+
+    text_input = page.locator('.bk-Slider')
+    expect(text_input).not_to_be_visible()
+
+    w1.visible = True
+
+    expect(text_input).not_to_be_visible()
```
dalthviz: Seems like the assert here should be `to_be_visible` since `visible=True`?

```suggestion
    w1.visible = True

    expect(text_input).to_be_visible()
```

philippjfr: No it's correct, since `collapsed=True`.

philippjfr: Meant to comment, the test was meant to highlight that the solution here is not sufficient.

dalthviz: I think the new changes added (https://github.com/holoviz/panel/pull/8274/commits/7efff7a2ebb968c3ddef20fa4591700b4819454d) should take care of that case (the new test passes with them).

Also, while doing the changes mentioned above, noticed that another test was failing (`panel/tests/ui/widgets/test_texteditor.py::test_texteditor_regression_copy_paste[chromium]`). It seems unrelated to the changes here so I tried adding a `flaky` decorator to it.

Let me know if something else is needed or if there is a better approach to take care of the case the new test illustrates. Let me know too if I should remove the flaky decorators that I added to some tests (`panel/tests/ui/widgets/test_texteditor.py::test_texteditor_regression_copy_paste` and `panel/tests/test_server.py::test_server_thread_pool_change_event`)

panel/models/card.ts
```diff
@@ -40,6 +40,13 @@ export class CardView extends ColumnView {
       this.child_views[0].el.style.backgroundColor = header_background
       this.header_el.style.backgroundColor = header_background
     })
+
+    for (const child_view of this.child_views.slice(1)) {
+      const {visible} = child_view.model.properties
+      this.on_change(visible, () => {
+        child_view.model.visible = !this.model.collapsed
+      })
```
philippjfr: Sorry for the long back and forth on this. I suspect this needs to be changed though, right now it'll add a new callback every time the component is rendered and never removed which means that you end up accumulating a bunch. Will have to dig in to figure out a good way to handle this.

dalthviz: No problem and thank you for all your feedback here! Thinking about a way to overcome that, maybe adding the logic above + some logic to disconnect remove old children related callbacks (if available) inside `update_children` could work ? 🤔 Anyhow, let me know if you find a better way to handle things!

dalthviz: Just to be sure, would it make sense for me to try the idea above (moving callback connections to the `update_children`) here or would be preferable to leave this PR as it is for the moment? Let me know!

philippjfr: Sorry I didn't follow I'll have a quick go at fixing this since you've been so patient and it would be great to just get this done.

dalthviz: No problem and thank you for checking this again!

panel/tests/ui/layout/test_card.py
philippjfr: Yeah those are just flaky, making that change was totally fine.

PR #8277, opened 2025-10-30
Enhance developer experience with the setup-dev command

maximlt:
As done in other HoloViz repos, as we found out the best way for contributors to get started is to have a single command (`pixi run setup-dev`) to get started, and to have a single `default` environment with pretty much everything they need to contribute (test deps, ui deps, docs deps, etc.), environment in which panel is already installed in editable mode and that has pre-commit already set up.

### DIFF
--- a/doc/developer_guide/index.md
+++ b/doc/developer_guide/index.md
@@ -33,8 +33,6 @@ To contribute to Panel, you will also need [Github account](https://github.com/j
 
 Developing all aspects of Panel requires a wide range of packages in different environments. To make this more manageable, Pixi manages the developer experience. To install Pixi, follow [this guide](https://pixi.sh/latest/#installation).
 
-:::
-
 :::{admonition} Note
 :class: info
 


maximlt:
I ran the full test CI and the jobs all passed (https://github.com/holoviz/panel/actions/runs/18968966032):

image

pixi.toml
```diff
@@ -1,4 +1,4 @@
-[project]
+[workspace]
```
hoxbro, line 1: Maybe we should add something like this https://github.com/holoviz/holoviews/pull/6699/files to the documentation.

doc/developer_guide/index.md
```diff
@@ -74,13 +74,18 @@ git push --tags
 
 ## Start developing
 
-To start developing, run the following command
+To start developing, run the following command, this will create an environment called `default` (in `.pixi/envs`), install Panel in [editable mode](https://pip.pypa.io/en/stable/topics/local-project-installs/#editable-installs), and install `pre-commit`:
```
hoxbro: Maybe one of these:

```suggestion
To start developing, run the following command, this will create an environment called `default` (in `.pixi/envs`), install Panel in [editable mode](https://pip.pypa.io/en/stable/topics/local-project-installs/#editable-installs), and install `pre-commit` hooks:
```

```suggestion
To start developing, run the following command, this will create an environment called `default` (in `.pixi/envs`), install Panel in [editable mode](https://pip.pypa.io/en/stable/topics/local-project-installs/#editable-installs), and install `pre-commit` git hooks:
```

```suggestion
To start developing, run the following command, this will create an environment called `default` (in `.pixi/envs`), install Panel in [editable mode](https://pip.pypa.io/en/stable/topics/local-project-installs/#editable-installs), and install `pre-commit` [git hooks](https://git-scm.com/book/en/v2/Customizing-Git-Git-Hooks):
```

hoxbro (approved):
Some small comments otherwise LGTM

pixi.toml
maximlt: Added in https://github.com/holoviz/panel/pull/8277/commits/8941b6a183a985a6d535d24e6057562cbf70b893

I don't know about you, but I'm getting a little tired about updating the developer docs the same way in every repo 🙃

hoxbro: I imagine this was how pyctdev started.

hoxbro (approved):

PR #8278, opened 2025-10-31
Introduce and use _descendents instead of param.concrete_descendents

maximlt:
The `concrete_descendents` utility offered by Param returns a dict of class name to class object. Unfortunately, this means that it clobbers classes on their name. Param intended to be released with a warning emitted by this function when this happens, suggesting users to use instead `descendents(cls, concrete=True)`. Testing this from a local branch, we found out that Panel was suffering from this clobbering in a few places, probably causing hard-to-find bugs in Panel. This PR fixes that by directly adding the `descendents` function code to Panel's utility module, and replacing all usage of `concrete_descendents`.

### DIFF
--- a/panel/util/__init__.py
+++ b/panel/util/__init__.py
@@ -563,7 +563,7 @@ def _descendents(class_: type, concrete: bool = False) -> list[type]:
     if not isinstance(class_, type):
         raise TypeError(f"descendents expected a class object, not {type(class_).__name__}")
     q = [class_]
-    out = []
+    out: list[type] = []
     while len(q):
         x = q.pop(0)
         out.insert(0, x)


panel/util/__init__.py
```diff
@@ -544,3 +545,35 @@ def camel_to_kebab(name):
...
+        return False
+
+
+def _descendents(class_: type, concrete: bool = False) -> list[type]:
+    """Added from Param to be used instead of concrete_descendents that clobber
+    class names."""
+    if not isinstance(class_, type):
+        raise TypeError(f"descendents expected a class object, not {type(class_).__name__}")
+    q = [class_]
+    out = []
```
philippjfr:
```suggestion
    out: list[type] = []
```

PR #8289, opened 2025-11-10
Improve robustness of Bokeh/Panel loading in notebooks

philippjfr:
In JupyterLab and other notebook environments, multiple versions of **Bokeh** and **Panel** can coexist or be loaded at different times (e.g. when opening older notebooks with embedded resources). Previously, Panel’s autoload logic made several assumptions that could lead to inconsistent or broken states when those assumptions didn’t hold.

1. **Existing script skip logic too aggressive**

   * We skipped loading scripts that already existed in the DOM, even if the corresponding global objects (`window.Bokeh` or `window.Bokeh.Panel`) had been deleted or were otherwise unavailable.
   * This meant that if a user or another extension removed `window.Bokeh`, Panel could never recover because the loader would still skip the `<script>` elements for those URLs.

2. **No verification of `window.Bokeh.Panel` presence**

   * Panel only checked for an existing `Bokeh` object and version match, not whether the `Panel` extension had actually been initialized.
   * In practice this could result in `Bokeh` being found, but `Panel` remaining undefined and the app failing to render.

3. **Unnecessary restoration of old Bokeh versions**

   * The previous caching logic always restored `root.Bokeh`, even when reloading or when the cached version lacked `Panel`. This prevented proper re-initialization of the correct version.

### Fixes Applied

* Introduced explicit regexes (`BK_RE`, `PN_RE`) to detect Bokeh and Panel CDN URLs.
* Adjusted **skip logic** to only skip script re-loading if both:

  * The script URL already exists, **and**
  * A valid `Bokeh` (with `Bokeh.Panel`) object is present.
* Updated **bokeh_loaded** detection to verify both `Bokeh.version` *and* `Bokeh.Panel`.
* Reworked version caching to:

  * Maintain versioned copies in `Bokeh.versions`
  * Restore `Bokeh.Panel` when missing
  * Avoid restoring broken or incomplete versions.

### DIFF
--- a/panel/_templates/autoload_panel_js.js
+++ b/panel/_templates/autoload_panel_js.js
@@ -285,8 +285,8 @@ calls it with the rendered model.
           if (Bokeh.Panel == null) {
             Bokeh.Panel = NewBokeh.Panel;
           }
+          root.Bokeh = Bokeh;
         }
-        root.Bokeh = Bokeh;
       });
     }
   }


panel/_templates/autoload_panel_js.js
```diff
@@ -273,16 +264,28 @@ calls it with the rendered model.
     } else {
       root._bokeh_is_initializing = true
       root._bokeh_onload_callbacks = []
-      const bokeh_loaded = root.Bokeh != null && (root.Bokeh.version === py_version || (root.Bokeh.versions !== undefined && root.Bokeh.versions.has(py_version)));
+      const bokeh_loaded = Bokeh != null && ((Bokeh.version === py_version && Bokeh.Panel) || (Bokeh.versions !== undefined && Bokeh.versions.has(py_version) && Bokeh.versions.has(py_version).Panel));
```
philippjfr:
```suggestion
      const bokeh_loaded = Bokeh != null && ((Bokeh.version === py_version && Bokeh.Panel) || (Bokeh.versions !== undefined && Bokeh.versions.has(py_version) && Bokeh.versions.get(py_version).Panel));
```

panel/_templates/autoload_panel_js.js
```diff
@@ -273,16 +264,28 @@ calls it with the rendered model.
...
+        if (Bokeh != undefined && !reloading) {
+          const NewBokeh = root.Bokeh;
+          if (Bokeh.versions === undefined) {
+            Bokeh.versions = new Map();
+          }
+          if (NewBokeh.version !== Bokeh.version) {
+            Bokeh.versions.set(NewBokeh.version, NewBokeh)
+          } else if (Bokeh.Panel == null) {
+            Bokeh.Panel = NewBokeh.Panel
+	  }
```
hoxbro:
```suggestion
	  }
```
tab

panel/_templates/autoload_panel_js.js
```diff
@@ -273,16 +264,28 @@ calls it with the rendered model.
...
         run_inline_js();
+        if (Bokeh != undefined && !reloading) {
+          const NewBokeh = root.Bokeh;
+          if (Bokeh.versions === undefined) {
+            Bokeh.versions = new Map();
+          }
+          if (NewBokeh.version !== Bokeh.version) {
+            Bokeh.versions.set(NewBokeh.version, NewBokeh)
+          } else if (Bokeh.Panel == null) {
+            Bokeh.Panel = NewBokeh.Panel
```
hoxbro: Shouldn't we always do this, like this:

```suggestion
          } 
          if (Bokeh.Panel == null) {
            Bokeh.Panel = NewBokeh.Panel
          }
```

panel/_templates/autoload_panel_js.js
```diff
@@ -138,7 +140,7 @@ calls it with the rendered model.
     for (let i = 0; i < js_urls.length; i++) {
       const url = js_urls[i];
       const escaped = encodeURI(url)
-      if (skip.indexOf(escaped) !== -1 || existing_scripts.indexOf(escaped) !== -1) {
+      if (skip.indexOf(escaped) !== -1 || existing_scripts.indexOf(escaped) !== -1 && !((Bokeh == null || Bokeh.Panel == null) && (BK_RE.test(escaped) || BK_RE.test(escaped)))) {
```
hoxbro: Not sure why `BK_RE.test(escaped)` is done twice and `PN_RE` is not used at all.

panel/_templates/autoload_panel_js.js
```diff
@@ -27,6 +27,8 @@ calls it with the rendered model.
   const py_version = '{{ version }}'.replace('rc', '-rc.').replace('.dev', '-dev.');
   const reloading = {{ reloading|default(False)|json }};
   const Bokeh = root.Bokeh;
+  const BK_RE = /^https:\/\/cdn\.bokeh\.org\/bokeh\/release\/bokeh-/;
+  const PN_RE = /^https:\/\/cdn\.holoviz\.org\/panel\/[^/]+\/dist\/panel/i
```
hoxbro, line 31: How would this work with inline and a local server? Mostly interested in the former tbh.

philippjfr: Inline scripts are always loaded, so yes. We don't support local server for notebooks.

panel/_templates/autoload_panel_js.js
philippjfr:
```suggestion
      if (skip.indexOf(escaped) !== -1 || existing_scripts.indexOf(escaped) !== -1 && !((Bokeh == null || Bokeh.Panel == null) && (BK_RE.test(escaped) || PN_RE.test(escaped)))) {
```

panel/_templates/autoload_panel_js.js
philippjfr: Sure, shouldn't happen but not harmful.

Issue #8290, opened 2025-11-10
Modernizing pn.extension: ESM Modules and Lazy Loading


philippjfr:
## The Problem

Panel’s current approach for loading JavaScript libraries globally via `<script>` tags and accessing them through `window.<library>` causes a range of problems in modern notebook environments, especially **JupyterLab** and **VSCode**. These issues become particularly visible as soon as multiple notebooks coexist in the same browser context. For the most part we have addressed these issues via clever tricks, but in scenarios where multiple environments are actively being used there are still corner cases that cannot be sufficiently addressed using the current approach.

### 1. **Global Namespace Collisions**

All libraries loaded by `pn.extension` are attached to `window`, meaning:

* Every notebook session on the same JupyterLab page shares the same global namespace.
* If Notebook A loads `Bokeh 3.5.0` and Notebook B loads `Bokeh 3.6.0`, whichever loads last *overwrites* the shared `window.Bokeh` object. We added a bunch of handling to support multiple versions here, but it is brittle and annoying to deal with.
* Components in the first notebook may now reference incompatible APIs, leading to subtle rendering or synchronization errors.

### 2. **Non-Isolated Outputs**

In JupyterLab, each notebook output is an independent `<div>` injected into the same page. Global libraries loaded by `<script>` tags pollute the outer context, but widget outputs often execute in inner contexts that may:

* Not have access to the global variable at all.
* Load the same script multiple times, triggering race conditions during library initialization.
* Fail silently when `window.Panel` or `window.Bokeh` are missing or mismatched.

### 3. **Version and Dependency Incompatibilities**

Panel components depend on a variety of third-party libraries (e.g. Tabulator, Vega, DeckGL, Leaflet, Plotly).
Under the current global model:

* Only *one version* of each library can exist on the page at once.
* Different components (or different versions of Panel) that require distinct versions cannot coexist.
* Updating or hot-reloading a component may unexpectedly upgrade/downgrade the global version of that library, breaking others.

### 4. **Lack of Caching and Dependency Tracking**

The global model gives no structured way to:

* Detect which version of a dependency is currently active.
* Share preloaded dependencies safely across notebooks.
* Avoid redundant network requests for already-loaded resources.

ESM-based imports, by contrast, allow consistent module identity, caching, and explicit dependency resolution.

## Solution Summary

[...]

PR #8291, opened 2025-11-11
Modernize notebook resource and extension load

philippjfr:
Implements https://github.com/holoviz/panel/issues/8290


MarcSkovMadsen:
I hope this will be a solution not only in notebooks. But also for custom extensions, pyodide/ pyscript, dynamic loading of elements etc. See https://github.com/holoviz/panel/issues/8290#issuecomment-3870666522.

philippjfr:
Yes, it would. Pre-loading would still be preferred but wouldn't be required.

PR #8304, opened 2025-11-25
Fix notebook reading by explicitly using UTF-8 encoding

### LINKED ISSUE #8302: 500 Error When Previewing Panel App in JupyterLab on Japanese Windows
When previewing a Panel app in a Jupyter notebook in JupyterLab, I get a 500: illegal multibyte sequence error. After debugging, I found that the issue is caused by the default encoding on my Japanese Windows system being cp932. Opening a notebook file with the default encoding triggers the error.
The notebook file should be opened with encoding="utf-8" instead.

The problematic line in the source code is here:

https://github.com/holoviz/panel/blob/42b09569a517a428f34d6ce0b8f01143246c8f84/panel/io/jupyter_server_extension.py#L246

ruoyu0088:
This PR fixes a decoding error that occurs when Panel reads a Jupyter notebook file while previewing a Panel app in JupyterLab. The code currently opens the .ipynb file using the system’s default text encoding, which may not be UTF-8. On systems where the default encoding is something else, decoding can fail and produce a 500 error.

Fixes #8302

### DIFF
--- a/panel/io/jupyter_server_extension.py
+++ b/panel/io/jupyter_server_extension.py
@@ -243,7 +243,7 @@ class PanelJupyterHandler(PanelBaseHandler):
         if self.request.arguments.get('kernel'):
             requested_kernel = self.request.arguments.pop('kernel')[0].decode('utf-8')
         elif notebook_path.suffix == '.ipynb':
-            with open(notebook_path) as f:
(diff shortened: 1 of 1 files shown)


philippjfr:
Greatly appreciated! Will cut a new release towards the end of this week.

Issue #8091, opened 2025-07-23
Tabulator does not work when ipywidgets extension is enabled


bakkiaraj:
#### ALL software version info
ipywidgets_bokeh version: 1.6.0
ipywidgets version: 8.1.7
panel version: 1.7.5

Python 3.12
OS : Linux , Rocky 8.9

#### Description of expected behavior and the observed behavior

I try to use both ipywidgets  and Tabulator  but this combination is not working.

#### Complete, minimal, self-contained example code that reproduces the issue

```python

import datetime as dt
import numpy as np
import pandas as pd
import panel as pn
from loguru import logger
import ipywidgets_bokeh
import ipywidgets as ipw
import panel_material_ui as pmui




def create_app():
        # Print versions
        logger.info(f"ipywidgets_bokeh version: {ipywidgets_bokeh.__version__}")
        logger.info(f"ipywidgets version: {ipw.__version__}")
        logger.info(f"panel version: {pn.__version__}")

        df = pd.DataFrame({
            'int': [1, 2, 3],
            'float': [3.14, 6.28, 9.42],
            'str': ['A', 'B', 'C'],
            'bool': [True, False, True],
            'date': [dt.date(2019, 1, 1), dt.date(2020, 1, 1), dt.date(2020, 1, 10)],
            'datetime': [dt.datetime(2019, 1, 1, 10), dt.datetime(2020, 1, 1, 12), dt.datetime(2020, 1, 10, 13)]
        }, index=[1, 2, 3])

        df_widget = pn.widgets.Tabulator(df, buttons={'Print': "<i class='fa fa-print'></i>"})
        
        app_laypout_templ = pmui.Page(
            title="CBB Fabric Network Visualizer",
            
            main=[pn.Column(df_widget, height=800)],
        )
        return app_laypout_templ


# Main 

# Initialize Panel extensions
pn.extension('tabulator')
pn.extension('ipywidgets') # ipywidgets-bokeh
pn.widgets.Tabulator.theme = 'materialize'

create_app().servable()

```

#### Screenshots or screencasts of the bug in action

<img width="1902" height="1067" alt="Image" src="https://github.com/user-attachments/assets/59677dda-0d7b-4789-8c9c-6cd1147d4fb3" />

philippjfr:
Thanks, seems like a clash between require.js and Tabulator.js internal require.

grapesmoker:
[...]

PR #8308, opened 2025-11-26
Ipywigets and Tabulator extensions not working if both loaded

maximejay:
Since Panel v1.6.2 Ipywidgets and Tabulator extensions was unusable together. This commit restore a small fraction of the code deleted by the commit a258611 (PR 7781).

This pull request addresses a fix for issue #8091.

Ipywidgets complains if require.js is not present in js_files. 
Tabulator get in conflict if require.je is loaded first. 

This commit move the require.js file at the end of the list js_files. require.js is loaded last. Tabulator and Ipywidgets works again if both loaded.

### DIFF
--- a/doc/how_to/caching/manual.md
+++ b/doc/how_to/caching/manual.md
@@ -27,4 +27,5 @@ data = pn.state.as_cached('data', load_data, *args, **kwargs)
 Now, the first time the app is loaded the data will be cached and subsequent sessions will simply look up the data in the cache, speeding up the process of rendering. If you want to warm up the cache before the first user visits the application you can also provide the `--warm` argument to the `panel serve` command, which will ensure the application is initialized as soon as it is launched. If you want to populate the cache in a separate script from your main application you may also provide the path to a setup script using the `--setup` argument to `panel serve`.
 
 ## Related Resources
+
 - If you want to periodically update the cache, consult the [How to > Schedule Tasks](../callbacks/schedule) guide.
--- a/doc/how_to/caching/memoization.md
+++ b/doc/how_to/caching/memoization.md
@@ -80,7 +80,65 @@ DataExplorer().servable()
 
 ## Disk Caching
 
-If you have `diskcache` installed, you can also cache the results to disk by setting `to_disk=True`. The `diskcache` library will then cache the value to the supplied `cache_path` (defaulting to `./cache`). Making use of disk caching allows you to cache items even if the server is restarted. You can configure the `cache_path` inline or globally by setting `pn.extension(cache_path=...)` or `pn.config.cache_path = ...`.
+Disk backed caching persists cached results across server restarts, making it ideal for expensive operations like data loading or model inference.
+
+### Prerequisites
+
+First, install the [`diskcache`](https://grantjenks.com/docs/diskcache/) library:
+
+```bash
+pip install diskcache
+```
+
+### Basic Disk Caching
+
+To enable disk caching, set `to_disk=True` when decorating your function:
+
+```{pyodide}
+import panel as pn
+import time
+from datetime import datetime
+
+@pn.cache(to_disk=True)
+def expensive_computation(n):
+    # Simulate expensive operation
+    time.sleep(2)
+    return datetime.now()
+
+# First call takes 2 seconds
+result1 = expensive_computation(5)
+
+# Subsequent calls are instant, even after server restart
+result2 = expensive_computation(5)
+
+pn.Column(result1, result2).servable()
+```
+
+By default, cached values are stored in a `./cache` directory relative to your application.
+
+### Configuring the Cache Path
+
+You can customize where cached values are stored in three ways:
+
+**1. Inline configuration:**
+
+```python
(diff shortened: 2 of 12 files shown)


philippjfr:
Thanks @maximejay, I'm okay with restoring this but will likely rework the JS loading machinery entirely fairly soon so hopefully this won't be needed at that point.

maximejay:
Thank you [philippjfr](https://github.com/philippjfr). 
Tell me if more work/tests are needed to approve this PR. 
At least this will restore the  functioning of the panel extensions Tabulator and Ipywidgets waiting the rework of the JS loading machinery...

Perhaps, we should add a test to prevent such regression in the future ?

philippjfr (approved):

philippjfr:
A test would be great, if you can manage it.

maximejay:
Hi,

I am not familiar with pytest and Panel code, but just in case :

Is a test  in panel/tests/ui/widgets/test_tabulator.py should do the job ?
And would this function work ?

```python
   def test_tabulator_with_loading_ipywidgets_no_console_error(page, df_mixed):

    extension("ipywidgets", "tabulator")

    widget = Tabulator(df_mixed)
    msgs, _ = serve_component(page, widget)
    page.wait_for_timeout(1000)

    assert [msg for msg in msgs if msg.type == 'error' and 'favicon' not in msg.location['url']] == []
```

philippjfr:
No need for the extension call there, otherwise that should work I think. If not I'll iterate on it.

maximejay:
Call panel.extension("ipywidgets", "Tabulator") (with both extension) was the cause of the problem above. If I remove this line, the test would correspond exactly to the test function  `def test_tabulator_no_console_error` on  (l. 103)
If no extension call is needed, the test already exist and pass with the current Panel version... Thus this is not the good way to trigger and test the problem above. But may be I misunderstand something...

philippjfr:
Ah, my bad, let's do something like this:

```python
def test_tabulator_with_loading_ipywidgets_no_console_error(page, df_mixed):

    def app(): 
        extension("ipywidgets", "tabulator")
        return Tabulator(df_mixed)

    msgs, _ = serve_component(page, app)
    page.wait_for_timeout(1000)

    assert [msg for msg in msgs if msg.type == 'error' and 'favicon' not in msg.location['url']] == []
```

The problem is that the extension should be run in the context of the particular session we are testing.

maximejay:
Ok thanks,
I am not sure the test is really useful. I tried it on my computer and the result are:
- with the fix suggested in this PR all Tabulator tests pass ! This is the good point !
- without the fix suggested in this PR several Tabulator tests failed. Thus it should failed since v1.6.2.
```
FAILED panel/tests/ui/widgets/test_tabulator.py::test_tabulator_with_loading_ipywidgets_no_console_error - playwright._impl._errors.TimeoutError: Timeout 20000ms exceeded.
FAILED panel/tests/ui/widgets/test_tabulator.py::test_tabulator_buttons_display - AssertionError: Locator expected to have count '10'
FAILED panel/tests/ui/widgets/test_tabulator.py::test_tabulator_selection_selectable_disabled - playwright._impl._errors.TimeoutError: Locator.wait_for: Timeout 20000ms exceeded.
FAILED panel/tests/ui/widgets/test_tabulator.py::test_tabulator_edit_event_empty_to_nan - playwright._impl._errors.TimeoutError: Locator.click: Timeout 20000ms exceeded.
FAILED panel/tests/ui/widgets/test_tabulator.py::test_tabulator_edit_event_integrations[local-header_filter-no_python_filter-no_sorter] - playwright._impl._errors.TimeoutError: Locator.click: Timeout 20000ms exceeded.
FAILED panel/tests/ui/widgets/test_tabulator.py::test_range_selection_on_sorted_data_downward[local] - playwright._impl._errors.TimeoutError: Locator.click: Timeout 20000ms exceeded.
```
I commit the new test `test_tabulator_with_loading_ipywidgets_no_console_error` and I let you decide if you approved it or not.
Thanks

philippjfr:
Interesting, suspect the extension call is still not correctly scoped in the session. I'll take a look once you've committed it. Appreciate your effort!

maximejay:
Hi,
The test has been committed (see above)  :  https://github.com/holoviz/panel/pull/8308/commits/ae6fee57dd59de42f748852b810735f363046176  ([Add test for testing the Tabulator extension when ipywidgets is loaded.](https://github.com/holoviz/panel/pull/8308/commits/ae6fee57dd59de42f748852b810735f363046176))
But 5 auto tests failed ... Should I rebase (or update) my  branch with "main" ?

philippjfr:
I'll look into the test now.

philippjfr:
Actually looks like the test was fine.

maximejay:
Good news !

So only this test test_tabulator_patch_no_horizontal_rescroll does not succeed. But it do not seems to be related to this commit. However all test passed on last Thursday with my panel fork (created the week before). It seems that something is getting wrong again with Tabulator since last week (library update, new test, probably related to the patch "no_horizontal_rescroll" ?)

Thank so much you for your efficient code review !

I hope everything will work fine with the next Panel release :)

PR #8309, opened 2025-11-27
Allow Tabulator.row_content to be an awaitable

maximlt:
This PR allows `row_content` to be an awaitable. When it is an awaitable, we display first a small spacer with a loading spinner, replaced by the content returned by the function.

Micro app to play around with this feature:
```python
import asyncio

import panel as pn
import pandas as pd

df = pd.DataFrame({"value": [0, 1, 2]})

async def async_row_content_cb(event):
    await asyncio.sleep(1)
    print("Event", event)
    return str(event)

def row_content_cb(event):
    print("Event", event)
    return str(event)

b = pn.widgets.Button(name="expanded")

t = pn.widgets.Tabulator(
    df,
    width=300,
    # row_content=row_content_cb,
    row_content=async_row_content_cb,
    # embed_content=True,
    # expanded=[1],
)

b.on_click(lambda e: setattr(t, 'expanded', [0]))

pn.Row(b, t).servable()
```

Draft as I'd like to test it in an app with a more complex `row_content` function.

### DIFF
--- a/panel/tests/widgets/test_tables.py
+++ b/panel/tests/widgets/test_tables.py
@@ -625,12 +625,9 @@ async def test_tabulator_content_embed_async(document, comm):
 
     table = Tabulator(df, embed_content=True, row_content=row_content)
 
-    # To create an event loop
-    serve_and_request(table)
-
     model = table.get_root(document, comm)
 
-    await async_wait_until(lambda: len(model.children) == len(df))
+    assert len(model.children) == len(df)
 
     for i, r in df.iterrows():
         await async_wait_until(lambda i=i, r=r: resolve_async_row_content_text(model, i) == f"&lt;pre&gt;{r.A}&lt;/pre&gt;")
--- a/panel/tests/util.py
+++ b/panel/tests/util.py
@@ -263,7 +263,7 @@ async def async_wait_until(fn, page=None, timeout=5000, interval=100):
             # None is returned when the function has an assert
             if result is None:
                 return
-            # When the function returns True or False
+            # When the function returns True
             if result:
                 return
             if timed_out():


panel/widgets/tables.py
```diff
@@ -1253,9 +1254,10 @@ class Tabulator(BaseTable):
     page_size = param.Integer(default=None, bounds=(1, None), doc="""
         Number of rows to render per page, if pagination is enabled.""")
 
-    row_content = param.Callable(doc="""
+    row_content = param.Callable(allow_refs=False, doc="""
```
maximlt, line 1276: I had to disable `allow_refs` as otherwise the async callback set (`Tabulator(row_content=async_row_content)`) was considered as a reference and immediately resolved.

philippjfr: Yes, that's sensible.

panel/widgets/tables.py
```diff
@@ -1625,10 +1627,31 @@ def _update_style(self, recompute=True):
...
+
+            async def _await_content():
+                try:
+                    resolved = await content
+                except Exception:
+                    placeholder.loading = False
+                    raise
+
+                placeholder.object = resolved
+                placeholder.loading = False
```
philippjfr:
```suggestion
                placeholder.param.update(object=resolved, loading=False)
```

panel/tests/widgets/test_tables.py
```diff
@@ -590,6 +590,61 @@ def test_tabulator_expanded_content(document, comm):
...
+        return False
+    child = model.children[idx]
+    if not getattr(child, "children", None):
+        return False
+    if not hasattr(child.children[0], "text"):
+        return False
+    return child.children[0].text
+
+
+def test_tabulator_expanded_content_async(document, comm):
```
philippjfr: Think you have to make these tests async as well

```suggestion
async def test_tabulator_expanded_content_async(document, comm):
```

panel/tests/widgets/test_tables.py
```diff
@@ -590,6 +590,61 @@ def test_tabulator_expanded_content(document, comm):
...
+
+    table = Tabulator(df, expanded=[0], row_content=row_content)
+
+    model = table.get_root(document, comm)
+
+    wait_until(lambda: resolve_async_row_content_text(model, 0) == "&lt;pre&gt;0.0&lt;/pre&gt;")
+    assert len(model.children) == 1
+
+
+def test_tabulator_content_embed_async(document, comm):
```
philippjfr:
```suggestion
async def test_tabulator_content_embed_async(document, comm):
```

Issue #6155, opened 2024-01-04
Textinput widget value updated too late after the widget loses focus


maximlt:
Ok that's not my best GIF but it proves the point: the `value` of the text input widget is not always the one displayed on the screen when clicking on the button, it sometimes lags behind (when I enter `5` and click, it prints `1234`).

I thought it was a Bokeh issue but I couldn't reproduce it with a pure Bokeh app (it doesn't mean it's not a Bokeh issue either, it might be that the overhead added by Panel makes it easier to expose it).

```python
import panel as pn

b = pn.widgets.Button(name='click')
ti = pn.widgets.TextInput()

def debug(e):
    print(ti.value)

b.on_click(debug)

pn.Column(ti, b).servable()
```

![panel_issue_button](https://github.com/holoviz/panel/assets/35924738/bf1e6ce9-8c6b-40a3-853a-673ee59aad84)

ahuang11:
I wonder what if you use `value_input`?

maximlt:
Using `value_input` instead is the trick I've used in apps that suffer from this issue, it seems to be a valid workaround.

PR #8311, opened 2025-11-30
Ensure the order of Bokeh property changes and events is preserved

maximlt:
Fixes https://github.com/holoviz/panel/issues/6155

To summarize the issue which is pretty simple:
- A user types some text in a TextInput and clicks on a button that has a callback registered with `on_click`
- The callback accesses the widget `value` attribute
- The string returned is not the one displayed but the older one (often an empty string in practice)

To reproduce the reported issue:
- I've been serving a small app, not tested yet in a notebook
- On my laptop, my trackpad is set with the "tap-to-click" mode, which I've used to reproduce the issue; a normal click wouldn't generally help me in reproducing

After some debugging, what I could see is that:
- The events were sent in the right order by Bokeh (bokeh `value` change and then click event)
- The bokeh `value` change is delayed by 50 ms in Panel (`doc.add_timeout_callback(cb, self._debounce)`, btw this is called debounce, but is it really doing debouncing?)
- If the click event happens quickly enough after the bokeh `value` change (which seems to always be the case when I do a "tap-to-click" on my trackpad, less so with a normal click), the on_click callback will be triggered before the bokeh `value` change triggers the widget `value` update.

I found two different ways that seem to fix the issue. I'm not yet sure which one is the best, or even if either is!

- Set `TextInput._priority_changes` to `['value']` to avoid "debouncing" `value` changes.
- Also "debounce" in `_process_bokeh_event`.

My guess is that the solution is a little bit more subtle than that.

TODO:
- [ ] Check in a notebook

### DIFF
--- a/panel/reactive.py
+++ b/panel/reactive.py
@@ -617,14 +617,14 @@ class Syncable(Renderable):
         if doc.session_context:
             event_id = uuid.uuid4().hex
             cb = partial(self._change_coroutine, doc, event_id=event_id)
+            if doc in state._change_callbacks:
+                state._change_callbacks[doc][event_id] = cb
+            else:
+                state._change_callbacks[doc] = {event_id: cb}
             if attr in self._priority_changes:
                 with set_curdoc(doc):
                     state.execute(cb, schedule=True)
             else:
-                if doc in state._change_callbacks:
-                    state._change_callbacks[doc][event_id] = cb
-                else:
-                    state._change_callbacks[doc] = {event_id: cb}
                 doc.add_timeout_callback(cb, self._debounce) # type: ignore
         else:
             try:


philippjfr:
Thanks for digging into this. The difference in handling here is to avoid excessive change events being triggered, so we delay processing of a change event and "accumulate" changes over that time period. For bokeh events each event represents a discrete action so it should not be "accumulated". That said the problem you describe is obviously real. The ideal solution would be that a discrete event would force any accumulated events in the session to be processed immediately, rather than delaying the discrete event as well. That said I'm not sure of the feasibility of that, but I'll take a quick look at it.

panel/reactive.py
```diff
@@ -604,10 +615,16 @@ def _server_change(
             return
 
         if doc.session_context:
-            cb = partial(self._change_coroutine, doc)
+            event_id = uuid.uuid4().hex
+            cb = partial(self._change_coroutine, doc, event_id=event_id)
             if attr in self._priority_changes:
-                doc.add_next_tick_callback(cb) # type: ignore
+                with set_curdoc(doc):
+                    state.execute(cb, schedule=True)
```
hoxbro, line 626: Does this always work? Couldn't it potentially exit the `set_curdoc` context manager before the callback is called in `state.execute`?

philippjfr: I think you asked the same question on another similar PR recently. The answer is yes, state.execute handles the curdoc appropriately, this just ensures that state.execute has access to the correct curdoc.

philippjfr:
I think in `Tabulator` we have explicitly depended on the fact that server events are handled before change events so I suspect that requires some careful fixes.

philippjfr:
Will merge and cut an RC release tonight. Would appreciate it if you could test with it in the morning.

maximlt:
It doesn't look like it's working as expected. I guess my test wasn't good enough 🙈 .

PR #8314, opened 2025-12-01
type: Fix type checking

hoxbro:
This was indirectly imported via `import bokeh.core.properties as bp`, which made it work, but a new release of mypy now flags this as an error.


philippjfr:
Not sure what's up with the JupyterLite UI tests yet, but that's a separate issue. Merging.

hoxbro:
Seems like a new version breaks our lock file hack. 

Screenshot From 2025-12-01 11-34-50

PR #8330, opened 2025-12-04
Support pandas dataframes in vega datasets

ahuang11:
Previously `if data:` resulted in
```python
ValueError: The truth value of a DataFrame is ambiguous. Use a.empty, a.bool(), a.item(), a.any() or a.all().
```

Then `data[0]` is invalid to get columns, so I just used `hasattr()` which theoretically also supports more than just pandas, like polars.

### DIFF
--- a/panel/pane/vega.py
+++ b/panel/pane/vega.py
@@ -384,20 +384,7 @@ class Vega(ModelPane):
                 # Handle geometry records types
                 datasets[name] = data
                 continue
-            if hasattr(data, 'columns'):
-                columns = data.columns
-            else:
-                columns = set(data[0]) if data else []
-            if self.is_altair(self.object):
-                import altair as alt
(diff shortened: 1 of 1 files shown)


philippjfr:
>Previously if data: resulted in

I don't see where `if data` is used. Also, this should be using Narwhals.

ahuang11:
Sorry, I meant `if data` 
`columns = set(data[0]) if data else []`

I could implement narwhals for 
```
            if hasattr(data, 'columns'):
                columns = data.columns
```
into
```
nw.from_native(data).columns
```

However, I'm not sure it has any benefits since it doesn't cover this case:
```
data = [
    {"a": 1, "b": 2},
    {"a": 3, "b": 4},
]
```
which I think `columns = set(data[0]) if data else []` is trying to address

philippjfr:
Sorry, appears I didn't read well. I pushed some changes, the entire piece of complexity was an optimization. So I've disabled that for now.

philippjfr (approved):

PR #8334, opened 2025-12-10
fix: Notebook erroring out

hoxbro:
Seeing this error;

image


philippjfr:
Hmm, no bueno, weird that I never saw this. New patch release I guess.

PR #8351, opened 2025-12-30
Ensure ESM component properties aren't transformed twice

philippjfr:
Fixes regression introduced in https://github.com/holoviz/panel/pull/8338


philippjfr:
Test failures are due to the partially broken 1.8.5 release this PR fixes. Merging.

PR #8357, opened 2026-01-04
Interactive widgets in header- fixes #8353

### LINKED ISSUE #8353: Support interactive widgets in Card Header without toggling expand/collapse
Issue #7045 was fixed by PR #7057 in a way which was specific to text inputs as widgets in Card Headers.  However, it would be useful to have other user-interaction widgets which stay visible when the Card is collapsed, particularly the ability to add buttons (e.g. to remove a Card, move a Card within a Column, trigger other action, etc).

Currently, including any user-interactive widgets other than text inputs in a Card Header will still trigger expand/collapse of the Card when the user clicks the widget.

User-interactive widgets (buttons, inputs, checkboxes, etc) could be included in Card Headers and could be usefully clickable without automatically toggling the card's collapsed state.

1. Attempted to block expand/collapse behavior by suppressing propagation of the user click event using Javascript.  Because of the Shadow DOM, this was not found to be workable.
2. Attempted to extend the previous filter of `if (path instanceof HTMLInputElement)` to be `if (path instanceof HTMLInputElement || path instanceof HTMLButtonElement)` but because the entire Header is wrapped in a Button element, all clicks in the Header would be caught by the conditional.
3. Attempted to restrict the selection to just `if (e.target !== this.button_el)` but other children element on the Header (especially the Row div) would also catch this, leaving a very small active area where the true button div was present. 

#### Other notes:

Secondarily, Issue #7045 mentions that pressing spacebar in a text input field in the Header will trigger expand/collapse.  This is still unresolved- should #7045 be re-opened, or should a new issue be opened?

emunsing:
Previous behavior:  Previous #7057 specifically allowed TextInput objects in Card Headers, but the fix was specific to the TextInput widget and did not support any other interactive widgets (e.g. Select, Button, Slider, etc).  

This PR, described in #8353 , generalizes based on CSS class to stop propagation of clicks which begin in any object which begins with `bk-panel-models-widgets-`  which could be tweaked if needed to catch the right level of specificity (I defer to the repo owners; please provide feedback).

UI test has been updated (this updated test with a slider in the header was previously failing)

Linted and ran UI tests locally with no regressions, but this is my first PR on this repo!

### DIFF
--- a/panel/models/card.ts
+++ b/panel/models/card.ts
@@ -200,8 +200,13 @@ export class CardView extends ColumnView {
   }
 
   _toggle_button(e: MouseEvent): void {
-    for (const path of e.composedPath()) {
-      if (path instanceof HTMLInputElement) {
+    const is_panel_widget = (el: EventTarget): boolean =>
+      el instanceof HTMLElement &&
+      Array.from(el.classList).some((c) => c.startsWith("bk-panel-models-widgets-"))
+
+    for (const el of e.composedPath()) {
+      // If the click came from any Panel widget in the header, don't toggle.
+      if (is_panel_widget(el)) {
         return
       }
     }
--- a/panel/models/card.ts
+++ b/panel/models/card.ts
@@ -201,8 +201,9 @@ export class CardView extends ColumnView {
 
   _toggle_button(e: MouseEvent): void {
     const is_panel_widget = (el: EventTarget): boolean =>
+      el instanceof HTMLInputElement || (
       el instanceof HTMLElement &&
-      Array.from(el.classList).some((c) => c.startsWith("bk-panel-models-widgets-"))
+	Array.from(el.classList).some((c) => c.startsWith("bk-panel-models-widgets-")))
 
     for (const el of e.composedPath()) {
       // If the click came from any Panel widget in the header, don't toggle.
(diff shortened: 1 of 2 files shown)


philippjfr (approved):

panel/models/card.ts
```diff
@@ -200,8 +200,13 @@ export class CardView extends ColumnView {
   }
 
   _toggle_button(e: MouseEvent): void {
-    for (const path of e.composedPath()) {
-      if (path instanceof HTMLInputElement) {
+    const is_panel_widget = (el: EventTarget): boolean =>
+      el instanceof HTMLElement &&
+      Array.from(el.classList).some((c) => c.startsWith("bk-panel-models-widgets-"))
```
philippjfr: Sadly I don't think this is really enough, because some widgets come from Bokeh and simply have a prefix like `bk-` and then there's panel-material-ui widgets which always just have this `bk-panel-models-esm-ReactComponent`

panel/models/card.ts
```diff
@@ -200,8 +200,14 @@ export class CardView extends ColumnView {
   }
 
   _toggle_button(e: MouseEvent): void {
-    for (const path of e.composedPath()) {
-      if (path instanceof HTMLInputElement) {
+    const is_panel_widget = (el: EventTarget): boolean =>
+      el instanceof HTMLInputElement || (
+      el instanceof HTMLElement &&
```
philippjfr:
```suggestion
        el instanceof HTMLElement &&
```

PR #8360, opened 2026-01-04
Set placeholder for IntInput (Fixes #8359)

### LINKED ISSUE #8359: IntInput with value=None displays a placeholder value of 0

<details>

```plaintext
panel                                   1.8.5.post1
bokeh                                   3.8.1
Browser                               Chrome 143.0.7499.42
```
</details>

A FloatInput with no value set or `value=None` will not display any placeholder text - this is great and expected, as it visually informs the user that the value has not been set.

An IntInput with no value set or `value=None` will display 0 as placeholder text, even if 0 would not be a valid input per the `start, stop` values: e.g. `IntInput(name='Foo', value=None, start=10, end=100)` displays 0

Overly robust example for demonstration and exploration:

```python
from panel.widgets import IntInput, StaticText, Button, FloatInput
from panel import Column
import numpy as np

input_field = IntInput(name='Foo', value=None, start=10, end=100)
# input_field = FloatInput(name='Foo', value=None, start=10, end=100)

output_text = StaticText(
    name="Consolidated Results",
    value="",
)

def show_input_value(event):
    print(f" Input Value: {input_field.value}")
    output_text.value += f"\nInput: {input_field.value}"

read_output_btn = Button(name="Read Input")
read_output_btn.on_click(show_input_value)

Column(input_field,
       read_output_btn,
       output_text).servable()

```

- [X] I may be interested in making a pull request to address this

emunsing:
1-liner fix to add a placeholder value to 8359.  I'm not sure how to create a UI test for a *placeholder* value; happy to create one with help but given that this is a 1-line aesthetic change I don't know if that's needed.

### DIFF
--- a/panel/tests/widgets/test_input.py
+++ b/panel/tests/widgets/test_input.py
@@ -483,3 +483,14 @@ def test_file_input_save_one_file(document, comm, tmpdir):
     assert fpath.exists()
     content = fpath.read_text()
     assert content == 'Some text\n'
+
+def test_int_input_placeholder(document, comm):
+
+    int_input = IntInput(placeholder='Foo', name='IntInput')
+
+    widget = int_input.get_root(document, comm=comm)
+
+    assert widget.placeholder == "Foo"
+
+    int_input.placeholder = "Bar"
+    assert widget.placeholder == "Bar"
--- a/panel/widgets/input.py
+++ b/panel/widgets/input.py
@@ -1069,6 +1069,9 @@ class IntInput(_SpinnerBase, _IntInputBase):
     >>> IntInput(name='Value', value=100, start=0, end=1000, step=10)
     """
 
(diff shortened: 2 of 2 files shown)


philippjfr:
Thanks @emunsing, I don't think we really need a UI test, a simply unit test checking that the underlying model inherits the correct placeholder value should be sufficient.

philippjfr:
```python

def test_int_input_placeholder(document, comm):

    int_input = IntInput(placeholder='Foo", name='IntInput')

    widget = int_input.get_root(document, comm=comm)

    assert widget.active == "Foo"
    
    int_input.placeholder = "Bar"
    assert widget.placeholder == "Bar"
```

philippjfr:
Thanks @emunsing, wanted to make sure you get the credit so just added the test here instead of making a new PR.

Issue #8374, opened 2026-01-16
Bad request in scheduled refresh token task leads to unreachable app


TheoMathurin:
I noticed the issue in a real-world app with Panel 1.7.5, which is fairly old but I see that there were no edits to the `auth.py` file since then. I don't believe other libs could be involved.

I have enabled automatic OAuth access token refresh in my app. Since then, from time to time I find the application completely unreachable. The logs show it's due to a failed refresh token request.

Now, the failure itself may not be due to flawed panel code, which seeks to refresh a token for a OAuth (keycloak) session that's no longer active, and I will investigate this in its own right. The real problem is that this crash leads the app to be completely unreachable from that point in time.

I can't share an MRE since this bug involves communication with a third-party OAuth server, and I haven't yet found a pattern for its occurrence.

Here is the full traceback:
```
2026-01-13 19:19:58,020 Exception in callback functools.partial(<bound method IOLoop._discard_future_result of <tornado.platform.asyncio.AsyncIOMainLoop object at 0x7fb68c023620>>, <Task finished name='Task-22455' coro=<OAuthProvider._scheduled_refresh() done, defined at /usr/local/lib/python3.12/site-packages/panel/auth.py:1323> exception=HTTPError()>)
Traceback (most recent call last):
  File "/usr/local/lib/python3.12/site-packages/panel/auth.py", line 231, in _fetch_access_token
    response = await http.fetch(req)
               ^^^^^^^^^^^^^^^^^^^^^
tornado.httpclient.HTTPClientError: HTTP 400: Bad Request

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/usr/local/lib/python3.12/site-packages/tornado/ioloop.py", line 758, in _run_callback
    ret = callback()
          ^^^^^^^^^^
  File "/usr/local/lib/python3.12/site-packages/tornado/ioloop.py", line 782, in _discard_future_result
    future.result()
  File "/usr/local/lib/python3.12/site-packages/panel/auth.py", line 1326, in _scheduled_refresh
    await self._refresh_access_token(user, refresh_token, application, request)
  File "/usr/local/lib/python3.12/site-packages/panel/auth.py", line 1359, in _refresh_access_token
    ) = await auth_handler._fetch_access_token(
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/site-packages/panel/auth.py", line 234, in _fetch_access_token
[...]

PR #8384, opened 2026-01-23
Don't attempt to refresh access token if there is no active user session

TheoMathurin:
The check on `state._active_users` could as well be in `_refresh_access_token`.

Attempt at fixing #8374.

I'm saying it's an attempt because there is still the possibility that `state._active_users` is not reliable enough here. But in any case, if this condition is met, there should definitely be no attempt to refresh the token.

### DIFF
--- a/panel/auth.py
+++ b/panel/auth.py
@@ -1198,6 +1198,8 @@ class OAuthProvider(BasicAuthProvider):
             state.schedule_task(task, refresh_cb, at=expiry_date)
 
     async def _scheduled_refresh(self, user, refresh_token, application, request, reschedule=True):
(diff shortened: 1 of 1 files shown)


TheoMathurin:
Oops, did not see #8383 created moments before

philippjfr:
Sorry! I'm keeping yours.

TheoMathurin:
From what I understand and explained [here](https://github.com/holoviz/panel/issues/8374#issuecomment-3790450959), we can say that this PR will not fix #8374. However, as said above it's definitely necessary to prevent refresh attempts there anyway.

PR #8385, opened 2026-01-23
Updates for pandas 3.0 compatibility

philippjfr:


panel/io/cache.py
```diff
@@ -218,6 +218,8 @@ def _io_hash(obj):
     dt.date      : lambda obj: f'{type(obj).__name__}{obj}'.encode(),
     # Fully qualified type strings
     'numpy.ndarray'              : _numpy_hash,
+    'pandas.DataFrame'           : _pandas_hash,
+    'pandas.Series'              : _pandas_hash,
```
hoxbro, line 223: I think other `pandas` imports also needs update:

image

doc/tutorials/basic/size.md
```diff
@@ -133,7 +133,7 @@ data = pd.DataFrame(
 )
 
 fig = figure(
-    x_range=data["Day"],
+    x_range=list(data["Day"]),
```
hoxbro, line 136: Most elegant way to make it work, but will need a new Bokeh release to work as before.

panel/io/cache.py
hoxbro: It does not look like `pandas.core.indexes.numeric.Int64Index` exists anymore so I have not added it.

examples/reference/indicators/Tqdm.ipynb
hoxbro, line 1: Fails because of the removal done in https://github.com/pandas-dev/pandas/pull/57444, and tqdm doesn't seem to be actively maintained anymore.

PR #8391, opened 2026-01-26
Guard against errors during Feed visible checks

philippjfr:
Sometimes entries in the `Feed` are removed while we check for visibility which should not cause errors.


philippjfr:
Test failures due to pyodide and the rc release. Will merge.

PR #8407, opened 2026-02-04
Support threaded session creation

philippjfr:
This PR aims to allow moving the creation of new sessions onto a thread with two primary goals:

1. In the short term this will allow any blocking I/O happening during the creation of a new session to no longer block interactivity on the server.
2. In the long term this is meant to prepare us for a future where free-threaded Python is fully supported.

As-is, using cp314t (i.e. free-threaded Python 3.14) this PR increases the number of sessions created per second by about 25% for a medium complexity app without any blocking I/O.

The approach is simple, we subclass bokeh's ApplicationContext and then reuse the thread pool created by setting `--num-threads` (in the CLI) and `config.nthreads` (in Python) and move the session creation onto there.


panel/io/session.py
```diff
@@ -1,18 +1,29 @@
 from __future__ import annotations
 
+import asyncio
 import logging
+import threading
+import weakref
 import typing as t
```
philippjfr, line 6:
```suggestion
import threading
import typing as t
import weakref
```

philippjfr:
No longer needed since Bokeh now handles this.

PR #8408, opened 2026-02-04
Fix CDN requests in server mode for loading.css and es-module-shims

nojaf:
## Summary

Fixes external CDN requests when running Panel in server mode, ensuring all resources are served locally.

## Problem

When deploying Panel applications in server mode with `PANEL_NPM_CDN` configured to an internal registry, the app still made external network requests to:

1. `https://cdn.holoviz.org/panel/<version>/dist/css/loading.css`
2. `https://cdn.jsdelivr.net/npm/es-module-shims@^1.10.0/dist/es-module-shims.min.js`

This is problematic for air-gapped environments or deployments with strict network policies that require all resources to be served locally.

### 1. `loading.css` (panel/reactive.py)

The loading stylesheet URL was hardcoded using `CDN_DIST`, which always points to `cdn.holoviz.org` regardless of the resource mode:

```python
stylesheets = [..., f'{CDN_DIST}css/loading.css']
```

### 2. `es-module-shims` (panel/io/resources.py)

This one is more subtle. The `ReactiveESM` class defines `__javascript_raw__` at class definition time (import time), capturing the *default* CDN URL before `PANEL_NPM_CDN` is read:

```python
__javascript_raw__ = [f"{config.npm_cdn}/es-module-shims@^1.10.0/dist/es-module-shims.min.js"]
```

Later, `bundled_files()` checks if URLs start with `config.npm_cdn` to map them to local files. But since `config.npm_cdn` now reflects the custom registry URL while the captured URL still has `cdn.jsdelivr.net`, the check fails and the file isn't served locally.

### 1. `loading.css`

Use `get_dist_path()` instead of `CDN_DIST`. This function respects the current resource mode and returns the local path in server mode.

### 2. `es-module-shims`

Expand the prefix matching in `bundled_files()` to recognize all common NPM CDN prefixes:

```python
npm_cdn_prefixes = [
    config.npm_cdn,
    'https://cdn.jsdelivr.net/npm',
    'https://unpkg.com',
]
```

This ensures URLs captured at import time with the default CDN are still correctly mapped to local bundled files, regardless of the runtime `PANEL_NPM_CDN` setting.

### DIFF
--- a/panel/io/resources.py
+++ b/panel/io/resources.py
@@ -312,6 +312,8 @@ def patch_stylesheet(stylesheet, dist_url):
         patched_url = url.replace(CDN_DIST, dist_url)
     elif url.startswith(LOCAL_DIST) and dist_url.lstrip('./').startswith(LOCAL_DIST):
         patched_url = url.replace(LOCAL_DIST, dist_url)
+    elif url.startswith(LOCAL_DIST) and dist_url != LOCAL_DIST:
+        patched_url = url.replace(LOCAL_DIST, dist_url)
     else:
         return
     version_suffix = f'?v={JS_VERSION}'
@@ -415,11 +417,15 @@ def bundled_files(model: Model, file_type: str = 'javascript') -> list[str]:
     bdir = BUNDLE_DIR / name
     shared = list((JS_URLS if file_type == 'javascript' else CSS_URLS).values())
     files = []
+    npm_cdn_prefixes = (config.npm_cdn, 'https://cdn.jsdelivr.net/npm', 'https://unpkg.com')
     for url in raw_files:
         if url.startswith(CDN_DIST):
             filepath = url.replace(f'{CDN_DIST}bundled/', '')
-        elif url.startswith(config.npm_cdn):
-            filepath = url.replace(config.npm_cdn, '')[1:]
+        elif url.startswith(npm_cdn_prefixes):
+            for prefix in npm_cdn_prefixes:
+                if url.startswith(prefix):
+                    filepath = url.replace(prefix, '')[1:]
+                    break
         else:
             filepath = url_path(url)
         test_filepath = filepath.split('?')[0]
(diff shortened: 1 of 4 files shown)


nojaf:
Related issue https://github.com/holoviz/panel/issues/6161

philippjfr:
Thanks @nojaf, this looks good.

nojaf:
@philippjfr thanks, sorry for all the formatting changes.
Is there some command I can run to format all the touched files?
My IDE does format on save 😅

hoxbro:
We don't have a formatter. I know VS Code have an option to `save without formatting` in the task menu (`ctrl+shift+p`)

I think the easiest way is doing something like `git checkout main -p -- panel/io/resources.py` and remove all the format changes + some manual changes if needed.

nojaf:
Thanks for the feedback on minimizing the diff @hoxbro ! I've force-pushed a cleaner version.

One thing that would help future contributions: would you consider adding a formatter (like `ruff format`) to the pre-commit hooks? It would prevent these formatting inconsistencies from creeping in and make reviews easier for everyone.

philippjfr:
Appreciate it! I do think we should adopt more formatting rules, i.e. consistent quoting would be nice. Will do that one of these days.

nojaf:
Sounds great! I do notice that some tests fail @philippjfr, I'm a bit confused if they are related or not? Any thoughts?

philippjfr:
Very confused too, just ran them locally without issue. I'll restart them but it's a bit worrying that some of the resource related tests are failing, seemingly randomly.

philippjfr:
Guess it's not totally random, ubuntu fails consistently.

philippjfr:
Well not that consistently apparently, one just passed and when I run it on my linux box and it works just fine.

philippjfr:
Okay, I think this was a result of bokeh document state leaking across tests. Should be resolved now.

nojaf:
Thanks for merging this, @philippjfr! Much appreciated. This problem has been on our radar for quite some time, and I'm excited to see it fixed!

PR #8409, opened 2026-02-04
enh: Use direct `html.escape` import

hoxbro:
This confused me. When seeing this `html.escape` import I instantly knows this is from the stdlib.

I have left the original import for now.


philippjfr (approved):
Thanks, this was left over from a time when `html.escape` behaved different for different python versions.

PR #8418, opened 2026-02-11
Add es_module_shims config option to disable es-module-shims polyfill

nojaf:
This is a follow-up of https://github.com/holoviz/panel/pull/8408

In our company everybody is on the latest chrome and we don't really need this es-module-shim.
Would be great if we had a flag to just skip it all together.

### DIFF
--- a/doc/api/config.md
+++ b/doc/api/config.md
@@ -180,6 +180,14 @@ Where to save json files for embedded state.
 
 Default: './' | Type: String
 
+### `es_module_shims` (`PANEL_ES_MODULE_SHIMS`)
+
+Whether to load the es-module-shims polyfill. Set to False
+if you only target browsers with native import map support
+(Chrome 89+, Firefox 108+, Safari 16.4+).
+
+Default: True | Type: Boolean
+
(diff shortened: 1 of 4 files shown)


nojaf:
Hi @philippjfr, any thoughts on this? Does this seem reasonable?

hoxbro:
I don't see any problem with this PR, but will let Philipp decide when he is back next week.

nojaf:
Got it, thanks for the reply @hoxbro!

philippjfr:
Just getting back after being OOO for a little while. I'm not convinced this works as is today, specifically we use the `importShim` for dynamically loading ESM modules in `ReactiveESM` and we'd have to handle that using native APIs (if that's even possible) before we can merge this.

philippjfr:
Yeah, sorry @nojaf. I'm going to have to close unfortunately. Despite the fact that modern browsers support most of the features implemented by the shim, none of the native API emulate the `importShim` function that implements the resolution of specifiers and URLs.

nojaf:
No worries, it was worth a shot.

PR #8424, opened 2026-02-19
fix: LaTeX rendering in Markdown pane for Pyodide

### LINKED ISSUE #8421: markdown latex not working in pyodide
You can reproduce by opening https://panel.holoviz.org/reference/panes/Markdown.html#latex-support and running the example.

Image

I see the same problem when trying to use latex in markdown in pyodide.

AR21SM:
Closes #8421

## Problem

`bokeh-mathjax` was excluded from the JS bundle in Pyodide/convert contexts,
causing `$$…$$` content to render as raw text. Bokeh's `_use_mathjax()` only
inspects its own model types (Div, TextAnnotation, etc.) and does not recognise
Panel's `HTML` model, which `Markdown`, `HTML`, and `Str` panes use.

## Fix

Added `_panel_use_mathjax(roots)` in `panel/io/resources.py` that detects
Panel's `HTML` Bokeh model using bokeh's own `_any()` tree walker. 

```python
def _panel_use_mathjax(roots) -> bool:
    """Whether any model in roots is a Panel HTML model (may need MathJax)."""
    from ..models.markup import HTML as PanelHTML
    from bokeh.embed.bundle import _any
    return _any(roots, lambda obj: isinstance(obj, PanelHTML))
```

### DIFF
--- a/panel/io/resources.py
+++ b/panel/io/resources.py
@@ -25,7 +25,7 @@ from typing import (
 import bokeh.embed.wrappers
 
 from bokeh.embed.bundle import (
-    CSS_RESOURCES as BkCSS_RESOURCES, URL, Bundle as BkBundle,
+    CSS_RESOURCES as BkCSS_RESOURCES, URL, Bundle as BkBundle, _any,
     _bundle_extensions, _use_mathjax, bundle_models, extension_dirs,
 )
 from bokeh.model import Model
@@ -37,6 +37,7 @@ from jinja2.loaders import FileSystemLoader
 from markupsafe import Markup
 
 from ..config import config, panel_extension as extension
+from ..models.markup import HTML as PanelHTML
 from ..util import _descendents, isurl, url_path
 from .state import state
 
@@ -449,6 +450,11 @@ def bundled_files(model: Model, file_type: str = 'javascript') -> list[str]:
             files.append(url)
     return files
 
+def _panel_use_mathjax(roots) -> bool:
+    """Whether any model in roots is a Panel HTML model (may need MathJax)."""
+    return _any(roots, lambda obj: isinstance(obj, PanelHTML))
+
+
 def bundle_resources(
     roots,
     resources: BkResources,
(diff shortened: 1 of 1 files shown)


panel/io/resources.py
```diff
@@ -449,6 +449,13 @@ def bundled_files(model: Model, file_type: str = 'javascript') -> list[str]:
             files.append(url)
     return files
 
+def _panel_use_mathjax(roots) -> bool:
+    """Whether any model in roots is a Panel HTML model (may need MathJax)."""
+    from ..models.markup import HTML as PanelHTML
+    from bokeh.embed.bundle import _any
```
philippjfr: I think we don't need to inline these.

AR21SM: Fair point, I’ll clean it up.

philippjfr: Thank you!!

philippjfr:
Thanks for catching this. Apart from my small nit (and the failing linter) this looks good and I'll merge once addressed.

AR21SM:
> Thanks for catching this. Apart from my small nit (and the failing linter) this looks good and I'll merge once addressed.

done :)

AR21SM:
I rebased the branch maybe it wasn’t required, but I think it’s fine.

panel/io/resources.py
```diff
@@ -37,6 +37,7 @@
 from markupsafe import Markup
 
 from ..config import config, panel_extension as extension
+from ..models.markup import HTML as PanelHTML
```
philippjfr: Looks like we do have to inline this one :/

```suggestion
```

panel/io/resources.py
```diff
@@ -449,6 +450,11 @@ def bundled_files(model: Model, file_type: str = 'javascript') -> list[str]:
             files.append(url)
     return files
 
+def _panel_use_mathjax(roots) -> bool:
+    """Whether any model in roots is a Panel HTML model (may need MathJax)."""
+    return _any(roots, lambda obj: isinstance(obj, PanelHTML))
```
philippjfr, line 455:
```suggestion
    from ..models.markup import HTML as PanelHTML
    return _any(roots, lambda obj: isinstance(obj, PanelHTML))
```

PR #8427, opened 2026-02-20
Fix embed_state merging for same-name Select widgets

### LINKED ISSUE #8335: `widget.Select` not working in exported HTML files if moren than one is present
<details>

```plaintext
bokeh==3.8.1
panel==1.8.4
python==3.10.18

Misc:
SO: Windows 10
browser: Edge (but also tried with Chrome)
```
</details>

Any HTML-based file that contains more than 1 dropdown-based widget (`Select`) will show an erratic behaviour. Usually, one of them is working and the other is like "frozen". There seems to be some kind of cross dependency as selecting an option in one of them also changes the selector on the other one. **This is working totally fine when I use `show()` instead of `save()`**

I am aware that exporting to HTML reduces quite a lot the functionalities of a dashboard. But if 1 widget is working, I would expect multiple ones to be working very similarly.

Different alternatives I tried:

- Many different options within `save`.
- Re-written the core with `pn.bind`
- Inserting any other kind of content inside. Not only plots. But the issue is always there even if there is only text.

It can be run with `uv run script.py`

```python
# /// script
# requires-python = "==3.10.18"
# dependencies = [
#     "bokeh==3.8.1",
#     "panel==1.8.4",
# ]
# ///

import os
from typing import Any

import bokeh
import panel as pn
from bokeh.plotting import figure

print(bokeh.__version__)
print(pn.__version__)


def group_plots_with_selector(plots: dict[str, Any]) -> pn.Column:
    """Create a selector with content."""
    selector = pn.widgets.Select(name="Select Plot", options=list(plots.keys()))
    plots_panelized = {name: pn.panel(plot) for name, plot in plots.items()}

    @pn.depends(selector.param.value)  # type: ignore
    def view_plot(plot_name: str) -> Any:
        return plots_panelized[plot_name]

    return pn.Column(selector, view_plot)


def create_simple_plot(title: str, color: str) -> figure:
    """Create a simple plot."""
    p = figure(width=400, height=300, title=title)
    p.line([1, 2, 3, 4, 5], [1, 4, 9, 16, 25], line_width=2, color=color)
    p.scatter([1, 2, 3, 4, 5], [1, 4, 9, 16, 25], size=8, color=color)
    return p


# First dropdown with plots
plots_1 = {
    "Plot A": create_simple_plot("Plot A", "blue"),
    "Plot B": create_simple_plot("Plot B", "red"),
    "Plot C": create_simple_plot("Plot C", "green"),
}

# Second dropdown with different plots
plots_2 = {
    "Plot X": create_simple_plot("Plot X", "purple"),
    "Plot Y": create_simple_plot("Plot Y", "orange"),
}

# Create two independent selectors
selector_1 = group_plots_with_selector(plots_1)
selector_2 = group_plots_with_selector(plots_2)

# Combine both selectors
dashboard = pn.Column(
    "# Bug: Multiple Dropdowns in HTML Export",
    "## Dropdown 1",
    selector_1,
    "---",
    "## Dropdown 2",
    selector_2,
    "---",
    "**Issue**: Both dropdowns work in the Panel app, but only one works in HTML export.",
)

# Save to HTML
embed_states = {
    selector_1[0]: {"value": "Plot A"},  # First selector widget
    selector_2[0]: {"value": "Plot X"},  # Second selector widget
}
dashboard.save("sample.html", embed=True, max_states=10)
os.startfile("sample.html")
```

Since the problem is only HTML-based, I am not sure how to retrieve this information. Sorry about it!

ujjwallsrivastavaa:
## Summary
This PR fixes an embedding bug where independent `Select` widgets with the same display name were incorrectly merged during state export.

## Problem
`embed_state` currently merges widgets by shared `name`.  
That behavior is useful for intentionally synchronized widgets, but incorrect for independent `Select` widgets that commonly reuse labels (e.g. "test", "category").

As a result, embedded state generation can miss valid combinations for independent selects.

## Solution
- Import `Select` in `panel/io/embed.py`
- Exclude `Select` widgets from the name-based merge path
- Keep existing merge behavior unchanged for other widget types

## Tests
Added regression test in `panel/tests/io/test_embed.py`:

- `test_embed_selects_with_same_name_not_merged`

This test verifies two independent same-name `Select` widgets are embedded as a full cross-product of states, rather than being merged.

## Why this is safe
- Scope is narrow and type-specific (`Select` only)
- Existing merge semantics for non-`Select` widgets are preserved
- Regression coverage ensures the reported case remains fixed

Fixes #8335

### DIFF
--- a/panel/io/embed.py
+++ b/panel/io/embed.py
@@ -220,7 +220,7 @@ def embed_state(panel, model, doc, max_states=1000, max_opts=3,
     from ..links import Link
     from ..models.state import State
     from ..pane import PaneBase
-    from ..widgets import DiscreteSlider, Widget
+    from ..widgets import DiscreteSlider, Select, Widget
 
     ref = model.ref['id']
     if isinstance(panel, PaneBase) and ref in panel.layout._models:
@@ -279,8 +279,10 @@ def embed_state(panel, model, doc, max_states=1000, max_opts=3,
             if not isinstance(w_model, w_type):
                 w_model = w_model.select_one({'type': w_type})
 
-        # If there is a widget with the same name, merge with it
-        if widget.name and widget.name in merged:
+        # If there is a widget with the same name, merge with it.
+        # Select widgets are excluded because independent selects often
+        # share display names but should not be synchronized.
+        if widget.name and widget.name in merged and not isinstance(widget, Select):
             merged[widget.name][0].append(w)
             merged[widget.name][1].append(w_model)
             continue
--- a/panel/tests/io/test_embed.py
+++ b/panel/tests/io/test_embed.py
@@ -559,7 +559,7 @@ def test_embed_selects_with_same_name_not_merged(document, comm):
 
     panel = Row(s1, s2, t1, t2)
     with config.set(embed=True):
-      model = panel.get_root(document, comm)
+        model = panel.get_root(document, comm)
     state_model = embed_state(panel, model, document)
 
     assert state_model is not None
--- a/panel/tests/io/test_embed.py
+++ b/panel/tests/io/test_embed.py
@@ -548,6 +548,26 @@ def test_embed_merged_sliders(document, comm):
     ]
 
 
+def test_embed_selects_with_same_name_not_merged(document, comm):
+    s1 = Select(name='test', options=['A', 'B'])
+    t1 = StaticText()
+    s1.param.watch(lambda event: setattr(t1, 'value', event.new), 'value')
+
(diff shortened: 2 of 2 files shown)


philippjfr:
I don't think this actually fixes anything. Won't this just overwrite one widget state with another? I think we'd have to change things so instead of keying off the widget name we'd have to key off an ID or something.

philippjfr:
Appreciate the contribution @ujjwallsrivastavaa, but in it's current state the change does not address the underlying issue here. Will close for now.

PR #8433, opened 2026-02-23
Fix resource resolution when pathname changes

philippjfr:
**Summary**
Changing the browser pathname early in the lifecycle can cause resources created just before the update to resolve against the wrong base URL. This PR ensures that resources generated prior to a pathname change continue to resolve relative to the original path, and keeps Panel’s internal relative path in sync for future resources.

**Changes**

* Gate URL updates on the Bokeh document idle signal so the initial batch of resources resolves against the original pathname.
* Update `pn.state.rel_path` when `pn.state.location.pathname` changes to ensure subsequently generated relative URLs resolve correctly.
* Use the History API for non-reload URL updates.

**Why**

Without this change:

* Early pathname updates can shift the document base too soon.
* Dynamically injected resources (JS/CSS/static) may resolve relative to the new path instead of the original one.
* Future resource generation can become inconsistent with the displayed URL.

**Result**

* Resources created before a pathname change resolve correctly.
* Resources created after the change use the updated relative base.
* URL updates remain SPA-friendly without unintended navigation side effects.

**Notes**

* The idle gating only affects the initial render window; subsequent updates proceed immediately.
* Backward compatible for existing Panel apps.

### DIFF
--- a/panel/io/location.py
+++ b/panel/io/location.py
@@ -140,20 +140,24 @@ class Location(Syncable):
     def _sync_pathname(self):
         """
         When in a server context and the pathname is overridden dynamically
-        we need to keep the state.rel_path in sync to ensure that resources
-        are resolved relative to the new URL.
+        we need to keep the state.rel_path and the Document dist_url template
+        variable in sync to ensure that resources are resolved relative to the new URL.
+        Additionally we clear the stylesheets cache which holds now outdated URLs.
         """
         session_context = state.curdoc.session_context
         if not (state._servers and session_context and session_context.server_context):
             return
         for server, _, _ in state._servers.values():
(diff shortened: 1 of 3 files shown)


panel/io/location.py
```diff
@@ -136,6 +136,26 @@ def _get_model(
...
+        When in a server context and the pathname is overridden dynamically
+        we need to keep the state.rel_path in sync to ensure that resources
+        are resolved relative to the new URL.
+        """
+        session_context = state.curdoc.session_context
+        if not (state._servers and session_context and session_context.server_context):
+            return
+        for server, _, _ in state._servers.values():
+            server_port = server.port if hasattr(server, 'port') else server.config.port
+            breakpoint()
```
hoxbro:
```suggestion
```

PR #8436, opened 2026-02-24
Ensure we only de-duplicate param value events

philippjfr:
Currently the `Param` pane tracks parameter events and blocks events from bouncing back and forth by tracking when a parameter value is changing. However we track this for all parameter attribute changes, including attributes like constant, label etc., and treat all of these the same. This means that when constant is toggled it can occur that we end up dropping events for another attribute or even the value. We now only apply this handling when the value changes.

### DIFF
--- a/panel/param.py
+++ b/panel/param.py
@@ -654,6 +654,8 @@ class Param(Pane):
             else:
                 updates['value'] = change.new
 
+            # Ensure we only de-duplicate value events since only
+            # the widget value can be controlled from the frontend 
             value_update = 'value' in updates or 'value_throttled' in updates
             try:
                 if value_update:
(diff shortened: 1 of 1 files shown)


panel/param.py
```diff
@@ -654,8 +654,10 @@ def action(event):
             else:
                 updates['value'] = change.new
 
+            value_update = 'value' in updates or 'value_throttled' in updates
```
philippjfr, line 659:
```suggestion
            # Ensure we only de-duplicate value events since only
            # the widget value can be controlled from the frontend 
            value_update = 'value' in updates or 'value_throttled' in updates
```

panel/param.py
```diff
@@ -654,8 +654,12 @@ def action(event):
             else:
                 updates['value'] = change.new
 
+            # Ensure we only de-duplicate value events since only
+            # the widget value can be controlled from the frontend
```
```suggestion
            # the widget value can be controlled from the frontend
```

PR #8441, opened 2026-02-25
Fix Number and String indicators defaulting to black in dark theme

### LINKED ISSUE #8373: `pn.indicators.Number` defaults to black font on value change in dark theme

<details>

```plaintext
panel=1.8.5
```
</details>

The text color should be persisted by default on value change for dark theme as well as for default. Currently when a value is changed the font color falls back to black, which makes it unreadable with the dark theme enabled.
The issue can be somewhat fixed by manually overwriting css rules on updates with `!important` rule set, but it is quite messy. Issue shows up both for reactive and bounded values.

I have only tested this for number indicators, but it could be similar for the others if using the same style updating code.

#### Complete, minimal, self-contained example code that reproduces the issue there as well.

```python
import panel as pn

pn.config.theme = 'dark'

btn_rx = pn.widgets.Button(name="rx++")
btn_val = pn.widgets.Button(name="val++")

ind_rx = pn.indicators.Number(name="rx", value=pn.rx(btn_rx.param.clicks))
ind_val = pn.indicators.Number(name="val", value=0)

def update_val_indicator(event):
    if not event:
        return
    ind_val.value += 1
    # Possible naive manual fix
    # ind_val.stylesheets = ["""
    #     div:nth-child(2) { color: white !important; }
    # """]

pn.bind(update_val_indicator, btn_val, watch=True)

pn.Column(btn_rx, btn_val, ind_rx, ind_val).servable()

```

\-

##### Before

Image

##### After

Image

AtharvaJaiswal005:
## Summary

Fixes Number and String indicators rendering black text in dark theme, making the value unreadable after updates.

Closes #8373

## Key Changes

**`panel/widgets/indicators.py`**
- Changed `default_color` from `'black'` to `'currentcolor'` on `Number` indicator
- Changed `default_color` from `'black'` to `'currentcolor'` on `String` indicator
- `currentcolor` is a CSS value that inherits the text color from the parent element, so it automatically adapts to both light and dark themes

**`panel/tests/widgets/test_indicators.py`**
- Updated `test_number_none` assertions to match new default color

## Tests

All 8 existing indicator tests pass. No regressions.

### DIFF
--- a/panel/tests/widgets/test_indicators.py
+++ b/panel/tests/widgets/test_indicators.py
@@ -10,11 +10,11 @@ def test_number_none(document, comm):
 
     model = number.get_root(document, comm)
 
-    assert model.text.endswith("&lt;div style=&quot;font-size: 54pt; color: black&quot;&gt;-&lt;/div&gt;")
+    assert model.text.endswith("&lt;div style=&quot;font-size: 54pt; color: currentcolor&quot;&gt;-&lt;/div&gt;")
 
     number.nan_format = 'nan'
 
-    assert model.text.endswith("&lt;div style=&quot;font-size: 54pt; color: black&quot;&gt;nan&lt;/div&gt;")
+    assert model.text.endswith("&lt;div style=&quot;font-size: 54pt; color: currentcolor&quot;&gt;nan&lt;/div&gt;")
 
 
 def test_number_thresholds(document, comm):
--- a/panel/widgets/indicators.py
+++ b/panel/widgets/indicators.py
@@ -331,7 +331,7 @@ class Number(ValueIndicator):
     >>> Number(name='Rate', value=72, format='{value}%', colors=[(80, 'green'), (100, 'red')]
     """
 
-    default_color = param.String(default='black', doc="""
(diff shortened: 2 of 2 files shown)


philippjfr:
This does seem like the best solution.

philippjfr:
Thanks @AtharvaJaiswal005! Test failures unrelated and the PR is simple enough. One quick note, we appreciate the PRs particularly for well scoped issues like this. For future reference, if you used AI assistance for a PR could you disclose that and ideally also indicate the model used.

AtharvaJaiswal005:
Thanks for the quick merge @philippjfr! For transparency - I use CodeRabbit for automated code review and Antigravity (which has models like Opus 4.6 and Gemini 3.1) as a coding assistant to help explore the codebase and iterate faster. The core understanding of the issue - `default_color` being hardcoded to `black` instead of using CSS `currentcolor` to inherit from the parent theme - was straightforward once I traced how the Number and String indicators render their styles. Will make sure to disclose AI tools on future PRs. Happy to keep contributing!

PR #8442, opened 2026-02-25
Prevent spacebar in Card header inputs from toggling collapse

### LINKED ISSUE #8358: Card Collapse/Expand triggered by Spacebar entry in a TextInput in Card Header
Issue #7045 reported that a Card would collapse/expand when a TextInput was placed in the Card Header, and the user *either* clicked *or* pressed spacebar in this item.  This was *partially* fixed by PR #7057 in a way which caught *clicks* from propagating, but **did not** catch spacebar entries in the textfield, which would still trigger collapse.

Based on my limited understanding, it seems that the default base HTMLTextInput object emulates click-like behavior when spacebar is pressed.  I don't understand enough of the raw DOM behavior to understand why this isn't caught by the fix in #7057.

<details>

```plaintext
panel                                   1.8.5.post1
```
</details>

**Expected**: A TextField in the Card Header can accept arbitrary text

**Observed**: Pressing spacebar in the TextField in a Card Header will toggle expand/collapse behavior.  E.g. if a user types "a quick brown fox" in a Textfield in the Header of an expanded Card, the user will see the card collapse, expand, and collapse again (ending in collapsed state), without having clicked anywhere else.

Updating `test_card_widget_not_collapsed` with the line `text_input.press("Space")` will cause the test to fail on the assert; the card will have collapsed. See below for a full version of the test with the new line, as of 1.8.5.post1

```python
def test_card_widget_not_collapsed(page, card_components):
    # Fixes https://github.com/holoviz/panel/issues/7045
    w1, w2 = card_components
    card = Card(w1, header=Row(w2))

    serve_component(page, card)

    text_input = page.locator('.bk-input[type="text"]')
    expect(text_input).to_have_count(1)

    text_input.click()

    text_input.press("F")
    text_input.press("Space")  # This is the only newline
    text_input.press("Enter")

    wait_until(lambda: w2.value == 'F', page)
    assert not card.collapsed  # This will now fail

```

AtharvaJaiswal005:
Fixes #8358

## Summary

PR #7057 fixed click propagation from interactive widgets in Card headers, but spacebar keydown events still propagated to the Card's collapse toggle. When a `TextInput` is placed in a Card header, pressing spacebar toggles the Card's collapse/expand state instead of inserting a space character.

**Root cause:** The Card header is a `<button>` element. When spacebar is pressed inside a nested `TextInput`, the `keyup` event bubbles to the button, which triggers the browser's native button activation (synthetic `click`). The existing `composedPath()` check in `_toggle_button` doesn't catch this because the synthetic click's path starts at the button, not the input.

**Fix:** Added a `keyup` event listener on the button that calls `preventDefault()` when spacebar originates from an `HTMLInputElement` (checked via `composedPath()`), preventing the synthetic click while allowing the space character to be typed normally.

## Changes
- `panel/models/card.ts` - Added `keyup` listener to prevent spacebar-triggered button activation from input elements
- `panel/tests/ui/layout/test_card.py` - Added `test_card_widget_spacebar_not_collapsed` test

## Test plan
- [ ] New test `test_card_widget_spacebar_not_collapsed` verifies typing "F B" (with space) in a TextInput in the Card header does not collapse the card
- [ ] Existing `test_card_widget_not_collapsed` still passes (click + Enter behavior unchanged)


panel/chat/interface.py
```diff
@@ -115,6 +115,10 @@ class ChatInterface(ChatFeed):
     user = param.String(default="User", doc="""
         Name of the ChatInterface user.""")
 
+    placeholder = param.String(default="Send a message", doc="""
```
philippjfr: This code is unrelated.

panel/models/card.ts
```diff
@@ -99,6 +99,16 @@ export class CardView extends ColumnView {
       header.el.style.backgroundColor = header_background != null ? header_background : ""
       this.button_el.appendChild(header.el)
       this.button_el.addEventListener("click", (e: MouseEvent) => this._toggle_button(e))
+      this.button_el.addEventListener("keyup", (e: KeyboardEvent) => {
+        if (e.key === " ") {
```
philippjfr:
```suggestion
        if (event.code === "Space") {
```

AtharvaJaiswal005:
Removed the unrelated ChatInterface changes and applied the `event.code === "Space"` suggestion. Sorry about the mix-up! Let me know if there's anything else that needs adjusting - happy to help.

panel/models/card.ts
```diff
@@ -99,6 +99,16 @@ export class CardView extends ColumnView {
       header.el.style.backgroundColor = header_background != null ? header_background : ""
       this.button_el.appendChild(header.el)
       this.button_el.addEventListener("click", (e: MouseEvent) => this._toggle_button(e))
+      this.button_el.addEventListener("keyup", (e: KeyboardEvent) => {
+        if (event.code === "Space") {
```
philippjfr:
```suggestion
        if (e.code === "Space") {
```

PR #8443, opened 2026-02-25
Fix: Pipeline "next" button stays disabled on back navigation when ready_parameter is already True (#8378)

### LINKED ISSUE #8378: Mismatch between pipeline's "next" button `disabled` state and stage's `ready_parameter` value
<details>

```plaintext
panel 1.8.5
```
</details>

It is possible to set a `ready_parameter` to a stage when some condition has to be met to proceed. Until it's set to `True`, the "next" button is disabled. I would expect that this behaviour would remain if you've proceeded to the next stage and then go back to the original stage.

Instead, what happens is that when the `ready_parameter` is set to `True`, the "next" button stays disabled so you can't proceed even though you should. On the second time it's set to `True` though, the "next" button is finally enabled. In fact, since the stage's state has not changed when you go back to it, the "next" button should not be disabled at all.

If I'm unclear, please refer to the attached screencast.

I took the code from the docs, added a `ready_parameter` to the first stage and a button that set it to `True` when clicked.

```python
import panel as pn
import param

pn.extension("katex")

pipeline = pn.pipeline.Pipeline()


class Stage1(param.Parameterized):
    a = param.Integer(default=2, bounds=(0, 10))
    b = param.Integer(default=3, bounds=(0, 10))
    ready = param.Boolean(default=False)

    @param.output(("c", param.Integer), ("d", param.Integer))
    def output(self):
        return self.a * self.b, self.a**self.b

    @param.depends("a", "b")
    def view(self):
        c, d = self.output()
        c_out = pn.pane.LaTeX(
            "${a} * {b} = {c}$".format(a=self.a, b=self.b, c=c),
            styles={"font-size": "2em"},
        )
        d_out = pn.pane.LaTeX(
            "${a}^{{{b}}} = {d}$".format(a=self.a, b=self.b, d=d),
            styles={"font-size": "2em"},
        )
        button = pn.widgets.Button(
            name="Ready",
            button_type="success",
            on_click=lambda x: setattr(self, "ready", True),
        )
        return pn.Column(
            c_out,
            d_out,
            button,
            margin=(40, 10),
            styles={"background": "#f0f0f0"},
        )

    def panel(self):
        return pn.Row(
            self.param,
            self.view,
        )


    c = param.Integer(default=6, bounds=(0, None))
    exp = param.Number(default=0.1, bounds=(0, 3))

    @param.depends("c", "exp")
    def view(self):
        out = pn.pane.LaTeX(
            "${%s}^{%s}={%.3f}$" % (self.c, self.exp, self.c**self.exp),
            styles={"font-size": "2em"},
        )
        return pn.Column(out, margin=(40, 10), styles={"background": "#f0f0f0"})

    def panel(self):
        return pn.Row(self.param, self.view)


pipeline.add_stage("Stage 1", Stage1, ready_parameter="ready")
pipeline.add_stage("Stage 2", Stage2)

pipeline.servable()
```

https://github.com/user-attachments/assets/3e2985aa-78a7-4074-8939-e4075100b863

DevilSpecial:
# Fix: Pipeline "next" button stays disabled on back navigation when `ready_parameter` is already `True`

Fixes #8378

---

## Summary

When navigating back to a pipeline stage whose `ready_parameter` is already `True`, the "Next" button was incorrectly displayed as **disabled**, requiring the user to set the parameter to `True` a second time before being able to proceed.

---

## Root Cause

In `_update_button` inside `pipeline.py`, the `ready_parameter` value was being read from **`stage`** (the registered class or original class-level instance) instead of **`self._state`** (the live parameterized instance the user is interacting with):

```python
# BEFORE — reads class-level default (always False), ignores instance value
disabled = (not getattr(stage, ready)) if ready in stage.param else False
```

When the user navigates back, `_update_button` re-runs and resets the button to `disabled=True` using the class **default** value, completely ignoring the instance's **actual** current value (`True`).

---

## Steps to Reproduce

```python
import panel as pn
import param

pn.extension("katex")
pipeline = pn.pipeline.Pipeline()

class Stage1(param.Parameterized):
    ready = param.Boolean(default=False)
    def panel(self):
        return pn.widgets.Button(
            name="Ready",
            on_click=lambda e: setattr(self, "ready", True)
        )

    def panel(self):
        return pn.pane.Str("Stage 2")

pipeline.add_stage("Stage 1", Stage1, ready_parameter="ready")
pipeline.add_stage("Stage 2", Stage2)
pipeline.servable()
```

1. Click **"Ready"** → Next button enables ✅
2. Click **"Next"** to advance to Stage 2
3. Click **"Previous"** to go back to Stage 1
4. **Bug**: Next button is **disabled** ❌ even though `ready` is still `True`
5. Click "Ready" again → Next button finally enables (requires a second click)

---

## Fix

One-line change in `_update_button` ([`panel/pipeline.py`](panel/pipeline.py)):

```diff
- disabled = (not getattr(stage, ready)) if ready in stage.param else False
+ disabled = (not getattr(self._state, ready)) if ready in self._state.param else False
```

`self._state` is always the live param instance, so this correctly reflects whatever value the user has already set on that instance.

---

## Changes

| File | Change |
|------|--------|
| `panel/pipeline.py` | L389: `getattr(stage, ready)` → `getattr(self._state, ready)` |
| `panel/tests/test_pipeline.py` | Added regression test `test_pipeline_ready_parameter_preserved_on_back_navigation` |

---

## Regression Test Added

```python
def test_pipeline_ready_parameter_preserved_on_back_navigation():
    """
    Regression test for https://github.com/holoviz/panel/issues/8378.

    When navigating back to a stage whose ready_parameter is already True,
    the 'next' button should remain enabled (not disabled).
    Previously, _update_button read from the stage class (always default=False)
    instead of the live instance (self._state), causing the button to be
    incorrectly disabled until the parameter was set to True a second time.
    """
    pipeline = Pipeline(ready_parameter='ready')
    pipeline.add_stage('Stage 1', Stage1)
    pipeline.add_stage('Stage 2', Stage2)

    # Initially disabled because ready=False
    assert pipeline.next_button.disabled

    # Set ready=True → button enables
    pipeline._state.ready = True
    assert not pipeline.next_button.disabled

    # Advance to Stage 2
    pipeline._next()
    assert isinstance(pipeline._state, Stage2)

    # Go back to Stage 1 — ready is still True on the instance
    pipeline._previous()
    assert isinstance(pipeline._state, Stage1)
    assert pipeline._state.ready is True

    # The next button must be enabled immediately (not disabled)
    assert not pipeline.next_button.disabled
```

---

The fix is minimal, surgical, and does not affect any other pipeline behaviour. All currently passing tests remain valid. Slight uncertainty in merge probability accounts for potential CHANGELOG requirements or maintainer preference for additional guard checks (e.g. `self._state is not None`).

### DIFF
--- a/panel/pipeline.py
+++ b/panel/pipeline.py
@@ -386,7 +386,7 @@ class Pipeline(Viewer):
             self.next_button.disabled = True
         else:
             ready = kwargs.get('ready_parameter', self.ready_parameter)
-            disabled = (not getattr(stage, ready)) if ready in stage.param else False
+            disabled = (not getattr(self._state, ready)) if ready in self._state.param else False
             self.next_button.disabled = disabled
 
     def _get_error_button(self, e):
--- a/panel/tests/test_pipeline.py
+++ b/panel/tests/test_pipeline.py
@@ -411,3 +411,37 @@ def test_pipeline_repr():
 
     pipeline.add_stage('Stage 2', Stage2)
     assert repr(pipeline) == 'Pipeline:\n    [0] Stage 1: Stage1()\n    [1] Stage 2: Stage2()'
+
+
+def test_pipeline_ready_parameter_preserved_on_back_navigation():
+    """
+    Regression test for https://github.com/holoviz/panel/issues/8378.
+
+    When navigating back to a stage whose ready_parameter is already True,
+    the 'next' button should remain enabled (not disabled).
+    Previously, _update_button read from the stage class (always default=False)
+    instead of the live instance (self._state), causing the button to be
+    incorrectly disabled until the parameter was set to True a second time.
+    """
+    pipeline = Pipeline(ready_parameter='ready')
+    pipeline.add_stage('Stage 1', Stage1)
+    pipeline.add_stage('Stage 2', Stage2)
+
+    # Initially disabled because ready=False
+    assert pipeline.next_button.disabled
+
+    # Set ready=True → button enables
+    pipeline._state.ready = True
+    assert not pipeline.next_button.disabled
+
+    # Advance to Stage 2
+    pipeline._next()
+    assert isinstance(pipeline._state, Stage2)
+
+    # Go back to Stage 1 — ready is still True on the instance
+    pipeline._previous()
+    assert isinstance(pipeline._state, Stage1)
+    assert pipeline._state.ready is True
+
+    # The next button must be enabled immediately (not disabled)
+    assert not pipeline.next_button.disabled


philippjfr (approved):

philippjfr:
Thank you @DevilSpecial, good catch!

PR #8454, opened 2026-03-01
fix: markdown docs render

### LINKED ISSUE #8422: Markdown reference guide not rendered
When I open https://panel.holoviz.org/reference/panes/Markdown.html I see no rendered output

Image

I believe this is a bug.

AR21SM:
Closes #8422 
The page was initializing `MathJax` for every example at the top, even though only LaTeX needs it.
That made normal Markdown examples wait on `MathJax` and sometimes appear blank until manual Run.
I moved pn.extension('mathjax') to the LaTeX section only, where it is actually required.
Now regular Markdown renders immediately, and LaTeX rendering still works as expected.
image

however, some lower cells in this page still require clicking Run to render.
image
i think what we can do is move the LaTeX block to the bottom (move New Lines section above) (but LaTeX section still need manual click :( )

### DIFF
--- a/examples/reference/panes/Markdown.ipynb
+++ b/examples/reference/panes/Markdown.ipynb
@@ -10,7 +10,7 @@
    "source": [
     "import panel as pn\n",
     "\n",
-    "pn.extension('mathjax')"
+    "pn.extension()"
    ]
   },
   {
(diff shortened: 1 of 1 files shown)


AR21SM:
>however, some lower cells in this page still require clicking Run to render.
>i think what we can do is move the LaTeX block to the bottom (move New Lines section above) (but LaTeX section still need manual click :( )

@philippjfr @MarcSkovMadsen any suggestions

philippjfr:
Appreciate your help here @AR21SM. Tracking that down helped me fix the actual issue in `nbsite`, which was that MathJax wasn't loaded in the docs and nothing would render until it was loaded (which is never).

You can see the fixed build here:

https://holoviz-dev.github.io/panel/reference/panes/Markdown.html

PR #8458, opened 2026-03-02
docs: Update changelog with correct names

hoxbro:
Looks like the wrong username were used for the last changelog this should fix that.


SuMayaBee:
Thanks @hoxbro for fixing the changelog! I noticed `doc/about/releases.md` needs the same fix, so I've addressed it in my PR #8459.

PR #8462, opened 2026-03-03
Only serialize ESM model definition if session is already loaded

philippjfr:
The fix introduced in https://github.com/holoviz/panel/pull/8432 caused a regression because it was not correctly scoped. Specifically we should only serialize model definitions if the session is already fully initialized. This change adds an additional condition using `state.loaded` to ensure that is the case.


panel/custom.py
```diff
@@ -188,7 +188,7 @@ def __init__(mcs, name: str, bases: tuple[type, ...], dict_: Mapping[str, Any]):
             p for p in Reactive.param
             if not issubclass(type(mcs.param[p].owner), ReactiveESMMetaclass)
         ]
-        mcs._data_model__initialized = not (state.curdoc and state.curdoc.session_context)
+        mcs._data_model__initialized = not (state.curdoc and state.curdoc.session_context and state.loaded)
```
```suggestion
        mcs._data_model__initialized = not (state.curdoc and state.curdoc.session_context and state._connected.get(state.curdoc, False))
```

PR #8468, opened 2026-03-04
fix datetime axis type mismatch in link_axes after scroll + plot update

### LINKED ISSUE #7795: Holoviews panels and axis sharing after plot update
panel==1.6.1

<details>

```plaintext
holoviews==1.20.1
bokeh==3.6.3
hvplot==-0.11.2
```
</details>

When two or more holoviews plots are on the same page and the axes are linked.  If you change the view (for example with mouse scroll) on one, then updated the data and the plots, it returns an error.  Steps are as follows:
- Show two holoviews panes
- Scroll to changes axes limits, both adjust as expected
- Refresh the data and update the plots. 

```python
import panel as pn
import pandas as pd
import hvplot.pandas
import holoviews as hv
import numpy as np
pn.extension('tabulator',sizing_mode='stretch_width')
hv.extension('bokeh', theme='dark')


class MREViewer(pn.viewable.Viewer):

    plot1 = pn.pane.HoloViews(sizing_mode='stretch_width')
    plot2 = pn.pane.HoloViews(sizing_mode='stretch_width')
    reload_data_button = pn.widgets.Button(name='Reload Data')

    @pn.depends('reload_data_button.clicks', watch=True,on_init=True))
    def update_plots(self):
        dates = pd.date_range('2021-01-01', periods=365)
        data = pd.DataFrame({'Date': dates, 'Value': np.random.randn(365).cumsum()})

        plot1 = data.hvplot(x='Date', y='Value', title='Plot 1')
        plot2 = data.hvplot(x='Date', y='Value', title='Plot 2')
        self.plot1.object = plot1
        self.plot2.object = plot2

    def __panel__(self):
        # Include the reload_data_button in the layout
        layout = pn.Column(self.reload_data_button, self.plot1, self.plot2)
        return layout

app = MREViewer()
app.servable()

```

 message: Message 'PATCH-DOC' content: {'events': [{'kind': 'MessageSent', 'msg_type': 'bokeh_event', 'msg_data': {'type': 'event', 'name': 'button_click', 'values': {'type': 'map', 'entries': [['model', {'id': 'p7196'}]]}}}]} 
 error: UFuncTypeError(<ufunc 'greater'>, (<class 'numpy.dtypes.DateTime64DType'>, <class 'numpy.dtypes._PyFloatDType'>, None))
numpy.exceptions.DTypePromotionError: The DType <class 'numpy.dtypes.DateTime64DType'> could not be promoted by <class 'numpy.dtypes._PyFloatDType'>. This means that no common DType exists for the given inputs. For example they cannot be stored in a single array unless the dtype is `object`. The full list of DTypes is: (<class 'numpy.dtypes.DateTime64DType'>, <class 'numpy.dtypes._PyFloatDType'>)

The above exception was the direct cause of the following exception:

...

if axis.start > axis.end:
       ^^^^^^^^^^^^^^^^^^^^^
numpy._core._exceptions._UFuncNoLoopError: ufunc 'greater' did not contain a loop with signature matching types (<class 'numpy.dtypes.DateTime64DType'>, <class 'numpy.dtypes._PyFloatDType'>) -> None

Image

Image

After reload data clicked
Image

Traceback shown above

SuMayaBee:
## Description

While investigating issue #7795, I found that when two or more HoloViews panes with linked datetime axes are displayed and the user scrolls to change the viewport on one of them, then triggers a data reload, a `UFuncTypeError` / `DTypePromotionError` is raised and the reload fails.

**Before fix:**
- Two linked datetime plots render correctly ✓
- Scrolling on one adjusts both axes as expected ✓
- Clicking "Reload Data" after scrolling raises a `UFuncTypeError` ✗
- Both plots fail to update properly ✗

https://github.com/user-attachments/assets/718501dd-7711-4414-93b3-d0a55de0a7ba

**After fix:**
- Two linked datetime plots render correctly ✓
- Scrolling on one adjusts both axes as expected ✓
- Clicking "Reload Data" after scrolling updates both plots correctly ✓ 

https://github.com/user-attachments/assets/1be62b23-d707-4ca8-9f8a-4d84c556be83

**Problem**:  
When the user scrolls on a plot with a datetime x-axis, the browser sends the new x-range back to the server as plain numbers (milliseconds since 1970) instead of real datetime values. As a result, the shared axis can end up holding a mix of types: some values as datetimes and some as numbers. Later, when Panel’s `link_axes()` code rebuilds the shared ranges and tries to compare these mixed values, it crashes because NumPy cannot compare a datetime with a plain number. The guard check:

```python
if axis.start > axis.end:
```

attempts to compare `numpy.datetime64` with `numpy.float64`, which NumPy cannot do, raising:

```text
numpy.exceptions.DTypePromotionError: The DType <class 'numpy.dtypes.DateTime64DType'> could not be promoted by <class 'numpy.dtypes.Float64DType'>
```

This is specific to datetime axes, numerical axes are unaffected because
`float64` vs `float64` comparisons always succeed.

**Solution**:  
Wrapped the inverted-axis guard in a `try/except Exception` block. When the types are compatible (normal case), the check works exactly as before. When a transient type mismatch occurs (datetime64 vs float64 after a scroll + reload), the exception is caught and the swap is safely skipped. The `max_range`-computed `start`/`end` values are then assigned to the axis immediately after, restoring both values to a consistent type.

**Technical Details**:
- **File**: `panel/pane/holoviews.py` → `link_axes()` function
- **Change**: Wrapped `if axis.start > axis.end:` in `try/except Exception`
- **Result**: Mixed-type comparisons no longer crash the reload; inverted-axis detection still works for all compatible-type cases

Fixes #7795

## How Has This Been Tested?

- **Manual Testing:** Reproduced the issue using the MRE from issue (two linked datetime HoloViews panes, scroll to change viewport, click reload). Confirmed the error no longer occurs and both plots update correctly after the fix.

### AI Disclosure

Gemini 3 Flash is used to review the codebase and identify the root cause of the issue.

### Checklist:

- [x] Tests are passing

### DIFF
--- a/panel/pane/holoviews.py
+++ b/panel/pane/holoviews.py
@@ -14,6 +14,7 @@ from typing import (
     TYPE_CHECKING, Any, ClassVar, Literal, TypedDict,
 )
 
+import numpy as np
 import param
 
 from bokeh.models import Range1d, Spacer as _BkSpacer
@@ -962,11 +963,10 @@ def link_axes(root_view, root_model):
                 (ax[-1].start, ax[-1].end) for ax in axes
                 if isinstance(ax[-1], Range1d)
             ])
-            try:
-                if axis.start > axis.end:
-                    end, start = start, end
-            except Exception:
-                pass
+            axis_start = axis.start.astype(np.int64) if isinstance(axis.start, np.datetime64) else axis.start
+            axis_end = axis.end.astype(np.int64) if isinstance(axis.end, np.datetime64) else axis.end
+            if axis_start > axis_end:
+                end, start = start, end
             axis.start = start
             axis.end = end
         for fig, p, pax, _ in axes[1:]:
--- a/panel/pane/holoviews.py
+++ b/panel/pane/holoviews.py
@@ -962,8 +962,11 @@ def link_axes(root_view, root_model):
                 (ax[-1].start, ax[-1].end) for ax in axes
                 if isinstance(ax[-1], Range1d)
             ])
-            if axis.start > axis.end:
-                end, start = start, end
+            try:
+                if axis.start > axis.end:
+                    end, start = start, end
+            except Exception:
+                pass
             axis.start = start
             axis.end = end
         for fig, p, pax, _ in axes[1:]:
--- a/panel/tests/pane/test_holoviews.py
+++ b/panel/tests/pane/test_holoviews.py
@@ -658,6 +658,29 @@ def test_holoviews_linked_axes_merged_ranges(document, comm):
 
 @pytest.mark.usefixtures("hv_bokeh")
 @hv_available
+def test_holoviews_linked_datetime_axes_after_scroll_and_update(document, comm):
+    dates = [dt.datetime(2021, 1, 1), dt.datetime(2021, 6, 1), dt.datetime(2021, 12, 31)]
+    c1 = hv.Curve((dates, [1, 2, 3]))
+    c2 = hv.Curve((dates, [3, 2, 1]))
+
+    hv1 = HoloViews(c1, backend='bokeh')
+    hv2 = HoloViews(c2, backend='bokeh')
+    layout = Row(hv1, hv2)
+
+    row_model = layout.get_root(document, comm=comm)
+    p1, p2 = row_model.select({'type': figure})
+
+    assert p1.x_range is p2.x_range
+
+    p1.x_range.start = 1609459200000.0
+    p1.x_range.end = 1614556800000.0
+
+    hv1.object = hv.Curve((dates, [4, 5, 6]))
+    hv2.object = hv.Curve((dates, [6, 5, 4]))
+
+
+@pytest.mark.usefixtures("hv_bokeh")
+@hv_available
 def test_holoviews_linked_x_axis(document, comm):
     c1 = hv.Curve([1, 2, 3])
     c2 = hv.Curve([1, 2, 3], vdims='y2')


ahuang11:
> When the user scrolls on a plot with a datetime x-axis, the browser sends the new x-range back to the server as plain numbers (milliseconds since 1970) instead of real datetime values. As a result, the shared axis can end up holding a mix of types: some values as datetimes and some as numbers. Later, when Panel’s link_axes() code rebuilds the shared ranges and tries to compare these mixed values, it crashes because NumPy cannot compare a datetime with a plain number.

Thanks for the PR. I imagine rather than a try/except, we can do better by not using a catch all Exception and narrow down to the corresponding exception type, or better yet, casting to proper datetime types, i.e. `if isinstance...`

SuMayaBee:
> > When the user scrolls on a plot with a datetime x-axis, the browser sends the new x-range back to the server as plain numbers (milliseconds since 1970) instead of real datetime values. As a result, the shared axis can end up holding a mix of types: some values as datetimes and some as numbers. Later, when Panel’s link_axes() code rebuilds the shared ranges and tries to compare these mixed values, it crashes because NumPy cannot compare a datetime with a plain number.
> 
> Thanks for the PR. I imagine rather than a try/except, we can do better by not using a catch all Exception and narrow down to the corresponding exception type, or better yet, casting to proper datetime types, i.e. `if isinstance...`

Thanks @ahuang11 for the feedback. I've updated the fix to use `isinstance` checks and cast `datetime64` values to `int64` before comparison, instead of a catch-all exception. This normalizes the types (datetime64 from fresh plots vs float64 from browser scrolls) so they can be compared safely. I also added a test case that reproduces the issue and verifies the fix.

philippjfr (approved):

PR #8475, opened 2026-03-06
compat: Bokeh 3.9.0

hoxbro:
## Description

Check compatibility with Bokeh 3.9 rc1

Also took the time to clean up the clean up to only use Python 3.14 as it now support all dependencies. 

## How Has This Been Tested?

CI

## Checklist


panel/io/fastapi.py
```diff
@@ -22,9 +22,11 @@
...
     )
     from fastapi.responses import FileResponse
-except ImportError:
-    msg = "bokeh_fastapi must be installed to use the panel.io.fastapi module."
-    raise ImportError(msg) from None
+except ImportError as e:
+    if e.name == "bokeh_fastapi":
+        msg = "bokeh_fastapi must be installed to use the panel.io.fastapi module."
+        raise ImportError(msg) from None
+    raise e
```
hoxbro, line 29: Improve error message:

Image

hoxbro: Have made a patch to conda-forge, which is why this no longer happen. We still need a official release though.

hoxbro: A new release has been made.

philippjfr (approved):

PR #8479, opened 2026-03-06
Docs: Recommend using bound function reference instead of watch=True in bind_component guide

### LINKED ISSUE #6271: Update docs to use the bound function as a reference
Often, coming from the `widget.param.watch()` historical past, I would use `watch=True` in pn.bind:
```python
import panel as pn
pn.extension()

def update_value(select_value, slider_value):
    value = select_value * slider_value
    text.value = value

slider = pn.widgets.IntSlider(value=5, start=1, end=10)
select = pn.widgets.Select(value="⭐", options=["⭐", "🐘"])
text = pn.widgets.StaticText()
pn.bind(update_value, select, slider, watch=True)

pn.Row(slider, select, text).show()
```

Others mentioned that was an anti-pattern, and shared that I can do this instead:
```python
import panel as pn
pn.extension()

def update_value(select_value, slider_value):
    value = select_value * slider_value
    return value

slider = pn.widgets.IntSlider(value=5, start=1, end=10)
select = pn.widgets.Select(value="⭐", options=["⭐", "🐘"])
text = pn.widgets.StaticText()
text.value = pn.bind(update_value, select, slider)

pn.Row(slider, select, text).show()
```

To ideally:
```python
import panel as pn
pn.extension()

def update_value(select_value, slider_value):
    value = select_value * slider_value
    return value

slider = pn.widgets.IntSlider(value=5, start=1, end=10)
select = pn.widgets.Select(value="⭐", options=["⭐", "🐘"])
text = pn.widgets.StaticText(value=pn.bind(update_value, select, slider))

pn.Row(slider, select, text).show()
```

I think this should be documented in the docs
https://panel.holoviz.org/how_to/interactivity/bind_component.html

The benefit of doing this is not recreating the object.

Macdrita:
## Description

Fixes #6271

This PR updates the documentation in `doc/how_to/interactivity/bind_component.md` to clarify the recommended pattern for using `pn.bind` when updating component parameters.

Users coming from `param.watch()` patterns may attempt to use `watch=True` with `pn.bind`. However, this pattern recreates the object whenever the inputs change and is generally considered an anti-pattern.

This PR adds a section explaining that a better approach is to assign the bound function directly to the relevant component parameter. For example:

```python
def update_value(select_value, slider_value):
    return select_value * slider_value

text = pn.widgets.StaticText(value=pn.bind(update_value, select, slider))
```

This approach keeps the component reactive while avoiding unnecessary object recreation and ensures that only the relevant parameter is updated.

The documentation change clarifies the intended usage of `pn.bind` and improves guidance for users working with reactive components.

---

## How Has This Been Tested?

This change only modifies documentation.

The updated section was reviewed locally to ensure:

* Correct Markdown formatting
* Proper placement within the existing guide
* Consistency with surrounding documentation examples

Since this is a documentation-only change, no runtime functionality or tests were affected.

---

## AI Disclosure

* [x] This PR contains AI-generated content.
* [x] I have tested all AI-generated content in my PR.
* [x] I take responsibility for all AI-generated content in my PR.

Tools: ChatGPT

---

## Checklist

* [ ] Tests added and is passing
* [x] Added documentation

### DIFF
--- a/doc/how_to/interactivity/bind_component.md
+++ b/doc/how_to/interactivity/bind_component.md
@@ -76,6 +76,7 @@ pn.Row(slider, size, select, pn.pane.Markdown(refs=irefs))
 ```
 
 In this way we can update both the current `object` and the `styles` **Parameter** of the `Markdown` pane simultaneously.
+
 ### Using the bound function as a reference
 
 In some cases it may be tempting to use `watch=True` when calling `pn.bind` to update a component:
--- a/doc/how_to/interactivity/bind_component.md
+++ b/doc/how_to/interactivity/bind_component.md
@@ -76,7 +76,26 @@ pn.Row(slider, size, select, pn.pane.Markdown(refs=irefs))
 ```
 
 In this way we can update both the current `object` and the `styles` **Parameter** of the `Markdown` pane simultaneously.
+### Using the bound function as a reference
 
+In some cases it may be tempting to use `watch=True` when calling `pn.bind` to update a component:
+
+```python
+pn.bind(update_value, select, slider, watch=True)
+```
+
+However, this pattern recreates the object whenever the inputs change and is generally considered an anti-pattern.
+
+A better approach is to assign the bound function directly to the component parameter that should update:
+
+```python
+def update_value(select_value, slider_value):
+    return select_value * slider_value
+
+text = pn.widgets.StaticText(value=pn.bind(update_value, select, slider))
+```
+
+This keeps the component reactive while avoiding unnecessary object recreation and ensures that only the relevant parameter is updated.
 ## Related Resources
 
 - Learn [how to use generators with `bind`](bind_generators)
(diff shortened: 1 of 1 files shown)


doc/how_to/interactivity/bind_component.md
```diff
@@ -76,7 +76,26 @@ pn.Row(slider, size, select, pn.pane.Markdown(refs=irefs))
 ```
 
 In this way we can update both the current `object` and the `styles` **Parameter** of the `Markdown` pane simultaneously.
+### Using the bound function as a reference
```
ahuang11, line 80:
```suggestion

### Using the bound function as a reference
```

doc/how_to/interactivity/bind_component.md
```diff
@@ -76,7 +76,26 @@ pn.Row(slider, size, select, pn.pane.Markdown(refs=irefs))
...
+A better approach is to assign the bound function directly to the component parameter that should update:
+
+```python
+def update_value(select_value, slider_value):
+    return select_value * slider_value
+
+text = pn.widgets.StaticText(value=pn.bind(update_value, select, slider))
+```
+
+This keeps the component reactive while avoiding unnecessary object recreation and ensures that only the relevant parameter is updated.
```
ahuang11, line 99:
```suggestion
This keeps the component reactive while avoiding unnecessary object recreation and ensures that only the relevant parameter is updated.

```

ahuang11 (commented):
Spacing

philippjfr (approved):

PR #8488, opened 2026-03-07
Improve documentation of VideoStream with alternative stream examples

### LINKED ISSUE #3823: Improve documentation of VideoStream
## Request

Improve the documentation of VideoStream by describing how to use alternative streams to the default web cam. It would be even better if the example (or a Gallery example) showed how to use live image recognition.

## Motivation

When reading the docs https://panel.holoviz.org/reference/widgets/VideoStream.html#widgets-gallery-videostream

"The VideoStream widget displays a video from a local stream (for example from a webcam) and allows accessing the streamed video data from Python."

It is indicated that the web cam is only an example. But it is not clear how to use anything else. For example a user is asking here how to use an alternative sensor https://discourse.holoviz.org/t/videostream-source/4225/3.

In the Streamlit community the VideoStream is very popular and I see many examples primary driven by the extension https://github.com/whitphx/streamlit-webrtc. I believe the Panel community could also do this and maybe even better because of the architecture of Panel vs Streamlit. But its difficult to get inspired and started because there is no example of this.

Macdrita:
## Description

Fixes #3823

This PR improves the documentation of the `VideoStream` widget by adding examples demonstrating how to process frames from the stream and explaining how alternative video sources can be used.

## How Has This Been Tested?

The examples were executed locally to verify that frames can be accessed and processed.

## AI Disclosure

- [ ] This PR contains AI-generated content.

## Checklist

- [ ] Tests added and is passing
- [x] Added documentation

### DIFF
--- a/examples/reference/widgets/VideoStream.ipynb
+++ b/examples/reference/widgets/VideoStream.ipynb
@@ -130,6 +130,63 @@
    "source": [
     "pn.Row(video_stream.controls(jslink=True), video_stream)"
    ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "## Using alternative video streams\n",
+    "\n",
+    "While `VideoStream` is commonly used with a webcam, it can also work with other video sources such as IP cameras, external sensors, or custom video pipelines.\n",
+    "\n",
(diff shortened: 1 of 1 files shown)


philippjfr:
The added example does not work. @Macdrita at this point I will formally notify you that we reserve the right to block future contributions if there is no AI disclosure and we find that your submissions were not actually checked by you personally.

PR #8496, opened 2026-03-10
ci: Fix jupyterlite workflow - round 2

hoxbro:
## Description

Continue working on the #8494, now it no longer publish to main, but the `github.event_name` was not set correctly. Causing no upload to happen.

image

## How Has This Been Tested?

None, will check again on next dev release.


hoxbro:
Looks like it works now 

image

PR #8497, opened 2026-03-10
Fix plotly escaping prototype

### LINKED ISSUE #8449: [BUG] Incorrect escaping in SyntaxError display and potential Prototype Pollution in Plotly model
Two logic issues in the frontend models were identified during a recent audit, affecting both display correctness and potential security:

1. **Incomplete Escaping in `formatError` (`panel/models/util.ts`):**
The current implementation uses `code.replace(">", "&lt;").replace("<", "&gt;")`. Since these are string-based replaces, they only handle the first occurrence of each character. This behavior, which I noticed while reviewing the error overlay logic, could lead to broken rendering or potential XSS if the source code contains multiple tags. Switching to a global Regex `replace(/>/g, "&lt;").replace(/</g, "&gt;")` would resolve this.

2. **Potential Prototype Pollution in Plotly Trace Processing (`panel/models/plotly.ts`):**
The `_get_trace` function splits column names by `.` and dynamically accesses properties on the `trace` object without validation. If column names are derived from untrusted user data (e.g., uploaded CSV headers), this could allow an attacker to pollute the object prototype. Implementing a guard to skip sensitive keys like `__proto__`, `constructor`, and `prototype` is recommended.

A Pull Request to fix the escaping issue in `util.ts` is ready if required, and further discussion on the best approach for hardening the Plotly model would be appreciated.

Macdrita:
## Description

This PR fixes two issues reported in #8449.

1. **Incorrect escaping in `formatError` (panel/models/util.ts)**
   Updated the escaping logic to use global regular expressions so that all `<` and `>` characters are properly escaped. This ensures error messages render correctly when displaying code snippets containing multiple HTML-like characters.

2. **Prototype pollution risk in Plotly trace processing (panel/models/plotly.ts)**
   Added a guard in `_get_trace` to ignore dangerous property names (`__proto__`, `constructor`, `prototype`) when mapping column names to Plotly trace properties. This prevents prototype pollution if a data source contains malicious or unexpected column names.

Fixes #8449

## How Has This Been Tested?

* Verified that error formatting correctly escapes multiple `<` and `>` characters.
* Confirmed Plotly traces render normally with standard column names.
* Ensured that dangerous property names are skipped during trace generation.

## AI Disclosure

* [ ] This PR contains AI-generated content.

## Checklist

* [ ] Tests added and is passing
* [ ] Added documentation


ahuang11:
@Macdrita Could you please rebase new PRs on main before pushing?

Additionally, please make sure the AI Disclosure checkbox is filled out correctly if the PR contains AI-generated content and also the Checklist section. Thanks!

philippjfr:
@Macdrita Thanks for the PR, however in it's current form this is not mergable because it has other changes and we don't have an AI disclosure from you. The pre-commit hooks aren't passing either.

I'm closing the issue via: https://github.com/holoviz/panel/pull/8501 because we would like to prioritize these fixes for an imminent 1.8.10 release.

PR #8498, opened 2026-03-11
Improve state.busy logic to avoid state mismatch

### LINKED ISSUE #582: Header spinner keeps spinning after browser back navigation
## Description

When navigating from a `pmui.Page` to another page via a link, then pressing the browser's **Back** button, the header progress spinner keeps spinning indefinitely. The page content loads correctly, but the spinner never stops.

Image

## Steps to Reproduce

1. Open `page` in the browser
2. Click the link to navigate to the *other page*
3. Press the browser **Back** button to return to `page`
4. Observe the header spinner — it keeps spinning and never stops

## Expected Behavior

The spinner should stop after the page finishes loading.

## Actual Behavior

The spinner continues spinning indefinitely after navigating back.

## Environment

- Panel: latest
- Panel Material UI: latest
- Browser: Chrome / Edge (PWA and regular tab)

## Minimal Reproducible Example [NOT WORKING]

I've not yet been able to create MRE. But I've seen it across many applications. The below is indicative.

Save as `page.py` then run:

```bash
panel serve page.py --index page --dev
```

**page.py:**

```python
import panel as pn
import panel_material_ui as pmui

    pmui.Page(
        title="Page A",
        main=[
            pn.pane.Markdown("# Page A"),
            pn.pane.Markdown("[EXTERNAL URL](https://panel-material-ui.holoviz.org/)"),
            pn.pane.Markdown("Click the link above, then press browser Back button."),
            pn.pane.Markdown("The header spinner will keep spinning after navigating back."),
        ],
    ).servable()
```

Steps:

1. Open `http://localhost:5006/page_a`
2. Click "EXTERNAL URL"
3. Press browser Back button
4. The header spinner on `page` keeps spinning indefinitely

Image

philippjfr:
The existing logic to track whether the application was busy was error prone because it relied on a simple counter. We now track actual event ids and clean them up should some event never get processed.

Fixes https://github.com/panel-extensions/panel-material-ui/issues/582

### DIFF
--- a/panel/io/callbacks.py
+++ b/panel/io/callbacks.py
@@ -6,12 +6,13 @@ import asyncio
 import inspect
 import logging
 import time
+import uuid
 
 from functools import partial
 
 import param
 
-from ..util import edit_readonly, function_name
+from ..util import function_name
 from .logging import LOG_PERIODIC_END, LOG_PERIODIC_START
 from .state import curdoc_locked, set_curdoc, state
 
@@ -78,7 +79,7 @@ class PeriodicCallback(param.Parameterized):
             self.stop()
             self.start()
 
-    def _exec_callback(self, post=False):
+    def _exec_callback(self, post=False, busy_event_id=None):
         try:
             with set_curdoc(self._doc):
                 if self.running:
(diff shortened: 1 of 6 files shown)


MarcSkovMadsen:
If this fixes the panel-material-ui issue it means the world to me. Then I can actually start using panel-material-ui in production when next version of panel and panel-material-ui is released. Thanks.

PR #8501, opened 2026-03-12
Address potentially security flaws in panel.js

### LINKED ISSUE #8449: [BUG] Incorrect escaping in SyntaxError display and potential Prototype Pollution in Plotly model

philippjfr:
## Description

Adds additional validation and escaping to avoid potential XSS and prototype pollution vulnerabilities.

Fixes https://github.com/holoviz/panel/issues/8449

## AI Disclosure

    Tools: ChatGPT plus manual review

## Checklist

- [ ] Tests added and is passing
- [ ] Added documentation

### DIFF
--- a/panel/models/plotly.ts
+++ b/panel/models/plotly.ts
@@ -453,8 +453,8 @@ export class PlotlyPlotView extends HTMLBoxView {
       const prop = propPath[propPath.length - 1]
       const propParent = getSafeParent(trace, propPath.slice(0, -1))
       if (!propParent || !prop || !isSafePath(propPath) || FORBIDDEN_KEYS.has(prop)) {
-	console.warn("Attempted prototype pollution detected via Plotly column resolution.")
-	continue
+        console.warn("Attempted prototype pollution detected via Plotly column resolution.")
+        continue
       }
 
       if (update && propPath.length == 1) {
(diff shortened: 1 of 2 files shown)


RinZ27:
@philippjfr verified the changes locally and they look solid. Implementing the global regex and the explicit key checks for Plotly path resolution effectively mitigates those logic flaws. Great work on the quick turnaround.

philippjfr:
@RinZ27 appreciate the report, and thanks for testing so quickly. We'll try to get an RC for 1.8.10 out today, followed shortly by a real release (hopefully tomorrow).

PR #8504, opened 2026-03-13
feat: Add PDF export support to Panel save()

### LINKED ISSUE #8340: Download Dashboard to PDF
On the community call, it was mentioned people had built workarounds to export dashboards to PDFs.

In the docs, it only mentions HTML and PNG:
https://panel.holoviz.org/how_to/export/index.html

It'd be nice to have a native way to export to PDF.

SuMayaBee:
## Description

**Problem:** This adds PDF export support to `save()` by reusing the existing screenshot-based export path that Panel already uses for PNG export.

**Solution:**
I looked through the current save flow and found that `Viewable.save()` already routes HTML and PNG through shared logic, so instead of introducing a separate export implementation, this change extends that path with PDF support.

**In practice, the change:**

- adds PDF handling to the `save()` API
- reuses the same browser screenshot/render pipeline used for PNG export
- factors the shared screenshot logic into a helper so PNG and PDF do not duplicate the same setup
- adds tests for `.pdf` dispatch and the explicit `as_pdf=True` path
- updates the export docs with a simple PDF example

The goal here was to keep the implementation small and consistent with the existing design, rather than creating a completely new export system.

## After the change:

https://github.com/user-attachments/assets/9ce28e38-07ef-4bbb-996d-76945511b570

## How Has This Been Tested?

- Added unit tests for:
  - saving with a `.pdf` filename
  - saving with `as_pdf=True`
- Manually tested in Jupyter with a simple Panel app using:
  - `app.save("demo_app.pdf")`
- Verified that the generated PDF file is created successfully when webdriver dependencies are available.

Example used for manual testing:

```python
import panel as pn

pn.extension()

app = pn.Column(
    "# PDF export test",
    pn.pane.Markdown("Hello from Panel"),
    pn.widgets.IntSlider(name="Value", start=0, end=10, value=5),
)

app.save("demo_app.pdf")
```

Fixes #8340

## AI Disclosure

        Tools: GitHub Copilot (GPT-5.4) for implementation support

## Checklist

- [x] Added documentation

### DIFF
--- a/doc/how_to/export/saving.md
+++ b/doc/how_to/export/saving.md
@@ -1,12 +1,12 @@
 # Save App to File
 
-This guide addresses how to export an app to a HTML or PNG file.
+This guide addresses how to export an app to a HTML, PNG or PDF file.
 
 ---
 
-In case you don't need an actual server or simply want to export a static snapshot of a panel app, you can use the ``save`` method, which allows exporting the app to a standalone HTML or PNG file.
+In case you don't need an actual server or simply want to export a static snapshot of a panel app, you can use the `save` method, which allows exporting the app to a standalone HTML, PNG or PDF file.
 
-By default, the HTML file generated will depend on loading JavaScript code for BokehJS from the online ``CDN`` repository, to reduce the file size. If you need to work in an air-gapped or no-network environment, you can declare that ``INLINE`` resources should be used instead of ``CDN``:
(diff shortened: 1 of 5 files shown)


philippjfr:
Apologies, this should have been made clearer in the PR but https://github.com/holoviz/panel/pull/5913 already implements a better solution for this. It's just waiting on bokeh to convert their export machinery from Selenium to Playwright for consistency.

PR #8507, opened 2026-03-14
Fix widget disabled attribute stuck after rapid toggling

### LINKED ISSUE #7389: Widget 'disabled' attribute breaks after toggling

<details>

```plaintext
Python 3.11.9
panel 1.5.2 (still working on 1.4.5)
bokeh 3.6.0
```
</details>

**Expected behaviour**
Setting the `disabled` attribute of a widget to `False` should result in the widget being editable.

**Observed behaviour**
When the `disabled` attribute of widgets is set to `True` and then set to `False`, the attribute can be updated one more time and then fails to update further.

In the reproduction code below, clicking the Button named "toggle" once will see no change in disabled state. Clicking the "toggle" Button twice will see the state of the dummy widgets permanently set to disabled.

This behaviour occurs when the switching of states is done quickly (i.e. in a single function), but doesn't seem to occur when the switching is done in multiple cells of a notebook.

```python
import panel as pn
pn.extension()

dummy_button = pn.widgets.Button(name="dummy", button_type='primary')
dummy_text = pn.widgets.TextAreaInput()
toggle_button = pn.widgets.Button(name="toggle")

def toggle(_):
    dummy_button.disabled = True
    dummy_text.disabled = True
    dummy_button.disabled = False
    dummy_text.disabled = False

toggle_button.on_click(toggle)

layout = pn.Column(dummy_button, dummy_text, toggle_button)
layout.servable()
```

**Before first click**
before_click
**After first click**
after_first_click
**After second click**
after_second_click

ghostiee-11:
## Description

Fixes #7389

When `disabled` is toggled rapidly within a single callback (e.g., `True` then immediately `False`), the widget gets permanently stuck as disabled. This happens because of a bug in `Syncable._update_model()` in `panel/reactive.py`.

The `_events` check at line 376 was meant to avoid overwriting frontend-originated property changes, but it had two problems:

1. It skipped adding the attribute to `_changing`, so `model.update()` still applied the change without boomerang protection
2. The resulting `_server_change` callback re-populated `_events` with the stale value, permanently contaminating it
3. Every subsequent Python update for that attribute hit the same check and got silently dropped

The fix removes the `continue` statement so the code falls through to the normal path. This clears the stale `_events` entry while still adding the attribute to `_changing` for proper boomerang protection.

**Reproducer (from the issue):**
```python
import panel as pn
pn.extension()

dummy_button = pn.widgets.Button(name="dummy", button_type='primary')
dummy_text = pn.widgets.TextAreaInput()
toggle_button = pn.widgets.Button(name="toggle")

def toggle(_):
    dummy_button.disabled = True
    dummy_text.disabled = True
    dummy_button.disabled = False
    dummy_text.disabled = False

toggle_button.on_click(toggle)
pn.Column(dummy_button, dummy_text, toggle_button).servable()
```

Without the fix, clicking "toggle" once or twice permanently disables the widgets. With the fix, widgets stay enabled after any number of toggles.

**Before fix** (widgets stuck disabled after clicking "toggle"):

7389_before_toggle

**After fix** (widgets remain enabled after 5 toggle clicks):

7389_after_5_toggles

## How Has This Been Tested?

- All 39 existing `test_reactive.py` tests pass (zero regressions)
- All 13 existing button UI tests pass
- Added 3 new unit tests:
  - `test_rapid_toggle_disabled`: toggles disabled True then False, verifies widget stays enabled
  - `test_rapid_toggle_disabled_repeated`: 5 rapid toggle cycles, verifies no permanent contamination
  - `test_python_update_clears_stale_events`: injects stale `_events` entry, verifies Python update overrides it
- Added 1 new UI test:
  - `test_widget_disabled_toggle`: serves the exact reproducer, clicks toggle twice via Playwright, asserts both Button and TextAreaInput remain enabled
- Manual testing with the reproducer confirms the fix through the full server path

## AI Disclosure

    Tools: Claude, used for assistance. I planned the fix, understood the root cause, and tested everything manually.

## Checklist

### DIFF
--- a/panel/reactive.py
+++ b/panel/reactive.py
@@ -377,10 +377,18 @@ class Syncable(Renderable):
                 del msg[attr]
                 continue
             elif attr in self._events:
-                # Clear stale frontend event but still apply the
-                # Python-originated update so _changing protection
-                # prevents boomerang re-contamination of _events
-                del self._events[attr]
+                # Compare against the last frontend-originated value
+                # to distinguish a boomerang echo from a newer Python
+                # update. Matching values are dropped; differing
+                # values fall through so the Python update wins.
+                frontend_value = self._events.pop(attr)
+                try:
+                    is_boomerang = bool(value == frontend_value)
+                except Exception:
+                    is_boomerang = value is frontend_value
+                if is_boomerang:
+                    del msg[attr]
+                    continue
 
             # Bokeh raises UnsetValueError if the value is Undefined.
             try:
--- a/panel/reactive.py
+++ b/panel/reactive.py
@@ -377,10 +377,10 @@ class Syncable(Renderable):
                 del msg[attr]
                 continue
             elif attr in self._events:
-                # Do not override a property value that was just changed
-                # on the frontend
+                # Clear stale frontend event but still apply the
+                # Python-originated update so _changing protection
+                # prevents boomerang re-contamination of _events
                 del self._events[attr]
-                continue
 
             # Bokeh raises UnsetValueError if the value is Undefined.
             try:
--- a/panel/tests/test_reactive.py
+++ b/panel/tests/test_reactive.py
@@ -20,7 +20,7 @@ from panel.pane import Markdown
 from panel.reactive import Reactive, ReactiveHTML
 from panel.viewable import Viewable
 from panel.widgets import (
-    Checkbox, IntInput, IntSlider, StaticText, TextInput,
+    Button, Checkbox, IntInput, IntSlider, StaticText, TextInput,
 )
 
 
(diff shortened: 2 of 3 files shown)


ghostiee-11:
The CI failures are all pre-existing flaky React UI tests in `test_custom.py`. They're Playwright timeouts on React ESM component tests (`test_send_event`, `test_after_render_lifecycle_hooks`, `test_nested_update`, etc.) across all three platforms. None of these are related to the widget disabled attribute changes in this PR. Re-running the failed jobs should clear them up.

panel/reactive.py
```diff
@@ -374,10 +374,10 @@ def _update_model(
                 del msg[attr]
                 continue
             elif attr in self._events:
-                # Do not override a property value that was just changed
-                # on the frontend
+                # Clear stale frontend event but still apply the
+                # Python-originated update so _changing protection
+                # prevents boomerang re-contamination of _events
                 del self._events[attr]
```
philippjfr: This makes the `_events` tracking kind of pointless. I'll have to look into this in more depth, seems like the `_events` needs to track the value as well so it doesn't discard all events.

ghostiee-11:
Thanks for the review! You're right, the previous approach cleared `_events[attr]` unconditionally which threw away legitimate frontend events. I've pushed a rework that leverages the fact that `_events` already stores `{attr: value}`.

The new logic in `_update_model`:
- If `msg[attr] == self._events[attr]`: boomerang echo, skip the update (preserves the original intent of the check)
- If they differ: clear the stale entry and fall through so the newer Python update wins

This keeps boomerang protection intact for genuine echoes while fixing the rapid-toggle case where the stale `_events` entry held an older value.

A couple of notes on the implementation:
1. The `==` is wrapped in `try/except Exception` because numpy arrays raise `ValueError` on truthiness conversion and some custom `__eq__` implementations can raise arbitrarily. The fallback uses `is` identity, which means numpy-array-valued params are always treated as non-boomerang (worst case: the value is re-sent to a frontend that already has it, which is the pre-existing behavior for such types). Happy to narrow to `(ValueError, TypeError)` if you prefer.
2. Added `test_boomerang_value_is_skipped` to cover the matching-value path; the existing three tests cover the differing-value path.

Let me know if you'd like a different shape here, e.g., tracking a sequence number instead of the value.

philippjfr (approved):

philippjfr:
Thank you @ghostiee-11!

ghostiee-11:
Thanks for the review and merge Philip :)

PR #8508, opened 2026-03-14
Fix Modal objects update rendering inline instead of in dialog

### LINKED ISSUE #7669: BUG - Updating Modal via modal.objects results in the display of the updated objects in place of modal

<details>

```plaintext
panel==1.6.0
```
</details>

Running the code below will show the modal window as expected. However, if the lines inside `__update_modal` are uncommented, then clicking the button inserts the modal contents into the main page and does not show the modal correctly. This bug prevents dynamic update of the modal contents.

```python
import param
from panel import Column, Modal, extension, template, widgets

extension("modal")


class TestApp(param.Parameterized):
    text = param.String(default="Hello, World!")
    slider = param.Number(default=0.0)

    def __init__(self, **params):
        super().__init__(**params)
        self.modal = Modal("TEST", widgets.TextInput(name="Text:", value=self.text), name="Basic FloatPanel", margin=20)
        self.toggle_button = widgets.Button(name="Toggle modal", button_type="primary")
        self.toggle_button.on_click(lambda event: self.__update_modal())

    def __update_modal(self):
        # self.modal.objects = [
        #     widgets.FloatSlider(name="Slider", value=self.slider),
        # ]
        self.modal.show()

    def __panel__(self):
        return Column(
            self.modal,
            self.toggle_button,
        )


test = TestApp()
app = template.VanillaTemplate(title="Basic Modal", main=[test])
app.show()
```

ghostiee-11:
## Description

Fixes #7669

When dynamically updating `modal.objects` (e.g., replacing children of a Modal), the new objects render inline on the main page instead of inside the modal popup. A page refresh would fix it, confirming it was a client-side DOM routing problem.

The root cause is in `panel/models/modal.ts`. The `ModalView.update_children()` method delegated to `LayoutDOMView.prototype.update_children`, which rebuilds child views and appends them to `shadow_el` (the host element). It has no knowledge of the custom `dialog > dialog_content` DOM structure that `create_modal()` builds, so updated children end up outside the dialog.

The fix:
- Store a reference to the `dialog_content` div as `this.content` during `create_modal()`
- Rewrite `update_children()` to route child views into `this.content` instead of `shadow_el`
- The new implementation mirrors `ColumnView.update_children()` (including the `matching_index` optimization), but targets the modal's content container

**Reproducer (from the issue):**
```python
import param
from panel import Column, Modal, extension, widgets

extension("modal")

class TestApp(param.Parameterized):
    text = param.String(default="Hello, World!")
    slider = param.Number(default=0.0)

    def __init__(self, **params):
        super().__init__(**params)
        self.modal = Modal("TEST", widgets.TextInput(name="Text:", value=self.text), margin=20)
        self.toggle_button = widgets.Button(name="Toggle modal", button_type="primary")
        self.toggle_button.on_click(lambda event: self.__update_modal())

    def __update_modal(self):
        self.modal.objects = [
            widgets.FloatSlider(name="Slider", value=self.slider),
        ]
        self.modal.show()

    def __panel__(self):
        return Column(self.modal, self.toggle_button)

test = TestApp()
test.__panel__().servable()
```

Without the fix, clicking "Toggle modal" renders the FloatSlider inline on the page. With the fix, it renders correctly inside the modal popup:

7669_after_fix

## How Has This Been Tested?

- All 5 existing `test_modal.py` UI tests pass (zero regressions)
- Added 4 new UI tests:
  - `test_modal_update_objects`: updates objects while modal is open, verifies content renders inside `#pnx_dialog_content`, verifies close/reopen still works
  - `test_modal_update_objects_while_closed`: updates objects while modal is closed, opens it, verifies content is inside the dialog
  - `test_modal_append_objects`: appends a child via `modal.append()`, verifies it renders inside the dialog
  - `test_modal_multiple_updates`: 3 rapid successive object replacements, verifies only final state is visible inside the dialog
- Manual testing with the reproducer confirms the fix through the full server path

## AI Disclosure

    Tools: Claude, used for assistance. I planned the fix, understood the root cause, and tested everything manually.

## Checklist

### DIFF
--- a/panel/models/modal.ts
+++ b/panel/models/modal.ts
@@ -6,7 +6,6 @@ import type {StyleSheetLike} from "@bokehjs/core/dom"
 import type {Attrs} from "@bokehjs/core/types"
 import {UIElementView} from "@bokehjs/models/ui/ui_element"
 import {isNumber} from "@bokehjs/core/util/types"
-import {LayoutDOMView} from "@bokehjs/models/layouts/layout_dom"
 
 import modal_css from "styles/models/modal.css"
 
@@ -42,6 +41,7 @@ export class ModalView extends BkColumnView {
 
   modal: A11yDialogView
   close_button: HTMLButtonElement
+  content: HTMLElement
 
   override connect_signals(): void {
     super.connect_signals()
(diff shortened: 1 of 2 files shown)


philippjfr:
Thanks, fix looked good and I was about to suggest the same with `self_target`.

ghostiee-11:
Thanks for the feedback! Glad the `self_target` approach lines up with what you had in mind. Much cleaner than duplicating the `update_children` internals.

philippjfr (approved):

PR #8509, opened 2026-03-15
fix: Trend indicator not working with date_range x-axis

### LINKED ISSUE #6158: Trend indicator not working with date_range x-axis.
I'm on the `main` branch of Panel trying to help rh1 in https://discourse.holoviz.org/t/how-to-use-pn-indicator-trend/6670.

I can see that a `date_range` does not work as the `plot_x` value for Trend indicator.

```python
import pandas as pd
import panel as pn
import numpy as np

data = pd.DataFrame(
    {
        "Date": pd.date_range(start="2020-01-01", periods=12, freq="M"),
        "Sales": [100, 120, 150, 130, 110, 140, 160, 180, 200, 220, 250, 230],
    }
)
data["x"] = list(range(len(data["Date"])))

pn.indicators.Trend(
    data=data,
    plot_x="Date",
    plot_y="Sales",
    styles={"color": "#1f77b4"},
    height=200,
    width=400,
).servable()
```

image

## Workaround

Use a simple column of integers

```python
import pandas as pd
import panel as pn
import numpy as np

data = pd.DataFrame(
    {
        "Date": pd.date_range(start="2020-01-01", periods=12, freq="M"),
        "Sales": [100, 120, 150, 130, 110, 140, 160, 180, 200, 220, 250, 230],
    }
)
data["x"] = list(range(len(data["Date"])))

pn.indicators.Trend(
    data=data,
    plot_x="x",
    plot_y="Sales",
    styles={"color": "#1f77b4"},
    height=200,
    width=400,
).servable()
```

image

SuMayaBee:
## Description

The `Trend` indicator was not rendering correctly when a datetime column was used as `plot_x`. Instead of a proper chart, the bars appeared as tiny invisible specks crammed into the bottom-right corner of the plot.

## Before the fix:

https://github.com/user-attachments/assets/3a1ceea7-3128-4f2b-99a3-05b48e6f53c1

## After the fix:
https://github.com/user-attachments/assets/330687ff-8d5c-4c83-888e-980c3257c417

**Root cause:**
`ColumnDataSource.from_df()` converts datetime columns to milliseconds since epoch (e.g. `2020-01-31` → `1580428800000.0`). The `VBar` glyph in the frontend has a hardcoded `width: 0.9`, which works fine for integer x-values but becomes completely invisible when x-values are ms timestamps, the bars end up 0.9 ms wide while data points are ~30 days apart.

**Fix:**
In `_get_data()`, after calling `ColumnDataSource.from_df()`, we check if the `plot_x` column has a datetime dtype and replace the ms-timestamps with simple integer indices `[0, 1, 2, ...]`. This is safe because the Trend indicator never displays x-axis labels or ticks, only the relative positions matter for the visual shape.

Fixes #6158

## How Has This Been Tested?

**To reproduce the bug** (before the fix), run:

```python
import pandas as pd
import panel as pn

data = pd.DataFrame({
    "Date": pd.date_range(start="2020-01-01", periods=12, freq="ME"),
    "Sales": [100, 120, 150, 130, 110, 140, 160, 180, 200, 220, 250, 230],
})

pn.indicators.Trend(
    data=data,
    plot_x="Date",
    plot_y="Sales",
    height=200,
    width=400,
).servable().show()
```

After the fix the chart renders correctly. 
A new unit test `test_trend_date_range_x_axis` was added to `panel/tests/widgets/test_trend_indicator.py` to cover this case.

## AI Disclosure

    Tools: Claude Sonnet 4.6 (Claude Code)

## Checklist

- [ ] Added documentation


panel/widgets/indicators.py
```diff
@@ -1174,7 +1174,14 @@ def _get_data(self):
...
         elif isinstance(self.data, dict):
             return self.data, self.data
-        return self.data, ColumnDataSource.from_df(self.data)
+        processed = ColumnDataSource.from_df(self.data)
+        if (
+            self.plot_x in getattr(self.data, 'columns', []) and
+            np.issubdtype(self.data[self.plot_x].dtype, np.datetime64)
+        ):
+            processed = dict(processed)
+            processed[self.plot_x] = np.arange(len(self.data), dtype=float)
```
philippjfr: Appreciate the fix but this seems problematic since it doesn't handle unevenly spaced data very well. We should just handle datetimes in the Trend model.

SuMayaBee: @philippjfr Thanks for the feedback! I've updated the fix to address the uneven spacing concern. Instead of using integer indices with np.arange, datetime values are now normalized by dividing by the minimum interval between consecutive points:

      x_ns = self.data[self.plot_x].astype(np.int64)
      diffs = np.diff(x_ns)
      min_diff = diffs[diffs > 0].min() if len(diffs) else 1
      processed[self.plot_x] = (x_ns - x_ns[0]) / min_diff

So unevenly spaced data like [Jan 1, Jan 2, Feb 1] becomes [0, 1, 31] rather than [0, 1, 2], preserving the relative gaps. The VBar width of 0.9 then works correctly for all cases. Tested manually and confirmed bars render correctly.

philippjfr:
Thank you @SuMayaBee!

PR #8512, opened 2026-03-16
Allow making CDN_ROOT configurable

philippjfr:
In certain air-gapped environments the default HoloViz CDN is not available. Therefore we allow users to configure the `CDN_ROOT` either via the `PANEL_CDN_ROOT` environment variable, e.g. you can swap it for an NPM CDN with 

`PANEL_CDN_ROOT="https://unpkg.com/@holoviz/panel@"`


JL2435:
In lieu of this change, an air-gapped environment would also need to be able to set the convert.py PYODIDE_URL. Therefore, it's probably worth exposing that as well.

PR #8515, opened 2026-03-16
Fix Gauge/ECharts crash when echarts library not yet loaded

### LINKED ISSUE #8514: Gauge indicator crashes entire page and ignores slider updates
## Description

I was testing the `Gauge` indicator alongside other widgets and ran into two issues:

### 1. Page crashes when ECharts is not loaded

When `pn.extension('echarts')` is not called before creating a Gauge, or when the ECharts CDN script hasn't finished loading by the time `render()` runs, the entire page crashes with a blank screen. The error propagates up and kills the entire Bokeh render pipeline, so every widget on the page (sliders, text, etc.) also fails to render.

### 2. Gauge value desyncs from slider

Even when the Gauge does render successfully, linking it to a slider via `param.watch` shows inconsistent behavior. The slider value and the Gauge value don't stay in sync properly. For example, moving the slider to 25 might show a different value on the Gauge, or updates are delayed/dropped during rapid interaction.

## Reproduction

```python
import panel as pn

pn.extension('echarts')

gauge = pn.indicators.Gauge(name="Speed", value=50, bounds=(0, 100))
slider = pn.widgets.IntSlider(name="Test Slider", start=0, end=100, value=25)

def update_gauge(event):
    gauge.value = event.new

slider.param.watch(update_gauge, 'value')

pn.Column(gauge, slider).servable()
```

Serve with `panel serve app.py` and drag the slider. The Gauge needle position does not always match the slider value.

Without `pn.extension('echarts')`, the page is completely blank.

## Expected behavior

- The page should render all widgets even if ECharts fails to load
- The Gauge value should update in sync with the slider when linked via `param.watch`

## Actual behavior

- Without `pn.extension('echarts')`: completely blank page with console error:
```
Error rendering Bokeh items: TypeError: Cannot read properties of undefined (reading 'init')
    at g.render (panel.min.js:207)
```
- With `pn.extension('echarts')`: Gauge renders but shows values that don't match the slider during interaction

## Screenshots

https://github.com/user-attachments/assets/fe804e84-0b26-4709-83b5-ffd444075672

Page crashes : 
Image

## Root cause

`EChartsView.render()` in `panel/models/echarts.ts` calls `(window as any).echarts.init(...)` without checking if `window.echarts` exists. Several other methods (`_resize`, `remove`, `connect_signals`) also access `this._chart` or `window.echarts` without null guards.

## Environment

- Panel: 1.8.9
- Bokeh: 3.8.2
- Browser: Chrome/Safari
- OS: macOS

ghostiee-11:
## Description

I was testing the `Gauge` indicator and found that it crashes the entire page with `TypeError: Cannot read properties of undefined (reading 'init')`. When a Gauge is placed alongside other widgets, all widgets fail to render and the page goes completely blank.

The root cause is that `EChartsView.render()` calls `window.echarts.init()` without checking if the ECharts library has loaded from CDN yet. The error propagates up through Bokeh's render pipeline and kills everything on the page.

Fixes #8514

### Changes

**`panel/models/echarts.ts`**
- Added null guards on all `window.echarts` and `this._chart` accesses (6 vulnerable code paths)
- Extracted `_init_chart()` for deferred initialization when library loads late
- Added `_await_echarts()` with script `onload` listener and polling fallback (50ms interval, 10s timeout)
- Added loading indicator ("Loading ECharts...") while waiting for library
- Fixed pre-existing bug in `_subscribe()` where the query null check was inverted (passing `null` as query selector instead of the actual query string)
- Replaced empty `catch` blocks with `console.debug` logging

**`panel/widgets/indicators.py`**
- Added Python-side `param.warning` in `Gauge.__init__` when echarts extension is not loaded
- Updated docstring to document `pn.extension('echarts')` requirement

### After fix

after_fix_with_extension

https://github.com/user-attachments/assets/1e2a2b4a-c42f-4951-aa73-77c364139e18

## How Has This Been Tested?
**Unit tests** (7 passing):
```
pytest panel/tests/widgets/test_indicators.py -k gauge
```
- `test_gauge_creation` - basic instantiation
- `test_gauge_colors` - color threshold params
- `test_gauge_custom_opts` - custom ECharts options
- `test_gauge_process_param_change` - full output validation
- `test_gauge_process_param_change_with_colors` - color in processed output
- `test_gauge_warns_without_extension` - warning fires when echarts not loaded

**UI tests** (3 new Playwright tests):
- `test_gauge_renders` - canvas element appears
- `test_gauge_does_not_crash_other_widgets` - regression test: Gauge + IntSlider on same page
- `test_gauge_value_update` - value change does not crash

**Manual verification** with screen recording showing before/after behavior.

## Checklist

- [ ] Added documentation

## AI Disclosure

Used AI for planning the fix approach and code scaffolding. All testing, debugging, and verification were done manually.

### DIFF
--- a/panel/models/echarts.ts
+++ b/panel/models/echarts.ts
@@ -106,7 +106,7 @@ export class EChartsView extends HTMLBoxView {
 
   _await_echarts(): void {
     // Try script onload listener first (event-driven, not polling)
-    const script = document.querySelector("script[src*='echarts']") as HTMLScriptElement | null
+    const script = document.querySelector("script[src*='echarts']")
     if (script != null) {
       const onLoad = () => {
         script.removeEventListener("load", onLoad)
--- a/panel/models/echarts.ts
+++ b/panel/models/echarts.ts
@@ -39,9 +39,12 @@ export class EChartsEvent extends ModelEvent {
 export class EChartsView extends HTMLBoxView {
   declare model: ECharts
 
-  container: Element
+  container: HTMLDivElement
   _chart: any
   _callbacks: Array<any>[] = []
+  _loading_interval: ReturnType<typeof setInterval> | null = null
+  _loading_timeout: ReturnType<typeof setTimeout> | null = null
+  _loading_el: HTMLDivElement | null = null
 
   override connect_signals(): void {
     super.connect_signals()
@@ -50,7 +53,9 @@ export class EChartsView extends HTMLBoxView {
     this.on_change([width, height], () => this._resize())
     this.on_change([theme, renderer], () => {
       this.render()
-      this._chart.resize()
+      if (this._chart != null) {
+        this._chart.resize()
+      }
     })
     this.on_change([event_config, js_events], () => this._subscribe())
   }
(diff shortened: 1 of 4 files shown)


ghostiee-11:
The Windows UI failures are unrelated to this PR - they're pre-existing flaky tests (`test_card_widget_not_collapsed`, `test_reload_app_on_touch`, and several `test_pyodide_test_convert_*` tests timing out on PyScript loading). All unit tests, other UI tests, and pre-commit pass green.

philippjfr (approved):
Thanks @ghostiee-11! Fine with this change but we probably want a generalized solution for this eventually.

PR #8518, opened 2026-03-16
Check the output of `_scheduled_refresh` to prevent `TypeError`

### LINKED ISSUE #8517: `TypeError` sometimes raised after scheduled OAuth refresh
<details>

```plaintext
panel 1.8.8
```
</details>

I've had several occurrences of 500 Internal Error on trying to access my app with OAuth and automatic refresh enabled. The logs show the following error:
```
  File "/usr/local/lib/python3.12/site-packages/tornado/web.py", line 1859, in _execute
    result = await result  # type: ignore
             ^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/site-packages/bokeh/server/views/auth_request_handler.py", line 79, in prepare
    self.current_user = await self.application.auth_provider.get_user_async(self)
                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/site-packages/panel/auth.py", line 1147, in get_user
    expires_in = expiry - now_ts
```

I cannot provide an MRE. However looking at the code there's a way for this to happen. If in the scheduled refresh, the code returns [here](https://github.com/holoviz/panel/blob/e6c2916539ea19b450d76271a3768cc071324309/panel/auth.py#L1208) for some reason, then `expiry` is `None` and there's no check like [here](https://github.com/holoviz/panel/blob/e6c2916539ea19b450d76271a3768cc071324309/panel/auth.py#L1098) before computing `expires_in`.

While potentially not addressing the root cause, an easy fix would be to force full re-authentication by returning `None` at that point if the output of `_scheduled_refresh` is not as expected.

TheoMathurin:
## Description
Fixes #8517

Several ways to go here, I went for the replacement of the `user not in state._oauth_user_overrides` condition for a direct check on the output of `_scheduled_refresh`. The former is only one path that leads to the outputs being `None`.

Also I chose to check on every single of the three outputs to be sure, but that's probably not strictly necessary.

Happy to consider alternatives. We could for instance, when the refresh fails, choose to raise an exception and catch it.

## How Has This Been Tested?
As I could not create an MRE, I could not test but the logic is pretty straightforward.


philippjfr:
I think there could be OAuth implementations that do not provide a `refresh_token`, which means this new condition would always `return None` and cause an auth loop. I think the same could be true of the expiry.

TheoMathurin:
Indeed, what about just checking on `access_token` then? We could also make the function return a single `None` instead of three and rearrange the bits where it's called.

philippjfr:
That makes sense to me.

TheoMathurin:
Can we rerun the tests? I don't think the failures are related to these changes.

PR #8521, opened 2026-03-18
Add geo_data parameter to ECharts for geo/map support

### LINKED ISSUE #8506: Echarts cannot show map data
# ECharts geo map fails with "Cannot read properties of undefined (reading 'regions')"

## Description

Using `pn.pane.ECharts` with a `geo` component and `"map": "world"` raises a JavaScript error at runtime:

```
TypeError: Cannot read properties of undefined (reading 'regions')
    at t.getFilledRegions (echarts.min.js)
    at e.optionUpdated (echarts.min.js)
```

ECharts requires world map GeoJSON to be registered via `echarts.registerMap('world', geoJson)` before a `geo` component can render. The ECharts bundle bundled with Panel does not include or register any map data, so the `regions` property is undefined.

## Minimal reproducible example

```python
import panel as pn

pn.extension("echarts")

echart = {
    "geo": {
        "map": "world",
        "roam": True,
    },
    "series": [{
        "type": "scatter",
        "coordinateSystem": "geo",
        "data": [{"name": "London", "value": [-0.12, 51.51]}],
    }],
}

pn.pane.ECharts(echart, height=400).servable()
```

Run with `panel serve script.py` and open the browser — the map area is blank and the browser console shows the `regions` TypeError.

Image

## Expected behaviour

The geo map renders with country outlines, or Panel raises a clear Python-level error explaining that map data must be registered before use.

## Environment

```
panel: 1.x
echarts: bundled with Panel
browser: Chrome / Firefox
```

## Possible fix

Allow users to pass a GeoJSON URL or dict to `pn.pane.ECharts` that Panel registers via `echarts.registerMap(name, geoJson)` before mounting the chart. Alternatively, bundle a lightweight world map GeoJSON and register it automatically when `"map": "world"` is detected in the option.

## More interesting example

```python
import panel as pn

pn.extension("echarts")

cities = [
    ("London", -0.12, 51.51, 42),
    ("Paris", 2.35, 48.85, 55),
    ("Berlin", 13.40, 52.52, 38),
    ("Madrid", -3.70, 40.42, 61),
    ("Rome", 12.50, 41.90, 49),
    ("New York", -74.01, 40.71, 78),
    ("Los Angeles", -118.24, 34.05, 92),
    ("Chicago", -87.63, 41.88, 65),
    ("Tokyo", 139.69, 35.69, 55),
    ("Beijing", 116.40, 39.90, 135),
    ("Shanghai", 121.47, 31.23, 118),
    ("Mumbai", 72.88, 19.08, 88),
    ("São Paulo", -46.63, -23.55, 72),
    ("Cairo", 31.24, 30.06, 95),
    ("Sydney", 151.21, -33.87, 33),
    ("Dubai", 55.30, 25.20, 45),
    ("Singapore", 103.82, 1.35, 40),
    ("Moscow", 37.62, 55.75, 70),
    ("Toronto", -79.38, 43.65, 48),
    ("Mexico City", -99.13, 19.43, 105),
]

scatter_data = [{"name": c[0], "value": [c[1], c[2], c[3]]} for c in cities]

echart = {
    "backgroundColor": "#1a2035",
    "title": {"text": "Urban Air Quality Index", "left": "center",
              "textStyle": {"color": "#eee", "fontSize": 16}},
    "tooltip": {"trigger": "item"},
    "visualMap": {
        "min": 30, "max": 140,
        "calculable": True,
        "inRange": {"color": ["#50fa7b", "#f1fa8c", "#ffb86c", "#ff5555"]},
        "textStyle": {"color": "#ccc"},
        "orient": "horizontal",
        "left": "center",
        "bottom": 10,
    },
    "geo": {
        "map": "world",
        "roam": True,
        "label": {"show": False},
        "itemStyle": {"areaColor": "#2a3a5c", "borderColor": "#4a6080"},
        "emphasis": {"itemStyle": {"areaColor": "#3a5a8c"}},
    },
    "series": [{
        "type": "effectScatter",
        "coordinateSystem": "geo",
        "data": scatter_data,
        "symbolSize": 12,
        "rippleEffect": {"brushType": "stroke"},
        "label": {"show": True, "formatter": "{b}", "position": "right",
                  "color": "#ccc", "fontSize": 10},
        "zlevel": 1,
    }],
}

pn.pane.ECharts(echart, height=520, sizing_mode="stretch_width").servable()
```

ghostiee-11:
## Description

Fixes #8506

I was testing ECharts geo/map charts in Panel and hit `TypeError: Cannot read properties of undefined (reading 'regions')`. The root cause is that ECharts requires `echarts.registerMap(name, geoJson)` before `setOption()`, but Panel had no way to do this.

I added a `geo_data` parameter that accepts GeoJSON dicts or URL strings and registers them before the chart renders. When the config references a map name (like `"world"`) that is not provided, Panel auto-fetches it from the ECharts CDN so the original MRE from the issue works without any extra parameters.

I also noticed a pre-existing bug in `_subscribe()` where the query null-check was inverted - it was passing `null` as the query filter and omitting the actual query string. Fixed that too.

**After:**
echarts_geo_map_after

## How Has This Been Tested?

16 unit tests and 1 UI test added. Verified both MREs from the issue render correctly with `panel serve`.

```python
# This now works with zero extra code
pn.pane.ECharts({"geo": {"map": "world"}, "series": [...]}, height=400)

# Or pass GeoJSON explicitly
pn.pane.ECharts(config, geo_data={"world": geojson_dict})

# Or pass a URL string
pn.pane.ECharts(config, geo_data={"world": "https://example.com/world.json"})
```

## Checklist

- [ ] Added documentation

## AI Disclosure

Used AI for planning the fix approach and code scaffolding. All testing, debugging, and verification were done manually.

### DIFF
--- a/examples/reference/panes/ECharts.ipynb
+++ b/examples/reference/panes/ECharts.ipynb
@@ -179,6 +179,41 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
+    "### Geo / map charts\n",
+    "\n",
+    "ECharts geo and map series require map data to be registered before the chart renders. ",
+    "Pass GeoJSON as a dict, or a URL string, via the ``geo_data`` parameter; ",
+    "the data is registered with ``echarts.registerMap`` in the browser before ``setOption`` is called. ",
+    "Map names referenced by the config but missing from ``geo_data`` are auto-fetched from the ECharts CDN."
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "world_map = {\n",
+    "    'geo': {'map': 'world', 'roam': True},\n",
+    "    'series': [{\n",
+    "        'type': 'scatter',\n",
+    "        'coordinateSystem': 'geo',\n",
+    "        'symbolSize': 10,\n",
+    "        'data': [\n",
+    "            {'name': 'London', 'value': [-0.12, 51.51]},\n",
+    "            {'name': 'New York', 'value': [-74.01, 40.71]},\n",
+    "            {'name': 'Tokyo', 'value': [139.69, 35.69]},\n",
+    "        ],\n",
+    "    }],\n",
+    "}\n",
+    "\n",
+    "pn.pane.ECharts(world_map, height=400)"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
     "## Events\n",
     "\n",
     "The `EChart` object allows you to listen to any event defined in the Javascript API, either by listening to the event in Python using the `on_event` method or by triggering a Javascript callback with the `js_on_event` method.\n",
--- a/panel/models/echarts.py
+++ b/panel/models/echarts.py
@@ -56,6 +56,8 @@ class ECharts(LayoutDOM):
 
     data = Nullable(Dict(String, Any))
 
+    geo_data = Dict(String, Any)
+
     options = Nullable(Dict(String, Any))
 
     event_config = Dict(String, Any)
--- a/panel/models/echarts.ts
+++ b/panel/models/echarts.ts
@@ -22,6 +22,38 @@ const events = [
 
 const all_events = mouse_events.concat(events)
 
+const ECHARTS_MAP_CDN = "https://cdn.jsdelivr.net/npm/echarts@4.9.0/map/json/{name}.json"
+const _geojson_cache: Map<string, Promise<unknown>> = new Map()
+
+function fetch_geojson(url: string): Promise<unknown> {
+  let p = _geojson_cache.get(url)
+  if (p == null) {
+    p = fetch(url).then((r) => {
+      if (!r.ok) {
+        throw new Error(`HTTP ${r.status}`)
+      }
+      return r.json()
+    })
+    _geojson_cache.set(url, p)
+  }
+  return p
+}
+
+function collect_map_names(data: any): Set<string> {
+  const names = new Set<string>()
+  const geo = data?.geo
+  if (geo != null) {
+    const arr = Array.isArray(geo) ? geo : [geo]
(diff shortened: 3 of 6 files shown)


philippjfr (commented):
Overall I'd be happy to merge this PR but do have one question. Does ECharts not have a mechanism to fetch the GeoJSON automatically on the frontend when rendering?

ghostiee-11:
Good question. ECharts itself does not auto-fetch — it requires `echarts.registerMap(name, geoJson)` before `setOption()`, otherwise the #8506 error fires.

This PR does the fetch server-side in `_resolve_geo_data` when the config references a map name not in `geo_data`, then ships the resolved dict to the frontend which just calls `registerMap`.

I could move the fetch to `echarts.ts` (browser does it, works in static exports, no server network) instead of Python. Happy to switch if you prefer the frontend pattern, let me know.

philippjfr:
@ghostiee-11 Yes, I think fetching on the frontend would be nicer here.

ghostiee-11:
Ohh okay philip!! Will switch this

ghostiee-11:
Updated this to do the GeoJSON fetching on the frontend instead of in Python, per the review suggestion. The browser fetches URL-string entries in `geo_data` and auto-fetches map names referenced by the config but missing from `geo_data`, then calls `echarts.registerMap` before `setOption`. No more blocking `urllib` calls on the server side, and it works in pyodide / static exports too.

While I was at it I trimmed unrelated changes from the diff so it only touches the geo path.

One thing I noticed while testing: ECharts 6.0.0 (the version Panel bundles today) actually ships world-map polygon data inside `echarts.min.js` itself, so `geo: {map: "world"}` already renders without registering anything. The original `regions` TypeError from the issue still reproduces on any non-built-in map though (china, france, japan, etc.), and the fix covers those. Verified locally: pre-PR shows the `regions` TypeError on a non-built-in map, with the fix the map renders and registers correctly.

PR #8522, opened 2026-03-20
Fix model.useState boomerang causing character loss during fast typing

### LINKED ISSUE #8523: `model.useState` boomerang causes character loss in React ESM text inputs
## Description

When typing fast in any `ReactComponent` that uses `model.useState()` for a controlled text input, characters are dropped. This is the React ESM equivalent of the boomerang issue fixed for Bokeh components in PR #7093.

Here is an example from my deployed panel-material-ui ChatInterface application.

https://github.com/user-attachments/assets/aa34a2f6-621a-492c-af0a-5a134cbfa96a

I typed "boomerang. Sometimes". But it became "boomerang. Sometims". I'm normally only able to reproduce for deployed applications. While developing on my jupyterhub I don't see the issue.

## AI Disclaimer

Below is AI Generated. I've

- Test the MRE can be served. But as I state I can not reproduce issues in development. Only in production. Same was the case for #7096.

## Minimal Reproducible Example

```python
import param
import panel as pn

from panel.custom import ReactComponent

pn.extension()


class TypingTest(ReactComponent):
    value_input = param.String(default="")

    _esm = """
    export function render({model}) {
      const [value_input, setValueInput] = model.useState("value_input")
      return (
        <div>
          <textarea
            value={value_input}
            onChange={(e) => setValueInput(e.target.value)}
          />
          <p>value_input: {value_input}</p>
          <p>length: {value_input.length}</p>
        </div>
      )
    }
    """

TypingTest().servable()
```

Run with `panel serve script.py` and type fast — characters will be dropped.

## Root Cause

`model.useState` in `panel/models/react_component.ts` (lines 497-540) creates a bidirectional sync with no boomerang prevention:

- **Effect 1** (Python → React): calls `setValue(targetModel.attributes[resolvedProp])` on any property change
- **Effect 2** (React → Python): calls `targetModel.setv({prop: value})` when React state changes

The race condition:

1. User types "a" → `setValueInput("a")` → Effect 2 sends "a" to Python
2. User types "b" → `setValueInput("ab")` → Effect 2 sends "ab" to Python
3. Python echoes "a" back → Effect 1 fires `setValue("a")` → **overwrites "ab"**
4. Character "b" is lost

## Real-world Impact

This affects [panel-material-ui](https://github.com/panel-extensions/panel-material-ui) ChatInterface and all text input widgets (TextInput, TextAreaInput, PasswordInput, AutocompleteInput) — any component using the `model.useState` + controlled input pattern.

## Precedent

This is the same class of bug fixed for Bokeh components in:

- **PR #7093** — "Avoid events boomeranging from frontend" (Panel v1.5.0)
- **Issue #7096** — ChatAreaInput drops characters when typing fast
- **Issue #6414** — TextInput loses input when typing quickly with thread pool
- **Issue #6738** — CodeEditor unusable with automatic threading

## Proposed Fix

Add boomerang detection to `model.useState` in `react_component.ts`. Track values sent to Python via a ref array. When the Python → React callback fires, check if the incoming value is an echo of a locally-sent value — if so, skip the `setValue()` call.

```typescript
const sentRef = React.useRef([])

// Effect 1: Python → React — skip echoes
React.useEffect(() => {
  const cb = () => {
    if (target.model.events.includes(resolvedProp)) {
      // ... event handling unchanged ...
    } else {
      const incoming = targetModel.attributes[resolvedProp]
      const idx = sentRef.current.indexOf(incoming)
      if (idx !== -1) {
        sentRef.current.splice(0, idx + 1)  // Echo — discard
      } else {
        sentRef.current.length = 0
        setValue(incoming)  // Genuine server update — apply
      }
    }
  }
  react_proxy.on(prop, cb, true)
  return () => react_proxy.off(prop, cb)
}, [])

// Effect 2: React → Python — record sent values
React.useEffect(() => {
  if (!target.model.events.includes(resolvedProp)) {
    sentRef.current.push(value)
    targetModel.setv({ [resolvedProp]: value })
  }
}, [value])
```

## Environment

- Panel: 1.8.10 (but affects all versions with React ESM support)
- Python: 3.12
- Browser: Chromium (all browsers affected)

MarcSkovMadsen:
## AI Disclaimer

Below is AI Generated. I will start reviewing and update.

## Summary

- Fixes character loss when typing fast in React ESM components that use `model.useState()` with controlled inputs
- Adds echo detection to `model.useState` in `react_component.ts` — tracks values sent to Python via a ref array and discards stale echoes from the Python → React callback
- This is the React ESM equivalent of the boomerang fix in PR #7093 (which fixed Bokeh components)

## Root Cause

`model.useState` creates two React effects:
1. **Python → React**: calls `setValue(targetModel.attributes[prop])` on any property change
2. **React → Python**: calls `targetModel.setv({prop: value})` when React state changes

When typing fast, the round-trip creates a race condition:
1. User types "a" → `setValue("a")` → syncs "a" to Python
2. User types "b" → `setValue("ab")` → syncs "ab" to Python  
3. Python echoes stale "a" back → `setValue("a")` → **overwrites "ab"** → character "b" lost

## Fix

A `sentRef` (React ref array) tracks values sent to Python. When the Python → React callback fires:
- If the incoming value is in `sentRef` → it's an echo → skip `setValue()`
- If not → it's a genuine server update → apply `setValue()`

## Real-world Impact

This fixes character loss in [panel-material-ui](https://github.com/panel-extensions/panel-material-ui) ChatInterface and all text input widgets (TextInput, TextAreaInput, PasswordInput, AutocompleteInput), as well as any custom `ReactComponent` using the `model.useState` + controlled input pattern.

## MRE

```python
import param
import panel as pn
from panel.custom import ReactComponent

pn.extension()

class TypingTest(ReactComponent):
    value_input = param.String(default="")
    _esm = """
    export function render({model}) {
      const [value_input, setValueInput] = model.useState("value_input")
      return (
        <div>
          <textarea
            value={value_input}
            onChange={(e) => setValueInput(e.target.value)}
          />
          <p>value_input: {value_input}</p>
        </div>
      )
    }
    """

TypingTest().servable()
```

Run with `panel serve script.py` and type fast — characters are dropped without this fix.

## Test plan

- [x] Playwright test: type "The quick brown fox jumps" at 30ms delay into custom ReactComponent — all characters present
- [x] Playwright test: same test with panel-material-ui ChatInterface — all characters present
- [ ] Verify no regressions in existing React ESM component tests
- [ ] Verify event-type properties (`param.Event`) still work correctly (code path unchanged)

### DIFF
--- a/panel/tests/test_custom.py
+++ b/panel/tests/test_custom.py
@@ -1,11 +1,12 @@
 import numpy as np
 import pandas as pd
 import param
+import pytest
 
 from bokeh.plotting import figure
 
-from panel.custom import PyComponent, ReactiveESM
-from panel.io.state import state
+from panel.custom import PyComponent, ReactComponent, ReactiveESM
+from panel.io.state import set_curdoc, state
 from panel.layout import Row
 from panel.pane import Bokeh, Markdown
 from panel.util import edit_readonly
@@ -225,3 +226,46 @@ def test_esm_parameter_override(document, comm):
     esm.width = 84
 
     assert model.width == 84
+
+
+class ESMInput(ReactiveESM):
+
+    value_input = param.String(default='')
+
+    _esm = """
+    export function render({model}) {}
+    """
+
+
+class ReactInput(ReactComponent):
+
+    value_input = param.String(default='')
+
+    _esm = """
+    export function render({model}) {
+      const [value_input, setValueInput] = model.useState("value_input")
+      return <input value={value_input} onChange={(e) => setValueInput(e.target.value)} />
+    }
+    """
+
+
+@pytest.mark.parametrize('component', [ESMInput, ReactInput])
+def test_esm_property_change_does_not_boomerang(document, comm, component):
+    widget = component(value_input='A')
+
+    model = widget.get_root(document, comm)
+
+    assert model.data.value_input == 'A'
+
+    # Simulate: frontend updates the property to 'B' (user typed)
+    model.data.value_input = 'B'
+
+    # Simulate: a different frontend event arrives for processing
+    with set_curdoc(document):
+        widget._process_events({'value_input': 'C'})
+
+    # The Bokeh model should retain 'B' (the latest frontend value),
+    # NOT be overwritten to 'C' (which would be a boomerang).
+    assert model.data.value_input == 'B'
+    # The Python param should still be updated to 'C'.
+    assert widget.value_input == 'C'
--- a/panel/tests/test_custom.py
+++ b/panel/tests/test_custom.py
@@ -230,7 +230,7 @@ def test_esm_parameter_override(document, comm):
 
 class ESMInput(ReactiveESM):
 
-    value_input = param.String(default='')
+    value_input = param.String(default='', doc="Text value updated on every key press.")
 
     _esm = """
     export function render({model}) {}
@@ -239,7 +239,7 @@ class ESMInput(ReactiveESM):
 
 class ReactInput(ReactComponent):
 
-    value_input = param.String(default='')
+    value_input = param.String(default='', doc="Text value updated on every key press.")
 
     _esm = """
     export function render({model}) {
--- a/panel/tests/ui/test_custom.py
+++ b/panel/tests/ui/test_custom.py
@@ -1325,3 +1325,34 @@ def test_esm_compile_shared(page, components):
 
     expect(page.locator(f'#{example[0].name}')).to_have_text('Rendered')
     expect(page.locator(f'#{example[1].name}')).to_have_text('Rendered')
+
+
+def test_usestate_no_character_loss(page):
+    """Fast typing into a controlled input using model.useState should not lose characters."""
+
+    class TypingTest(ReactComponent):
+        value_input = param.String(default="")
+        _esm = """
+        export function render({model}) {
+          const [value_input, setValueInput] = model.useState("value_input")
+          return (
+            <textarea
+              id="typing-test"
+              value={value_input}
+              onChange={(e) => setValueInput(e.target.value)}
+            />
+          )
+        }
+        """
+
+    test_string = "The quick brown fox jumps"
+    widget = TypingTest()
+    serve_component(page, widget)
+
+    textarea = page.locator("#typing-test")
+    expect(textarea).to_be_visible(timeout=10000)
+    textarea.click()
+    textarea.press_sequentially(test_string, delay=30)
+
+    wait_until(lambda: widget.value_input == test_string, page, timeout=10000)
+    assert widget.value_input == test_string
--- a/panel/custom.py
+++ b/panel/custom.py
@@ -600,10 +600,16 @@ class ReactiveESM(ReactiveCustomBase, metaclass=ReactiveESMMetaclass):
         root: Model, model: Model, doc: Document, comm: Comm | None
     ) -> None:
         model_msg, data_msg, data_resets  = {}, {}, {}
+        curdoc_events = self._in_process__events.get(doc, {})
         for prop, v in list(msg.items()):
             if prop in list(Reactive.param)+['esm', 'importmap']:
                 model_msg[prop] = v
             elif prop in model.children:
+                continue
+            elif prop in curdoc_events and v is curdoc_events[prop]:
+                # Do not apply change that originated directly from
+                # the frontend since this may cause boomerang if a
+                # new property value is already in-flight
                 continue
             else:
                 data_msg[prop] = v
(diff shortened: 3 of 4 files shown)


philippjfr (approved):

philippjfr:
Fix looks reasonable.

MarcSkovMadsen:
## AI Disclaimer

Below is AI Generated but validated by a human (MarcSkovMadsen)

## Test validation: unit test fails on `main`, passes with fix

To verify the tests are meaningful regression tests, I ran the unit test against both `origin/main` (unfixed) and `fix/usestate-boomerang` (fixed) using a git worktree.

### Unit test: `test_esm_property_change_does_not_boomerang`

This test simulates the Python-side boomerang: the frontend sets `value_input = 'B'` on the Bokeh model, then `_process_events({'value_input': 'C'})` arrives. The model should retain `'B'` (not be overwritten to `'C'`).

**On `main` (unfixed): FAILED**
```
> assert model.data.value_input == 'B'
E AssertionError: assert 'C' == 'B'
```
The `ReactiveESM._update_model` echoes `'C'` back to the Bokeh model, overwriting the frontend's `'B'`.

**On `fix/usestate-boomerang`: PASSED** (both `ESMInput` and `ReactInput` parametrizations)

### Playwright test: `test_usestate_no_character_loss`

This test types "The quick brown fox jumps" at 30ms delay. In local in-process testing the round-trip is nearly instant, so the JS-side boomerang doesn't reliably trigger (same as the original issue — character loss only manifests in deployed apps with network latency). The test serves as an additional safety net but the unit test is the reliable regression test for CI.

### Note on the JS-side fix

The `react_component.ts` fix (`sentRef` echo detection in `model.useState`) prevents the JS-side boomerang that causes character loss in production. This cannot be reliably tested without simulating network latency, but the mechanism is analogous to the Python-side fix verified above — both prevent stale echoed values from overwriting fresher state.

panel/tests/ui/test_custom.py
```diff
@@ -1325,3 +1325,34 @@ def test_esm_compile_shared(page, components):
 
     expect(page.locator(f'#{example[0].name}')).to_have_text('Rendered')
     expect(page.locator(f'#{example[1].name}')).to_have_text('Rendered')
+
+
+def test_usestate_no_character_loss(page):
```
MarcSkovMadsen, line 1330: This does not fail on existing `main`. We believe its because we don't have same network latency for dev/ test as in production. But should we then add the test?

panel/tests/test_custom.py
```diff
@@ -225,3 +226,46 @@ def test_esm_parameter_override(document, comm):
...
+    _esm = """
+    export function render({model}) {
+      const [value_input, setValueInput] = model.useState("value_input")
+      return <input value={value_input} onChange={(e) => setValueInput(e.target.value)} />
+    }
+    """
+
+
+@pytest.mark.parametrize('component', [ESMInput, ReactInput])
+def test_esm_property_change_does_not_boomerang(document, comm, component):
```
MarcSkovMadsen, line 253: According to AI tests this fails on the existing `main` branch.

panel/tests/test_custom.py
```diff
@@ -225,3 +226,46 @@ def test_esm_parameter_override(document, comm):
...
+
+    # Simulate: a different frontend event arrives for processing
+    with set_curdoc(document):
+        widget._process_events({'value_input': 'C'})
+
+    # The Bokeh model should retain 'B' (the latest frontend value),
+    # NOT be overwritten to 'C' (which would be a boomerang).
+    assert model.data.value_input == 'B'
+    # The Python param should still be updated to 'C'.
+    assert widget.value_input == 'C'
```
MarcSkovMadsen, line 271: I see this is a test similar to original bokeh test. But I don't fully understand why its ok for model.data.value_input and widget.value_input to get out of sync in the example?

MarcSkovMadsen:
I've tried to do the work understanding and fixing this. I believe this is the right fix because its a fix similar to fix for original bokeh issue.

But as I don't fully understand the js <-> bokeh <-> panel/ param communication I don't fully understand the quality of this PR @philippjfr . 

I really tried with the help of AI to find a playwright test that could reproduce this on the `main` branch. But I was not able to.

panel/tests/ui/test_custom.py
philippjfr: Vibe coded but does something like this work:

```python
def add_ws_latency(page, pattern="**/ws", up_ms=300, down_ms=800):
    context = page.context

    def handle_ws(ws):
        server = ws.connect_to_server()

        def from_page(message):
            import time
            time.sleep(up_ms / 1000)
            server.send(message)

        def from_server(message):
            import time
            time.sleep(down_ms / 1000)
            ws.send(message)

        ws.on_message(from_page)
        server.on_message(from_server)

    context.route_web_socket(pattern, handle_ws)


add_ws_latency(page.context)
```

philippjfr:
Closing, as https://github.com/holoviz/panel/pull/8617 provides the more correct, backend only fix.

PR #8528, opened 2026-03-20
Optimize Tabulator resize events

philippjfr:
Tabulator resize events are expensive (100s of milliseconds) and even though we had debouncing implemented on them it turned out that `after_layout` didn't use the debounced version and the `ResizeObserver` fired very eagerly while a parent was rendering. A common example might be a Column containing some widgets and a Tabulator that stretches in height, the widgets might render after the Tabulator, which then causes a resize event to occur when the widgets render and take up space. Instead of resizing then we should wait until the parent (or rather the root) says all its children are idle, i.e. resizing is the last thing that should happen.

Here we await the `root.ready` Promise to ensure everything has settled before the trigger the resize. Additionally we measure the width and height and avoid resizing the table unless it has measurably changed.


panel/models/tabulator.ts
```diff
@@ -549,24 +554,77 @@ export class DataTabulatorView extends HTMLBoxView {
...
+        return
+      }
+      await this.root.ready
+      // `remove()` can clear `_resize_pending` while awaiting `root.ready`.
+      /* eslint-disable-next-line @typescript-eslint/no-unnecessary-condition -- cleared asynchronously in `remove()` */
+      if (!this._resize_pending) {
+        continue
+      }
+      if (this._is_scrolling || this._initializing || this._building || this.container === null || this.is_drawing) {
+        await new Promise<void>((resolve) => setTimeout(resolve, 0))
```
mattpap: There's `defer()` in `@bokehjs/core/util/defer` for this purpose and also `delay()` for non-zero timeout.

philippjfr: Nice, thanks!

Issue #5000, opened 2023-06-01
Tabulator Patch always resets scroll position


Wijnko:
#### ALL software version info
Panel 1.0.4
Bokeh 3.1.1
MacOS 12.6
Chrome Version 113.0.5672.126

#### Description of expected behavior and the observed behavior

Expected:
Scroll position is untouched on data patching

Observed:
Table scrolls to top after every data patch

#### Complete, minimal, self-contained example code that reproduces the issue

```
import panel as pn
import pandas as pd
pn.extension('tabulator')

sample_data = pd.DataFrame({'Number': [0]*10})

last_row_index = sample_data.index[-1]
last_row_value = sample_data['Number'].loc[last_row_index]

table = pn.widgets.Tabulator(sample_data, height=200)

button = pn.widgets.Button(name='Increase the value of the last row')

def increase_last_number(event):
    global last_row_value
    last_row_value += 1
    table.patch({'Number': [(last_row_index, last_row_value)]})

button.on_click(increase_last_number)

pn.Column(
    table,
    button
).servable()

```

#### Screenshots or screencasts of the bug in action

![panel_tabulator_pitch_scroll_issue](https://github.com/holoviz/panel/assets/17784946/80402bef-72bc-496f-874e-c1c7188e1bde)

#### Note:

By changing the `updateOrAddData` method and removing one line in `/panel/models/tabulator.ts`, I was able to get the expected behaviour:

Changes:

- Removed the line `this.tabulator.rowManager.element.scrollTop = this._lastVerticalScrollbarTopPosition;` from the patching event handler (https://github.com/holoviz/panel/blob/main/panel/models/tabulator.ts#L352).

The `_lastVerticalScrollbarTopPosition` variable never gets set to a value other than 0 on Initialisation (https://github.com/holoviz/panel/blob/main/panel/models/tabulator.ts#L290)

- Changed `this.tabulator.setData(data)` to `this.tabulator.updateOrAddData(data)` in updateOrAddData method (https://github.com/holoviz/panel/blob/main/panel/models/tabulator.ts#L877-L885)

```
  updateOrAddData(): void {
    // To avoid double updating the tabulator data
    if (this._tabulator_cell_updating)
      return

    let data = transform_cds_to_records(this.model.source, true)
    this.tabulator.updateOrAddData(data)
    this.postUpdate()
  }
```

Result:

[...]

PR #8531, opened 2026-03-25
fix(tabulator): Scroll reset when filtering on patch 

hoxbro:
## Description

Started out with improving flaky UI test. But found a bug where scroll would reset if patch was used with filtering.

## AI Disclosure

Tools and Models: Cursor + Sonnet 4.6, used to try to fix the flaky tabulator test.

## Checklist


hoxbro:
I think this should also fix https://github.com/holoviz/panel/issues/5000

philippjfr (approved):

Wijnko:
Thanks for re-finding and fixing this after three years! ;-)

PR #8539, opened 2026-03-29
Fix DataFrame comparison in Tabulator.from_param

### LINKED ISSUE #8538: Tabulator.from_param fails as soon as the value is changed.

<details>

```plaintext
panel   1.8.10
pandas  3.0.1
param   2.3.2
bokeh   3.8.2
python  3.13.5
Ubuntu  24.04.3 LTS (in WSL)
```
</details>

When I create a Tabulator with `Tabulator.from_param`, updating the value fails with `ValueError`.

#8492 introduced a comparison between the current and the new value and expected to result to be a bool (see `panel/param.py:575`). In this case the values are DataFrames that do not return a bool but another DataFrame (`type(pd.DataFrame() != pd.DataFrame()) == pd.DataFrame`).

```python
import pandas as pd
import panel as pn
import param


class MyObject(param.Parameterized):
    data = param.DataFrame()


my_object = MyObject(data=pd.DataFrame({"a": [1, 2, 3]}))

tabulator = pn.widgets.Tabulator.from_param(my_object.param.data)

tabulator.value = pd.DataFrame({"a": [1, 2, 5]})
```

```
  File "***/test.py", line 14, in <module>
    tabulator.value = pd.DataFrame({"a": [1, 2, 5]})
    ~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^
  File "***/.venv/lib/python3.13/site-packages/param/parameterized.py", line 2970, in trigger
    self_.update(dict(params, **triggers))
    ~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "***/.venv/lib/python3.13/site-packages/param/parameterized.py", line 2788, in update
    restore = dict(self_._update(arg, **kwargs))
                   ~~~~~~~~~~~~~^^^^^^^^^^^^^^^
  File "***/.venv/lib/python3.13/site-packages/param/parameterized.py", line 2821, in _update
    self_._batch_call_watchers()
    ~~~~~~~~~~~~~~~~~~~~~~~~~~^^
  File "***/.venv/lib/python3.13/site-packages/param/parameterized.py", line 3038, in _batch_call_watchers
    ~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^
  File "***/.venv/lib/python3.13/site-packages/param/parameterized.py", line 3000, in _execute_watcher
    watcher.fn(*args, **kwargs)
    ~~~~~~~~~~^^^^^^^^^^^^^^^^^
  File "***/.venv/lib/python3.13/site-packages/panel/param.py", line 575, in link_widget
    if not reset and current_val != new:
                     ^^^^^^^^^^^^^^^^^^
  File "***/.venv/lib/python3.13/site-packages/pandas/core/generic.py", line 1513, in __bool__
    raise ValueError(
    ...<2 lines>...
    )
ValueError: The truth value of a DataFrame is ambiguous. Use a.empty, a.bool(), a.item(), a.any() or a.all().
```

zhangdomi:
## Description

This PR fixes a bug in Tabulator.from_param where updating the widget value would incorrectly trigger updates back to the parameter due to DataFrame comparison issues.

Fixes #8538 
 
## How Has This Been Tested?

Created unit test `panel/tests/widgets/test_tabulator.py`
## AI Disclosure

Tools and Models: Claude Sonnet 4.5

## Checklist

### DIFF
--- a/panel/param.py
+++ b/panel/param.py
@@ -21,7 +21,6 @@ from functools import partial
 from types import FunctionType
 from typing import TYPE_CHECKING, Any, ClassVar
 
-import numpy as np
 import param
 
 try:
@@ -38,6 +37,8 @@ from param.parameterized import (
 )
 from param.reactive import rx
 
+from panel.io.cache import is_equal
+
 from .config import config
 from .io import state
 from .layout import (
@@ -50,9 +51,7 @@ from .util import (
     abbreviated_repr, flatten, full_groupby, fullpath, is_parameterized,
     param_name, recursive_parameterized, to_async_gen,
 )
-from .util.checks import (
-    is_array, is_dataframe, is_mpl_axes, is_series,
-)
+from .util.checks import is_dataframe, is_mpl_axes, is_series
 from .viewable import Layoutable, Viewable
 from .widgets import (
     ArrayInput, Button, Checkbox, ColorPicker, DataFrame, DatePicker,
@@ -549,18 +548,6 @@ class Param(Pane):
             watchers, updating = [], []
             widgets = {p_name: widget}
 
-        def _equal(current, new):
-            if (is_dataframe(current) and is_dataframe(new)) or (
-                is_series(current) and is_series(new)
-            ):
-                return current.equals(new)
-            if is_array(current) and is_array(new):
-                return np.array_equal(current, new)
-            try:
-                return bool(current == new)
-            except Exception:
-                return True
-
         def link_widget(change):
             p_key = p_name if config.nthreads is None else (threading.get_ident(), p_name)
             if p_key in updating:
@@ -587,7 +574,7 @@ class Param(Pane):
                 if reset:
                     widget.value = new
                 current_val = getattr(parameterized, p_name)
-                if not reset and not _equal(current_val, new):
+                if not reset and not is_equal(current_val, new):
                     is_w = isinstance(widget, Row) and len(widget) == 2
                     target = widget[0] if is_w else widget
                     target.param.update(value=current_val)
--- a/panel/param.py
+++ b/panel/param.py
@@ -21,6 +21,7 @@ from functools import partial
 from types import FunctionType
 from typing import TYPE_CHECKING, Any, ClassVar
 
+import numpy as np
 import param
 
 try:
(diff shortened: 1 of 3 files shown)


panel/param.py
```diff
@@ -546,6 +546,16 @@ def widget(self_or_cls, p_name: str, parameterized: param.Parameterized | None =
...
             widgets = {p_name: widget}
 
+        def _equal(current, new):
+            if (is_dataframe(current) and is_dataframe(new)) or (
+                is_series(current) and is_series(new)
+            ):
+                return current.equals(new)
+            try:
+                return current == new
+            except Exception:
```
dejfh: This except-clause will most likely never be called. In the case of a DataFrame the problem occurs not during `current == new`, but when using it in a condition.
One could do `return bool(current == new)` to force the exception here.

panel/param.py
```diff
@@ -546,6 +546,16 @@ def widget(self_or_cls, p_name: str, parameterized: param.Parameterized | None =
...
+                return False
```
dejfh: Returning `False` here triggers an update of the widget parameter. Is that the intended fallback behavior? IMHO that is a bit dangerous, since the comparison should be a protection against infinite loops.

panel/param.py
zhangdomi: I agree, I updated the code

panel/param.py
zhangdomi: Good point, thanks. I’ve updated the fallback to return `True` so that in uncertain cases we conservatively avoid triggering updates. I also added a dedicated check for NumPy arrays using np.array_equal to handle ambiguous equality results similar to the DataFrame case.

panel/param.py
```diff
@@ -546,6 +549,18 @@ def widget(self_or_cls, p_name: str, parameterized: param.Parameterized | None =
             watchers, updating = [], []
             widgets = {p_name: widget}
 
+        def _equal(current, new):
```
philippjfr: Should probably just use `panel.io.cache.is_equal` here.

zhangdomi: Good call, I've updated the code to use `panel.io.cache.is_equal` instead. Thanks for the suggestion!

panel/param.py
```diff
@@ -37,6 +37,8 @@ class Skip(Exception):  # type: ignore
 )
 from param.reactive import rx
 
+from panel.io.cache import is_equal
+
 from .config import config
 from .io import state
```
```suggestion
from .config import config
from .io import state
from .io.cache import is_equal
```

philippjfr (approved):

philippjfr:
Thank you @zhangdomi, much appreciated!

Issue #5849, opened 2023-11-11
Speed up initial load 5-10% by implementing faster self.param.values().items()


MarcSkovMadsen:
I'm on the `main` branch of Panel trying to identify performance issues. I've tried profiling different apps and every things `values` shows up in the profiler taking 6-10% of the time.

On further inspecting I can see `values` in many places. For example here

![image](https://github.com/holoviz/panel/assets/42288570/8ad712f3-a4cc-414a-86c5-421ad9f44c0a)

To me this looks highly inefficient. But I don't know `param` well enough to know if you really have to use this to get the correct dictionary of parameter names and values in all cases.

## Reproduce

```python
import panel as pn

N=10

pn.Column(
    *[pn.Row(
        *[pn.pane.Markdown("Hello World") for i in range(N)]
    ) for i in range(N)]
).servable()
```

```bash
panel serve script.py --autoreload --index script --admin --profiler pyinstrument
```

Open the app 5 times and inspect the pyinstrument report.

![image](https://github.com/holoviz/panel/assets/42288570/0508c11a-af42-4413-a90b-4cb7841b4dd0)

![image](https://github.com/holoviz/panel/assets/42288570/09be94d8-eef7-4dde-8daa-7b6ef74e205f)

## Fix

My simple approach would be to just loop `.param` and look up the value on the fly.

```python
from pathlib import Path
import pytest
from pyinstrument import Profiler
import panel as pn

TESTS_ROOT = Path.cwd()

@pytest.fixture(autouse=True)
def auto_profile(request):
    PROFILE_ROOT = (TESTS_ROOT / ".profiles")
    # Turn profiling on
    profiler = Profiler()
    profiler.start()

    yield  # Run test

    profiler.stop()
    PROFILE_ROOT.mkdir(exist_ok=True)
    results_file = PROFILE_ROOT / f"{request.node.name}.html"
    profiler.write_html(results_file)

item = pn.pane.Markdown("Hello World")

def method1():
    return {k: v for k, v in item.param.values().items()}

def method2():
    return {k: getattr(item,k) for k in item.param}

def test_performance():
    N = 10000
    for i in range(N):
        method1()

def test_performance2():
    N = 10000
    for i in range(N):
        method2()

def test_same_result():
    assert method1()==method2()
```

run

```bash
pytest script_test.py
```

And compare the two pyinstrument reports

As you can see there is a 75% speed up by using simple iteration.

[...]

PR #8547, opened 2026-04-02
Optimizations

philippjfr:
Various optimizations to help improve general speed:

- Freeze bokeh document while adding models to the document
- Avoid the `.param.values` API where possible (implements https://github.com/holoviz/panel/issues/5849)
- Add an `lru_cache` to the `relative_to` utility to speed up path handling 
- Disable bokeh validation unless in `--dev` (autoreload) mode

### DIFF
--- a/panel/config.py
+++ b/panel/config.py
@@ -22,7 +22,6 @@ from weakref import WeakKeyDictionary
 
 import param
 
-from bokeh.settings import settings as _bk_settings
 from pyviz_comms import (
     JupyterCommManager as _JupyterCommManager, extension as _pyviz_extension,
 )
@@ -372,12 +371,6 @@ class _config(_base_config):
                 setattr(self, p+'_', None)
         if self.log_level:
             panel_log_handler.setLevel(self.log_level)
-
-    @param.depends('autoreload', watch=True, on_init=True)
-    def _set_bokeh_validation(self):
-        _bk_settings.perform_document_validation.set_value(self.autoreload)
-        if hasattr(_bk_settings, 'perform_error_diagnostics'):
-            _bk_settings.perform_error_diagnostics.set_value(self.autoreload)
 
     @param.depends('_nthreads', watch=True, on_init=True)
     def _set_thread_pool(self):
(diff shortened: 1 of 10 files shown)


panel/layout/tabs.py
```diff
@@ -202,9 +202,13 @@ def _get_objects(self, model, old_objects, doc, root, comm=None):
...
             elif hidden:
-                child = BkSpacer(**{k: v for k, v in pane.param.values().items()
-                                    if k in Layoutable.param and v is not None and
-                                    k not in ('name', 'design')})
+                params = {}
+                for k in pane.param:
+                    if k not in Layoutable.param or k in ('name', 'design'):
+                        continue
+                    v = getattr(pane, k)
+                    params[k] = v
```
philippjfr:
```suggestion
                    if (v := getattr(pane, k)) is not None:
                        params[k] = v
```

panel/custom.py
```diff
@@ -139,7 +139,7 @@ def _create__view(self):
             view = ParamMethod(self.__panel__, lazy=True)
         else:
             view = panel(self.__panel__())
-        params = view.param.values()
+        params = {p: getattr(self, p) for p in view.param}
```
```suggestion
        params = {p: getattr(view, p) for p in view.param}
```

panel/config.py
```diff
@@ -372,6 +373,12 @@ def __init__(self, **params):
         if self.log_level:
             panel_log_handler.setLevel(self.log_level)
 
+    @param.depends('autoreload', watch=True, on_init=True)
+    def _set_bokeh_validation(self):
+        _bk_settings.perform_document_validation.set_value(self.autoreload)
+        if hasattr(_bk_settings, 'perform_error_diagnostics'):
+            _bk_settings.perform_error_diagnostics.set_value(self.autoreload)
```
hoxbro: Two things:

1. Should we ensure we don't overwrite a user set config?
2. I would do a version check for `perform_error_diagnostics`. Even though it is not released yet, I hope to see it in Bokeh 3.9.1, cc @mattpap.

philippjfr: I think I'm going to revert this for now.

PR #8566, opened 2026-04-22
Sanitize javscript string replacement tokens in python code

### LINKED ISSUE #4356: Panel convert edge case bug when have ${..} in f-string
I'm on Panel 0.14.2. I was working on contributing #4350 and wanted to share my app via `panel convert`. But the app did not work.

After some root cause analysis I could see the issue is the f-string `f"Portfolio Total ${portfolio_total}"`. 

Javascript believes this is a template expression and cannot find the `portfolio_total` js variable.

As I see it we can do one of three things

- Document this edge case
- Validate the `code` and raise an error if `${...}` is in it.
- Somehow find out how to avoid this issue.

## Minimum Reproducible Example

```python
import panel as pn

portfolio_total="100.27"
text = f"Portfolio Total ${portfolio_total}"
pn.panel(text).servable()
```

```bash
panel convert script.py --to pyodide-worker --out pyodide
python3 -m http.server
```

image

image

chucklay:
## Description

Fixes #4356

## How Has This Been Tested?

This has been something I've had to manually fix in the output files in the past. There is a chance it could break conversions where users intentionally have JS-style format strings in python code, but this also adds a flag to disable sanitizing if needed.

## Checklist

- [x] Added documentation


doc/how_to/wasm/convert.md
```diff
@@ -27,6 +27,8 @@ The ``panel convert`` command has the following options:
     --watch               Watch the files
     --num-procs NUM_PROCS
                           The number of processes to start in parallel to convert the apps.
+    --preserve-format-strings
+                          Whether to preserve javiscript style format strings. If not passed, the ${ pattern will be replaced by the escaped equivalent, &#36;{.
```
philippjfr:
```suggestion
                          Whether to preserve javascript style format strings. If not passed, the ${ pattern will be replaced by the escaped equivalent, &#36;{.
```

philippjfr:
Thanks @chucklay, this looks good. I guess my main question is, should this be the default and do you think it's harmful to simply always keep it enabled?

chucklay:
@philippjfr I would advocate for it being the default, we run in to it frequently with tools that take format strings to display currency. As-is, a `panel convert` of any file with a `${` in it results in the convert reporting a success, but outputting a broken JS file, and the resulting error message isn't super helpful in tracking down the problem.

I don't want to say for certain that it would be harmless always clean things, but I really struggle to think of a valid use case. In theory someone could be using JS format strings intentionally to grab/inject things from outside of the python code when this is actually running, but I don't think that would ever be supported/intended behavior.

philippjfr:
I'm +1 on making it a default then. If it results in broken files anyway I don't think we should even offer the option of disabling it. The only other comment I have then is that it'd be good to figure out if there's somewhere earlier we could apply the escape.

panel/io/convert.py
```diff
@@ -724,7 +734,8 @@ def convert_apps(
         'inline': inline,
         'verbose': verbose,
         'compiled': compiled,
-        'local_prefix': local_prefix
+        'local_prefix': local_prefix,
+        'preserve_format_strings': preserve_format_strings,
```
philippjfr:
```suggestion
```

panel/io/convert.py
```diff
@@ -691,6 +699,8 @@ def convert_apps(
         Whether to inline resources.
     compiled: bool
         Whether to use the compiled and faster version of Pyodide.
+    preserve_format_strings: bool
+        Whether to preserve javascript-style format strings within python code
```
philippjfr:
```suggestion
```

panel/io/convert.py
```diff
@@ -647,6 +654,7 @@ def convert_apps(
     inline: bool = False,
     compiled: bool = False,
     verbose: bool = True,
+    preserve_format_strings: bool = False,
```
philippjfr:
```suggestion
```

panel/io/convert.py
```diff
@@ -505,6 +511,7 @@ def convert_app(
...
```
philippjfr:
```suggestion
```

panel/io/convert.py
```diff
@@ -376,6 +379,9 @@ def script_to_html(
 
     # Execution
     post_code = POST_PYSCRIPT if runtime == 'pyscript' else POST
+    if not preserve_format_strings:
+        # Escape javascript-style format strings.
+        source = source.replace('${', '&#36;{')
```
philippjfr:
```suggestion
    # Escape javascript-style format strings.
    source = source.replace('${', '&#36;{')
```

panel/io/convert.py
```diff
@@ -350,6 +351,8 @@ def script_to_html(
         Whether to inline resources.
     compiled: bool
         Whether to use pre-compiled pyodide bundles.
+    preserve_format_strings: bool
+        If '${' patterns should be preserved.
```
philippjfr:
```suggestion
```

panel/io/convert.py
```diff
@@ -320,7 +320,8 @@ def script_to_html(
     local_prefix: str = LOCAL_PREFIX,
     manifest: str | None = None,
     inline: bool = False,
-    compiled: bool = True
+    compiled: bool = True,
+    preserve_format_strings: bool = False,
```
philippjfr:
```suggestion
    compiled: bool = True
```

panel/command/convert.py
```diff
@@ -95,6 +95,11 @@ class Convert(Subcommand):
             default = 1,
             help    = "The number of processes to start in parallel to convert the apps."
         )),
+        ('--preserve-format-strings', Argument(
+            default = False,
+            action  = 'store_true',
+            help    = "Whether to preserve javascript-style format strings within python code."
+        ))
```
philippjfr:
```suggestion
```

panel/command/convert.py
```diff
@@ -149,7 +154,8 @@ def invoke(self, args: argparse.Namespace) -> None:
                     max_workers=args.num_procs,
                     http_patch=not args.disable_http_patch,
                     compiled=args.compiled,
-                    verbose=True
+                    verbose=True,
+                    preserve_format_strings=args.preserve_format_strings,
```
philippjfr:
```suggestion
                    verbose=True
```

doc/how_to/wasm/convert.md
```diff
@@ -27,6 +27,8 @@ The ``panel convert`` command has the following options:
     --watch               Watch the files
     --num-procs NUM_PROCS
                           The number of processes to start in parallel to convert the apps.
+    --preserve-format-strings
+                          Whether to preserve javascript style format strings. If not passed, the ${ pattern will be replaced by the escaped equivalent, &#36;{.
```
philippjfr:
```suggestion
```

PR #7945, opened 2025-05-26
test: Fix flaky UI test

hoxbro:
Have downloaded all the logs from the last month to see which tests are flaky, the following results are only from the main branch.

Have collected the results here: [result.json](https://github.com/user-attachments/files/20548117/result.json), but some light processing show these results:

| failures                                                                                                  |   count |
|:----------------------------------------------------------------------------------------------------------|--------:|
| panel/tests/ui/io/test_jupyterlite.py::test_jupyterlite_execution                                         |      10 |
| panel/tests/ui/widgets/test_tabulator.py::test_tabulator_patch_no_horizontal_rescroll                     |       7 |
| panel/tests/ui/io/test_convert.py::test_pyodide_test_error_handling_worker                                |       4 |
| panel/tests/ui/layout/test_column.py::test_column_scroll_position_init                                    |       4 |
| panel/tests/ui/layout/test_feed.py::test_feed_reset_visible_range                                         |       4 |
| panel/tests/ui/layout/test_feed.py::test_feed_view_latest                                                 |       4 |
| panel/tests/ui/widgets/test_texteditor.py::test_texteditor_regression_click_toolbar_cursor_stays_in_place |       3 |
| panel/tests/ui/widgets/test_indicators.py::test_tooltip_text_updates                                      |       3 |
| panel/tests/ui/widgets/test_indicators.py::test_plaintext_tooltip[str]                                    |       3 |
| panel/tests/ui/widgets/test_tabulator.py::test_tabulator_sorter_default_number                            |       3 |
| panel/tests/ui/layout/test_feed.py::test_feed_scroll_to_latest_within_limit                               |       3 |
| panel/tests/ui/widgets/test_tabulator.py::test_tabulator_editors_panel_date                               |       3 |
| panel/tests/ui/io/test_convert.py::test_pyodide_test_convert_button_app[pyodide-worker]                   |       3 |
| panel/tests/ui/widgets/test_indicators.py::test_plaintext_tooltip[Tooltip]                                |       3 |
| panel/tests/ui/io/test_convert.py::test_pyodide_test_convert_button_app[pyscript]                         |       2 |
| panel/tests/ui/widgets/test_tabulator.py::test_tabulator_frozen_rows                                      |       2 |
| panel/tests/ui/widgets/test_time_picker.py::test_time_picker                                              |       2 |
| panel/tests/ui/layout/test_feed.py::test_feed_scroll_to_latest_disabled_when_limit_zero                   |       2 |
[...]

PR #8572, opened 2026-04-30
Fix flaky Gauge and Tabulator UI tests

ghostiee-11:
## Description

Two intermittent UI test failures on `ui:test-ui:macos-latest` and `ui:test-ui:ubuntu-latest` propagate failures to every open PR. Fixing both with the established readiness-wait patterns from this codebase.

### 1. Gauge tests (#8515)

`test_gauge_renders`, `test_gauge_does_not_crash_other_widgets`, and `test_gauge_value_update` flake intermittently with:

```
AssertionError: Locator expected to have count '1'
```

`serve_component` only waits for the Panel WebSocket and `networkidle`, neither guarantees the ECharts JS module has finished loading. Gauge renders through ECharts, so the canvas only appears after the module fetch + parse + setOption. On slow CI runners this chain can take longer than 10 seconds.

**Fix:** added a small `_wait_for_echarts(page)` helper that polls for `window.echarts` to be defined (timeout 30s), called before each canvas-count assertion.

### 2. Tabulator test (#7945-class flake)

`test_tabulator_frozen_columns_with_positions` flakes with:

```
playwright._impl._errors.Error: Locator.scroll_into_view_if_needed:
Element is not attached to the DOM
```

The dict-based `frozen_columns={"float": "left", "int": "right"}` path re-orders columns and re-renders rows. Between Playwright matching `text=\"2019-01-01 10:00:00\"` and the scroll firing, the row can detach.

**Fix:** same pattern as commit `7411f182` (which fixed `test_tabulator_patch_no_horizontal_rescroll` for the same flake class):
- `expect(cell).to_be_attached()` before `scroll_into_view_if_needed()` to eliminate the match-then-detach race
- `@pytest.mark.flaky(reruns=3, reruns_delays=2)` as defense in depth

## How Has This Been Tested?

Both changes are test-only and follow established readiness-wait patterns already used in this codebase. CI on this PR is the authoritative verifier. Latest run on the previous push showed only the tabulator test still failing, which this commit addresses; the gauge tests now pass.

### DIFF
--- a/panel/tests/ui/widgets/test_indicators.py
+++ b/panel/tests/ui/widgets/test_indicators.py
@@ -85,12 +85,23 @@ def test_tooltip_text_updates(page):
     expect(tooltip).to_have_text("Updated")
 
 
+def _wait_for_echarts(page, timeout=30000):
+    """Wait for the ECharts JS module to be loaded on the page.
+
+    Gauge renders via ECharts. serve_component only waits for the Panel
+    websocket and networkidle, not for the ECharts module fetch. On slow
+    CI runners the canvas can appear well after networkidle, so polling
+    for `window.echarts` is the reliable readiness signal.
+    """
+    page.wait_for_function("typeof window.echarts !== 'undefined'", timeout=timeout)
+
+
 def test_gauge_renders(page):
     gauge = Gauge(name="Speed", value=75, bounds=(0, 200))
 
     serve_component(page, gauge)
 
-    # Wait for ECharts to load and render a canvas element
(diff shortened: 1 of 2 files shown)


ghostiee-11:
Latest CI: macOS and Ubuntu UI both green, gauge and tabulatorfixes verified. Windows UI failures are pre-existing flakes in test_convert.py (pyodide WASM) and test_jupyter_server_extension.py, not regressions from this PR. Same pattern fails on main.

philippjfr (approved):

PR #8586, opened 2026-05-13
Drop support for bokeh 3.7

philippjfr:
The last bokeh 3.7.x release was over a year ago and we should generally keep users within a 2 minor release window.

### DIFF
--- a/panel/util/__init__.py
+++ b/panel/util/__init__.py
@@ -47,7 +47,6 @@ from .parameters import (  # noqa
 log = logging.getLogger('panel.util')
 
 bokeh_version = Version(Version(bokeh.__version__).base_version)
-BOKEH_GE_3_6 = bokeh_version >= Version('3.6')
 BOKEH_GE_3_8 = bokeh_version >= Version('3.8')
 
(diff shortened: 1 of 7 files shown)


hoxbro (approved):

hoxbro:
Lets move https://raw.githubusercontent.com/holoviz/panel/main/doc/_static/logo.png etc. to an S3 bucket. No need to be done in this PR.

panel/util/__init__.py
```diff
@@ -47,7 +47,6 @@
 log = logging.getLogger('panel.util')
 
 bokeh_version = Version(Version(bokeh.__version__).base_version)
-BOKEH_GE_3_6 = bokeh_version >= Version('3.6')
 BOKEH_GE_3_8 = bokeh_version >= Version('3.8')
```
hoxbro: This is also no longer needed

hoxbro: Pushed in [3c1f4d3](https://github.com/holoviz/panel/pull/8586/commits/3c1f4d374faa5f4b22c0cb23f98add6f1d52db7d)

PR #8588, opened 2026-05-14
Add support for wildcard route handling

philippjfr:
## Description

This PR adds first-class dynamic route support across both server implementations and makes route context available inside apps in a backend-consistent way.

Users can now serve apps on dynamic routes and reliably read:

- `pn.state.route_params` for captured path variables
- `pn.state.app_url` as the concrete matched app path for the current session (not new)

It also improves route matching behavior for wildcard routes, ensures websocket/session flows preserve route context, and updates docs to present path-template syntax as the primary recommended style.

### Supported route syntax

- **Path-template syntax (recommended, cross-backend):**
  - `/user/{name}`
  - `/files/{filepath:path}`

- **Regex syntax (Tornado):**
  - `/user/([^/]+)`
  - `/org/(?P<org>[^/]+)/user/(?P<user>[^/]+)`

For Tornado, path templates are normalized to Tornado-compatible patterns.  
For FastAPI, path templates are passed through naturally and now preserve route metadata in session state.

### Behavior improvements included

- Wildcard/dynamic routes no longer interfere with core app endpoints (`/ws`, `/metadata`, `/autoload.js`).
- Root index no longer auto-redirects to an invalid wildcard-only route.
- Session creation and websocket-first flows both carry route metadata so app state is consistent regardless of connection path.
- Server docs now include a dedicated wildcard routes how-to and backend compatibility guidance.

#### Tornado
```python
import panel as pn
pn.extension()

def app():
    return {
        "route_params": pn.state.route_params,
        "app_url": pn.state.app_url,
    }

pn.serve({"/user/{name}": app}, show=False)
# /user/alice -> {"route_params": {"name": "alice"}, "app_url": "/user/alice"}
```

#### FastAPI
```python
import panel as pn
from panel.io.fastapi import serve
pn.extension()

def app():
    return {
        "route_params": pn.state.route_params,
        "app_url": pn.state.app_url,
    }

serve({"/files/{filepath:path}": app}, show=False)
# /files/a/b/c.txt -> {"route_params": {"filepath": "a/b/c.txt"}, "app_url": "/files/a/b/c.txt"}
```

## AI Disclosure

- Tool & Model: Cursor + Codex 5.3 / GPT 5.5
- Usage: Used to investigate routing internals, implement dynamic route support across Tornado/FastAPI, add tests, and update documentation.

## Checklist

- [x] Added documentation

### DIFF
--- a/doc/how_to/server/wildcard_routes.md
+++ b/doc/how_to/server/wildcard_routes.md
@@ -1,0 +1,115 @@
+# Serving applications on wildcard routes
+
+When serving multiple apps with `pn.serve`, the route keys are usually static paths (for example `"/app"`). You can also use wildcard-style route patterns to match dynamic URL segments.
+
+## Serve an app on a wildcard route
+
+Use a route pattern as the dictionary key:
+
+```python
+import panel as pn
+
+pn.extension()
+
+def app():
+    return pn.pane.JSON(
+        {
+            "route_params": pn.state.route_params,
+            "app_url": pn.state.app_url,
+        },
+        depth=2,
+    )
+
+pn.serve(
+    {
+        "/user/{name}": app,
+    },
+    show=False,
+)
+```
+
+Now:
+
+- `/user/alice` and `/user/bob` will both resolve to the same app
+- `pn.state.route_params` will contain the captured values for the current request
+
+For the two URLs above, `pn.state.route_params` will be:
+
+- `{"name": "alice"}`
+- `{"name": "bob"}`
+
+Path converters are supported as well:
+
+```python
+pn.serve(
+    {
+        "/files/{filepath:path}": app,
+    },
+    show=False,
+)
+```
+
+Then `pn.state.route_params` contains:
+
+- `{"filepath": "a/b/c.txt"}`
+
+Template converters currently supported are:
+
+- `str` (default) for single path segments
+- `path` for slash-containing paths
+- `int` for signed integers (for example `-2`, `0`, `+3`)
+- `float` for signed decimal/scientific formats (for example `-1.5`, `.5`, `1.`, `6.02e23`)
+- `uuid` for canonical UUID strings
+
+## Regex syntax (Tornado-style)
+
+On Tornado you can also use regular expression routes directly:
+
+```python
+pn.serve(
+    {
+        "/user/([^/]+)": app,
+    },
+    show=False,
+)
+```
+
+For this route, `pn.state.route_params` contains positional captures:
+
+- `{"0": "alice"}`
+
+Unlike path-template routes (which use parameter names), raw regex captures are exposed as positional string keys (`"0"`, `"1"`, ...).
+
+### Named parameters with regex syntax
+
+You can also use named groups:
+
+```python
+pn.serve(
+    {
+        "/org/(?P<org>[^/]+)/user/(?P<user>[^/]+)": app,
+    },
+    show=False,
+)
+```
+
+Then `pn.state.route_params` contains the named captures:
+
+- `{"org": "acme", "user": "alice"}`
+
+## Backend compatibility
+
+- **Tornado (`panel serve`, `panel.io.server.serve`)**
+  - Supports regex-style routes, e.g. `"/user/([^/]+)"` and named groups.
+  - Supports path-template syntax, which is normalized to Tornado-compatible patterns (for example `"/user/{name}"`).
+- **FastAPI (`panel.io.fastapi.serve`)**
+  - Supports path-template syntax, e.g. `"/user/{name}"`, `"/files/{filepath:path}"`.
+  - Route parameters are available on `pn.state.route_params`.
+  - Concrete matched paths are reflected in `pn.state.app_url`.
+
+## Best practices
+
+- Prefer path-template syntax (`{name}`, `{name:path}`) when writing apps intended to run on both Tornado and FastAPI.
+- If you use regex syntax, prefer non-greedy segment patterns such as `([^/]+)` for path segments.
+- Prefer explicit groups over very broad patterns like `(.*)` unless you specifically need to capture slashes.
+- Use `pn.state.app_url` when you need the concrete matched application URL for the current session.
--- a/doc/how_to/server/wildcard_routes.md
+++ b/doc/how_to/server/wildcard_routes.md
@@ -104,8 +104,6 @@ Then `pn.state.route_params` contains the named captures:
   - Supports path-template syntax, which is normalized to Tornado-compatible patterns (for example `"/user/{name}"`).
 - **FastAPI (`panel.io.fastapi.serve`)**
   - Supports path-template syntax, e.g. `"/user/{name}"`, `"/files/{filepath:path}"`.
-  - Route parameters are available on `pn.state.route_params`.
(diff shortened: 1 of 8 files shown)


panel/io/fastapi.py
```diff
@@ -44,6 +57,159 @@
...
+        payload = dict(payload)
+        if route_params:
+            payload['route_params'] = route_params
+        if app_path:
+            payload['app_path'] = app_path
+        expiry = int(payload.pop('session_expiry', 0))
+        expires_in = max(
+            1,
+            expiry - calendar.timegm(dt.datetime.now(tz=dt.timezone.utc).timetuple())
+        ) if expiry else 300
```
hoxbro:
```suggestion
        expires_in = max(1, expiry - int(time.time())) if expiry else 300
```

Nit, but easier.

panel/io/server.py
```diff
@@ -100,6 +108,16 @@ class TokenPayload(t.TypedDict):
 
 INDEX_HTML = os.path.join(os.path.dirname(__file__), '..', '_templates', "index.html")
 DEFAULT_TITLE = "Panel Application"
+_PATH_TEMPLATE_PATTERN = re.compile(
+    r'^\{(?P<name>[A-Za-z_][A-Za-z0-9_]*)(?::(?P<converter>str|path|int|float|uuid))?\}$'
+)
+_PATH_TEMPLATE_CONVERTERS = {
+    'str': r'[^/]+',
+    'path': r'.*',
```
hoxbro, line 118: Is this too broad?

panel/io/server.py
```diff
@@ -374,6 +420,67 @@ def get_login_url(self):
...
+
+    @property
+    def app_path(self) -> str | None:
+        return self._app_path
+
+    def __getattr__(self, name: str) -> t.Any:
+        if not name.startswith("_") and isinstance(self._request, dict):
+            if name in self._request:
+                return self._request[name]
+            raise AttributeError(name)
```
hoxbro: Just want to be sure we don't want it to fallback here?

panel/io/server.py
```diff
@@ -374,6 +420,67 @@ def get_login_url(self):
...
+    if not app_path:
+        app_path = '/'
+    if not app_path.startswith('/'):
+        app_path = '/' + app_path
+    return app_path
+
+
+def _route_params(args: tuple[t.Any, ...], kwargs: dict[str, t.Any]) -> dict[str, str]:
+    route_params = {str(i): str(value) for i, value in enumerate(args)}
+    route_params.update({str(key): str(value) for key, value in kwargs.items()})
```
hoxbro: Ok, to overwrite here?

panel/io/server.py
```diff
@@ -587,17 +694,54 @@ async def get(self, *args, **kwargs) -> None:
...
+    @classmethod
+    def _is_concrete_route(cls, route: str) -> bool:
+        return not any(token in route for token in cls._regex_tokens)
+
     @authenticated
     async def get(self, *args, **kwargs):
-        if self.use_redirect and len(self.applications) == 1:
+        if (
+            self.use_redirect and len(self.applications) == 1 and
+            self._is_concrete_route(next(iter(self.applications)))
```
hoxbro, line 801:
```suggestion
            self._is_concrete_route(self.applications[0])
```

If you can take length of it we should also be possible to take the first item.

doc/how_to/server/wildcard_routes.md
```diff
@@ -0,0 +1,115 @@
...
+
+## Backend compatibility
+
+- **Tornado (`panel serve`, `panel.io.server.serve`)**
+  - Supports regex-style routes, e.g. `"/user/([^/]+)"` and named groups.
+  - Supports path-template syntax, which is normalized to Tornado-compatible patterns (for example `"/user/{name}"`).
+- **FastAPI (`panel.io.fastapi.serve`)**
+  - Supports path-template syntax, e.g. `"/user/{name}"`, `"/files/{filepath:path}"`.
+  - Route parameters are available on `pn.state.route_params`.
+  - Concrete matched paths are reflected in `pn.state.app_url`.
```
hoxbro: Is this only available for FastAPI backend?

panel/io/application.py
```diff
@@ -130,17 +130,34 @@ def _set_session_prefix(self, doc):
...
+                app_path = session_context.token_payload.get('app_path', app_path)
+            except AssertionError:
+                pass
+        if not app_path.startswith('/'):
+            app_path = f'/{app_path}'
+
+        request_path = getattr(request, 'path', None) or app_path
+        prefix = ''
+        if app_path != '/' and request_path.endswith(app_path):
+            prefix = request_path[:-len(app_path)]
```
hoxbro: Should this be using a stricter matching?

panel/io/fastapi.py
```diff
@@ -44,6 +57,159 @@
...
+    tornado_request.app_path = app_path
+    simple_cookies = SimpleCookie()
+    for name, cookie in request.cookies.items():
+        cookie_value = cookie.value if hasattr(cookie, "value") else str(cookie)
+        simple_cookies[name] = cookie_value
+    tornado_request._cookies = simple_cookies
+
+    headers = dict(tornado_request.headers)
+    cookies = {
+        name: cookie.value if hasattr(cookie, "value") else str(cookie)
```
hoxbro, line 124:
```suggestion
        name: getattr(cookie, "value", str(cookie)) 
```

panel/io/fastapi.py
```diff
@@ -44,6 +57,159 @@
...
+            method=request.method,
+            uri=uri,
+            headers=HTTPHeaders(request.headers),
+        )
+    tornado_request.route_params = route_params
+    tornado_request.app_path = app_path
+    simple_cookies = SimpleCookie()
+    for name, cookie in request.cookies.items():
+        cookie_value = cookie.value if hasattr(cookie, "value") else str(cookie)
+        simple_cookies[name] = cookie_value
```
hoxbro:
```suggestion
    simple_cookies = SimpleCookie({
        name: getattr(cookie, "value", str(cookie))
        for name, cookie in request.cookies.items()
    })
```

I think this should also work.

hoxbro (approved):
Only reviewed the code, not tested the functionality.

panel/io/server.py
philippjfr: Yes, that's how it works, path params are passed to the handler as args and kwargs and then we create our path_params dictionary from those.

panel/io/server.py
philippjfr: It's a dictionary afaik.

doc/how_to/server/wildcard_routes.md
philippjfr: No, for both, will clarify.

panel/io/fastapi.py
philippjfr: Meh, will leave it as is.

panel/io/server.py
philippjfr: It's broad but correct I think.

panel/io/fastapi.py
philippjfr: Will update.

panel/io/server.py
philippjfr: Good catch, I think we do.

PR #8592, opened 2026-05-16
Add links to related API docs and include pn.rx in Functions vs. Classes doc

Coderambling:
First of all, great document @philippjfr ! Ties a lot of high level concepts together.

Updated the doc to include links to related API documents and added mention of pn.rx .

Not sure if pn.rx is mentioned in the right place in the updated doc, but it should definitely be included, for completeness, and to clarify where it sits in / relative to functions and classes.

Also, there seems to be some inconsistency / lack of clarity across (API explanation) documents.

It feels like concepts and terms like API, approach, reactive and declarative are sometimes used with different scope / intent / meaning across docs.

Makes it easy for new users to get things muddled, which contributes to the "beginner's hump" that users often experience.

Will try to raise a separate issue for this.

### DIFF
--- a/doc/explanation/api/functions_vs_classes.md
+++ b/doc/explanation/api/functions_vs_classes.md
@@ -1,6 +1,6 @@
 ## Classes vs functions in Panel: understanding the tradeoff
 
-In Panel, there are two main ways to structure reactive code: binding functions with `pn.bind` (the reactive API), or building classes that extend `param.Parameterized` or `pn.viewable.Viewer` (the declarative API). Both are valid, and the right choice depends on what you're building.
+In Panel, there are two main ways to structure reactive code: binding functions with `pn.bind` and `pn.rx`(the [reactive API](../../explanation/api/reactive.md)), or building classes that extend `param.Parameterized` or `pn.viewable.Viewer` (the [declarative API](../../explanation/api/parameterized.md)). Both approacheas are valid, and the right choice depends on what you're building.
 
 ### Why functions feel natural at first
 
@@ -16,7 +16,7 @@ Consider a dashboard that filters a dataset by year range and manufacturer. With
 
 ### What classes give you: a single source of truth
 
-The class-based approach solves this by giving state a home. Instead of free-floating widget variables, you create a dedicated object whose job is to hold the current selection as `param.Parameter` attributes. Any part of the app that needs to know what the user has selected asks that object directly.
+The class-based approach solves this by utlizing the declarative API (of Panel/param(?)), giving state a home. Instead of free-floating widget variables, you create a dedicated object whose job is to hold the current selection as `param.Parameter` attributes. Any part of the app that needs to know what the user has selected asks that object directly.
 
 Other objects can declare, using `@param.depends`, that they should recompute when those parameters change. The logic for what to do when state changes lives in one place, not spread across multiple `pn.bind` calls. Any number of views can all reference the same state-holding object, and when something changes they update automatically — without any view needing to know anything about the widgets that caused the change.
 
--- a/doc/explanation/api/functions_vs_classes.md
+++ b/doc/explanation/api/functions_vs_classes.md
@@ -16,7 +16,7 @@ Consider a dashboard that filters a dataset by year range and manufacturer. With
 
 ### What classes give you: a single source of truth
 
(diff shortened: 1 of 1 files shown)


doc/explanation/api/functions_vs_classes.md
```diff
@@ -16,7 +16,7 @@ Consider a dashboard that filters a dataset by year range and manufacturer. With
 
 ### What classes give you: a single source of truth
 
-The class-based approach solves this by giving state a home. Instead of free-floating widget variables, you create a dedicated object whose job is to hold the current selection as `param.Parameter` attributes. Any part of the app that needs to know what the user has selected asks that object directly.
+The class-based approach solves this by utilizing the declarative API (of Panel/param(?)), giving state a home. Instead of free-floating widget variables, you create a dedicated object whose job is to hold the current selection as `param.Parameter` attributes. Any part of the app that needs to know what the user has selected asks that object directly.
```
philippjfr:
```suggestion
The class-based approach solves this by declaratively expressing state as parameters, giving state a home. Instead of free-floating widget variables, you create a dedicated object whose job is to hold the current selection as `param.Parameter` attributes. Any part of the app that needs to know what the user has selected asks that object directly.
```

doc/explanation/api/functions_vs_classes.md
```diff
@@ -16,7 +16,7 @@ Consider a dashboard that filters a dataset by year range and manufacturer. With
 
 ### What classes give you: a single source of truth
 
-The class-based approach solves this by giving state a home. Instead of free-floating widget variables, you create a dedicated object whose job is to hold the current selection as `param.Parameter` attributes. Any part of the app that needs to know what the user has selected asks that object directly.
+The class-based approach solves this by declaratively expressing state as parameters, giving state a home. Instead of free-floating widget variables, you create a dedicated object whose job is to hold the current selection as `param.Parameter` attributes. Any part of the app that needs to know what the user has selected asks that object directly.
 
 Other objects can declare, using `@param.depends`, that they should recompute when those parameters change. The logic for what to do when state changes lives in one place, not spread across multiple `pn.bind` calls. Any number of views can all reference the same state-holding object, and when something changes they update automatically — without any view needing to know anything about the widgets that caused the change.
```
Coderambling, line 21:
```suggestion
Other objects can declare, using [`@param.depends`(https://panel.holoviz.org/how_to/param/dependencies.html), that they should recompute when the parameters of a state-holding object change. 

_what_ should happen is defined by the functional logic of a particular (view) object.
_when_ they happen is determined by state changes that live in one place (the state-holding object), not spread across multiple `pn.bind` calls.

Any number of views can now reference the same state-holding object, and when something changes, they all update automatically according to their own logic— without any view needing to know anything about the widgets that caused the change.

The result is a clean separation of concerns.

Note:
`@param.depends` should not be confused (?) with [`@pn.depends`](https://panel.holoviz.org/explanation/api/reactive.html#example), which is a shorthand for pn.bind .
```

The above is probably not all correct @philippjfr , but I hope it also serves as an illustration of how confusing things could get for users that are new to panel. This was also the motivation to make the above text slightly more verbose and explicit.

Also concerned about users confusing `@param.depends` and `@pn.depends`. Currently, `@pn.depends` is not mentioned yet in the document. Maybe it should be included in the pn.bind section?

And when reading their respective documentation pages, it is quite hard to understand the nuances of when to use which.

Obviously totally fine with (parts of) these proposed changes happening in a later release.

doc/explanation/api/functions_vs_classes.md
Coderambling, line 27:
```suggestion
One of the main practical gains is easy testability. Because state lives in `param.Parameter` attributes, rather than widget values, you can test data transformation logic by constructing an object, setting its parameters directly, and asserting on the output in a [testing](https://panel.holoviz.org/developer_guide/index.html#testing) script or Github Actions, without the need for widgets,  layout, or a browser. 

You can change how something is displayed without touching any of the underlying computation. With a pure function approach, this separation is much harder to maintain, because the layout code names the widgets and maps them to specific function arguments, so changes to either tend to ripple across both.
```

Coderambling (commented):
Got some inspiration to to look through the document again @philippjfr . Totally fine if this elements of these proposed changes are considered in a later release.

PR #8628, opened 2026-06-02
docs: Build llms.txt and markdown docs during build

philippjfr:
Initial tests


ahuang11:
Will update skills to point to these after this is merged https://holoviz-dev.github.io/holoviz-skills/developing-with-holoviz/panel/#component-reference

PR #8629, opened 2026-06-03
type: Fix type errors

hoxbro:
## Description

Fixes failing types seen in https://github.com/holoviz/panel/actions/runs/26874892906/job/79259737195 and also enable `param.mypy_plugin`

image


panel/io/admin.py
```diff
@@ -43,7 +43,7 @@
 
 PROCESSES: dict[int, Process] = {}
 
-log_sessions = []
+log_sessions: list = []
```
hoxbro, line 46: Not sure why this is needed?

philippjfr (approved):

PR #8633, opened 2026-06-04
Ensure ready parameter and next button are in sync after previous button

### LINKED ISSUE #8631: Ready parameter set to `True` but next button stays disabled in pipeline
<details>

```plaintext
panel 1.9.3
```
</details>

I thought https://github.com/holoviz/panel/pull/8443 would fix this but it's not the case.

In one of my apps, there's a pipeline stage where I reset the content and the ready parameter whenever the stage is entered. I do this because if the stage has already been visited, I need the user to do something to compute the outputs from the new inputs and not be able to proceed to the next stage with the old outputs.

However, in this situation if I get to this stage from the next by clicking the _previous_ button, the _next_ button stays disabled when the ready parameter is set to `True` once. You have somehow to set it to `True` a second time for the _next_ button to be enabled.

```python
import panel as pn
import param

pipeline = pn.pipeline.Pipeline()


class Stage1(param.Parameterized):
    ready = param.Boolean(default=False)

    def __init__(self):
        super().__init__()
        self.ready_button = pn.widgets.Button(name="Ready", on_click=self._ready)

    def _ready(self, _):
        self.ready = True

    def panel(self):
        self.ready = False
        return pn.Column(self.ready_button)


    def panel(self):
        return "Stage 2"


pipeline.add_stage("Stage 1", Stage1, ready_parameter="ready")
pipeline.add_stage("Stage 2", Stage2)

pipeline.servable()
```

https://github.com/user-attachments/assets/dd82e945-ddb9-47dc-8ef1-3080bf1a86bc

TheoMathurin:
## Description
Fixes #8631

When coming back to a stage, the `_block` attribute set to `True` meant that on the first time the `ready` parameter was set to `True`, the _next_ button stayed disabled and, for stages with `auto_advance`, the pipeline did not proceed.

In this minimal example, you can try various combinations of `auto_advance` and automatic reset of the `ready` parameter in the `panel` method. All are included in the now expanded test introduced by #8443.
```python
import panel as pn
import param

pipeline = pn.pipeline.Pipeline()


class Stage1(param.Parameterized):
    ready = param.Boolean(default=False)

    def __init__(self):
        super().__init__()
        self.ready_button = pn.widgets.Button(name="Ready", on_click=self._ready)

    def _ready(self, _):
        self.ready = True

    def panel(self):
        # self.ready = False
        return pn.Column(self.ready_button)


    def panel(self):
        return "Stage 2"


auto_advance = True
pipeline.add_stage("Stage 1", Stage1, ready_parameter="ready", auto_advance=auto_advance)
pipeline.add_stage("Stage 2", Stage2)

pipeline.servable()
```

NB: there are small but significant changes in this new implementation, including that there is no need for a `_block` attribute anymore.

## Checklist

### DIFF
--- a/panel/pipeline.py
+++ b/panel/pipeline.py
@@ -180,7 +180,6 @@ class Pipeline(Viewer):
         self._states = {}
         self._state = None
         self._linear = True
-        self._block = False
         self._error = None
         self._graph = {}
         self._route = []
@@ -248,18 +247,12 @@ class Pipeline(Viewer):
         return self._stages[index][0]
 
     def _unblock(self, event):
-        if self._state is not event.obj or self._block:
-            self._block = False
+        self.next_button.disabled = not event.new
+        if not event.new:
             return
 
-        button = self.next_button
-        if button.disabled and event.new:
-            button.disabled = False
-        elif not button.disabled and not event.new:
-            button.disabled = True
-
         stage_kwargs = self._stages[self._stage][-1]
-        if event.new and stage_kwargs.get('auto_advance', self.auto_advance):
+        if stage_kwargs.get('auto_advance', self.auto_advance):
             self._next()
 
     def _select_next(self, event):
@@ -444,7 +437,6 @@ class Pipeline(Viewer):
                 self.stage[0] = self._state.panel()
             else:
                 self.stage[0] = self._init_stage()
-            self._block = True
         except Exception as e:
             self.error[:] = [self._get_error_button(e)]
             self._error = self._stage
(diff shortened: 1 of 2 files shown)


TheoMathurin:
Note that I don't know about the original rationale behind the `_block` attribute. It was only ever set to `True` in the `_previous` method, but I hope I'm not missing something.

Also there's another bit of logic with the `self._state is not event.obj` check, that is now gone and which I could reinstate if needed. I don't know in which situation this could occur.

philippjfr:
Appreciate it! `Pipeline` hasn't seen any love in years. Probably want to either modernize it or deprecate it.

philippjfr:
Type errors unrelated.

philippjfr (approved):

PR #8635, opened 2026-06-05
Forward additional `oauth_extra_params` to the OAuth authorization endpoint

MarcSkovMadsen:
## Summary

`OAuthLoginHandler.get_authenticated_user` only forwarded `audience` (and `scope`) from `config.oauth_extra_params` to the OAuth2/OIDC **authorize** request, silently dropping other standard authorization-request parameters such as `prompt`, `login_hint` and `domain_hint`.

This makes it impossible to, for example, force account selection on Azure Entra ID via `prompt=login` / `prompt=select_account` — a common requirement for users who are signed into multiple Microsoft accounts in the same browser.

## Change

Forward all user-provided `oauth_extra_params` to the authorize request, skipping the keys the provider implementations consume internally to build their URLs/scopes (`tenant`, `subdomain`, `url`, `server`, the generic `*_URL`/`USER_KEY`, and the explicitly handled `audience`/`scope`). These are collected in a new `RESERVED_OAUTH_PARAMS` set.

So a config like:

```bash
panel serve app.py --oauth-provider=azure \
  --oauth-extra-params "{'tenant': '...', 'prompt': 'login'}"
```

now actually forwards `prompt=login` to the authorize URL, while `tenant` continues to be used only to build the provider URL (not leaked as a query param).

## Tests

Adds `test_get_authenticated_user_forwards_extra_oauth_params`, which asserts that `prompt`/`login_hint` are forwarded into the authorize request's `extra_params`, that `audience` is still forwarded, and that the provider-internal `tenant` key is not.

### DIFF
--- a/panel/auth.py
+++ b/panel/auth.py
@@ -36,6 +36,18 @@ log = logging.getLogger(__name__)
 STATE_COOKIE_NAME = 'panel-oauth-state'
 CODE_COOKIE_NAME = 'panel-oauth-code'
 
+# Keys in ``config.oauth_extra_params`` that the provider implementations
+# consume internally (to build provider URLs, scopes, etc.) and which must
+# therefore NOT be forwarded as query parameters to the OAuth2/OIDC
+# authorization endpoint.
+RESERVED_OAUTH_PARAMS = {
+    'scope',  # forwarded as a top-level param via _SCOPE, not in extra_params
+    'tenant',  # Azure AD
+    'subdomain',  # Auth0
+    'url', 'server',  # Okta / GitLab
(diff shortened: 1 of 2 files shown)


MarcSkovMadsen:
We've been using this for a while and now it was time to upstream. We need `--oauth-extra-params "{{'tenant': '{TENANT_ID}', 'prompt': 'login'}}"`` working because many of our users have multiple Azure accounts (normal user + admin user).

The failing type tests are not due to this PR.

philippjfr (approved):

philippjfr:
Thanks @MarcSkovMadsen!

PR #8636, opened 2026-06-05
Set `Cache-Control: no-store` on the Bokeh document page

MarcSkovMadsen:
## Summary

The served document page embeds a short-lived Bokeh **session token**. Because Panel's `DocHandler` sends no cache headers, browsers may cache the HTML page. After an oauth logout/login cycle (or when returning to an app later), the browser can serve the **stale cached page carrying an expired token**, so the subsequent WebSocket upgrade fails with *"Could not open websocket"* and the user is stuck in a disconnect loop.

## Change

Send `Cache-Control: no-store` from `DocHandler.get`, so the document page (and the token embedded in it) is always re-fetched.

This mirrors what Panel already does for the Jupyter server extension handler (`panel/io/jupyter_server_extension.py`), which sets `Cache-Control: no-cache, no-store, must-revalidate`.

## Tests

Adds `test_server_doc_page_is_not_cached`, which serves an app and asserts the response carries `Cache-Control: no-store`.

### DIFF
--- a/panel/io/server.py
+++ b/panel/io/server.py
@@ -635,6 +635,11 @@ class DocHandler(LoginUrlMixin, BkDocHandler):
 
     @authenticated
     async def get(self, *args, **kwargs):
+        # Prevent the browser from caching the document page. The page embeds
+        # a short-lived Bokeh session token; a cached page served after a
+        # logout/login cycle would carry a stale token and the subsequent
+        # WebSocket connection would fail (see e.g. holoviz/panel#8634).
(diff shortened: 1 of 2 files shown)


MarcSkovMadsen:
We have used this for our apps a while with success. And now it was time to upstream.

The type tests are not caused by this PR.

philippjfr (approved):

philippjfr:
Will resolve the type errors separately.

Issue #8634, opened 2026-06-05
Azure OAuth not working with multiple accounts


MarcSkovMadsen:
We have for the last two years been trying to setup Azure Oauth for our Panel apps without success. Many problems have been solved along the way. But the problem remaining is that Azure Oauth does not work when having multiple azure accounts (personal + admin).

The problem is well known and described here https://panel.holoviz.org/how_to/authentication/trouble_shooting.html#cause-multiple-accounts.

You can see the problem in action below where it oauth only works for me if I use incognito mode.

<img width="2552" height="563" alt="Image" src="https://github.com/user-attachments/assets/97ce6b44-4fa6-4f82-a526-0828bb80060c" />

```bash
bokeh.min.js?v=2630124d82b67a204b9c9da925d3d456e9ec585bb0cf8f839a1c704dbee3f200b70d237916a9fd429755022dc83ac334505cd7385ebb4453a7b17aa1d3699c35:249 WebSocket connection to 'wss://...REMOVED.../panel-oauth-test/panel_app/ws' failed: 
connect @ bokeh.min.js?v=2630124d82b67a204b9c9da925d3d456e9ec585bb0cf8f839a1c704dbee3f200b70d237916a9fd429755022dc83ac334505cd7385ebb4453a7b17aa1d3699c35:249
bokeh.min.js?v=2630124d82b67a204b9c9da925d3d456e9ec585bb0cf8f839a1c704dbee3f200b70d237916a9fd429755022dc83ac334505cd7385ebb4453a7b17aa1d3699c35:249 [bokeh 3.9.0] Failed to connect to Bokeh server: Could not open websocket
_on_error @ bokeh.min.js?v=2630124d82b67a204b9c9da925d3d456e9ec585bb0cf8f839a1c704dbee3f200b70d237916a9fd429755022dc83ac334505cd7385ebb4453a7b17aa1d3699c35:249
bokeh.min.js?v=2630124d82b67a204b9c9da925d3d456e9ec585bb0cf8f839a1c704dbee3f200b70d237916a9fd429755022dc83ac334505cd7385ebb4453a7b17aa1d3699c35:248 [bokeh 3.9.0] Failed to load Bokeh session pJKROf71epgaXLhoJlnQRikGVYIql2VFTj0IIGl1A9gV: Error: Could not open websocket
[...]

Issue #7909, opened 2025-05-10
Token issue when using oauth encryption key


MarcSkovMadsen:
# WebSocket connection failures with Azure OAuth due to large token size

## Issue Description

When using Panel with Azure OAuth authentication on a Kubernetes deployment behind an ingress controller, WebSocket connections fail due to excessively large authentication tokens (10,183 bytes), causing the application to be unusable. The issue only occurs when using the `--oauth-encryption-key` parameter, which significantly increases token size (from 7,530 to 10,183 bytes).

## Environment

- Panel version: Current development version
- Deployment: Kubernetes with NGINX ingress controller
- Authentication: Azure OAuth
- Browser: Chrome/Edge (latest)

## Steps to Reproduce

1. Deploy a simple Panel application in Kubernetes with Azure OAuth authentication
2. Configure the application with `--oauth-encryption-key` parameter
3. Access the application through an ingress controller
4. Attempt to log in

## Expected Behavior

The application should successfully establish a WebSocket connection after authentication and work as expected.

## Actual Behavior

After authentication, the WebSocket connection fails with errors in the browser console:
<details>
<summary>Console Error Details</summary>

```
[bokeh 3.7.2] setting log level to: 'trace'
WebSocket connection to 'wss://example.org/apps/panel-oauth-app/script/ws' failed
[bokeh 3.7.2] Failed to connect to Bokeh server: Could not open websocket
[bokeh 3.7.2] Failed to load Bokeh session XXXXXXXXXXXXXXXXXXXXXXXXXXXX: Error: Could not open websocket
Error rendering Bokeh items: Error: Could not open websocket
[bokeh 3.7.2] Lost websocket 0 connection, 1006 ()
[bokeh 3.7.2] Websocket connection 0 disconnected, will not attempt to reconnect
```

The issue appears to be in the JavaScript code where it attempts to create a WebSocket connection:
```javascript
new WebSocket(e,["bokeh", this.token])
```
where `this.token` is 10,183 bytes when using `--oauth-encryption-key`.
</details>

## Analysis

I've confirmed that the large token size is causing the issue by testing with and without the `--oauth-encryption-key` parameter:

1. **With encryption key** (fails):
   - Token size: 10,183 bytes
   - Command:
<details>
<summary>Panel command with encryption key</summary>

```bash
panel serve script.py --oauth-provider=azure \
  --oauth-key='CLIENT_ID' \
  --oauth-secret='CLIENT_SECRET' \
[...]

PR #8638, opened 2026-06-06
Warn when the Bokeh session token may exceed proxy request-header limits

MarcSkovMadsen:
## Summary

Approved in #8634 ([comment](https://github.com/holoviz/panel/issues/8634#issuecomment-4630656272)).

When OAuth is enabled, Bokeh embeds the request's cookies/headers into the **session token**, which the browser echoes back in the **`Sec-WebSocket-Protocol`** header when opening the WebSocket. If that header exceeds the proxy's per-request-header limit (nginx `large_client_header_buffers`, **default 8&nbsp;KB**), the proxy rejects the upgrade with `HTTP 400` and the user sees only a cryptic **`Could not open websocket`** — with nothing in the Panel server logs, because the request never reaches Panel. This was the root cause in #8634 (and #7909) and took the reporter ~2 years to diagnose.

## Change

`DocHandler.get` now logs a **one-time, informational warning per app** when the generated session token approaches the common 8&nbsp;KB proxy limit. The message is deliberately conditional — a large token is only a problem if a proxy actually rejects it, so it tells the operator to ignore the message if the app connects fine, and only points at remedies (`--exclude-cookies`, raising the proxy header buffer, the troubleshooting docs) for the case where they *do* hit `Could not open websocket`. Non-fatal; purely diagnostic.

```
The Bokeh session token for '/app' is 8123 bytes. The browser sends this token in the
'Sec-WebSocket-Protocol' header when opening the WebSocket connection. Some proxies reject
request headers larger than 8 kB (e.g. nginx 'large_client_header_buffers'); if yours does,
the connection will fail with 'Could not open websocket'. This is only a problem if you
actually see that error - if the app connects fine you can ignore this message. Otherwise,
raise the proxy's request-header buffer or reduce the token size (e.g. `panel serve
--exclude-cookies ...`). See https://panel.holoviz.org/how_to/authentication/trouble_shooting.html
```

The threshold (`WS_TOKEN_SIZE_WARNING_THRESHOLD = 7600`) leaves headroom under the 8&nbsp;KB default for the `"bokeh, "` subprotocol prefix.

## Tests

`test_warn_if_ws_token_too_large` verifies a small token does not warn, a large token warns, and the warning fires at most once per app path.

Refs #8634, #7909.

### DIFF
--- a/panel/io/server.py
+++ b/panel/io/server.py
@@ -125,6 +125,43 @@ _MAX_ROUTE_PARAM_KEY_CHARS = 128
 _MAX_ROUTE_PARAM_VALUE_CHARS = 1024
 _MAX_APP_PATH_CHARS = 2048
 
+# The browser echoes the Bokeh session token back in the `Sec-WebSocket-Protocol`
+# request header when opening the WebSocket connection. Many proxies reject
+# request headers larger than 8 kB (e.g. nginx `large_client_header_buffers`),
+# which surfaces as a cryptic "Could not open websocket". Warn once per app when
+# the token approaches that size so the cause is discoverable. See
+# https://panel.holoviz.org/how_to/authentication/trouble_shooting.html
+WS_TOKEN_SIZE_WARNING_THRESHOLD = 7600
+_ws_token_size_warned: set[str] = set()
+
+
+def _warn_if_ws_token_too_large(path: str, token: str) -> None:
+    """
+    Warn (once per app path) when the session token is large enough to risk
+    rejection by a proxy's request-header size limit.
+
+    The browser sends this token in the ``Sec-WebSocket-Protocol`` header when
+    opening the WebSocket connection; proxies such as nginx reject request
+    headers larger than 8 kB by default, which surfaces as a cryptic
+    "Could not open websocket".
+    """
+    if len(token) <= WS_TOKEN_SIZE_WARNING_THRESHOLD or path in _ws_token_size_warned:
+        return
+    _ws_token_size_warned.add(path)
+    logger.warning(
+        "The Bokeh session token for %r is %d bytes. The browser sends this "
+        "token in the 'Sec-WebSocket-Protocol' header when opening the "
+        "WebSocket connection. Some proxies reject request headers larger than "
+        "8 kB (e.g. nginx 'large_client_header_buffers'); if yours does, the "
+        "connection will fail with 'Could not open websocket'. This is only a "
+        "problem if you actually see that error - if the app connects fine you "
+        "can ignore this message. Otherwise, raise the proxy's request-header "
+        "buffer or reduce the token size (e.g. `panel serve --exclude-cookies "
+        "...`). See "
+        "https://panel.holoviz.org/how_to/authentication/trouble_shooting.html",
+        path, len(token),
+    )
+
 
 def _has_prefix_boundary(path: str, prefix: str) -> bool:
     if path == prefix:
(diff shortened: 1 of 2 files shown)


MarcSkovMadsen:
Manual test

<details><summary>script.py</summary>

```python
"""Manual test for ``panel.io.server._warn_if_ws_token_too_large`` (PR #8638).

Run on the ``warn-large-ws-token`` branch with:

    python script.py

It calls the helper directly with controlled token sizes and prints what
happens, so you can eyeball the one-time-per-app warning behaviour without
needing a real oversized OAuth token. Expected:

    [1] small token            -> no warning
    [2] large token /big-app   -> WARNING (once)
    [3] large token /big-app   -> no warning (already warned this path)
    [4] large token /other-app -> WARNING (new path)
"""
import logging
import sys

# Make the panel.io.server logger visible on stdout (so warnings interleave in
# order with the print() lines below) regardless of how Panel configures
# logging in this environment.
_logger = logging.getLogger("panel.io.server")
_logger.setLevel(logging.WARNING)
_handler = logging.StreamHandler(sys.stdout)
_handler.setFormatter(logging.Formatter("  >> %(levelname)s %(name)s: %(message)s"))
_logger.addHandler(_handler)
_logger.propagate = False  # avoid duplicate lines if a root handler also exists

from panel.io.server import (  # noqa: E402
    WS_TOKEN_SIZE_WARNING_THRESHOLD, _warn_if_ws_token_too_large,
    _ws_token_size_warned,
)

threshold = WS_TOKEN_SIZE_WARNING_THRESHOLD
big = "x" * (threshold + 1)
small = "x" * 100

# Start from a clean slate in case the module already warned for some path.
_ws_token_size_warned.clear()

print(f"WS_TOKEN_SIZE_WARNING_THRESHOLD = {threshold} bytes\n")

print(f"[1] small token ({len(small)} bytes) on /small-app -> expect NO warning")
_warn_if_ws_token_too_large("/small-app", small)

print(f"\n[2] large token ({len(big)} bytes) on /big-app -> expect ONE warning")
_warn_if_ws_token_too_large("/big-app", big)

print(f"\n[3] large token ({len(big)} bytes) on /big-app again -> expect NO new warning")
_warn_if_ws_token_too_large("/big-app", big)

print(f"\n[4] large token ({len(big)} bytes) on /other-app -> expect a NEW warning")
_warn_if_ws_token_too_large("/other-app", big)

print(f"\nPaths that have warned: {sorted(_ws_token_size_warned)}")
```
</details>

```bash
$ python script.py
WS_TOKEN_SIZE_WARNING_THRESHOLD = 7600 bytes

[1] small token (100 bytes) on /small-app -> expect NO warning

[2] large token (7601 bytes) on /big-app -> expect ONE warning
  >> WARNING panel.io.server: The Bokeh session token for '/big-app' is 7601 bytes. The browser sends this token in the 'Sec-WebSocket-Protocol' header when opening the WebSocket connection. Some proxies reject request headers larger than 8 kB (e.g. nginx 'large_client_header_buffers'); if yours does, the connection will fail with 'Could not open websocket'. This is only a problem if you actually see that error - if the app connects fine you can ignore this message. Otherwise, raise the proxy's request-header buffer or reduce the token size (e.g. `panel serve --exclude-cookies ...`). See https://panel.holoviz.org/how_to/authentication/trouble_shooting.html

[3] large token (7601 bytes) on /big-app again -> expect NO new warning

[4] large token (7601 bytes) on /other-app -> expect a NEW warning
  >> WARNING panel.io.server: The Bokeh session token for '/other-app' is 7601 bytes. The browser sends this token in the 'Sec-WebSocket-Protocol' header when opening the WebSocket connection. Some proxies reject request headers larger than 8 kB (e.g. nginx 'large_client_header_buffers'); if yours does, the connection will fail with 'Could not open websocket'. This is only a problem if you actually see that error - if the app connects fine you can ignore this message. Otherwise, raise the proxy's request-header buffer or reduce the token size (e.g. `panel serve --exclude-cookies ...`). See https://panel.holoviz.org/how_to/authentication/trouble_shooting.html

Paths that have warned: ['/big-app', '/other-app']
```

MarcSkovMadsen:
Failing UI tests are unrelated to this PR.

PR #8660, opened 2026-07-01
[codex] Fix repeated location sync mappings

### LINKED ISSUE #8654: Bug with multiple pn.state.location.sync calls

<details>
    <summary>panel==1.5.4</summary>

```plaintext
No change of code on this area until v1.9.3
```
</details>

When calling `pn.state.location.sync` twice in the same class, the synchronization of parameters between the class and the URI doesn't work for all parameters in the second call. An error is thrown whenever we change a parameter from the second list.

`test_sync.py` file content:
```python
import panel as pn
import param

class Main_App(param.Parameterized):
    theme = param.Selector(
        default="light",
        objects=["light", "dark"],
    )
    font_size = param.Integer(default=12)

    def __init__(self, **params):
        super().__init__(**params)
        if pn.state.location:
            pn.state.location.sync(
                self,
                ["theme", "font_size"],
            )

    def panel(self):
        return pn.Param(self)

class UserSettings(Main_App):
    language = param.Selector(
        default="en",
        objects=["en", "fr", "es"],
    )
    notifications = param.Boolean(default=True)

    def __init__(self, **params):
        super().__init__(**params)
        if pn.state.location:
            pn.state.location.sync(
                self,
                ["language", "notifications"],
            )

    def panel(self):
        return pn.Param(self)

# Create an instance of UserSettings
user_settings = UserSettings()

# Serve the panel
user_settings.panel().servable()
```

Execute the following:
```bash
panel serve test_sync.py
```

Access http://localhost:5006/test_sync then change the language.

Following error appears:

```python
  File "<...>/.pixi/envs/default/lib/python3.10/site-packages/panel/io/location.py", line 212, in _update_query
    serialized[matches[0][e.name]] = val
KeyError: 'language'
```

#### PR proposition

I would turn the private parameter `pn.state.location._synced` into a dict where the key is the variable `parameterized` from `pn.state.location.sync`, I think it would also simplify other functions like `_server_destroy` and `unsync`

EltonChang1:
Fixes #8654.

## What changed
- Updated `Location._update_query` to choose only sync mappings that contain the changed parameter.
- Added a regression test covering repeated `Location.sync` calls for the same `Parameterized` object.

## Why
Calling `Location.sync` multiple times for one object stores multiple parameter mappings. Parameter updates from later sync calls previously indexed the first mapping and raised `KeyError` for parameters that only existed in a later mapping.

## Validation
- `.venv/bin/python -m pytest panel/tests/io/test_location.py -k same_object_multiple_calls -q`
- `.venv/bin/python -m pytest panel/tests/io/test_location.py -q`
- `git diff --check`
- Manual reproduction snippet for issue #8654 now updates `language` without raising.

## Limitations
Full test suite and browser UI tests were not run locally.

## AI disclosure
This draft PR was prepared with assistance from OpenAI Codex/ChatGPT. The code change, regression test, PR body, and local validation workflow were AI-assisted and reviewed in the local checkout before submission.

### DIFF
--- a/panel/io/location.py
+++ b/panel/io/location.py
@@ -224,7 +224,10 @@ class Location(Syncable):
             return
         serialized = query or {}
         for e in events:
-            matches = [ps for o, ps, _, _ in self._synced if o in (e.cls, e.obj)]
+            matches = [
+                ps for o, ps, _, _ in self._synced
+                if o in (e.cls, e.obj) and e.name in ps
+            ]
             if not matches:
                 continue
             owner = e.cls if e.obj is None else e.obj
--- a/panel/tests/io/test_location.py
+++ b/panel/tests/io/test_location.py
@@ -94,6 +94,20 @@ def test_location_sync_query(location):
     assert location._synced == []
     assert location.search == ""
 
+def test_location_sync_same_object_multiple_calls(location):
+    p = SyncParameterized(integer=1, string='abc')
+    location.sync(p, ['integer'])
+    location.sync(p, ['string'])
+
+    p.string = 'def'
+    assert location.search == "?integer=1&string=def"
+    p.integer = 2
+    assert location.search == "?integer=2&string=def"
+
+    location.unsync(p)
+    assert location._synced == []
+    assert location.search == ""
+
 def test_location_sync_param_init(location):
     p = SyncParameterized()
     location.search = "?integer=1&string=abc"


philippjfr:
Test failures unrelated.

philippjfr:
Thank you @EltonChang1, much appreciated!

PR #8664, opened 2026-07-05
Serialize shapely geometry as WKT for Bokeh

### LINKED ISSUE #8663: Tabulator cannot serialize a GeoDataFrame geometry column (bokeh SerializationError)
When a `Tabulator` is backed by a `GeoDataFrame`, the geometry column's raw shapely objects end up in the widget's `ColumnDataSource`. On sync/patch to the browser, bokeh raises `SerializationError: can't serialize <class 'shapely.geometry.polygon.Polygon'>`, which breaks the session.

The initial render can appear fine, but a data update (patching the `ColumnDataSource`) triggers the failure, so it shows up as a silent stall in a served app.

#### Reproduce 

```python
import geopandas as gpd
import panel as pn
from shapely.geometry import Polygon
from bokeh.core.serialization import Serializer

gdf = gpd.GeoDataFrame(
    {"v": [1, 2], "geometry": [Polygon([(0, 0), (1, 0), (1, 1)]),
                               Polygon([(2, 0), (3, 0), (3, 1)])]},
    crs="EPSG:4326",
)
w = pn.widgets.Tabulator(gdf)
doc = Document()
w.get_root(doc)
w.value = gdf.copy()               # patches the ColumnDataSource
for m in doc.models:
    if hasattr(m, "data"):
        Serializer().encode(m.data)   # SerializationError: can't serialize Polygon
```

#### Expected

Tabulator should be able to display a geometry column (for example by serializing it as WKT text), or at minimum document that geometry columns must be stringified before being passed to Tabulator.

#### Environment
panel 1.8.10a0, bokeh 3.8.2, geopandas 1.1.4, shapely 2.1.2, Python 3.14.3

ghostiee-11:
## Description

Fixes #8663

Bokeh cannot serialize shapely geometry, so a GeoDataFrame geometry column crashes any Panel component that syncs it (Tabulator, hover, holoviews glyphs) and the render stalls. This registers Bokeh serializers that encode shapely geometries as WKT text so the data reaches the frontend, wired into `add_to_doc` and `diff`. A no-op until shapely is imported.


maximlt:
Is the `geometry` column meant to be displayed at all in a table? As a user, I'm not sure that's useful. For polygons/lines,  the serialized wkt might be huge.

ghostiee-11:
yeah fair. it's less about displaying geometry in a table and more that Tabulator(gdf) just hard-crashes today, SerializationError: can't serialize Polygon the moment it serializes. wkt is only so it doesn't crash, and it's the smallest lossless option I checked (wkb and geojson both come out bigger than json since the size is just the coordinate data).

Just checked, lumen already handles geometry per-view on its own, so it doesn't really need this.
happy to close it.

maximlt:
> Just checked, lumen already handles geometry per-view on its own, so it doesn't really need this.

How did you hit this problem if it's handled in Lumen? By using Panel directly maybe?

---

I haven't checked but it sounds like it's a generic problem, Tabulator(df) will break if the dataframe contains a column with objects that can't be serialized by Bokeh. If that's the case, one approach is to make this more robust by not erroring.

ghostiee-11:
i hit it building geopandas support in lumen, the geometry column reached bokeh's serializer. i handle it per view in my geopandas pr (holoviz/lumen#1903): table -> wkt, deckgl -> geojson, and hvplot already goes to coord arrays, so lumen won't depend on this. the wkt serializer here was just the approach i tried for the crash, but you're right it's the generic serialization case.

Coderambling:
> Is the `geometry` column meant to be displayed at all in a table? As a user, I'm not sure that's useful. For polygons/lines, the serialized wkt might be huge.

A user could potentially want to display coordinates in a table?

So If the Shapely object in the geometry column is a coordinate, i.e. POINT (long lat), could that be a use case for displaying the column in a table?

maximlt:
> A user could potentially want to display coordinates in a table?

Yes. Is this the most common scenario? I tend to believe that's not the case. For lines and polygons, the values are going to be difficult to parse for most datasets, only the type would be useful to see. For points, yes, maybe that's fine, but they might as well create x/y or lat/lon columns, which would be easier to view, sort, etc.

Let's not forget Tabulator is a widget component that allows editing the table content. Should we support editing the geometry column?

Btw I'm not again having special handling of the geometry column of geodataframes. I'm just thinking that the default of displaying is as WKT is imo not the best default.

Coderambling:
1. > > A user could potentially want to display coordinates in a table?

 > Yes. Is this the most common scenario? I tend to believe that's not the case. For lines and polygons, the values are going to be difficult to parse for most datasets, only the type would be useful to see. For points, yes, maybe that's fine, but they might as well create x/y or lat/lon columns, which would be easier to view, sort, etc.

2. > Let's not forget Tabulator is a widget component that allows editing the table content. Should we support editing the geometry column?
3. > Btw I'm not again having special handling of the geometry column of geodataframes. I'm just thinking that the default of displaying is as WKT is imo not the best default.


Makes sense @maximlt .

1. Agree.

A scenario could be to have Tabulator handle POINT and MULTIPOINT values, and show type only for all other Shapely objects, in combination with a warning / explanation, but it does not seem ideal.

2. Hm. Maybe not. Is there a need for this in practice for users working with GeoDataframes? How do you see this @ghostiee-11 ? Is manual editing of GeoDataframes a relevant use case?

3. Special handling by Tabulator? Or at a higher level in Panel?

For now, should there at be a separate PR so that Tabulator (and other components?) throw a graceful error (with a link to a doc) when a GeoDataframe is used?

Broadly agree with all 3 points @maximlt .

I see 2 potential PR's. Please advise, and I wil create them.


1. Graceful handling / emitting a warning (with link to guidance) when users feed  a `geometry` colum to Tabulator? And / or to other components?

> I haven't checked but it sounds like it's a generic problem, Tabulator(df) will break if the dataframe contains a column with objects that can't be serialized by Bokeh. If that's the case, one approach is to make this more robust by not erroring.

So does the above mean that Tabulator will not crash anymore, if this PR (#8664 ) is implemented @ghostiee-11? Will Tabulator instead just display what is in the column (even if it is not very useful in most cases)?

So, for example, will it just show a string: [Polygon([(0, 0), (1, 0), (1, 1)]) , if that is what is in the `geometry` colum?


2. Documentation / discoverability of Holoviz GeoPandas capabilities:

I could not find any mention of GeoPandas in the Panel website, and only 2 in https://holoviews.org :

https://holoviews.org/user_guide/Geometry_Data.html
https://holoviews.org/releases.html

And only 2 in https://hvplot.org .

Should be at least be mentioned in Panel docs, with a reference to the holoviews or https://hvplot.holoviz.org website, i.e. a link to https://hvplot.holoviz.org/en/docs/latest/gallery/index.html#geospatial

For example add a mention in this doc: https://panel.holoviz.org/gallery/hvplot_explorer.html

This would enable / improve discoverability of this Holoviz functionality for Panel website visitors, and for seach engines and AI .

ghostiee-11:
On editing, manual geometry editing isn't a real use case, hand-typing WKT for a polygon isn't practical, so it's best kept for display only.

And yeah, this pr stops the crash. Today, a live Tabulator(gdf) hard-crashes` (SerializationError: can't serialize Polygon) `because the column holds raw shapely objects. With it the geometry serializes as wkt text, so the cell shows POLYGON ((0 0, 1 0, 1 1, 0 0)) or POINT (5 6), not the shapely repr. Agree wkt reads fine for points, but is long and unwieldy for polygons and lines though.

Both PRs imo make sense. The graceful "don't error" one is a better default than wkt and matches the earlier suggestion, and the docs/discoverability one is a nice add. Lumen already handles geometry per-view, so it doesn't depend on this either way.

PR #8665, opened 2026-07-06
Fix typo and optimize phrasing

Coderambling:
Let me know what you think @philippjfr . Is greatly optimized testability an exaggeration? Maybe just improved testability is better?

I assume the phrase is meant as: improved testability versus using functions.

Or does it also mean do say that the user _gains_ the ability to test purely via test scripts (so no widgets, no layout, no browser needed)?

Maybe also point out that this advantage grows (exponentially) as a codebase becomes larger?


philippjfr (approved):
Agree with the statement. Thanks @Coderambling

PR #8679, opened 2026-07-23
Compatibility with bokeh 3.10 dev

philippjfr:
Adapts Panel's bokeh models and Python code to bokeh 3.10's refactored view/stylesheet APIs.

### TypeScript model changes

The core of the PR adapts to bokeh 3.10's rewritten DOMComponentView rendering and stylesheet pipeline.

- Render pipeline (card.ts, reactive_esm.ts, reactive_html.ts): the old `_update_stylesheets()` / `_update_css_classes()` / `_apply_styles()` / `_update_css_variables()` sequence collapsed into `_apply_stylesheets()` + `_apply_html_attributes()`, matching the new base class.
- children() → `_children_views()` (button.ts, checkbox_button_group.ts, radio_button_group.ts, icon.ts, file_download.ts): bokeh replaced the `*children(): IterViews` generator with `_children_views(): ChildView[]` returning an array (the base filters nulls).
- `react_component.ts`: the largest change. The new `InlineStyleSheet` drops install/uninstall/persistent; inline sheets go into `adoptedStyleSheets` via a reactive `to_native()` and links via `to_element()`. Since Panel's non-shadow-DOM path shares one shadow root across sibling React components, the override merges adopted sheets into the root (rather than clobbering), tracks them for clean removal, and preserves sibling dedup by CSS text and href. HostedStyleSheet loses its persistent arg, and self_style is reassigned instead of the now-getter-only style.
- `tabs.ts`: bokeh's Tabs became a vDOM component, so the dead update_active override (referencing the removed header_els) was deleted.
- `layout.ts`: fixed `watch_stylesheets()` in both `PanelMarkupView` and `HTMLBoxView`, since `_applied_stylesheets` is now `HTMLElement[]` directly rather than objects with an `.el` property.
- package.json: bumped @bokeh/bokehjs to ^3.10.0-dev.7 (with lockfile update).

### Python changes

- _param.py, models/vega.py: `enumeration()` no longer unpacks a `typing.Literal`, so both call sites now pass `*t.get_args(...)` explicitly.
- pyproject.toml: bokeh caps raised from <3.10.0 to <3.11.0 (build and runtime).
- pixi.toml: bokeh pinned to ==3.10.0.dev7.


pixi.toml
hoxbro:
```suggestion
channels = ["pyviz/label/dev", "bokeh/label/dev", "conda-forge"]
```

mattpap:
> `enumeration()` no longer unpacks a `typing.Literal` (...)

It actually still does and more, but you have to use `type` syntax, e.g.

```py
In [1]: from bokeh.core.enums import enumeration

In [2]: from typing import Literal

In [3]: type ABC = Literal["a", "b", "c"]

In [4]: enumeration(ABC)
Out[4]: Enumeration(a, b, c)
```

mattpap:
> pyproject.toml: bokeh caps raised from <3.10.0 to <3.11.0 (build and runtime).

We dropped two versions of Python in this release, so the minimum is 3.12 right now.

PR #8680, opened 2026-07-23
Ensure explicitly set sizing options are respected

philippjfr:
During the last bokeh layout changes we added `sizing_mode` inference from a layout component's children. This achieved backward compatibility but meant that explicitly set sizing options were not always respected. This addresses this problem by tracking which options were set explicitly.


philippjfr:
UI test failures unrelated.

PR #8685, opened 2026-07-24
Support movable Tabulator columns

### LINKED ISSUE #8567: Tabulator: movableColumns is hardcoded to false, ignoring user configuration

`pn.widgets.Tabulator` does not expose a way to enable Tabulator.js's native column
reordering (drag-and-drop on column headers). Passing the option through the
`configuration` parameter has no effect, because `getConfiguration()` in
`panel/models/tabulator.ts` hardcodes `movableColumns: false` *after* spreading the
user-provided configuration, silently overriding it:

```ts
// panel/models/tabulator.ts (line 821 in Panel 1.8.9)
const configuration = {
  ...transformJsPlaceholders(this.model.configuration),
  index: "_index",
  nestedFieldSeparator: false,
  movableColumns: false,   // <-- overrides user value
  ...
}
```

Minimal reproducer (expect drag-and-drop, observe it is disabled):

```python
import panel as pn
import pandas as pd
pn.extension('tabulator')

df = pd.DataFrame({'A': [1, 2, 3], 'B': [4, 5, 6], 'C': [7, 8, 9]})

pn.widgets.Tabulator(
    df,
    configuration={'movableColumns': True},
).servable()
```

End users have no straightforward workaround: neither `configuration=` nor any
dedicated Param lets them enable the feature from Python.

Expose `movable_columns` as a first-class Param on `pn.widgets.Tabulator`, following
the existing pattern used for `frozen_rows`, `hidden_columns`, `header_filters`, etc.

Three coordinated changes:

1. **`panel/widgets/tables.py`** — declare the Param:
   ```python
   movable_columns = param.Boolean(default=False, doc="""
       If True, columns can be reordered by dragging their headers.""")
   ```

2. **`panel/models/tabulator.py`** — declare the Bokeh model property:
   ```python
   movable_columns = Bool(default=False)
   ```

3. **`panel/models/tabulator.ts`** — read the property instead of hardcoding:
   ```ts
   movableColumns: this.model.movable_columns,
   ```

Default remains `False`, so existing code is unaffected. Enabling the feature becomes
the idiomatic:

```python
pn.widgets.Tabulator(df, movable_columns=True)
```

Happy to submit a PR with the changes, bundled-JS rebuild (`panel build panel`), and a
Playwright UI test, if the maintainers agree with this approach.

1. **Respect `configuration` pass-through without a new Param** — change line 821 to
   `movableColumns: this.model.configuration?.movableColumns ?? false`. This is less
   invasive (one-line fix, no new public API) but leaves the feature undiscoverable
   from docstrings / tab-completion, and inconsistent with how other Tabulator
   features are surfaced in Panel.

2. **Name the Param `reorderable`** instead of `movable_columns`. Shorter and more
   Pythonic, but diverges from the underlying Tabulator.js option name, which may hurt
   discoverability for users coming from Tabulator docs.

3. **Status quo (client-side hack)** — patch the installed `panel.min.js` locally.
   Works, but does not scale and is lost on every `pip install --upgrade panel`.

#### Additional context

- Panel version tested: 1.8.9
- Bundled Tabulator.js: 6.3.1 (per `dist/bundled/datatabulator/tabulator-tables@6.3.1/`)
- The hardcoding appears to predate several releases. If there is a historical reason
  for disabling `movableColumns` by default (e.g. interaction with `selectable='checkbox'`
  or `frozen_columns`), I'd be glad to document it and guard against the problematic
  combinations rather than disabling the feature outright.
- Related Tabulator.js docs: https://tabulator.info/docs/6.3/layout#movable-columns

Kkkakania:
## What this changes

Adds a first-class `movable_columns` parameter to `Tabulator`. When enabled, the value is synchronized through the Bokeh model and passed to Tabulator.js as `movableColumns`, allowing users to reorder columns. The default remains `False`, preserving existing behavior. Runtime changes rebuild the frontend table with the updated setting.

The parameter is included in the Tabulator reference and covered by Python model-sync and browser integration tests.

Closes #8567.

## Validation

- `pytest panel/tests/widgets/test_tables.py::test_tabulator_movable_columns -q`
- `pytest panel/tests/ui/widgets/test_tabulator.py::test_tabulator_movable_columns -q --ui` (Chromium)
- `panel build panel`
- `pre-commit run --files examples/reference/widgets/Tabulator.ipynb panel/models/tabulator.py panel/models/tabulator.ts panel/widgets/tables.py panel/tests/widgets/test_tables.py panel/tests/ui/widgets/test_tabulator.py`


Kkkakania:
CI note: the type job currently fails only on the existing `panel/io/document.py:313` and `panel/io/server.py:257` coroutine callback annotations; neither file is touched here. The PR-specific checks are green so far: pre-commit, core on Bokeh 3.8, and core on the current dependency set. Locally, the new model-sync test and Chromium integration test pass, and `panel build panel` compiles all 85 TypeScript files.

Kkkakania:
CI follow-up:

- The new `test_tabulator_movable_columns` passes in the workflow, and the complete Windows and macOS UI jobs are green. Core tests, all unit-test matrices, the frontend build, and pre-commit are also green.
- Ubuntu UI completed 762 tests and failed only three pre-existing tests (`test_serve_page_with_reactive_html_css_on_nested_route`, `test_tabulator_frozen_rows`, and `test_tabulator_update_hidden_columns`); none exercises `movable_columns`. The failures are timeout/missing-element symptoms, while the same UI suite is green on Windows and macOS.
- The type job still reports only the two pre-existing errors in `panel/io/document.py:313` and `panel/io/server.py:257`.

I attempted to rerun the failed jobs, but GitHub requires repository admin rights for that action. A maintainer rerun should distinguish the Ubuntu UI flake without changing this patch.

philippjfr:
My only question/concern is whether there's any issue with editing a table where the columns have been moved around.

Kkkakania:
Good question. Tabulator keeps the field identity when a column is moved, so edits should still be routed by the column field rather than its visual position. I added a browser regression test in `7486ff4a` that:

1. moves column `A` after `B`,
2. edits the moved `A` cell through the rendered numeric editor,
3. waits for the edit to reach Python, and
4. asserts `widget.value["A"]` changed while `widget.value["B"]` remained untouched.

The updated Chromium UI test passes locally, as does the Python model synchronization test. I also merged the latest `main`, including the revert of #8680 that caused the all-platform failures in the previous merge snapshot, so CI is now rerunning on the corrected baseline.

Kkkakania:
One more CI note: the fresh unit failure is now on `test_markdown_codeblocks[how_to/layout/size.md]`, where current `main` passes the deprecated `Widget.name` argument and warnings are treated as errors. It is unrelated to movable columns. I reproduced it against `main`, fixed it with `label=`, and opened the focused follow-up #8688; the exact docs test passes there.

philippjfr:
Test failures are unrelated. Will fix those shortly.

Issue #8692, opened 2026-07-29
hold() leaves events queued on the Document when the session is not connected


SimonHeybrock:
#### ALL software version info

<details>
    <summary>Software Version Info</summary>

```plaintext
Panel: main @ c97e8098
Bokeh: 3.9.1
Python: 3.11
OS: Linux
```
</details>

#### Description of expected behavior and the observed behavior

When a `hold()` block exits while the Document is not connected, the `finally` chain in `panel/io/document.py` takes

```python
elif not state._connected.get(doc):
    doc.callbacks._hold = None
```

which clears the hold policy but leaves `doc.callbacks._held_events` populated. Bokeh's `unhold()` returns early when `_hold is None`, so nothing drains them there. The events sit on the Document until some later, unrelated `hold()`/`unhold()` pair flushes them — by which point the session may be connected, and they are dispatched well outside the context they were recorded in.

Expected: leaving a `hold()` block leaves nothing queued behind it, whether the events are dispatched or deliberately discarded.

#### Complete, minimal, self-contained example code that reproduces the issue

```python
import time

import panel as pn
import requests

from panel.io.state import state

info, docs = {}, []

def app():
    md = pn.pane.Markdown("initial")
    docs.append(state.curdoc)

    def before_ready():
        doc = state.curdoc
        with pn.io.hold():
            md.object = "changed"
        info['connected'] = state._connected.get(doc)
        info['hold'] = doc.callbacks.hold_value
        info['queued'] = [type(e).__name__ for e in doc.callbacks._held_events]

    state.curdoc.add_next_tick_callback(before_ready)
    return md

pn.serve(app, port=5099, show=False, threaded=True)
time.sleep(1)
requests.get("http://localhost:5099")
time.sleep(1)

print("on exiting the hold:", info)
print("still queued afterwards:", [type(e).__name__ for e in docs[0].callbacks._held_events])
```

Output:

```
on exiting the hold: {'connected': None, 'hold': None, 'queued': ['SessionCallbackAdded']}
still queued afterwards: ['SessionCallbackAdded']
```

The hold is gone but the event it collected is not, and nothing on the Document will drain it until an unrelated hold happens to.

[...]

PR #8693, opened 2026-07-29
Claim the Document thread before running app code

### LINKED ISSUE #8691: pn.io.hold() in the app callable leaks the hold, causing duplicate session-callback registration

Panel 1.9.3, Bokeh 3.9.2, Python 3.11, Linux.

A session callback registered in the app callable *after* a `pn.io.hold()` block is registered twice by Bokeh, so every session open logs an unhandled `ValueError: A callback of the same type has already been added with this ID`.

`hold()` decides whether it was called off the document's thread with

```python
threaded = state._current_thread != state._thread_id
```

`state._thread_id` is only assigned in `init_doc()`, which runs *after* the app callable returns. During the app callable it is therefore `None` and `threaded` is unconditionally `True`, so `hold()` takes the off-thread path: rather than unholding on exit it schedules `_unhold` as a next-tick callback and restores `_hold`. The document stays held for the rest of the build.

Bokeh then constructs the `ServerSession`, whose `__init__` registers every session callback already on the document. The deferred `_unhold` runs afterwards and replays the queued `SessionCallbackAdded` event, registering the same callback a second time.

Note that `state._unblocked()` resolves the same ambiguity the other way — `self._thread_id in (self._current_thread, None)` treats an unset thread id as *not* off-thread — so the two are inconsistent.

Beyond the log noise: `DocumentCallbackManager.unhold()` dispatches in a plain loop, so the exception drops every event still queued behind the offending one. In a real app that was 38 of the 39 events queued during the initial render.

#### Complete, minimal, self-contained example code

```python
import panel as pn


def app():
    with pn.io.hold():
        pass
    pn.state.add_periodic_callback(lambda: None, period=1000)
    return pn.pane.Markdown("hello")


pn.serve(app, port=5051, show=False)
```

Open http://localhost:5051 — the traceback below is logged once per session opened.

#### Stack traceback

```
  File ".../bokeh/server/session.py", line 226, in with_document_locked
    return func(*args, **kwargs)
  File ".../bokeh/document/callbacks.py", line 193, in remove_then_invoke
    return callback()
  File ".../panel/io/document.py", line 622, in _unhold
    doc.unhold()
  File ".../bokeh/document/document.py", line 831, in unhold
    self.callbacks.unhold()
  File ".../bokeh/document/callbacks.py", line 450, in unhold
    self.trigger_on_change(event)
  File ".../bokeh/document/events.py", line 876, in dispatch
    cast(SessionCallbackAddedMixin, receiver)._session_callback_added(self)
  File ".../bokeh/server/session.py", line 261, in _session_callback_added
    self._callbacks.add_session_callback(wrapped)
  File ".../bokeh/server/callbacks.py", line 198, in add_session_callback
    self._group.add_periodic_callback(callback_obj.callback, callback_obj.period, callback_obj.id)
  File ".../bokeh/util/tornado.py", line 191, in _assign_remover
    raise ValueError("A callback of the same type has already been added with this ID")
ValueError: A callback of the same type has already been added with this ID
```

#### Suggested fix

```python
threaded = state._thread_id is not None and state._current_thread != state._thread_id
```

Assigning `state._thread_id_[doc]` before the app callable runs (the equivalent condition) makes the error disappear, and leaves no hold and no queued events after the build.

SimonHeybrock:
## Description

`state._thread_id` is only recorded in `init_doc()`, which runs *after* the app callable returns. While the app callable is running it is therefore `None`, and `hold()` decides whether it was called off the Document's thread with `threaded = state._current_thread != state._thread_id` — so it unconditionally concludes it is off-thread. Rather than unholding on exit it schedules `_unhold` as a next-tick callback and restores `_hold`, leaving the Document held for the rest of the build. Bokeh then constructs the `ServerSession`, whose `__init__` registers every session callback already on the Document, and the deferred unhold afterwards replays the queued `SessionCallbackAdded` event — registering the same callback a second time.

Every session opened by an app that registers a session callback after a `hold()` block therefore logs `ValueError: A callback of the same type has already been added with this ID`:

```python
import panel as pn

def app():
    with pn.io.hold():
        pass
    pn.state.add_periodic_callback(lambda: None, period=1000)
    return pn.pane.Markdown("hello")

pn.serve(app, port=5051, show=False)
```

Beyond the log noise, `DocumentCallbackManager.unhold()` dispatches in a plain loop, so the exception drops every event still queued behind the offending one. In the app where I hit this that was 38 of the 39 events queued during the initial render.

The Document's thread is knowable before user code runs, so `Application.initialize_document` now claims it. That writes the same value `init_doc` writes moments later — `init_doc` runs synchronously inside the same build — just early enough for the app callable to see it. `init_doc` keeps its own assignment, which still covers Panel objects embedded in a plain Bokeh `Application`.

The alternative I originally suggested on the issue, guarding the comparison with `state._thread_id is not None and ...`, resolves the ambiguity by guessing that an unset thread id means "on-thread". That guess also fires where the thread id is legitimately never set — notably notebooks, where `init_doc` returns early because there is no `session_context` — and would silently move `hold()` there onto a different path, so I went with making the thread id known instead.

Fixes #8691

### A question on #8619

This moves a build-time `hold()` from the `threaded` branch of the `finally` chain onto the `not state._connected` branch, which drops the hold policy but leaves `_held_events` populated for an unrelated later `unhold()` to flush. Build-time holds behaved exactly that way before #8619 reordered the chain, so this restores the older behaviour rather than introducing something new — but was that reordering meant to change what happens to events queued while the app is still being built? Raised separately in #8692, since the leftover events outlive this PR either way.

## AI Disclosure

Usage: Claude Code diagnosed the root cause against the Panel and Bokeh sources, chose where the fix should live, wrote the fix and the regression test, and ran the verification — the reproducer above before and after, the regression test confirmed failing on the parent commit, and the full non-UI test suite compared against a baseline run (no new failures; the 18 pre-existing ones are missing JS bundles and `bokeh_fastapi` in my environment). I reported the original issue and reviewed the change.

## Checklist

### DIFF
--- a/panel/io/application.py
+++ b/panel/io/application.py
@@ -6,6 +6,7 @@ from __future__ import annotations
 import json
 import logging
 import os
+import threading
 import typing as t
 
 from functools import partial
@@ -165,6 +166,10 @@ class Application(BkApplication):
 
     def initialize_document(self, doc):
         logger.info(LOG_SESSION_LAUNCHING, id(doc))
+        # Claim the Document for the current thread before user code runs,
+        # so APIs that behave differently when invoked off the Document's
+        # thread (e.g. hold) can tell the two cases apart during the build.
+        state._thread_id_[doc] = threading.get_ident()
         self._set_session_prefix(doc)
         super().initialize_document(doc)
         if doc in state._templates and doc not in state._templates[doc]._documents:
--- a/panel/tests/io/test_document.py
+++ b/panel/tests/io/test_document.py
@@ -78,6 +78,27 @@ def test_hold_does_not_get_stuck_with_threaded_callbacks(threads):
     wait_until(lambda: not doc.callbacks.hold_value, timeout=5000)
 
 
+@pytest.mark.xdist_group(name="server")
+def test_hold_in_app_callable_does_not_leak_hold():
+    build = {}
+
+    def app():
+        with hold():
+            pass
+        # A hold leaked here defers the unhold past ServerSession
+        # construction, which registers this callback a second time
+        # when the queued SessionCallbackAdded event is replayed.
+        pn.state.add_periodic_callback(lambda: None, period=10000)
+        build['thread_id'] = state._thread_id
+        build['hold'] = state.curdoc.callbacks.hold_value
+        return IntSlider()
+
+    serve_and_request(app)
+
+    assert build['thread_id'] is not None
+    assert build['hold'] is None
+
+
 class _FakeProtocol:
     def create(self, msgtype, events):
         return object()


philippjfr (approved):

SimonHeybrock:
@philippjfr I am not sure why the CI run failed, could you have a look?

philippjfr (approved):

philippjfr:
Thanks @SimonHeybrock, test failure was just flaky, merging.

PR #8708, opened 2026-08-10
Inline versioned stylesheets on the notebook update path

### LINKED ISSUE #8707: config.inline leaves loading.css as a remote CDN import after the first render, because replace_inline_css keeps the ?v= query in the local path

<details>

```plaintext
panel        1.9.3
bokeh        3.9.2
pyviz_comms  3.0.6
python       3.12.13
OS           Linux (python:3.12-slim container, no network)
```

Also checked against `main` at `b6b1d99fb8bfd84e55c3499a475cbaa4b9f36fba`: `panel/io/notebook.py` is
unchanged between `v1.9.3` and that commit, so the code path below is the same on both.
</details>

With `pn.config.inline = True` in a notebook, Panel is expected to serve every one of its own
stylesheets from the local `panel/dist` directory, so a machine with no route to
`cdn.holoviz.org` still renders correctly. That is what #5167 asked for and what #7701 delivered.

It holds for the first render, but not for components created afterwards. Anything added to an
already rendered object goes out with `css/loading.css` still pointing at the CDN:

```
https://cdn.holoviz.org/panel/1.9.3/dist/css/loading.css?v=1.9.3
```

The cause is the version query. `patch_stylesheet` (`panel/io/resources.py:305-326`) appends
`?v={JS_VERSION}` to every URL it rewrites. `replace_inline_css`
(`panel/io/notebook.py:370-376`) then builds the local path by string-replacing the CDN prefix
and keeps that query in the filename:

```python
path = DIST_DIR / stylesheet.url.replace(CDN_DIST, '')  # -> 'css/loading.css?v=1.9.3'
if not path.exists():                                   # False, so the CDN URL is returned as-is
    return stylesheet
```

`panel/dist/css/loading.css` exists; `panel/dist/css/loading.css?v=1.9.3` does not. So Panel
produces the one URL shape its own inliner cannot resolve.

Why only after the first render: `_repr_mimebundle_` (`panel/viewable.py:887`) wraps
`_render_model` in `set_resource_mode('inline')`, so `get_dist_path()` returns `CDN_DIST`,
every branch of `patch_stylesheet` falls through, and no `?v=` is ever produced. Models built
outside that context manager get the ambient `server` resource mode instead, `get_dist_path()`
returns `LOCAL_DIST`, `patch_stylesheet` rewrites LOCAL to CDN and appends the version, and then
`panel/viewable.py:660` tries to inline the result and cannot.

The first block is the fault on its own, with no notebook needed:

```python
from bokeh.models import ImportedStyleSheet
from panel.io.notebook import replace_inline_css
from panel.io.resources import CDN_DIST, DIST_DIR, JS_VERSION

plain = f'{CDN_DIST}css/loading.css'
print(type(replace_inline_css(ImportedStyleSheet(url=plain))).__name__)
# InlineStyleSheet

versioned = f'{plain}?v={JS_VERSION}'
print(type(replace_inline_css(ImportedStyleSheet(url=versioned))).__name__)
# ImportedStyleSheet, i.e. still fetched from the CDN

print((DIST_DIR / 'css/loading.css').exists())                    # True
print((DIST_DIR / f'css/loading.css?v={JS_VERSION}').exists())    # False
```

And the same thing through the public API, on the update path a notebook actually uses:

```python
import panel as pn
import pyviz_comms
from bokeh.models import ImportedStyleSheet
from panel.io.resources import CDN_DIST, DIST_DIR

pn.config.inline = True
pn.extension()

col = pn.Column(pn.widgets.FloatSlider(name='a'))
doc, comm = Document(), pyviz_comms.Comm()
root = col.get_root(doc, comm)
pn.state._views[root.ref['id']] = (col, root, doc, comm)

col.append(pn.widgets.FloatSlider(name='b'))   # everything after the first render

for model in [root, *root.references()]:
    for sts in getattr(model, 'stylesheets', []) or []:
        if isinstance(sts, ImportedStyleSheet):
            print('still remote:', sts.url)
            print('  exists locally without the query:',
                  (DIST_DIR / sts.url.replace(CDN_DIST, '').split('?')[0]).exists())
```

No traceback: nothing raises, the stylesheet is silently left as a remote import. Output of the
second script, run in a container with `--network none`:

```
still remote: https://cdn.holoviz.org/panel/1.9.3/dist/css/loading.css?v=1.9.3
  exists locally without the query: True
```

On a `pn.Column(Tabulator)` the same probe inlines 12 stylesheets and leaves exactly that one.

I have not attached one. The fault is a resource that stays a network request rather than a
rendered difference on this machine, and its visible form (missing loading-spinner styling) only
appears in a live kernel that is genuinely firewalled, which I cannot capture honestly here. The
console output above is the whole observable difference I measured. Happy to add a rendered
capture if that would help.

#### Possible fix

Splitting the query off before the `exists()` check is enough, and it matches what
`panel/io/resources.py:432` already does in the same situation, stripping the query off a
URL-derived filename before testing it against the filesystem:

```python
path = DIST_DIR / stylesheet.url.replace(CDN_DIST, '').split('?')[0]
```

ebarkhordar:
## Description

Ran a Panel notebook on a machine with no route to `cdn.holoviz.org`, with `pn.config.inline =
True` set for exactly that reason. The first render was fine. Every component added afterwards
still asked the CDN for `css/loading.css`, so the loading indicator lost its styling and there
was no error anywhere to explain why.

`patch_stylesheet` (`panel/io/resources.py:305-326`) appends `?v={JS_VERSION}` to every URL it
rewrites. `replace_inline_css` then builds the local path by string-replacing the CDN prefix,
which carries that query into the filename:

```python
path = DIST_DIR / stylesheet.url.replace(CDN_DIST, '')  # 'css/loading.css?v=1.9.3'
if not path.exists():                                   # False
    return stylesheet                                   # returned as a remote import
```

`panel/dist/css/loading.css` exists, `panel/dist/css/loading.css?v=1.9.3` does not, so Panel
produces the one URL shape its own inliner cannot resolve. This only bites after the first
render: `_repr_mimebundle_` (`panel/viewable.py:887`) wraps `_render_model` in
`set_resource_mode('inline')`, so `patch_stylesheet` falls through and no `?v=` is ever
produced. Models built outside that context manager get the ambient `server` mode, pick up the
version suffix, and then `panel/viewable.py:660` cannot inline them.

The fix splits the query off before the `exists()` check, which is what
[`panel/io/resources.py:432`](https://github.com/holoviz/panel/blob/b6b1d99fb8bfd84e55c3499a475cbaa4b9f36fba/panel/io/resources.py#L432)
already does when it derives a filename from a URL to test against the filesystem.

Fixes #8707

### Reproducing it

```python
from bokeh.models import ImportedStyleSheet
from panel.io.notebook import replace_inline_css
from panel.io.resources import CDN_DIST, JS_VERSION

url = f'{CDN_DIST}css/loading.css'
print(type(replace_inline_css(ImportedStyleSheet(url=url))).__name__)
print(type(replace_inline_css(ImportedStyleSheet(url=f'{url}?v={JS_VERSION}'))).__name__)
```

Before: `InlineStyleSheet`, then `ImportedStyleSheet`. After: `InlineStyleSheet` both times.

Issue #8707 has the same thing through the public API, on the update path a notebook actually
uses. Run in a `--network none` container, it reports one stylesheet still fetched remotely
before the change and none after. On a `pn.Column(Tabulator)` the same probe inlines 12
stylesheets and leaves exactly that one.

I have not attached a screenshot. The change is a resource that stops being a network request,
and its visible form only appears in a live kernel that is genuinely firewalled, which I could
not capture honestly here. Glad to add a rendered capture if you would like one.

### Verification

- `test_replace_inline_css_ignores_version_query` fails on `main` (`assert isinstance(versioned,
  InlineStyleSheet)`) and passes with the change.
- `panel/tests/io/{test_notebook,test_resources,test_save}.py` on Python 3.12, panel 1.9.3
  environment: same 14 failures before and after, one additional pass (the new test). The
  failures are this container missing built JS bundles, not related to the change.
- `ruff check` and `isort --check-only` clean on both changed files.
- Not checked: a real browser render, and the Pyodide path, where `config.inline` is bypassed
  earlier and I did not exercise it.

## AI Disclosure

Usage: written by an autonomous agent running under this account. It located the fault, ran the
reproduction and the test suites quoted above in a container, and wrote the patch, the test and
this description.

## Checklist

### DIFF
--- a/panel/io/notebook.py
+++ b/panel/io/notebook.py
@@ -370,7 +370,7 @@ class Mimebundle:
 def replace_inline_css(stylesheet: ImportedStyleSheet):
     if not stylesheet.url.startswith(CDN_DIST):
         return stylesheet
-    path = DIST_DIR / stylesheet.url.replace(CDN_DIST, '')  # type: ignore
+    path = DIST_DIR / stylesheet.url.replace(CDN_DIST, '').split('?')[0]  # type: ignore
     if not path.exists():
         return stylesheet
     return InlineStyleSheet(css=path.read_text(encoding='utf-8'))
--- a/panel/tests/io/test_notebook.py
+++ b/panel/tests/io/test_notebook.py
@@ -5,8 +5,10 @@ pytest.importorskip("IPython")
 from bokeh.models import ImportedStyleSheet, InlineStyleSheet
 
 from panel.config import config, panel_extension
-from panel.io.notebook import ipywidget
-from panel.io.resources import CDN_ROOT, set_resource_mode
+from panel.io.notebook import ipywidget, replace_inline_css
+from panel.io.resources import (
+    CDN_DIST, CDN_ROOT, JS_VERSION, set_resource_mode,
+)
 from panel.pane import Str
 from panel.widgets import TextEditor
 
@@ -76,3 +78,12 @@ def test_notebook_inline_css_stylesheets(nb_loaded):
     model = list(widget._models.values())[0][0]
     for stylesheet in model.stylesheets[:len(model.__css__)]:
         assert isinstance(stylesheet, InlineStyleSheet)
+
+def test_replace_inline_css_ignores_version_query():
+    url = f'{CDN_DIST}css/loading.css'
+    unversioned = replace_inline_css(ImportedStyleSheet(url=url))
+    versioned = replace_inline_css(ImportedStyleSheet(url=f'{url}?v={JS_VERSION}'))
+
+    assert isinstance(unversioned, InlineStyleSheet)
+    assert isinstance(versioned, InlineStyleSheet)
+    assert versioned.css == unversioned.css


philippjfr:
Thank you @ebarkhordar!

ebarkhordar:
Thanks for the quick turnaround, Philipp. Flagging once more from the PR body: the verification was headless, not a real browser render, and the Pyodide path bypasses `config.inline` earlier so it is untouched by this change.

PR #8712, opened 2026-08-12
Update value_throttled when a slider value is set from Python (closes #2675)

### LINKED ISSUE #2675: Setting slider value programmatically does not update value_throttled parameter
panel 0.12.1
param 1.11.1
bokeh 2.3.3

This is a continuation of #1256.

When changing the `value` parameter of a slider from python _after_ instantiation, `value_throttled` is not updated. Then it's only when the user changes the slider value in the browser that the discrepancy between `value_throttled` and `value` disappears. I think all sliders are affected.

For instance with a FloatSlider, if you have the following:

```python
import panel as pn

slider = pn.widgets.FloatSlider(name='Slider', start=0, end=1, value=0, step=0.1)
slider.value = 1

print(slider.value_throttled)
```

you will get 0 instead of 1.

This can have detrimental consequences, because if the slider `value` is changed based on some event, you cannot reliably make things depend on the `value_throttled` parameter, which can be at odds with `value` and with the state of the widget as displayed in the browser.

Because of this, there are situations where you can't use `value_throttled` when it's needed, for instance as arguments of a DynamicMap when you have high computational costs associated with a change of parameters.

Kayvan-Zahiri:
Closes #2675.

### The bug

Setting a slider's `value` from Python leaves `value_throttled` stale forever, so
the two attributes disagree and every callback bound to `param.value_throttled`
silently never fires.

```python
s = pn.widgets.FloatSlider(start=0, end=1, value=0, step=0.1)
s.value = 1
s.value, s.value_throttled          # (1, 0)

seen = []
s2 = pn.widgets.IntSlider(start=0, end=10, value=0)
pn.bind(lambda v: seen.append(v), s2.param.value_throttled, watch=True)
s2.value = 7
seen                                # []   expected [7]
```

Verified on panel 1.8.7 across all four families: `FloatSlider`, `IntSlider`,
`DiscreteSlider` (3 -> throttled 1) and `RangeSlider` ((2, 5) -> throttled (0, 1)).

The reporter's case is a `DynamicMap` driven by `value_throttled` that quietly
renders against the previous value. There is no error, so it looks like it works.

### Why this is a bug and not the intended contract

`_SliderBase.__init__` already does this:

```python
if 'value' in params and 'value_throttled' in self.param:
    params['value_throttled'] = params['value']
```

So the invariant "value_throttled tracks value unless a drag is in progress" is
already established at construction. A later programmatic assignment just fails
to maintain it. Nothing is being throttled when Python sets the value, because
there is no handle being dragged.

### The fix

A watcher on `value` that syncs `value_throttled`, skipping changes that came
from the frontend. Panel applies frontend property changes inside param's
`_syncing` context (`reactive.py`, `_process_events`), so `_param__private.syncing`
distinguishes the two cases. A drag still leaves `value_throttled` pinned until
the handle is released.

The watcher is appended to `_internal_callbacks`, matching `input.py` and
`codeeditor.py`. Without that, `test_widget_untracked_watchers` fails for eight
slider types. I hit exactly that and it is why the registration is explicit.

### One thing I would like your call on

The guard reads `self._param__private.syncing`, which is param-internal. Panel
already imports `param.parameterized._syncing` in `reactive.py` and `custom.py`,
so this is consistent with existing usage, but if you would rather I set an
explicit flag on the widget in `_process_events` instead, say so and I will
change it.

### Verification

- `panel/tests/widgets/test_slider.py`: 67 passed.
- `panel/tests/widgets/`: 3 failed, 890 passed, identical to an unmodified
  checkout. The 3 pre-existing failures are async tests that need a plugin the
  local environment lacks.
- Four of the five new tests fail on `main`. The fifth,
  `test_slider_value_throttled_not_updated_while_dragging`, passes both before
  and after by design: it pins the drag behaviour that must not change.

### DIFF
--- a/panel/tests/widgets/test_slider.py
+++ b/panel/tests/widgets/test_slider.py
@@ -10,9 +10,9 @@ from bokeh.models import (
 
 from panel import config
 from panel.widgets import (
-    DateRangeSlider, DateSlider, DatetimeRangeSlider, DatetimeSlider,
-    DiscreteSlider, EditableFloatSlider, EditableIntSlider,
-    EditableRangeSlider, FloatSlider, IntSlider, RangeSlider, StaticText,
+    DateRangeSlider, DateSlider, DatetimeRangeSlider, DatetimeSlider, DiscretePlayer,
+    DiscreteSlider, EditableFloatSlider, EditableIntSlider, EditableRangeSlider, FloatInput,
+    FloatSlider, IntInput, IntSlider, Player, RangeSlider, StaticText,
 )
 
 
@@ -1011,3 +1011,18 @@ def test_slider_value_throttled_not_updated_while_dragging(document, comm):
 
     slider._process_events({'value_throttled': 4.0})
     assert slider.value_throttled == 4.0
+
+
+@pytest.mark.parametrize('widget, kwargs, new', [
+    (Player, dict(start=0, end=10, value=1), 4),
+    (DiscretePlayer, dict(options=[1, 2, 3], value=1), 3),
+    (IntInput, dict(value=1), 7),
+    (FloatInput, dict(value=1.0), 7.0),
+])
+def test_non_slider_value_throttled_follows_programmatic_value(widget, kwargs, new):
+    # PlayerBase and _SpinnerBase seed value_throttled from value in __init__ the
+    # same way _SliderBase does, so they had the same stale-forever bug. The sync
+    # lives on Widget so every family that declares value_throttled gets it.
+    w = widget(**kwargs)
+    w.value = new
+    assert w.value_throttled == new
--- a/panel/tests/widgets/test_slider.py
+++ b/panel/tests/widgets/test_slider.py
@@ -10,9 +10,10 @@ from bokeh.models import (
 
 from panel import config
 from panel.widgets import (
-    DateRangeSlider, DateSlider, DatetimeRangeSlider, DatetimeSlider, DiscretePlayer,
-    DiscreteSlider, EditableFloatSlider, EditableIntSlider, EditableRangeSlider, FloatInput,
-    FloatSlider, IntInput, IntSlider, Player, RangeSlider, StaticText,
+    DateRangeSlider, DateSlider, DatetimeRangeSlider, DatetimeSlider,
+    DiscretePlayer, DiscreteSlider, EditableFloatSlider, EditableIntSlider,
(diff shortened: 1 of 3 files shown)


ebarkhordar (commented):
This closes #2675, but the players have the same gap: the watcher is registered in `_SliderBase.__init__`, and `PlayerBase` inherits `Widget`, so `Player` and `DiscretePlayer` are untouched. On panel 1.9.3, `Player(start=0, end=10, value=1)` then `p.value = 4` leaves `value_throttled` at 1, and `DiscretePlayer` behaves the same. TheoMathurin raised the player case in #2675 itself (2022-10-06), and `PlayerBase.__init__` already seeds `value_throttled` from `value` exactly like the slider does. Widen this PR, or keep #2675 open for the players?

Kayvan-Zahiri:
Thanks for checking the players, that's exactly the case I'd want someone to poke at. I think they're already covered though.

The watcher isn't in `_SliderBase.__init__`, it's in `Widget.__init__` in `panel/widgets/base.py`, gated on whether the widget declares the parameter at all:

```python
if 'value_throttled' in self.param:
    self._internal_callbacks.append(
        self.param.watch(self._sync_value_throttled, 'value')
    )
```

`PlayerBase` inherits `Widget`, so it picks this up without needing anything player-specific. On this branch:

```
Player(start=0, end=10, value=1);  p.value = 4  ->  value_throttled == 4
DiscretePlayer(options=[1,2,3], value=1); d.value = 3  ->  value_throttled == 3
```

Both are stale on main and both follow here. There's a test for it too, `test_non_slider_value_throttled_follows_programmatic_value`, which parametrizes over `Player`, `DiscretePlayer`, `IntInput` and `FloatInput`. I included the spinners because `_SpinnerBase` seeds `value_throttled` in `__init__` the same way, so it had the same gap.

That's also why the fix went on `Widget` instead of the slider base: TheoMathurin's point in #2675 is that this isn't really a slider bug, it's every family that declares `value_throttled`.

If you were testing against released 1.9.3 rather than this branch, that would explain what you saw. Let me know if there's a case I've still missed and I'll add it.

philippjfr:
@Kayvan-Zahiri Thanks for the fix, hope you don't mind, I pushed a small fix for CompositeWidget types.

philippjfr:
Test failures unrelated.

PR #8719, opened 2026-08-17
Fix Tabulator 'starts' and 'ends' header filters

### LINKED ISSUE #8717: Tabulator 'starts' and 'ends' header filters raise AttributeError

```
panel  1.9.4rc3 (also reproduces on 1.8.9)
pandas 2.x
python 3.12
```

The `starts` and `ends` header filter types should filter the table the way Tabulator's client side filters do. Instead both raise `AttributeError` for any value, so neither filter type has ever worked.

The cause is that `_get_header_filters` calls `col.str.startsWith(val)` and `col.str.endsWith(val)`. Those are JavaScript method names. pandas spells them `startswith` and `endswith`.

This dates back to #2815 (October 2021), which added header filters. There is no test covering either operator, which is why it went unnoticed.

```python
import pandas as pd
import panel as pn

df = pd.DataFrame({'word': ['alpha', 'alphabet', 'beta']})
table = pn.widgets.Tabulator(df, header_filters=True)

table.filters = [{'field': 'word', 'type': 'starts', 'value': 'al'}]
table.current_view
```

#### Stack Traceback

```
  File "panel/widgets/tables.py", line 611, in _get_header_filters
    filters.append(col.str.startsWith(val))
                   ^^^^^^^^^^^^^^^^^^
AttributeError: 'StringMethods' object has no attribute 'startsWith'. Did you mean: 'startswith'?
```

The same happens for `type: 'ends'` at line 613.

ghostiee-11:
The `starts` and `ends` header filters called `col.str.startsWith` and `col.str.endsWith`, which are JavaScript names, so both raised `AttributeError` for every value. Now using the pandas spelling, matched case insensitively to mirror Tabulator's own client side filters so local and remote pagination agree.

Fixes #8717

### DIFF
--- a/panel/tests/widgets/test_tables.py
+++ b/panel/tests/widgets/test_tables.py
@@ -3072,6 +3072,47 @@ def test_header_filters_categorial_dtype():
     widget.filters = [{'field': 'model', 'type': 'like', 'value': 'A'}]
     assert widget.current_view.size == 1
 
+
+@pytest.fixture
+def df_words():
+    return pd.DataFrame({'word': ['alpha', 'alphabet', 'beta', 'Alphanumeric']})
+
+
+@pytest.mark.parametrize('pagination', ['local', 'remote', None])
+def test_header_filter_starts(df_words, pagination):
+    widget = Tabulator(df_words, header_filters=True, pagination=pagination)
+
+    widget.filters = [{'field': 'word', 'type': 'starts', 'value': 'bet'}]
+
+    assert list(widget.current_view['word']) == ['beta']
+
+
+@pytest.mark.parametrize('pagination', ['local', 'remote', None])
+def test_header_filter_starts_is_case_insensitive(df_words, pagination):
+    widget = Tabulator(df_words, header_filters=True, pagination=pagination)
+
+    widget.filters = [{'field': 'word', 'type': 'starts', 'value': 'ALPHA'}]
+
+    assert list(widget.current_view['word']) == ['alpha', 'alphabet', 'Alphanumeric']
+
+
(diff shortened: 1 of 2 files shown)


panel/widgets/tables.py
```diff
@@ -608,9 +608,9 @@ def _get_header_filters(self, df: pd.DataFrame) -> list[pd.Series | np.ndarray]:
             elif op == 'like':
                 filters.append(col.str.contains(val, case=False, regex=False))
             elif op == 'starts':
-                filters.append(col.str.startsWith(val))
+                filters.append(col.str.lower().str.startswith(val.lower()))
```
hoxbro, line 611: I'm not sure if is correct to lower the string for comparison.

ghostiee-11: Went with this because Tabulator's own `starts` filter lowercases both sides. Python only filters when `pagination='remote'`, so case sensitive here would quietly disagree with what the JS path shows for the same input.

panel/tests/widgets/test_tables.py
```diff
@@ -3072,6 +3072,64 @@ def test_header_filters_categorial_dtype():
     widget.filters = [{'field': 'model', 'type': 'like', 'value': 'A'}]
     assert widget.current_view.size == 1
 
+
+@pytest.fixture
+def df_words():
+    return pd.DataFrame({'word': ['alpha', 'alphabet', 'beta', 'Alphanumeric']})
```
hoxbro: I would just inline this.

ghostiee-11: Sure

hoxbro (approved):

PR #8720, opened 2026-08-17
Raise a clear type error for unsupported Perspective objects

### LINKED ISSUE #8718: Perspective raises AttributeError instead of a type error for unsupported objects

```
panel  1.9.4rc3
python 3.12
```

Passing an unsupported object to a pane normally raises a clear error:

```
ValueError: DataFrame pane does not support objects of type 'DataFrame'.
```

`Perspective` instead crashes inside `deconstruct_pandas` with an `AttributeError` about a pandas attribute the object does not have. The message gives no indication that the object type is the problem.

`Perspective.applies()` correctly returns `False` for these objects, so the check exists. It just never runs. `Perspective` mixes in `ReactiveData`, whose `__init__` builds the column data inside `super().__init__()`, and that happens before `PaneBase.__init__` reaches its `applies()` check:

```
panel/pane/perspective.py:359  __init__()
panel/pane/base.py:323         __init__()
panel/pane/base.py:159         __init__()          <- super() call
panel/reactive.py:1044         __init__()          <- ReactiveData builds data here
panel/reactive.py:1098         _update_cds()
panel/pane/perspective.py:370  _get_data()
panel/pane/perspective.py:102  deconstruct_pandas()
```

Assigning an unsupported object to `.object` after construction reaches the same crash, and that path has no validation at all.

```python
import panel as pn

pn.pane.Perspective('not tabular')
```

and, for the assignment path:

```python
import pandas as pd
import panel as pn

p = pn.pane.Perspective(pd.DataFrame({'a': [1]}))
p.object = 'not tabular'
```

#### Stack Traceback

```
  File "panel/pane/perspective.py", line 371, in _get_data
    ncols = len(df.columns)
                ^^^^^^^^^^
AttributeError: 'str' object has no attribute 'columns'
```

`Perspective` is the only pane that mixes in `ReactiveData`, so as far as I can tell no other pane is affected.

ghostiee-11:
`Perspective` mixes in `ReactiveData`, which builds the data inside `super().__init__()`, before `PaneBase` reaches its `applies()` check, so an unsupported object crashed with an `AttributeError` instead of the standard "does not support objects of type" error. Guard added in `_get_data`, which both the construction and the `.object` assignment paths go through.

Fixes #8718

### DIFF
--- a/panel/pane/perspective.py
+++ b/panel/pane/perspective.py
@@ -363,6 +363,11 @@ class Perspective(ModelPane, ReactiveData):
     def _get_data(self):
         if self.object is None:
             return {}, {}
+        # ReactiveData builds the data during super().__init__, before PaneBase
+        # gets to validate the object, so the type check has to happen here to
+        # cover both construction and later assignment to .object.
+        if self.applies(self.object) is False:
+            self._type_error(self.object)
         if isinstance(self.object, dict):
             ncols = len(self.object)
             df = data = self.object
--- a/panel/tests/pane/test_perspective.py
+++ b/panel/tests/pane/test_perspective.py
@@ -1,3 +1,5 @@
+import pytest
+
 from panel.pane import Perspective
 
 data = {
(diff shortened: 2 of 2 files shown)


philippjfr:
Thank you, test failures were unrelated.

Issue #8721, opened 2026-08-17
RFC: DataFrame backend parity for Panel's tabular components via Narwhals


ghostiee-11:
### Summary

Panel already depends on `narwhals >=2`, and Bokeh already accepts Polars and PyArrow frames as `ColumnDataSource` input. Panel's tabular components accept neither. I would like to close that gap, and I have done enough analysis to know where the hard part is, which is not rendering.

I am asking for agreement on three design questions before writing anything substantial.

### Current state

Run against `main` today:

```python
import pandas as pd, polars as pl, pyarrow as pa, panel as pn
from bokeh.models import ColumnDataSource

data = {'a': [1, 2, 3], 'b': ['x', 'y', 'z']}
for backend, df in [('pandas', pd.DataFrame(data)),
                    ('polars', pl.DataFrame(data)),
                    ('pyarrow', pa.table(data))]:
    for label, fn in [('bokeh CDS.from_df',    ColumnDataSource.from_df),
                      ('panel Tabulator',      pn.widgets.Tabulator),
                      ('panel DataFrame pane', pn.pane.DataFrame),
                      ('panel Perspective',    pn.pane.Perspective)]:
        try:
            fn(df); print(f'OK    {label:22} {backend}')
        except Exception as e:
            print(f'FAIL  {label:22} {backend}  {type(e).__name__}')
```

| Entry point | pandas | polars | pyarrow | Fails at |
| --- | --- | --- | --- | --- |
| `ColumnDataSource.from_df` | OK | OK | OK | |
| `widgets.Tabulator` | OK | FAIL | FAIL | `widgets/tables.py:757` |
| `widgets.DataFrame` | OK | FAIL | FAIL | `widgets/tables.py:757` |
| `pane.DataFrame` | OK | FAIL | FAIL | `pane/base.py:211` |
| `pane.Perspective` | OK | FAIL | FAIL | `pane/perspective.py:102` |

Bokeh passes 4 of 4 non pandas cases. Panel passes 0 of 8. Identical on the released 1.8.9, so this is not a recent regression.

PyArrow fails in exactly the same places as Polars, which suggests these are general pandas API assumptions rather than anything Polars specific.

### Why this seems worth doing now

* Bokeh completed its side in bokeh#13780 (closed as completed, February 2025). `ColumnDataSource._data_from_df` already imports `narwhals.stable.v1` and branches on `is_pandas_like_dataframe`, so the serialization layer beneath Panel is already backend neutral.
[...]

PR #8723, opened 2026-08-17
Accept DataFrames without a row index in the table widgets

ghostiee-11:
Panel already depends on narwhals and Bokeh already accepts any narwhals compatible frame as `ColumnDataSource` input, but Panel's tabular components are pandas only. Adds an internal seam and makes `Tabulator` and `DataFrame` construct from Polars and PyArrow, with every new branch behind a `has_index()` check so the pandas paths are untouched.

Read path only, rendering still needs backend neutral dtype handling and stays a strict xfail. Part of #8721.


panel/util/dataframe.py
```diff
@@ -0,0 +1,74 @@
+"""
```
philippjfr, line 1: Can we move this into `panel/utils`?

PR #8726, opened 2026-08-18
Fix Tabulator keywords header filter with header_filters=True

### LINKED ISSUE #8725: Tabulator 'keywords' header filter raises AttributeError when header_filters=True

```
panel  1.9.4 (main @ f981408cc)
python 3.12
```

The `keywords` header filter raises `AttributeError` when `header_filters=True`.

`_get_header_filters` reads the per column filter definition with:

```python
filt_def = getattr(self, 'header_filters', {}) or {}
```

`header_filters` is declared as `param.ClassSelector(class_=(bool, dict))`, so when it is `True` that expression evaluates to `True` rather than a dict, and the `keywords` branch then calls `.get` on a bool:

```python
match_all = filt_def.get(col_name, {}).get('matchAll', False)
sep = filt_def.get(col_name, {}).get('separator', ' ')
```

Only the `keywords` branch reads `filt_def`, which is why the other filter types are unaffected.

Passing `header_filters` as a dict works, so this only affects the boolean form.

```python
import pandas as pd
import panel as pn

table = pn.widgets.Tabulator(
    pd.DataFrame({'w': ['alpha', 'beta']}), header_filters=True
)
table.filters = [{'field': 'w', 'type': 'keywords', 'value': 'al ph'}]
table.current_view
```

Works as expected with the dict form:

```python
table = pn.widgets.Tabulator(
    pd.DataFrame({'w': ['alpha', 'beta']}),
    header_filters={'w': {'type': 'input', 'func': 'keywords'}},
)
table.filters = [{'field': 'w', 'type': 'keywords', 'value': 'al'}]
table.current_view   # ['alpha']
```

#### Stack Traceback

```
  File "panel/widgets/tables.py", line 615, in _get_header_filters
    match_all = filt_def.get(col_name, {}).get('matchAll', False)
                ^^^^^^^^^^^^
AttributeError: 'bool' object has no attribute 'get'
```

I found this while adding backend parametrised filter tests. There is no existing test for `keywords` combined with `header_filters=True`, which is why it has gone unnoticed.

ghostiee-11:
`header_filters` is a `ClassSelector(class_=(bool, dict))`, and `getattr(self, 'header_filters', {}) or {}` leaves it as `True` for the boolean form, so the keywords branch raised `AttributeError` when it read options off it. It is the only branch that reads that config, which is why no other filter type was affected.

Fixes #8725


ebarkhordar (commented):
The bool guard is right. The same expression has a second shape: `_get_filter_spec` accepts a plain string per column (`filter_type = self.header_filters[column.field]`, then `isinstance(filter_type, dict)`), so `header_filters={'word': 'input'}` is valid config and `filt_def.get(col_name, {})` returns a str.

On `71acd0f5`, panel 1.9.4, with `filters=[{'field': 'word', 'type': 'keywords', 'value': 'lph bet'}]`: `True` and the dict form both filter correctly now, `{'word': 'input'}` still raises `AttributeError: 'str' object has no attribute 'get'`. Normalising the per column entry rather than the container covers all three:

```python
col_def = filt_def.get(col_name)
if not isinstance(col_def, dict):
    col_def = {}
```

Only reachable by setting `filters` directly, so not a blocker.

The red 3.14 legs are the `test_vtk.py` NumPy 2.5 failures that #8729 fixed on main, not your change.

PR #8731, opened 2026-08-19
Make Tabulator edits more robust

### LINKED ISSUE #3620: TableEditEvent when losing focus on a NaN date cell being
Opening the date widget of a Tabulator date cell and hitting `Escape` when that cell value is `NaN` triggers a `TableEditEvent` while it shouldn't:

issue_tabulator_date_editevent

```python
import numpy as np
import pandas as pd
import panel as pn

pn.extension('tabulator')

df = pd._testing.makeMixedDataFrame()
df.at[0, 'D'] = np.nan

receiver = pn.pane.Str('Edit Event Log:\n\n', width=400)
t = pn.widgets.Tabulator(df, widths=100)
i = 0

def callback(e):
    global i
    receiver.object += f'Edit Event #{i} {e.value}\n'
    i += 1

t.on_edit(callback)
pn.Row(t, receiver).servable()
```

philippjfr:
`on_edit` callbacks could intermittently report a stale or incorrect value/old pair (e.g. old == value for an edit that actually changed the data), flaky on CI and worse on some bokeh versions but hard to reproduce locally. Editing a row so its new value no longer matched an active header filter could also silently corrupt a sibling row on the same page under remote pagination.

Root cause: the frontend sent a TableEditEvent immediately before and after patching the ColumnDataSource, and _process_event derived the reported value/old by reading self.value/self._old_value back off the widget. That patch and that event are dispatched via independent, asynchronously-scheduled paths on the server, so there was no guarantee the patch had actually been applied by the time the event was processed, the callback could fire before or after the data actually landed, depending on scheduling. Bokeh's own event-payload serialization only auto-unpacks certain JS constructs, and none of the raw values were dtype-converted, so a naive "just send the value from JS" fix would silently break numeric/date columns.

## Fix

- panel/models/tabulator.ts / panel/models/tabulator.py — TableEditEvent now
  carries the raw value/old straight from the edited Tabulator cell, captured
  at edit time. A JS-side stash (_pending_old_values) works around Tabulator's
  own numeric-empty-to-NaN coercion re-triggering the edit handler and
  clobbering its oldValue bookkeeping before it can be read.
- panel/widgets/tables.py — _process_event now converts that raw value/old
  through _convert_column (the same dtype/NaN coercion _process_data uses)
  instead of reading self.value, so the reported values are correctly typed
  without depending on the patch having landed.
- Simplified the message flow: rather than sending a paired pre/post event
  around the ColumnDataSource.patch() call, the frontend now sends a single
  event before the patch. The Python side records it as a pending edit
  (self._pending_edits, keyed by column + row) and only fires on_edit from
  _update_column — the exact code path that mutates self.value/current_view —
  once that value has actually landed. This removes the two-message dance
  entirely and ties on_edit to the data mutation as a guaranteed side effect
  rather than a message racing against it.
- Fixed an inverted condition in the header-filter/_edited_indexes bookkeeping
  that could desync the row range _update_column writes from the
  (filter-masked) array _process_data hands it, corrupting a neighboring row's
  value.
- As a side effect of only firing on_edit when a value actually changes, this
  also fixes #3620: clicking away from an unedited NaN datetime cell no longer
  fires a spurious edit event (removed the xfail on
  test_tabulator_editor_datetime_nan).
- Added a defensive guard against a malformed/valueless edit event (e.g. one
  constructed without value/old) so it's a no-op rather than a crash.

## AI Disclosure

Fixed with the help of Claude Sonnet 5

### DIFF
--- a/panel/models/tabulator.ts
+++ b/panel/models/tabulator.ts
@@ -22,12 +22,12 @@ import {schedule_when, transformJsPlaceholders} from "./util"
 import tabulator_css from "styles/models/tabulator.css"
 
 export class TableEditEvent extends ModelEvent {
-  constructor(readonly column: string, readonly row: number, readonly pre: boolean) {
+  constructor(readonly column: string, readonly row: number, readonly pre: boolean, readonly value?: unknown, readonly old?: unknown) {
     super()
   }
 
   protected override get event_values(): Attrs {
-    return {model: this.origin, column: this.column, row: this.row, pre: this.pre}
+    return {model: this.origin, column: this.column, row: this.row, pre: this.pre, value: this.value, old: this.old}
   }
 
   static {
@@ -376,6 +376,7 @@ export class DataTabulatorView extends HTMLBoxView {
   columns: Map<string, any> = new Map()
   container: HTMLDivElement | null = null
   _tabulator_cell_updating: boolean=false
+  _pending_old_values: Map<string, any> = new Map()
   _updating_page: boolean = false
   _updating_expanded: boolean = false
   _updating_sort: boolean = false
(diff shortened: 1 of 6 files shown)


panel/models/tabulator.ts
```diff
@@ -1619,20 +1620,36 @@ export class DataTabulatorView extends HTMLBoxView {
...
+    if (this._pending_old_values.has(key)) {
+      old_value = this._pending_old_values.get(key)
+      this._pending_old_values.delete(key)
+    }
+    // value/old are sent explicitly rather than left for Python to read back
+    // from self.value once the ColumnDataSource patch below lands: that
+    // patch is applied via a different, independently-scheduled server-side
+    // path than this event, so there is no ordering guarantee between them.
+    // The event is still sent first so the Python side can record it as a
+    // pending edit before the patch (which it keys off of) arrives.
```
philippjfr:
```suggestion
```

PR #8733, opened 2026-08-21
Fix conda build issue

philippjfr:
For making dev releases the conda build test step may need packages from the `pyviz/label/dev` channel. This ensures that during a dev release the channel is added.


harunqltkn:
scripts gg ez win w l f

ü

PR #8736, opened 2026-08-24
Clean up GridSpec children when deleting cells

### LINKED ISSUE #8710: GridSpec never cleans up removed children: in-place objects mutation defeats the diff in _get_objects
#### Description

Removing or replacing a child of a rendered `GridSpec` never calls the child's `_cleanup(root)`. For panes with document/stream hooks this leaks live objects: a `pn.pane.HoloViews` wrapping a stream-driven `DynamicMap` keeps its rendered plot subscribed to the stream, so each slot replacement permanently adds one more plot doing full render work on every stream event.

Cause: `GridSpec.__setitem__`/`__delitem__` mutate `self.objects` **in place** and then `self.param.trigger('objects')`. The removed-child cleanup in `GridSpec._get_objects` diffs `old_objects` against the current objects — but after an in-place mutation the event's old value is the same dict, so the diff is always empty and no removed child is ever cleaned. Other layouts reassign `objects` and are unaffected (control below). Present in 1.9.3; the mutation pattern is unchanged on `main`.

#### Complete, minimal, self-contained example

```python
import holoviews as hv, panel as pn
hv.extension('bokeh')

pipe = hv.streams.Pipe(data=hv.Curve([1, 2, 3]))
dmap = hv.DynamicMap(lambda data: hv.Curve([]), streams=[pipe])

grid = pn.GridSpec()
grid[0, 0] = pn.pane.HoloViews(dmap)
root = grid.get_root()
print(len(pipe.subscribers))  # 1

del grid[0, 0]
print(len(pipe.subscribers))  # 1  <- removed pane was never cleaned up

grid[0, 0] = pn.pane.HoloViews(dmap)
print(len(pipe.subscribers))  # 2  <- one leaked live plot per replacement
```

Control showing `Column` cleans up correctly (only `GridSpec` is affected):

```python
pipe2 = hv.streams.Pipe(data=hv.Curve([1, 2, 3]))
dmap2 = hv.DynamicMap(lambda data: hv.Curve([]), streams=[pipe2])
col = pn.Column(pn.pane.HoloViews(dmap2))
root2 = col.get_root()
print(len(pipe2.subscribers))  # 1
col[:] = []
print(len(pipe2.subscribers))  # 0
```

Versions: panel 1.9.3, bokeh 3.9.2, holoviews 1.23.1, Python 3.12, Linux.

Found while chasing a production leak (scipp/esslivedata#1224): each rebuild of a grid cell hosting a live plot permanently doubled the render work for that cell's streams.

Related: holoviz/holoviews#6988 — `Plot.cleanup()` currently severs all weakly-wrapped subscribers on shared streams, which matters for how a fix orders cleanup vs. rendering of a replacement child.

EltonChang1:
## Summary

`GridSpec.__delitem__` now assigns a copied `objects` mapping instead of mutating the current mapping in place. The parameter event therefore retains the old children, allowing the existing diff logic to clean up removed pane models.

The replacement path on current `main` already assigns a new mapping; this closes the corresponding deletion path and adds a regression test for model cleanup.

Fixes #8710

## Testing

- `pytest panel/tests/layout/test_grid.py -q` — 22 passed
- `ruff check panel/layout/grid.py panel/tests/layout/test_grid.py`
- `isort --check-only panel/layout/grid.py panel/tests/layout/test_grid.py`

### DIFF
--- a/panel/layout/grid.py
+++ b/panel/layout/grid.py
@@ -466,10 +466,10 @@ class GridSpec(Panel):
             deleted = dict([list(o)[0] for o in subgrid.flatten()])
         else:
             deleted = [list(subgrid)[0][0]]
-        new_objects = dict(self.objects)
-        for key in deleted:
-            del new_objects[key]
-        self.objects = new_objects
+        self.objects = {
+            k: v for k, v in self.objects.items()
+            if k not in deleted
+        }
 
     def __getitem__(self, index):
         if isinstance(index, tuple):
--- a/panel/layout/grid.py
+++ b/panel/layout/grid.py
@@ -466,9 +466,10 @@ class GridSpec(Panel):
             deleted = dict([list(o)[0] for o in subgrid.flatten()])
         else:
             deleted = [list(subgrid)[0][0]]
+        new_objects = dict(self.objects)
         for key in deleted:
-            del self.objects[key]
-        self.param.trigger('objects')
+            del new_objects[key]
+        self.objects = new_objects
 
     def __getitem__(self, index):
         if isinstance(index, tuple):
--- a/panel/tests/layout/test_grid.py
+++ b/panel/tests/layout/test_grid.py
@@ -23,6 +23,20 @@ def test_gridspec_cleanup(document, comm):
     assert ref not in spacer._models
 
 
+def test_gridspec_delete_cleans_up_removed_pane(document, comm):
+    spacer = Spacer()
+    gspec = GridSpec()
+    gspec[0, 0] = spacer
+
+    model = gspec.get_root(document, comm)
+    ref = model.ref['id']
+
+    del gspec[0, 0]
+
+    assert ref not in spacer._models
+    assert model.children == []
+
+
 def test_gridspec_integer_setitem():
     div = Div()
     gspec = GridSpec()


philippjfr (approved):

panel/layout/grid.py
```diff
@@ -466,9 +466,10 @@ def __delitem__(self, index):
             deleted = dict([list(o)[0] for o in subgrid.flatten()])
         else:
             deleted = [list(subgrid)[0][0]]
+        new_objects = dict(self.objects)
         for key in deleted:
-            del self.objects[key]
-        self.param.trigger('objects')
+            del new_objects[key]
```
philippjfr:
```suggestion
        new_objects = {k: v for k, v in self.objects.items() if key not in deleted}
```

panel/layout/grid.py
```diff
@@ -466,9 +466,10 @@ def __delitem__(self, index):
...
+        self.objects = new_objects
```
philippjfr:
```suggestion
        self.objects = {
            k: v for k, v in self.objects.items()
            if key not in deleted
        }
```

philippjfr:
```suggestion
        self.objects = {
             k: v for k, v in self.objects.items()
             if k not in deleted
         }
```

PR #8743, opened 2026-08-27
Fix DatetimeSlider same-day bounds

### LINKED ISSUE #8742: DatetimeSlider / DateSlider / DatetimeRangeSlider render an empty container (no slider track) when served — no console errors
## Description

`DatetimeSlider`, `DateSlider` and `DatetimeRangeSlider` render an **empty container** when served via `pn.serve` (Bokeh server): the widget's DOM node exists and the slider title/label renders, but the slider track itself never appears — there is nothing to drag. `IntSlider` / `DiscreteSlider` render fine on the same page.

No JS console errors are logged; BokehJS reports `Bokeh items were rendered successfully`.

## Versions tested (all reproduce)

| panel | bokeh | python | OS |
|-------|-------|--------|-----|
| 1.9.4 | 3.8.1 | 3.12 | macOS arm64 |
| 1.9.4 | 3.9.2 | 3.12 | macOS arm64 |
| 1.8.4 | 3.8.1 | 3.12 | macOS arm64 |

(panel 1.9.4 requires `bokeh>=3.7.0,<3.10.0`, so bokeh 3.10 is unreachable with current panel.)

## Minimal repro

```python
from datetime import datetime

import panel as pn

pn.extension()
slider = pn.widgets.DatetimeSlider(
    name="t",
    start=datetime(2026, 8, 21, 8, 45),
    end=datetime(2026, 8, 21, 13, 30),
    value=datetime(2026, 8, 21, 8, 45),
)
pn.serve(slider, port=5203, show=True)
```

## Verification (browser devtools)

```js
// shadow-DOM piercing query:
const collect = (root, out) => {
  for (const e of root.querySelectorAll("*")) { out.push(e); if (e.shadowRoot) collect(e.shadowRoot, out); }
  return out;
};
const dt = collect(document, []).find(e => typeof e.className === "string" && e.className.includes("-DatetimeSlider"));
dt.children.length   // → 0
dt.outerHTML.length  // → 66 (just the empty <div class="bk-panel-models-datetime_slider-DatetimeSlider"></div>)
```

The slider title element (`bk-slider-title`) does render, which makes the widget look present — but the noUi track/handles are never attached, so it cannot be interacted with.

## Expected behavior

An interactive date slider, as `IntSlider`/`DiscreteSlider` behave on the same page.

## Workaround

`DiscreteSlider` with string options (e.g. `HH:MM` labels mapped to timestamps) works correctly.

EltonChang1:
## Description

Fixes #8742.

`DatetimeSlider` converted its value to a millisecond timestamp but left its
`start` and `end` bounds as Python datetimes. Bokeh's datetime property then
normalized same-day bounds to the same midnight value, producing an invalid
slider model with equal bounds.

This converts datetime-valued `start`, `end`, and `value` fields consistently
and adds a regression test for same-day bounds with non-midnight times.

## Validation

- `python3 -m ruff check panel/widgets/slider.py panel/tests/widgets/test_slider.py`
- `.venv312/bin/python -m pytest panel/tests/widgets/test_slider.py -q` (63 passed)
- `git diff --check`

## Limitations

The local in-app browser blocked capture of the localhost before/after page.
The regression test validates the underlying rendered Bokeh model values: the
two datetime bounds remain distinct millisecond timestamps instead of both
becoming midnight.

## AI Disclosure

- **Tool & Model:** AI coding assistant (GPT-5)
- **Usage:** Used to investigate the reported serialization path, implement the
  focused conversion change, and draft the regression test and PR text.
- [x] I tested the contribution and reviewed the resulting diff.
- [x] I take responsibility for the contribution.

## Checklist

- [x] The contribution is linked to an existing issue.
- [x] Tests cover the behavioral change.
- [x] No documentation change is needed for this bug fix.
- [x] No unrelated generated files are included.

### DIFF
--- a/panel/tests/widgets/test_slider.py
+++ b/panel/tests/widgets/test_slider.py
@@ -253,6 +253,18 @@ def test_datetime_slider(document, comm, value, start, end):
         assert widget.value == 1620777600000
 
 
+def test_datetime_slider_preserves_time_in_bounds(document, comm):
+    start = datetime(2026, 8, 21, 8, 45)
+    end = datetime(2026, 8, 21, 13, 30)
+    datetime_slider = DatetimeSlider(value=start, start=start, end=end)
+
+    widget = datetime_slider.get_root(document, comm=comm)
+
+    epoch = datetime(1970, 1, 1)
+    assert widget.start == (start - epoch).total_seconds() * 1000
+    assert widget.end == (end - epoch).total_seconds() * 1000
+
+
 def test_datetime_slider_np_datetime64(document, comm):
     start = np.datetime64('2018-09-01')
     end = np.datetime64('2018-09-10')
--- a/panel/widgets/slider.py
+++ b/panel/widgets/slider.py
@@ -306,11 +306,13 @@ class DateSlider(_SliderBase):
 
     def _process_param_change(self, params: dict[str, t.Any]) -> dict[str, t.Any]:
         props = super()._process_param_change(params)
-        if 'value' in props:
-            value = props['value']
+        for key in ('value', 'start', 'end'):
+            if key not in props:
+                continue
+            value = props[key]
             if isinstance(value, dt.datetime):
                 value = datetime_as_utctimestamp(value)
-            props['value'] = value
+            props[key] = value
         return props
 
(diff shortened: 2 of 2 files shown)


netrampandey855-sketch:
Fire hack free fire hack

PR #8746, opened 2026-09-01
Make flaky tests more robust

philippjfr:
Mark some more tests as flaky.

### DIFF
--- a/panel/tests/util.py
+++ b/panel/tests/util.py
@@ -139,7 +139,7 @@ def check_layoutable_properties(layoutable, model):
     assert model.height_policy == 'min'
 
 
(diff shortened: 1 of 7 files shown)


panel/tests/util.py
```diff
@@ -139,7 +139,7 @@ def check_layoutable_properties(layoutable, model):
     assert model.height_policy == 'min'
 
 
-def wait_until(fn, page=None, timeout=5000, interval=100):
+def wait_until(fn, page=None, timeout=10000, interval=100):
```
hoxbro: Is this change really worth it? My feeling is if something is flaky it is not because we don't wait long enough but because of mistiming.

philippjfr: Agreed.

Issue #8326, opened 2025-12-03
Define migration plan for integrating panel-material-ui into core Panel


philippjfr:
### Overview

As Panel continues to evolve its UI and widget ecosystem, we want to define a clear, intentional migration path toward integrating the work in `panel-material-ui` into core Panel.

The Material UI widgets provide a more modern look and feel, improved consistency, and a cleaner API in several areas, but they are not 100% API-compatible with existing `panel.widgets`. Rather than forcing a sudden or breaking transition, we should introduce a structured, opt-in path that allows us to:

* Incrementally adopt the new widgets
* Preserve backwards compatibility
* Resolve any remaining API inconsistencies
* Align visual and behavioral aspects across the ecosystem

This issue proposes a staged migration plan centered around introducing a new `panel.v2` namespace (with `panel.ui` noted as a possible alternative).

### Goals

* Provide a clear, low-risk migration path
* Encourage adoption of modern widgets
* Improve long-term API coherence and visual consistency
* Give contributors and users a stable target for “future Panel” development

#### 1. Update examples and documentation

* Update existing examples to use panel-material-ui widgets
* Over time:
  * Make `panel.v2` the default in “Getting Started” and primary docs
  * Clearly label `panel.widgets` examples as legacy

#### 2. Introduce `label` ↔ `name` compatibility

To reduce friction and support shared mental models:

* Alias `label` and `name` for widgets where applicable
* Ensure both work consistently across:

  * Core Panel widgets
  * `panel-material-ui` widgets
  * `panel.v2` API
* Standardize on `label` in new APIs and documentation

#### 3. Introduce a `panel.v2` namespace (opt-in)

Create a new namespace that exposes the next-generation widget API backed by Material UI components:

```python
import panel.v2 as pn2

pn2.Select(label="Fruit", options=["Apple", "Banana"])
```

Key properties:

* `panel.v2` is **explicitly opt-in**
* No breaking changes to `panel.widgets`
* Documentation and examples can gradually move to `panel.v2`
* Allows API cleanup and alignment without destabilizing existing apps

> **Alternative:** `panel.ui`
> A `panel.ui` namespace could serve a similar purpose and may age better long-term than a versioned namespace. This remains an open design decision to be resolved during implementation.

#### 4. Address API inconsistencies incrementally

Use `panel.v2` as the place to clean up long-standing inconsistencies such as:

* Parameter naming (`name` vs `label`, etc.)
* Event semantics and update timing
* Validation and error display
[...]

PR #8752, opened 2026-09-05
Panel UI foundations

philippjfr:
Implements https://github.com/holoviz/panel/issues/8326

It makes `panel.ui` a real, public, complete component namespace built on a `panel-material-ui`
dependency:

```python
import panel as pn

pn.ui.Button(label='Click')          # Material, from panel-material-ui
pn.ui.widgets.Button(label='Click')  # same class, structured import
pn.ui.Tabulator(value=df)            # classic, no Material equivalent, no wrapper
```

Nothing is vendored here. The first commit teaches Panel's core the extension points a design system
needs, so that panel-material-ui no longer has to monkeypatch Panel to make its widgets show up
where Panel generates components on the user's behalf, no longer has to reimplement the loading
indicator because the CSS driven one is hardcoded to `config.loading_spinner`, and no longer has to
carry its own file watcher, base template and build hook. The second commit builds `panel/ui/` on top
of those seams as a composition layer: the Material components come from the dependency, the classic
components that have no Material equivalent are re-exported alongside them, and Panel 2.0 replaces
the imports with the moved implementation without changing a single name users write. Taking the
dependency first means the API decisions and the upstream cleanup in §6.3 can land and ship to real
users while the stakes are still low.

## Design resolution

`Design` gained a `component_mapping` and a `widget_mapping`, plus `resolve_component()` and
`resolve_widget()` classmethods, with module level `panel.theme.resolve_component()` and
`resolve_widget()` helpers that dispatch to the currently configured design. A design system can
now declare "when Panel would generate a `Select`, generate mine instead" without patching
anything.

The resolution is wired into every place Panel picks a component for the user:

- `Param.widget_type()` and `Param`'s `input_widgets` fallbacks
- `panel.widgets.widget()`, i.e. `widget_from_single_value`, `widget_from_tuple` and
  `widget_from_iterable`, which is what `interact` uses
- `HoloViews._resolve_widget()` for the widgets generated from HoloMap and DynamicMap dimensions

Explicit user choices always win. A `widget_type` passed through `Param.widgets`, a
`default_widgets` override on `HoloViews`, or a subclass of a mapped type is never substituted:
`resolve_component` only replaces exact matches, since a subclass may implement behavior the
design system's equivalent does not.

`widget_mapping` maps a `Parameter` type to a widget type or to a callable receiving the
`Parameter`, and takes precedence over `component_mapping` so a design system is free to split or
merge the classic mapping, e.g. resolving different widgets for `dict`, `list` and `tuple`
parameters where Panel resolves one.

## Design lookup by name and aliases

The string handling for `pn.extension(design=...)` moved out of `panel_extension` into
`panel.theme.resolve_design()`, and `config.design` now accepts a string directly through a
`_design_hook`, so `config.design = 'material'` works the same way `pn.extension(design='material')`
does.

`DESIGN_ALIASES` is a registry mapping a name to a `'module.path.ClassName'` reference, so designs
that do not live in `panel.theme` and designs whose class name differs from the name users
reference them by can be resolved, which is what `panel.ui` needs.

## `config.set_if_unset()`

Design systems and extensions want to declare the defaults they prefer without clobbering
anything the user set. `config.set_if_unset(**params)` applies a value only if the parameter is
still at its default and returns the names it applied.

## Loading indicator through the Design

The loading indicator was assembled inline from `config.loading_spinner`, `loading_color` and
`loading_max_height` in nine different places, so a design system could not change how it looks
without patching all of them. It is now a `Design` responsibility:

- `Design.loading_options()`, `loading_css_classes()`, `loading_css()` and `loading_resources()`
- a `_loading_options` class attribute for a design to declare its own spinner, color and max
  height, with explicit user config always taking precedence
- module level equivalents in `panel/io/loading.py` that fall back to the `Design` baseclass, which
  keeps the classic behavior, so calling code does not need to know whether a design is active

Callers updated to go through it: `panel/io/resources.py`, `panel/io/convert.py`,
`panel/io/document.py`, `panel/io/application.py`, `panel/io/pyodide.py`, `panel/io/server.py`,
`panel/links.py`, `panel/reactive.py`, `panel/template/base.py` and the editable template.
`start_loading_spinner`/`stop_loading_spinner` and the per component stylesheets honor the design a
component was actually rendered with over the globally configured one.

Two fixes fell out of this. The JS link callback for the `loading` property used
`if (!(css in css_classes))`, which tests array *indices* rather than values, so the loading classes
were appended on every toggle; it now uses `indexOf`. And the loading overlay color is no longer
hardcoded to `rgba(255, 255, 255, 0.5)`, it reads `--panel-loading-overlay-color` with a
`--panel-background-rgb` fallback so a dark design does not get a white flash.

A `material` spinner was added (`panel/dist/assets/material_spinner.svg`, a new
`loading_spinner` option) and `Material` declares it through `_loading_options`.

## `base.html` extension points

`panel/_templates/base.html` gained the blocks and variables a design system needs to render its
own loader and metadata, which is the merge of what panel-material-ui's forked template did:

- a `meta` block driven by a `meta` dict supporting `title`, `name`, `description`, `keywords`,
  `author`, `refresh` and `viewport`
- `loader`, `loader_css` and `loader_script` blocks so a design can render a loading screen before
  Bokeh boots
- `data-theme-managed` on the `<html>` element so the client can tell whether the design manages
  theming itself
- named `endblock`s, so subclasses can override individual blocks

The `preamble` block also had an inverted condition, `{% if apple_icon is not defined %}`, which
emitted `<link rel="apple-touch-icon" href="">` whenever no icon was set and never emitted the link
when one was. It now renders the icons when they are defined.

## Shared path watcher

`ReactiveESM` started one `watchfiles.awatch()` task per component instance, which is fine for a
handful of custom components but not for a design system where every widget in a shared bundle
watches the same file. `panel/io/watcher.py` adds a process wide `PathWatcher`, one per path, that
holds its subscribers weakly and notifies each by calling the method it subscribed with.
`ReactiveESM._setup_autoreload()` subscribes and `_cleanup()` unsubscribes.

A failing subscriber no longer tears down reloading for the others: the exception goes to
`config.exception_handler` when one is set and is logged otherwise. The stop event is registered on
`state._watch_events`, so the server stops these watchers on shutdown alongside its own.

## Build pipeline for the bundle

`hatch_build.py` gained `build_ui_bundle()`, which compiles the ESM components in `panel.ui` into
`panel/dist/ui/` via `panel.io.compile`, and can be skipped with `PANEL_LITE` or
`PANEL_SKIP_UI_BUNDLE`. Because `compile_components` can report a build failure on stdout while
still returning 0, the hook also checks that the artifact was actually written. It is a no-op while
there is no JSX under `panel/ui`, which is the state this PR leaves it in: keying the no-op on the
directory not existing would not do, because `find_module_bundles()` inspects a module's namespace
and would pick up the re-exported pmui components and try to rebuild their bundle into `panel/ui`.
It starts doing real work in Phase 3, when the JSX arrives.

No packaging or serving change was needed: `panel/dist` is already force included in the wheel and
sdist, and it is a registered extension directory, so the bundle is served from
`static/extensions/panel/ui/`.

Two pixi tasks make the development loop workable without a reinstall:

```
pixi run build-ui   # compile the bundle once
pixi run watch-ui   # recompile on every JSX change, unminified
```

`scripts/verify_build_size.py` gained a `ui` target with a 4 MB ceiling, skipped while the bundle
does not exist, and the pip build job in CI runs it.

## Layout of `panel/ui`

`panel/ui/` holds `__init__.py`, `base.py`, `theme.py`, `designs.py`, `notifications.py` and seven
component subpackages: `widgets`, `layout`, `pane`, `chat`, `indicators`, `template` and `wrappers`.
Every export is an explicit import with an `__all__`, no module `__getattr__` and no lazy tricks, so
Pyright, mypy and Jedi resolve `pn.ui.Button` to `panel_material_ui.widgets.Button` today and to
`panel.ui.widgets.Button` after the move. Both are resolvable classes, which was Marc's tooling
objection in the thread.

180 components are exported flat and each one is also reachable through the submodule it belongs to.
The lists were generated from runtime introspection of both namespaces rather than typed by hand,
filtering by `cls.__module__` so that pmui's leaky wildcard exports (`Column` and `Row` showing up in
`panel_material_ui.indicators`, `MaterialComponent` in `panel_material_ui.designs`) do not become
part of Panel's public API. Those leaks are fixed upstream in §6.3.

`panel.ui.base` re-exports the Material base classes (`MaterialComponent`, `MaterialUIComponent`,
`MaterialWidget`, `MaterialLayout`, `MaterialListLike`, `MaterialNamedListLike`, `MaterialPaneBase`)
so that writing a component that participates in the design does not require importing from
`panel_material_ui` directly.

## Namespace decisions

The flat namespace is the one place where two components can want the same name, so the curation is
explicit and the completeness test asserts against it rather than against a hand-maintained list:

- `Alert` is the Material `panel.ui.Alert`. The classic `panel.pane.Alert` is superseded and is not
  re-exported anywhere in `panel.ui`.
- `DataFrame` is the rendering pane, `panel.pane.DataFrame`. It was the one name two classic modules
  claimed with different classes, and it is resolved by deprecating the widget (below) rather than by
  picking a winner, so nothing about the resolution has to be remembered.
- `Spinner` is omitted, being a deprecated alias of `NumberInput`, and the `panel()` function is
  omitted, being reachable as `panel.panel`.
- Abstract bases (`Widget`, `ListLike`, `PaneBase`, `MaterialWidget`, ...) and helper classes
  (`Grammar`, `GrammarList`, `Utterance`, `Voice`) are submodule-only, which is what Panel's own flat
  namespace does.
- The submodule names are deliberately *not* in `panel.ui.__all__`, so that `from panel.ui import *`
  yields components only and does not rebind names like `widgets` or `layout` in the caller's
  namespace.

Everything else that is in one namespace is in the other under the same name. `pn.ui.Tabulator is
pn.widgets.Tabulator`, with no wrapper class, which also means `panel.widgets` and friends cannot be
deleted in 3.0; they stop being documented instead.

Where `panel.ui` and `panel` export different classes under the same name it is deliberate and
limited to the seven layouts pmui reimplements: `Accordion`, `Card`, `Column`, `Feed`, `FlexBox`,
`Row` and `Tabs`. Mixing them is supported, a Material layout takes classic children and vice versa.

## The `DataFrame` widget is deprecated

`panel.widgets.DataFrame` now warns on instantiation and points at `Tabulator`, with removal in 3.0.
Its docstring and reference page have said "will eventually be replaced with the `Tabulator` widget,
and so new code should be written to use `Tabulator` instead" since 1.0, and `Tabulator` has been the
better table in every respect for as long. Making that explicit is worth doing on its own, and it
also decides the only genuine name collision in `panel.ui` on the merits instead of by convention.

Two consequences. `Param` now maps `param.DataFrame` to `Tabulator` unconditionally, where it used to
return the deprecated widget unless `panel.models.tabulator` happened to be imported already, so
Panel no longer hands users a component it warns about. The conditional was there to avoid loading
Tabulator's bundle, which on-demand resource loading (#8745) has since taken care of. And the
namespace generator now raises on an ambiguous flat name rather than accepting an override, since
there is no longer an entry to justify the mechanism.

## Design selection

`config.design` is set to `MaterialUIDesign` when `panel.ui` is imported, but only if the user has
not chosen a design themselves, which is the §3.5 behavior `config.set_if_unset()` above exists for.
panel-material-ui 0.11.2 still assigns `config.design` unconditionally at import time, silently
overriding `pn.extension(design='fast')`, so `panel/ui/__init__.py` captures the design before
importing it and puts it back:

```python
_prior_design = _global_design()

from . import ...   # imports panel_material_ui

if _global_design() is not _prior_design:
    _set_global_design(_prior_design)

if config.design is None:
    _set_global_design(MaterialUIDesign)
```

The write goes through `param.Parameterized.__setattr__` rather than `config.design = ...` because
`config.__setattr__` scopes assignments to the current session, and `panel/ui/__init__.py` runs once.
Setting it per session would leave every session after the first on the classic design.

`panel.theme.base.DESIGN_ALIASES` gains `'material-ui'`, so `pn.extension(design='material-ui')` and
`config.design = 'material-ui'` resolve through the alias registry, importing `panel.ui` on the way.
The design is exported as `MaterialUIDesign`, the name pmui renames `MaterialDesign` to in §6.3, so
that panel's spelling is right from the start and distinct from the existing
`panel.theme.material.Material`.

## `pn.ui` is lazy

`panel/__init__.py` resolves `ui` through a PEP 562 module `__getattr__` with a `TYPE_CHECKING`
import, which Pyright handles correctly. The plan had it eager, but eager is not an option during the
depend stage: panel-material-ui imports panel, so `from . import ui` at the top of
`panel/__init__.py` is circular, and importing `panel.ui` selects the Material design, which every
Panel app would then get whether or not it asked for one. `ui` is also absent from `panel.__all__`,
so `from panel import *` does not pull it in. `import panel` still does not import
panel-material-ui, which `test_no_blocklist_imports` now asserts.

## Dependency pin

`panel-material-ui` moves from `>=0.10.0` to `>=0.11.2,<0.12` in `pyproject.toml` and `pixi.toml`.
The upper bound matters as much as the lower one: `panel.ui`'s public surface is versioned by another
repository for this cycle, and an open ceiling lets a pmui minor change Panel's documented API
without a Panel release. `panel/ui/__init__.py` carries the same bounds as `_PMUI_MIN_VERSION` and
`_PMUI_NEXT_VERSION` and checks them at import, raising `ImportError` below the floor so a stale
install fails with a sentence instead of a missing name, and warning above the ceiling. The floor
moves to whichever release carries the §6.3 closeout. The conda recipe and the runtime-versus-build
bootstrap noted in §6.1 are feedstock work and are not part of this PR.

## Tests

For the core foundations: `panel/tests/theme/test_resolve.py` (design and component resolution,
aliases, `set_if_unset`), `panel/tests/theme/test_loading.py` (loading options, CSS classes,
resources, inlining, and that the JS link and convert paths follow the design),
`panel/tests/io/test_base_template.py` (rendered template output for icons, metadata, the new blocks,
and that a template extending `base.html` can override them) and `panel/tests/io/test_watcher.py`
(`PathWatcher` sharing, weak subscribers, error isolation, and that two ESM components sharing a file
share one watcher).

For the namespace, `panel/tests/ui_components/test_namespace.py` is the file that keeps "`panel.ui` is
complete" true, a full release before the move makes a gap expensive to fix. It asserts that every
public name in `panel.widgets`, `panel.layout`, `panel.pane`, `panel.chat` and `panel.indicators` is
reachable both from the matching `panel.ui` submodule and from the flat namespace, modulo the
omissions above, and the same for every component panel-material-ui itself defines. On top of that it
checks that flat and structured forms are the same objects, that no flat name resolves to different
classes in different submodules, that classic components are re-exported and not wrapped, that
`__all__` is sorted and complete, that no module in `panel.ui` defines a `__getattr__`, that the
`'material-ui'` alias resolves, and, in subprocesses, that importing `panel.ui` selects the Material
design, that an explicit design survives it, and that it changes no other config parameter.

`panel/tests/widgets/test_tables.py` gains `test_dataframe_widget_is_deprecated` and a module-level
`filterwarnings` mark, since the rest of the file has to keep testing the widget until it is removed.

Verified separately, not committed: instantiating and rendering every non-abstract component in
`panel.ui.__all__` to a `Document` succeeds for 169 of 179. The ten that fail are classic components
that need constructor arguments (`DiscreteSlider`, `DatetimeRangeInput`, `ParamMethod`) or an
uninstalled optional dependency (`Reacton`, `RGGPlot`), and they fail the same way through
`panel.pane` and `panel.widgets`. The roundtrip render test in §9 is Phase 3 work.

## Notes for review

panel-material-ui is a hard dependency of Panel on this branch, and importing it mutates
`Param.mapping`, `Param.input_widgets` and `HoloViews.default_widgets` in place, reassigns
`panel.io.resources.BASE_TEMPLATE` and `panel.io.convert.BASE_TEMPLATE`, and replaces
`panel.io.convert.loading_resources` with a lambda. The tests are written to be hermetic under that:

- The resolution tests derive the expected substitution from whatever the classic mapping currently
  resolves to, rather than hardcoding Panel's own widget classes, so they test the wiring and not
  the contents of a mutable class attribute.
- The template tests resolve `base.html` by name through the Jinja environment rather than through
  the `BASE_TEMPLATE` symbol.
- `test_convert_loading_resources` skips itself when `panel.io.convert.loading_resources` is not
  Panel's own function. The skip is the signal that pmui is still patching; it starts running again
  when §6.3 removes the patch.
- `test_namespace.py` imports `panel.ui` through a module-scoped fixture rather than at module scope,
  and the fixture undoes what the import did. Importing it during collection would pull
  panel-material-ui into the process before the rest of the suite is collected, and the tests that
  parametrise over `Viewable.__subclasses__()` would silently start covering third party components.
  Undoing it matters just as much: the `Param.mapping` patch leaks into every test that runs
  afterwards, and one consequence is that `Widget.controls(jslink=True)` raises for every classic
  widget with an `Action` or `Event` parameter, because the Material `Button` it now substitutes has a
  non-jslinkable `value`. Without the restore,
  `panel/tests/widgets/test_base.py::test_widget_linkable_params[Button]` fails whenever
  pytest-randomly happens to order this module first.

`test_importing_panel_ui_does_not_patch_core` asserts that the fixture found nothing to restore. It is
`xfail(strict=False)` today and flips to passing when §6.3 deletes those blocks, at which point the
restore can go too. That makes it the acceptance check for that work rather than a test that has to be
remembered and written later.

I compared full suite runs before and after (`panel/tests`, excluding the Playwright suite). The only
differences are the new tests going from failing to passing, plus two flaky filesystem and network
dependent tests that pass on rerun (`test_reload_on_update`, `test_loading_a_image_from_url`,
`test_init_avatar_bytes`). The 170-odd failures common to both runs are pre-existing on this branch in
an environment with pmui and Lumen installed, mostly Panel's generic widget and docstring tests
parametrizing over third party components.

## Not in this PR

The docs landing page for `panel.ui`, the classic-component styling audit under the Material design
(§6.5), and the upstream closeout in panel-material-ui (§6.3) are separate items. Per §6.4 the bundle,
its CDN root and the reference docs stay panel-material-ui's for now, which is the version skew the
pin and the import-time check exist to contain.

## AI Disclosure

Created with heavy assistance from Claude Opus 5


panel/reactive.py
```diff
@@ -36,11 +36,12 @@
 )
 
 from .io.document import hold, unlocked
+from .io.loading import loading_css
 from .io.notebook import push
 from .io.resource_spec import lazy_load_available, resource_spec
 from .io.resources import (
     CDN_DIST, get_dist_path, loading_css, patch_stylesheet, process_raw_css,
```
philippjfr:
```suggestion
    CDN_DIST, get_dist_path, patch_stylesheet, process_raw_css,
```

PR #8755, opened 2026-09-07
Yield to the event loop when a periodic callback overruns its period

### LINKED ISSUE #8751: PeriodicCallback starves the event loop when a callback is slower than its period

```
panel  == 1.9.4
bokeh  == 3.9.2
python == 3.12.3
Linux (also reaches users via Pyodide/WebAssembly)
```

Still present on `main` today: `panel/io/callbacks.py` line 154.

`PeriodicCallback` yields to the event loop **only when a callback finishes inside its period**. A synchronous callback that returns promptly on every call, but simply takes longer than the period, starves the event loop completely for as long as that lasts.

`_async_repeat` is:

```python
while True:
    start = time.monotonic()
    await func()
    timeout = (self.period/1000.) - (time.monotonic()-start)
    if timeout > 0:
        await asyncio.sleep(timeout)
```

`await func()` on a coroutine is not itself a suspension point, and `_periodic_callback` only awaits when the callback is awaitable, so for a synchronous callback that conditional sleep is the loop's **only** yield. When a frame overruns the period it is skipped, and nothing else on the loop runs at all.

I would expect an overrunning callback to make the app *slow*. It makes it *unresponsive*, which is a different thing: events are not queued and delivered late, they are never processed.

The docstring documents that an overrunning callback "will run again immediately", so overruns are anticipated. What seems unintended is that "immediately" means without ever returning to the event loop.

This is not specific to threading configuration: `config.nthreads` defaults to `None`, so `state._thread_pool` is `None` and a synchronous callback runs inline on the event loop by default.

```python
import asyncio
import time

import panel as pn

PERIOD_MS = 33
FRAMES = 20


def busy(seconds):
    """A callback that returns promptly every call -- it just takes `seconds`."""
    def step():
        end = time.monotonic() + seconds
        while time.monotonic() < end:
            pass
    return step


async def measure(frame_s):
    """How many times can another asyncio task run while the callback loops?"""
    ran = 0

    async def other_task():
        nonlocal ran
        while True:
            ran += 1
            await asyncio.sleep(0)

    task = asyncio.ensure_future(other_task())
    cb = pn.state.add_periodic_callback(
        busy(frame_s), period=PERIOD_MS, count=FRAMES, start=True)
    started = time.monotonic()
    while cb.running and time.monotonic() - started < 30:
        await asyncio.sleep(0)
    task.cancel()
    return ran


async def main():
    print("period = %d ms, %d frames\n" % (PERIOD_MS, FRAMES))
    for frame_s in (0.020, 0.050):
        ran = await measure(frame_s)
        print("  frame %3d ms (%s period): other task ran %8d times"
              % (frame_s * 1000,
                 "under" if frame_s * 1000 < PERIOD_MS else "over", ran))


asyncio.run(main())
```

Output:

```
period = 33 ms, 20 frames

  frame  20 ms (under period): other task ran    66078 times
  frame  50 ms (over period): other task ran        1 times
```

The exact count varies between runs; the contrast does not. It is a cliff at
the period, not a gradient. Measured against the same loop shape at a 33 ms period: a competing task was scheduled 48734 times on 30 ms frames, and **zero** times on 34 ms and 70 ms frames.

#### Where this bites in practice

In `pyodide-worker` mode the worker's `onmessage` -- which applies widget patches -- is a JS task, so it cannot run either. A slider dragged during an animation has no effect, and Pause does not take effect, until some frame happens to come in under the period. Because frame cost is usually uneven, that presents to a user as controls freezing *sometimes*.

Measured in a browser on an interactive model at a setting whose frames run ~85 ms against a 33 ms period: three interleaved A/B repetitions, all three unresponsive with a synchronous callback (no response in 30 s, Pause ineffective) and all three responsive with an `async` one (0.3-0.9 s latency, Pause immediate), with throughput unchanged.

#### Not the same as #7578

[#7578](https://github.com/holoviz/panel/issues/7578) was an infinite `while True` inside the callback, correctly closed as user error. Here the callback returns on every call; it is only slower than its period.

That thread is relevant in one way, though. The advice there was that `time.sleep(0.5)` at `period=1000` still lets a stop call through -- which is true, but only because 500 ms is *under* the period, so the conditional sleep fires. At `sleep(1.5)` it would not.

#### Suggested fix

Yield unconditionally, so the loop stays cooperative even when a callback overran:

```python
await asyncio.sleep(timeout if timeout > 0 else 0)
```

This preserves the documented "run again immediately" behaviour -- a zero-length sleep does not delay the next frame measurably -- while giving the event loop one turn per frame, which is enough to deliver pending events.

A user-side workaround, for anyone who finds this issue: make the callback a coroutine ending in `await asyncio.sleep(0)`. `_periodic_callback` awaits an awaitable callback, so the yield then happens inside the frame.

ebarkhordar:
`asyncio.sleep(0)` is still a yield, so dropping the conditional keeps the documented "run again immediately" behaviour and gives the loop a turn on every iteration:

```python
timeout = (self.period/1000.) - (time.monotonic()-start)
await asyncio.sleep(max(timeout, 0))
```

`asyncio.sleep` already routes `delay <= 0` to a bare yield, so the old `if timeout > 0` guard was not avoiding a sleep, it was only removing the yield.

One scope note: this is the `_async_repeat` branch only. Session-scoped callbacks in a served app go to `doc.add_periodic_callback` instead, and I have not checked whether that path behaves the same way.

Your measure() script goes from 1 to 20 other-task runs on the 50 ms frame for me. Added two tests in `panel/tests/io/test_callbacks.py`; the overrunning one fails on main.

Closes #8751

### DIFF
--- a/panel/io/callbacks.py
+++ b/panel/io/callbacks.py
@@ -151,8 +151,9 @@ class PeriodicCallback(param.Parameterized):
             start = time.monotonic()
             await func()
             timeout = (self.period/1000.) - (time.monotonic()-start)
-            if timeout > 0:
-                await asyncio.sleep(timeout)
+            # sleep(0) still yields, so an overrunning callback runs again
+            # immediately without holding the loop for the next iteration.
+            await asyncio.sleep(max(timeout, 0))
 
     def _cleanup(self, session_context):
         self.stop()
--- a/panel/tests/io/test_callbacks.py
+++ b/panel/tests/io/test_callbacks.py
@@ -1,0 +1,72 @@
+"""Tests for PeriodicCallback's loop-driven path.
+
+`_async_repeat` is the branch `start()` takes when there is a running event loop but
+no bokeh session: Pyodide and WASM, notebooks, and scripts. Session-scoped callbacks
+go to `doc.add_periodic_callback` instead and are covered in `panel/tests/test_server.py`.
+"""
+
+import asyncio
+import time
+
+import pytest
+
+from panel.io.callbacks import PeriodicCallback
+
+
+class _Stop(Exception):
+    """Raised by a test callback to end the otherwise infinite `_async_repeat` loop."""
+
+
+async def _drive(period, frames, body):
+    """Run `_async_repeat` for `frames` iterations of `body`, counting loop turns.
+
+    Returns (number of callback invocations, number of times a competing task ran).
+    A competing task can only advance when `_async_repeat` suspends, so the second
+    number is how many times the loop handed control back to the scheduler.
+    """
+    turns = 0
+
+    async def competing():
+        nonlocal turns
+        while True:
+            turns += 1
+            await asyncio.sleep(0)
+
+    calls = 0
+
+    async def callback():
+        nonlocal calls
+        calls += 1
+        body()
+        if calls >= frames:
+            raise _Stop
+
+    competitor = asyncio.create_task(competing())
+    await asyncio.sleep(0)
+
+    cb = PeriodicCallback(callback=callback, period=period)
+    try:
+        with pytest.raises(_Stop):
+            await cb._async_repeat(callback)
+    finally:
+        competitor.cancel()
+    return calls, turns
+
+
+async def test_overrunning_callback_still_yields_to_the_loop():
+    frames = 5
+    # 5 ms of work against a 1 ms period, so every frame overruns and the remaining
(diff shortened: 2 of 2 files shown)


philippjfr (approved):

philippjfr:
Thank you @ebarkhordar!

PR #8757, opened 2026-09-08
Keep the count and timeout budgets across a period change

### LINKED ISSUE #8756: Changing period on a running PeriodicCallback restarts its count and timeout budgets

<details>

```plaintext
panel 1.9.4 (wheel) and main at 5b27854
bokeh 3.9.2 with the wheel, 3.10.0 against main
Python 3.12.14, Linux, python:3.12-slim container
```
</details>

`add_periodic_callback(..., count=6)` should invoke the callback six times. If `period` is changed while it is running, it invokes it six more, so ten in the example below.

`_update_period` does a period change as `stop()` then `start()`. `stop()` sets `counter` back to 0 and `start()` re-stamps `_start_time`, so both budgets start over. `cb.counter` also drops to 0 mid-run, which anyone reading it will see.

`timeout` goes the same way. With `timeout=200` and the period changed at 150 ms the callback runs on to about 370 ms instead of stopping at 200.

Different route to the same symptom as #5339, which was the threaded race producing runs past `count`.

```python
import asyncio
import panel as pn

async def main():
    calls = 0

    def body():
        nonlocal calls
        calls += 1

    cb = pn.state.add_periodic_callback(body, period=20, count=6)
    await asyncio.sleep(0.07)
    print("counter before:", cb.counter)   # 4
    cb.period = 21
    print("counter after: ", cb.counter)   # 0
    while cb.running:
        await asyncio.sleep(0.02)
    print("invocations:", calls)           # 10, expected 6

asyncio.run(main())
```

Nothing raises.

Not applicable, no UI involved.

ebarkhordar:
## Description

Fixes #8756

`_update_period` restarts the callback to pick up the new interval, and `stop()`/`start()` also reset the two budgets on the way through: `stop()` zeroes `counter`, `start()` re-stamps `_start_time`. So a period change gives a `count=6` callback six more runs and pushes a `timeout` out by however long it had already been running.

The reset in `stop()` is wanted when you actually stop, so I carried `counter` and `_start_time` across the internal restart instead of changing `stop()`. `counter` is restored under `param.discard_events` so watchers do not see it bounce.

Repro is in the issue. Two tests in `panel/tests/io/test_callbacks.py`, one per budget; both fail on main (`assert 9 == 6` and `assert 0.815 < 0.7`) and pass here. `panel/tests/io` goes 304 to 306 passing with no new failures.

The timeout test asserts on elapsed wall time with a 200 ms margin either side, so tell me if you would rather it did not sit in the matrix.

## AI Disclosure

Usage: helped locate the reset, draft the patch and write the two tests. I reproduced the fault and ran the tests myself.

## Checklist

### DIFF
--- a/panel/io/callbacks.py
+++ b/panel/io/callbacks.py
@@ -80,8 +80,14 @@ class PeriodicCallback(param.Parameterized):
     @param.depends('period', watch=True)
     def _update_period(self):
         if self._cb:
+            # stop() zeroes the counter and start() re-stamps the start time,
+            # so both are carried across the restart.
+            counter, start_time = self.counter, self._start_time
             self.stop()
             self.start()
+            with param.discard_events(self):
+                self.counter = counter
+            self._start_time = start_time
 
     def _exec_callback(self, post=False, busy_event_id=None):
         try:
--- a/panel/tests/io/test_callbacks.py
+++ b/panel/tests/io/test_callbacks.py
@@ -63,6 +63,50 @@ async def test_overrunning_callback_still_yields_to_the_loop():
     assert turns >= frames
 
 
+async def _run_until_stopped(cb, limit=400):
+    for _ in range(limit):
+        if not cb.running:
+            return
+        await asyncio.sleep(0.01)
+
+
+async def test_period_change_keeps_the_count_budget():
+    calls = 0
+
+    def body():
+        nonlocal calls
+        calls += 1
+
+    cb = PeriodicCallback(callback=body, period=20, count=6)
+    cb.start()
+    try:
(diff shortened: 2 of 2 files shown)


philippjfr (approved):

philippjfr:
Thank you @ebarkhordar!

ebarkhordar:
The timeout test is the part I would watch: it asserts elapsed wall time under 700 ms against a 500 ms deadline, so there is only 200 ms of headroom on a slow runner. If it does start going red there, I would rather drive the clock in the test than widen the bound.

PR #8764, opened 2026-09-15
fix: Rerender `Matplotlib` and `RGGPlot` panes once per parameter change

hoxbro:
## Description

 `PaneBase` no longer registers a watcher twice for a parameter listed twice. The removals of `_rerender_params` are because they are duplicates. 

Maybe this should be upstreamed directly to param? I can't think of a reason where this would make sense.

## AI Disclosure

Tool & Model: Claude:claude-opus-5
Usage: This is part of a long running session which tires to speed up docs build of HoloViews. 

## Checklist

- [ ] Added documentation


hoxbro:
Upstreamed it

PR #8765, opened 2026-09-15
Fix: Fix links to tabler icons site

ATL2001:
## Description
I noticed the links to the tabler icons were broken in the [FileDownload widget](https://panel.holoviz.org/reference/widgets/FileDownload.html#icons) page, but then when I went to fix it there, I noticed the same link existed in a handful of other pages.  It looks like the website moved to a new address. this PR just updates the links.


philippjfr:
Thank you!

PR #8766, opened 2026-09-18
fix: Fix the root causes of flaky tests

hoxbro:
## Description

Fixes the causes of flaky CI instead of rerunning them. Most were real bugs that show up on slow machines or networks.

**Fixes**
- `Feed`: scroll to latest lost when stylesheets load late, when a visibility report arrives first, or when the latest children were still being built; a `Feed` outside the scroll limit now stays where it is
- `Column`: initial scroll and auto-scroll applied before children or stylesheets are ready
- `Tabulator`: scroll position lost on data changes and redraws; header filter list closed by a redraw; data that arrived while the table was building was dropped, leaving it stale
- `Plotly`: moving straight to another point reported the previous point on hover and click
- `Card`: child left hidden by an echoed `visible`
- Changes made before `document_ready` dropped, unsent models declared synced; a document without a session is no longer taken for one a client has
- Threaded server shutdown stranding callbacks and skipping Panel's cleanup; a cancelled autoreload watcher no longer keeps the server up
- Notebook: components rendered before their module globals were assigned (e.g. `FileDropper`)
- Stuck flags that could wedge a view: a `Feed` that never becomes scrollable, a leftover hidden Tabulator filter list, and `auto_scroll_limit = 0`

**Tests**
- Test servers bound and loaded on `127.0.0.1`; wider ephemeral port range on Windows CI
- Waits on real state (stylesheets, server data, laid-out cells, kernel ready) instead of sleeps and fixed coordinates
- Requests to test servers time out, so a server that stops answering fails its test instead of the job
- Tests that download remote files carry the `internet` marker, which now also grants reruns; a notebook cell that stops making progress times out after a minute and is rerun, along with the cells that depend on it
- Fixtures no longer leak event loops, thread pools, `panel serve` children or nbclassic kernels between tests
- `flaky` markers removed from fixed tests. Still marked: jupyter, jupyterlite, reload, convert, nbclassic, and `test_autoreload_app_local_module` (autoreload misses modules imported after the watcher starts, tracked by an xfail test)

## Follow-ups

- autoreload misses modules imported after the watcher starts, tracked by an xfail test
- The example notebooks are network-bound (18 fetch remote data); the cell timeout and reruns above keep a reset connection from hanging the job, but the run is still only as reliable as the network
- Panel's admin log handler streams every log line into a Tabulator and reprocesses the whole table each time, which slows a server thread down as a run progresses; probably its own issue
- macOS stalls at the end of a UI run when VTK renders a window off the main thread; worked around here, the deadlock itself is unfixed

## AI Disclosure

Usage: Reproduced the flaky tests under CI sized limits, diagnosed them with instrumentation, and wrote the fixes and tests.

## Checklist


.github/workflows/test.yaml
```diff
@@ -225,11 +225,8 @@ jobs:
           echo "PANEL_TEST_AUTH=1" >> $GITHUB_ENV
       - name: Test UI
         run: |
-          # Create a .uicoveragerc file to set the concurrency library to greenlet
-          # https://github.com/microsoft/playwright-python/issues/313
-          echo "[run]\nconcurrency = greenlet" > .uicoveragerc
```
hoxbro, line 230: This never worked... `\n` would not make a new line, so removed it

panel/tests/conftest.py
```diff
@@ -190,6 +191,10 @@ def pytest_collection_modifyitems(config, items):
         else:
             skipped.append(item)
 
+    for item in selected:
+        if item.get_closest_marker("internet") and not item.get_closest_marker("flaky"):
+            item.add_marker(pytest.mark.flaky(reruns=3, reason="Downloads remote files, which can fail on a bad connection"))
```
hoxbro, line 201: I added reruns to internet markers as they would sometime fail because of bad internet in the CI. 

A small script was created to see which test was going out on the internet.

hoxbro:
Likely there is still some flaky test, but will suggest we make a list of them over the coming weeks and then do a follow up PR.

philippjfr (approved):

PR #8767, opened 2026-09-21
compat: Python 3.15 (rc)

hoxbro:
## Description

Add test for Python 3.15


.github/workflows/test.yaml
```diff
@@ -162,7 +164,7 @@ jobs:
         run: |
           pixi run -e "${ENV}" test-unit $COV
       - name: Test subprocess
-        if: contains(matrix.os, 'ubuntu')
+        if: contains(matrix.os, 'ubuntu') && matrix.environment != 'test-315'
```
hoxbro, line 167: Getting this which seems like we should investigate at some point:

Image

